PR #1166, opened 2026-04-02
Add the Skip softmax for diffusion

jingyu-ml:
### What does this PR do?

Type of change: new feature, new example  

## Summary

- Add skip-softmax sparse attention (BLASST) for diffusion models via dedicated Triton kernels — an inference kernel with tile skipping and a calibration kernel with vectorized multi-threshold sparsity measurement
- Add `triton_skip_softmax` method with exponential model calibration (`scale_factor = a * exp(b * sparsity)`) and log-space fitting for diffusion models
- Add Triton kernel backends for diffusers and LTX attention dispatch
- Fix calibration to skip RULER dataset generation when user provides their own `forward_loop` (required for non-LLM models)

### Triton kernels (`modelopt/torch/kernels/triton_fa.py`)
- **`_attn_fwd`**: Forward kernel with optional tile skipping — tiles whose max attention score is far below the running softmax max are skipped entirely (no V load, no softmax, no accumulation). Runtime sparsity measurement via atomic counters.
- **`_attn_fwd_calibrate`**: Calibration kernel that computes full attention while measuring how many tiles would be skipped at each of N thresholds simultaneously. Uses per-program output buffers (zero atomic contention) and vectorized multi-threshold comparison.
- **`attention()`** / **`attention_calibrate()`**: Python wrappers for inference and calibration kernels.

### Kernel backends (`modelopt/torch/sparsity/attention_sparsity/kernels/`)
- **`diffusers_triton_attention.py`**: Registers `modelopt_triton` backend in diffusers' attention dispatch. Handles [B, S, H, D] → varlen layout conversion, calibration/inference mode switching, thread-local configuration, and counter accumulation.
- **`ltx_triton_attention.py`**: Patches `ltx_core.Attention` modules for Triton dispatch with the same calibration/inference modes.

### Method (`modelopt/torch/sparsity/attention_sparsity/methods/triton_skip_softmax.py`)
- `TritonSkipSoftmaxMethod`: Context managers for calibration (→ calibration kernel) and inference (→ forward kernel with tile skipping). Three threshold priority levels: raw threshold > calibrated scale_factor > static threshold.

### Calibration (`modelopt/torch/sparsity/attention_sparsity/calibration/`)
- **`calibrator.py`**: `DynamicThresholdCalibrator` with `fit_logspace` option — fits exponential model in log space (minimizes relative error) for diffusion models where scale_factors span many orders of magnitude. Records observed sparsity range for extrapolation warnings.
[...]

PR #1227, opened 2026-04-09
[2/3] Implicit Gemm NVFP4

jingyu-ml:
### What does this PR do?

Type of change: new feature 

- Add Conv3D implicit GEMM kernel with BF16 WMMA tensor cores and fused NVFP4 activation quantization for video diffusion VAE layers
- Integrate into _QuantConv3d via QuantModuleRegistry — automatically dispatched when NVFP4 quantization is applied to nn.Conv3d
- Move kernel from `experimental/conv/ to modelopt/torch/kernels/conv/`; move tests to `tests/gpu/torch/quantization/kernels/`

### Testing

- Added test cases to measure the difference between cuDNN and our CUDA implicit GEMM kernel
- Added an NVFP4 fake quantization test using CUDA code

## Summary by CodeRabbit

* **New Features**
  * Per-backbone quantization/export in a single run with per-backbone checkpoints and backbone-aware quant filters
  * Configurable NVFP4 block-size via CLI/config; improved NVFP4 Conv3D inference path and Wan 2.2 quantization support
* **Bug Fixes**
  * Video-model calibration now respects extra params and forces video decoding during calibration
* **Documentation**
  * Added comprehensive Conv3D implicit‑GEMM kernel documentation; removed experimental Conv3D prototype docs/benchmark
* **Tests**
  * New Wan 2.2 quantization/export tests and expanded Conv3D/FP4 kernel test coverage


sychen52 (approved):

Edwardf0t1 (commented):
## Review

### Overview
Promotes the Conv3D implicit GEMM CUDA kernel (BF16 WMMA + fused NVFP4 activation quant) from `experimental/` to `modelopt/torch/kernels/conv/`, wires it into `_QuantConv3d` via `QuantModuleRegistry` so NVFP4-configured `nn.Conv3d` layers dispatch automatically, and adds `WanCausalConv3d` support for diffusers VAE. Also overhauls the diffusers quantization example for per-backbone (transformer + VAE) quantize/restore/export, and ships a large batch of correctness tests (+1121/-503, CI fully green).

### Strengths
- **Excellent kernel test coverage**: input validation, dtype preservation (FP16/BF16), non-contiguous inputs, combined conv params, FP4 advanced paths, zero/degenerate inputs, numerical stability, and exact tile-boundary tests. Monotonicity simplified sensibly.
- Clean calibration→inference split in `_QuantConv3d.forward`: falls back to cuDNN during `_if_calib and not _if_quant` for speed.
- End-to-end Wan 2.2 tests exercise quantize + restore + HF export for INT8/FP8/FP4.
- Per-backbone iteration (`iter_backbones`) + per-backbone checkpoints (`{name}.pt`) is a nice generalization over the single-backbone assumption.
- README is honest about limitations (inference-only, BF16 rounding error scales with √K, groups=1).

### Issues

**Correctness / risk**
- **QAT / training mode not guarded**: `_should_use_implicit_gemm()` returns True as long as both quantizers are NVFP4 and the kernel imports. The README says the kernel is inference-only (`@torch.no_grad`), but nothing in the dispatch checks `self.training`. A user who runs QAT on an NVFP4 Conv3D will silently enter the implicit-GEMM path and lose gradients. Recommend gating on `self.training is False` (or `not torch.is_grad_enabled()`).
- **Stale dispatch cache**: `_should_use_implicit_gemm` memoizes the answer on `self._use_implicit_gemm`. If a user later toggles a quantizer (disable/enable, change block_size) on a live module, the cache is wrong. Either drop the cache or invalidate it in a quantizer setter hook.
- **`check_conv_and_mha` behavior change**: previously `Conv3d` was auto-disabled for FP4; now it is not. Combined with the new NVFP4 Conv3D path, this silently *enables* FP4 quantization on Conv3D layers inside non-VAE backbones for any existing user workflow that relied on the old behavior. Worth calling out in the PR description / CHANGELOG more explicitly than "new feature."
- **Weight-quantizer block layout assumption**: `_nvfp4_quantize_weight_along_k` flattens weight to `[Cout, Cin*kD*kH*kW]` and passes it to `weight_quantizer`. This only produces correct blocks if the quantizer's `block_sizes` maps `-1: N` (block on last axis). `_is_nvfp4_quantizer` doesn't verify that — if a user configures block_sizes on a named axis (e.g., `block_sizes={2: 16}` for the Conv3D K axis), the guard passes, the flatten silently mis-aligns blocks, and results drift. Add an assertion or make the check stricter.

**Private-API coupling**
- `self.input_quantizer._get_amax(input)`, `self.input_quantizer._block_sizes.get(-1, 16)`, `self.input_quantizer._if_calib`, `self.input_quantizer._if_quant` — four separate underscore-prefixed attributes. `_is_nvfp4_quantizer` uses the public `block_sizes`, so the naming is inconsistent. Please expose public accessors or at least document why these private fields are load-bearing.

**`_QuantDiffusersWanCausalConv3d`**
- Duplicates the forward dispatch logic from `_QuantConv3d.forward` (the `_should_use_implicit_gemm + _if_calib/_if_quant` gate). If `_QuantConv3d.forward` ever changes (e.g., the QAT gate above), the child silently diverges. Factor the gate into a shared helper.
- `torch.nn.Conv3d.forward(self, input)` is used to bypass the Wan-specific causal padding — works today but depends on WanCausalConv3d's internal decision to set `self.padding=(0,0,0)`. A comment explaining that dependency would help reviewers.
- `except ImportError: pass` around the whole registration silently hides any non-ImportError bugs in registration (e.g., a `QuantModuleRegistry.register` AttributeError).

**Example script**
- The `forward_loop` closure in `main()` is defined inside the `for backbone_name, backbone in …` loop. The closure captures nothing backbone-specific, so it happens to work, but a future edit that references `backbone` inside the lambda would silently late-bind to the last value. Either define it outside or add a comment.
- Per-backbone quant config is re-derived each iteration via `get_quant_config(...)`, but `copy.deepcopy` inside only runs when `block_size != 16`. When block_size is 16 (default), the same dict is shared across backbones; if `quantize_model` mutates it (e.g., via `mtq.quantize` post-processing), later backbones see the mutation. Deepcopy unconditionally, or confirm `mtq.quantize` doesn't mutate.

**Tests**
- No end-to-end test for the NVFP4 implicit-GEMM integration in `_QuantConv3d` — all kernel tests call `cuda_conv3d` directly; there is no test that exercises `_QuantConv3d.forward → _implicit_gemm_forward` and compares against the default cuDNN path. Easy miss for the dispatch guard logic and the weight-flatten path.
- `test_wan22_int8_smoothquant` is not GPU-gated; fine if CI's Linux runner has CUDA, but worth confirming.
- Integration tests depend on `_test_utils.torch.diffusers_models.create_tiny_wan22_pipeline_dir`, which I believe was introduced in PR #1166 — if this lands first, the tests will fail to import.

**Minor**
- Deleted `experimental/conv/bench_implicit_gemm.py` removes the only perf-reporting tool; consider moving it to `tools/` or `benchmarks/` instead of dropping it.
- `copy` imported inside `get_quant_config` — hoist to module-level.

### Recommendation
**Approve with changes.** The kernel move + test expansion is solid. Before merge please add: (1) QAT/training-mode guard, (2) stricter NVFP4 block-layout check (or doc assumption), (3) at least one end-to-end test that goes through `_QuantConv3d.forward` with the implicit-GEMM path, and (4) shared dispatch helper between `_QuantConv3d` and `_QuantDiffusersWanCausalConv3d`. The private-API coupling is a smell but not a blocker.

Edwardf0t1 (approved):
LGTM, please check AI's reviews.

modelopt/torch/quantization/nn/modules/quant_conv.py
```diff
@@ -62,12 +64,87 @@ class QuantConv2d(_LegacyQuantLinearConvBaseMixin, nn.Conv2d):
...
+        return self.output_quantizer(output)
+
+    def forward(self, input, *args, **kwargs):
+        """Forward with implicit GEMM for NVFP4, default path otherwise."""
+        if not self._should_use_implicit_gemm():
+            return super().forward(input, *args, **kwargs)
+
+        if self.training:
+            warnings.warn(
+                "Implicit GEMM Conv3D kernel is inference-only and does not support training. "
```
mxinO, line 133: For NVFP4 (or block-wise quant), Is there a reason to fall back the general gemm quantization path? If it can't be used in inference, there is no reason to keep that path, then we should raise an error instead if "NVFP4 Weight & input + training" is using, or the fallback should be im2col, but that cost is too high I assume.

mxinO (commented):
I noticed the kernel is now putting at a top-level folder `torch/kernels`, our previous kernel is putting at subfolders, e.g. `quantization/src`. This folder seems from @kaix-nv 's attention kernel. Did we get aligned about this? For me it's more clear that putting quantization kernels in quantization folder and sparse kernels in sparse folder. cc @kevalmorabia97 @realAsma

shengliangxu:
> I noticed the kernel is now putting at a top-level folder `torch/kernels`, our previous kernel is putting at subfolders, e.g. `quantization/src`. This folder seems from @kaix-nv 's attention kernel. Did we get aligned about this? For me it's more clear that putting quantization kernels in quantization folder and sparse kernels in sparse folder. cc @kevalmorabia97 @realAsma

I actually like the kernel being in their own top level dir. It's more clear overall architecture wise

kevalmorabia97:
> > I noticed the kernel is now putting at a top-level folder `torch/kernels`, our previous kernel is putting at subfolders, e.g. `quantization/src`. This folder seems from @kaix-nv 's attention kernel. Did we get aligned about this? For me it's more clear that putting quantization kernels in quantization folder and sparse kernels in sparse folder. cc @kevalmorabia97 @realAsma
> 
> I actually like the kernel being in their own top level dir. It's more clear overall architecture wise

I personally would like algorithm specific kernels in their subfolders and general kernels at top level so top-level is clean and someone working on say quantization knows what kernels they need to look at and not be worried about other unrelated kernels

mxinO:
> I actually like the kernel being in their own top level dir. It's more clear overall architecture wise.
Even we have a top-level folder, we should have a clear subfolder structure, e.g. 
```
kernels/
    common
    quanitzation
         conv
         gemm
    sparisity
         atttention
         gemm
```
If we choose to have top-level kernel folder, then we need move previous kernels here.

kevalmorabia97 (approved):

modelopt/torch/quantization/nn/modules/quant_conv.py
jingyu-ml: Resolved, thanks! Go ahead and remove the cuDNN path.

jingyu-ml:
> > I actually like the kernel being in their own top level dir. It's more clear overall architecture wise.
> > Even we have a top-level folder, we should have a clear subfolder structure, e.g.
> 
> ```
> kernels/
>     common
>     quanitzation
>          conv
>          gemm
>     sparisity
>          atttention
>          gemm
> ```
> 
> If we choose to have top-level kernel folder, then we need move previous kernels here.

This make more sense to me, I will change the kernel structure in the next PR

PR #1235, opened 2026-04-10
[OMNIML-3349] Add SwinTransformer support for torch_onnx quantization workflow

ajrasane:
## Summary
- Enable end-to-end quantize → ONNX export → TRT engine pipeline for SwinTransformer models (v1 and v2) across FP8, INT8, MXFP8, NVFP4, and auto precision modes
- Add Conv2d quantization overrides for TRT compatibility (TRT only supports FP8/INT8 for convolutions)
- Fix FP8 LayerNorm type mismatch in TRT stronglyTyped mode by adding `LayerNormalization` to `change_casts_to_fp16`
- Fix `cast_initializer_to_dtype` crash when node has no initializer inputs
- Simplify `download_example_onnx.py` to a single `--timm_model_name` (required) flag, removing redundant `--vit` and `--llama` flags
- Add vision model support matrix to README (ViT, Swin, SwinV2)
- Rewrite tests: parametrize over (ViT, Swin, SwinV2) × (fp8, int8, mxfp8, nvfp4, auto) with TRT engine build verification

## Test plan
- [ ] `python -m pytest tests/examples/torch_onnx/test_torch_quant_to_onnx.py -v` — 15 tests (3 models × 5 modes), all pass
- [ ] Verified Swin accuracy on ImageNet-1k across all precisions (FP8: 81.29%, INT8: 81.12%, MXFP8: 81.32%, NVFP4: 80.79%, Auto: 80.84% TRT top-1 vs 81.37% base)
- [ ] INT4_AWQ deferred (TODO in test file) — requires INT4 exporter changes for non-MatMul/Gemm consumer patterns

## Summary by CodeRabbit

* **New Features**
  * ONNX export supports arbitrary timm vision models with auto device selection and new CLI options (--timm_model_name, --model_kwargs, --no_pretrained); batch-size/input sizing is now model-generic.

* **Bug Fixes**
  * Expanded FP16/BF16 cast handling to additional ONNX ops.
  * Disabled inplace ReLU before auto-quantization to avoid incorrect transforms.
  * Conv2d quantization overrides added for improved TensorRT compatibility.
  * Safer handling when initializers are missing during dtype casting.

* **Documentation**
  * README updated with supported models table, quantization mappings, and example CLI usage.

* **Tests**
  * Tests expanded to multiple architectures/quant modes and now verify TensorRT engine build.

### DIFF
--- a/examples/onnx_ptq/README.md
+++ b/examples/onnx_ptq/README.md
@@ -56,7 +56,7 @@ Most of the examples in this doc use `vit_base_patch16_224.onnx` as the input mo
 
 ```bash
 python download_example_onnx.py \
-    --vit \
+    --timm_model_name=vit_base_patch16_224 \
     --onnx_save_path=vit_base_patch16_224.onnx \
     --fp16 # <Optional, if the desired output ONNX precision is FP16>
 ```
--- a/examples/onnx_ptq/download_example_onnx.py
+++ b/examples/onnx_ptq/download_example_onnx.py
@@ -51,6 +51,12 @@ if __name__ == "__main__":
         help="Export timm/vit_base_patch16_224 model to ONNX.",
     )
     parser.add_argument(
+        "--timm_model_name",
+        type=str,
+        default="vit_base_patch16_224",
+        help="Export any timm model to ONNX (e.g., swin_tiny_patch4_window7_224).",
+    )
+    parser.add_argument(
         "--llama",
         action="store_true",
         help="Export meta-llama/Llama-3.1-8B-Instruct to ONNX with KV cache.",
@@ -62,7 +68,7 @@ if __name__ == "__main__":
         "--batch_size",
         type=int,
         default=1,
-        help="Batch size for the exported ViT model.",
+        help="Batch size for the exported model.",
     )
     parser.add_argument(
         "--fp16",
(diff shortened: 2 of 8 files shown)


cjluo-nv (changes requested):
## Review: Swin Transformer ONNX Quantization + Export

Good PR — the Conv2d quantization overrides for TRT compatibility are well-designed and follow existing `quant_cfg` patterns. The end-to-end test coverage with `trtexec` engine builds is a strong addition. A few issues to address before merging:

### Issues to Fix

**1. `--timm_model_name` default is always truthy (`examples/onnx_ptq/download_example_onnx.py:55`)**

The argument has `default="vit_base_patch16_224"`, so `if args.timm_model_name:` (line 99) is always `True`. This means running with no args exports ViT unintentionally, and `--vit` exports it twice. Fix: change default to `None` and use `elif`.

**2. `# nosec` on `subprocess.run` (`tests/examples/torch_onnx/test_torch_quant_to_onnx.py:53`)**

Per project security guidelines, `# nosec` bypasses are prohibited. The call is safe (list args, no `shell=True`), so just remove the comment.

**3. PR description claims a `cast_initializer_to_dtype` fix not present in the diff**

The summary says "Fix `cast_initializer_to_dtype` crash when node has no initializer inputs" but there are no changes to `modelopt/onnx/quantization/qdq_utils.py`. The bug is real (`input_id` is unbound at line 1168 when no inputs are initializers), but it's not fixed here. Either include the fix or remove the claim.

**4. PR description says "10 tests (2 models × 5 modes)" but the code parametrizes 3 models × 5 modes = 15 tests.** Also verify swinv2_tiny + auto actually passes, since the README marks it ✅.

### Non-blocking Notes

- `_disable_inplace_relu` is a good defensive fix for autograd backward hooks in auto_quantize.
- Extended `change_casts_to_fp16` op list (`LayerNormalization`, `Clip`, `Mul`, `Exp`) is reasonable for Swin's op patterns.
- `cpb_mlp` / `downsample` exclusions in `filter_func` are correct Swin-specific layers to skip for accuracy.
- `copy.deepcopy` on config dicts in `get_quant_config` correctly avoids mutating globals.
- Using `--no_pretrained` with reduced depth keeps tests fast — nice.

cjluo-nv (approved):
LGTM — all four issues from my previous review are addressed cleanly in 807a9a3.

One optional suggestion: `_verify_trt_engine_build()` will raise `FileNotFoundError` if `trtexec` is not in PATH. A `shutil.which("trtexec")` guard with `pytest.skip` would make the test more robust in environments without TensorRT CLI. Non-blocking.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

PR #1241, opened 2026-04-13
feat: parallelize fakequant export across GPUs via ThreadPoolExecutor

kinjalpatel27:
### What does this PR do?

Type of change: New feature

Refactors `export_hf_vllm_fq_checkpoint()` Step 1 (weight fake-quantization) to process weights concurrently across GPUs using `ThreadPoolExecutor`.

**Motivation**: When exporting large models with compute-heavy quantization backends (e.g., barboqt exhaustive search), the export loop processes one weight tensor at a time on a single GPU while other GPUs sit idle. On clusters with idle GPU reapers (e.g., cw-dfw), this causes jobs to be killed after 30 minutes of idle GPU time. Parallel export keeps all GPUs active during the quantize phase.

**Changes**:
- Collects all quantizer work items, groups by device
- One thread per GPU, each processes its weight partition concurrently
- Falls back to sequential on single GPU or CPU (no thread overhead)
- Adds `parallel: bool = True` parameter (default on, backwards compatible)
- GIL released during CUDA kernels → true parallelism
- `state_dict[key] = value` writes are atomic (single-key dict assignment)

**Measured speedup** on Qwen3-8B with `barboqt_exhaustive_2b` (8× H100 DGX):
- Quantize step: 293s → 49s (**6.0x**)
- Total export: 312s → 66s (**4.7x**)

Verified bitwise identical output (SHA256 checksums match) across FP8, NVFP4, NVFP4+rotation, psx_luts, psx_formats, luts/lo-bcq, and barboqt backends.

### Usage

```python
from modelopt.torch.export import export_hf_vllm_fq_checkpoint

# Default: parallel=True (uses all GPUs automatically)
export_hf_vllm_fq_checkpoint(model, export_dir="./export")

# Opt out if needed:
export_hf_vllm_fq_checkpoint(model, export_dir="./export", parallel=False)
```

No changes needed in `hf_ptq.py` or downstream scripts — `parallel=True` is the default and falls back gracefully on single GPU.

### Testing

- **Unit test (new)**: `tests/gpu/torch/export/test_vllm_fakequant_hf_parallel_export.py`
  - `test_parallel_vs_sequential_identical`: Requires ≥2 GPUs, verifies bitwise identical HF weights + quantizer state between parallel and sequential export
  - `test_single_gpu_fallback`: Verifies `parallel=True` on single GPU works without error
- **Existing test**: `test_vllm_fakequant_hf_export.py` passes unchanged (uses `parallel=True` default, falls back to sequential on single GPU)
- **Manual validation**: SHA256 file checksums verified identical for FP8, NVFP4, NVFP4_rotate presets on 8× H100

- Is this change backward compatible?: ✅ (`parallel=True` default; single GPU falls back to sequential; no API break)
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A (no copied code, no new deps)
- Did you write any new necessary tests?: ✅ (2 new GPU tests)
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ❌ (will update if requested)

### Additional Information

- Concurrent `ThreadPoolExecutor` is safe because: (1) `state_dict` is a detached copy, (2) each `TensorQuantizer` has per-instance state, (3) `torch.inference_mode()` context, (4) CUDA kernels release GIL, (5) `same_device_as()` context manager handles per-thread device switching
- For models loaded with `device_map="auto"`, weights are naturally distributed across GPUs → parallel export uses all GPUs with no configuration needed
- `device_map="sequential"` on small models may place everything on GPU 0 → parallel export gracefully falls back to sequential

## Release Notes

* **New Features**
  * Added parallel export capability to vLLM checkpoint quantization, enabling faster processing across multiple GPU devices.

* **Tests**
  * Added comprehensive test coverage verifying export results remain identical between parallel and sequential execution.
  * Added single-GPU fallback test to ensure compatibility on systems with fewer GPUs.

### DIFF
--- a/modelopt/torch/export/plugins/vllm_fakequant_hf.py
+++ b/modelopt/torch/export/plugins/vllm_fakequant_hf.py
@@ -82,6 +82,7 @@ def _collect_quant_work(
                     hasattr(candidate, "_pre_quant_scale")
                     and candidate._pre_quant_scale is not None
                     and candidate._disabled
+                    and getattr(candidate, "_enable_pre_quant_scale", True)
                 ):
                     inp_q = candidate
                     inp_q_key = get_unwrapped_name(
@@ -179,22 +180,28 @@ def export_hf_vllm_fq_checkpoint(
                 num_cuda_devices,
                 ", ".join(f"{d}: {len(items)} weights" for d, items in device_groups.items()),
             )
-            all_results: list[tuple[str, torch.Tensor, str | None]] = []
             with ThreadPoolExecutor(max_workers=num_cuda_devices) as pool:
-                futures = []
+                # Submit GPU batches first (non-blocking)
+                futures = [
+                    pool.submit(_process_device_batch, items, device)
+                    for device, items in device_groups.items()
+                    if device.type == "cuda"
+                ]
+                # Process CPU weights inline while GPU futures run
                 for device, items in device_groups.items():
-                    if device.type == "cuda":
-                        futures.append(pool.submit(_process_device_batch, items, device))
-                    else:
-                        # CPU weights: process inline (no thread needed).
-                        all_results.extend([_process_weight(item) for item in items])
+                    if device.type != "cuda":
+                        for sd_key, w_quant, inp_q_key in map(_process_weight, items):
+                            state_dict[sd_key] = w_quant
+                            fakequant_weights.add(sd_key)
+                            if inp_q_key is not None:
+                                input_quantizers_folded_pqs.add(inp_q_key)
+                # Collect GPU results
                 for future in futures:
-                    all_results.extend(future.result())
-            for sd_key, w_quant, inp_q_key in all_results:
-                state_dict[sd_key] = w_quant
-                fakequant_weights.add(sd_key)
-                if inp_q_key is not None:
-                    input_quantizers_folded_pqs.add(inp_q_key)
+                    for sd_key, w_quant, inp_q_key in future.result():
+                        state_dict[sd_key] = w_quant
+                        fakequant_weights.add(sd_key)
+                        if inp_q_key is not None:
+                            input_quantizers_folded_pqs.add(inp_q_key)
         else:
             # Sequential fallback (single GPU, CPU, or parallel=False).
             for item in work_items:
--- a/modelopt/torch/export/plugins/vllm_fakequant_hf.py
+++ b/modelopt/torch/export/plugins/vllm_fakequant_hf.py
@@ -14,6 +14,11 @@
 # limitations under the License.
 """Export HuggingFace model to vLLM fakequant checkpoint."""
 
+import logging
+import time
+from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
 from pathlib import Path
 
 import torch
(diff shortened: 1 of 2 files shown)


realAsma (approved):

cjluo-nv (commented):
Good PR with a well-motivated performance optimization. The parallelization approach is sound - grouping work by device, one thread per GPU, using ThreadPoolExecutor with GIL-releasing CUDA kernels. The refactoring into `_collect_quant_work` / `_process_weight` / `_process_device_batch` is clean.

However, there are a few issues:

1. **Dead code in tests**: `_quantize_model` helper is defined but never called.
2. **Unrelated behavioral change**: The `_enable_pre_quant_scale` check addition and the `orig_rotate is not None` guard are correctness improvements but are unrelated to the parallelization feature. They should ideally be in a separate PR or at least called out explicitly.
3. **Test limitation**: The `test_parallel_vs_sequential_identical` test uses `device_map="auto"` which may not actually distribute a tiny 4-layer model across GPUs on all systems, potentially making it test the sequential fallback path instead of the parallel path. Consider adding an assertion that the parallel path was actually taken.
4. **Redundant `torch.inference_mode()`**: The outer `with torch.inference_mode()` in `export_hf_vllm_fq_checkpoint` and the inner one in `_process_device_batch` are redundant (though harmless).

tests/gpu/torch/export/test_vllm_fakequant_hf_parallel_export.py
```diff
@@ -0,0 +1,139 @@
...
+from _test_utils.torch.transformers_models import create_tiny_llama_dir
+from transformers import AutoModelForCausalLM
+
+import modelopt.torch.quantization as mtq
+from modelopt.torch.export import export_hf_vllm_fq_checkpoint
+
+
+def _quantize_model(tmp_path, suffix=""):
+    """Create and quantize a tiny LLaMA model. Returns (model, export_dir)."""
+    tiny_model_dir = create_tiny_llama_dir(tmp_path / f"model{suffix}", num_hidden_layers=4)
```
cjluo-nv, line 28: This `_quantize_model` helper function is defined but never called by any test in this file. Please remove it to avoid dead code confusion.

tests/gpu/torch/export/test_vllm_fakequant_hf_parallel_export.py
```diff
@@ -0,0 +1,139 @@
...
+        pytest.skip("Need >= 2 GPUs for parallel export test")
+
+    # Create a tiny model and spread across GPUs.
+    tiny_model_dir = create_tiny_llama_dir(tmp_path / "model", num_hidden_layers=4)
+    model = AutoModelForCausalLM.from_pretrained(
+        tiny_model_dir, device_map="auto", torch_dtype=torch.float16
+    )
+    model.eval()
+
+    def forward_loop(model):
```
cjluo-nv, line 56: With only 4 hidden layers in a tiny model, `device_map="auto"` may place everything on a single GPU depending on the system's memory profile, causing the parallel path to silently fall back to sequential. Consider adding a check after quantization that weights are actually on multiple devices, or manually distributing layers:

```python
# After quantization, verify weights are on multiple GPUs
devices = {p.device for p in model.parameters()}
assert len([d for d in devices if d.type == 'cuda']) > 1, (
    f"Model not spread across GPUs: {devices}. Test won't exercise parallel path."
)
```

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -28,6 +33,93 @@
...
+            # Check for pre_quant_scale folding eligibility.
+            inp_q = None
+            inp_q_key = None
+            inp_attr = attr_name.replace("weight_quantizer", "input_quantizer")
+            if hasattr(module, inp_attr):
+                candidate = getattr(module, inp_attr)
+                if (
+                    hasattr(candidate, "_pre_quant_scale")
+                    and candidate._pre_quant_scale is not None
+                    and candidate._disabled
```
cjluo-nv, line 84: The `_enable_pre_quant_scale` check is a correctness improvement (prevents double-folding when `disable_pre_quant_scale_and_resmooth` was called), but it's a behavioral change unrelated to parallelization. Consider noting this in the PR description or splitting it into a separate fix for easier review/bisect.

Also, `getattr(candidate, "_enable_pre_quant_scale", True)` always returns `True` for the default because `TensorQuantizer.__init__` sets `self._enable_pre_quant_scale = True`. The `getattr` with default is unnecessary — `candidate._enable_pre_quant_scale` would suffice (since `candidate` is guaranteed to be a `TensorQuantizer` here).

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -166,4 +278,5 @@ def export_hf_vllm_fq_checkpoint(
 
     for wq, orig_rotate in wqs_to_restore:
         wq.enable()
-        wq._rotate = orig_rotate
+        if orig_rotate is not None:
+            wq._rotate = orig_rotate
```
cjluo-nv, line 282: The `if orig_rotate is not None:` guard is a behavioral change unrelated to parallelization. Looking at `TensorQuantizer.__init__` → `set_from_attribute_config`, `_rotate` is always set (defaults to `False`), so it should never be `None`. If this guard is protecting against some edge case, please document which scenario produces `_rotate = None`. Otherwise, this changes the restore semantics — the old code would always restore, the new code skips restoration if `None`.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -28,6 +33,93 @@
...
+    w = item.weight
+    w_quant = item.quantizer(w.float()).to(w.dtype)
+    if item.inp_q is not None:
+        scale = item.inp_q._pre_quant_scale.squeeze().to(device=w_quant.device)
+        w_quant = (w_quant * scale[None, :]).to(w_quant.dtype)
+    return item.sd_key, w_quant.cpu(), item.inp_q_key
+
+
+def _process_device_batch(items: list[_WeightQuantWork], device: torch.device):
+    """Process all weight items on a single GPU. Runs in a dedicated thread."""
```
cjluo-nv, line 117: Nit: `_process_device_batch` wraps with `torch.inference_mode()` but it's always called from within the outer `with torch.inference_mode():` block in `export_hf_vllm_fq_checkpoint`. The nested context is harmless (inference_mode is re-entrant) but redundant for the sequential/CPU-inline paths. For the threaded GPU path, it IS necessary since `torch.inference_mode()` is thread-local. Consider adding a comment explaining why this inner context is needed (thread-local inference mode).

PR #1242, opened 2026-04-13
vLLM fakequant export update for AWQ checkpoint

kinjalpatel27:
### What does this PR do?

Type of change: Bug

Enables end-to-end AWQ checkpoint export and reload in the vLLM fake-quant serving path (`MODELOPT_STATE_PATH`). Previously, the `input_quantizer` was using incorrect `pre_quant_scale` especially with grouped quantizers like `qkv_proj`, using simply the first `input_quantizer.pre_quant_scale`. This MR adds `_resmooth_experts_for_export` that non-mutatively averages `pre_quant_scale` across MoE experts and unifies input `_amax`, required because vLLM uses a single input quantizer per expert group. Adds `merge_amax_tensors_for_group` (element-wise max for same-shape, `cat` for GQA, scalar-max fallback) replacing the scalar-collapsing `torch.stack().max()` that  dropped per-channel `_amax` structure. 

### Usage

```python
# Export AWQ checkpoint from HF model
  from modelopt.torch.export.plugins.vllm_fakequant_hf import export_hf_vllm_fq_checkpoint
  export_hf_vllm_fq_checkpoint(model, export_dir="./awq_vllm_checkpoint")      
```

### Testing
 **Step 1 — Export the quantized checkpoint:**                                                                                                                                                                
  ```bash                    
    --pyt_ckpt_path <MODEL_PATH> \                                                                                                                                                                             
    --recipe <AWQ_RECIPE> \
    --calib_size 512 \                                                                                                                                                                                         
    --export_path <EXPORT_DIR> \                                                                                                                                                                               
    --vllm_fakequant_export
```
  This produces `<EXPORT_DIR>/vllm_fq_modelopt_state.pth` with the averaged per-expert                                                                                                                           
  pre_quant_scale and unified _amax now included.                                                                                                                                                              
   

 Step 2 — Serve via vLLM fakequant worker:                                                                                                                                                                    
```bash
  MODELOPT_STATE_PATH=<EXPORT_DIR>/vllm_fq_modelopt_state.pth \
    python examples/vllm_serve/vllm_serve_fakequant.py \                                                                                                                                                       
      <EXPORT_DIR> --tensor-parallel-size <TP>   
```

Tested for quantization configurations:
```
FP8_DEFAULT_CFG
FP8_DEFAULT_CFG (input_q disabled)
INT8_SMOOTHQUANT_CFG
INT8_WEIGHT_ONLY_CFG
NVFP4_DEFAULT_CFG
NVFP4_AWQ_LITE_CFG
INT4_AWQ_CFG
NVFP4_AWQ_CFG
NVFP4_DEFAULT_CFG (input_q disabled)
```

## Summary by CodeRabbit

* **New Features**
  * Added Nemotron-style MoE export support and group-aware AWQ resmoothing with optional requantization during export.
  * Improved handling for shared-input / expert groups and tensor-parallel sharding of pre-quantization scales.

* **Bug Fixes**
  * Removed AWQ reload limitation from known issues; improved checkpoint validation and safer save/load behavior.
  * Better detection and handling of enabled weight-quantizers and clearer warnings for mismatched checkpoint keys.

### DIFF
--- a/examples/vllm_serve/README.md
+++ b/examples/vllm_serve/README.md
@@ -98,5 +98,5 @@ QUANT_CFG=<quant_cfg> QUANT_FILE_PATH=<quantizer_state.pth> python vllm_serve_fa
 ## Known Problems
 
 1. **MCore reload does not use `MODELOPT_STATE_PATH`**; use `QUANT_FILE_PATH` and make sure `QUANT_CFG` matches the quantization recipe used for the original MCore model (otherwise quantizer keys/config won’t align).
-3. KV cache quantization export and reload is not supported in MCore yet.
-4. **`NVFP4_KV_CFG` and `NVFP4_AFFINE_KV_CFG` require `--enforce-eager`**; these configs use a dynamic-block Triton kernel for KV-cache quantization that is incompatible with CUDA graph capture (the kernel grid is computed from Python-level tensor shapes, which get baked in at capture time). Without `--enforce-eager`, the captured grid will be wrong for different batch sizes, producing incorrect outputs.
+2. KV cache quantization export and reload is not supported in MCore yet.
+3. **`NVFP4_KV_CFG` and `NVFP4_AFFINE_KV_CFG` require `--enforce-eager`**; these configs use a dynamic-block Triton kernel for KV-cache quantization that is incompatible with CUDA graph capture (the kernel grid is computed from Python-level tensor shapes, which get baked in at capture time). Without `--enforce-eager`, the captured grid will be wrong for different batch sizes, producing incorrect outputs.
--- a/examples/vllm_serve/README.md
+++ b/examples/vllm_serve/README.md
@@ -98,6 +98,5 @@ QUANT_CFG=<quant_cfg> QUANT_FILE_PATH=<quantizer_state.pth> python vllm_serve_fa
 ## Known Problems
 
 1. **MCore reload does not use `MODELOPT_STATE_PATH`**; use `QUANT_FILE_PATH` and make sure `QUANT_CFG` matches the quantization recipe used for the original MCore model (otherwise quantizer keys/config won’t align).
-2. AWQ reload is not supported yet
 3. KV cache quantization export and reload is not supported in MCore yet.
 4. **`NVFP4_KV_CFG` and `NVFP4_AFFINE_KV_CFG` require `--enforce-eager`**; these configs use a dynamic-block Triton kernel for KV-cache quantization that is incompatible with CUDA graph capture (the kernel grid is computed from Python-level tensor shapes, which get baked in at capture time). Without `--enforce-eager`, the captured grid will be wrong for different batch sizes, producing incorrect outputs.
--- a/examples/vllm_serve/fakequant_worker.py
+++ b/examples/vllm_serve/fakequant_worker.py
@@ -31,7 +31,7 @@ from vllm_reload_utils import (
 )
 
 import modelopt.torch.quantization as mtq
-from modelopt.torch.export.hf_vllm_quantizer_merge import is_weight_quantizer_state_key
+from modelopt.torch.export.plugins.vllm_fakequant_hf import is_weight_quantizer_state_key
 from modelopt.torch.quantization.plugins.vllm import (
     disable_compilation,
     post_restore_vllm_parallel_linears,
--- a/examples/vllm_serve/fakequant_worker.py
+++ b/examples/vllm_serve/fakequant_worker.py
@@ -15,6 +15,7 @@
 
 
 import os
+import warnings
 from typing import Any
 
 import torch
@@ -62,8 +63,11 @@ def _fakequant_run_prolog_worker(self) -> None:
         model = model.unwrap()
     if quant_config["modelopt_state_path"]:
         print(f"Loading modelopt state from {quant_config['modelopt_state_path']}")
-        # Load on CPU to avoid failures when the checkpoint was saved from a different
-        # GPU mapping
+        # map_location="cpu": load tensors on CPU so device ids in the file need not match this worker.
+        # weights_only=False: ``vllm_fq_modelopt_state.pth`` is a full ModelOpt pickle (metadata,
+        # nested dicts, dtypes, etc.); PyTorch's ``weights_only=True`` rejects that and only
+        # allows tensor-only checkpoints. Loading arbitrary pickles can execute stored code—use
+        # paths you trust (your own exports or verified checkpoints).
         modelopt_state = torch.load(
             quant_config["modelopt_state_path"], weights_only=False, map_location="cpu"
         )
@@ -73,16 +77,12 @@ def _fakequant_run_prolog_worker(self) -> None:
             if hasattr(self.model_runner.model, "hf_to_vllm_mapper")
             else None
         )
-        # convert modelopt state to vllm format
         modelopt_state = convert_modelopt_state_to_vllm(modelopt_state, map_fun=map_fun)
-        # restore model from modelopt state
         restore_from_modelopt_state_vllm(model, modelopt_state)
 
         if modelopt_weights is not None:
-            # convert quantizer state values to vllm format
             modelopt_weights = convert_dict_to_vllm(modelopt_weights, map_fun=map_fun)
             mtq.utils.set_quantizer_state_dict(model, modelopt_weights)
-            # Log any pqs keys in modelopt_weights that were not loaded (key mismatch).
             if not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
                 from modelopt.torch.quantization.nn import TensorQuantizer
                 from modelopt.torch.utils import get_unwrapped_name
@@ -93,22 +93,22 @@ def _fakequant_run_prolog_worker(self) -> None:
                     if isinstance(m, TensorQuantizer)
                 }
                 pqs_in_weights = {
-                    k for k, v in modelopt_weights.items()
+                    k
+                    for k, v in modelopt_weights.items()
                     if isinstance(v, dict) and "_pre_quant_scale" in v
                 }
                 unmatched_pqs = pqs_in_weights - loaded_keys
                 if unmatched_pqs:
-                    print(
-                        f"[fakequant_worker] WARNING: {len(unmatched_pqs)} pqs key(s) in "
-                        f"modelopt_weights have no matching quantizer in the model:\n"
-                        + "\n".join(f"  {k}" for k in sorted(unmatched_pqs)[:20])
+                    sample = sorted(unmatched_pqs)[:20]
+                    warnings.warn(
+                        f"{len(unmatched_pqs)} checkpoint pre_quant_scale key(s) have no "
+                        f"matching TensorQuantizer in the model (showing up to 20): {sample}",
+                        stacklevel=2,
                     )
-            # set_quantizer_state_dict does not invoke modelopt_post_restore (unlike restore_quantizer_state).
+            # set_quantizer_state_dict does not run modelopt_post_restore (unlike restore_quantizer_state).
             post_restore_vllm_parallel_linears(model)
-            # For RowParallelLinear layers, pqs was saved at full H_in width (single-GPU
-            # export). Slice it to [H_in / tp_world_size] so it matches the local shard.
-            _debug_pqs = os.environ.get("MODELOPT_DEBUG_PQS", "0").lower() in ("1", "true")
-            shard_pre_quant_scale_for_tp(model, debug=_debug_pqs)
+            # Row-parallel: slice full-width exported pre_quant_scale to this TP rank.
+            shard_pre_quant_scale_for_tp(model)
 
     else:
         if quant_config["quant_file_path"]:
@@ -127,15 +127,13 @@ def _fakequant_run_prolog_worker(self) -> None:
 
         quant_cfg = get_quant_config(quant_config, model)
 
-        # quantize model
         with disable_compilation(model):
             print("Quantizing model...")
             mtq.quantize(model, quant_cfg, forward_loop=calibrate_loop)
 
         quantizer_file_path = quant_config["quant_file_path"]
         if quantizer_file_path:
-            # Get amax and other quantizer state from the quantizer file
-            # this can be used with Megatron-LM exported model using export_mcore_gpt_to_hf_vllm_fq
+            self.model_runner._dummy_run(1)
             current_state_dict = load_state_dict_from_path(self, quantizer_file_path, model)
             model.load_state_dict(current_state_dict)
 
--- a/examples/vllm_serve/fakequant_worker.py
+++ b/examples/vllm_serve/fakequant_worker.py
@@ -92,8 +92,10 @@ def _fakequant_run_prolog_worker(self) -> None:
                     for n, m in model.named_modules()
                     if isinstance(m, TensorQuantizer)
                 }
+                # Same namespace as ``loaded_keys``: checkpoint keys may include DDP/FSDP
+                # prefixes that ``convert_dict_to_vllm`` does not strip.
                 pqs_in_weights = {
-                    k
+                    get_unwrapped_name(k, model)
                     for k, v in modelopt_weights.items()
                     if isinstance(v, dict) and "_pre_quant_scale" in v
                 }
@@ -107,7 +109,7 @@ def _fakequant_run_prolog_worker(self) -> None:
                     )
             # set_quantizer_state_dict does not run modelopt_post_restore (unlike restore_quantizer_state).
             post_restore_vllm_parallel_linears(model)
-            # Row-parallel: slice full-width exported pre_quant_scale to this TP rank.
+            # Must follow post_restore: shard_pre_quant_scale_for_tp uses weight H_in vs pqs length.
             shard_pre_quant_scale_for_tp(model)
 
     else:
--- a/examples/vllm_serve/fakequant_worker.py
+++ b/examples/vllm_serve/fakequant_worker.py
@@ -31,10 +31,12 @@ from vllm_reload_utils import (
 )
 
 import modelopt.torch.quantization as mtq
+from modelopt.torch.export.hf_vllm_quantizer_merge import is_weight_quantizer_state_key
 from modelopt.torch.quantization.plugins.vllm import (
     disable_compilation,
     post_restore_vllm_parallel_linears,
 )
+from modelopt.torch.utils import safe_load
 from modelopt.torch.utils.dataset_utils import get_dataset_dataloader
 
 quant_config: dict[str, Any] = {
@@ -63,14 +65,8 @@ def _fakequant_run_prolog_worker(self) -> None:
         model = model.unwrap()
     if quant_config["modelopt_state_path"]:
         print(f"Loading modelopt state from {quant_config['modelopt_state_path']}")
-        # map_location="cpu": load tensors on CPU so device ids in the file need not match this worker.
-        # weights_only=False: ``vllm_fq_modelopt_state.pth`` is a full ModelOpt pickle (metadata,
-        # nested dicts, dtypes, etc.); PyTorch's ``weights_only=True`` rejects that and only
-        # allows tensor-only checkpoints. Loading arbitrary pickles can execute stored code—use
-        # paths you trust (your own exports or verified checkpoints).
-        modelopt_state = torch.load(
-            quant_config["modelopt_state_path"], weights_only=False, map_location="cpu"
-        )
+        # Load on CPU to avoid failures when the checkpoint was saved from a different GPU mapping.
+        modelopt_state = safe_load(quant_config["modelopt_state_path"], map_location="cpu")
         modelopt_weights = modelopt_state.pop("modelopt_state_weights", None)
         map_fun = (
             self.model_runner.model.hf_to_vllm_mapper.apply_dict
@@ -148,7 +144,7 @@ def _fakequant_run_prolog_worker(self) -> None:
 
     mtq.fold_weight(model)
     for name, module in model.named_modules():
-        if name.endswith("weight_quantizer"):
+        if is_weight_quantizer_state_key(name):
             assert not module.is_enabled, f"quantizer {name} is still enabled"
 
 
--- a/examples/vllm_serve/fakequant_worker.py
+++ b/examples/vllm_serve/fakequant_worker.py
@@ -26,6 +26,7 @@ from vllm_reload_utils import (
     convert_modelopt_state_to_vllm,
     load_state_dict_from_path,
     restore_from_modelopt_state_vllm,
+    shard_pre_quant_scale_for_tp,
 )
 
 import modelopt.torch.quantization as mtq
@@ -64,7 +65,7 @@ def _fakequant_run_prolog_worker(self) -> None:
         # Load on CPU to avoid failures when the checkpoint was saved from a different
         # GPU mapping
         modelopt_state = torch.load(
-            quant_config["modelopt_state_path"], weights_only=True, map_location="cpu"
+            quant_config["modelopt_state_path"], weights_only=False, map_location="cpu"
         )
         modelopt_weights = modelopt_state.pop("modelopt_state_weights", None)
         map_fun = (
@@ -81,8 +82,33 @@ def _fakequant_run_prolog_worker(self) -> None:
             # convert quantizer state values to vllm format
             modelopt_weights = convert_dict_to_vllm(modelopt_weights, map_fun=map_fun)
             mtq.utils.set_quantizer_state_dict(model, modelopt_weights)
+            # Log any pqs keys in modelopt_weights that were not loaded (key mismatch).
+            if not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
+                from modelopt.torch.quantization.nn import TensorQuantizer
+                from modelopt.torch.utils import get_unwrapped_name
+
+                loaded_keys = {
+                    get_unwrapped_name(n, model)
+                    for n, m in model.named_modules()
+                    if isinstance(m, TensorQuantizer)
+                }
+                pqs_in_weights = {
+                    k for k, v in modelopt_weights.items()
+                    if isinstance(v, dict) and "_pre_quant_scale" in v
+                }
+                unmatched_pqs = pqs_in_weights - loaded_keys
+                if unmatched_pqs:
+                    print(
+                        f"[fakequant_worker] WARNING: {len(unmatched_pqs)} pqs key(s) in "
+                        f"modelopt_weights have no matching quantizer in the model:\n"
+                        + "\n".join(f"  {k}" for k in sorted(unmatched_pqs)[:20])
+                    )
             # set_quantizer_state_dict does not invoke modelopt_post_restore (unlike restore_quantizer_state).
             post_restore_vllm_parallel_linears(model)
+            # For RowParallelLinear layers, pqs was saved at full H_in width (single-GPU
+            # export). Slice it to [H_in / tp_world_size] so it matches the local shard.
+            _debug_pqs = os.environ.get("MODELOPT_DEBUG_PQS", "0").lower() in ("1", "true")
+            shard_pre_quant_scale_for_tp(model, debug=_debug_pqs)
 
     else:
         if quant_config["quant_file_path"]:
--- a/examples/vllm_serve/fakequant_worker.py
+++ b/examples/vllm_serve/fakequant_worker.py
@@ -144,8 +144,11 @@ def _fakequant_run_prolog_worker(self) -> None:
 
     mtq.fold_weight(model)
     for name, module in model.named_modules():
-        if is_weight_quantizer_state_key(name):
-            assert not module.is_enabled, f"quantizer {name} is still enabled"
+        if is_weight_quantizer_state_key(name) and module.is_enabled:
+            raise RuntimeError(
+                f"Weight quantizer {name!r} is still enabled after fold_weight — "
+                "double-quantization would corrupt activations."
+            )
 
 
 class FakeQuantWorker(BaseWorker):
--- a/examples/vllm_serve/vllm_reload_utils.py
+++ b/examples/vllm_serve/vllm_reload_utils.py
@@ -23,10 +23,7 @@ from typing import Any
 import torch
 from vllm.distributed.parallel_state import get_tp_group
 
-from modelopt.torch.export.plugins.vllm_fakequant_hf import (
-    is_weight_quantizer_state_key,
-    merge_amax_tensors_for_vllm_group,
-)
+from modelopt.torch.export.plugins.vllm_fakequant_hf import is_weight_quantizer_state_key
 from modelopt.torch.opt.conversion import (
     ModelLikeModule,
     ModeloptStateManager,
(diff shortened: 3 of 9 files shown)


cjluo-nv (commented):
This PR makes significant improvements to enable AWQ checkpoint export and reload in the vLLM fakequant serving path. The changes are substantial (~473 lines across 6 files) and touch critical export/reload paths. While the overall design is sound and addresses a real gap (AWQ export was listed as unsupported), there are several correctness concerns, missing tests for new utility functions, and a potentially risky silent-failure pattern that should be addressed before merging.

**What's good:**
- Proper refactoring of TP sharding into `_narrow_tensor_to_tp_local_shard` (eliminates duplication)
- `merge_amax_tensors_for_group` correctly handles same-shape, different-shape (GQA), and fallback cases
- Non-mutative resmoothing via `_resmooth_experts_for_export` avoids modifying the live model
- Use of `safe_load`/`safe_save` for security best practices
- Good use of `is_weight_quantizer_state_key` regex for consistent matching

**Key concerns:**
1. **No unit tests for new utility functions** (`merge_amax_tensors_for_group`, `is_weight_quantizer_state_key`, `_resmooth_experts_for_export`, `shard_pre_quant_scale_for_tp`, `_infer_prefix_remap`). These are critical correctness-sensitive functions.
2. A silent `contextlib.suppress(Exception)` in `_resmooth_experts_for_export` that could mask real failures.
3. The `_collect_expert_pre_quant_scales` function name mentions "experts" but is used for all shared-input groups including dense GQA projections, which could be confusing.
4. The existing test `test_hf_vllm_export` only covers FP8, but AWQ is the main feature being added — no AWQ test case was added.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +62,161 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+
+        avg_pqs_out = avg_pqs.detach().clone()
+        for m in modules:
+            nm = id_to_name.get(id(m))
+            if nm is None:
+                continue
+            out[get_unwrapped_name(f"{nm}.input_quantizer", model)] = (avg_pqs_out, synced_amax)
+
+    # MoE expert groups — must be enumerated by name because MoE routing sends
+    # different tokens to each expert, so forward hooks cannot detect them as
```
cjluo-nv, line 420: Bug concern: `contextlib.suppress(Exception)` here will silently swallow **any** failure during the dummy forward — including actual bugs like shape mismatches, device mismatches, or import errors. Since `collect_shared_input_modules` relies on hooks that fire during this forward, a silent failure means the dense GQA resmoothing silently becomes a no-op.

At minimum, log or warn when the forward fails so users aren't left debugging silent accuracy regressions:
```python
def _dummy_forward() -> None:
    try:
        model(torch.ones([1, 2], dtype=torch.long, device=dev))
    except Exception as e:
        import logging
        logging.getLogger(__name__).debug(
            "Dummy forward for shared-input detection failed (expected for VLMs): %s", e
        )
```

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +62,161 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+def requant_weights_for_export(
+    quantizer: TensorQuantizer,
+    w: torch.Tensor,
+) -> torch.Tensor:
+    """Requantize weights for export."""
+    quantizer_copy = copy.deepcopy(quantizer)
+    quantizer_copy.eval()
+    quantizer_copy.reset_amax()
+    enable_stats_collection(quantizer_copy)
+    quantizer_copy(w)
```
cjluo-nv: Minor: `requant_weights_for_export` does `deep_copy` of the quantizer but never places it on the same device as the weight tensor `w`. If `w` is on GPU and the deepcopy lands on the same device this works, but if model parameters get offloaded (e.g., CPU offload), there could be a device mismatch. Consider explicitly moving `quantizer_copy` to `w.device`.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +62,161 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+            return
+
+        avg_pqs = torch.stack(pqs_list).mean(0)
+        avg_pqs = avg_pqs.clamp(min=torch.finfo(torch.float32).tiny)
+
+        for m in modules:
+            nm = id_to_name.get(id(m))
+            if nm is None or f"{nm}.weight" not in state_dict:
+                continue
+            old_pqs = m.input_quantizer._pre_quant_scale
```
cjluo-nv, line 385: The weight adjustment formula `W' = W * old_pqs / avg_pqs` is correct for the mathematical equivalence: if the original computation is `(x * old_pqs) @ W^T`, then with new scale `avg_pqs`, we need `(x * avg_pqs) @ W'^T = (x * old_pqs) @ W^T`, so `W' = W * old_pqs / avg_pqs`. Good.

However, after adjusting the weight, `requant_weights` is needed because the old weight quantizer's amax was calibrated against the original weight. The `requant_weights_for_export` function does this correctly by re-running stats collection. 👍

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +62,161 @@ def disable_rotate(quantizer: TensorQuantizer):
     return False
 
 
+def _collect_expert_pre_quant_scales(
+    experts: list[nn.Module],
+) -> list[torch.Tensor] | None:
+    """Return per-expert ``pre_quant_scale`` tensors if every expert can be averaged; else None.
```
cjluo-nv, line 223: Naming: `_collect_expert_pre_quant_scales` is also called for dense GQA q/k/v projections (not just MoE experts). Consider renaming to `_collect_group_pre_quant_scales` to better reflect its dual usage in `_process_group`.

examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -195,6 +197,38 @@ def _merge_values_require_identical(merged_key: str, key_value_pairs: list[tuple
...
+    It must be **2-D** (mappers expect matrix weights; 1-D often errors). Only the returned key
+    path is used; values are ignored. A CPU tensor is enough for typical HF↔vLLM name mapping.
+    """
+    prefix_remap: dict[str, str] = {}
+    probe_weight = torch.empty((1, 1))
+    for key in quantizer_keys:
+        first_component = key.split(".")[0]
+        if first_component in prefix_remap:
+            continue
+        last_dot = key.rfind(".")
```
cjluo-nv: The `_infer_prefix_remap` approach is fragile: it constructs synthetic `*.weight` keys and probes `map_fun` to discover prefix remappings. This works for simple cases (e.g., `backbone` → `model`) but could produce incorrect results if `map_fun` does more complex transformations (e.g., reordering layers, renaming intermediate components). 

Also, what happens if two different quantizer keys map to different prefix remappings for the same `first_component`? The first one wins due to `if first_component in prefix_remap: continue`.

This should at minimum have unit tests covering edge cases.

examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -317,7 +370,14 @@ def _has_buffers(state: dict) -> bool:
             }
             # Add state for quantizers in model but not in metadata (e.g. disabled/excluded)
             for k in model_keys - filtered.keys():
-                filtered[k] = model_qstate[k]
+                state = model_qstate[k]
+                # Weight quantizers absent from exported metadata were disabled during export
+                # (weights are already fake-quantized and pre_quant_scale is folded in).
+                # Keep them disabled on reload so fold_weight does not re-quantize the
+                # already-folded weights (re-quantizing distorts the pqs-scaled values).
+                if is_weight_quantizer_state_key(k) and not state.get("_disabled"):
```
cjluo-nv, line 370: The logic to force-disable weight quantizers that are absent from exported metadata is important for correctness (prevents re-quantization of already-folded weights). However, this creates a subtle contract: the export side must _always_ strip weight quantizer keys, and the reload side must _always_ disable them. Any mismatch would corrupt weights silently.

Consider adding an explicit comment or assertion that validates this invariant, e.g., checking that all weight quantizers in the model are either (a) present in the filtered metadata or (b) have been force-disabled.

examples/vllm_serve/fakequant_worker.py
```diff
@@ -121,8 +143,11 @@ def _fakequant_run_prolog_worker(self) -> None:
```
cjluo-nv, line 144: Good improvement: `is_weight_quantizer_state_key(name)` properly handles `SequentialQuantizer` entries (e.g., `weight_quantizer.0`) whereas the old `name.endswith("weight_quantizer")` did not. However, note that `module.is_enabled` is checked on the **module** found by `model.named_modules()`, but `name` here is the full dotted path. Since `is_weight_quantizer_state_key` uses a regex that matches substrings, it will also match on things like `some_layer.weight_quantizer.0._amax` if such paths exist as named modules. Verify this doesn't produce false positives.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +62,161 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+    quantizer_copy(w)
+    finish_stats_collection(quantizer_copy)
+    return quantizer_copy(w.float()).to(w.dtype)
+
+
+def merge_amax_tensors_for_group(tensors: list[torch.Tensor]) -> torch.Tensor:
+    """Combine `_amax` buffers from a merge group into a single tensor.
+
+    Used when HuggingFace module names are folded to vLLM names (e.g. q/k/v → qkv_proj).
+
```
cjluo-nv, line 283: The `merge_amax_tensors_for_group` function's `torch.cat` fallback for different shapes assumes dimension 0 is the right concat axis. For GQA where q_amax has shape `[N_q]` and k_amax has shape `[N_kv]`, this is correct (concatenating per-output-channel amaxes). But for higher-dimensional amaxes (e.g., per-block), `dim=0` may not be right. The scalar fallback via `torch.max(flat)` handles this but loses structure.

This seems acceptable for now given the use cases, but worth documenting the assumption that `dim=0` concat is for per-channel amax.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -96,7 +289,7 @@ def export_hf_vllm_fq_checkpoint(
                         if (
```
cjluo-nv: Bug fix: Changed from `inp_q._disabled` to `not inp_q.is_enabled`. This is correct — `_disabled` is an internal flag, while `is_enabled` is the public property that accounts for all disable conditions. Good catch.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -133,7 +326,7 @@ def export_hf_vllm_fq_checkpoint(
```
cjluo-nv: Using `is_weight_quantizer_state_key(key)` instead of `key.endswith("weight_quantizer")` is correct and handles `SequentialQuantizer` entries. Good.

examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -153,7 +155,7 @@ def _merge_values_by_max_or_concat(merged_key: str, key_value_pairs: list[tuple[
         for dict_key in values[0]:
```
cjluo-nv, line 179: Good change: replacing `torch.stack(tensors).max(dim=0)[0]` with `merge_amax_tensors_for_group(tensors)` correctly handles the GQA case where q/k/v amax tensors may have different shapes.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: addressed in [248a98e](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/248a98e31b58fb073a22ae7b0c4bf407da2e6298)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: addressed in [248a98e](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/248a98e31b58fb073a22ae7b0c4bf407da2e6298)

examples/vllm_serve/vllm_reload_utils.py
kinjalpatel27: addressed in [248a98e](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/248a98e31b58fb073a22ae7b0c4bf407da2e6298)

examples/vllm_serve/vllm_reload_utils.py
kinjalpatel27: addressed in [248a98e](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/248a98e31b58fb073a22ae7b0c4bf407da2e6298)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: addressed in [248a98e](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/248a98e31b58fb073a22ae7b0c4bf407da2e6298)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: addressed in [248a98e](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/248a98e31b58fb073a22ae7b0c4bf407da2e6298)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -116,15 +418,20 @@ def export_hf_vllm_fq_checkpoint(
...
-                    attr_name.endswith("weight_quantizer")
-                    and isinstance(quantizer, TensorQuantizer)
-                    and quantizer.is_enabled
-                ):
+                if not (attr_name.endswith("weight_quantizer") and quantizer.is_enabled):
+                    continue
+                if isinstance(quantizer, SequentialQuantizer):
+                    quantizer.disable()
+                    for sub in quantizer:
+                        orig_rotate = sub._rotate
```
cjluo-nv: **Critical (unresolved from previous review):** The weight quantizer disable/restore pattern is still not wrapped in `try/finally`. If `get_quantizer_state_dict()`, `safe_save()`, or `model.save_pretrained()` raises, quantizers remain disabled and `_rotate` remains patched.

```python
wqs_to_restore: list[tuple[TensorQuantizer, Any]] = []
try:
    for _, module in model.named_modules():
        # ... disable quantizers, populate wqs_to_restore ...
    
    # ... all save/export operations ...
    safe_save(modelopt_state, export_dir / "vllm_fq_modelopt_state.pth")
    model.save_pretrained(export_dir, state_dict=clean_sd, save_modelopt_state=False)
finally:
    for wq, orig_rotate in wqs_to_restore:
        wq.enable()
        wq._rotate = orig_rotate
```

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +137,195 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+        avg_pqs = avg_pqs.clamp(min=torch.finfo(torch.float32).tiny)
+
+        for m in modules:
+            nm = id_to_name.get(id(m))
+            if nm is None or f"{nm}.weight" not in state_dict:
+                continue
+            old_pqs = m.input_quantizer._pre_quant_scale
+            avg_pqs_dev = avg_pqs.to(device=old_pqs.device, dtype=old_pqs.dtype)
+            if torch.equal(old_pqs, avg_pqs_dev):
+                continue
```
cjluo-nv, line 388: **Critical (unresolved from previous review):** When `pqs_list` contains fp16/bf16 tensors, `torch.stack(pqs_list).mean(0)` stays in that dtype. Then `avg_pqs.clamp(min=torch.finfo(torch.float32).tiny)` is useless because `torch.finfo(torch.float32).tiny ≈ 1.175e-38` underflows to 0.0 in fp16 (smallest fp16 normal is ~6e-5). This allows division by zero in line 280, producing `inf` weights.

Fix: compute in fp32:
```python
avg_pqs = torch.stack([p.float() for p in pqs_list]).mean(0)
avg_pqs = avg_pqs.clamp(min=torch.finfo(torch.float32).tiny)
```
Then cast back to original dtype at line 295 when writing `avg_pqs_out`.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +137,195 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+            except NotImplementedError:
+                pass
+            else:
+                for experts in expert_groups:
+                    if experts:
+                        _process_group(experts)
+
+    # Dense shared-input groups (e.g. q/k/v in GQA attention) — detected via forward
+    # hooks so any architecture is covered regardless of projection attribute names.
+
```
cjluo-nv, line 444: **Minor (unresolved from previous review):** Wrap the dummy forward in `torch.inference_mode()` to avoid building autograd graphs during this probe forward. The caller `collect_shared_input_modules` already uses `torch.no_grad()`, but `inference_mode` is stronger and avoids unnecessary tensor versioning overhead.

```python
def _dummy_forward() -> None:
    with torch.inference_mode():
        try:
            model(torch.ones([1, 2], dtype=torch.long, device=dev))
        except Exception as e:
            ...
```

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -14,19 +14,118 @@
...
+        if not isinstance(module, QuantModule):
+            continue
+        for attr_name, quantizer in module.named_children():
+            if attr_name.endswith("weight_quantizer") and isinstance(
+                quantizer, (TensorQuantizer, SequentialQuantizer)
+            ):
+                assert not quantizer.is_enabled, (
+                    f"vLLM fakequant export: {attr_name!r} must be disabled before saving "
+                    f"quantizer_state (weights already folded). "
+                    f"See filter_modelopt_state_quantizer_state_for_model in vllm_reload_utils."
```
cjluo-nv: **Minor (per codebase guidelines):** Bare `assert` for a runtime invariant check can be optimized away with `python -O`. Use an explicit `RuntimeError` instead:
```python
if quantizer.is_enabled:
    raise RuntimeError(
        f"vLLM fakequant export: {attr_name!r} must be disabled ..."
    )
```

examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -195,6 +196,38 @@ def _merge_values_require_identical(merged_key: str, key_value_pairs: list[tuple
...
+    quantizer_keys: dict[str, Any],
+    map_fun: Callable[[dict[str, Any]], dict[str, Any]],
+) -> dict[str, str]:
+    """Map HF root name → vLLM root (e.g. ``backbone`` → ``model``) using ``map_fun`` on ``*.weight`` keys.
+
+    Quantizer keys never go through ``map_fun`` later, so we probe with a tiny CPU placeholder.
+    It must be **2-D** (mappers expect matrix weights; 1-D often errors). Only the returned key
+    path is used; values are ignored. A CPU tensor is enough for typical HF↔vLLM name mapping.
+    """
+    prefix_remap: dict[str, str] = {}
```
cjluo-nv: **Minor:** This `_infer_prefix_remap` is a less robust duplicate of `infer_quantizer_prefix_remap` in `vllm_fakequant_hf.py`. The public version validates consistency (raises on conflicting root mappings for the same HF prefix), while this version silently takes the first mapping via `if first_component in prefix_remap: continue`.

Consider delegating to the public API:
```python
from modelopt.torch.export.plugins.vllm_fakequant_hf import infer_quantizer_prefix_remap
```
Or at minimum, add the same consistency check here.

examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -317,7 +369,25 @@ def _has_buffers(state: dict) -> bool:
...
-                filtered[k] = model_qstate[k]
+                state = model_qstate[k]
+                # Weight quantizers absent from exported metadata were disabled during export
+                # (weights are already fake-quantized and pre_quant_scale is folded in).
+                # Keep them disabled on reload so fold_weight does not re-quantize the
+                # already-folded weights (re-quantizing distorts the pqs-scaled values).
+                if is_weight_quantizer_state_key(k) and not state.get("_disabled"):
+                    state = {**state, "_disabled": True}
+                filtered[k] = state
+
```
cjluo-nv, line 373: **Minor (per codebase guidelines):** Bare `assert` for a runtime invariant that validates export/reload contract. Should use `RuntimeError` — `assert` can be optimized away with `-O`.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +137,195 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+def merge_amax_tensors_for_group(tensors: list[torch.Tensor]) -> torch.Tensor:
+    """Combine `_amax` buffers from a merge group into a single tensor.
+
+    Used when HuggingFace module names are folded to vLLM names (e.g. q/k/v → qkv_proj).
+
+    - If every tensor has the same shape, take the element-wise maximum over the group
+      (conservative when each branch carried the same axis layout).
+    - If shapes differ: ``torch.cat(..., dim=0)`` assumes **1D per-channel** amaxes in
+      fused order (e.g. GQA q/k/v → ``[N_q]`` + ``[N_kv]`` + ``[N_kv]``), matching vLLM’s
+      grouped quantizer. Not valid for 2D blockwise amax; on failure, **scalar**
```
cjluo-nv, line 288: **Bug concern in `requant_weights_for_export`:** The calibration pass (lines 206-211) runs the quantizer on `weight` (for single-quantizer) or `weight.float()` (for chain). Then after `finish_stats_collection`, the re-application pass (lines 214-216) always uses `weight.float()`. The single-quantizer calibration path may collect different amax stats than the application path uses if `weight` is not float32. This is subtle — the calibration sees the original dtype (e.g. fp16) values, but then the quantizer applies to float32 values. For correctness both passes should use the same input.

Also, the result of the calibration pass (line 206-211) is assigned to `weight_quantized` but then immediately overwritten at line 214 — this is confusing even if intentional. Consider adding a comment or using `_` for the discarded result.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -86,7 +385,10 @@ def export_hf_vllm_fq_checkpoint(
                 )
                 if sd_key in state_dict:
```
cjluo-nv: **Minor:** Bare `assert` at a runtime check:
```python
assert sd_key not in fakequant_weights, ...
```
Should use `RuntimeError` for the same reason as other `assert` instances.

cjluo-nv (commented):
This is a substantial and well-designed PR that enables AWQ checkpoint export/reload in vLLM fakequant. Many previous review comments have been addressed (renamed functions, `is_enabled` instead of `_disabled`, `merge_amax_tensors_for_group` usage, dummy forward logging, device alignment for `deepcopy`, invariant assertions, tests added for `infer_quantizer_prefix_remap` and `merge_amax_tensors_for_group`).

**However, several critical issues from previous reviews remain unresolved:**

1. **Missing `try/finally` for `wqs_to_restore`** (lines 430-493 of `vllm_fakequant_hf.py`) — weight quantizers are disabled but restoration only happens at the end. If `safe_save` or `save_pretrained` throws, the model is left in a mutated state. This was flagged in the second review iteration and is still unresolved.

2. **`avg_pqs` computed in half precision** (line 271) — `torch.stack(pqs_list).mean(0)` preserves the original fp16/bf16 dtype, then clamping with `torch.finfo(torch.float32).tiny` is meaningless because that value underflows to zero in fp16. Division by zero may produce inf weights. This was flagged as Major in the 4th review.

3. **Missing `torch.inference_mode()` in `_dummy_forward`** (line 307) — Builds unnecessary autograd graphs during the probe forward. Flagged in 4th review.

4. **Bare `assert` in `_check_all_weight_quantizers_disabled`** (line 127) — can be optimized away with `-O`. Per codebase guidelines, use `RuntimeError`.

5. **Bug in `requant_weights_for_export`** — The single-quantizer calibration path runs `sequence_quantizers[0](weight)` without `.float()` but the application pass uses `weight.float()`. While the intent seems correct (calibrate then apply), the inconsistency could lead to precision-dependent amax values.

6. **Duplicate `_infer_prefix_remap`** — Two implementations exist: `infer_quantizer_prefix_remap` (in `vllm_fakequant_hf.py`, with consistency checks) and `_infer_prefix_remap` (in `vllm_reload_utils.py`, without consistency checks). The reload-side function should delegate to the public API to avoid divergence.

The test coverage has improved with tests for `merge_amax_tensors_for_group` and `infer_quantizer_prefix_remap`, but key functions like `_resmooth_experts_for_export`, `requant_weights_for_export`, `is_weight_quantizer_state_key`, and `shard_pre_quant_scale_for_tp` still lack tests.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

examples/vllm_serve/vllm_reload_utils.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

examples/vllm_serve/vllm_reload_utils.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: fixed in [609207c](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/609207c0d51a8e0e9d7f971856d06310a34ca1d4)

cjluo-nv (approved):
All critical issues from previous reviews have been addressed:

1. **`try/finally` for `wqs_to_restore`** — ✅ Fixed. Weight quantizer disable/restore is now properly wrapped in try/finally.
2. **fp32 averaging of `avg_pqs`** — ✅ Fixed. Uses `torch.stack([p.float() for p in pqs_list]).mean(0)` to avoid fp16/bf16 underflow.
3. **`torch.inference_mode()` in `_dummy_forward`** — ✅ Fixed.
4. **Bare `assert` → `RuntimeError`** — ✅ Fixed across all changed files.
5. **`_disabled` → `is_enabled` public API** — ✅ Fixed.
6. **`merge_amax_tensors_for_group` usage** — ✅ Used for all amax merging, preserving per-channel structure.
7. **Duplicate `_infer_prefix_remap`** — ✅ Eliminated. Reload side now imports `infer_quantizer_prefix_remap` from the export module.
8. **Weight quantizer invariant assertion** — ✅ Added proper validation.
9. **`deepcopy` device alignment** — ✅ `requant_weights_for_export` does `.to(device=weight.device)`.
10. **`safe_save`/`safe_load`** — ✅ Used instead of raw `torch.save`/`torch.load`.

Tests have been meaningfully expanded: GPU test now covers INT4_AWQ_CFG in addition to FP8, and unit tests cover `infer_quantizer_prefix_remap` (7 cases) and `merge_amax_tensors_for_group` (3 cases). The code is well-structured with clear documentation and reasonable design decisions throughout.

tests/gpu/torch/export/test_vllm_fakequant_hf_export.py
```diff
@@ -25,7 +25,15 @@
 from modelopt.torch.utils import safe_load
 
 
-@pytest.mark.parametrize("quant_cfg", [mtq.FP8_DEFAULT_CFG])
+@pytest.mark.parametrize(
```
meenchen: Can you add a test for MoE?

meenchen (approved):
Change LGTM overall, request to add a unit test for MoE

tests/gpu/torch/export/test_vllm_fakequant_hf_export.py
kinjalpatel27: added one in [9a6f5d2](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/9a6f5d2b051310bb9ba38f82d1d1ca50bd6aa662)

examples/vllm_serve/README.md
```diff
@@ -98,6 +98,5 @@ QUANT_CFG=<quant_cfg> QUANT_FILE_PATH=<quantizer_state.pth> python vllm_serve_fa
 ## Known Problems
 
 1. **MCore reload does not use `MODELOPT_STATE_PATH`**; use `QUANT_FILE_PATH` and make sure `QUANT_CFG` matches the quantization recipe used for the original MCore model (otherwise quantizer keys/config won’t align).
-2. AWQ reload is not supported yet
```
mxinO, line 101: Change 3, 4-> 2, 3

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,6 +139,206 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+        enable_stats_collection(quantizer_copy)
+    # Match legacy single-quantizer path: first calib uses ``w`` as-is; chains use float.
+    if len(quantizers) == 1:
+        weight_quantized = quantizers[0](weight)
+    else:
+        weight_quantized = weight
+        for quantizer_copy in quantizers:
+            weight_quantized = quantizer_copy(weight_quantized)
+    for quantizer_copy in quantizers:
+        finish_stats_collection(quantizer_copy)
```
mxinO, line 270: This logic looks can be simplified, distinguish the `len(quantizers) == 1` is redundant, the loop in the else just works.

mxinO (commented):
LGTM. Thanks.
Further question.
We are solving the the pqs fusing at the export time which is sub-optimal, this is actually a problem of calibration time, since all our inference frameworks use fused q,k,v and experts, any thought to solve it at calibration time? cc @realAsma

examples/vllm_serve/README.md
kinjalpatel27: Good catch, fixed in [4b66f05](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/4b66f05a5f00023b9cc9a6a7051daec40f0e339d)

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: Good catch fixed in [4b66f05](https://github.com/NVIDIA/Model-Optimizer/pull/1242/commits/4b66f05a5f00023b9cc9a6a7051daec40f0e339d)

kevalmorabia97 (approved):

PR #1245, opened 2026-04-13
copy custom modeling files to pruned checkpoint dirs

j-rausch:
## Summary

pruned FFN checkpoints created by `init_child_from_parent` were missing custom `.py` files (e.g. `modeling_nemotron_h.py`, `configuration_nemotron_h.py`). without them, `AutoConfig.from_pretrained(..., trust_remote_code=True)` fails silently and the checkpoint is excluded from the replacement library, reducing MIP candidate diversity.

adds `copy_remote_code_files(source_dir, output_dir)` to `checkpoint_utils_hf.py` that copies `*.py` files from the source checkpoint root. called from `init_child_from_parent` after `copy_tokenizer` when `descriptor.requires_trust_remote_code()` is true.

## Summary by CodeRabbit

* **Refactoring**
  * Improved remote-code copying for checkpoints: reliably copies top-level Python files from local or hub sources and ensures destination directories exist.
  * Checkpoint initialization now performs remote-code persistence earlier for consistent loading across workflows.
* **Tests**
  * Added unit tests validating local and hub remote-code copying behavior; removed an older GPU test module.


modelopt/torch/puzzletron/tools/checkpoint_utils_hf.py
```diff
@@ -448,3 +449,26 @@ def save_model_config(model_config: PretrainedConfig, checkpoint_dir: Path | str
             for conf in model_config.block_configs
         ]
     model_config.save_pretrained(checkpoint_dir)
+
+
+def copy_remote_code_files(source_dir: Path | str, output_dir: Path | str) -> None:
```
kevalmorabia97: This does not work if input is HF hub path. Can you use this approach from test utils as it works for both: https://github.com/NVIDIA/Model-Optimizer/blob/feature/puzzletron/tests/_test_utils/torch/puzzletron/utils.py#L198-L206

We can then use this function in the test utils.py.

j-rausch: done

kevalmorabia97: I think my approach may not work for local model path so perhaps we need both? Please check

kevalmorabia97 (approved):

PR #1246, opened 2026-04-13
vLLM fakequant fold weight_quantizer for megatron export

kinjalpatel27:
### What does this PR do?

Type of change: Bug fix                                                                                                                                                                        
                                                                                                                                                                                                             
 During Megatron→vLLM fakequant export (`export_mcore_gpt_to_hf_vllm_fq`), the   `weight_quantizer` is now applied as fake-quantization (quantize + dequantize) directly into the exported weight tensor, and its amax is no longer saved to `quantizer_state.pth`.  On reload, if `weight_quantizer` keys are absent from the checkpoint (because they were folded at export time), the corresponding quantizer modules are disabled.
This change is useful especially when amax across experts are not synced for `weight_quantizer`, this allows the `weight_quantizer` to keep them different for better accuracy.

### Usage
```python                                                                                                                                                                                                    
# Unchanged — export API is the same                                                                                                                                                                         
export_mcore_gpt_to_hf_vllm_fq(model, pretrained_model_name_or_path=..., export_dir=...)
```
 
### Testing
Step 1 — Quantize (run from Megatron-LM `examples/post_training/modelopt`):                                                                                                                              
  ```bash
HF_MODEL_CKPT=<path/to/hf/weights> MLM_MODEL_SAVE=<quant-ckpt-name> \
bash quantize.sh <hf-model-id> NVFP4_DEFAULT_CFG
```  

Step 2 — Export for vLLM fakequant:                                                                                                                                                                          
```bash  
MLM_EXTRA_ARGS=--export-vllm-fq \ 
HF_MODEL_CKPT=<path/to/hf/weights> \ 
MLM_MODEL_CKPT=<quant-ckpt-name> \ 
EXPORT_DIR=<export-dir> \ 
bash export.sh <hf-model-id> 
```

Step 3 — Serve (run from examples/vllm_serve):                                                                                                                                                               
```bash
 QUANT_CFG=NVFP4_DEFAULT_CFG \ 
 QUANT_FILE_PATH=<export-dir>/quantizer_state.pth \ 
 python3 vllm_serve_fakequant.py <export-dir> \ 
 -tp 1 --served-model-name <model-name> \ 
 --host 0.0.0.0 --port 8000 \ 
--trust-remote-code --enforce-eager \ 
 --disable-custom-all-reduce \ 
--gpu-memory-utilization 0.8          
```

### Additional Information
<!-- E.g. related issue. -->


## Summary by CodeRabbit

* **Bug Fixes**
  * Better handling when loading checkpoints: missing weight-quantizer entries are validated and corresponding modules are disabled to avoid load failures.

* **Improvements**
  * Export now folds enabled weight quantizers into exported weights when present and omits internal weight-quantizer tensors from the exported state to produce cleaner exports.


examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -438,20 +437,30 @@ def load_state_dict_from_path(
...
-    # Ensure counts match
-    if checkpoint_quant_count != model_quant_count:
-        warnings.warn(
-            f"Mismatch in quantizer state key counts: checkpoint has {checkpoint_quant_count} "
-            f"quant keys but model has {model_quant_count} quantizer state keys. "
-            f"This can happen if the model is using PP."
-        )
+    for name, module in model.named_modules():
+        if name in missing_wq_module_paths and hasattr(module, "disable"):
+            module.disable()
```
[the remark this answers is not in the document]
kinjalpatel27: `weight_quantizer` is added as part of vllm.py. Unless user specifically register a custom module with custom quantizer name, this scenario should not occur.

modelopt/torch/export/plugins/vllm_fakequant_megatron.py
```diff
@@ -133,14 +137,41 @@ def _get_quantized_state(
         block_size = 0
 
         if hasattr(module, "weight") and module.weight is not None:
-            weight = module.weight.to(dtype).cpu()
-            name_to_value["weight"] = weight
+            weight = module.weight.to(dtype)
+            # Fold the weight_quantizer into the weight by applying fake-quantization
+            # (quantize then dequantize). The weight_quantizer amax is not exported;
```
jenchen13, line 142: is the amax currently exported into the HF weights? also how do we handle the quant config for new/unrecognized quant formats? how does VLLM recognize the new quant formats (e.g. PSX LUTS)

kinjalpatel27: 1. amax is exported in the separate file for input quantizer. Weight quantizers are folded in weights, so we dont export them separately.
2. QUANT_CFG is currently user's responsibility to pass the same config as MCore. Any configuration which is registered using `register_quant_backend` is recognized by vllm fakequant: [get_quant_config](https://github.com/NVIDIA/Model-Optimizer/blob/main/examples/vllm_serve/vllm_ptq_utils.py#L142)

realAsma (approved):

modelopt/torch/export/plugins/vllm_fakequant_megatron.py
jenchen13: 1. makes sense
2. by "user's responsibility" you mean the user must call `get_quant_config` and copy the quant config to the HF `config.json`? will VLLM serve work without this step?

kinjalpatel27: Currently user need to use the same QUANT cfg they used for quantizing MCore with vLLM fakequant serve. If the config is only part of MCore, they will have to add it to Modelopt code and use it with `QUANT_CFG`

jenchen13 (approved):

modelopt/torch/export/plugins/vllm_fakequant_megatron.py
jenchen13: is there a way to automatically detect quant cfg in a future PR and have it be part of config.json?

kinjalpatel27: I will create a ticket to track that for future PR.

meenchen (commented):
### 1. Correctness — Stricter Missing Key Handling May Break PP

**[QUESTION] Removed count-mismatch warning may regress pipeline parallelism**
The old code warned when checkpoint and model quantizer key counts didn't match (e.g., PP splits) and proceeded. The new code raises `ValueError` for any missing non-weight-quantizer key. If PP causes missing `input_quantizer` keys in the checkpoint (model has more layers than this rank's checkpoint), this would now error where it previously warned. Has this been tested with PP > 1?

### 2. Correctness — `module.disable()` without type check

**[SUGGESTION] No `isinstance` guard before calling `disable()`**
```python
    if name in missing_wq_module_paths and hasattr(module, "disable"):
        module.disable()
```
The `hasattr(module, "disable")` check is a reasonable duck-type guard, but if any non-TensorQuantizer module at that path happens to have a `disable()` method, it would be called unintentionally. Consider adding `isinstance(module, TensorQuantizer)` for safety.

### 3. Design — Device juggling in `_get_quantized_state`

**[SUGGESTION] Temporary CUDA lift for quantizer buffers is fragile**
When weight is on CPU, buffers are moved to CUDA, the quantizer runs, then buffers are restored. If the quantizer forward internally creates new buffers or registers state, those won't be moved back. The `finally` block only restores buffers captured before the forward. Consider a comment noting this assumption, or use `quantizer.to(quant_device)` / `quantizer.to(orig_device)` for a cleaner round-trip.

cjluo-nv (approved):

kinjalpatel27:
> 1. Correctness — Stricter Missing Key Handling May Break PP
> [QUESTION] Removed count-mismatch warning may regress pipeline parallelism
> The old code warned when checkpoint and model quantizer key counts didn't match (e.g., PP splits) and proceeded. The new code raises ValueError for any missing non-weight-quantizer key. If PP causes missing input_quantizer keys in the checkpoint (model has more layers than this rank's checkpoint), this would now error where it previously warned. Has this been tested with PP > 1?

Addressed it in [3e2ec17](https://github.com/NVIDIA/Model-Optimizer/pull/1246/commits/3e2ec17fed0bf1bde6ad798762478a7296f7bebd). Verified it works with PP>1

> 2. Correctness — module.disable() without type check
> [SUGGESTION] No isinstance guard before calling disable()
> 
> for name, module in model.named_modules():
>     if name in missing_wq_module_paths and hasattr(module, "disable"):
>         module.disable()
> The hasattr(module, "disable") check is a reasonable duck-type guard, but if any non-TensorQuantizer module at that path happens to have a disable() method, it would be called unintentionally. Consider adding isinstance(module, TensorQuantizer) for safety.

Addressed it in [3e2ec17](https://github.com/NVIDIA/Model-Optimizer/pull/1246/commits/3e2ec17fed0bf1bde6ad798762478a7296f7bebd)
 

> 3. Design — Device juggling in _get_quantized_state
> [SUGGESTION] Temporary CUDA lift for quantizer buffers is fragile
> When weight is on CPU, buffers are moved to CUDA, the quantizer runs, then buffers are restored. If the quantizer forward internally creates new buffers or registers state, those won't be moved back. The finally block only restores buffers captured before the forward. Consider a comment noting this assumption, or use quantizer.to(quant_device) / quantizer.to(orig_device) for a cleaner round-trip.

Addressed it in [3e2ec17](https://github.com/NVIDIA/Model-Optimizer/pull/1246/commits/3e2ec17fed0bf1bde6ad798762478a7296f7bebd)

meenchen (approved):

mxinO:
The loading logic changes seem mainly for disabling weight quantizers, disabling weight quantizers seem an often used case, a cleaner way may be directly add an env `LOADING_FOLDED_CKPT` in `fakequant_worker.py`, and append a `{"quantizer_name": "*weight_quantizer", "enable": False}` to the quant config if set, so all these handling can be avoided.

kinjalpatel27:
> The loading logic changes seem mainly for disabling weight quantizers, disabling weight quantizers seem an often used case, a cleaner way may be directly add an env `LOADING_FOLDED_CKPT` in `fakequant_worker.py`, and append a `{"quantizer_name": "*weight_quantizer", "enable": False}` to the quant config if set, so all these handling can be avoided.
I think adding `_union_quantizer_keys_across_ranks` is helpful regardless since it avoids the warning if PP > 1. The main difference is checking for weight_quantizer is being disabled if the keys are missing in ckpt. I wanted to avoid one more environment variable which user has to manage. Eventually, we will want to add support s.t. user will have to pass just checkpoint directory (and may be quantizer state file) and  and we should be able to reload the exported checkpoint

PR #1251, opened 2026-04-13
Add layerwise calibration for large models

realAsma:
## Summary

Adds **performant layerwise calibration** for quantizing large models (e.g. DeepSeek-R1 671B) that don't fit entirely on GPU. ([Example commands](#example-commands))

1. **Performant calibration for large models** — Each decoder layer is moved from CPU/disk to GPU (accelerate) or unsharded (FSDP2) **only once** and kept on GPU for the entire calibration step. Previously, every calibration batch triggered weight transfer for every layer — O(num_batches) weight movements per layer. Now it is O(1) per layer. This also means you can **increase batch size** since only one layer's weights occupy GPU at a time — e.g. DeepSeek-R1 on a single node (8×80GB) with `batch_size=16` and `gpu_max_mem_percentage=0.5`.
2. **Checkpoint save/resume** — Saves progress after each layer, so jobs that exceed cluster time limits (e.g. 4-hour Slurm windows for 100+ layer MoE models) can resume from the last completed layer.
3. **Rename** `sequential_calibrate` → `layerwise_calibrate` for clarity.

### Design details

The existing layerwise state machine (skip/run/capture) already processes one layer at a time, but skip-mode layers still kept their parameters in the ModuleList — so frameworks transferred all weights every forward pass. This PR adds:
- **`_SkipLayer`**: replaces fully-calibrated layers with a parameter-free dummy in the ModuleList, so framework hooks have nothing to transfer
- **`persistent_materialization`**: keeps the active layer on GPU for the entire calibration step, avoiding repeated offload/reload cycles

Checkpoint save is per-layer; restore is bulk — quantizer state and weights for layers 0..K-1 are restored once at the end of calibration, keeping the hot path fast.

### Example commands

**Qwen3-8B** (NVFP4+GPTQ, single GPU):
```bash
python hf_ptq.py \
    --recipe nvfp4_gptq_sequential.yaml \
    --calib_size 64 \
    --batch_size 16 \
    --dataset cnn_dailymail \
    --export_path outputs/qwen3_8b_nvfp4_gptq_seq \
    --gpu_max_mem_percentage 0.5 \
    --use_seq_device_map \
    --vllm_fakequant_export
```

**DeepSeek-R1** (NVFP4 experts-only + FP8 KV, 8×80GB):
```bash
python hf_ptq.py \
    --model unsloth/DeepSeek-R1-0528-BF16 \
    --recipe ../../modelopt_recipes/general/ptq/nvfp4_experts_only-fp8_kv.yaml \
    --dataset cnn_dailymail \
    --batch_size 16 \
    --calib_size 64 \
    --calib_seq 512 \
    --gpu_max_mem_percentage 0.5 \
    --use_seq_device_map \
    --trust_remote_code \
    --export_path output/DeepSeek-R1-BF16-nvfp4-experts-only-fp8-kv \
    --vllm_fakequant_export
```

### Example: NVFP4+GPTQ layerwise calibration on Qwen3-8B (36 layers, single GPU — 20 GB peak)

**Initial run** (killed after layer 11):
```
Layerwise calibration: Found 36 transformer layers
Calibrating layer 1/36 | capture: [1]
Computing Hessians for 7 linear layers...
GPTQ time: 51.39s
Calibrating layer 2/36 | run: [1] | capture: [2]
Checkpoint: saved layer 0
GPTQ time: 50.06s
Calibrating layer 3/36 | skip: 1 | run: [2] | capture: [3]
Checkpoint: saved layer 1
...
Calibrating layer 12/36 | skip: 10 | run: [11] | capture: [12]
Checkpoint: saved layer 10
<killed>
```

**Resumed run** (picks up from layer 11, finishes all 36):
```
Layerwise calibration: Found 36 transformer layers
Checkpoint: resuming layerwise calibration from layer 11/36
Calibrating layer 12 (resumed)
GPTQ time: 51.45s
Calibrating layer 13/36 | skip: 11 | run: [12] | capture: [13]
Checkpoint: saved layer 11
...
Calibrating layer 36/36 | skip: 34 | run: [35] | capture: [36]
Checkpoint: saved layer 34
GPTQ time: 50.33s
Checkpoint: saved layer 35 (final)
Checkpoint: restored 11 previously calibrated layers
Layerwise calibration completed
Quantized model exported to: outputs/qwen3_8b_nvfp4_gptq_seq
GPU 0: Peak memory usage = 20.42 GB
```

## TODO
- [ ] Update CHANGELOG

## Test plan
- `tests/unit/torch/quantization/test_layerwise_calibrate.py` — unit tests for skip/swap/restore
- `tests/unit/torch/quantization/test_sequential_checkpoint.py` — checkpoint save/resume correctness
- `tests/gpu/torch/quantization/plugins/test_accelerate_gpu.py` — CPU-offloaded layerwise + GPTQ + checkpoint resume
- `tests/gpu/torch/quantization/test_fsdp2.py` — FSDP2 layerwise calibration

### Verified
- [x] Qwen3-8B: layerwise calibration + checkpoint save/restore + fakequantized checkpoint export + vLLM serve
- [x] DeepSeek-R1: checkpoint resume tested
- [x] DeepSeek-R1: fakequantized checkpoint export verified


modelopt/torch/quantization/plugins/accelerate.py
```diff
@@ -50,32 +50,68 @@ def _get_cpu_offload_hook(hook):
...
 @contextmanager
 def weight_access_and_writeback_context(module):
-    """Context manager for weight access and writeback for modules managed by accelerate."""
+    """Context manager for weight access and writeback for modules managed by accelerate.
+
+    Handles two cases:
+    1. **Single-module**: the module's own ``_hf_hook`` is an offload hook.
+    2. **Sub-module**: the module's hook is non-offloading, but its children have
+       offload hooks (common with ``SequentialHook`` on sub-modules placed by
+       ``load_checkpoint_and_dispatch``).
```
realAsma: what about the third case - both module and submodule have offload hooks. How does this method work?

modelopt/torch/quantization/plugins/accelerate.py
```diff
@@ -50,32 +50,68 @@ def _get_cpu_offload_hook(hook):
...
-        # TODO: Add support for disk-offloaded models if needed (they will be really slow, hence low priority)
-
-        # This will load the weights from CPU state_dict and move it to the GPU from meta device
         align_hook.pre_forward(module)
+        try:
+            yield
+        finally:
+            _writeback_params_to_weights_map(module, align_hook)
+            align_hook.post_forward(module, None)
+        return
```
realAsma: this try -> catch does not look good to me. Can we do this in a more deterministic way? is it possible to elegantly generalize weight_access_and_writeback_context?

modelopt/torch/quantization/utils/activation_collector.py
```diff
@@ -328,8 +348,264 @@ def get_input_activations(self, layer: torch.nn.Module, forward_loop: ForwardLoo
             )
 
         inputs = list(info.collected_inputs)
-        # After capture, set to original so calib_func can call the layer's
-        # real forward directly.  The layer will transition to run → skip
-        # in subsequent iterations via _set_layer_states.
```
realAsma, line 333: revert this change. I want the comments.

realAsma, line 333: revert this change

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -328,8 +348,264 @@ def get_input_activations(self, layer: torch.nn.Module, forward_loop: ForwardLoo
...
+def _clone_to_cpu(obj: Any) -> Any:
+    """Recursively deep-clone tensors to CPU. Non-tensors are returned as-is."""
+    if isinstance(obj, torch.Tensor):
+        return obj.detach().cpu().clone()
+    if isinstance(obj, dict):
+        return {k: _clone_to_cpu(v) for k, v in obj.items()}
+    if isinstance(obj, (list, tuple)):
+        cloned = [_clone_to_cpu(v) for v in obj]
+        return type(obj)(cloned)
+    return obj
```
realAsma: why do we need a seperate method - can we use _move_to_device instead of this?

modelopt/torch/quantization/model_calib.py
```diff
@@ -1577,30 +1585,77 @@ def sequential_calibrate(
...
+                ckpt.save_prev(transformer_layers, layer_inputs)
 
             def _layer_forward_loop(m, _inputs=layer_inputs):
                 for args, kwargs_input in _inputs:
                     m(*args, **kwargs_input)
 
             calib_func(layer, _layer_forward_loop, **calib_kwargs)
 
+            if ckpt:
+                ckpt.stash(layer_idx, layer, model)
```
realAsma: why do we need a seperate stash step? can we reduce this to one save call and simplify the workflow?

modelopt/torch/utils/network.py
```diff
@@ -590,21 +622,37 @@ def get_unwrapped_name(name: str, model: nn.Module | None = None) -> str:
...
+            module._hf_hook = accelerate_hook
+
+            def new_forward(mod, *args, **kwargs):
+                args, kwargs = mod._hf_hook.pre_forward(mod, *args, **kwargs)
+                if mod._hf_hook.no_grad:
+                    with torch.no_grad():
+                        output = mod._old_forward(*args, **kwargs)
+                else:
+                    output = mod._old_forward(*args, **kwargs)
+                return mod._hf_hook.post_forward(mod, output)
```
realAsma: no I dont want to define this method -> can we cache the previous patch before calling remove_hook_from_module and restore it here?

modelopt/torch/utils/network.py
```diff
@@ -590,21 +621,27 @@ def get_unwrapped_name(name: str, model: nn.Module | None = None) -> str:
...
     accelerate_hook = None
+    patched_forward = None
     if hasattr(module, "_hf_hook"):
-        # A module with forward method patched by accelerate
-        from accelerate.hooks import add_hook_to_module, remove_hook_from_module
+        from accelerate.hooks import remove_hook_from_module
 
         accelerate_hook = module._hf_hook
+        patched_forward = module.forward
         remove_hook_from_module(module)
```
realAsma: why do we need to call remove_hook_from_module(module) now? can we directly do module.forward = (forward method before accelerate patching ??) - I am worried about calling remove_hook_from_module(module) bu not calling add_hook_to_module.

tests/unit/torch/quantization/test_sequential_calibrate.py
```diff
@@ -517,19 +517,110 @@ def forward_loop(m):
...
+    for layer in originals:
+        collector.get_input_activations(layer, forward_loop)
+    collector._unpatch_all_layers()
+
+    for i, orig in enumerate(originals):
+        assert model.layers[i] is orig, f"Layer {i} not restored to original after cleanup"
+        assert not hasattr(orig, "_seq_calib"), f"Layer {i} still has _seq_calib"
+
+
+def test_skip_dummy_has_no_hf_hook(monkeypatch):
```
realAsma: can we move this test to accelerate (torch/gpu/torch/quantization/plugsin/. ?)

modelopt/torch/quantization/mode.py
```diff
@@ -240,14 +241,12 @@ def wrapped_calib_func(
         if sequential:
             if forward_loop is None:
                 raise ValueError("forward_loop is required for calibration but got None.")
```
cjluo-nv, line 253: Bug/Regression: The assertion `method in ["max", "gptq"]` was removed without replacement. This means methods like `awq`, `smoothquant`, `svdquant`, and `mse` can now be passed with `use_sequential=True`, but these methods expect a full model as input, not a single decoder layer. When `sequential_calibrate` calls `calib_func(layer, _layer_forward_loop, **calib_kwargs)`, the `layer` is a single decoder block — methods like `awq` will fail or silently produce incorrect results.

Consider either:
1. Keeping a guard (possibly expanded): `assert method in ["max", "gptq", "mse", ...], ...`
2. Or adding a clear runtime check inside `sequential_calibrate` that the calib_func is compatible with per-layer invocation.

Without this, users get confusing errors deep in the calibration stack instead of a clear message.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -0,0 +1,630 @@
...
+    output_meta: tuple,
+    next_inputs: list | None,
+    num_layers: int,
+) -> None:
+    """Save a single layer checkpoint and update the manifest atomically."""
+    d = _layer_dir(checkpoint_dir, idx)
+    if os.path.isdir(d):
+        shutil.rmtree(d)
+    os.makedirs(d)
+    torch.save(weights, os.path.join(d, "weights.pt"))
```
cjluo-nv, line 506: Security: `weights_only=False` allows arbitrary code execution during unpickling. Since these checkpoints are written and read locally, the immediate risk is low, but this will be flagged by security scanners and PyTorch deprecation warnings.

For `output_meta.pt` (which contains tuples of shapes/dtypes), consider `weights_only=True` — it should work since the data is simple Python/torch types. For `weights.pt` and `next_inputs.pt` it may also work if they only contain tensors and basic containers. `quantizer_state.pt` may need `weights_only=False` depending on what's stored.

modelopt/torch/quantization/model_calib.py
```diff
@@ -1577,27 +1585,52 @@ def sequential_calibrate(
...
+        layer_inputs = input_getter.get_first_layer_inputs(
+            start_layer, resumed_inputs, forward_loop
+        )
 
-            def _layer_forward_loop(m, _inputs=layer_inputs):
-                for args, kwargs_input in _inputs:
+        for layer_idx in range(start_layer, num_layers):
+            layer = transformer_layers[layer_idx]
+
+            def _layer_forward_loop(m):
```
cjluo-nv: Nit: `_layer_forward_loop` captures `layer_inputs` from the enclosing scope. This works correctly because `calib_func` is called synchronously before `layer_inputs` is reassigned, but it's a subtle pattern that could break if `calib_func` ever stores the forward_loop for deferred execution. Consider making the capture explicit:
```python
def _layer_forward_loop(m, _inputs=layer_inputs):
    for args, kwargs_input in _inputs:
        m(*args, **kwargs_input)
```
This was actually the pattern used in the old code.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -0,0 +1,630 @@
...
+    """
+    manifest = _read_manifest(checkpoint_dir)
+    if manifest is None:
+        return None
+    last = manifest.get("last_completed_layer")
+    total = manifest.get("num_layers")
+    if last is None or total is None:
+        return None
+    if last + 1 >= total:
+        return None
```
cjluo-nv, line 528: Correctness concern: `restore_quantizer_state` is called with `{"quantizer_state": qstate}` where `qstate` was saved via `quantizer_state(layer)` (a flat dict keyed by quantizer names). However, `restore_quantizer_state` expects `metadata["quantizer_state"]` to have keys that match `quantizer_state(model)` which uses `get_unwrapped_name(n, model)`. But here you're passing `layer` (not the full model) as the first argument to `restore_quantizer_state`, so the keys should match since both save and restore use the layer scope.

Please verify this works correctly for layers with FSDP wrapping where `get_unwrapped_name` may produce different keys depending on whether `model` or `layer` is passed as the second argument.

modelopt/torch/quantization/plugins/accelerate.py
```diff
@@ -50,32 +50,79 @@ def _get_cpu_offload_hook(hook):
...
+        # Guard: the sub-module branch below is not reached when the parent has
+        # an offload hook.  Assert that no children also carry offload hooks,
+        # which would require a combined writeback strategy.
+        assert not any(
+            _get_cpu_offload_hook(mod._hf_hook)
+            for mod in module.modules()
+            if mod is not module and hasattr(mod, "_hf_hook")
+        ), (
+            "Both the module and one of its sub-modules have CPU-offload hooks. "
+            "weight_access_and_writeback_context does not support this layout yet."
```
cjluo-nv: Good defensive check. One edge case to consider: if the parent module's `_hf_hook` is an offload hook AND a child also has one, this assert will fire. Is this combination actually possible with `load_checkpoint_and_dispatch`? If so, this would be an unsupported configuration that crashes at runtime rather than being handled gracefully. Consider downgrading to a warning + fallback or documenting when this can occur.

modelopt/torch/quantization/model_calib.py
```diff
@@ -1577,27 +1585,52 @@ def sequential_calibrate(
...
+                for args, kwargs_input in layer_inputs:
                     m(*args, **kwargs_input)
 
             calib_func(layer, _layer_forward_loop, **calib_kwargs)
 
+            # Run one more forward to get next layer's inputs and set
```
cjluo-nv, line 1632: Minor: For the last layer, `cache_outputs_for_next_layer_calib` is not called, so `layer._seq_calib.output_meta` will be `None` (the layer was calibrated via `calib_func` which calls the layer's real forward, not the patched one that sets `output_meta`). The fallback in `save()` handles this with `torch.zeros(1)`, but this means the saved `output_meta.pt` for the last layer is a dummy. This is fine since it's never used for skip mode, but worth a comment noting this intentional behavior.

cjluo-nv (commented):
This is a substantial PR (~1500 lines) that adds checkpoint save/resume for sequential calibration, extends support to FSDP2 and accelerate-offloaded models, and renames `activation_collector.py` → `layerwise_calib.py`. The changes are cohesive and well-tested (unit + GPU tests for checkpoint, resume, offload, FSDP2 scenarios).

**Key issues found:**

1. **Removed guard on sequential calibration methods** — The assertion restricting sequential calibration to `max` and `gptq` was removed without replacement. Methods like `awq`, `smoothquant`, and `svdquant` operate on the full model (not per-layer) and will break silently or produce incorrect results when used with `use_sequential=True`.

2. **`weights_only=False` security concern** — `torch.load(..., weights_only=False)` is used for loading checkpoints, which can execute arbitrary code. While the checkpoints are locally generated, this is flagged by security scanners and should use `weights_only=True` where possible.

**Minor observations:**
- PR size is above ~1000 lines but the changes are cohesive and hard to split
- Good test coverage for the new functionality
- The `temporarily_remove_accelerate_hook` rewrite is a nice improvement avoiding the `init_hook` pitfall
- `_writeback_params_to_weights_map` properly handles all parameters (not just `weight`)
- FSDP2 context manager correctly generalized to handle all DTensor parameters

modelopt/torch/quantization/mode.py
realAsma: The old  guard was needed when  received a raw method string and dispatched internally. The new  receives  as a callable -- it never sees a method string at all. The dispatch is now handled by the config/mode system:  and  each declare  support and pass the appropriate calibration function. Unsupported methods (like ) simply don't set  in their config, so they never reach . The type-safe config system replaces the runtime assertion.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -0,0 +1,664 @@
...
+            prev = self._decoder_layers[layer_idx - 1]._layerwise_calib
+            if not prev.collected_inputs:
+                raise RuntimeError(
+                    f"Layer {layer_idx - 1} ({prev.name!r}) has no collected inputs to replay. "
+                    "Layers must be calibrated sequentially — ensure get_input_activations() "
+                    "was called for every preceding layer in order."
+                )
+            prev.mode = "run"
+            prev.cached_inputs = deque(prev.collected_inputs)
+            prev.collected_inputs = []
```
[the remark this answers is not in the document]
realAsma: Pushing back on this one. The  transition is a deliberate one-shot state machine: each layer transitions exactly once through . There is no retry mechanism -- if  raises, the calibration aborts via the  block in  which calls . Restoring previous-layer state for a retry that never happens would add complexity (saving/restoring deques, mode strings) with no benefit. The current cleanup correctly resets the *current* layer so the exception propagates cleanly.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: Good catch on verifying the key matching concern. We looked into this \u2014 both  and  use the same  scope (not the full model), so the keys are consistent by construction. Additionally, checkpoint resume is currently gated to single-process only (line 540-545), so FSDP2 wrapping doesn't come into play. Adding a runtime key-match assertion here would be unnecessary overhead for a condition that can't arise in the current code path. If/when we lift the single-process restriction, we'll revisit this as part of the FSDP2 support work.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: Added inline security comments above all four  calls explaining the files are internally generated by . Also added explicit  to the two loads that were missing it. Fixed in 43d1888b.

modelopt/torch/quantization/plugins/accelerate.py
realAsma: Replaced the bare  with  with a user-friendly message. Fixed in 43d1888b.

modelopt/torch/quantization/model_calib.py
realAsma: Added a comment explaining the dummy  is a placeholder for the last layer whose  is never used for skip mode. Fixed in 43d1888b.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -0,0 +1,630 @@
...
+        if self.start_layer == 0:
+            return None
+
+        last_ckpt = self.start_layer - 1
+
+        for i in range(self.start_layer):
+            d = _layer_dir(self.checkpoint_dir, i)
+            meta = torch.load(
+                os.path.join(d, "output_meta.pt"), map_location="cpu", weights_only=False
+            )
```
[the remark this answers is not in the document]
realAsma: Added inline comment. Fixed in 43d1888b.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -0,0 +1,630 @@
...
+            layer_device = get_module_device(layers[i])
+            meta = _remap_output_metadata_device(meta, layer_device)
+            layers[i]._seq_calib.output_meta = meta
+
+        d = _layer_dir(self.checkpoint_dir, last_ckpt)
+        next_inputs_path = os.path.join(d, "next_inputs.pt")
+        if not os.path.isfile(next_inputs_path):
+            raise FileNotFoundError(f"Cannot resume: next_inputs.pt missing for layer {last_ckpt}")
+        next_inputs = torch.load(next_inputs_path, map_location="cpu", weights_only=False)
```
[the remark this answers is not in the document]
realAsma: Added inline comment. Fixed in 43d1888b.

modelopt/torch/quantization/config.py
```diff
@@ -1217,16 +1217,26 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
...
 
+    checkpoint_dir: str | None = ModeloptField(
+        default=None,
+        title="Checkpoint directory for layerwise calibration.",
+        description=(
+            "If set together with use_layerwise=True, per-layer checkpoints are saved to this "
+            "directory during calibration. On restart, calibration resumes from the last "
+            "completed layer."
+        ),
+    )
```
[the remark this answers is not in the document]
realAsma: Added  that raises  if  is set but  is . Fixed in 43d1888b.

modelopt/torch/quantization/plugins/accelerate.py
```diff
@@ -50,32 +50,83 @@ def _get_cpu_offload_hook(hook):
...
+            continue
+        if isinstance(align_hook.weights_map, PrefixedDataset):
+            key = align_hook.weights_map.prefix + name
+            w_map = align_hook.weights_map.dataset.state_dict
+        else:
+            w_map = align_hook.weights_map
+            key = name
+        if key in w_map:
+            w_map[key] = param.data.to(w_map[key].device, dtype=w_map[key].dtype)
+
```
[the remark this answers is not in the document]
realAsma: Changed to  and  so both parameters and buffers are written back. Fixed in 43d1888b.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -0,0 +1,635 @@
...
+        if not checkpoint_dir:
+            return None
+        os.makedirs(checkpoint_dir, exist_ok=True)
+        info = detect_resume_point(checkpoint_dir)
+        start = info[0] if info else 0
+        if start > 0:
+            print_rank_0(
+                f"Checkpoint: resuming layerwise calibration from layer {start}/{num_layers}"
+            )
+        return cls(checkpoint_dir, num_layers, start_layer=start)
```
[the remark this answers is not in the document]
realAsma: Added  mismatch validation in  that raises  before any restore work. Fixed in 43d1888b.

modelopt/torch/utils/network.py
```diff
@@ -90,12 +90,43 @@ def is_parallel(model: nn.Module) -> bool:
...
+        return None
+
+    dev = getattr(hook, "execution_device", None)
+    if dev is not None:
+        return torch.device(dev)
+
+    for h in getattr(hook, "hooks", ()):
+        dev = getattr(h, "execution_device", None)
+        if dev is not None:
+            return torch.device(dev)
```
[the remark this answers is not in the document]
realAsma: Added  check with  for both the top-level hook and SequentialHook loop. Fixed in 43d1888b.

tests/gpu/torch/quantization/plugins/test_accelerate_gpu.py
```diff
@@ -73,3 +77,405 @@ def test_cpu_offloaded_tinyllama(tmp_path, quant_cfg):
...
+    device_map["model.layers.0"] = "cpu"
+    model_ref = load_checkpoint_and_dispatch(model_ref, tiny_llama_dir, device_map=device_map)
+    mtq.quantize(model_ref, seq_ckpt_cfg, lambda model: model(inputs))
+    output_ref = model_ref(inputs)
+
+    # Simulate crash after layer 0 by truncating the manifest
+    manifest_path = os.path.join(ckpt_dir, "manifest.json")
+    with open(manifest_path, "w") as f:
+        json.dump({"last_completed_layer": 0, "num_layers": num_layers}, f)
+
```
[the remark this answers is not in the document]
realAsma: Now removing stale layer directories above  after rewriting  in all three test locations. Fixed in 43d1888b.

examples/llm_ptq/example_utils.py
```diff
@@ -854,3 +855,35 @@ def copy_custom_model_files(source_path: str, export_path: str, trust_remote_cod
...
     else:
         print("No custom model files found to copy")
+
+
+def needs_checkpoint_path_update(quant_cfg: dict) -> bool:
+    """Check if quant_cfg has a checkpoint_dir that should be auto-resolved to a unique subpath."""
+    algorithm = quant_cfg.get("algorithm")
+    if algorithm is None or isinstance(algorithm, str):
+        return False
+    return algorithm.get("checkpoint_dir") is not None
```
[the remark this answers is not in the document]
realAsma: Changed guard to  to cover all non-mapping types. Fixed in 43d1888b.

realAsma (commented):
@sugunav14 **Note on `layerwise_calib.py`**: Most of this file is moved from `modelopt/torch/quantization/utils/activation_collector.py` (deleted in this PR). Git does not detect the rename because the file nearly doubled in size with new checkpoint/resume logic. To see the actual diff against the original, use:

```
git diff origin/main...HEAD -M10% -- modelopt/torch/quantization/utils/
```

This shows it as a rename with ~393 insertions and ~47 deletions, rather than a full 681-line new file + 335-line deletion.

meenchen (commented):
### 1. Design — Method Guard Removed

**[QUESTION] Calibration method guard removed with TODO**
The old code asserted that only `max` and `gptq` methods could be used with sequential calibration. This PR removes that assertion with a TODO: "add a method guard here." This means any calibration method (AWQ, SmoothQuant, etc.) can now be silently used with layerwise, even if it doesn't support per-layer invocation. What's the plan for re-adding validation?

### 2. Correctness — Checkpoint Corruption Recovery

**[SUGGESTION] Manifest corruption restarts silently from layer 0**
`_read_manifest()` returns `None` on corrupt/missing JSON, causing `from_folder()` to treat it as a fresh run. If a user's 50-layer calibration checkpoint is partially corrupted, they'd restart from scratch without knowing why. Add a warning when the manifest exists but can't be parsed — "Checkpoint manifest found but unreadable; starting from layer 0."

### 3. Correctness — _SkipLayer Proxy Masks AttributeErrors

**[SUGGESTION] _SkipLayer `__getattr__` catches all AttributeErrors**
```python
def __getattr__(self, name):
    try:
        return super().__getattr__(name)
    except AttributeError:
        return getattr(object.__getattribute__(self, "_original"), name)
```
If the original layer also doesn't have the attribute, the error message will reference `_original` instead of the skip layer — confusing for debugging. Consider catching only the first `super().__getattr__` and letting the second propagate naturally (which it does), but add context to the error.

### 4. Design — Distributed Checkpoint Blocked

**[SUGGESTION] Multi-rank checkpointing blocked at runtime, should fail at config time**
`_CheckpointState.__init__` raises `RuntimeError` if `dist.size() > 1`. Users discover this only after model loading and initial setup. Validate this in the config validator alongside `layerwise_checkpoint_dir` — if running distributed, reject the config early.

cjluo-nv (approved):
All critical comments from previous reviews have been addressed:

**Critical issues — all resolved:**
1. ✅ `weights_only=False` security: All 6 `torch.load` calls now have inline justification comments explaining the files are internally generated.
2. ✅ `_writeback_params_to_weights_map` now uses `state_dict(keep_vars=True)` to write back both parameters and buffers.
3. ✅ `from_folder()` validates `num_layers` mismatch between checkpoint manifest and current model.
4. ✅ Config validator `validate_layerwise_checkpoint_dir` rejects `layerwise_checkpoint_dir` when `layerwise=False`.
5. ✅ `_get_execution_device_from_hook` handles integer GPU ordinals with `isinstance(dev, int)`.
6. ✅ `needs_checkpoint_path_update` guards against non-dict algorithm values.
7. ✅ Tests now clean up stale layer directories above `last_completed_layer` during crash simulation.
8. ✅ `weight_access_and_writeback_context` now raises `RuntimeError` (instead of bare assert) for unsupported dual-hook layouts.
9. ✅ Last layer's dummy `output_meta` has a clear comment explaining it's a placeholder.
10. ✅ `_layer_forward_loop` uses default argument capture (`_inputs=layer_inputs`) for explicit binding.

**Design decisions accepted by reviewers:**
- Author pushed back on restoring previous-layer state on exception (no retry mechanism exists), and CodeRabbit agreed.
- Method guard removal: The `layerwise` field is on the base `QuantizeAlgorithmConfig`, so technically any method could set `layerwise=True`. However, the TODO is clearly marked, and in practice the typed config dispatch system means standard recipes won't hit this path for unsupported methods. This is acceptable risk with the TODO in place.

**Code quality:**
- Excellent test coverage: unit tests for skip/swap/restore, checkpoint save/resume, GPU integration tests for CPU-offloaded models, FSDP2, GPTQ combinations.
- The `_SkipLayer` pattern for replacing calibrated layers is clean and avoids framework hook overhead.
- The `persistent_materialization` context manager is well-designed for both FSDP2 and accelerate.
- The `_forward_loop` change to disable `use_cache` during calibration is a good correctness fix.
- The `calib_utils.py` change to handle `meta` device in `GPTQHelper.__init__` correctly supports CPU-offloaded models.

**PR size:** ~1900 lines is above the soft 1000-line guideline, but the changes are cohesive (rename + layerwise calibration + checkpoint + framework integration) and the author notes most of `layerwise_calib.py` is moved from the deleted `activation_collector.py`.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -0,0 +1,683 @@
...
+            # weights_only=False is safe: file is internally generated by _save_layer, not user-supplied
+            meta = torch.load(
+                os.path.join(d, "output_meta.pt"), map_location="cpu", weights_only=False
+            )
+            layer_device = get_module_device(layers[i])
+            meta = _remap_output_metadata_device(meta, layer_device)
+            layers[i]._layerwise_calib.output_meta = meta
+
+        d = _layer_dir(self.checkpoint_dir, last_ckpt)
+        next_inputs_path = os.path.join(d, "next_inputs.pt")
```
sugunav14, line 599: Are we going to be saving the input activations for a layer for resume? Is it feasible even for large calibration data samples/sequence lengths? Specifically in the case where almost full model is calibrated (all layerwise weights) + activations to the last layer: Is there a chance RAM would not be enough?

modelopt/torch/quantization/model_calib.py
```diff
@@ -1552,53 +1556,85 @@ def postprocess(module, name):
...
+                ckpt.save(layer_idx, layer, model, transformer_layers, next_inputs)
 
             del layer_inputs
             torch.cuda.empty_cache()
+            layer_inputs = next_inputs  # noqa: F841 (used in next iteration's closure)
     finally:
         input_getter._unpatch_all_layers()
 
-    print_rank_0("Sequential calibration completed")
+    if ckpt:
```
sugunav14, line 1649: Why do we need this final full_restore?

modelopt/torch/quantization/mode.py
```diff
@@ -237,17 +238,16 @@ def wrapped_calib_func(
                 module._moe_calib_experts_ratio = moe_calib_experts_ratio
 
     if func is not None:
-        if sequential:
+        if layerwise:
+            # TODO: add a method guard here — not all calib methods support per-layer invocation
```
sugunav14: Why is the assert removed?

modelopt/torch/quantization/config.py
```diff
@@ -1217,16 +1217,36 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
...
     )
 
-    use_sequential: bool = ModeloptField(
+    layerwise: bool = ModeloptField(
         default=False,
-        title="Enable sequential layer-by-layer calibration.",
+        title="Enable layerwise (layer-by-layer) calibration.",
         description=(
-            "If True, the calibration algorithm is applied sequentially to each decoder block. "
+            "If True, the calibration algorithm is applied to each decoder layer independently. "
```
shengliangxu: If I understand it correctly, it is the opposite of independently.

The quantization of the layers are sequentially dependent on the prior layers.

my understanding of "independent" would be the default behavior when all the quantizers collect stats as if it's the only quantizer in calibration.

And layerwise vs sequential, I feel sequential is a better word, unless I misunderstand, in addition, vllm compressor has the sequential calibration pipeline:

https://docs.vllm.ai/projects/llm-compressor/en/0.9.0/reference/llmcompressor/pipelines/sequential/

aligning the naming sounds reasonable to me. how do you think?

modelopt/torch/quantization/model_calib.py
realAsma: Good question.  is a no-op in the non-resume path ( returns early), so it only does work when resuming.\n\nDuring the hot calibration loop, already-calibrated layers are replaced in the ModuleList with  (parameter-free stand-in). In a fresh run, each layer's calibrated weights + quantizer state are kept on the original layer object and restored into the ModuleList in .\n\nOn resume, however, layers  were calibrated in a **previous process** \u2014 their quantized weights and quantizer state only live on disk, not in memory. The model loaded from HF at resume time has raw, unquantized weights for those layers.  walks those layers, reads  and  from disk, re-applies the quantizer state (so the layer structure has s in the right places), and loads the quantized weights \u2014 leaving the final model with all N layers properly quantized before export.\n\nWithout this step, a resumed run would produce an exported checkpoint where only layers  are quantized, and layers  are their original unquantized weights.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: Valid concern, flagging as a follow-up.\n\nTo clarify the current behavior:  is the list of captured  tuples for every calibration batch fed through the next layer \u2014 i.e.  entries, each holding a  tensor plus auxiliary kwargs (attention masks, position IDs, etc.). Only **one layer's worth** of activations is kept in RAM at a time (the previous layer's are freed before the next iteration). The on-disk  is only read during resume.\n\nFor the config in the PR description (DeepSeek-R1, , , , hidden=7168, bf16):  per layer's activations \u2014 manageable. But for much longer sequences or larger calib sets (e.g. 1024 samples \u00d7 4096 seq_len) this scales linearly and could become a bottleneck for both RAM and on-disk checkpoint size.\n\nMitigations to consider in a follow-up: (a) stream  batches through a generator instead of materializing a list, (b) checkpoint  with offload to pinned-CPU/disk between saves, (c) document a recommended upper bound on . I'll file this as a follow-up rather than expand scope here \u2014 the PR's target configs (DS-R1 at 64\u00d7512) sit well within the envelope.

modelopt/torch/quantization/config.py
realAsma: Two points in your comment \u2014 I agree with one and want to push back on the other.\n\n**Agreed \u2014 "independently" is misleading.** You're right: the wording contradicts the very next sentence ("each layer's inputs ... reflects the quantization of all preceding layers"). I'll fix the description to say something like "applied to one decoder layer at a time" or "applied layer by layer, with each layer's inputs reflecting the quantization of all preceding layers."\n\n**Pushback on the  vs  rename.** The rename to  is a deliberate choice documented in the PR description (item #3). Rationale:\n\n1. In this codebase,  previously meant something narrower ( only, one specific layer-by-layer flow). This PR broadens it to support GPTQ too and adds checkpoint resume \u2014 a renaming signals the behavioral change.\n2.  describes *what* the algorithm does (operates on one layer at a time);  could just as easily describe batch/sample sequencing, data pipelining, etc. For a config field surfaced to users,  is less ambiguous.\n3. vllm-compressor's "sequential" naming is one convention among many; TRT-Model-Optimizer's existing API surface already uses  terminology in internal docs and class names (, ). Aligning the user-facing flag with that internal naming is more valuable than aligning with an external library.\n\nI'll keep the  name but fix the docstring wording.

modelopt/torch/quantization/mode.py
realAsma: You're right \u2014 the TODO comment on the next line even admits this should have a method guard. I removed the old  because this PR broadens layerwise calibration to also support , but I should have replaced the assert with the new allowlist instead of dropping it.\n\nWill add:\n\n\n(Same concern was raised earlier by @cjluo-nv in #3081916767 \u2014 tracking as a fix here, not closing this thread until it's addressed.)

modelopt/torch/quantization/model_calib.py
realAsma: (My previous reply had code formatting stripped \u2014 re-posting with plain text.)\n\nGood question. full_restore is a no-op in the non-resume path (start_layer == 0 returns early), so it only does work when resuming.\n\nDuring the hot calibration loop, already-calibrated layers are replaced in the ModuleList with _SkipLayer (a parameter-free stand-in). In a fresh run, each layer's calibrated weights + quantizer state are kept on the original layer object and restored into the ModuleList in the finally-block _cleanup_layers() call. So after a full single-process run there is nothing to do at the end \u2014 all layers are already quantized in memory.\n\nOn resume, however, layers 0..start_layer-1 were calibrated in a **previous process** \u2014 their quantized weights and quantizer state only live on disk, not in memory. The model loaded from HF at resume time has raw, unquantized weights for those layers. full_restore walks those layers, reads weights.pt and quantizer_state.pt from disk, re-applies the quantizer state (so the layer structure has TensorQuantizer modules in the right places), and loads the quantized weights \u2014 leaving the final model with all N layers properly quantized before export.\n\nWithout this step, a resumed run would produce an exported checkpoint where only layers start_layer..N-1 are quantized, and layers 0..start_layer-1 are their original unquantized weights.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: (My previous reply had code formatting stripped \u2014 re-posting with plain text.)\n\nValid concern, flagging as a follow-up.\n\nTo clarify the current behavior: next_inputs is the list of captured (args, kwargs) tuples for every calibration batch fed through the next layer \u2014 i.e. calib_size / batch_size entries, each holding a (batch_size, seq_len, hidden) tensor plus auxiliary kwargs (attention masks, position IDs, etc.). Only **one layer's worth** of activations is kept in RAM at a time (the previous layer's list is freed before the next iteration). The on-disk next_inputs.pt is only read during resume.\n\nFor the config in the PR description (DeepSeek-R1, calib_size=64, batch_size=16, calib_seq=512, hidden=7168, bf16): 64 * 512 * 7168 * 2 bytes is about 470 MB per layer's activations \u2014 manageable. But for much longer sequences or larger calib sets (e.g. 1024 samples \u00d7 4096 seq_len) this scales linearly and could become a bottleneck for both RAM and on-disk checkpoint size.\n\nMitigations to consider in a follow-up: (a) stream next_inputs batches through a generator instead of materializing a list, (b) checkpoint next_inputs with offload to pinned-CPU/disk between saves, (c) document a recommended upper bound on calib_size * calib_seq. I'll file this as a follow-up rather than expand scope here \u2014 the PR's target configs (DS-R1 at 64\u00d7512) sit well within the envelope.

modelopt/torch/quantization/config.py
realAsma: (My previous reply had code formatting stripped \u2014 re-posting with plain text.)\n\nTwo points in your comment \u2014 I agree with one and want to push back on the other.\n\n**Agreed \u2014 "independently" is misleading.** You're right: the wording contradicts the very next sentence ("each layer's inputs ... reflects the quantization of all preceding layers"). I'll fix the description to say something like "applied to one decoder layer at a time" or "applied layer by layer, with each layer's inputs reflecting the quantization of all preceding layers."\n\n**Pushback on the sequential vs layerwise rename.** The rename to 'layerwise' is a deliberate choice documented in the PR description (item #3). Rationale:\n\n1. In this codebase, sequential_calibrate previously meant something narrower ('max' only, one specific layer-by-layer flow). This PR broadens it to support GPTQ too and adds checkpoint resume \u2014 a rename signals the behavioral change.\n2. 'layerwise' describes *what* the algorithm does (operates on one layer at a time); 'sequential' could just as easily describe batch/sample sequencing or data pipelining. For a config field surfaced to users, 'layerwise' is less ambiguous.\n3. vllm-compressor's "sequential" naming is one convention among many; TRT-Model-Optimizer's existing API surface already uses 'layerwise' terminology in internal docs and class names (LayerActivationCollector, _layerwise_calib). Aligning the user-facing flag with the codebase's own internal naming is more valuable than aligning with an external library.\n\nI'll keep the 'layerwise' name but fix the docstring wording.

modelopt/torch/quantization/mode.py
realAsma: (My previous reply had code formatting stripped \u2014 re-posting with plain text.)\n\nYou're right \u2014 the TODO comment on the next line even admits this should have a method guard. I removed the old assert that checked method in ["max"] because this PR broadens layerwise calibration to also support gptq, but I should have replaced the assert with the new allowlist instead of dropping it.\n\nWill add an assert of the form:\n\n    assert method in ["max", "gptq"], f"Layerwise calibration currently only supports max and gptq, got {method}"\n\n(Same concern was raised earlier by @cjluo-nv in #3081916767 \u2014 tracking as a fix here, not closing this thread until it's addressed.)

modelopt/torch/quantization/config.py
sugunav14: Just FYI, previously, sequential_calibrate wasn't narrower/limited only to max. It supported GPTQ too

modelopt/torch/quantization/mode.py
realAsma: Intentional \u2014 layerwise calibration is designed as a general mechanism that should work out of the box with any PTQ algorithm, not just /. All currently implemented algorithms (max, gptq, awq, smoothquant, svdquant, mse) are compatible with per-layer invocation, so an allowlist assert would be unnecessarily restrictive and would block valid usage.\n\nI've replaced the  with an explanatory comment documenting this contract: a guard is only needed if a future algorithm requires full-model context at calibration time.

realAsma: (Correction \u2014 previous reply had shell-mangled wording.) Intentional: layerwise calibration is designed as a general mechanism that should work out of the box with any PTQ algorithm, not just max/gptq. All currently implemented algorithms (max, gptq, awq, smoothquant, svdquant, mse) are compatible with per-layer invocation, so an allowlist assert would be unnecessarily restrictive and would block valid usage. I've replaced the TODO with an explanatory comment documenting this contract: a guard is only needed if a future algorithm requires full-model context at calibration time.

modelopt/torch/quantization/config.py
realAsma: Good catch on the wording \u2014 fixed the docstring in e9840ade. Reworded to "applied layer by layer" to avoid the "independently" / sequentially-dependent contradiction. On the naming: keeping "layerwise" for consistency with the surrounding code (LayerActivationCollector, _layerwise_calib); happy to revisit in a follow-up if the team prefers to align with llm-compressor's "sequential" term.

modelopt/torch/quantization/mode.py
sugunav14: I don't think it is safe to allow all algorithms given that we haven't tested them with sequential flow yet. I think its still best to leave the assert as is.

realAsma: Fair point \u2014 I'll restore the allowlist assert scoped to the methods this PR has actually exercised ( and ). Will push the fix shortly.

realAsma: (Correction \u2014 previous reply had shell-mangled wording.) Fair point. I will restore the allowlist assert, scoped to the methods this PR has actually exercised: max and gptq. Fix coming in the next push.

realAsma: On reflection, keeping the generic comment \u2014 layerwise is designed to be general across PTQ algorithms; adding an allowlist would be unnecessarily restrictive. Retracting the earlier commitment to restore the assert.

modelopt/torch/quantization/config.py
realAsma: Sorry for the bot comment. Will work on it to give useful replies.

realAsma: @shengliangxu  I had consulted with Suguna and other on this. I felt Sequential was confusing - here we are sequentially iterative decoder block by block.

realAsma: @shengliangxu  @sugunav14  Do you feel sequential was better - let me know. I will revert.

modelopt/torch/quantization/mode.py
realAsma: @sugunav14  All the algorithms we have in ModelOpt (Max, GPTQ, mse, svdquant , awq etc are compatible with sequential layer by layer quantization. I can add unittests for all these algorithms with layer wise calibration. 

Will you be okay with that?

sugunav14: Yeah, I think having unit tests will help. It should work for other algos, given that each algo has an initial call to max_calibrate to set the scales.

modelopt/torch/quantization/config.py
sugunav14: I don't have a strong preference on the naming. Either works for me.

examples/llm_ptq/example_utils.py
realAsma, line 1: @shengliangxu  I am overwriting the layer wise save from the yaml recipe this way. Does this look goo to you?

modelopt/torch/export/plugins/vllm_fakequant_hf.py
realAsma, line 1: @kinjalpatel27  could you please review this file? I had to add inlace memory update for large models (DSR1 weights cost 1.3 TB CPU memory, Kimi cost > 2 TB CPU memory - without in place memory efficient single node export could crash due to CPU OOM)

realAsma:
@meenchen @sugunav14 — added `_supports_layerwise` (per-mode opt-out) on `BaseCalibrateModeDescriptor` in `modelopt/torch/quantization/mode.py`. Defaults to `True`; `SVDQuantModeDescriptor` opts out. `wrapped_calib_func` now raises a clear `ValueError` when `layerwise=True` is requested on an unsupported mode. See commit `baaf80f4`.

modelopt/torch/export/plugins/vllm_fakequant_hf.py
kinjalpatel27: These changes LGTM. Thank you

sugunav14 (approved):

sychen52 (approved):

modelopt/torch/export/plugins/vllm_fakequant_hf.py
```diff
@@ -38,9 +40,75 @@ def disable_rotate(quantizer: TensorQuantizer):
...
+        else:
+            assert state_dict is not None
+            state_dict[sd_key] = w_quant.cpu()
+        fakequant_weights.add(sd_key)
+
+
 def export_hf_vllm_fq_checkpoint(
     model: nn.Module,
     export_dir: Path | str,
+    inplace_mem_efficient: bool = False,
```
kinjalpatel27, line 111: Can you extend the tests/gpu/torch/export/test_vllm_fakequant_hf_export.py to test for offload?

realAsma: Added `test_hf_vllm_export_offload` in c9b7f827 — exercises the `inplace_mem_efficient=True` path on a CPU-offloaded tiny LLaMA (layer 0 on CPU via accelerate), asserts the inplace path actually mutates offloaded weights, and verifies reloaded HF weights match a deepcopy+fold_weight reference.

Edwardf0t1 (approved):
LGTM in general. @realAsma It would be great to test GLM5.1 as well. cc @Fridah-nv 

There's a draft PR for GLM5/5.1:
https://github.com/NVIDIA/Model-Optimizer/pull/985

modelopt/torch/export/plugins/vllm_fakequant_hf.py
realAsma, line 230: @kinjalpatel27  why are we doing this specially for quantize mode?

kinjalpatel27: we need to remove disabled weight_quantizer from metadata, otherwise the reload creates an issue.

PR #1256, opened 2026-04-14
Exclude small-k and small-n Matmul nodes from Int8 quantization

nv-samcheng:
### What does this PR do?
Exclude small-dimension MatMul nodes from INT8 quantization. MatMuls with N or K < 16 cannot efficiently use INT8, causing performance regressions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quantization exclusions so MatMul/Gemm ops with derived K<16 or N<16 are skipped, honoring Gemm transB, using inferred and runtime-determined shapes, and avoiding duplicate outputs.

* **Tests**
  * Expanded unit tests to cover constant, inferred, and runtime-derived shapes, Gemm transB behavior, small-dimension edge cases, and output deduplication.

* **Documentation**
  * Added changelog entry documenting the new small-dimension exclusion thresholds and transB handling.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ Changelog
 - Fix Megatron utility functions for generation (with pipeline parallelism) and ~10x speedup in MMLU score evaluation (by batching prefill passes).
 - Fix Minitron pruning (``mcore_minitron``) for MoE models. Importance estimation hooks were incorrectly registered for MoE modules and NAS step was hanging before this.
 - Fix TRT support for remote autotuning in ONNX Autotune from 10.16+ to 10.15+ and fix TRT versioning check to the ``trtexec`` version instead of the TRT Python API when using ``trtexec`` backend.
+- Exclude MatMul/Gemm nodes with K or N < 16 from ONNX INT8 and FP8 quantization. Such small-dimension GEMMs cannot efficiently use INT8/FP8 Tensor Cores and the added Q/DQ layers cause perf regressions in TensorRT. Honors Gemm ``transB`` when deriving K.
 
 **Misc**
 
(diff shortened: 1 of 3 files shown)


ajrasane (changes requested):
## Review: Exclude small-k and small-n Matmul nodes from Int8 quantization

Good change overall — the motivation is clear and the implementation is clean. A few items to address before merging:

### Issues

**1. Missing `transB` handling for Gemm nodes (Medium)**
`find_nodes_from_matmul_to_exclude` collects both `MatMul` and `Gemm` nodes (line 1116: `node.op in {"MatMul", "Gemm"}`). `_get_inp_b_k_dim` always reads K from axis `[-2]`, which is correct for MatMul (`[K, N]`) but **wrong for Gemm with `transB=1`** where B is `[N, K]` and K is at axis `[-1]`. This could cause:
- **False negatives:** a Gemm with small K but large N would read N as K and skip exclusion
- **False positives:** a Gemm with large K but small N would read N as K and exclude incorrectly

Suggested fix: detect `transB` attribute on the node and set `k_axis = -1 if transB > 0 else -2`, then use `k_axis` across all three fallback paths in `_get_inp_b_k_dim`.

**2. Small-gemm check applies unconditionally for all quantize modes (Low-Medium)**
The new small-gemm check fires for all invocations of `find_nodes_from_matmul_to_exclude`, but the threshold `_MIN_MATMUL_DIM_INT8 = 16` is specifically for INT8 (as the name implies). The calling context may invoke this for FP8 quantization too. Consider either:
- Gating the check behind a `quantize_mode` parameter (similar to how `find_nodes_from_convs_to_exclude` does it)
- Or documenting explicitly that this is intentionally applied to all modes

**3. Tests only cover shape-inference path (Low)**
All new tests exercise `_exclude_matmuls_by_shape_inference`. There are no tests for `_exclude_matmuls_by_inference` (the runtime inference path). The runtime path has the same logic but uses `output_map` — adding at least one test would increase confidence.

ajrasane (approved):

PR #1263, opened 2026-04-14
[OMNIML-3349] Add ResNet50 support for torch_onnx quantization workflow

ajrasane:
## Summary
- Add end-to-end ResNet50 support in the torch_onnx quantization → ONNX export → TRT engine pipeline
- Fix multiple Conv2d-related export issues that blocked Conv2d-heavy models from working with FP8/INT8/MXFP8/NVFP4/auto quantization modes
- Fix `configure_linear_module_onnx_quantizers` to handle all modules with block quantization (not just `nn.Linear`), fixing NVFP4/MXFP8 export for models with quantized non-Linear modules
- Add `--trt_build` flag to `torch_quant_to_onnx.py` and simplify test infrastructure

### Files Changed
- `modelopt/torch/_deploy/utils/torch_onnx.py` — Disable FP8 Conv2d weight quantizers and autocast during ONNX export
- `modelopt/torch/quantization/export_onnx.py` — Fix `configure_linear_module_onnx_quantizers` for all module types with block quantization
- `examples/torch_onnx/torch_quant_to_onnx.py` — Add `--trt_build` flag, calibration for FP8 override quantizers, Conv2d→FP8 override for auto mode, filter_func updates
- `examples/torch_onnx/README.md` — Add ResNet50 to supported models table
- `tests/examples/torch_onnx/test_torch_quant_to_onnx.py` — Add ResNet50 test entry, simplify using `--trt_build`
- `tests/_test_utils/torch/vision_models.py` — Add ResNet50 to timm model registry

### Quantization modes passing
- ✅ FP8, INT8, MXFP8, NVFP4, Auto (all 5 modes pass export + TRT build)
- INT4_AWQ excluded (pre-existing limitation for all models)

## Test plan
- [x] All 5 resnet50 test modes pass: `pytest tests/examples/torch_onnx/test_torch_quant_to_onnx.py -k resnet50` (5/5 passed)
- [x] Full regression: 18 passed, 2 failed (pre-existing swinv2_tiny fp8/int8 failures)

## Summary by CodeRabbit

* **New Features**
  * Added ResNet50 to supported ONNX export vision models with FP8, INT8, MXFP8, and NVFP4 support.
  * Optional TensorRT engine build after export via a new CLI flag.

* **Improvements**
  * Enhanced quantization calibration and export flows for FP8/INT8 models, including broader block-quantization support across module types and safer export handling.
  * Tests updated to include ResNet50 in the model matrix.


gcunhase:
Does this quantized model perform similarly to the same model quantized via the ONNX path? The reason I'm asking is because previous experiments showed that the Torch path does not add Q/DQ nodes in the Residual branches, causing perf regression when compared to its ONNX counterpart.

cjluo-nv (approved):
The PR cleanly adds ResNet50 support to the torch_onnx quantization workflow with well-considered fixes:

1. **`configure_linear_module_onnx_quantizers` fix** (export_onnx.py): Correctly generalizes from `isinstance(module, nn.Linear)` to checking `block_sizes` on any module with quantizers. This is more precise (only sets export attributes when block quantization is actually used) and handles non-Linear modules like MaxPool2d that get NVFP4/MXFP8 input quantizers.

2. **FP8 Conv weight quantizer disable** (torch_onnx.py): Well-implemented context manager with proper try/finally cleanup. The issue (TorchScript exporter needs static kernel shapes, FP8 DequantizeLinear produces dynamic shapes) is clearly documented.

3. **Autocast disable for FP8/INT8**: Correct — these quantized models should not use autocast during export since it could interfere with the QDQ patterns.

4. **`_calibrate_uncalibrated_quantizers`**: Clean solution for the gap where MXFP8/NVFP4 pipelines skip standard calibration but FP8 Conv overrides need it. The detection heuristic (enabled, no block_sizes, no _amax) is sound.

5. **Test simplification**: Moving TRT build into the main script via `--trt_build` reduces test boilerplate and ensures the build logic is available for non-test use too.

6. **PR size**: 159 additions, 58 deletions — well-scoped and cohesive.

All 5 quant modes tested for ResNet50 (fp8, int8, mxfp8, nvfp4, auto).

gcunhase:
> 3. **Autocast disable for FP8/INT8**: Correct — these quantized models should not use autocast during export since it could interfere with the QDQ patterns.

Doesn't the high-precision tensors will be in FP32 instead of FP16? This will cause perf regressions when using `--stronglyTyped`.

ajrasane:
> Doesn't the high-precision tensors will be in FP32 instead of FP16? This will cause perf regressions when using --stronglyTyped.

We are using the convert_float_to_float16 API as the autocast API fails for these precisions. I will put out a follow up PR to make sure that we can run all the precisions with the AutoCast API.

PR #1265, opened 2026-04-15
Handle zero-amax per-channel activation scaling for MoE export

AEON-7:
## What

`NVFP4QTensor.get_activation_scaling_factor` asserts:

```python
assert torch.all(activation_scaling_factor > 0), (
    f" activation scaling factor {activation_scaling_factor} not positive."
)
```

On MoE models, some per-channel activation amax entries are exactly zero because routing sparsity leaves certain input slots on rarely-routed experts un-activated during calibration. The derived scaling factor (`amax / (maxbound * 448)`) is then zero and the assertion trips.

## How to reproduce

Any MoE model with per-expert-decomposed linears quantized using `NVFP4_AWQ_FULL_CFG`. On SuperGemma4 26B (128 experts, ~6% activation rate per expert per token), this fires on the first expert whose calibration-time coverage left even a single channel dark. It is the routine case, not the edge case.

## The fix

Detect zero entries in the computed `activation_scaling_factor` tensor and replace them with the minimum positive value in the same tensor via `torch.where`. Fall back to a small positive floor (`1e-8`) for the pathological case where every channel in the tensor is zero (block entirely un-activated).

## Why this is numerically safe

A zero amax channel means no activation was ever observed there during calibration. Any value flowing through that channel at inference is therefore statistically near-zero relative to the observed distribution. Scaling that near-zero value by the "quietest live channel's" scaling factor quantizes it to near-zero and dequantizes back to near-zero — the same end result as a genuinely zero scale, minus the NaN/division hazards.

The assertion after the fix remains strict (`torch.all(scale > 0)`), so downstream code that relies on the positivity invariant is unaffected.

## Validation

End-to-end on SuperGemma4 26B (Gemma 4 MoE, 128 experts, per-expert-decomposed plugin) with `NVFP4_AWQ_FULL_CFG`:

- **Before**: `AssertionError:  activation scaling factor tensor([...]) not positive.` on a per-channel tensor whose printed head hides zeros in the `...` ellipsis.
- **After**: scaling factors land strictly positive; export produces a valid NVFP4 checkpoint. Side-by-side sampled generation (seed=42, temperature=0.7, top_p=0.9) against the BF16 baseline matches on fact-recall ("The capital of France is Paris." — identical) and produces coherent, well-formed outputs on creative (haiku) and technical (neural network explanation) prompts.

The resulting quantized model ships at [AEON-7/supergemma4-26b-abliterated-multimodal-nvfp4](https://huggingface.co/AEON-7/supergemma4-26b-abliterated-multimodal-nvfp4).

## Companion PR

Depends on / pairs with #1264 (non-scalar input amax in `preprocess_linear_fusion`). Both are orthogonal bugs on the same NVFP4 + per-expert-MoE export path; this PR fixes the bug that fires *after* #1264's fix unblocks the fusion step.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an edge-case in quantization where exact-zero scaling factors could occur; those zeros are now replaced with safe positive defaults to prevent downstream dequantization errors and ensure stable, correct behavior in rare inputs.


AEON-7:
Good catch — you're right, `<= 0` would let negative entries (which only arise from upstream bugs, not routing sparsity) be silently "repaired" alongside genuine zeros, masking real problems.

Fixed in `c6edb1685f`: `zero_mask` now matches exactly-zero entries only. Negative entries are left untouched so the existing positivity assertion below still catches them. Also tightened the `positive` selection to filter on `> 0` rather than `~zero_mask`, so the replacement value can never be negative even if one somehow slipped through. The all-zero fallback path is preserved.

Diff:
```python
-        zero_mask = activation_scaling_factor <= 0
+        zero_mask = activation_scaling_factor == 0
         if zero_mask.any():
-            positive = activation_scaling_factor[~zero_mask]
-            if positive.numel() > 0:
-                activation_scaling_factor = torch.where(
-                    zero_mask, positive.min(), activation_scaling_factor
-                )
-            else:
-                activation_scaling_factor = torch.full_like(
-                    activation_scaling_factor, 1e-8
-                )
+            positive = activation_scaling_factor[activation_scaling_factor > 0]
+            replacement = (
+                positive.min()
+                if positive.numel() > 0
+                else torch.tensor(
+                    1e-8,
+                    device=activation_scaling_factor.device,
+                    dtype=activation_scaling_factor.dtype,
+                )
+            )
+            activation_scaling_factor = torch.where(
+                zero_mask, replacement, activation_scaling_factor
+            )
```

Thanks for the review.

meenchen:
@AEON-7 could you check if the issue can be reproduced on the main? We have fixed similar issues with MoE calibration

PR #1268, opened 2026-04-15
Centralize 'trtexec' subprocess runs in ONNX into a single function

gcunhase:
### What does this PR do?

Type of change: code improvement

Subprocess is needed to run `trtexec` commands but those require `nosec` approval. This PR centralizes it into a single function for the ONNX workflow to avoid future approval triggers. The torch workflow has that centralized in `run_command` (renamed to `_run_trtexec_streamed`).

Original request: https://github.com/NVIDIA/Model-Optimizer/pull/1259#issuecomment-4253630765

### Usage

```python
results = _run_trtexec(cmd)
```

### Testing
N/A

## Summary by CodeRabbit

* **Refactor**
  * Unified how the app runs and logs the external conversion/benchmarking tool across build, profile, and benchmark flows for consistent behavior and clearer logs.
* **Behavior Change**
  * The tool is now invoked by name from PATH (configurable path option removed).
* **Bug Fixes**
  * Missing-tool errors now instruct to ensure the external tool is available in PATH.
* **Tests**
  * Tests updated to match the new execution/logging entrypoints.

### DIFF
--- a/modelopt/onnx/quantization/autotune/benchmark.py
+++ b/modelopt/onnx/quantization/autotune/benchmark.py
@@ -185,7 +185,6 @@ class TrtExecBenchmark(Benchmark):
         self.latency_pattern = r"\[I\]\s+Latency:.*?median\s*=\s*([\d.]+)\s*ms"
 
         self._base_cmd = [
-            self.trtexec_path,
             f"--avgRuns={self.timing_runs}",
             f"--iterations={self.timing_runs}",
             f"--warmUp={self.warmup_runs}",
--- a/modelopt/onnx/quantization/autotune/benchmark.py
+++ b/modelopt/onnx/quantization/autotune/benchmark.py
@@ -31,7 +31,6 @@ import importlib.util
 import os
 import re
 import shutil
-import subprocess  # nosec B404
 import tempfile
 import time
 from abc import ABC, abstractmethod
@@ -42,7 +41,7 @@ import numpy as np
 import torch
 
 from modelopt.onnx.logging_config import logger
-from modelopt.onnx.quantization.ort_utils import _check_for_trtexec
+from modelopt.onnx.quantization.ort_utils import _check_for_trtexec, _run_trtexec
 
 TRT_AVAILABLE = importlib.util.find_spec("tensorrt") is not None
 if TRT_AVAILABLE:
@@ -269,7 +268,7 @@ class TrtExecBenchmark(Benchmark):
 
             cmd = [*self._base_cmd, f"--onnx={model_path}"]
             self.logger.debug(f"Running: {' '.join(cmd)}")
-            result = subprocess.run(cmd, capture_output=True, text=True)  # nosec B603
+            result = _run_trtexec(cmd)
             self._write_log_file(
                 log_file,
                 "\n".join(
--- a/modelopt/onnx/quantization/autotune/benchmark.py
+++ b/modelopt/onnx/quantization/autotune/benchmark.py
@@ -158,7 +158,6 @@ class TrtExecBenchmark(Benchmark):
         warmup_runs: int = 5,
         timing_runs: int = 10,
         plugin_libraries: list[str] | None = None,
-        trtexec_path: str = "trtexec",
         trtexec_args: list[str] | None = None,
     ):
         """Initialize the trtexec benchmark.
@@ -168,14 +167,11 @@ class TrtExecBenchmark(Benchmark):
             warmup_runs: See :meth:`Benchmark.__init__`.
             timing_runs: See :meth:`Benchmark.__init__`.
             plugin_libraries: See :meth:`Benchmark.__init__`.
-            trtexec_path: Path to trtexec binary. Defaults to 'trtexec' which
-                         looks for the binary in PATH.
             trtexec_args: Additional command-line arguments to pass to trtexec.
                          These are appended after the standard arguments.
                          Example: ['--fp16', '--workspace=4096', '--verbose']
         """
         super().__init__(timing_cache_file, warmup_runs, timing_runs, plugin_libraries)
-        self.trtexec_path = trtexec_path
         self.trtexec_args = trtexec_args if trtexec_args is not None else []
         self.temp_dir = tempfile.mkdtemp(prefix="trtexec_benchmark_")
         self.engine_path = os.path.join(self.temp_dir, "engine.trt")
@@ -299,8 +295,9 @@ class TrtExecBenchmark(Benchmark):
             self.logger.info(f"TrtExec benchmark (median): {latency:.2f} ms")
             return latency
         except FileNotFoundError:
-            self.logger.error(f"trtexec binary not found: {self.trtexec_path}")
-            self.logger.error("Please ensure TensorRT is installed and trtexec path is correct")
+            self.logger.error(
+                "'trtexec' binary not found. Please ensure TensorRT is installed and 'trtexec' is in PATH."
+            )
             return float("inf")
         except Exception as e:
             self.logger.error(f"Benchmark failed: {e}")
--- a/modelopt/onnx/quantization/autotune/benchmark.py
+++ b/modelopt/onnx/quantization/autotune/benchmark.py
@@ -262,13 +262,14 @@ class TrtExecBenchmark(Benchmark):
                 self.logger.debug(f"Wrote model bytes to temporary file: {model_path}")
 
             cmd = [*self._base_cmd, f"--onnx={model_path}"]
-            self.logger.debug(f"Running: {' '.join(cmd)}")
+            full_cmd = ["trtexec", *cmd]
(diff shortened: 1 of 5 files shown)


modelopt/onnx/quantization/ort_utils.py
```diff
@@ -44,6 +44,15 @@ def _check_lib_in_ld_library_path(ld_library_path, lib_pattern):
     return False, None
 
 
+def _run_trtexec(cmd, timeout=None):
+    """Run a 'trtexec' command via subprocess."""
+    # Ensure that this command is a trtexec run
+    assert any("trtexec" in c for c in cmd), "Subprocess can only execute 'trtexec' commands"
```
gcunhase: @kevalmorabia97 please let me know if this is enough to ensure only `trtexec` runs or if this is even necessary. Thanks.

kevalmorabia97: Can we not pass `trtexec` in `cmd` and only pass arguments to `trtexec` so this function prepends `trtexec` to the `cmd`? 

Please add type annotation also

modelopt/onnx/quantization/ort_utils.py
kevalmorabia97, line 1: Can you unify with https://github.com/NVIDIA/Model-Optimizer/blob/07ae8e71281a1c17898991b0cd76db5788f490bb/modelopt/torch/_deploy/_runtime/tensorrt/engine_builder.py#L62 also

gcunhase: I was trying to avoid coupling `modelopt/torch` with `modelopt/onnx`. What do you think is best?

kevalmorabia97: `modelopt.torch._deploy` already has some dependencies on `modelopt.onnx` so its fine. We can keep this in `modelopt.onnx.trt_utils` or whatever names makes sense


For example this: https://github.com/NVIDIA/Model-Optimizer/blob/07ae8e71281a1c17898991b0cd76db5788f490bb/modelopt/torch/_deploy/_runtime/trt_client.py

gcunhase: Good point, will look into this today.

modelopt/onnx/quantization/ort_utils.py
gcunhase: done, please check

modelopt/onnx/quantization/ort_utils.py
gcunhase: Hi Keval, I've kept the ONNX and torch `trtexec` functions separate for now for a few reasons:
- I/O capture: In-memory vs Streamed to NamedTemporaryFile on disk
- Return type: subprocess.CompletedProcess vs tuple[int, bytes] (returncode, combined output)
- Use case: Short calls (version probe, trtexec with no args, 5s timeout) vs Long calls (engine build / profile (large logs, no fixed timeout).

Intentional split:
- For the version probe, in-memory + timeout is correct — the output is tiny and you want to bail if trtexec hangs.
- For engine builds, in-memory would buffer megabytes of verbose output in RAM, you want it on disk; failures need to surface to the user via logging; build time is unbounded so a fixed timeout is wrong.                                                                                                                                                                                             

I've modified both function to only run `trtexec` though, so no other commands can exploit this function.

Let me know if that makes sense or if merging them is still preferable and I'll look into it.

Thanks!

modelopt/torch/_deploy/_runtime/tensorrt/engine_builder.py
```diff
@@ -41,25 +40,29 @@
 )
 
 
-# TODO: Get rid of this function or get approval for `# nosec` usage if we want to include this
-#   as a non-compiled python file in the release.
-def _run_command(cmd: list[str], cwd: Path | None = None) -> tuple[int, bytes]:
-    """Util function to execute a command.
+def _run_trtexec_streamed(args: list[str], cwd: Path | None = None) -> tuple[int, bytes]:
```
kevalmorabia97: can you move this also to modelopt/onnx/quantization/ort_utils.py?
Maybe name it modelopt/onnx/quantization/trtexec_utils.py

modelopt/torch/_deploy/_runtime/tensorrt/engine_builder.py
kevalmorabia97: can you move this `import subprocess` in the function also?

modelopt/torch/_deploy/_runtime/tensorrt/engine_builder.py
```diff
@@ -41,25 +40,29 @@
...
+        A tuple of (returncode, output) where output is the combined stdout/stderr bytes.
     """
+    cmd = ["trtexec", *args]
     logging.info(" ".join(cmd))
     with NamedTemporaryFile("w+b") as log:
-        p = subprocess.Popen(cmd, stdout=log, stderr=log, cwd=str(cwd) if cwd else None)  # nosec
+        p = subprocess.Popen(  # nosec B603 - cmd[0] is hardcoded "trtexec"
+            cmd, stdout=log, stderr=log, cwd=str(cwd) if cwd else None
+        )
         p.wait()
```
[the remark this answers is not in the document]
gcunhase: @ajrasane what do you think about this comment? This is related to Torch TRT engine building.

modelopt/torch/_deploy/_runtime/tensorrt/engine_builder.py
gcunhase: Done

modelopt/torch/_deploy/_runtime/tensorrt/engine_builder.py
gcunhase: Is this still needed? I see it's been marked as resolved.

kevalmorabia97 (approved):

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -42,7 +41,7 @@
 import torch
 
 from modelopt.onnx.logging_config import logger
-from modelopt.onnx.quantization.ort_utils import _check_for_trtexec
+from modelopt.onnx.quantization.ort_utils import _check_for_trtexec, _run_trtexec
```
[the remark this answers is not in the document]
gcunhase: Centralizing 'nosec' commands. This exception will be allowed.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -269,7 +263,7 @@ def run(
 
             cmd = [*self._base_cmd, f"--onnx={model_path}"]
             self.logger.debug(f"Running: {' '.join(cmd)}")
-            result = subprocess.run(cmd, capture_output=True, text=True)  # nosec B603
+            result = _run_trtexec(cmd)
```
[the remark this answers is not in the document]
gcunhase: Resolved

modelopt/torch/_deploy/_runtime/tensorrt/engine_builder.py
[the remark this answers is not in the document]
ajrasane: I think it makes sense to add a timeout. Should we add a timeout of 30 minutes? I don't think any of the trt engine builds for the models we support takes longer than this.

gcunhase:
@ajrasane can you please check the modifications in `engine_builder.py` and see if you agree?

I've listed the advantages of `_run_trtexec_with_logging` over `_run_trtexec_streamed` below:

1. **No tempfile.** Dropped NamedTemporaryFile and the with block. Output stays in memory via subprocess.PIPE instead of being written to disk and read back — peak memory is approximately the same since both versions ultimately materialize the full output as bytes.
     
2. **Automatic process cleanup on timeout.** `subprocess.run(timeout=3600)` kills and reaps the child internally before raising TimeoutExpired. The streamed version had to do `p.kill(); p.wait()` manually to avoid orphan trtexec processes holding GPU memory — easy to get wrong.

3. **Output preserved on timeout, for free.** TimeoutExpired.stdout carries whatever was captured before the kill. The streamed version had to remember to `log.seek(0); log.read()` before the tempfile context exited (otherwise the file got deleted and the captured output was lost forever).
                                                     
4. **Less code, fewer footguns.** Gone: NamedTemporaryFile import, Popen setup, wait(timeout), kill()/wait() cleanup, two log.seek(0)/log.read() blocks. ~15 lines down to ~10.

5. The **name was changed** from `_streamed` to `_with_logging` to better describe what it does after the modification.
                                                                                                                                                                                                      
6. **FileNotFoundError handled consistently.** Same wrapped-with-helpful-message pattern as _run_trtexec, so users get the same actionable error from both helpers.

**The original behavior is still preserved**: 60-min timeout, cwd support, INFO logging of cmd, ERROR logging of output on non-zero returncode or timeout, (returncode, bytes) return shape, hardcoded "trtexec" security boundary.                                                           

Let me know if that sounds good.

ajrasane (approved):

PR #1269, opened 2026-04-15
Skip Softmax diffusion export

jingyu-ml:
### What does this PR do?

Type of change: New Feature 

Adds HuggingFace `config.json` export of skip-softmax sparse-attention calibration for diffusion pipelines (e.g. Wan 2.2), on top of the base skip-softmax work.

- **`_export_diffusers_checkpoint`** walks every `nn.Module` component of a diffusers pipeline, calls `export_sparse_attention_config`, and writes the result into that component's `config.json` under the `sparse_attention_config` key. The sparse config lives **only** in `config.json` — there is no standalone `sparse.yaml`.
- **`export_sparse_attention_config`** emits a `config_groups` schema where each algorithm's parameters are nested inside its own group; only `config_groups` and `producer` are top-level:
  - skip-softmax group → `algorithm: "skip_softmax"`, `targets`, `ignore` (layers kept dense — e.g. cross-attention + first/last blocks), `initial_disabled_steps` (opt-in, user-set; emitted only when `> 0`), `threshold_scale_factor` (`a * exp(b * target_sparsity)`), and `target_sparsity`.
  - N:M group → `algorithm: "sparse_softmax"` with `sparsity_n`/`sparsity_m`, `dense_sink_tokens`, `dense_recent_tokens` flattened into the group.
- **Deploy reader** (`modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_config.py`) reads these per-group params back, keeping the export↔load round-trip consistent.
- **Example wiring**: `examples/diffusers/sparsity/wan22_skip_softmax.py` gains `--export-dir`, `--skip-softmax-threshold`, and `--initial-disabled-steps`. `--export-dir` runs `export_hf_checkpoint(pipe, export_dir=...)` after calibration.
- Updated `CHANGELOG.rst`.

### Usage

```bash
python examples/diffusers/sparsity/wan22_skip_softmax.py \
    --model-path Wan-AI/Wan2.2-T2V-A14B-Diffusers \
    --calibrate --target-sparsity 0.5 --calib-size 4 \
    --initial-disabled-steps 5 \
    --export-dir ./wan22_skip_softmax_ckpt
```

Resulting layout — a `config.json` per component, **no `sparse.yaml`**:

```
wan22_skip_softmax_ckpt/
├── transformer/config.json        # carries sparse_attention_config
├── transformer_2/config.json      # carries sparse_attention_config
├── vae/ …  text_encoder/ …  tokenizer/ …  scheduler/ …
└── model_index.json
```

A representative `config.json` entry for a diffusion transformer:

```json
"sparse_attention_config": {
  "config_groups": {
    "group_0": {
      "algorithm": "skip_softmax",
      "targets": ["WanAttention"],
      "ignore": ["blocks.0.attn1", "blocks.0.attn2", "…"],
      "initial_disabled_steps": 5,
      "threshold_scale_factor": {
        "formula": "a * exp(b * target_sparsity)",
        "prefill": {"a": 1443.49, "b": 4.30}
      },
      "target_sparsity": {"prefill": 0.5}
    }
  },
  "producer": {"name": "modelopt", "version": "0.45.0..."}
}
```

The N:M variant adds a second group:

```json
"group_1": {
  "algorithm": "sparse_softmax",
  "targets": ["WanAttention"],
  "sparsity_n": 2, "sparsity_m": 4,
  "dense_sink_tokens": 0, "dense_recent_tokens": 64
}
```

### Testing

- `tests/examples/diffusers_sparsity/test_sparsity.py`: baseline / triton-baseline / fixed-threshold runs of the Wan 2.2 example, plus a Python-API calibrate → **export** test asserting the nested `sparse_attention_config` (`threshold_scale_factor`, `target_sparsity`, `ignore`, `initial_disabled_steps`) and the absence of any `sparse.yaml`.
- `tests/unit/torch/sparsity/attention_sparsity/test_sparse_attention_conversion.py` and `test_sparse_attn_config.py`: unit coverage of the per-group export schema and the deploy-reader round-trip (writer nests → reader reads from groups → internal mtsa config unchanged).
- Validated end-to-end on Wan 2.2 T2V-A14B: full 4-prompt / 40-step / 81-frame calibration; the exported checkpoint carries the nested schema in both `transformer` and `transformer_2` `config.json`, and runtime measurement shows ~47–49% tile sparsity at a 0.5 target.

- Is this change backward compatible?: ❌ The exported `sparse_attention_config` schema was renamed and nested per-group during 0.45.x development, and the loader reads only the new layout — checkpoints exported by earlier 0.45.x builds must be re-exported. No released version is affected. 


tests/examples/diffusers/test_sparsity.py
```diff
@@ -0,0 +1,278 @@
...
+        assert "raw_threshold" in group_0
+        assert group_0["raw_threshold"] == -5.0
+        assert "disabled_layers" in group_0  # cross-attn layers are disabled
+        assert "producer" in sa_config
+
+        # Check model weights were saved
+        weight_files = list(component_dir.glob("*.safetensors")) + list(component_dir.glob("*.bin"))
+        assert len(weight_files) > 0, f"No weight files for {component}"
+
+    # Check unified sparse.yaml at top level (combines all components)
```
shengliangxu: do we have the downstream systems agree to read and parse this sparse.yaml file?

jingyu-ml: Yes, TRTLLM will read this.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1019,8 +1023,33 @@ def _export_diffusers_checkpoint(
...
 
+    # Step 8.5: Write unified sparse.yaml at the top-level export directory.
+    # Combines sparse configs from all components (keyed by pipeline component name)
+    # so downstream consumers get the full pipeline's sparse config in one file.
+    if pipeline_sparse_configs:
+        yaml_path = export_dir / "sparse.yaml"
+        with open(yaml_path, "w") as file:
+            yaml.dump(pipeline_sparse_configs, file, default_flow_style=False, sort_keys=False)
+        print(f"Saved unified sparse config to {yaml_path}")
+
```
zhenhuaw-me: What's the purpose of this YAML file?

modelopt/torch/sparsity/attention_sparsity/conversion.py
```diff
@@ -349,73 +349,144 @@ def update_sparse_attention_metadata(
...
+                entry["log_a"] = params["log_a"]
+            threshold_scale_factor[phase] = entry
+
+        group_0["threshold_scale_factor"] = threshold_scale_factor
+
+    if raw_threshold is not None:
+        group_0["raw_threshold"] = raw_threshold
+
+    if disabled_layer_names:
+        group_0["disabled_layers"] = disabled_layer_names
```
zhenhuaw-me: Does it make sense to align the key naming with quant configs like this? https://huggingface.co/nvidia/Wan2.2-T2V-A14B-Diffusers-NVFP4/blob/main/transformer/config.json#L45

modelopt/torch/export/unified_export_hf.py
jingyu-ml: this is the ask from the previous trtllm team, that these sparse config needs to be inside a yaml file

modelopt/torch/sparsity/attention_sparsity/conversion.py
jingyu-ml: it matched already: 

```python
                "ignore": [
                    "blocks.0.attn1",
                    "blocks.0.attn2",
                    "blocks.1.attn1",
                    "blocks.1.attn2",
                    "blocks.2.attn2",
...
```

modelopt/torch/export/unified_export_hf.py
jingyu-ml: In the latest commit, I removed the YAML file. please check again

kaix-nv (approved):

examples/diffusers/README.md
```diff
@@ -314,7 +314,7 @@ By following these steps, your PEFT LoRA model should be efficiently quantized u
 
 ## Sparse Attention (Skip-Softmax)
 
-Skip-softmax sparse attention skips KV tiles whose attention scores are negligible during the softmax computation, reducing FLOPs without retraining. An exponential model (`scale_factor = a * exp(b * target_sparsity)`) is calibrated once, then the target sparsity can be adjusted at runtime without recalibration.
+Skip-softmax sparse attention skips KV tiles whose attention scores are negligible during the softmax computation, reducing FLOPs without retraining. An exponential model (`scale_factor = a * exp(b * target_sparsity)`) is calibrated once, then the target sparsity can be adjusted at runtime without recalibration. Calibrated coefficients can be exported as a Hugging Face checkpoint (per-component `config.json` plus a top-level `sparse.yaml`) consumed directly by TRT-LLM's `SkipSoftmaxAttentionConfig.resolve_for_target_sparsity` — no extra conversion needed downstream.
```
cjluo-nv: > _Bot comment._

This sentence says the export writes "per-component `config.json` plus a top-level `sparse.yaml`", but the PR body and the new tests (`test_wan22_export_sparse_checkpoint`, `test_wan22_calibrated_export`) explicitly assert that no `sparse.yaml` is produced — the sparse config lives only inside each component's `config.json`. Please drop the `sparse.yaml` mention so the docs match the implementation.

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_config.py
```diff
@@ -170,11 +170,19 @@ def load_from_checkpoint_metadata(hf_config) -> tuple[dict, str] | None:
     config_groups = sparse_meta.get("config_groups", {})
     if not isinstance(config_groups, dict):
         return None
-    algos = {grp.get("sparse_algo") for grp in config_groups.values() if isinstance(grp, dict)}
-    if "softmax_skip" in algos and _has_calibrated_threshold_scale_factor(
-        sparse_meta.get("threshold_scale_factor")
+    algos = {grp.get("algorithm") for grp in config_groups.values() if isinstance(grp, dict)}
```
cjluo-nv, line 177: > _Bot comment._

The exporter renames `sparse_algo` → `algorithm`, `softmax_skip` → `skip_softmax`, and relocates `threshold_scale_factor` / `target_sparse_ratio` / `sparse_softmax` from the top level into per-group dicts (with `target_sparse_ratio` further renamed to `target_sparsity`). This reader only reads the new layout, so any checkpoint produced by the previous exporter will silently fall through to `return None` here (i.e. sparse attention disappears at deploy time). The PR body claims backward compatibility, but this is a hard break for on-disk checkpoints. Either accept the legacy keys here as a fallback, or call out the format break explicitly in the CHANGELOG so users know they need to re-export.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1128,6 +1128,19 @@ def _export_diffusers_checkpoint(
         else:
             _save_component_state_dict_safetensors(component, component_export_dir)
 
+        # Step 8: Update config.json with sparse attention info (both quantized and non-quantized)
```
cjluo-nv: > _Bot comment._

Both this block and the quantization block above it are commented `# Step 8:` — please relabel one to keep the step numbering monotonic.

modelopt/torch/sparsity/attention_sparsity/conversion.py
```diff
@@ -430,36 +445,28 @@ def export_sparse_attention_config(model: nn.Module) -> dict[str, Any] | None:
             )
         if sparse_softmax_config is None:
             sparse_softmax_config = _get_sparse_softmax_export_config(module)
+        if not initial_disabled_steps:
```
cjluo-nv, line 450: > _Bot comment._

`initial_disabled_steps` is taken from the first enabled module that reports a non-zero value, so any per-module variation is silently collapsed to a single value in the exported config. If that's intentional (single global setting), please add a comment to that effect; otherwise consider asserting all enabled modules agree.

cjluo-nv (commented):

Functionality and tests look reasonable, but two issues to address before merge:

1. **Doc inconsistency in `examples/diffusers/README.md`** — the new prose claims the export produces "per-component `config.json` plus a top-level `sparse.yaml`", but the PR body, the example flow, and the unit tests all assert there is *no* `sparse.yaml` (it explicitly checks `not (export_dir / "sparse.yaml").exists()`). This README line will mislead users. Drop the `sparse.yaml` mention.

2. **Backward-incompatible on-disk schema change without a compat path.** The exported `sparse_attention_config` JSON has been renamed in three ways at once:
   - top-level key `sparse_algo` → `algorithm`
   - algorithm value `softmax_skip` → `skip_softmax`
   - `threshold_scale_factor` / `target_sparse_ratio` / `sparse_softmax` moved from top level into per-group dicts, with `target_sparse_ratio` renamed to `target_sparsity`
   
   The deploy reader (`plugins/sparse_attn_config.py`) only understands the new layout; it silently returns `None` (i.e. no sparse attention) for any checkpoint produced by the prior exporter. The PR body asserts "Backward compatible: ✅" but this format break is not backward compatible with already-exported checkpoints. Either keep a fallback in `load_from_checkpoint_metadata` for the old keys, or call out the break explicitly in the CHANGELOG so users know to re-export. (The team has previously been bitten by silent reinterpretation of exported `(a,b)` checkpoints — worth being loud here.)

Minor:
- `_export_diffusers_checkpoint` now has two blocks both labeled `# Step 8:` — relabel for clarity.
- `initial_disabled_steps` is harvested from "the first enabled module that has a non-zero value", so per-module variation is silently collapsed to one value. Probably fine for current usage but worth documenting.

examples/diffusers/README.md
Edwardf0t1: **Doc contradiction:** "plus a top-level `sparse.yaml`" contradicts the rest of this PR. The PR description, the sibling `examples/diffusers/sparsity/README.md` ("lives only in config.json — never a standalone yaml"), and two tests that assert `not (export_dir / "sparse.yaml").exists()` all say there is no `sparse.yaml`. Please drop the "plus a top-level `sparse.yaml`" clause.

modelopt/torch/sparsity/attention_sparsity/conversion.py
```diff
@@ -430,36 +445,28 @@ def export_sparse_attention_config(model: nn.Module) -> dict[str, Any] | None:
...
-        "producer": {
-            "name": "modelopt",
-            "version": mo_version,
-        },
-    }
-
-    if calibration_params is not None:
-        # Build threshold_scale_factor with model parameters
+        if disabled_layer_names:
+            skip_group["ignore"] = disabled_layer_names
```
Edwardf0t1, line 467: `ignore` is emitted here, but the in-repo deploy reader (`plugins/sparse_attn_config.py::_build_calibrated_softmax_skip_config`) never reads it back — it builds the same `{"*attn*": {enable: True}, "default": {enable: False}}` config regardless. So when that plugin is the consumer, deployment sparsifies *all* `*attn*` modules, including the cross-attention (`attn2`) layers calibration deliberately kept dense — exactly the failure mode the comment above says this is preventing.

If `ignore` is meant to be consumed by TRT-LLM rather than this vLLM plugin, that's fine, but then the PR's "round-trip consistent" claim overstates what the modelopt-side reader does, and there's no test covering `ignore` being honored on load. Could you either wire `ignore` into the reader or clarify who consumes it?

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_config.py
```diff
@@ -95,15 +95,15 @@ def _add_sparse_softmax_params(layer_cfg: dict, sparse_meta: dict, config_groups
...
+    """Build a vLLM Triton sparse config from a skip_softmax config group."""
     return {
         "sparse_cfg": {
             "*attn*": {
                 "method": "triton_skip_softmax",
-                "threshold_scale_factor": sparse_meta["threshold_scale_factor"],
+                "threshold_scale_factor": skip_group["threshold_scale_factor"],
                 "target_sparse_ratio": _normalize_target_sparse_ratio(
-                    sparse_meta.get("target_sparse_ratio")
+                    skip_group.get("target_sparsity")
```
Edwardf0t1: This reader consumes `threshold_scale_factor` and `target_sparsity` from `skip_group` but drops the `ignore` entry entirely, so the dense-layer subset recorded at export time is lost on reload (see related comment in `conversion.py`). Consider honoring `skip_group.get("ignore")` so the loaded `sparse_cfg` keeps the same layers dense that were calibrated dense.

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_config.py
```diff
@@ -170,11 +170,19 @@ def load_from_checkpoint_metadata(hf_config) -> tuple[dict, str] | None:
...
         return None
-    algos = {grp.get("sparse_algo") for grp in config_groups.values() if isinstance(grp, dict)}
-    if "softmax_skip" in algos and _has_calibrated_threshold_scale_factor(
-        sparse_meta.get("threshold_scale_factor")
+    algos = {grp.get("algorithm") for grp in config_groups.values() if isinstance(grp, dict)}
+    skip_group = next(
+        (
+            grp
+            for grp in config_groups.values()
+            if isinstance(grp, dict) and grp.get("algorithm") == "skip_softmax"
```
Edwardf0t1, line 182: **Backward-compat:** the on-disk schema keys were renamed (`sparse_algo`→`algorithm`, `softmax_skip`→`skip_softmax`, `target_sparse_ratio`→`target_sparsity`, and `threshold_scale_factor`/`target_sparsity` moved into the group). This reader now keys solely off `algorithm`/`"skip_softmax"` with no fallback, so any checkpoint exported by a prior modelopt version silently fails to load its sparse config (returns `None`). The PR checklist marks "backward compatible ✅", which isn't accurate for the on-disk format. Since the feature is new (0.45.x) this is likely fine in practice — please either note the format break or keep a legacy-key fallback.

examples/diffusers/sparsity/wan22_skip_softmax.py
```diff
@@ -190,15 +191,33 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument(
         "--calib-frames",
         type=int,
-        default=151,
-        help="Number of frames for calibration",
+        default=None,
+        help="Number of frames for calibration (default: same as --num-frames).",
```
Edwardf0t1, line 195: Minor: changing the `--calib-frames` default from `151` to `None` (= `--num-frames`) is a behavior change for the example. It's reasonable and aligns with the new "Resolution-dependent fit" guidance — just flagging it as intentional.

Edwardf0t1 (commented):
Reviewed the schema refactor + diffusion export. The move to nest each algorithm's params inside its own `config_groups` group is a clear improvement, and the writer/reader/test changes are internally consistent. Imports and helpers all check out. A few things worth addressing before merge — left them inline. The biggest is the `ignore` list being written but not consumed on load, and a doc line that contradicts the rest of the PR.

tests/examples/diffusers/sparsity/test_sparsity.py
kevalmorabia97, line 1: can you move `tests/examples/diffusers_sparsity` to `tests/examples/diffusers/sparsity` to match `examples/diffusers/sparsity` structure

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_config.py
jingyu-ml: Agreed — this is a hard break for on-disk checkpoints, so it's not backward compatible. We're keeping the new naming and per-group schema; I've updated the PR's *backward-compatible* checkbox to ❌ with a note that checkpoints exported by earlier 0.45.x builds need to be re-exported. Since the skip-softmax export is unreleased, no released checkpoints are affected.

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_config.py
jingyu-ml: Agreed — this is a hard break for on-disk checkpoints, so it's not backward compatible. We're keeping the new naming and per-group schema; I've updated the PR's *backward-compatible* checkbox to ❌ with a note that checkpoints exported by earlier 0.45.x builds need to be re-exported. Since the skip-softmax export is unreleased, no released checkpoints are affected.

examples/diffusers/README.md
jingyu-ml: Fixed in 6ec6d4a08 — dropped the `sparse.yaml` mention; the line now reads that the calibrated coefficients are embedded in each component's `config.json` under `sparse_attention_config`. Thanks!

jingyu-ml: Fixed in 6ec6d4a08 — dropped the "plus a top-level `sparse.yaml`" clause; it now matches the rest of the PR (config lives only in each component's `config.json`).

modelopt/torch/export/unified_export_hf.py
jingyu-ml: Fixed in 6ec6d4a08 — relabeled the sparse-attention `config.json` block to `# Step 9`.

modelopt/torch/sparsity/attention_sparsity/conversion.py
jingyu-ml: Intentional — a single run-wide value. Added a comment in 6ec6d4a08 clarifying it's harvested from the first enabled module that sets it, the same pattern as `calibration_params` / `target_sparse_ratio`.

modelopt/torch/sparsity/attention_sparsity/conversion.py
jingyu-ml: Good catch — wired `ignore` into the in-repo reader in 6ec6d4a08. `_build_calibrated_softmax_skip_config` now emits `{enable: False}` for each ignored layer ahead of the catch-all `*attn*` rule (first match wins in `match_sparse_config`), so the dense (e.g. cross-attention) layers stay dense on load. Added `test_calibrated_softmax_skip_honors_ignore` to cover it.

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_config.py
jingyu-ml: Done in 6ec6d4a08 — the reader now honors `skip_group['ignore']` (emits `{enable: False}` entries before `*attn*`), with a unit test. The round-trip is now consistent on the modelopt side too.

examples/diffusers/sparsity/wan22_skip_softmax.py
jingyu-ml: Confirmed intentional — defaulting `--calib-frames` to `--num-frames` aligns with the resolution-dependent-fit guidance (calibrate at the deployment frame count).

tests/examples/diffusers/sparsity/test_sparsity.py
jingyu-ml: Done in 6ec6d4a08 — moved to `tests/examples/diffusers/sparsity/test_sparsity.py` to match the `examples/diffusers/sparsity` layout.

kevalmorabia97:
/ok to test [4323134](https://github.com/NVIDIA/Model-Optimizer/pull/1269/commits/4323134257080d17dce32040a424386e51a9c386)

kevalmorabia97 (approved):

kevalmorabia97:
/ok to test [80610e9](https://github.com/NVIDIA/Model-Optimizer/pull/1269/commits/80610e974f25113a94e7ab3c48478dfdc857ab0a)

Edwardf0t1 (approved):
LGTM.

PR #1275, opened 2026-04-16
Removed version fixes for torch transformers in windows ptq example requirements

hthadicherla:
### What does this PR do?

Type of change: Bug fix

Removed version fixes for torch and transformers

### Testing
Tested quantization with a couple of models . Working as expected.

## Summary by CodeRabbit

* **Chores**
  * Relaxed dependency specs: removed strict pin for torch to allow latest compatible installs, and constrained transformers to <5.0.0 for broader compatibility and easier updates.

### DIFF
--- a/examples/windows/onnx_ptq/genai_llm/requirements.txt
+++ b/examples/windows/onnx_ptq/genai_llm/requirements.txt
@@ -1,3 +1,3 @@
 datasets>=2.14.5
-torch==2.9.0
-transformers==4.57.3
+torch
+transformers
--- a/examples/windows/onnx_ptq/genai_llm/requirements.txt
+++ b/examples/windows/onnx_ptq/genai_llm/requirements.txt
@@ -1,3 +1,3 @@
 datasets>=2.14.5
 torch
-transformers
+transformers<5.0.0


examples/windows/onnx_ptq/genai_llm/requirements.txt
```diff
@@ -1,3 +1,3 @@
 datasets>=2.14.5
-torch==2.9.0
-transformers==4.57.3
+torch
+transformers
```
kevalmorabia97: does this work for transformers v5 also?

shengliangxu: agreed, do we want to specify a range instead?

hthadicherla: Atleast for the quantization example, I saw that it worked with transformers v5 for the models I tested(llama, qwen, gemma). I could however lock it install highest  version below 5 if you think it could cause an issue.

kevalmorabia97: Do you support MoE models? Thats the main one which is affected by transformers v5. If you only care about dense llms, then you are fine using latest transformers

hthadicherla: I don't know whether it is used for MOE models. Atleast in our internal automation , we only test for dense models. 

@vishalpandya1990 do you know if our quantization example supports MOE models ?

hthadicherla: Adding upper bound to < 5 anyway. 
If there is a need in future for transformers v5 for this particular example, I can add it then.

shengliangxu (approved):

PR #1279, opened 2026-04-16
[Serialization]: remove explicit weights_only default from safe_load to allow user to bypass if needed 

kevalmorabia97:
## Summary

- Remove the `kwargs.setdefault("weights_only", True)` call from `safe_load`, deferring to torch's built-in default (which is `True` for torch>=2.6)
- This allows users to override via the `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1` env var when they trust a checkpoint but hit `pickle.UnpicklingError`
- Add a test that verifies the default fails on unsafe objects and the env var bypass works

## Test plan

- [x] `python -m pytest tests/unit/torch/utils/test_serialization.py -v`

## Summary by CodeRabbit

* **Bug Fixes**
  * Serialization utility now respects PyTorch's default behavior and environment-variable configuration instead of forcibly enforcing parameter overrides, providing greater configuration flexibility.

* **Tests**
  * Added test coverage validating environment-variable override functionality and default behavior in the serialization utility.

### DIFF
--- a/modelopt/torch/utils/serialization.py
+++ b/modelopt/torch/utils/serialization.py
@@ -54,9 +54,11 @@ def safe_save(obj: Any, f: str | os.PathLike | BinaryIO, **kwargs) -> None:
 
 
 def safe_load(f: str | os.PathLike | BinaryIO | bytes, **kwargs) -> Any:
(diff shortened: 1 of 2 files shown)


shengliangxu (approved):
LGTM

PR #1281, opened 2026-04-16
[minor] Add custom calibration backend registry

Fridah-nv:
### What does this PR do?

Type of change: ? 

### Usage

```python
```

## Summary by CodeRabbit

* **New Features**
  * Added a public backend-specific calibrator registration API to support FP8 scale-sweep calibration, allowing backends to supply custom calibrators used during FP8 tuning.

* **Tests**
  * Added unit tests confirming registry insertion/overwrite, that registered calibrators are invoked when FP8 scale-sweep is enabled, are not invoked when disabled, and that calibration falls back to defaults when no backend is registered.


realAsma (approved):

realAsma:
LGTM!

modelopt/torch/quantization/model_calib.py
```diff
@@ -55,11 +55,30 @@
...
     "sequential_calibrate",
     "smoothquant",
     "svdquant",
 ]
 
+# Registry for backends that provide a custom calibrator factory for mse_calibrate().
+_FP8_SWEEP_CALIBRATOR_REGISTRY: dict[str, type] = {}
+
+
+def register_fp8_sweep_calibrator(backend: str, calibrator_factory) -> None:
```
shengliangxu: who will call this API?

and what is calibrator_factory? can we make it typed?

Fridah-nv: This will be used by internal repo. 

> can we make it typed?

Sure, let me add type hint

modelopt/torch/quantization/model_calib.py
```diff
@@ -52,14 +53,38 @@
...
 ]
 
+CalibratorFactory: TypeAlias = Callable[
+    [torch.Tensor, int | tuple | list | None, Callable[..., torch.Tensor]], _Calibrator
+]
+
+_FP8_SWEEP_CALIBRATOR_REGISTRY: dict[str, CalibratorFactory] = {}
+
+
+def register_fp8_sweep_calibrator(backend: str, calibrator_factory: CalibratorFactory) -> None:
```
shengliangxu: do we intend to make it an API that our external users would use?

Fridah-nv: Advanced users can use it for new calibration algorithms, though I not sure if we have such use case in external usages.

shengliangxu: OK, then I would recommend that we keep it as a private API, prefix it with _

`
def _register_fp8_sweep_calibrator
`

If we we want to expose it as public API, then we also need documentation about it. And would need to at least implement some calibrator_facotry for user to choose.

How do you think?

Fridah-nv: Thanks for the suggestion, make the update

modelopt/torch/quantization/model_calib.py
```diff
@@ -54,10 +54,10 @@
 
 __all__ = [
     "CalibratorFactory",
+    "_register_fp8_sweep_calibrator",
```
shengliangxu: For a private API, we do not want to put it into `__all__`.

`__all__` will automatically import if user write:

`from model_calib import *`

without putting into `__all__` user can still explicitly import using:

`import model_calib._register_fp8_sweep_calibrator`
or
`from model_calib import _register_fp8_sweep_calibrator`

only public APIs should be put into `__all__`

shengliangxu (approved):

PR #1289, opened 2026-04-17
[OMNIML-3349][6114983] Add FP8 MHA quantization support for HuggingFace ViT

ajrasane:
## Summary
Enables TensorRT attention-v2 fusion for vision transformers when exported to ONNX with FP8 Q/DQ. The core library changes are architecture-agnostic (drop-in for any FP8 ONNX export); coverage is exercised by the existing `examples/torch_onnx/torch_quant_to_onnx.py` pipeline.

- **`modelopt/onnx/export/fp8_exporter.py`** — new post-processing passes: move attention-scaling `Mul` and K `Transpose` to the Q-side so DQ feeds MatMul directly, pre-transpose constant weights, and insert FP8 Q/DQ on Softmax outputs (fixed `1/448` scale, data-independent) for MHA-v2 fusion. Rewrites only fire when every downstream consumer is a MatMul so non-attention branches are never perturbed.
- **`modelopt/onnx/utils.py`** — `fold_dq_fp32_to_fp16_casts` / `fold_q_fp16_to_fp32_casts` remove the Cast nodes `convert_float_to_float16` inserts around Q/DQ and rewrite scale initializers to FP16 so TRT fuses DQ into the downstream GEMM. Guarded behind opset >= 19 (FP16 Q/DQ scale requirement). Warns on FP16 overflow/underflow.
- **`modelopt/torch/_deploy/utils/torch_onnx.py`** — calls the fold helpers for FP8-quantized models after `convert_float_to_float16`.
- **`modelopt/torch/quantization/export_onnx.py`** — keeps FP8 Q/DQ scale in the native input dtype so no Cast is emitted between graph and Q/DQ. Removes the now-unused `trt_high_precision_dtype` parameter from `_fp8_quantize`/`_fp8_dequantize`.
- **`modelopt/torch/quantization/nn/modules/quant_layernorm.py`** (new) — registers `nn.LayerNorm` in `QuantModuleRegistry` so LayerNorm output quantizers are honored.
- **`modelopt/torch/quantization/plugins/huggingface.py`** — skips `*Attention` wrappers whose children are also `*Attention` per-instance (not per-class) to avoid double-patching `eager_attention_forward` (e.g. `ViTAttention` vs `ViTSelfAttention`).
- **`examples/torch_onnx/torch_quant_to_onnx.py`** — adds a `_FP8_MHA_OVERRIDE` config block to FP8 mode that enables LayerNorm output quantizer + disables its input quantizer for TRT attention fusion.
- **Unit tests** (12 CPU tests, ~1.2s total) — fp8_exporter rewrites + fanout safety, fold-cast helpers + opset guard, LayerNorm quant-wrapper identity, per-instance nested-attention detection.

## Benchmarks
ViT-base-patch16-224, RTX 6000 Ada, strongly-typed FP8 via `trtexec`. Accuracy on 2 000 ImageNet-1k validation samples (streaming).

**Batch = 1 (latency-bound)**
| Model | Top-1 | Top-5 | TRT latency | Speedup |
|---|---|---|---|---|
| FP16 baseline | 80.96% | 95.80% | 0.722 ms | 1.00x |
| Torch FP8 MHA | 80.66% | 95.75% | 0.657 ms | **1.10x** |
| ONNX PTQ FP8 | — | — | 0.589 ms | **1.23x** |

**Batch = 64 (throughput-bound, realistic inference)**
| Model | TRT latency | Speedup | Images/s |
|---|---|---|---|
| FP16 baseline | 23.40 ms | 1.00x | 1152 |
| Torch FP8 MHA | 15.89 ms | **1.47x** | 1152 |
| ONNX PTQ FP8 | 15.89 ms | **1.47x** | 1216 |

Top-1 accuracy stays within 0.30 pp of FP16; at batch=64 the Torch FP8 MHA path matches ONNX PTQ wall-time — attention is the bottleneck there and both paths achieve full FP8 attention fusion (36/36 attention MatMuls with QDQ in ViT-base).

## Test plan
- [x] CPU unit tests (new): \`python -m pytest tests/unit/onnx/quantization/test_fp8_mha_exporter.py tests/unit/onnx/test_fold_casts.py tests/unit/torch/quantization/test_quant_layernorm.py tests/unit/torch/quantization/plugins/test_nested_attention_skip.py\`
- [x] Existing ONNX / quantization unit suites unaffected: \`python -m pytest tests/unit/onnx tests/unit/torch/quantization\`
- [x] End-to-end ViT FP8 export: \`python examples/torch_onnx/torch_quant_to_onnx.py --timm_model_name vit_base_patch16_224 --quantize_mode fp8 --onnx_save_path vit_base_fp8.onnx\` — expect log lines \`Folded 48 weight Transpose nodes\`, \`Inserted FP8 weight DequantizeLinear for 1 Conv nodes\`, and \`Attention QDQ rewrites: ... inserted QDQ on 12 Softmax outputs\`
- [x] trtexec FP8 strongly-typed build: \`trtexec --onnx=vit_base_fp8.onnx --fp8 --stronglyTyped\`
- [x] Accuracy within ~0.3 pp of FP16 baseline on ImageNet-1k subset

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,10 @@ Changelog
   - ``from modelopt.torch.quantization.src.conv.implicit_gemm_cuda import ...`` → ``from modelopt.torch.kernels.quantization.conv.implicit_gemm_cuda import ...``
   - ``from modelopt.torch.sparsity.attention_sparsity.kernels import ...`` → ``from modelopt.torch.kernels.sparsity.attention import ...``
 
+**New Features**
+
+- Add FP8 MHA quantization support for vision transformers. Adds an attention-aware ONNX post-processing pass (scale Mul / K-transpose move before Q, Q→DQ insertion on softmax output) in :class:`FP8QuantExporter <modelopt.onnx.export.fp8_exporter.FP8QuantExporter>`, per-instance nested-attention-wrapper skipping in the HF plugin, and ``nn.LayerNorm`` registration in ``QuantModuleRegistry`` so BMM input quantizers and LayerNorm output quantizers defined in FP8_DEFAULT_CFG are honored end-to-end. See `examples/torch_onnx/torch_quant_to_onnx.py <https://github.com/NVIDIA/Model-Optimizer/tree/main/examples/torch_onnx/torch_quant_to_onnx.py>`_ for the general timm-model quantize→ONNX workflow.
+
 0.44 (2026-05-xx)
 ^^^^^^^^^^^^^^^^^
 
--- a/modelopt/onnx/export/fp8_exporter.py
+++ b/modelopt/onnx/export/fp8_exporter.py
@@ -17,6 +17,7 @@
 
 import time
 
+import numpy as np
 import onnx
 import onnx_graphsurgeon as gs
 import torch
@@ -25,6 +26,11 @@ from onnx_graphsurgeon.ir.tensor import LazyValues
 from modelopt.onnx.logging_config import logger
 
 from .base_exporter import ONNXQuantExporter
+
+# FP8 E4M3 max representable magnitude; softmax output in [0, 1] saturates exactly at 1.0
+# when using 1/448 as the Q scale (single fixed value — softmax range is data-independent).
+_FP8_E4M3_MAX = 448.0
+_FP8_E4M3_SOFTMAX_SCALE = 1.0 / _FP8_E4M3_MAX
 
 
 class FP8QuantExporter(ONNXQuantExporter):
@@ -62,6 +68,8 @@ class FP8QuantExporter(ONNXQuantExporter):
         # Fold constants is required since the scale is not constant yet.
         graph.cleanup().toposort().fold_constants().cleanup()
 
+        n_t_folded = 0
+
         for node in graph.nodes:
             if node.op == "TRT_FP8QuantizeLinear":
                 # Should not remove input QDQ (only process weight quantization)
@@ -74,9 +82,44 @@ class FP8QuantExporter(ONNXQuantExporter):
                 torch_scale = torch.from_numpy(scale.values)
                 quantizer_name = scale.name.rsplit("/", 1)[0]
                 dq_op = node.outputs[0].outputs[0]
-                assert dq_op.op == "TRT_FP8DequantizeLinear", (
-                    f"QDQ does not occur in pairs. You reached {dq_op.op}"
-                )
+                if dq_op.op != "TRT_FP8DequantizeLinear":
+                    raise RuntimeError(f"QDQ does not occur in pairs. You reached {dq_op.op}")
+
+                # Pre-transpose constant weights if DQ feeds ``Transpose → MatMul`` (or
+                # ``Cast → Transpose → MatMul`` after fp16 conversion) so TRT sees DQ→MatMul.
+                # Control flow: scan candidates; a Cast-wrapped candidate is accepted only if it
+                # leads to a Transpose; a bare Transpose whose all consumers are MatMul wins and
+                # breaks the loop. Any other shape defaults `cast_to_remove` back to None and
+                # continues scanning.
+                transpose_to_remove = None
+                cast_to_remove = None
+                for candidate in list(dq_op.outputs[0].outputs):
+                    if candidate.op == "Cast":
+                        cast_to_remove = candidate
+                        candidate = next(
+                            (c for c in candidate.outputs[0].outputs if c.op == "Transpose"),
+                            None,
+                        )
+                        if candidate is None:
+                            cast_to_remove = None
+                            continue
+                    if candidate.op != "Transpose":
+                        cast_to_remove = None
+                        continue
+                    t_consumers = list(candidate.outputs[0].outputs)
+                    # Only fold the transpose when every downstream consumer is MatMul; otherwise
+                    # non-MatMul consumers would observe the un-transposed weights.
+                    if t_consumers and all(c.op == "MatMul" for c in t_consumers):
+                        perm = candidate.attrs.get("perm", None)
+                        torch_weights = (
+                            torch_weights.permute(*perm).contiguous()
+                            if perm is not None
+                            else torch_weights.T.contiguous()
+                        )
+                        transpose_to_remove = candidate
+                    else:
+                        cast_to_remove = None
+                    break
 
                 # Replace it with Dequantize with FP8 weights. This is a WAR because numpy does not support fp8.
                 numpy_weights = (
@@ -94,9 +137,23 @@ class FP8QuantExporter(ONNXQuantExporter):
                 dq_op.inputs[0] = onnx_weights_fp8
                 dq_op.op = "DequantizeLinear"
                 dq_op.outputs[0].dtype = dq_op.inputs[1].dtype
+                dq_op.outputs[0].shape = list(numpy_weights.shape)
+
+                if transpose_to_remove is not None:
+                    t_out = transpose_to_remove.outputs[0]
+                    for consumer in list(t_out.outputs):
+                        for i, inp in enumerate(consumer.inputs):
+                            if inp is t_out:
+                                consumer.inputs[i] = dq_op.outputs[0]
+                    transpose_to_remove.outputs.clear()
+                    if cast_to_remove is not None:
+                        cast_to_remove.outputs.clear()
+                    n_t_folded += 1
 
         graph.cleanup().toposort()
         end_time = time.time()
+        if n_t_folded > 0:
+            logger.info(f"Folded {n_t_folded} weight Transpose nodes during weight compression")
         print(f"fp8 qdq replaced with only dq completed in {end_time - start_time}s.")
 
         return gs.export_onnx(graph)
@@ -121,7 +178,6 @@ class FP8QuantExporter(ONNXQuantExporter):
         Returns:
             Number of Conv weight DQ nodes inserted.
         """
-        fp8_max = 448.0
         count = 0
 
         for node in list(graph.nodes):
@@ -142,7 +198,7 @@ class FP8QuantExporter(ONNXQuantExporter):
             amax = torch_weights.abs().max().float()
             if amax == 0:
                 continue
-            scale_val = (amax / fp8_max).item()
+            scale_val = (amax / _FP8_E4M3_MAX).item()
 
             # Quantize weights to FP8 (WAR: numpy doesn't support fp8)
             fp8_data = (torch_weights / scale_val).to(torch.float8_e4m3fn).view(torch.uint8).numpy()
@@ -155,8 +211,6 @@ class FP8QuantExporter(ONNXQuantExporter):
             )
 
             # Scale in FP16 — DQ output type matches scale dtype, must match activation type
-            import numpy as np
-
             scale_constant = gs.Constant(
                 node.name + "/weight_quantizer/scale",
                 np.array(scale_val, dtype=np.float16),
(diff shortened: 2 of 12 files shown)


modelopt/onnx/export/fp8_exporter.py
```diff
@@ -78,6 +80,33 @@ def compress_weights(onnx_model: onnx.ModelProto) -> onnx.ModelProto:
                     f"QDQ does not occur in pairs. You reached {dq_op.op}"
```
cjluo-nv: Bug (existing but expanded): This `assert` will be stripped with `-O` flag. Per codebase conventions, use `raise ValueError(...)` or `RuntimeError(...)` instead for runtime validation:
```python
if dq_op.op != "TRT_FP8DequantizeLinear":
    raise RuntimeError(f"QDQ does not occur in pairs. You reached {dq_op.op}")
```

modelopt/onnx/export/fp8_exporter.py
```diff
@@ -78,6 +80,33 @@ def compress_weights(onnx_model: onnx.ModelProto) -> onnx.ModelProto:
...
+                            (c for c in candidate.outputs[0].outputs if c.op == "Transpose"),
+                            None,
+                        )
+                        if candidate is None:
+                            cast_to_remove = None
+                            continue
+                    if candidate.op != "Transpose":
+                        cast_to_remove = None
+                        continue
+                    if any(c.op == "MatMul" for c in candidate.outputs[0].outputs):
```
cjluo-nv: Potential issue: When `dq_op.outputs[0]` has multiple consumers and one is `Cast` followed by `Transpose`, the loop breaks after the first candidate. If the first consumer is a Cast that doesn't lead to a Transpose, `cast_to_remove` is set to `None` and then `continue` re-enters the loop at the next candidate — but `break` at line 109 exits regardless. This means if the first candidate is a non-Cast/non-Transpose node, we skip checking remaining consumers. Consider restructuring to check all candidates:

```python
for candidate in list(dq_op.outputs[0].outputs):
    if candidate.op == "Cast":
        ...
        if candidate is None:
            cast_to_remove = None
            continue  # <-- this continues to next candidate, good
    if candidate.op != "Transpose":
        cast_to_remove = None
        continue  # <-- this also continues, good  
    ...
    break  # <-- but this break exits even if we didn't find a match
```

Actually looking more carefully, the logic resets `cast_to_remove = None` when it's not a Transpose, then `continue`s — so it does check the next candidate. But if `candidate.op` is a `Transpose` without a MatMul consumer, you `break` without resetting state. The logic works but is fragile. A comment would help.

modelopt/onnx/export/fp8_exporter.py
```diff
@@ -94,20 +123,232 @@ def compress_weights(onnx_model: onnx.ModelProto) -> onnx.ModelProto:
...
+            t_output = transpose_node.outputs[0]
+            if not any(c.op == "MatMul" for c in t_output.outputs):
+                continue
+
+            new_t_output = gs.Variable(q_input.name + "_transposed", dtype=q_input.dtype)
+            graph.nodes.append(
+                gs.Node(
+                    op="Transpose",
+                    name=transpose_node.name + "_moved",
+                    inputs=[q_input],
```
cjluo-nv, line 355: The hardcoded `1.0/448.0` is correct (FP8 E4M3 max = 448, softmax output ∈ [0,1]), and the docstring explains this well. However, consider defining this as a named constant (e.g., `_FP8_E4M3_SOFTMAX_SCALE = 1.0 / 448.0`) since the same value is also used in `export_onnx.py` (`export_fp8_mha` uses `1.0` for softmax amax which becomes `448.0/1.0` then `1.0/448.0`). This would make the connection explicit.

modelopt/onnx/utils.py
```diff
@@ -1415,6 +1415,96 @@ def _bypass_cast_node(model: onnx.ModelProto, node: onnx.NodeProto) -> None:
...
+
+def _scale_fp32_to_fp16(scale_init: onnx.TensorProto) -> None:
+    """Convert a scalar Q/DQ scale initializer in-place from FP32 to FP16."""
+    import numpy as np
+
+    if scale_init.data_type != onnx.TensorProto.FLOAT:
+        return
+    scale_data = np.frombuffer(scale_init.raw_data, dtype=np.float32)
+    if not scale_data.size:
+        scale_data = np.array(scale_init.float_data, dtype=np.float32)
```
cjluo-nv, line 1431: Potential precision issue: `_scale_fp32_to_fp16` does not check if the FP32 scale value is outside FP16 representable range. If a scale is very small (< ~5.96e-8 for FP16 subnormals) or very large (> 65504), the conversion will silently produce 0 or inf. Consider adding a warning:
```python
import warnings
fp16_val = scale_data.astype(np.float16)
if np.any(np.isinf(fp16_val)) or np.any(fp16_val == 0) and np.any(scale_data != 0):
    logger.warning("Q/DQ scale overflows/underflows when converting to FP16")
```

modelopt/onnx/utils.py
```diff
@@ -1415,6 +1415,96 @@ def _bypass_cast_node(model: onnx.ModelProto, node: onnx.NodeProto) -> None:
...
 
 
+_DQ_OPS = {"DequantizeLinear", "TRT_FP8DequantizeLinear"}
+_Q_OPS = {"QuantizeLinear", "TRT_FP8QuantizeLinear"}
+
+
+def _scale_fp32_to_fp16(scale_init: onnx.TensorProto) -> None:
+    """Convert a scalar Q/DQ scale initializer in-place from FP32 to FP16."""
+    import numpy as np
+
```
cjluo-nv, line 1424: Minor: `np` is already imported at the top of the file. The local `import numpy as np` is unnecessary.

modelopt/torch/quantization/export_onnx.py
```diff
@@ -219,25 +219,16 @@ def _fp8_quantize(
     trt_high_precision_dtype: str,
 ):
     """Helper Function for Quantization."""
+    # Emit the scale in the native input dtype so no Cast is inserted between the
```
cjluo-nv, line 221: The removed `assert` that validated `trt_high_precision_dtype in (input_type, "Float")` was a useful safety check. Now that the scale is emitted in the native input dtype, the `trt_high_precision_dtype` parameter is unused in `_fp8_quantize`. Should it be removed from the signature (and callers updated), or should there be a new validation? Currently `trt_high_precision_dtype` is passed but never used, which is confusing.

modelopt/torch/quantization/export_onnx.py
```diff
@@ -249,23 +240,14 @@ def _fp8_dequantize(
 ):
     """Helper Function for Dequantization."""
     output_shape = sym_help._get_tensor_sizes(inputs)
-    assert trt_high_precision_dtype in (otype, "Float"), (
-        "TRT StronglyType requires both weights and amax to be in the BF16/FP16, or the QDQ in Float."
-    )
     scale = g.op(
         "Constant",
         value_t=torch.tensor(scale_inv, dtype=torch_dtype_map[otype]),  # type: ignore[index]
```
cjluo-nv, line 243: Same as above: `trt_high_precision_dtype` is now unused in `_fp8_dequantize` — the DQ output type is set to `otype` directly. The parameter should either be removed or have a deprecation comment explaining it's kept for API compatibility.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -286,10 +286,25 @@ def register_hf_attentions_on_the_fly(model):
 
     attention_cls = set()
     registered_attn_module = False
+
+    # Skip attention wrappers that contain a nested "Attention" child (e.g. ViTAttention
+    # wraps ViTSelfAttention). Patching both would double-quantize eager_attention_forward.
+    parent_attention_types = {
+        _QuantAttention.get_attn_type(module)
+        for _, module in model.named_modules()
+        if type(module).__name__.endswith("Attention")
```
cjluo-nv: Good approach for detecting parent attention wrappers! One edge case: `named_modules()` includes the module itself, so `child is not module` is the correct check (which you have). However, this only checks one level of nesting. If there's a 3-level hierarchy like `OuterAttention > MiddleAttention > InnerAttention`, both `OuterAttention` and `MiddleAttention` would be in `parent_attention_types`, and only `InnerAttention` would be registered. Is that the intended behavior? For ViT it's fine (only 2 levels), but worth a comment.

modelopt/torch/quantization/nn/modules/quant_layernorm.py
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Quantized layer normalization module."""
+
+import torch.nn as nn
+
+from .quant_module import QuantInputBase, QuantModuleRegistry
+
+QuantModuleRegistry.register({nn.LayerNorm: "nn.LayerNorm"})(QuantInputBase)
```
cjluo-nv, line 28: This follows the exact pattern of `quant_batchnorm.py` — good. However, unlike BatchNorm which has 1d/2d/3d variants, LayerNorm only has one class, so this is clean. Consider adding a brief module-level docstring noting that this enables LayerNorm output quantizers to be honored during quantization (the file docstring says "Quantized layer normalization module" which is a bit terse).

examples/torch_onnx/vit_mha_quantization.py
```diff
@@ -0,0 +1,538 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv: **Missing unit tests**: This 538-line example is the only "test" for the entire PR. The core library changes (graph rewrites in `fp8_exporter.py`, cast folding in `utils.py`, attention skipping in `huggingface.py`, LayerNorm quantization) need proper unit tests that:

1. Test `_move_mul_before_qdq` with a synthetic ONNX graph
2. Test `_move_transpose_before_qdq` with a synthetic ONNX graph
3. Test `_insert_qdq_after_softmax` with a synthetic ONNX graph
4. Test `fold_dq_fp32_to_fp16_casts` and `fold_q_fp16_to_fp32_casts`
5. Test `parent_attention_types` detection with a mock ViT-like model
6. Test that `nn.LayerNorm` is properly registered in `QuantModuleRegistry`

These can all be CPU-only tests with small synthetic models/graphs.

examples/torch_onnx/vit_mha_quantization.py
```diff
@@ -0,0 +1,538 @@
...
+        onnx_path=fp16_onnx_path,
+        quantize_mode="fp8",
+        output_path=output_path,
+    )
+    print(f"  ONNX PTQ FP8 model saved to: {output_path}")
+    return output_path
+
+
+# ──────────────────────────────────────────────────────────────────────────────
+# Step 4: TRT latency benchmarking
```
cjluo-nv: `contextlib.suppress(Exception)` is used in multiple places in the benchmark parsing. Per codebase conventions, prefer explicit handling — suppressing all exceptions can mask real bugs. At minimum, suppress only `(IndexError, ValueError)` which is what you're already doing inline:
```python
with contextlib.suppress(IndexError, ValueError):
```

cjluo-nv (commented):
This PR adds FP8 MHA quantization support for HuggingFace ViT models with ONNX export optimizations. The implementation is well-structured and addresses a real gap (NVBug 6078291). However, there are several issues to address:

**Critical issues:**
1. **No unit tests** — This is ~933 lines of new/changed library code across core export paths (`fp8_exporter.py`, `utils.py`, `export_onnx.py`, `huggingface.py`) with zero unit tests. The only "test" is the example script which requires GPU, ImageNet data, and TRT. The graph rewrite functions in `fp8_exporter.py`, the cast folding helpers in `utils.py`, the attention skipping logic in `huggingface.py`, and the LayerNorm quantization registration all need unit tests.

2. **Bare `assert` for runtime validation** in `fp8_exporter.py` — the existing `assert` on QDQ pair validation will be stripped with `-O`.

3. **Silent `contextlib.suppress(Exception)` in the example** — can mask real failures during benchmark parsing.

**Minor issues:**
4. The `_scale_fp32_to_fp16` helper doesn't handle the case where the scale value overflows or underflows to inf/0 in FP16 — this could silently produce bad quantization results for extreme scales.

5. The `_move_mul_before_qdq` rewrite assumes a single scalar const Mul for attention scaling; if the model architecture changes, these pattern-matching rewrites could silently become no-ops without any warning.

6. The `_insert_qdq_after_softmax` hardcodes scale=1/448 which is correct for E4M3 but should at minimum document why this specific value and that it's tied to the FP8 E4M3 max representable value.

**Positive aspects:**
- Clean separation of graph rewrites as static methods
- Good docstrings on the new functions
- The parent_attention_types detection for avoiding double-patching is well done
- The LayerNorm registration follows existing patterns exactly

examples/torch_onnx/vit_mha_quantization.py
kevalmorabia97, line 1: can you update changelog for 0.45 to mention this new feature details?

CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Changelog
...
+- Add FP8 MHA quantization support for vision transformers. Adds an attention-aware ONNX post-processing pass (scale Mul / K-transpose move before Q, Q→DQ insertion on softmax output) in :class:`FP8QuantExporter <modelopt.onnx.export.fp8_exporter.FP8QuantExporter>`, per-instance nested-attention-wrapper skipping in the HF plugin, and ``nn.LayerNorm`` registration in ``QuantModuleRegistry`` so BMM input quantizers and LayerNorm output quantizers defined in FP8_DEFAULT_CFG are honored end-to-end. See `examples/torch_onnx/torch_quant_to_onnx.py <https://github.com/NVIDIA/Model-Optimizer/tree/main/examples/torch_onnx/torch_quant_to_onnx.py>`_ for the general timm-model quantize→ONNX workflow.
```
kevalmorabia97: This should go to 0.45 section

ajrasane: Done

examples/torch_onnx/vit_mha_quantization.py
ajrasane: Done

cjluo-nv (approved):
All critical issues from previous reviews have been addressed: unit tests added (4 test files), bare `assert` replaced with `RuntimeError`, all graph rewrites guarded with `all(MatMul)` fanout checks, opset guards added to FP16 scale folding, FP16 overflow warning added, per-instance nested attention detection implemented, and unused parameter documented. The remaining minor items (unreachable dead code in `_insert_qdq_after_softmax`, copyright year 2024 on new files) are non-blocking.

PR #1291, opened 2026-04-17
add gptq fused kernel

sychen52:
### What does this PR do?

Add gptq fused kernel to improve speed.

### Usage

check unittest

### Testing
added a unittest

## Summary by CodeRabbit

* **New Features**
  * Fused GPTQ backend for faster blockwise weight updates, toggleable via a new "fused" option.
  * Shared NVFP4 quantization primitives exposed for reuse.

* **Refactor**
  * Consolidated FP4 scale/quantization logic into reusable utilities and centralized Hessian inversion handling.

* **Tests**
  * Expanded GPU tests comparing fused vs unfused GPTQ, added Triton-availability gating and a local benchmark entrypoint.

### DIFF
--- a/modelopt/torch/quantization/triton/gptq_fused_kernel.py
+++ b/modelopt/torch/quantization/triton/gptq_fused_kernel.py
@@ -53,7 +53,6 @@ def _gptq_scalar_kernel(
     n_scale_blocks,
     quant_block_size,
     block_start,
-    n_cols,
     BLOCK_SIZE: tl.constexpr,
 ):
     row = tl.program_id(0)
@@ -66,7 +65,7 @@ def _gptq_scalar_kernel(
     scales_base = scales_ptr + row * n_scale_blocks
 
     j_range = tl.arange(0, BLOCK_SIZE)
-    w_full = tl.load(w_base + j_range, mask=j_range < n_cols, other=0.0)
+    w_full = tl.load(w_base + j_range)
 
     for col in range(0, BLOCK_SIZE, 1):
         scale = tl.load(scales_base + (block_start + col) // quant_block_size)
@@ -85,7 +84,7 @@ def _gptq_scalar_kernel(
         tl.store(err_base + col, err_val)
         tl.store(qw_base + col, q_scalar)
 
-        remaining = (j_range > col) & (j_range < n_cols)
+        remaining = j_range > col
         hinv_row = tl.load(hinv_ptr + col * BLOCK_SIZE + j_range, mask=remaining, other=0.0)
         w_full = w_full - err_val * hinv_row
 
@@ -96,7 +95,6 @@ def gptq_fused_block_scalar(
     h_inv_cho_blk: torch.Tensor,
     quant_block_size: int,
     block_start: int,
-    n_cols: int,
 ) -> tuple[torch.Tensor, torch.Tensor]:
     """Run scalar GPTQ (NVFP4) column loop for one block in a single Triton kernel launch.
 
@@ -106,7 +104,6 @@ def gptq_fused_block_scalar(
         h_inv_cho_blk:   Block of upper-Cholesky inverse Hessian ``[block_size, block_size]``.
         quant_block_size: Number of elements sharing one scale factor.
         block_start:     Column offset of this block in the full weight matrix.
-        n_cols:          Number of active columns in this block.
 
     Returns:
         ``(qw_block, err_block)`` each ``[num_rows, block_size]``.
@@ -126,7 +123,6 @@ def gptq_fused_block_scalar(
         scales_2d.shape[1],
         quant_block_size,
         block_start,
-        n_cols,
         BLOCK_SIZE=block_size,
     )
 
--- a/modelopt/torch/quantization/triton/gptq_fused_kernel.py
+++ b/modelopt/torch/quantization/triton/gptq_fused_kernel.py
@@ -15,30 +15,28 @@
 
 """Fused Triton kernels for GPTQ blockwise weight-update.
 
-A kernel for scalar (NVFP4) quantization.
-Each kernel fuses quantization + per-column GPTQ error propagation into
+A kernel for scalar (NVFP4) quantization with inline two-level scale computation.
+Fuses scale computation + quantization + per-column GPTQ error propagation into
 one launch per GPTQ block, avoiding the Python-level per-column loop.
 
-Architecture (both kernels):
+Architecture:
   - One Triton program per output row.
   - ``w_full [BLOCK_SIZE]`` register tensor holds working weights.
-  - Per-column error propagation: ``w_full -= err * h_inv_row``.
-
-Scalar kernel (``_gptq_scalar_kernel``):
-  - Calls ``nvfp4_scalar_quant()`` from ``nvfp4_quant.py`` per column.
+  - Per-column: calls ``nvfp4_scalar_qdq()`` for FP4 QDQ with inline scale
+    computation, then propagates error via ``w_full -= err * h_inv_row``.
 """
 
 import torch
 import triton
 import triton.language as tl
 
-from .nvfp4_quant import nvfp4_scalar_quant
+from .nvfp4_quant import nvfp4_scalar_qdq
 
 __all__ = ["gptq_fused_block_scalar"]
 
 
 # ---------------------------------------------------------------------------
-# Scalar kernel — NVFP4 quantization + error propagation
+# Scalar kernel — NVFP4 QDQ + error propagation
 # ---------------------------------------------------------------------------
 
 
@@ -47,10 +45,11 @@ def _gptq_scalar_kernel(
     w_ptr,
     qw_ptr,
     err_ptr,
-    scales_ptr,
+    amax_ptr,
+    global_scale,
     hinv_ptr,
     num_rows,
-    n_scale_blocks,
+    n_amax_blocks,
     quant_block_size,
     block_start,
     BLOCK_SIZE: tl.constexpr,
@@ -62,19 +61,20 @@ def _gptq_scalar_kernel(
     w_base = w_ptr + row * BLOCK_SIZE
     qw_base = qw_ptr + row * BLOCK_SIZE
     err_base = err_ptr + row * BLOCK_SIZE
-    scales_base = scales_ptr + row * n_scale_blocks
+    amax_base = amax_ptr + row * n_amax_blocks
 
     j_range = tl.arange(0, BLOCK_SIZE)
     w_full = tl.load(w_base + j_range)
 
     for col in range(0, BLOCK_SIZE, 1):
-        scale = tl.load(scales_base + (block_start + col) // quant_block_size)
+        block_amax = tl.load(amax_base + (block_start + col) // quant_block_size)
 
         w_scalar = tl.sum(tl.where(j_range == col, w_full, 0.0))
         q_scalar = tl.sum(
-            nvfp4_scalar_quant(
+            nvfp4_scalar_qdq(
                 tl.full([1], w_scalar, dtype=tl.float32),
-                scale,
+                block_amax,
+                global_scale,
                 1,
             )
         )
@@ -91,16 +91,22 @@ def _gptq_scalar_kernel(
 
 def gptq_fused_block_scalar(
     w_block: torch.Tensor,
-    scales_2d: torch.Tensor,
+    block_amax: torch.Tensor,
+    global_scale: float,
     h_inv_cho_blk: torch.Tensor,
     quant_block_size: int,
     block_start: int,
 ) -> tuple[torch.Tensor, torch.Tensor]:
     """Run scalar GPTQ (NVFP4) column loop for one block in a single Triton kernel launch.
 
+    Computes FP8-quantized scales from per-block amax inline via
+    :func:`nvfp4_scalar_qdq`, then performs NVFP4 fake quantization and
+    GPTQ error propagation per column.
+
     Args:
         w_block:         Working weights ``[num_rows, block_size]`` (float32).
-        scales_2d:       Pre-computed scales ``[num_rows, n_scale_blocks]`` (float32).
+        block_amax:      Per-block amax values ``[num_rows, n_amax_blocks]`` (float32).
+        global_scale:    Pre-computed ``global_amax / (6.0 * 448.0)`` (scalar).
         h_inv_cho_blk:   Block of upper-Cholesky inverse Hessian ``[block_size, block_size]``.
         quant_block_size: Number of elements sharing one scale factor.
         block_start:     Column offset of this block in the full weight matrix.
@@ -117,10 +123,11 @@ def gptq_fused_block_scalar(
         w_block.contiguous(),
         qw_block,
         err_block,
-        scales_2d.contiguous(),
+        block_amax.contiguous(),
+        global_scale,
         h_inv_cho_blk.contiguous(),
         num_rows,
-        scales_2d.shape[1],
+        block_amax.shape[1],
         quant_block_size,
         block_start,
         BLOCK_SIZE=block_size,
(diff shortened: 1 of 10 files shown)


modelopt/torch/quantization/triton/nvfp4_quant.py
```diff
@@ -0,0 +1,95 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv, line 1: Copyright year is `2026` — should be `2025`.

modelopt/torch/quantization/triton/gptq_fused_kernel.py
```diff
@@ -0,0 +1,133 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv, line 1: Copyright year is `2026` — should be `2025`.

modelopt/torch/quantization/triton/nvfp4_quant.py
```diff
@@ -0,0 +1,95 @@
...
+Single source of truth for FP4 decision-boundary rounding.  Used by:
+  - ``fp4_kernel.py``         (standalone blockwise fake quant)
+  - ``fp4_kernel_hopper.py``  (Hopper block-pointer variant)
+  - ``gptq_fused_kernel.py``  (fused GPTQ scalar path)
+
+FP4 (E2M1) representable magnitudes: {0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0}
+"""
+
+import triton
+import triton.language as tl
```
cjluo-nv, line 27: Missing `__all__` definition. Since `__init__.py` does `from .nvfp4_quant import *`, the internal Triton JIT helpers `fp4_round_magnitude` and `nvfp4_scalar_quant` will leak into the public API of `modelopt.torch.quantization.triton`. Consider adding:
```python
__all__: list[str] = []  # No public symbols; used internally by sibling kernels
```
or explicitly listing only what should be public.

modelopt/torch/quantization/triton/gptq_fused_kernel.py
```diff
@@ -0,0 +1,133 @@
...
+    w_full = tl.load(w_base + j_range, mask=j_range < n_cols, other=0.0)
+
+    for col in range(0, BLOCK_SIZE, 1):
+        scale = tl.load(scales_base + (block_start + col) // quant_block_size)
+
+        w_scalar = tl.sum(tl.where(j_range == col, w_full, 0.0))
+        q_scalar = tl.sum(
+            nvfp4_scalar_quant(
+                tl.full([1], w_scalar, dtype=tl.float32),
+                scale,
```
cjluo-nv: Nit: The `w_block.contiguous()` call here is redundant — the caller already passes `w[:, bs:be].clone().contiguous()`. Not a bug, just unnecessary work (the `contiguous()` is a no-op on an already-contiguous tensor, but it reads confusingly as if the caller might not guarantee contiguity).

tests/gpu/torch/quantization/test_gptq.py
```diff
@@ -231,3 +234,221 @@ def test_gptq_e2e_flow(quant_cfg):
...
+
+_NVFP4_BENCH_IDS = [f"qbs{qbs}_gbs{gbs}_{of}x{d}" for qbs, gbs, of, d in _NVFP4_BENCH_CONFIGS]
+
+
+@requires_triton
+@pytest.mark.parametrize(
+    ("quant_block_size", "gptq_block_size", "out_features", "dim"),
+    _NVFP4_BENCH_CONFIGS,
+    ids=_NVFP4_BENCH_IDS,
+)
```
cjluo-nv: This benchmark test has no assertions — it can never fail. Consider either:
1. Adding a soft check like `assert speedup > 0.5` (sanity, not a strict perf gate), or
2. Marking it `@pytest.mark.benchmark` so it can be easily excluded from CI and run separately.

As-is, it only prints results which are invisible in CI unless someone digs into logs.

cjluo-nv (commented):
Good PR overall: clean deduplication of FP4 rounding logic into `nvfp4_quant.py`, well-structured fused GPTQ kernel, and solid correctness tests. A few issues to address:

1. **Missing `__all__` in `nvfp4_quant.py`** (minor but important): Since `__init__.py` does `from .nvfp4_quant import *`, internal Triton JIT helpers (`fp4_round_magnitude`, `nvfp4_scalar_quant`) leak into the public API. These are internal building blocks, not user-facing.

2. **Copyright year 2026** in both new files — should be 2025.

3. **Benchmark tests in CI**: `test_fused_nvfp4_benchmark` is purely a performance test with no assertions. It prints results but can't fail for correctness. Consider marking it with a custom marker (e.g., `@pytest.mark.benchmark`) or adding a soft speedup assertion to make it more useful, or just being explicit that it's informational-only.

modelopt/torch/quantization/triton/nvfp4_quant.py
sychen52: This year is 2026

modelopt/torch/quantization/triton/gptq_fused_kernel.py
sychen52: same as above.

modelopt/torch/quantization/triton/nvfp4_quant.py
sychen52: removed from .nvfp4_quant import *

tests/gpu/torch/quantization/test_gptq.py
sychen52: add note

modelopt/torch/quantization/triton/gptq_fused_kernel.py
sychen52: Current call has contiguous, but we cannot force this to the caller. No change.

tests/gpu/torch/quantization/test_gptq.py
```diff
@@ -14,15 +14,18 @@
...
 import torch
 from _test_utils.torch.transformers_models import get_tiny_llama, get_tiny_tokenizer
+from conftest import requires_triton
 
 import modelopt.torch.quantization as mtq
 from modelopt.torch.export.unified_export_hf import _export_quantized_weight
 from modelopt.torch.quantization.model_calib import gptq
 from modelopt.torch.quantization.qtensor.nvfp4_tensor import NVFP4QTensor
+from modelopt.torch.quantization.triton.fp4_kernel import compute_fp4_scales
 from modelopt.torch.quantization.utils.calib_utils import update_hessian
```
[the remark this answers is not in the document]
sychen52: Done.

tests/gpu/torch/quantization/test_gptq.py
```diff
@@ -231,3 +234,221 @@ def test_gptq_e2e_flow(quant_cfg):
...
+
+    t_fused = _bench(run_fused)
+    t_unfused = _bench(run_unfused)
+    speedup = t_unfused / t_fused if t_fused > 0 else float("inf")
+
+    tag = f"qbs{quant_block_size}_gbs{gptq_block_size}_{out_features}x{dim}"
+    print(
+        f"\n[{tag}] Fused: {t_fused * 1e3:8.2f} ms | "
+        f"Unfused: {t_unfused * 1e3:8.2f} ms | Speedup: {speedup:.1f}x"
+    )
```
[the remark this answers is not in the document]
sychen52: removed from pytest only there to run by python directly.

modelopt/torch/quantization/triton/gptq_fused_kernel.py
cjluo-nv, line 1: Minor: Copyright year says 2026 — unless you're time-traveling, this should be 2025. Same in `nvfp4_quant.py`.

modelopt/torch/quantization/triton/gptq_fused_kernel.py
```diff
@@ -0,0 +1,133 @@
...
+    scales_base = scales_ptr + row * n_scale_blocks
+
+    j_range = tl.arange(0, BLOCK_SIZE)
+    w_full = tl.load(w_base + j_range, mask=j_range < n_cols, other=0.0)
+
+    for col in range(0, BLOCK_SIZE, 1):
+        scale = tl.load(scales_base + (block_start + col) // quant_block_size)
+
+        w_scalar = tl.sum(tl.where(j_range == col, w_full, 0.0))
+        q_scalar = tl.sum(
```
cjluo-nv, line 73: Nit: When `col >= n_cols`, the kernel still iterates and reads from `scales_base + (block_start + col) // quant_block_size`. In the current usage `BLOCK_SIZE == n_cols` always (caller passes `w[:, bs:be]`), so this is fine. But if someone calls `gptq_fused_block_scalar` with `n_cols < block_size` in the future, this would read past the valid scale range. Consider adding a `if col >= n_cols: break`-equivalent guard (or a comment documenting that `n_cols == BLOCK_SIZE` is assumed).

cjluo-nv (commented):
**Re-review of PR #1291: add gptq fused kernel**

**Addressed (critical/important):**
- ✅ Module-level `compute_fp4_scales` import moved inside function (was breaking pytest collection without Triton)
- ✅ Benchmark test moved out of pytest into `if __name__ == "__main__"` guard
- ✅ Wildcard import of `nvfp4_quant` removed from `__init__.py`
- ✅ `@requires_triton` skip marker properly applied to fused kernel tests

**Still outstanding (minor):**
- ❌ Copyright year 2026 in `nvfp4_quant.py` and `gptq_fused_kernel.py` — author argued "this year is 2026" but it's 2025
- ❌ Hessian initialized on CPU then moved to CUDA (minor nitpick, not blocking)

**Correctness assessment:** The fused kernel logic is correct — the per-column GPTQ error propagation matches the unfused reference, the `nvfp4_scalar_quant` deduplication is clean, and the test (`test_fused_vs_unfused_nvfp4`) provides meaningful correctness assertions with `torch.testing.assert_close`. The PR is well-structured at ~490 lines with good code reuse of FP4 rounding logic.

modelopt/torch/quantization/triton/gptq_fused_kernel.py
sychen52: Let's not argue about this. You need to trust me on this.

modelopt/torch/quantization/triton/gptq_fused_kernel.py
sychen52: Done.

modelopt/torch/quantization/triton/gptq_fused_kernel.py
```diff
@@ -0,0 +1,129 @@
...
+
+    j_range = tl.arange(0, BLOCK_SIZE)
+    w_full = tl.load(w_base + j_range)
+
+    for col in range(0, BLOCK_SIZE, 1):
+        scale = tl.load(scales_base + (block_start + col) // quant_block_size)
+
+        w_scalar = tl.sum(tl.where(j_range == col, w_full, 0.0))
+        q_scalar = tl.sum(
+            nvfp4_scalar_quant(
```
sugunav14: In the future, if we want to add support for more scalar formats do we just call the appropriate quant function here?

sychen52: Yes.

modelopt/torch/quantization/utils/calib_utils.py
```diff
@@ -74,6 +74,42 @@ def update_hessian(input, hessian, n_samples):
     return hessian, n_samples
 
 
+def compute_hessian_inverse(hessian, weight, perc_damp):
```
sugunav14, line 77: Why do we need this as a function outside GPTQHelper?

sychen52: To use it in unit test

modelopt/torch/quantization/utils/calib_utils.py
```diff
@@ -231,6 +222,115 @@ def _print_mse_error(self, hessian):
...
+    from modelopt.torch.quantization.triton.gptq_fused_kernel import gptq_fused_block_scalar
+
+    num_cols = weight.shape[1]
+    for bs in range(0, num_cols, block_size):
+        be = min(bs + block_size, num_cols)
+        qw, err = gptq_fused_block_scalar(
+            weight[:, bs:be].clone().contiguous(),
+            scales_2d,
+            h_inv[bs:be, bs:be].contiguous(),
+            quant_block_size,
```
sugunav14, line 305: quant_block_size corresponds to the number of columns considered for QDQ right? For scalar formats we would still like to do QDQ->GPTQ error update one column at a time i.e. set quant_block_size=1, but it seems like we are extracting it from TensorQuantizer.

modelopt/torch/quantization/utils/calib_utils.py
```diff
@@ -242,3 +342,7 @@ def register_gptq_helper(backend: str, factory: type[GPTQHelper]) -> None:
     construct ``factory`` instead of the default ``GPTQHelper``.
     """
     _GPTQ_HELPER_REGISTRY[backend] = factory
+
+
+# Built-in registrations
+register_gptq_helper("fused_gptq_nvfp4", FusedScalarGPTQHelper)
```
sugunav14: The GPTQ registry dispatches handles based on "backend" attribute in the TensorQuantizer. Would this work?

modelopt/torch/quantization/utils/calib_utils.py
sychen52: This is the block size for quantizing to NVFP4. i.e. how many nvfp4 shares the same block scale factor. GPTQ update is still 1 column at a time.

modelopt/torch/quantization/utils/calib_utils.py
sychen52: added config instead as we discussed.

cjluo-nv:
@sugunav14 could you approve if the PR is at good shape?

cjluo-nv (approved):
## Confirmation: `fp4_kernel.py` and `fp4_kernel_hopper.py` changes are pure no-ops

**`fp4_kernel.py`**: The inline FP4 quantization logic in `static_blockwise_fp4_fake_quant_kernel` is replaced by a call to `nvfp4_scalar_quant()` which implements identical logic. The only micro-difference is the NaN check: `(scale != scale)` → `libdevice.isnan(scale)`, which are semantically equivalent per IEEE 754. The `compute_fp4_scales` extraction from `static_blockwise_fp4_fake_quant` is a pure code-movement refactor. **No behavioral change.**

**`fp4_kernel_hopper.py`**: The `fp4_round_magnitude()` substitution is identical to the removed inline code. The `fp8_quantize_scale()` substitution has one subtle difference: old code multiplied by `global_scale` while the new code multiplies by `global_scale_safe`. These differ only when `global_scale ≤ 0` (yielding 0 vs ~4.5e-10), but the downstream guard `tl.where(block_max_quant >= 1e-5, block_max_quant, 1.0)` catches both cases, making the final result identical. **Effectively a no-op.**

## Previous review comments status
- ✅ Module-level `compute_fp4_scales` import (critical) — moved inside function
- ✅ Benchmark test out of pytest (critical) — moved to `if __name__ == "__main__"` guard
- ✅ Wildcard import of `nvfp4_quant` removed from `__init__.py` (minor)
- ✅ Guard for `col >= n_cols` in kernel (minor) — addressed
- ❌ Copyright year 2026 (minor) — disputed by author; not blocking
- ❌ Hessian CPU init (minor) — not addressed; cosmetic
- ❌ `zero_cols.unsqueeze(-1)` flagged as bug by CodeRabbit — **this is actually NOT a bug**; PyTorch advanced indexing handles the [N,1,1] shape correctly, and this code is pre-existing (just moved to a standalone function)

## Overall assessment
Clean PR: well-structured deduplication of FP4 rounding logic into `nvfp4_quant.py`, correct fused GPTQ kernel, solid correctness test with `torch.testing.assert_close`. The `fused` config plumbing is clean and properly gated behind `_is_nvfp4_static_quantizer` check.

sugunav14 (approved):

PR #1302, opened 2026-04-20
fix: auto-compute dp_replicate_size from world_size

yeyu-nvidia:
## Summary
- When `dp_shard_size < world_size` (e.g., `dp_shard_size=4` on 8 GPUs across 2 nodes), `ParallelismConfig` raises `total_size (4) does not match num_processes (8)` because `dp_replicate_size` defaults to 1
- Auto-compute `dp_replicate_size = world_size // (dp_shard_size * cp_size)` so intra-node FSDP2 sharding + inter-node data-parallel replication works without manual config
- This enables `dp_shard_size` to be set to per-node GPU count (better NVLink utilization) while automatically creating replicas across nodes

## Test plan
- [ ] Verify single-node training (dp_shard_size == world_size, dp_replicate_size == 1) unchanged
- [ ] Verify multi-node with dp_shard_size < world_size creates correct replica groups
- [ ] Verify existing EAGLE3/DFlash configs still work

## Summary by CodeRabbit

* **Refactor**
  * Improved parallelism setup in the speculative decoding example to better detect and validate available devices, derive replication size, and ensure consistent distributed-training configuration.

### DIFF
--- a/examples/speculative_decoding/main.py
+++ b/examples/speculative_decoding/main.py
@@ -212,8 +212,15 @@ def train():
             "Either data.data_path or data.offline_data_path must be set in the config."
         )
     if training_args.cp_size > 1 or training_args.dp_shard_size > 1:
+        # Auto-compute dp_replicate_size so that
+        # dp_replicate_size * dp_shard_size * cp_size == world_size.
+        world_size = int(os.environ.get("WORLD_SIZE", torch.cuda.device_count()))
+        parallel_size = training_args.dp_shard_size * training_args.cp_size
+        dp_replicate_size = world_size // parallel_size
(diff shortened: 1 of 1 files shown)


kevalmorabia97 (approved):

ChenhanYu (approved):
Correct fix for multi-node FSDP2 where dp_shard_size < world_size. One suggestion: add a divisibility guard (`if world_size % parallel_size != 0: raise ValueError(...)`) to catch misconfigurations early instead of letting ParallelismConfig fail with a confusing error. Also note the `torch.cuda.device_count()` fallback returns per-node count, not world size — correct for single-node but worth a comment. LGTM.

PR #1128, opened 2026-03-27
add: DFlash block diffusion speculative decoding

ChenhanYu:
NOTE: closing this PR and opt for https://github.com/NVIDIA/Model-Optimizer/pull/1211

Implement DFlash (Block Diffusion for Flash Speculative Decoding) as a new mode in ModelOpt's speculative decoding framework.

Key architecture:
- Feature Fusion: extract hidden states from uniformly sampled target model layers, project via FC layer
- KV Injection: fused target features injected as K/V entries in every draft decoder layer's attention (not just first layer input)
- Parallel Drafting: all tokens in a block predicted simultaneously using learnable mask embeddings and bidirectional within-block attention

Files:
- dflash/ module: DFlashModel, DFlashConfig, conversion, default config
- plugins/hf_dflash.py: HFDFlashModel with DFlashAttention (KV injection), DFlashModule (feature fusion + decoder), training forward pass with random anchor sampling and exponential position decay loss
- main.py: --mode dflash support in training script

Reference: "DFlash: Block Diffusion for Flash Speculative Decoding" (arXiv:2602.06036)

### What does this PR do?

Type of change: ? 

### Usage

```python
# Add a code snippet demonstrating how to use this
```

### Before your PR is "*Ready for review*"

Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md) and your commits are signed (`git commit -s -S`).

Make sure you read and follow the [Security Best Practices](https://github.com/NVIDIA/Model-Optimizer/blob/main/SECURITY.md#security-coding-practices-for-contributors) (e.g. avoiding hardcoded `trust_remote_code=True`, `torch.load(..., weights_only=False)`, `pickle`, etc.).

- Is this change backward compatible?: ✅ / ❌ / N/A 
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ / ❌ / N/A 
- Did you write any new necessary tests?: ✅ / ❌ / N/A 
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ / ❌ / N/A 

## Summary by CodeRabbit

* **New Features**
  * Added DFlash speculative-decoding mode with end-user config options and sensible defaults (block size, decoder layers, architecture JSON, torch.compile toggle, mask token).
  * CLI/training accepts and runs in DFlash mode; dataset flow supports DFlash.
[...]

PR #1309, opened 2026-04-21
fix: prevent gh-pages repo bloat from doc preview artifacts

kevalmorabia97:
### What does this PR do?

Type of change: Bug fix

Fixes gh-pages branch bloat that grew from ~26 MB to ~441 MB in four weeks (nvbug 6099503). Three compounding causes were identified and addressed:

1. **Sphinx `.doctrees/` cache published to gh-pages** — `sphinx-build` was writing its build cache inside `build/html/` which was then uploaded verbatim. Accounts for ~3.3 GB uncompressed across history.
2. **`JamesIves/github-pages-deploy-action` appending a commit on every push** — main-site files accumulated forever with `single-commit: false` (default).
3. **PR preview deploying on every `synchronize` event for all PRs** — `rossjrw/pr-preview-action` re-deployed the full site for every push to any PR regardless of whether docs changed (e.g. PR #1128 triggered 64 preview deploys × ~11 MB each).

Changes:
- Pass `-d /tmp/doctrees` to `sphinx-build` so `.doctrees/` is never written into `build/html/`
- Add `paths: [docs/**, modelopt/**]` filter to `pull_request` trigger so the docs workflow only runs on PRs that touch docs or source code
- Set `single-commit: true` on the deploy action so main-site pushes squash into one commit
- Deduplicate docs build: `deploy-preview` now downloads the artifact from `build-docs` instead of running a second `sphinx-build`
- Set `retention-days: 1` on the artifact since it is only needed for the duration of the workflow run

The one-time cleanup (force-push squashed orphan to gh-pages) was already applied separately — repo is now ~59 MB for a full clone vs ~441 MB before.

### Usage

N/A — CI/workflow change only.

### Testing

- Workflow logic reviewed manually.
- The one-time cleanup was verified: `git rev-list --objects --disk-usage origin/gh-pages` now reports ~28 MB; full clone is ~59 MB.

### Additional Information

nvbug 6099503

## Summary by CodeRabbit

* **Chores**
  * Optimized documentation build and preview workflow: PR previews now run only for doc-related changes, builds complete faster with shorter timeouts, and previews depend on completed doc builds.
  * Simplified artifact handling with unconditional uploads during runs and shorter retention.
  * Standardized Pages deployment to produce a single consolidated commit.
  * Stabilized docs build by directing temporary build state to a fixed temp path.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -4,10 +4,6 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, closed]
     branches: [main, release/*, feature/*]
-    paths:
-      - "docs/**"
-      - "modelopt/**"
-      - ".github/workflows/pages.yml"
   push:
     branches: [main]
   schedule:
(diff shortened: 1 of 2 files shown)


shengliangxu (approved):
LGTM and let's see if it solve the issue

PR #1310, opened 2026-04-21
fix: layerwise calibration backward-compat, recipe split, batch-size guard

realAsma:
## Summary

Follow-up to #1251 (which renamed `use_sequential` → `layerwise`). Three related fixes bundled:

1. **Backward-compatible config loading.** PTQ checkpoints saved before #1251 store the
   legacy `use_sequential` key in the calibration-algorithm config, so loading them now
   raises `ValidationError: Extra inputs are not permitted (use_sequential)` because
   `QuantizeAlgorithmConfig` uses `extra='forbid'`. Accept `use_sequential` as an alias
   for `layerwise` via `AliasChoices`. The field still serializes as `layerwise`, so
   round-trips through the current schema are clean.

2. **Recipe split.** `nvfp4_experts_only-fp8_kv` previously enabled layerwise calibration
   by default, which changes the calibration flow materially. Split into two recipes:
   - `nvfp4_experts_only-fp8_kv.yaml` — default (no layerwise)
   - `nvfp4_experts_only-fp8_kv_layerwise.yaml` — layerwise variant

3. **`hf_ptq` batch-size guard.** Auto batch-size detection is not supported together
   with layerwise calibration. Default to `batch_size=1` when layerwise is enabled and
   the user hasn't set a batch size explicitly.

Originally reported by Jenny Chen while resuming a PTQ checkpoint via
`restore_sharded_modelopt_state`:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for MaxCalibConfig
use_sequential
  Extra inputs are not permitted [type=extra_forbidden, input_value=False, input_type=bool]
```

## Test plan

- [x] `tests/unit/torch/quantization/test_config_validation.py` — legacy alias accepted, current name accepted, dump serializes under current name, `extra='forbid'` still rejects unknown keys.
- [x] `pre-commit run` — clean.

### Before your PR is *Ready for review*

- Is this change backward compatible?: ✅ (restores compatibility for pre-#1251 checkpoints)
- New PIP dependency: N/A
- New necessary tests: ✅
- Changelog update: N/A (bug fix)

## Release Notes

* **New Features**
  * Added new PTQ recipe for efficient layerwise calibration of large models.
  * Automatic batch size optimization for layerwise calibration recipes.
  * Backward compatibility support for legacy input naming conventions.

* **Documentation**
  * Updated recipe guides and changelog with new layerwise calibration recipe.

* **Tests**
  * Added validation tests for configuration compatibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1310)


modelopt/torch/quantization/config.py
```diff
@@ -1237,6 +1237,21 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
         ),
     )
 
+    @model_validator(mode="before")
+    @classmethod
+    def _rename_legacy_fields(cls, values):
```
shengliangxu: can we bring back the use_sequential? and mark it as deprecated and frozen as:

```
    # 1. Deprecated for documentation
    # 2. Frozen to prevent explicit reassignment after creation
    use_sequential: bool | None = Field(
        default=None, 
        deprecated="Use 'layerwise' instead.",
        frozen=True 
    )
```

this would stop future use of `use_sequential`

shengliangxu: Then we can just use field_validator for the deprecated field, each deprecated field can then have their own deprecating logic.

realAsma: adressed

modelopt/torch/quantization/config.py
```diff
@@ -1237,6 +1243,12 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
         ),
     )
 
+    @model_validator(mode="after")
+    def _migrate_use_sequential(self):
+        if self.use_sequential is not None and not self.layerwise:
+            object.__setattr__(self, "layerwise", self.use_sequential)
+        return self
```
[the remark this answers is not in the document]
shengliangxu: This AI comment is a valid problem and good fix

modelopt/torch/quantization/config.py
```diff
@@ -1227,6 +1227,12 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
         ),
     )
 
+    use_sequential: bool | None = Field(
```
jenchen13: Instead of defining a `use_sequential` field, add an `alias` in layerwise to make it point to the old name 
e.g. `validation_alias='use_sequential'` so that layerwise will accept a value from `use_sequential` but will still serialize to `layerwise`
https://pydantic.dev/docs/validation/latest/concepts/fields#field-aliases

modelopt/torch/quantization/config.py
```diff
@@ -1237,6 +1243,12 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
         ),
     )
 
+    @model_validator(mode="after")
```
jenchen13: this is not necessary with my comment above

jenchen13 (changes requested):

modelopt/torch/quantization/config.py
```diff
@@ -1219,6 +1219,7 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
 
     layerwise: bool = ModeloptField(
         default=False,
+        validation_alias=AliasChoices("layerwise", "use_sequential"),
```
jenchen13, line 591: i'm pretty sure you don't need AliasChoices and can just do `validation_alias="use_sequential"` which will enable validation on both fields

examples/llm_ptq/hf_ptq.py
```diff
@@ -955,6 +955,18 @@ def quantize_main(
...
+    recipe = None
+    if args.recipe is not None and not args.auto_quantize_bits:
+        print(f"Use recipe {args.recipe} for quantization")
+        recipe = load_recipe(args.recipe)
+        assert isinstance(recipe, ModelOptPTQRecipe), (
+            f"Expected PTQ recipe, but got {type(recipe).__name__} from {args.recipe}"
+        )
+
+    recipe_algorithm = recipe.quantize.model_dump().get("algorithm") if recipe else None
+    is_layerwise = isinstance(recipe_algorithm, dict) and recipe_algorithm.get("layerwise", False)
```
jenchen13: you can check is layerwise before converting the recipe to a dict, when it is still a pydantic config via `recipe.algorithm.layerwise`

modelopt_recipes/general/ptq/nvfp4_experts_only-fp8_kv_layerwise.yaml
```diff
@@ -0,0 +1,94 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metadata:
+  recipe_type: ptq
+  description: NVFP4 static weight and dynamic activation for expert layers only (W4A4), FP8 KV cache, max layerwise calibration.
+quantize:
+  algorithm:
+    method: max
+    # Max calibration is fast and does not typically need checkpointing.
```
jenchen13: what does this comment mean? "Max calibration is fast and does not typically need checkpointing."

modelopt/torch/quantization/config.py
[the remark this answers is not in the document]
realAsma: Addressed in ad91b295f6 \u2014 the _migrate_use_sequential validator was removed in favor of validation_alias=AliasChoices("layerwise", "use_sequential") on the layerwise field (config.py:1220-1229), so there is no model_validator path that can drop an explicit layerwise=False. Pydantic resolves the alias during field parsing, and an explicit layerwise: false in YAML wins over a legacy use_sequential: true via normal field precedence.

modelopt/torch/quantization/config.py
realAsma: Addressed in ad91b295f6 \u2014 the separate use_sequential field is gone; layerwise now uses validation_alias=AliasChoices("layerwise", "use_sequential") (config.py:1222), matching your suggestion.

modelopt/torch/quantization/config.py
realAsma: Addressed in ad91b295f6 \u2014 the _migrate_use_sequential validator has been removed; only validate_layerwise_checkpoint_dir (an unrelated after-validator) remains.

modelopt/torch/quantization/config.py
realAsma: We verified this in Pydantic v2: a bare validation_alias="use_sequential" replaces the field name entirely, so only use_sequential would be accepted on input and layerwise: true in YAML would stop validating. Accepting BOTH the field name and the legacy alias requires either AliasChoices("layerwise", "use_sequential") or model_config = ConfigDict(populate_by_name=True). The project-wide ModeloptBaseRule base doesn't set populate_by_name, so AliasChoices is the minimal and most-local fix \u2014 it avoids changing validation behavior for every other field in the hierarchy.

modelopt_recipes/general/ptq/nvfp4_experts_only-fp8_kv_layerwise.yaml
realAsma: The comment explains why layerwise_checkpoint_dir is intentionally omitted from this recipe: method: max calibration is cheap enough per-layer that resuming from a partial run isn't worth the disk overhead, unlike GPTQ/SmoothQuant where checkpointing is valuable. Happy to reword for clarity, e.g. "layerwise_checkpoint_dir intentionally unset: max calibration is fast and does not typically need per-layer checkpointing."

examples/llm_ptq/hf_ptq.py
shengliangxu: Agreed, we do not need to do model_dump, unless have to.

We have places do model_dump just to adapt to the existing codebase.

cjluo-nv (commented):
Overall a clean, well-motivated PR with good tests. Two issues worth addressing:

1. **Copyright year**: The new `nvfp4_experts_only-fp8_kv_layerwise.yaml` uses `Copyright (c) 2024` but the project's `LICENSE_HEADER` says `2026`. Should use the current year.

2. **Bare `assert` for runtime validation**: The recipe type check in `hf_ptq.py` uses `assert isinstance(recipe, ModelOptPTQRecipe)`. Since `assert` is stripped under `-O`, this should be a proper `ValueError`/`TypeError` raise.

modelopt_recipes/general/ptq/nvfp4_experts_only-fp8_kv_layerwise.yaml
```diff
@@ -0,0 +1,94 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv: The project's `LICENSE_HEADER` specifies `Copyright (c) 2026`. This new file has `Copyright (c) 2024`. Please update to match the canonical header.

examples/llm_ptq/hf_ptq.py
```diff
@@ -955,6 +955,18 @@ def quantize_main(
     default_pad_token,
     device: torch.device,
 ):
+    # Load the recipe up front so we can detect layerwise calibration before batch-size probing.
+    recipe = None
+    if args.recipe is not None and not args.auto_quantize_bits:
+        print(f"Use recipe {args.recipe} for quantization")
+        recipe = load_recipe(args.recipe)
```
cjluo-nv, line 995: This `assert` will be stripped under `python -O`. Since this validates user-supplied input at runtime, prefer an explicit `if not isinstance(...): raise TypeError(...)` (or `ValueError`). This is a pre-existing pattern from the moved code, but worth fixing while you're touching it.

examples/llm_ptq/hf_ptq.py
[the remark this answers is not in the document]
realAsma: Addressed in f78ac5046f \u2014 replaced the dict-only inspection with a small recursive helper that handles list-form QuantizeAlgoCfgType (as well as ModelOptPTQRecipe / QuantizeAlgorithmConfig / None). See examples/llm_ptq/hf_ptq.py:967-974.

realAsma: Addressed in f78ac5046f \u2014 the helper now reads recipe.quantize.algorithm directly as a pydantic config (via a ModelOptPTQRecipe branch that recurses into it), no more .model_dump() round-trip. See examples/llm_ptq/hf_ptq.py:967-974.

modelopt_recipes/general/ptq/nvfp4_experts_only-fp8_kv_layerwise.yaml
realAsma: Addressed in f78ac5046f \u2014 bumped the header year to 2026 to match LICENSE_HEADER.

cjluo-nv (approved):

shengliangxu (approved):
LGTM

modelopt_recipes/general/ptq/nvfp4_experts_only-fp8_kv_layerwise.yaml
```diff
@@ -0,0 +1,94 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
meenchen: Okay for now. Can you compose this yaml based on the `modelopt_recipes/general/ptq/nvfp4_experts_only-fp8_kv.yaml` once the composable recipes PR is merged?

jenchen13 (approved):

PR #1313, opened 2026-04-22
Support NVFP4 W4A16 quantization

hychiang-git:
### What does this PR do?

Type of change: ? New feature

This PR supports NVFP4 W4A16 quantization.

### Usage

```python
  python examples/llm_ptq/scripts/huggingface_example.sh \
      --model Qwen/Qwen3-8B \
      --quant nvfp4_w4a16 \
      --calib 1 \
      --kv_cache_quant none \
      --tasks quant
```

### Additional Information

NVFP4 W4A16 is currently supported on vLLM only. TensorRT-LLM and SGLang do not support this format.                                                              
   
  vLLM loads NVFP4 W4A16 checkpoints via its `CompressedTensorsW4A16Fp4` scheme, which requires the checkpoint to be in `compressed-tensors` format. The raw ModelOpt export from this PR is still not directly loadable by vLLM. After running hf_ptq.py, the checkpoint must be converted by:                                                                                                     
  1. Renames tensors: .weight (uint8) → .weight_packed, .weight_scale_2 (fp32) → .weight_global_scale (inverted: 1 / weight_scale_2) 
  2. Rewrites config.json: sets quant_method: "compressed-tensors", format: "nvfp4-pack-quantized", quantization_status: "compressed", and derives the ignore list
  automatically from layers that lack a weight_packed tensor.  

## Summary by CodeRabbit

* **New Features**
  * NVFP4 W4A16 weight-only quantization added (FP4 weights, group_size=16; BF16 activations; no calibration-forward pass). Selectable via qformat/CLI and included in export flow.
  * New --exclude_modules CLI option (and EXCLUDE_MODULES env support in example script) to skip quantizing specific modules.

* **Documentation**
  * Changelog entry describing NVFP4 W4A16 usage and vLLM deployment via compressed-tensors conversion.

* **Tests**
  * Added test coverage for NVFP4 W4A16 export.


cjluo-nv:
I'm now thinking if we should land this PR or just use W4A4 checkpoint in deployment frameworks and use it for W4A16 deployment (ignoring the input quantizers). I think we need extra discussion before landing it.

examples/llm_ptq/hf_ptq.py
```diff
@@ -1242,6 +1261,17 @@ def parse_args() -> argparse.Namespace:
         default=False,
         action="store_true",
     )
+    parser.add_argument(
```
meenchen: I feel this is something we want to avoid, and if users want to exclude some modules, they should specify them in the quant_config or recipe.

modelopt/torch/quantization/config.py
```diff
@@ -794,6 +794,7 @@ def _nvfp4_selective_quant_cfg(
 NVFP4_EXPERTS_ONLY_CFG = _nvfp4_selective_quant_cfg(["*mlp.experts*", "*block_sparse_moe*"])
 NVFP4_MLP_ONLY_CFG = _nvfp4_selective_quant_cfg(["*mlp*", "*block_sparse_moe*"])
 NVFP4_OMLP_ONLY_CFG = _nvfp4_selective_quant_cfg(["*o_proj*", "*mlp*", "*block_sparse_moe*"])
+NVFP4_W4A16_CFG = _nvfp4_selective_quant_cfg(["*"], weight_only=True)
```
jenchen13: could this be a YAML recipe in modelopt_recipes since it's a more experimental format? We want more established formats in config.py

examples/llm_ptq/hf_ptq.py
jenchen13: +1 use the quant_config/yaml to exclude certain modules
e.g. add `{"quantizer_name': "<name>", enabled: False}` at the end of the quant config/yaml (last one overrides previous quantizers)

examples/llm_ptq/scripts/huggingface_example.sh
```diff
@@ -53,9 +53,9 @@ esac
 IFS=","
 for qformat in $QFORMAT; do
     case $qformat in
-    fp8 | fp8_pc_pt | fp8_pb_wo | int8_wo | int8_sq | int4_awq | w4a8_awq | fp16 | bf16 | nvfp4 | nvfp4_awq | nvfp4_mse | w4a8_nvfp4_fp8 | w4a8_mxfp4_fp8 | nvfp4_experts_only | nvfp4_mlp_only | nvfp4_omlp_only | nvfp4_svdquant | mxfp8 | nvfp4_local_hessian) ;;
+    fp8 | fp8_pc_pt | fp8_pb_wo | int8_wo | int8_sq | int4_awq | w4a8_awq | fp16 | bf16 | nvfp4 | nvfp4_awq | nvfp4_mse | w4a8_nvfp4_fp8 | w4a8_mxfp4_fp8 | nvfp4_experts_only | nvfp4_mlp_only | nvfp4_omlp_only | nvfp4_svdquant | mxfp8 | nvfp4_local_hessian | nvfp4_w4a16) ;;
```
jenchen13: @meenchen @cjluo-nv instead of having aliases to the quant configs, could we just use the full quant config name from config.py? this way there's less maintenance w/o having to add aliases

examples/llm_ptq/hf_ptq.py
hychiang-git: Sure! fixed.

examples/llm_ptq/scripts/huggingface_example.sh
```diff
@@ -53,9 +53,9 @@ esac
...
     *)
-        echo "Unknown quant argument: Expected one of: [fp8, fp8_pc_pt, fp8_pb_wo, int8_wo, int8_sq, int4_awq, w4a8_awq, fp16, bf16, nvfp4, nvfp4_awq, nvfp4_mse, w4a8_nvfp4_fp8, w4a8_mxfp4_fp8, nvfp4_experts_only, nvfp4_mlp_only, nvfp4_omlp_only, nvfp4_svdquant, mxfp8, nvfp4_local_hessian]" >&2
+        echo "Unknown quant argument: Expected one of: [fp8, fp8_pc_pt, fp8_pb_wo, int8_wo, int8_sq, int4_awq, w4a8_awq, fp16, bf16, nvfp4, nvfp4_awq, nvfp4_mse, w4a8_nvfp4_fp8, w4a8_mxfp4_fp8, nvfp4_experts_only, nvfp4_mlp_only, nvfp4_omlp_only, nvfp4_svdquant, mxfp8, nvfp4_local_hessian, nvfp4_w4a16]" >&2
```
juhi10071998: Hi team, @hychiang-git - do we have a convention on how the naming should look like for the schemes, should it be `w4a16_nvfp4 `or `nvfp4_w4a16.`

I see for some schemes we have `w4a8_nvfp4_fp8` where we first have the wxay representation followed by the precision for w and a resepectively. 
Curious if we follow any pattern/ have any preference here? I am working on adding corresponding ModelOpt method in vLLM and would like the names to be aligned.

meenchen: It should be `w4a16_nvfp4`. Let's follow the current convention like `w4a8_nvfp4_fp8`

juhi10071998: Thanks @meenchen for confirmation.
@hychiang-git , @ajrasane for reference.
@hychiang-git  could you update this PR to reflect this?

examples/llm_ptq/scripts/huggingface_example.sh
meenchen: config.py is a superset of the supported quant formats. I think that we can gradually move on to the recipe system for new formats.

examples/llm_ptq/scripts/huggingface_example.sh
hychiang-git: done :)

examples/llm_ptq/hf_ptq.py
```diff
@@ -785,6 +786,12 @@ def export_quantized(
                     extra_state_dict=mtp_state_dict,
                 )
 
+                if args.qformat == "w4a16_nvfp4":
+                    warnings.warn(
+                        "TensorRT-LLM and SGLang do not support this format. "
+                        "To serve on vLLM, convert the NVFP4 W4A16 checkpoint to compressed-tensors format."
+                    )
```
meenchen, line 793: Let's mention that the support for vLLM deployment is in progress instead.

CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Changelog
 
 **New Features**
 
+- Add NVFP4 W4A16 weight-only quantization (``w4a16_nvfp4``): FP4 weights with group_size=16, BF16 activations, no calibration forward pass required. Use ``mtq.W4A16_NVFP4_CFG`` or ``--qformat w4a16_nvfp4`` in ``hf_ptq.py``. Exported checkpoints can be served on vLLM after conversion to compressed-tensors format.
```
meenchen: Same here, we can mention the vLLM deployment support is WIP instead.

hychiang-git: Fixed.

examples/llm_ptq/hf_ptq.py
hychiang-git: Fixed.

jenchen13:
If none of the deployment frameworks support this format, please move it to a YAML recipe under `modelopt_recipes` instead of adding it as a aliased quant config. Aliased quant configs inside our main examples should be already widely supported

jenchen13 (changes requested):
we shouldn't be adding quant formats not supported in any deployment framework into the main example -- please convert to yaml recipe for experimental format

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -900,6 +900,26 @@ def forward(self, *args, **kwargs):
         self._down_proj_linear = False
         return super().forward(*args, **kwargs)
 
+    def iter_weights_for_calibration(self):
```
meenchen, line 903: What's this function for?

hychiang-git: W4A4 always provides a forward_loop (calibration data is required to observe activation ranges), so `weight_only_quantize` is never called

https://github.com/NVIDIA/Model-Optimizer/blob/d30ebbd455a16716605f4a52beb8e2b1cec0f941/modelopt/torch/quantization/model_calib.py#L154-L160

The bug only surfaces for W4A16 where weights are the only thing to calibrate and no forward pass is needed.

CHANGELOG.rst
```diff
@@ -28,6 +28,7 @@ Changelog
 
 **New Features**
 
+- Add NVFP4 W4A16 weight-only quantization (``w4a16_nvfp4``): FP4 weights with group_size=16, BF16 activations, no calibration forward pass required. Use ``mtq.W4A16_NVFP4_CFG`` or ``--qformat w4a16_nvfp4`` in ``hf_ptq.py``. vLLM deployment support is in progress.
 - Support full Transformer Engine spec for Minitron pruning (``mcore_minitron``). Now we no longer need to use custom ModelOpt spec. Note that this does not affect the usage of the pruning workflow but makes pruning slightly faster and may result in slightly different pruned model because of different kernel and numerics.
```
juhi10071998, line 34: hi @hychiang-git - quick question, would this work identically for the MoE models also? and there the quant_algo will be *W4A16_NVFP4* right?

modelopt_recipes/general/ptq/nvfp4_weight_only-kv_fp16.yaml
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+imports:
+  base_disable_all: configs/ptq/units/base_disable_all
+  default_disabled_quantizers: configs/ptq/units/default_disabled_quantizers
+  w4a16_nvfp4: configs/ptq/units/w4a16_nvfp4
+
+metadata:
```
cjluo-nv, line 21: actually, should we default to MSE based weight quantization for w4a16? @meenchen WDYT?

cjluo-nv (commented):

The core NVFP4 W4A16 feature addition is small and cleanly threads `QUANTIZATION_W4A16_NVFP4` through existing NVFP4 code paths (reusing `_nvfp4_selective_quant_cfg(..., weight_only=True)` and extending `to_quantized_weight`, `_export_quantized_weight`, `convert_hf_quant_config_format`, `get_weight_scaling_factor[_2]`, and `_get_quantization_from_layer`). Existing recipe infrastructure is reused (no new subsystem), so the design-review protocol check is satisfied. However, several things warrant a human look before merge:

1. **Missing test coverage for the MoE path.** The new `_QuantFusedExperts.iter_weights_for_calibration` override in `modelopt/torch/quantization/plugins/huggingface.py` is only exercised when `forward_loop is None` (weight-only calibration). The new test case in `test_unified_hf_export_and_check_safetensors.py` is only for dense `tiny_llama`; there is no `w4a16_nvfp4` case for `tiny_qwen3_moe` or `tiny_gpt_oss`, so the fused-experts override added here is effectively untested end-to-end.

2. **Two unrelated drive-by changes** are bundled into a "Support NVFP4 W4A16" PR and should either be split out or justified in the PR body:
   - `examples/llm_ptq/hf_ptq.py`: `if args.qformat in QUANT_CFG_CHOICES:` → `if args.recipe is not None or args.qformat in QUANT_CFG_CHOICES:` is a recipe-dispatch fix that has no connection to NVFP4 W4A16 (w4a16_nvfp4 is already in QUANT_CFG_CHOICES).
   - `tools/launcher/core.py`: adding `USER` to the docker env is a launcher bugfix for `getpass.getuser()` crashes.

3. **PR body is out of sync with the diff.** The CodeRabbit summary lists a "new `--exclude_modules` CLI option (and EXCLUDE_MODULES env support in example script)" but no such option appears in the diff. Either remove this from the description or add the option.

4. **Deployment gap.** The PR body acknowledges that the raw ModelOpt export is not directly loadable by vLLM and needs a post-export conversion (rename `.weight`→`.weight_packed`, `.weight_scale_2`→`.weight_global_scale` with inversion, rewrite `config.json`). That conversion tool is not part of the PR, so users following the documented usage will end up with an unusable checkpoint and only a post-hoc `warnings.warn` to hint at it. Consider including the converter (or at least linking to it) and surfacing the warning before the long export runs, not after.

5. **Minor: warning ordering.** In `hf_ptq.py`, the "TensorRT-LLM and SGLang do not support this format" warning fires after `export_hf_checkpoint()` returns — friendlier to warn at the top of `quantize_main` / in `parse_args` so users aren't surprised after a multi-minute export.

No licensing concerns: the new YAMLs carry the standard NVIDIA Apache-2.0 header, matching the existing sibling recipes in `modelopt_recipes/configs/ptq/units/`.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -900,6 +900,26 @@ def forward(self, *args, **kwargs):
...
+        and its corresponding quantizer directly.
+        """
+        for weight_name, quantizers_name in (
+            ("gate_up_proj", "gate_up_proj_weight_quantizers"),
+            ("down_proj", "down_proj_weight_quantizers"),
+        ):
+            weight = getattr(self, weight_name, None)
+            quantizers = getattr(self, quantizers_name, None)
+            if weight is None or quantizers is None:
+                continue
```
cjluo-nv, line 917: > _Bot comment._

The new `iter_weights_for_calibration` override is only triggered by `weight_only_quantize()` in `model_calib.py`, which is the path `max_calibrate` takes when `forward_loop is None` — i.e. exactly the w4a16_nvfp4 flow. But the test matrix in `tests/gpu/torch/export/test_unified_hf_export_and_check_safetensors.py` only adds a dense-Llama case for `w4a16_nvfp4`; no MoE variant exercises this override. Please add a `('w4a16_nvfp4', 'tiny_qwen3_moe-w4a16-nvfp4', ...)` case (or equivalent) so the per-expert calibration path this PR adds is actually tested.

examples/llm_ptq/hf_ptq.py
```diff
@@ -1128,7 +1135,7 @@ def quantize_main(
             quant_cfg = copy.deepcopy(quant_cfg)
             force_weight_quantizers_static(quant_cfg["quant_cfg"])
 
-        if args.qformat in QUANT_CFG_CHOICES:
+        if args.recipe is not None or args.qformat in QUANT_CFG_CHOICES:
```

This change (`if args.recipe is not None or args.qformat in QUANT_CFG_CHOICES:`) is a recipe-dispatch fix that is unrelated to NVFP4 W4A16 — `w4a16_nvfp4` is added to `QUANT_CFG_CHOICES` above, so the original guard already handles it. Please either split this into its own PR or mention in the PR body why it's bundled here, and double-check that downstream code in `mono_quantize` works when `quant_cfg` came from a recipe (e.g., the post-hoc `_set_kv_cache_constant_amax`, MTP exclusion, and `resolve_checkpoint_dir` branches all assume `quant_cfg["quant_cfg"]` is present and mutable).

examples/llm_ptq/hf_ptq.py
```diff
@@ -785,6 +786,12 @@ def export_quantized(
                     extra_state_dict=mtp_state_dict,
                 )
 
+                if args.qformat == "w4a16_nvfp4":
```
cjluo-nv, line 789: > _Bot comment._

Minor: this `warnings.warn` fires only after `export_hf_checkpoint()` returns, i.e. after a potentially long export. Since the user has already committed to `--qformat w4a16_nvfp4` by the time `quantize_main` starts, this warning would be more useful emitted at the top of `quantize_main` (or in `parse_args`) so users learn about the TensorRT-LLM/SGLang/vLLM gap before they wait for the export.

tools/launcher/core.py
```diff
@@ -50,6 +50,9 @@ def get_default_env(experiment_title=None):
         "HF_HOME": f"/{title}/hf-cache",
         "HF_TOKEN": os.getenv("HF_TOKEN", ""),
         "MLM_SKIP_INSTALL": "1",
+        # DockerExecutor runs as the host UID, which may not be in the container's
+        # /etc/passwd; setting USER prevents getpass.getuser() from calling pwd.getpwuid().
+        "USER": os.getenv("USER", "docker"),
```
cjluo-nv: > _Bot comment._

This USER-env fix for `DockerExecutor` / `getpass.getuser()` is unrelated to NVFP4 W4A16 quantization and looks like a drive-by change. Please split it into its own PR (or at minimum note it in the PR body) — bundling unrelated launcher fixes makes the quantization change harder to bisect if something regresses.

modelopt_recipes/configs/ptq/units/w4_nvfp4.yaml
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# W4A16 NVFP4: NVFP4 E2M1 dynamic weight quantizer only; activations remain in BF16.
+
+# modelopt-schema: modelopt.torch.quantization.config.QuantizerCfgListConfig
+imports:
+  nvfp4: configs/numerics/nvfp4
+---
+  - quantizer_name: '*weight_quantizer'
```
shengliangxu, line 22: ~~I am not exactly sure if our quantizer support it, but we should make it explicit that the activation here should be configured as fp16/bf16~~

~~if our current quantizer does not allow it then we should add this capability, but not in this PR.~~

shengliangxu: Actually, rename this unit to just say that it is w4_nvfp4.yaml

I think this is the correct unit name.

modelopt_recipes/general/ptq/nvfp4_weight_only-kv_fp16.yaml
meenchen: I think we should still keep max as the default. I feel the MSE calibration is not mature/need more support for a new model, e.g., we are seeing issues with Qwen3.6 MSE, but max calibration works ootb.

meenchen (approved):
LGTM overall

examples/llm_ptq/hf_ptq.py
```diff
@@ -1147,7 +1154,7 @@ def _is_layerwise(obj):
             quant_cfg = copy.deepcopy(quant_cfg)
             force_weight_quantizers_static(quant_cfg["quant_cfg"])
 
-        if args.qformat in QUANT_CFG_CHOICES:
+        if args.recipe is not None or args.qformat in QUANT_CFG_CHOICES:
```
cjluo-nv: how about we just do `if quant_cfg:` here?

cjluo-nv (approved):

modelopt/torch/export/model_config.py
```diff
@@ -38,6 +38,7 @@
 QUANTIZATION_MXFP4 = "mxfp4"
 QUANTIZATION_MXFP8 = "mxfp8"
 QUANTIZATION_W4A8_MXFP4_FP8 = "w4a8_mxfp4_fp8"
+QUANTIZATION_W4A16_NVFP4 = "w4a16_nvfp4"  # weight-only FP4
```
jenchen13: remove this comment

modelopt/torch/export/quant_utils.py
```diff
@@ -641,6 +644,10 @@ def _get_quantization_from_layer(layer, quantizer_attr_names: QuantizerAttrNames
                 return QUANTIZATION_NVFP4_AWQ
             if getattr(layer, "fused_with_prequant", False):
                 return QUANTIZATION_NVFP4_AWQ
+            # W4A16 weight-only: input_quantizer absent or disabled
```
jenchen13: unnecessary comment

tools/launcher/core.py
jenchen13: is this docker change necessary?

jenchen13 (approved):

modelopt/torch/export/model_config.py
hychiang-git: done

PR #1324, opened 2026-04-22
[NVBug 6102977] Add _disable_use_cache context manager to fix PTQ AttributeError on custom configs

meenchen:
### What does this PR do?

Type of change: Bug fix 

- Summary: Running hf_ptq.py on stepfun-ai/Step-3.5-Flash (and any model whose custom HF config doesn't assign use_cache) crashed in get_max_batch_size() with AttributeError: 'Step3p5Config' object has no attribute 'use_cache' before calibration could start.
-  Extract the existing "disable KV cache during calibration" logic into a _disable_use_cache(model) context manager, apply it to both get_max_batch_size and _forward_loop. The CM sets config.use_cache = False unconditionally (not only when the attribute exists) and restores the prior value on exit if one was set.
-  Behavior unchanged for normal configs; the NemotronH hybrid-cache correctness guarantee from #1251 is preserved.

### Usage

```python
```

### Testing

Step-3.5-Flash PTQ now passes get_max_batch_size

## Summary by CodeRabbit

* **Refactor**
  * Improved memory handling during model evaluation and calibration by consistently disabling KV cache for both single-batch probes and full dataloader runs, simplifying and stabilizing inference flow and ensuring cache state is managed reliably.

* **Tests**
  * Added unit tests verifying cache-state handling across models with and without cache settings, including correct restoration behavior even when errors occur.

### DIFF
--- a/modelopt/torch/utils/dataset_utils.py
+++ b/modelopt/torch/utils/dataset_utils.py
@@ -511,7 +511,7 @@ def get_max_batch_size(
             )
             * sample_memory_usage_ratio
         )
-        if mem_diff_per_data_batch <= 0:
+        if mem_diff_per_data_batch <= 0:  # pragma: no cover - GPU memory probe edge case
             print(
                 "Warning: No measurable memory usage found for a single batch. "
                 "Falling back to batch_size=1."
@@ -533,7 +533,7 @@ def get_max_batch_size(
                 try:
                     infer_method(target_input)
                     break
-                except torch.cuda.OutOfMemoryError:
+                except torch.cuda.OutOfMemoryError:  # pragma: no cover - GPU OOM retry path
                     target_data_batch = target_data_batch // 2
 
     # Regulate the data batch target to be 1, 2, 4, 8, 12, ..., capped at 64
--- a/modelopt/torch/utils/dataset_utils.py
+++ b/modelopt/torch/utils/dataset_utils.py
@@ -512,11 +512,11 @@ def get_max_batch_size(
             * sample_memory_usage_ratio
         )
         if mem_diff_per_data_batch <= 0:  # pragma: no cover - GPU memory probe edge case
-            print(
+            print(  # pragma: no cover
                 "Warning: No measurable memory usage found for a single batch. "
                 "Falling back to batch_size=1."
             )
-            target_data_batch = 1
+            target_data_batch = 1  # pragma: no cover
         else:
             target_data_batch = max(int(free_mem_before / mem_diff_per_data_batch), 1)
         target_input = sample_input_single_batch.expand(
@@ -534,7 +534,7 @@ def get_max_batch_size(
                     infer_method(target_input)
                     break
                 except torch.cuda.OutOfMemoryError:  # pragma: no cover - GPU OOM retry path
-                    target_data_batch = target_data_batch // 2
+                    target_data_batch = target_data_batch // 2  # pragma: no cover
 
     # Regulate the data batch target to be 1, 2, 4, 8, 12, ..., capped at 64
     if target_data_batch < 2:
--- a/modelopt/torch/utils/dataset_utils.py
+++ b/modelopt/torch/utils/dataset_utils.py
@@ -19,7 +19,7 @@ import copy
 import json
 import os
 from collections.abc import Callable, Iterator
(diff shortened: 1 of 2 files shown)


cjluo-nv (commented):
The context manager has a state-leak bug: when `config` didn't originally have `use_cache`, the attribute is set to `False` but never cleaned up on exit. After the context manager completes, the config permanently has `use_cache = False` that wasn't there before. The `finally` block needs to `delattr` in the `not had_attr` case.

Also, no unit tests are included. A simple test for `_disable_use_cache` covering the three cases (no config, config with existing `use_cache`, config without `use_cache`) would be straightforward and valuable.

modelopt/torch/utils/dataset_utils.py
```diff
@@ -437,6 +438,33 @@ def get_supported_datasets() -> list[str]:
...
+    if config is None:
+        yield
+        return
+    had_attr = hasattr(config, "use_cache")
+    prev = config.use_cache if had_attr else None
+    config.use_cache = False
+    try:
+        yield
+    finally:
+        if had_attr:
```
cjluo-nv, line 464: Bug: When `had_attr` is `False` (the config didn't originally have `use_cache`), the context manager sets `config.use_cache = False` on entry but never removes it on exit. This leaks a new attribute onto the config object.

The `finally` block should clean up:
```python
finally:
    if had_attr:
        config.use_cache = prev
    else:
        delattr(config, "use_cache")
```

Without this, after calling `get_max_batch_size` or `_forward_loop`, a config that never had `use_cache` will now permanently have `use_cache = False`, which could change model behavior for subsequent inference.

modelopt/torch/utils/dataset_utils.py
```diff
@@ -437,6 +438,33 @@ def get_supported_datasets() -> list[str]:
     return list(SUPPORTED_DATASET_CONFIG.keys())
```
cjluo-nv, line 438: Missing tests: Please add unit tests for `_disable_use_cache` covering:
1. Model with no `config` attribute (no-op)
2. Model whose config already has `use_cache` (restored on exit)
3. Model whose config lacks `use_cache` (attribute should not persist after exit)

These are simple to write with a mock `nn.Module` and would directly validate the bug fix.

modelopt/torch/utils/dataset_utils.py
```diff
@@ -467,42 +500,43 @@ def _get_free_gpu_mem():
...
-                target_data_batch = target_data_batch // 2
+        # For some models on multi GPU, we observe the memory per batch is not a constant.
+        # So we just test the target batch size and make sure we do not go OOM.
+        while target_data_batch > 1:
+            with torch.set_grad_enabled(enable_grad):
+                try:
+                    infer_method(target_input)
+                    break
+                except torch.cuda.OutOfMemoryError:
+                    target_data_batch = target_data_batch // 2
```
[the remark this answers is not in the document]
meenchen: Good catch, but this is pre-existing — target_input is built outside the OOM retry loop on the parent commit too (verified via git show 204daafc^:modelopt/torch/utils/dataset_utils.py; the loop body is unchanged here, the lines only show in the diff   because we re-indented to wrap them in with _disable_use_cache(model):). Will file a separate fix to move target_input =   ...expand(...) inside the while loop. Out of scope for this PR.

cjluo-nv (approved):

All critical issues from the previous review have been addressed:

1. **State-leak bug (critical) — Fixed**: The `finally` block now properly calls `delattr(config, "use_cache")` (wrapped in `suppress(AttributeError)`) when the attribute didn't exist before entry. This prevents leaking `use_cache = False` onto configs that never had it.

2. **Missing unit tests (critical) — Fixed**: Four well-structured tests added covering all cases: no config attribute (no-op), existing `use_cache` restored on exit (parametrized True/False), missing `use_cache` cleaned up on exit (no leak), and restoration on exception.

3. **OOM retry loop stale `target_input` (pre-existing)**: Author correctly identified this as a pre-existing bug unrelated to this PR — the re-indentation just made it visible in the diff. Appropriate to fix separately.

The context manager is clean, well-documented, and correctly applied to both `get_max_batch_size` and `_forward_loop`. The refactoring from inline try/finally to a shared `@contextmanager` is a good simplification.

PR #1334, opened 2026-04-23
feat(recipes): add KV cache cast variants (fp8_cast / nvfp4_cast)

cjluo-nv:
## Summary

- Adds three built-in PTQ recipes that express the KV-cache *cast* variants directly in YAML, using the existing `use_constant_amax: true` quantizer field. These are recipe equivalents of `--kv_cache_qformat=fp8_cast` / `nvfp4_cast`:
  - `general/ptq/fp8_default-fp8_cast_kv`
  - `general/ptq/nvfp4_default-fp8_cast_kv`
  - `general/ptq/nvfp4_default-nvfp4_cast_kv`
- Makes `--recipe` authoritative in `examples/llm_ptq/hf_ptq.py`: the post-hoc `_set_kv_cache_constant_amax` override now only runs when `--recipe is None`, so a recipe YAML fully determines KV-cache config instead of being silently overridden by the default `--kv_cache_qformat=fp8_cast`. Updated help text on both flags.
- Extends the recipe loader smoke test to cover the three new recipes.

## Motivation

Before this change, the cast variants lived only in argparse (`_KV_CAST_FORMATS = {"fp8_cast", "nvfp4_cast"}`) and were layered on top of any recipe-loaded config. That meant `--recipe nvfp4_default-fp8_kv` would silently become a cast recipe due to the `--kv_cache_qformat` default. Now the recipe is self-contained: its YAML either sets `use_constant_amax: true` on the `*[kv]_bmm_quantizer` entry (cast) or doesn't (data-driven calibration).

## Test plan

- [x] `pytest tests/unit/recipe/test_loader.py` — all 24 tests pass, including the three new parametrized recipes.
- [x] Verified each new recipe round-trips through `load_recipe()` with `use_constant_amax: True` surviving Pydantic validation on the KV entry.
- [x] End-to-end run on `/models/Qwen/Qwen3-8B` (RTX 6000 Ada, 4 samples, seq_len=128) for all three new recipes:
  - After `mtq.quantize(model, recipe.quantize.model_dump(), forward_loop=...)`, all 72 `k_bmm_quantizer` / `v_bmm_quantizer` modules have `_use_constant_amax=True` and `_get_amax()` returns `448.0` (FP8 E4M3 max).
  - Weight quantizers still calibrate from data normally (sample amax values: q_proj=0.5508, k_proj=0.6250, v_proj=0.1689, o_proj=0.7266).
- [x] Verified the `--recipe` authoritative behavior change:
  - Non-cast recipe + default `--kv_cache_qformat=fp8_cast` → KV entry does NOT get `use_constant_amax` (no silent override).
  - Cast recipe + contradictory `--kv_cache_qformat=fp8` → KV entry keeps `use_constant_amax=True` (recipe wins).

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI now preserves KV cache quantization settings specified in recipe YAML (no forced constant amax override).

* **New Features**
  * Added three PTQ recipes for FP8 and NVFP4 with cast-KV variants and optimized KV cache handling.

* **Documentation**
  * Expanded CLI help for --recipe and KV cache options with cast-recipe examples and behavior notes.

* **Tests**
  * Added unit test coverage to load and validate the new PTQ recipes.

### DIFF
--- a/modelopt_recipes/general/ptq/fp8_default-fp8_cast_kv.yaml
+++ b/modelopt_recipes/general/ptq/fp8_default-fp8_cast_kv.yaml
@@ -1,0 +1,70 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metadata:
+  recipe_type: ptq
+  description: >-
+    FP8 per-tensor weight and activation (W8A8), FP8 KV cache with constant amax
+    (skips KV calibration; amax hardcoded to FP8 E4M3 max 448.0), max calibration.
+quantize:
+  algorithm: max
+  quant_cfg:
+    - quantizer_name: '*'
+      enable: false
+    - quantizer_name: '*input_quantizer'
+      cfg:
+        num_bits: e4m3
+        axis:
+    - quantizer_name: '*weight_quantizer'
+      cfg:
+        num_bits: e4m3
+        axis:
+    - quantizer_name: '*[kv]_bmm_quantizer'
+      enable: true
+      cfg:
+        num_bits: e4m3
+        use_constant_amax: true
+    - quantizer_name: '*block_sparse_moe.gate*'
+      enable: false
+    - quantizer_name: '*linear_attn.conv1d*'
+      enable: false
+    - quantizer_name: '*lm_head*'
+      enable: false
+    - quantizer_name: '*mixer.conv1d*'
+      enable: false
+    - quantizer_name: '*mlp.gate.*'
+      enable: false
+    - quantizer_name: '*mlp.shared_expert_gate.*'
+      enable: false
+    - quantizer_name: '*output_layer*'
+      enable: false
+    - quantizer_name: '*proj_out.*'
+      enable: false
+    - quantizer_name: '*router*'
+      enable: false
+    - quantizer_name: 'output.*'
+      enable: false
+    - parent_class: 'nn.BatchNorm1d'
+      quantizer_name: '*'
+      enable: false
+    - parent_class: 'nn.BatchNorm2d'
+      quantizer_name: '*'
+      enable: false
+    - parent_class: 'nn.BatchNorm3d'
+      quantizer_name: '*'
+      enable: false
+    - parent_class: 'nn.LeakyReLU'
+      quantizer_name: '*'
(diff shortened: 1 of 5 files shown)


meenchen (commented):
The PR is well-motivated and the logic is correct. The YAML recipes are consistent with existing ones, and the `hf_ptq.py` change correctly gates the post-hoc override. One minor issue: copyright years in new files should be 2025, not 2024. Also noting that these three new YAML files share ~90% of their content with the existing non-cast variants — the upcoming `$import` system (PR #1253) would be a great fit to deduplicate, but that's not blocking.

modelopt_recipes/general/ptq/fp8_default-fp8_cast_kv.yaml
```diff
@@ -0,0 +1,70 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
meenchen, line 1: Minor: Copyright year should be 2025 for new files, not 2024 (same applies to the other two new YAML files).

modelopt_recipes/general/ptq/nvfp4_default-nvfp4_cast_kv.yaml
```diff
@@ -0,0 +1,82 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metadata:
+  recipe_type: ptq
+  description: >-
+    NVFP4 W4A4, NVFP4 KV cache with constant amax (skips KV calibration; amax
+    hardcoded to FP8 E4M3 max 448.0), max calibration.
```
meenchen: Nit: The description says "amax hardcoded to FP8 E4M3 max 448.0" but this recipe uses NVFP4 (e2m1) for KV cache. The constant amax value for e2m1 would be different from FP8 E4M3's 448.0. Is the description accurate, or should it say something like "amax hardcoded to the format's max representable value"?

meenchen (approved):

shengliangxu (approved):
LGTM

modelopt_recipes/general/ptq/nvfp4_default-nvfp4_cast_kv.yaml
cjluo-nv: Good catch on reading it literally — but 448.0 is intentional here for NVFP4 KV too, not a typo.

`use_constant_amax` always returns `torch.finfo(torch.float8_e4m3fn).max` at runtime, regardless of the quantizer's `num_bits`:

https://github.com/NVIDIA/Model-Optimizer/blob/main/modelopt/torch/quantization/nn/modules/tensor_quantizer.py#L644-L647

This is deliberate — at deployment, the attention kernel upcasts the NVFP4 KV cache values to FP8 before the attention math, so the scale has to land in the FP8 E4M3 range (448.0). The existing docstring on `QuantizerAttributeConfig.use_constant_amax` calls this out explicitly:

> This is used for KV cache quantization where the downstream engine uses FP8 attention math for both FP8 and NVFP4 quantization, so the amax is hardcoded to the FP8 range.

So the YAML description is accurate — "amax hardcoded to FP8 E4M3 max 448.0" correctly reflects what happens even for the NVFP4 KV recipe. I'll add a short note in the YAML description pointing at this rationale so the next reader isn't surprised.

PR #1341, opened 2026-04-24
DSV4 PTQ example with dequant on the fly

mxinO:
### What does this PR do?

Type of change: new example 

Add deepseek v4 official modeling ptq example

### Usage

See readme, and it requires the vllm PR: https://github.com/vllm-project/vllm/pull/42209

### Testing
Tested with ptq and export of dsv4 flash and served with vllm.

## Summary by CodeRabbit

* **New Features**
  * Added DeepSeek‑V4 routed‑expert post‑training quantization and an NVFP4 checkpoint conversion utility.

* **Documentation**
  * Expanded DeepSeek quantization guide with directory layout, updated V3/V3.2 workflows, and detailed V4 routed‑expert calibration, single/multi‑node examples, and export guidance.

* **Chores**
  * Made example quantization scripts location‑independent.
  * Updated pre‑commit license hook to skip DeepSeek example quantization files.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1341?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -0,0 +1,524 @@
...
+from modelopt.torch.utils.dataset_utils import get_dataset_dataloader
+from modelopt.torch.utils.distributed import ParallelState
+
+
+_DEFAULT_V4_DIR = (
+    Path(__file__).resolve().parent.parent.parent.parent / "models/DeepSeek-V4-Pro/inference"
+)
+
+
+def _inject_v4_module(v4_inference_dir: Path) -> None:
```
cjluo-nv, line 85: Can we move this to examples/deepseek instead of creating a new example?

Edwardf0t1:
@mxinO could we land this PR as we plan to release dsv4 nvfp4 ckpt?

kevalmorabia97 (approved):
approving as pre-commit codeowner

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
```diff
@@ -0,0 +1,654 @@
...
+
+        <path>.weight          NVFP4-packed uint8, shape (out, in//2)
+        <path>.weight_scale    per-block scale (E4M3), shape (out, in//16)
+        <path>.weight_scale_2  per-tensor scale (FP32 scalar)
+        <path>.input_scale     per-tensor activation scale (FP32 scalar)
+
+  * An updated ``model.safetensors.index.json`` reflecting dropped/added keys.
+  * ``config.json`` keeps the source FP8 quantization metadata and adds
+    ``"moe_quant_algo": "NVFP4"`` for DeepSeek-V4 loaders.
+  * An ``hf_quant_config.json`` manifest listing the NVFP4-quantized layers.
```
Edwardf0t1: Is it possible to follow our standardized quant config format? 
e.g., config.json: https://huggingface.co/nvidia/Kimi-K2.6-NVFP4/blob/main/config.json#L17-L174 
hf_quant_config.json: https://huggingface.co/nvidia/Kimi-K2.6-NVFP4/blob/main/hf_quant_config.json

Otherwise the deployment fw needs special handlings, see a review comment in SGLang: https://github.com/sgl-project/sglang/pull/25820#discussion_r3314045651

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -0,0 +1,548 @@
...
+    quantized layer paths. The manifest is built by scanning the model for
+    ``TensorQuantizer`` instances whose path contains ``.experts.<n>.w``
+    (i.e. routed-expert quantizers only, excluding ``shared_experts``); we do
+    *not* rely on ``modelopt.torch.export.quant_utils.get_quant_config``
+    because its introspection doesn't see weights stored on nested
+    submodules — ``QuantExpert``'s ``w{1,2,3}`` are submodules of the
+    container, not direct parameters. The downstream export script
+    (``quantize_to_nvfp4.py``) uses this manifest as ground truth for which
+    tensor paths to replace with NVFP4 packed weight + scales.
+    """
```
Edwardf0t1, line 354: It's not a docstring, because the triple-quoted string follows executable statements, it's a no-op expression, not a docstring (won't show in help()/tooling). Move it to the top of the function.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -0,0 +1,548 @@
...
+                gate.topk = orig_topk
+
+    mtq.register(original_cls=deekseep_v4_model.Expert, quantized_cls=QuantExpert)
+    # CalibMoE registration is skipped for first-sanity runs — see
+    # --calib_all_experts flag below. Registering it forces every token
+    # through every local expert which populates amax on all experts but
+    # multiplies per-batch work ~64x and surfaces NCCL-timing issues on
+    # multi-node runs. Default behavior: natural top-k routing, fewer
+    # experts observed per batch.
+    if os.environ.get("CALIB_ALL_EXPERTS", "0") == "1":
```
Edwardf0t1: There is no such argparse flag in V4's main() — that flag exists in the V3 ptq.py (per CHANGELOG). This looks like a copy-paste leftover.

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
```diff
@@ -0,0 +1,654 @@
...
+    filesystems, which is common with container mounts.
+    """
+    for item in src.iterdir():
+        if item.name in _SKIP_TOP_LEVEL:
+            continue
+        if _HF_SHARD_RE.match(item.name):
+            continue
+        target = dst / item.name
+        if item.is_file():
+            if target.exists():
```
cjluo-nv, line 410: > _Bot comment._

The emitted `hf_quant_config.json` schema (`producer`/`quantization.quantized_layers`) doesn't match the standardized NVIDIA NVFP4 layout @Edwardf0t1 pointed to (e.g. `nvidia/Kimi-K2.6-NVFP4`). Per the linked SGLang review thread, deviating here forces deployment frameworks to add special-case handling. Please confirm whether this format is what vLLM PR #42209 actually consumes, and if not, align it with the standard format before landing.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -0,0 +1,548 @@
...
+        mtq.print_quant_summary(model)
+    _trace("ptq() returning")
+    return model
+
+
+def save_amax_and_quant_config(model, output_path: str):
+    def _trace(msg):
+        print(f"[rank {os.getenv('RANK', '0')}] SAVE: {msg}", flush=True)
+
+    _trace("entered")
```
cjluo-nv: > _Bot comment._

This triple-quoted string is still a no-op statement (it sits after `_trace("entered")`), not a docstring — `help(save_amax_and_quant_config)` won't show it. @Edwardf0t1 and CodeRabbit both flagged this in the previous round. Move it directly under the `def` line and put the `_trace` call after it.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+def load_deepseek_v4(
+    model_config: str, model_path: str, batch_size: int, dummy_weights: bool = False
+):
+    world_size = int(os.getenv("WORLD_SIZE", "1"))
+    rank = int(os.getenv("RANK", "0"))
+    local_rank = int(os.getenv("LOCAL_RANK", "0"))
+    if world_size > 1:
+        dist.init_process_group("nccl")
```
cjluo-nv, line 225: > _Bot comment._

The `CALIB_ALL_EXPERTS` env-var-only switch is undocumented in the README and inconsistent with V3, which exposes `--calib_all_experts` on argparse. Please either add the CLI flag here or document the env var explicitly in `examples/deepseek/README.md` so users have a discoverable path.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -0,0 +1,548 @@
...
+        We do NOT do a second "real" forward pass — MoE has an internal
+        ``dist.all_reduce(y)`` and the double-forward form kept collectives
+        misaligned when TileLang JIT compile times varied across ranks,
+        producing 10-minute NCCL timeouts. Calibration output is discarded
+        by ``calibrate_loop`` anyway so a single all-expert forward is
+        sufficient.
+
+        Empty ``_setup`` because this wrapper installs no quantizer state;
+        it exists solely to override ``forward``. ``DynamicModule.convert``
+        still requires the method to exist."""
```
cjluo-nv: > _Bot comment._

The comment "We do NOT do a second 'real' forward pass" describes a design that isn't visible here — `forward` only calls `super().forward(x, input_ids)` once, with `topk` temporarily widened. Reading it cold suggests something is missing. Either tighten the comment to describe the single-pass behavior or drop the historical context.

cjluo-nv (commented):

Several previous critical review comments still appear unaddressed in the latest commit:

- :speech_balloon: @Edwardf0t1 asked about following the standardized NVIDIA NVFP4 quant config format (referencing Kimi-K2.6-NVFP4 `config.json` / `hf_quant_config.json`) so deployment frameworks (vLLM, SGLang) don't need special handling. The current `_write_index_and_manifest` and `_rewrite_config_json` in `deepseek_v4/quantize_to_nvfp4.py` still emit a custom shape (`producer`, `quantization.quantized_layers`, `moe_quant_algo: "NVFP4"`). I don't see a reply or commit addressing this — given the SGLang reviewer feedback linked, this looks like a deployment blocker.
- :speech_balloon: @Edwardf0t1 flagged the descriptive triple-quoted string in `save_amax_and_quant_config` (now around line 388) as a no-op expression because it follows executable statements (`_trace("entered")`). It still does. CodeRabbit raised the same point. Trivial to fix.
- :speech_balloon: @Edwardf0t1 noted the V4 `main()` has no `--calib_all_experts` argparse flag (only the `CALIB_ALL_EXPERTS` env var controls it). The README still documents `--calib_all_experts` for the V3 path at line 88, but that command also still uses bare `ptq.py` rather than `deepseek_v3/ptq.py` — CodeRabbit flagged this as an "outside diff" issue and it remains unfixed. This is a copy/paste hazard for users.
- Test coverage: this is a 1292-line new example with no unit tests and no CI smoke test. The PR body says "Tested with ptq and export of dsv4 flash and served with vllm" but there's nothing in-tree to keep that wired up. Given the export script's complexity (amax fusing for w1/w3, MTP handling, w13 weight-amax overrides, hard-link aux dir walker), some unit-level coverage on at least `_fuse_w1_w3_amax`, `_validate_paths`, and `_build_w13_weight_amax_overrides` would help.

Smaller concerns:
- `import re as _re` and `import time as _time` inside functions — no apparent reason for these to be local; please move to module top.
- `os.environ.get("CALIB_ALL_EXPERTS", "0")` is read inside `install_quant_registry` and surfaces nowhere on the CLI; either expose it as `--calib_all_experts` (matching V3) or document the env var in the README.
- `CalibMoE` docstring claims "we do NOT do a second 'real' forward pass", but `forward` does call `super().forward(...)` which is the only forward — the comment phrasing reads like leftover from an earlier two-pass design and could confuse future readers.

Recommend addressing the standardized quant config question with @Edwardf0t1 before merging since it directly affects whether the released checkpoint is consumable by vLLM/SGLang without bespoke handling.

---

- `examples/deepseek/README.md:88` — > _Bot comment._

Still references `ptq.py` rather than `deepseek_v3/ptq.py` like the other DeepSeek V3 commands above. CodeRabbit flagged this in the prior review. As written, the documented command will fail from the `examples/deepseek/` root.

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
cjluo-nv, line 410: > _Bot comment._

**Quant config schema deviates from the standardized NVIDIA NVFP4 layout.** @Edwardf0t1 flagged this in the prior review (see `nvidia/Kimi-K2.6-NVFP4/hf_quant_config.json`) and the SGLang reviewer pointed out that deviating here forces deployment frameworks to add a special-case path. The current emission uses `producer` / `quantization.quantized_layers` / `moe_quant_algo: "NVFP4"` which doesn't match. Please either align with the standard schema or document explicitly in the PR body why vLLM PR #42209 requires this divergent format and confirm SGLang owners are okay with the special-case handling. Treating this as a merge blocker.

examples/deepseek/deepseek_v4/ptq.py
cjluo-nv: > _Bot comment._

Still a no-op expression, not a docstring — it sits after `_trace("entered")`. `help(save_amax_and_quant_config)` won't surface it. Move the triple-quoted block immediately under the `def` line and put `_trace("entered")` after it. Flagged in the previous round by @Edwardf0t1 and CodeRabbit.

examples/deepseek/deepseek_v4/ptq.py
cjluo-nv, line 225: > _Bot comment._

`CALIB_ALL_EXPERTS` is read from the env here but isn't an argparse flag and isn't mentioned in the README, while V3 exposes the equivalent as `--calib_all_experts`. Please either add the CLI flag for parity with V3 or document the env-var contract in `examples/deepseek/README.md` so users have a discoverable path.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -0,0 +1,548 @@
...
+    ``TensorQuantizer`` instances whose path contains ``.experts.<n>.w``
+    (i.e. routed-expert quantizers only, excluding ``shared_experts``); we do
+    *not* rely on ``modelopt.torch.export.quant_utils.get_quant_config``
+    because its introspection doesn't see weights stored on nested
+    submodules — ``QuantExpert``'s ``w{1,2,3}`` are submodules of the
+    container, not direct parameters. The downstream export script
+    (``quantize_to_nvfp4.py``) uses this manifest as ground truth for which
+    tensor paths to replace with NVFP4 packed weight + scales.
+    """
+    import re as _re
```
cjluo-nv: > _Bot comment._

`import re as _re` (and `import time as _time` inside `ptq` around L342) — these should be at module top per the project's coding conventions. There's no circular-import or heavy-optional-dep reason for them to be local.

examples/deepseek/deepseek_v4/ptq.py
cjluo-nv: > _Bot comment._

The comment "We do NOT do a second 'real' forward pass" describes a design that isn't visible in this code — `forward` only calls `super().forward(x, input_ids)` once. Reading cold, it sounds like something is missing. Either rewrite the comment to describe the actual single-pass behavior (and why double-forward was rejected — the NCCL timeout rationale is good context) or trim the historical framing.

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
```diff
@@ -0,0 +1,654 @@
...
+    return match.group("experts")
+
+
+def _validate_paths(source_ckpt: Path, output_ckpt: Path) -> None:
+    source_resolved = source_ckpt.resolve()
+    output_resolved = output_ckpt.resolve()
+    if (
+        output_resolved == source_resolved
+        or source_resolved in output_resolved.parents
+        or output_resolved in source_resolved.parents
```
cjluo-nv, line 551: > _Bot comment._

Operator-context note: this PR is a 1292-line new example with no unit tests. `_fuse_w1_w3_amax`, `_validate_paths`, and `_build_w13_weight_amax_overrides` are all pure-Python with well-defined contracts — adding even a small `tests/unit/examples/deepseek/test_v4_quantize_to_nvfp4.py` covering these would protect against silent regressions on future format changes. The `_validate_paths` fix (parent-direction check) noted in commit a21c478 looks correct.

cjluo-nv (commented):

Per the operator's request, here's a substantive note on **why `deepseek_v3/` and `deepseek_v4/` cannot reasonably be merged into a single example** even though there is surface-level duplication (argparse scaffolding, `load_model` flow, calibration loop, the `save_amax_and_quant_config` skeleton). The split is driven by genuine architectural divergence in the upstream DeepSeek code, not by author preference:

1. **Upstream module geometry is different.** In V3's `inference/model.py`, MoE experts expose `w1/w2/w3` as `Linear` instances directly, so V3 can register `QuantLinear`/`QuantColumnParallelLinear`/`QuantRowParallelLinear` once and let pattern matching handle the rest. In V4, the routed expert is a single `Expert` *module* whose `w{1,2,3}` are submodules; ModelOpt's introspection (`get_quant_config`) doesn't descend into them, which is why `save_amax_and_quant_config` in V4 has to walk `named_modules()` and build its own manifest. Reusing V3's registry on V4 would either miss the experts entirely or, if you wrapped `Linear` globally, force *every* linear (attention, gate, shared expert, lm_head) through a BF16 dequant + `F.linear` path — exactly what the V4 recipe is trying to avoid (see the design-notes block at the top of `deepseek_v4/ptq.py`).

2. **On-disk weight formats are incompatible.** V3 stores experts as FP8 E4M3 + FP32 block scale and uses `modelopt.torch.kernels.quantization.gemm.weight_dequant`. V4 mixes MXFP4 (`float4_e2m1fn_x2`) + UE8M0 1×32 scale on routed experts with FP8 E4M3 + UE8M0 128×128 scale elsewhere. The existing `weight_dequant` Triton kernel does not accept UE8M0 scales, so V4 needs its own `_fp8_ue8m0_blockwise_to_bf16` and `MXFP4QTensor.dequantize` paths inline. These are not the same dequant routine with a flag — they consume different scale dtypes and block sizes.

3. **Calibration topology differs.** V3's `CalibMoe.forward` runs the model twice (once with `topk = n_routed_experts`, once with the real top-k). V4's `CalibMoE` deliberately does *only* the all-expert pass because V4's `MoE.forward` contains an internal `dist.all_reduce(y)` and TileLang JIT-compile time varies across ranks; the double-forward form produced 10-minute NCCL timeouts in multi-node runs. That's a hard constraint specific to V4's TileLang kernels, not a stylistic choice.

4. **Export-side geometry is different.** V3's `quantize_to_nvfp4.py` rebuilds an HF release from MP-sharded amax dumps via the `quantize_fp8_to_nvfp4.sh` pipeline. V4 goes original-HF → new-HF directly, must fuse w1/w3 weight amax for vLLM's fused GEMM1, has bespoke MTP handling (`--include_mtp_experts`), and a hard-link/copy aux walker for cross-filesystem container mounts. None of that maps cleanly onto V3's pipeline.

So the *correct* refactor (if you wanted to reduce duplication) is not to merge the examples, it's to factor out the small genuinely-shared bits — argparse skeleton, `_run_quantized_generate`, the rank-aware `_trace` logger — into something like `examples/deepseek/_common.py`. Trying to unify the registries, dequant kernels, calib MoE, and export pipelines would be strictly worse than what's here.

---

**That said, I'm leaving this as `comment` rather than approve/nudge because several critical items from the prior round are still unaddressed:**

- :speech_balloon: @Edwardf0t1 asked to align the emitted `hf_quant_config.json` with the standardized NVIDIA NVFP4 layout (referencing `nvidia/Kimi-K2.6-NVFP4`). The current `_write_index_and_manifest` still emits a custom `producer` / `quantization.quantized_layers` shape with `moe_quant_algo: "NVFP4"`. SGLang's reviewer flagged this as forcing special-case handling. No reply or commit visible. **This is a deployment blocker** — please confirm with @Edwardf0t1 before landing.
- :speech_balloon: The triple-quoted string in `save_amax_and_quant_config` (`deepseek_v4/ptq.py` ~L388) was flagged twice as a no-op expression rather than a docstring. Still after `_trace("entered")`. Two-line fix.
- :speech_balloon: README L88 (`--calib_all_experts` example) still uses bare `ptq.py` instead of `deepseek_v3/ptq.py` — flagged by CodeRabbit and cjluo-nv, still broken as written.
- `CALIB_ALL_EXPERTS` in V4 is env-var-only and undocumented; V3 exposes the equivalent as `--calib_all_experts`. Please either add the CLI flag or document the env var in the README.
- No tests at all for ~1.3k lines of new code. At minimum, unit coverage on `_fuse_w1_w3_amax`, `_validate_paths`, and `_build_w13_weight_amax_overrides` would be valuable since these are pure-Python with clear contracts.
- `import re as _re` / `import time as _time` inside functions in `deepseek_v4/ptq.py` — no stated reason, please move to module top per the project's coding conventions.

The design-level separation is fine. The deployment-format and doc/test issues should be resolved before merge.

---

- `examples/deepseek/README.md:88` — > _Bot comment._

This `--calib_all_experts` example still invokes `ptq.py` (bare), but the surrounding V3 examples were updated to `deepseek_v3/ptq.py`. As written, the documented command will fail when run from `examples/deepseek/`. Flagged by CodeRabbit and cjluo-nv in the prior round.

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
mxinO: Thanks for pointing it out, I have changed the config format, the config.json now includes something like this, tested with nightly vllm,
```
"quantization_config": {
    ..., # nvfp4 format
    "fmt": "e4m3",
    "group_size": 16,
    "ignore": [
      "*.attn.*",
      "*.ffn.shared_experts.*",
      "head",
      "mtp.*"
    ],
    "kv_cache_quant_algo": null,
    "moe_quant_algo": "NVFP4",
    "producer": {
      "name": "modelopt",
      "version": "dsv4-nvfp4-experts"
    },
    "quant_algo": "MIXED_PRECISION",
    "quant_method": "fp8",
    "quantized_layers": {
      "layers.0.ffn.experts": {
        "group_size": 16,
        "quant_algo": "NVFP4"
      },
      ...
     }
}

examples/deepseek/deepseek_v4/ptq.py
mxinO: Thanks, removed, forcing to use all experts is not safe.

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
mxinO: A unit test is not necessary, it's quantizing the dsv4's official implementation, that doesn't change usually, and we don't have unit tests for dsv3 either. The example is more like to illustrate how to quantize a custom model.

Edwardf0t1 (approved):
LGTM - Is there an exported NVFP4 checkpoint we can take a look? I assume the new ckpt is only tested in vLLM not SGLang, is that right?

cjluo-nv (commented):

Re-review of the DSV4 PTQ example (1271 LOC, 7 files). Most prior critical comments have been resolved in the latest revision; flagging a few remaining items for human sign-off.

**Resolved since previous round:**
- :speech_balloon: Standardized NVFP4 quant config format — author updated `_rewrite_config_json` to emit the standard `config_groups` / `ignore` / `quant_algo: MIXED_PRECISION` / `quant_method: fp8` / `kv_cache_quant_algo` / `weight_block_size` fields, tested against nightly vLLM. Edwardf0t1 LGTM'd this resolution.
- :speech_balloon: `--calib_all_experts` copy-paste flag — author removed it (and the `CALIB_ALL_EXPERTS` env var) from V4. ✅
- :speech_balloon: Triple-quoted string no-op in `save_amax_and_quant_config` — now correctly placed as the first statement under `def`, before the `_trace` helper. ✅
- :speech_balloon: `import re as _re` / `import time as _time` inside functions — moved to module top. ✅
- :speech_balloon: `_validate_paths` parent-direction check — fixed in commit a21c478, both source-in-output and output-in-source are blocked. ✅
- :speech_balloon: README L88 — now uses `deepseek_v3/ptq.py`. ✅
- :speech_balloon: V3/V4 split vs single example — cjluo-nv themselves wrote a detailed analysis concluding the split is architecturally correct (different module geometry, on-disk formats, calib topology, export pipeline). Design question settled.
- :speech_balloon: `CalibMoE` confusing comment — class has been removed entirely. ✅

**Remaining points worth a human eye before merge:**
- :speech_balloon: Author replied: *"A unit test is not necessary… we don't have unit tests for dsv3 either. The example is more like to illustrate how to quantize a custom model."* — consistent with the V3 precedent, but pure-Python helpers like `_fuse_w1_w3_amax`, `_validate_paths`, `_build_w13_weight_amax_overrides`, and `_amax_to_nvfp4_scale_2` are easy to unit-test and would protect against silent regressions on future V4 release format changes. Worth a maintainer judgment call given how recently the config schema was reshaped.
- The emitted `config.json` still embeds non-standard fields alongside the standardized ones (`producer`, `moe_quant_algo`, `quantized_layers` map). Edwardf0t1 LGTM'd, but if SGLang ingestion is also in scope it's worth confirming the extra fields don't trip their loader (referenced SGLang review thread asked for the standardized layout specifically).
- Size: 1271 LOC in a single PR is on the large side, but the changes are cohesive (one new example) and not reasonably splittable without losing the ptq → export round-trip story.

No correctness blockers spotted in the current revision. Edwardf0t1 has already approved and is requesting merge for the DSV4 NVFP4 release; recommend a human reviewer make the final landing call.

cjluo-nv (approved):

PR #1342, opened 2026-04-24
Support EP mcore import for TE  Spec and Fix mamba moe config

jenchen13:
### What does this PR do?

Type of change: Bug fix

- Enable EP (expert parallelism) import for HF to MCore when using TE Spec
- Fix bug in mamba moe config which doesn't skip attention layers properly in MCore (Mcore uses different naming for attention layers than HF)
- Add getter for Quant Config (used in MLM modelopt examples to get quant cfg fields)

### Usage

```python
# In Megatron-LM/examples/post_training/modelopt
 MLM_EXTRA_ARGS="--export-default-te-spec --trust-remote-code --moe-router-dtype fp32" EP=4 HF_MODEL_CKPT=</path/to/hf> MLM_MODEL_SAVE=<save/path> ./convert.sh nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected expert-slice assignment so each expert-parallel rank loads the proper expert slice.
  * Improved detection of pipeline-parallel layer indices in submodule names.

* **Improvements**
  * Relaxed constraints between local and global expert counts for grouped-local-expert imports.
  * Added typed helpers for managing quantization configuration entries and expanded quantizer disable patterns.
  * Exporter now accepts an additional hybrid model type when available.


modelopt/torch/quantization/config.py
```diff
@@ -170,6 +170,34 @@ class QuantizerCfgEntry(TypedDict, total=False):
...
 
 
+QuantizerCfg = list[QuantizerCfgEntry]
+
+
+def _set_quant_cfg_entry(
+    quant_cfg: QuantizerCfg, quantizer_name: str, entry_cfg: QuantizerCfgEntry
+) -> None:
+    """Set a QuantizerCfgEntry by quantizer_name in the QuantizerCfg."""
+    for entry in quant_cfg:
```
shengliangxu: don't do this, just append, the latter one will overwrite the former one.

if you to keep the function, rename it to something like:

def _overwrite_quant_cfg

shengliangxu: Oh, one difference:

the dict update will do a merge, the semantic of an arbitrary merge is not clear to me.

jenchen13: oh i see, so the setter is not necessary because we can just append the QuantizerCfgEntry, but the getter is useful? I removed the setter

modelopt/torch/quantization/config.py
```diff
@@ -170,6 +170,17 @@ class QuantizerCfgEntry(TypedDict, total=False):
     enable: bool | None  # toggles matched quantizers on/off; independent of cfg
 
 
+QuantizerCfg = list[QuantizerCfgEntry]
+
+
+def _get_quant_cfg_entry(quant_cfg: QuantizerCfg, quantizer_name: str) -> QuantizerCfgEntry | None:
```
shengliangxu: find_quant_cfg_entry_by_path is already doing this.

Although, a simple equality get is really not meaningful because the quantizer_name is actually a wildcard string.

But too many places just simply use equality comparison to locate an existing config entry, so have to keep that function.

jenchen13: hmm ok but it raises ValueError on non-existent key which is very strict. I can use `find_quant_cfg_entry_by_path` but curious why it needs to raise an error

shengliangxu: I actually let Claude code written for me, it chooses to raise an error instead of return a None, both are fine to me, so I just let it be. lol.

shengliangxu: I think the major reason is I just do not like this function at all for the above reason, so do not bother try to improve it in any way.

shengliangxu (approved):

ChenhanYu (approved):

kevalmorabia97 (approved):

PR #1345, opened 2026-04-24
[OMNIML-4021]: align local JSONL loading with HF datasets path + keep original behaviour

shengliangxu:
## Summary

Local `.jsonl` paths fed to `get_dataset_samples` / `get_dataset_dataloader` previously went through a `text`-key-only reader, while HF dataset names flowed through an auto-preprocess pipeline that recognizes `messages` / `conversations` / `prompt` / `text` / `input` columns. This split meant a calibration dataset behaved differently depending on whether it lived on HF Hub or on disk.

This PR routes local `.jsonl` through HF's `json` builder and the same auto-preprocess pipeline so the format is detected from columns, not the source. The legacy `text`-field reader is preserved as a fallback for files where the HF builder fails (e.g. PyArrow schema unification across heterogeneous rows). Existing callers passing a plain `{"text": ...}` JSONL file keep working unchanged; chat-shaped JSONL now works without a separate code path.

## Changes

- **`modelopt/torch/utils/dataset_utils.py`**
  - Route `.jsonl` paths through `load_dataset(path="json", data_files=...)` + `_auto_preprocess_sample`.
  - Wrap the HF load/preprocess block in `try/except`; on failure for `.jsonl`, fall back to `get_jsonl_text_samples`. If the fallback also fails, re-raise the original HF error so the diagnostic is preserved.
  - Update docstrings; clarify the `splits = ["train"]` invariant for HF's file-based builders.

## Tests

`tests/unit/torch/utils/test_dataset_utils.py` — three new test classes (18 cases, all passing):

- `TestLocalJsonlLoading` — text / messages / conversations / prompt+completion / input+output columns; `num_samples` honored; `tools` kwarg forwarded; ValueError on unrecognized columns; legacy `text`-key fallback on schema-unification failure.
- `TestGetDatasetDataloaderBlending` — single JSONL, list of JSONL files concatenated, mixed-format JSONL files blended, length-mismatch assertion.
- `TestHfTinyDataset` — uses `hf-internal-testing/dataset_with_data_files` (10 rows x {train, test}) for end-to-end coverage: single split, multiple splits, default split, HF -> JSONL -> reload round-trip, two-HF-dataset blending, HF + local-JSONL mixing.

## Test plan

- [x] `pytest tests/unit/torch/utils/test_dataset_utils.py` — 28 passed
- [x] `mypy modelopt/torch/utils/dataset_utils.py` — no new errors
- [x] `ruff check` — clean

## Summary by CodeRabbit

* **New Features**
  * Local JSONL files now load through the HuggingFace JSON pipeline so local and HF datasets share the same auto-detection and preprocessing; lists can mix dataset identifiers and JSONL paths (local JSONL restricted to train).

* **Bug Fixes**
  * Fields explicitly set to null/None are no longer treated as present text, avoiding incorrect concatenation or chat-template application.
  * Narrower legacy fallback with clearer warnings when schema/parsing failures occur.

* **Documentation**
  * Updated to describe mixed-source handling and concatenation-before-tokenization.

* **Tests**
  * Expanded unit and integration tests covering schemas, sparse/null handling, concatenation, sample limits, tool forwarding, and fallback paths.


shengliangxu:
@coderabbitai review

pyproject.toml
```diff
@@ -307,7 +307,9 @@ skips = [
 addopts = "-v -ra --instafail --cov-report=term-missing --cov-report=html --cov-report=xml:coverage.xml --cov-config=pyproject.toml --durations=50 --strict-markers"
 pythonpath = ["tests/"]
 markers = [
+    "integration: Tests that require external services or other non-hermetic dependencies",
     "manual: Only run when --run-manual is given",
+    "network: Tests that require network access",
```
kevalmorabia97: there are many other tests that require network. Should this marker be added everywhere?

kevalmorabia97 (approved):

pyproject.toml
shengliangxu: yeah, it's unintentional change, will remove it.

PR #1142, opened 2026-03-30
[NVBug: 6000530] Fix AWQ crash for uncalibrated MoE experts

cjluo-nv:
## Summary
- Fixes NVBugs 6000530: `AttributeError: 'float' object has no attribute 'pow'` when running AWQ lite with `moe_calib_experts_ratio < 1.0` on MoE models (e.g. Qwen3-30B-A3B).
- **Root cause**: When `moe_calib_experts_ratio=0.5`, some MoE experts receive zero tokens during the AWQ cache phase, leaving `act_scale` as a Python float `0.0` instead of a tensor. This causes two failures:
  1. **Search phase crash**: Uncalibrated experts crash in `get_scale()` because `float.pow()` doesn't exist.
  2. **Export crash**: Calibrated experts have `pre_quant_scale` but uncalibrated ones don't, causing `torch.stack()` to fail on mixed `None`/tensor values in `preprocess_linear_fusion()`.
- **Fix**: Handle uncalibrated experts (`num_cache_steps == 0`) in two stages:
  1. **Before search**: Disable AWQ search (`is_enabled = False`) to prevent `get_scale()` crash on float `act_scale`.
  2. **During postprocessing**: Max calibrate weights and apply a neutral (all-ones) `pre_quant_scale` so export can stack scaling factors consistently across all experts. The `pre_quant_scale` buffer must be registered outside `enable_weight_access_and_writeback` because HF accelerate's `post_forward` hook drops newly-registered submodule buffers.

## Test plan
- [x] Reproduce with `Qwen/Qwen3-30B-A3B`, `--qformat int4_awq`, `--moe_calib_experts_ratio 0.5` — verify no crash during calibration and export

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1354, opened 2026-04-27
[NVBUG: 6103846] Fix nvfp4_awq export for uncalibrated MoE experts

cjluo-nv:
## Summary

- NVBug: [6103846](https://nvbugspro.nvidia.com/bug/6103846) — `Qwen3-30B-A3B nvfp4_awq` quantization fails at export with `AssertionError: Modules have different quantization formats`.
- Root cause: in `model_calib.awq_lite`, MoE experts that end up disabled (NaN in act/weight scales, or no search-pass tokens) get `max_calibrate`-d but no `pre_quant_scale`. `get_quantization_format` then returns `nvfp4` for those experts while siblings stay `nvfp4_awq`. `unified_export_hf.requantize_resmooth_fused_llm_layers` groups all 128 experts of each linear name (gate_proj/down_proj/up_proj) and calls `preprocess_linear_fusion(..., resmooth_only=True)`, which asserts uniform format → fires for any single mismatched expert.
- Fix: unify the disabled-expert paths in the awq_lite postprocess loop so any expert with `is_enabled == False` (no cache hits, NaN scales, or no search-pass tokens) receives `max_calibrate` + a neutral all-ones `pre_quant_scale`, matching the existing behavior for `num_cache_steps == 0`. Emit a warning so users notice that calibration coverage is incomplete and accuracy may degrade.

## Test plan

- [x] `pytest tests/unit/torch/quantization/test_calib.py -k 'awq'` → 5 passed
- [x] End-to-end on `Qwen/Qwen3-30B-A3B` with `NVFP4_AWQ_LITE_CFG` and a small calib set that leaves many experts uncalibrated:
  - All 6144 gate_proj/up_proj/down_proj expert linears report `nvfp4_awq` (no mismatch)
  - `export_hf_checkpoint` succeeds with no `AssertionError`
  - The new "Forcing pre_quant_scale=1 ... may degrade accuracy" warning fires for each affected expert
- [x] Re-run via `examples/llm_ptq/hf_ptq.py` with the bug-report CLI (cnn_dailymail, batch_size=8, calib_size=64 — scaled down from 512 to fit budget) on B200:
  - 36 "the second time did not forward data through ..experts.X.{gate,up,down}_proj" warnings — i.e. the exact bug-triggering condition from the original NVBug log naturally reproduces
  - 2058 "Forcing pre_quant_scale=1" warnings — fix path activates for uncalibrated/disabled experts
  - 0 `AssertionError`s — export completes
  - `Quantized model exported to: /tmp/test_plan_qwen3-30b-a3b-nvfp4_awq` and post-PTQ generation works

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,10 +19,6 @@ Changelog
 
 - Add FP8 MHA quantization support for vision transformers. Adds an attention-aware ONNX post-processing pass (scale Mul / K-transpose move before Q, Q→DQ insertion on softmax output) in :class:`FP8QuantExporter <modelopt.onnx.export.fp8_exporter.FP8QuantExporter>`, per-instance nested-attention-wrapper skipping in the HF plugin, and ``nn.LayerNorm`` registration in ``QuantModuleRegistry`` so BMM input quantizers and LayerNorm output quantizers defined in FP8_DEFAULT_CFG are honored end-to-end. See `examples/torch_onnx/torch_quant_to_onnx.py <https://github.com/NVIDIA/Model-Optimizer/tree/main/examples/torch_onnx/torch_quant_to_onnx.py>`_ for the general timm-model quantize→ONNX workflow.
 
-**Bug Fixes**
-
-- Fix ``nvfp4_awq`` export ``AssertionError: Modules have different quantization formats`` for MoE models (e.g. Qwen3-30B-A3B) when some experts are not exercised by the calibration data. ``awq_lite`` now applies a neutral all-ones ``pre_quant_scale`` to any expert that ends up disabled (no cache-pass tokens, NaN scales, or no search-pass tokens) so its format remains ``nvfp4_awq``, consistent with the rest of the MoE block. A warning is emitted whenever this fallback fires.
-
 0.44 (2026-05-xx)
 ^^^^^^^^^^^^^^^^^
 
@@ -51,6 +47,7 @@ Changelog
 - Fix Minitron pruning (``mcore_minitron``) for MoE models. Importance estimation hooks were incorrectly registered for MoE modules and NAS step was hanging before this.
 - Fix TRT support for remote autotuning in ONNX Autotune from 10.16+ to 10.15+ and fix TRT versioning check to the ``trtexec`` version instead of the TRT Python API when using ``trtexec`` backend.
 - Exclude MatMul/Gemm nodes with K or N < 16 from ONNX INT8 and FP8 quantization. Such small-dimension GEMMs cannot efficiently use INT8/FP8 Tensor Cores and the added Q/DQ layers cause perf regressions in TensorRT. Honors Gemm ``transB`` when deriving K.
+- Fix ``nvfp4_awq`` export ``AssertionError: Modules have different quantization formats`` for MoE models (e.g. Qwen3-30B-A3B) when some experts are not exercised by the calibration data. ``awq_lite`` now applies a neutral all-ones ``pre_quant_scale`` to any expert that ends up disabled (no cache-pass tokens, NaN scales, or no search-pass tokens) so its format remains ``nvfp4_awq``, consistent with the rest of the MoE block. A warning is emitted whenever this fallback fires.
 
 **Misc**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,10 @@ Changelog
 
 - Add FP8 MHA quantization support for vision transformers. Adds an attention-aware ONNX post-processing pass (scale Mul / K-transpose move before Q, Q→DQ insertion on softmax output) in :class:`FP8QuantExporter <modelopt.onnx.export.fp8_exporter.FP8QuantExporter>`, per-instance nested-attention-wrapper skipping in the HF plugin, and ``nn.LayerNorm`` registration in ``QuantModuleRegistry`` so BMM input quantizers and LayerNorm output quantizers defined in FP8_DEFAULT_CFG are honored end-to-end. See `examples/torch_onnx/torch_quant_to_onnx.py <https://github.com/NVIDIA/Model-Optimizer/tree/main/examples/torch_onnx/torch_quant_to_onnx.py>`_ for the general timm-model quantize→ONNX workflow.
 
+**Bug Fixes**
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -49,6 +47,7 @@ Changelog
 - Fix Minitron pruning (``mcore_minitron``) for MoE models. Importance estimation hooks were incorrectly registered for MoE modules and NAS step was hanging before this.
 - Fix TRT support for remote autotuning in ONNX Autotune from 10.16+ to 10.15+ and fix TRT versioning check to the ``trtexec`` version instead of the TRT Python API when using ``trtexec`` backend.
 - Exclude MatMul/Gemm nodes with K or N < 16 from ONNX INT8 and FP8 quantization. Such small-dimension GEMMs cannot efficiently use INT8/FP8 Tensor Cores and the added Q/DQ layers cause perf regressions in TensorRT. Honors Gemm ``transB`` when deriving K.
+- Fix ``nvfp4_awq`` export ``AssertionError: Modules have different quantization formats`` for MoE models (e.g. Qwen3-30B-A3B) when some experts are not exercised by the calibration data. ``awq_lite`` now applies a neutral all-ones ``pre_quant_scale`` to any expert that ends up disabled (no cache-pass tokens, NaN scales, or no search-pass tokens) so its format remains ``nvfp4_awq``, consistent with the rest of the MoE block. A warning is emitted whenever this fallback fires.
```
kevalmorabia97, line 50: Is this bug intruced and fixed both in current release - if so, it should not be mentioned here as it wont be relevant for external users. Only if it was fixed in current release but was an issue in previous we should add here

meenchen (commented):
The fix correctly addresses the `AssertionError: Modules have different quantization formats` bug for MoE models with uncalibrated experts during nvfp4_awq export.

**Root cause & fix**: Previously, three different "disabled expert" paths had inconsistent handling:
1. `num_cache_steps == 0` → neutral `pre_quant_scale=1` ✓
2. NaN in scales → `is_enabled = False`, fell through to `postprocess()` which did `max_calibrate` but **no** `pre_quant_scale` ✗
3. `num_cache_steps > 0 and num_search_steps == 0` → same as #2 ✗

Cases 2 and 3 resulted in `get_quantization_format()` returning `nvfp4` (no `_pre_quant_scale` attribute) instead of `nvfp4_awq`, which triggers the assertion in `preprocess_linear_fusion(..., resmooth_only=True)` when grouping experts.

The PR unifies all three cases under a single `not module.awq_lite.is_enabled` check that applies `max_calibrate` + neutral `pre_quant_scale=1`, ensuring consistent `nvfp4_awq` format. The hoisting of the `num_search_steps == 0` check before the branch is clean and doesn't change semantics for the enabled path.

**Minor note**: For the `num_cache_steps > 0 and num_search_steps == 0` case, the old code path through `postprocess()` would re-enable `input_quantizer` (if `is_input_quantized`), while the new unified path does not. However, this is consistent with the existing `num_cache_steps == 0` path which also doesn't re-enable it, and the export path handles this correctly since `get_quantization_format()` checks `hasattr(input_quantizer, "_pre_quant_scale")` without requiring `is_enabled`.

The change is small, well-scoped, has appropriate warnings, and the PR description is thorough.

Testing: Test plan has 1 unchecked item(s) out of 3. Finish or remove them before approving.

meenchen (commented):
Clean, well-scoped bug fix that unifies three inconsistent disabled-expert paths in `awq_lite` into a single `not module.awq_lite.is_enabled` branch. All disabled experts now consistently receive `max_calibrate` + neutral all-ones `pre_quant_scale`, ensuring uniform `nvfp4_awq` format across MoE expert groups and preventing the `AssertionError` in `preprocess_linear_fusion`.

Previous review comments addressed: test plan is confirmed finished per additional instructions. The CHANGELOG question from kevalmorabia97 is an editorial/project decision, not a blocking code issue.

Testing: Test plan has 1 unchecked item(s) out of 3. Finish or remove them before approving.

CHANGELOG.rst
cjluo-nv: The bug was present in 0.43 — the changelog entry should stay.

  The two ingredients of the bug both predate the 0.43 release (2026-04-16):

  ┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────┬────────────┬───────────────┐
  │                                                                             Component                                                                             │               Commit                │    Date    │    Author     │
  ├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┼────────────┼───────────────┤
  │ Export-side assertion preprocess_linear_fusion(...) "Modules have different quantization formats"                                                                 │ 73d6af785f                          │ 2025-01-29 │ Keval Morabia │
  ├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┼────────────┼───────────────┤
  │ MoE expert grouping that hits it (requantize_resmooth_fused_llm_layers calling preprocess_linear_fusion(modules, resmooth_only=True) at unified_export_hf.py:344) │ b8dbfc0931                          │ 2025-10-22 │ sugunav14     │
  ├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┼────────────┼───────────────┤
  │ Calibration-side asymmetric handling (num_cache_steps == 0 got ones-scale, but NaN-disabled / num_search_steps == 0 didn't)                                       │ ada1e26ba0 (#1142, prior NVBug fix) │ 2026-03-31 │ me            │
  └───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────┴────────────┴───────────────┘

  ada1e26ba0 only patched the num_cache_steps == 0 subpath; it left the NaN and "search-pass-missed" subpaths still emitting nvfp4 while siblings emitted nvfp4_awq. So on 0.43, any user running hf_ptq.py --qformat nvfp4_awq on an MoE model with calibration data that triggers either of those
  subpaths (e.g. cnn_dailymail at moderate calib_size on Qwen3-MoE — exactly what the original bug report reproduced) would hit the same export-side AssertionError.

  I'll leave the changelog entry under 0.44 Bug Fixes as-is.

meenchen (approved):
Re-review: All previous comments addressed. The test plan is now complete per the additional instructions. The changelog discussion was resolved by the author with a clear justification. The code change itself is small (+24/-11), correct, and well-scoped — it unifies three inconsistent disabled-expert paths into a single `not module.awq_lite.is_enabled` branch with appropriate warnings.

meenchen (approved):

PR #1363, opened 2026-04-28
Support Mixed precision & Static MSE in MCore;  Nemotron Super v3 NVFP4 recipe

jenchen13:
### What does this PR do?

Type of change: New recipe + Bug Fixes

 MCore and MSE fixes
  - support mixed precision export in MCore by detecting mixed precision layers in HF Quant Config
  - Restore static quantizer in MCore checkpoint restore as `NVFP4QTensor` (not TensorQuantizer which can call max calibrate. we want to skip max calibrate for static quantizer during restore) --> fixes bug during MCore export for MSE
  - Skip MSE calibration for any non-NVFP4 quantization format
  - Fix dynamic block quantizer detection when `block_sizes` is dict-backed.                                                                                                       
  -  Add a YAML quantization recipe that roughly mirrors Nemotron 3 Super NVFP4 `hf_quant_config.json`
  Export bug fixes     
  - copy .py files properly from original HF ckpt (for reasoning parser etc)                                                                 
                                                                                                                       
                                                                                                                                                                                   
## Super recipe

Mirrors the published nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4 hf_quant_config.json:
 - MoE routed experts (mixer.experts.<N>.{up,down}_proj):       NVFP4 W4A4 weight MSE, group_size 16
 - MoE shared experts (mixer.shared_experts.{up,down}_proj):    FP8 per-tensor
- Mamba mixer linears (mixer.{in,out}_proj):                   FP8 per-tensor
- KV cache:                                                    FP8
rest: not quantized

### Usage

```python
```

### Testing
TODO test in HF and MCore PTQ on Nemotron model

- Is this change backward compatible?: ✅ / ❌ / N/A <!--- If ❌, explain why. -->
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ / ❌ / N/A <!--- Mandatory -->
- Did you write any new necessary tests?: ✅ / ❌ / N/A <!--- Mandatory for new features or examples. -->
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ / ❌ / N/A <!--- Only for new features, API changes, critical bug fixes or backward incompatible changes. -->

### Additional Information
<!-- E.g. related issue. -->


## Summary by CodeRabbit

* **New Features**
  * Added three new quantization recipes for Nemotron-3-Super-120B with NVFP4 and FP8 calibration strategies.
  * Added configurable FP8 scale sweep stride parameter for fine-tuning quantization calibration.
  * Improved per-layer quantization metadata collection during model export.

* **Improvements**
  * Enhanced mixed-precision quantization and MoE expert handling across distributed processes.
  * Refined KV-cache and attention layer quantization configuration export.

* **Tests**
  * Updated quantization export verification tests.

### DIFF
--- a/modelopt/torch/export/unified_export_megatron.py
+++ b/modelopt/torch/export/unified_export_megatron.py
@@ -61,6 +61,7 @@ from .quant_utils import (
     get_weight_block_size,
     get_weight_scaling_factor,
     get_weight_scaling_factor_2,
+    process_layer_quant_config,
     to_quantized_weight,
 )
 
@@ -169,6 +170,7 @@ class GPTModelExporter:
         self.all_rules = self._populate_rule_book()
         self.rules = self.all_rules[self.arch]
         self.exclude_modules = []
+        self.layer_config_dict = {}
 
         if not hasattr(model, "_modelopt_state"):
             return
@@ -324,22 +326,32 @@ class GPTModelExporter:
                 print(f"Successfully loaded {len(mtp_state_dict)} MTP tensors")
 
         combined_exclude_modules = self._gather_exclude_modules()
+        combined_layer_config_dict = self._gather_layer_config_dict()
+        # kv_cache_dtype is only set on attention-owning ranks; writer rank may not be one.
+        gathered_kv_cache_dtype = self._gather_kv_cache_dtype()
 
         if is_last_stage_main_rank and quantization is not None:
+            if combined_layer_config_dict:
+                quantization_config = process_layer_quant_config(combined_layer_config_dict)
+                quantization_config["exclude_modules"] = combined_exclude_modules
+            else:
+                quantization_config = {
+                    "quant_algo": quantization,
+                    "exclude_modules": combined_exclude_modules,
+                }
+                if quantization == "NVFP4":  # update block size
+                    quantization_config["group_size"] = 16
+
+            if gathered_kv_cache_dtype is not None:
+                quantization_config["kv_cache_quant_algo"] = gathered_kv_cache_dtype
+
             self._hf_quant_config = {
                 "producer": {
                     "name": "modelopt",
                     "version": __version__,
                 },
-                "quantization": {
-                    "quant_algo": quantization,
-                    "exclude_modules": combined_exclude_modules,
-                },
+                "quantization": quantization_config,
             }
-            if quantization == "NVFP4":  # update block size
-                self._hf_quant_config["quantization"]["group_size"] = 16
-            if hasattr(self, "kv_cache_dtype"):
-                self._hf_quant_config["quantization"]["kv_cache_quant_algo"] = self.kv_cache_dtype
             with open(save_directory + "/hf_quant_config.json", "w") as f:
                 json.dump(self._hf_quant_config, f, indent=4)
 
@@ -359,10 +371,11 @@ class GPTModelExporter:
         # Newer versions of VLLM expect config.json with hf_quant_config
         config_json_file = save_directory + "/config.json"
         if self._hf_quant_config and os.path.exists(config_json_file):
-            converted_quant_config = convert_hf_quant_config_format(self._hf_quant_config)
             with open(config_json_file) as f:
                 config_dict = json.load(f)
-            config_dict["quantization_config"] = converted_quant_config
+            config_dict["quantization_config"] = convert_hf_quant_config_format(
+                self._hf_quant_config
+            )
             with open(config_json_file, "w") as f:
                 json.dump(config_dict, f, indent=4)
 
@@ -803,9 +816,7 @@ class GPTModelExporter:
         name_to_value = {}
         qformat: str = self._get_quantization_format(module)
         if qformat is None and "norm" not in prefix:
-            # Add exclude layers for hf_quant_config. Note that if the prefix is not an empty
-            # string then it usually ends with "." which needs to be removed.
-            self.exclude_modules.append(prefix.removesuffix("."))
+            self._record_excluded_module(prefix)
         block_size = get_weight_block_size(module)
 
         name_to_value = self._get_weight_bias(module, dtype, name_to_value)
@@ -850,6 +861,27 @@ class GPTModelExporter:
 
         return weight_scale, weight_scale_2
 
+    def _record_layer_quant_config(self, prefix: str, qformat: str | None, block_size: int):
+        """Record per-HF-layer quantization metadata for mixed precision exports."""
+        if qformat in (None, QUANTIZATION_NONE):
+            return
+
+        layer_name = prefix.removesuffix(".")
+        if "{" in layer_name or not layer_name:
+            return
+
+        self.layer_config_dict[layer_name + ".quantization"] = qformat
+        self.layer_config_dict[layer_name + ".awq_block_size"] = block_size
+
+    def _record_excluded_module(self, prefix: str):
+        """Record an unquantized HF module prefix for hf_quant_config."""
+        layer_name = prefix.removesuffix(".")
+        if "{" in layer_name or not layer_name:
+            return
+
+        if layer_name not in self.exclude_modules:
+            self.exclude_modules.append(layer_name)
+
     def _name_remapping(
         self,
         module: torch.nn.Module | torch.Tensor,
@@ -866,6 +898,7 @@ class GPTModelExporter:
             return
 
         name_to_value, qformat, block_size = self._get_quantized_state(module, dtype, prefix=prefix)
+        self._record_layer_quant_config(prefix, qformat, block_size)
 
         weight = name_to_value.pop("weight")
         weight_scale, weight_scale_2 = self._get_weight_scales(name_to_value, qformat)
@@ -906,6 +939,8 @@ class GPTModelExporter:
 
         gate_proj_prefix = prefix + gate_proj_name + "."
         up_proj_prefix = prefix + up_proj_name + "."
+        self._record_layer_quant_config(gate_proj_prefix, qformat, block_size)
+        self._record_layer_quant_config(up_proj_prefix, qformat, block_size)
 
         ffn_hidden_size = module.config.ffn_hidden_size
         gate_proj_weight = weight[:ffn_hidden_size, :]
@@ -986,6 +1021,7 @@ class GPTModelExporter:
 
         for expert_id in range(num_experts):
             expert_prefix = prefix.format(expert_id) + "."
+            self._record_layer_quant_config(expert_prefix, qformat, block_size)
             weight_key = f"weight{expert_id}"
 
             if weight_key not in state_dict:
@@ -1030,6 +1066,16 @@ class GPTModelExporter:
         q_proj_prefix = prefix + q_proj_name + "."
         k_proj_prefix = prefix + k_proj_name + "."
         v_proj_prefix = prefix + v_proj_name + "."
+        self._record_layer_quant_config(q_proj_prefix, qformat, block_size)
+        self._record_layer_quant_config(k_proj_prefix, qformat, block_size)
+        self._record_layer_quant_config(v_proj_prefix, qformat, block_size)
+        if qformat in (None, QUANTIZATION_NONE):
+            # Split fused linear_qkv exclude into per-HF-name q/k/v_proj entries.
+            fused_prefix = prefix.removesuffix(".")
+            self.exclude_modules = [m for m in self.exclude_modules if m != fused_prefix]
+            self._record_excluded_module(q_proj_prefix)
+            self._record_excluded_module(k_proj_prefix)
+            self._record_excluded_module(v_proj_prefix)
 
         config = module.config
         hidden_size = config.hidden_size
@@ -1179,6 +1225,7 @@ class GPTModelExporter:
             weight_scale_list.append(weight_scale)
             weight_scale_2_list.append(weight_scale_2)
             input_scale_list.append(input_scale)
+            self._record_layer_quant_config(prefix, qformat, block_size)
 
         merged_weight = torch.stack(weight_list, dim=0)
 
@@ -1247,6 +1294,7 @@ class GPTModelExporter:
             weight_scale_2_list.append(weight_scale_2)
             input_scale_list.append(input_scale)
             bias_list.append(bias)
+            self._record_layer_quant_config(prefix, qformat, block_size)
 
         merged_weight = torch.stack(weight_list, dim=0)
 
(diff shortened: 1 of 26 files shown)


modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4-fp8-sweep-stride4.yaml
```diff
@@ -0,0 +1,135 @@
...
+# Mirrors the published nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4 hf_quant_config.json:
+#   - MoE routed experts:     NVFP4 W4A4 weight MSE, group_size 16
+#     HF names:               mixer.experts.<N>.{up,down}_proj
+#     Megatron-Core names:    mlp.experts.local_experts.<N>.linear_fc{1,2}
+#   - MoE shared experts:     FP8 per-tensor
+#     HF names:               mixer.shared_experts.{up,down}_proj
+#     Megatron-Core names:    mlp.shared_experts.linear_fc{1,2}
+#   - Mamba mixer linears (mixer.{in,out}_proj):                   FP8 per-tensor
+#   - KV cache:                                                    FP8
+#   - Attention linears ({q,k,v}_proj):                            BF16 (not quantized)
```
jinhangchoi, line 26: Can we double check attention out linear? IIRC, attention o_proj should be FP8.

jenchen13: responded in slack, only 2/9 attention layers had `o_proj` FP8 in final Super NVFP4 ckpt, but we can always add it later to test if accuracy degradation is minimal

jenchen13:
@coderabbitai full review

jenchen13:
@claude review

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -803,9 +814,7 @@ def _get_quantized_state(
         name_to_value = {}
         qformat: str = self._get_quantization_format(module)
         if qformat is None and "norm" not in prefix:
-            # Add exclude layers for hf_quant_config. Note that if the prefix is not an empty
-            # string then it usually ends with "." which needs to be removed.
-            self.exclude_modules.append(prefix.removesuffix("."))
+            self._record_excluded_module(prefix)
         block_size = get_weight_block_size(module)
```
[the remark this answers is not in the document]
jenchen13: QUANTIZATION_NONE = None (defined in `export/model_config.py` so this will still work

jenchen13:
@coderabbit full review

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -324,22 +326,32 @@ def save_pretrained(
...
-                    "exclude_modules": combined_exclude_modules,
-                },
+                "quantization": quantization_config,
             }
-            if quantization == "NVFP4":  # update block size
-                self._hf_quant_config["quantization"]["group_size"] = 16
-            if hasattr(self, "kv_cache_dtype"):
-                self._hf_quant_config["quantization"]["kv_cache_quant_algo"] = self.kv_cache_dtype
+            # Use one serving-facing config for both hf_quant_config.json and config.json.
+            self._hf_quant_config = convert_hf_quant_config_format(raw_hf_quant_config)
```
shengliangxu: do we change the format of hf_quant_config.json by this change?

modelopt/torch/quantization/qtensor/nvfp4_tensor.py
```diff
@@ -55,7 +55,16 @@ def get_e2m1_bounds(cls, device):
     @classmethod
     def _is_static_quantizer(cls, weight_quantizer) -> bool:
         """Check if the weight quantizer is a static NVFP4 quantizer with pre-computed amax."""
-        return hasattr(weight_quantizer, "global_amax") and weight_quantizer.global_amax is not None
+        global_amax = cls._get_static_global_amax(weight_quantizer)
+        return global_amax is not None
+
+    @classmethod
+    def _get_static_global_amax(cls, weight_quantizer):
```
Fridah-nv, line 62: Thanks for catching the bug! I'd like to understand why in mcore path we use `_global_amax` instead of `global_amax`, is there any general guidance for when to use private attribute name and when to use public ones? I think we have similar issue with `amax` vs `_amax`

Fridah-nv (approved):
Changes in MSE flow LGTM

modelopt/torch/quantization/qtensor/nvfp4_tensor.py
jenchen13: during checkpoint resume in MCore it initializes the quantizers as generic TensorQuantizer and those have `_global_amax` parameter (instead of NVFP4QTensor which has `global_amax`)

modelopt/torch/export/unified_export_megatron.py
jenchen13: yes to align it with the quant config inside config.json. previously `hf_quant_config` was still using the older json format whereas the quant config inside config.json was being converted to a newer format

modelopt/torch/quantization/qtensor/nvfp4_tensor.py
jenchen13: @Fridah-nv I have changed MCore static quantizer restore to restore as NVFP4QTensor in mtq.plugins.custom

modelopt/torch/quantization/model_calib.py
realAsma, line 1: @jenchen13  could you please take another pass at this file?

Can you please see if there are any scope for simplification? The comments are super verbose

realAsma:
> _Bot comment._

Review findings from a public pass on `41b440c`:

**Blocker: sparse experts that receive no calibration tokens can silently miss the MSE/static path.** In `mse_calibrate()`, `NVFP4MSECalibrator` replacement and static NVFP4 promotion only happen inside the `hasattr(module, "_amax")` branch (`modelopt/torch/quantization/model_calib.py:576-622`). A routed expert weight quantizer that was not exercised during the initial calibration forward pass has no `_amax`, so it is still collected for the one-at-a-time weight pass but keeps its original max calibrator. The later repair block only recomputes static per-block/global state for quantizers already promoted to `NVFP4StaticQuantizer` (`model_calib.py:686-804`). That means the recipe can silently mix true MSE-calibrated static expert weights with max/dynamic fallback expert weights depending on calibration routing. Please promote/setup every enabled static NVFP4 weight quantizer from its weight tensor before the MSE weight pass, or fail loudly when the static MSE path cannot be applied.

**Test gap:** the PR adds mixed per-layer MCore export behavior, but the MCore export test matrix still covers uniform quantization configs only. Please add a mixed case matching the new recipe shape, for example NVFP4 routed experts plus FP8 shared/mamba/KV paths, and assert the emitted `config.json`/`hf_quant_config.json` compressed-tensors groups, ignored modules, and per-layer targets.

realAsma:
> _Bot comment._

Minimal-dev follow-up on `modelopt/torch/quantization/model_calib.py`:

**Design / scope:** the `mse_calibrate()` changes add a large post-hoc repair pass for missing/meta/non-finite weight quantizer state (`model_calib.py:686-804`). Can we handle this at the weight-MSE setup boundary instead? For weight quantizers, if the initial max-calibration forward did not populate `_amax`, the one-at-a-time weight loop already has the actual weight tensor. Initializing per-block `_amax` / `global_amax` from that weight before constructing the MSE calibrator would keep the intended weight-MSE behavior for unvisited experts and remove most of the late repair/fallback block.

**Keep the PR surgical:** the new batched DP/EP/TP amax sync helpers rewrite a broad part of generic `max_calibrate()` (`model_calib.py:157-325`, `377-509`). The recipe/export change seems to need recursive `SequentialQuantizer` traversal plus correct static/dynamic block predicates, not a distributed-sync batching rewrite. Could we keep the existing recursive sync flow here and make only the narrow predicate/traversal change, then split batching/perf work into a separate PR with distributed sync-focused coverage?

**Avoid duplicate state policy:** invalid amax handling now lives in both `mse_calibrate()` and export (`model_calib.py:686-804`, `modelopt/torch/export/quant_utils.py:273-340`). Please move the “missing/meta/non-finite amax” validation and “recompute static NVFP4 state from weight” operation into one shared quantizer/state helper, and have calibration/export call that. Otherwise these paths can drift on what counts as invalid and how buffers are registered.

**Comment/style cleanup:** the long incident-style block comment at `model_calib.py:686-708` is useful context, but it is too much inline history for production code. A short invariant comment is enough here; the detailed failure narrative belongs in the regression test or PR discussion.

**Name the predicate once:** the static NVFP4 condition is repeated with private fields in multiple places (`model_calib.py:582-587`, `model_calib.py:1008-1014`, `modelopt/torch/quantization/utils/core_utils.py:960-965`). Since this PR already adds `is_static_block_quant` / `is_dynamic_block_quant`, please add one `is_static_nvfp4_block_quant`-style helper/property and use it at the call sites.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -324,22 +326,32 @@ def save_pretrained(
...
+                quantization_config["exclude_modules"] = combined_exclude_modules
+            else:
+                quantization_config = {
+                    "quant_algo": quantization,
+                    "exclude_modules": combined_exclude_modules,
+                }
+                if quantization == "NVFP4":  # update block size
+                    quantization_config["group_size"] = 16
+
+            if hasattr(self, "kv_cache_dtype"):
```
jinhangchoi: If PP>1, and last_stage_main_rank doesn't have attention layer, then `kv_cache_quant_algo` will be lost in JSON config.

modelopt/torch/quantization/model_calib.py
jenchen13: I reverted a lot of changes, previously they were claude trying to optimize the syncs by batching them before collective, but that only saved like 5 seconds

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -190,7 +190,12 @@ def quant_module_set_extra_state(self, state: Any):
     if quantizer_state is not None:
         for name, module in self.named_modules():
             if isinstance(module, TensorQuantizer):
-                module.set_from_modelopt_state(quantizer_state[name], properties_only=False)
+                quantizer_substate = quantizer_state[name]
```
jenchen13, line 224: TODO create a helper in mtq.conversion.py and use that

modelopt/torch/quantization/model_calib.py
```diff
@@ -210,26 +210,88 @@ def _has_expert_parallelism(module: nn.Module) -> bool:
...
-    for group in [
-        parallel_state.data_parallel_group,
-        parallel_state.expert_model_parallel_group,
-        parallel_state.tensor_parallel_group,
-    ]:
-        if not group.is_initialized():
+    yield quantizer
+
+
+def _check_static_block_tp_supported(model: nn.Module) -> None:
```
jenchen13: TODO move to megatron/plugins, make shorter

modelopt/torch/quantization/model_calib.py
```diff
@@ -210,26 +210,88 @@ def _has_expert_parallelism(module: nn.Module) -> bool:
...
+    a different set of experts; their amaxes describe different shards. Input
+    quantizers are replicated across EP and should be synced. Set
+    ``sync_expert_weight_amax=True`` for TEGroupedMLP where all routed experts
+    share one weight quantizer.
+    """
+    if not quantizer.is_block_quant:
+        return True
+    if "input_quantizer" in child_name:
+        return True
+    return sync_expert_weight_amax and "weight_quantizer" in child_name
```
jenchen13: TODO reduce docstring, check if `sync_expert_weight_amax` is correct
check with @realAsma

modelopt/torch/quantization/model_calib.py
```diff
@@ -449,13 +487,23 @@ def mse_calibrate(
...
+                if not (
+                    getattr(module, "_num_bits", None) == (2, 1)
+                    and bs is not None
+                    and bs.get("scale_bits") == (4, 3)
+                ):
+                    fmt = f"num_bits={module._num_bits} block_sizes={bs}"
+                    skipped_non_nvfp4[fmt] = skipped_non_nvfp4.get(fmt, 0) + 1
+                    continue
                 initial_amax = module._amax.clone().detach()
                 is_nvfp4_static = module.is_nvfp4_static
```
realAsma, line 506:
```suggestion
if not is_nvfp4_static:
     ....
```

modelopt/torch/quantization/model_calib.py
```diff
@@ -449,13 +487,23 @@ def mse_calibrate(
...
             if module._calibrator is not None and not module._dynamic and hasattr(module, "_amax"):
+                bs = getattr(module, "block_sizes", None)
+                if not (
+                    getattr(module, "_num_bits", None) == (2, 1)
+                    and bs is not None
+                    and bs.get("scale_bits") == (4, 3)
+                ):
+                    fmt = f"num_bits={module._num_bits} block_sizes={bs}"
+                    skipped_non_nvfp4[fmt] = skipped_non_nvfp4.get(fmt, 0) + 1
+                    continue
```
realAsma, line 503:
```suggestion

```

can we is `is_nvfp4_static` flag to avoid this check redundancy?

modelopt/torch/quantization/model_calib.py
```diff
@@ -210,26 +218,54 @@ def _has_expert_parallelism(module: nn.Module) -> bool:
...
-        parallel_state.tensor_parallel_group,
-    ]:
-        if not group.is_initialized():
+    yield quantizer
+
+
+def _check_moe_calibration_complete(quantizer, parallel_state):
+    """Raise error if MoE calibration is incomplete across distributed MoE ranks."""
+    for leaf_quantizer in _iter_leaf_quantizers(quantizer):
+        if leaf_quantizer.is_block_quant:
```
realAsma: This will skip checks for activation quantizers as well right?

realAsma: why do we need this change?

jenchen13: sorry I've removed that block quant skip logic

modelopt/torch/quantization/algorithms.py
realAsma, line 280: @jenchen13  we need EP support in `get_score` and `get_cost`

realAsma: change 1 for EP w/ AutoQuantize

realAsma: change 2 for EP w/ AutoQuantize

modelopt/torch/quantization/plugins/megatron.py
jinhangchoi, line 402: If TP>1, but ETP=1, this gets problem as each ETP expects to have replicated amax, whereas TP expects sharding. We should pass megatron's expert TP/DP groups here.
```suggestion
        tp_pg = self.parallel_state.tensor_parallel_group
        dp_pg = self.parallel_state.data_parallel_group
        tp_group = tp_pg.group if tp_pg.is_initialized() else None
        dp_cp_group = dp_pg.group if dp_pg.is_initialized() else None
        sharded_state_dict.update(
            **make_sharded_tensors_for_checkpoint(
                quantizer_state_dict, prefix, sharded_axis_dict, sharded_offsets
                quantizer_state_dict,
                prefix,
                sharded_axis_dict,
                sharded_offsets,
                tp_group=tp_group,
                dp_cp_group=dp_cp_group,
            )
        )
```

jenchen13:
Moving this PR to https://github.com/NVIDIA/Model-Optimizer/pull/1521

PR #1378, opened 2026-04-30
Use trtexec_safe on safety platforms when using remoteAutoTuning

dthienan-nv:
Previously there was a bug when using TrtExecBenchmark with the safety runtime and remoteAutoTuning that resulted in invalid latency measurements.  This was due to using the local trtexec instead of the remote trtexec_safe.
This PR fixes this issue by using scp to push the resulting engine to the remote device and then remotely calling trtexec_safe to generate latency measurements.
This PR changes the behavior when specifying remote auto tuning such that an incorrect configuration will no longer gracefully fallback but rather fail hard and fast so that the user may correct the configuration.

Before:
```
[modelopt][onnx] - WARNING - Could not parse median latency from trtexec output
[modelopt][onnx] - WARNING - Benchmark failed: /tmp/tmp9s0yckb3/baseline.onnx
[modelopt][onnx] - INFO - Baseline latency: inf ms
[modelopt][onnx] - INFO - Baseline: inf ms
```

After:
```
[modelopt][onnx] - INFO - TrtExec benchmark (median): 3.16 ms
[modelopt][onnx] - INFO - Baseline latency: 3.165 ms
[modelopt][onnx] - INFO - Baseline: 3.16 ms
```
## Security Exception
Requesting an exception since we need to call subprocesses to push data to a remote device for remote profiling.

## Summary by CodeRabbit

* **Improvements**
  * Distinct latency measurement for safe vs standard benchmark modes for more accurate results.
  * Strict parsing and validation of remote autotuning configuration as URL-style input.
  * Automatic transfer of built artifacts and SSH-based remote benchmarking with a fallback execution path.
  * Timeouts for remote operations and clearer error handling that surfaces missing dependencies and yields infinite latency on failures.


modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -204,9 +211,56 @@ def __init__(
...
+            # parsed.username, parsed.password, parsed.hostname, parsed.port, parsed.query
+            self.remote_user = parsed.username
+            self.remote_password = parsed.password
+            self.remote_ip = parsed.hostname
+            self.remote_port = parsed.port
+            # Parse query options into a dict
+            self.remote_options = {
+                k: v[0] if len(v) == 1 else v for k, v in parse_qs(parsed.query).items()
+            }
+            self.remote_bin_path = os.path.dirname(str(self.remote_options["remote_exec_path"]))
```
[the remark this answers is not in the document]
dthienan-nv: Fixed

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -293,10 +352,34 @@ def run(
...
+                )
+                trt_path = f"{os.path.join(self.remote_bin_path, 'trtexec_safe')}"
+                trtexec_safe_cmd = [
+                    "ssh",
+                    "-p",
+                    f"{self.remote_port}",
+                    f"{self.remote_user}:{self.remote_password}@{self.remote_ip}",
+                    f"{ld_path} {trt_path} --loadEngine={self.remote_engine_path}",
+                ]
+                result = subprocess.run(trtexec_safe_cmd, capture_output=True, text=True)  # nosec B603
```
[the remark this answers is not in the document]
dthienan-nv: Partially addressed.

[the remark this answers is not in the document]
dthienan-nv: Partially addressed.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -292,10 +358,59 @@ def run(
...
+                        f"{self.remote_port}",
+                        f"{self.remote_user}@{self.remote_ip}",
+                        f"{ld_path} {shlex.quote(trt_path)} --safe --loadEngine={shlex.quote(self.remote_engine_path)}",
+                    ]
+                    trtexec_safe_cmd = ssh_pass + trtexec_safe_cmd
 
-            if not (match := re.search(self.latency_pattern, result.stdout, re.IGNORECASE)):
-                self.logger.warning("Could not parse median latency from trtexec output")
-                self.logger.debug(f"trtexec stdout:\n{result.stdout}")
+                    result = subprocess.run(trtexec_safe_cmd, capture_output=True, text=True)  # nosec B603
```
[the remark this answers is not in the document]
dthienan-nv: Done

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -204,9 +212,65 @@ def __init__(
...
+            if not remote_config_str.startswith("ssh://"):
+                raise ValueError("Only 'ssh://' remote autotuning config URLs are supported")
+            parsed = urlparse(remote_config_str)
+            # parsed.username, parsed.password, parsed.hostname, parsed.port, parsed.query
+            self.remote_user = parsed.username
+            self.remote_password = parsed.password
+            self.remote_ip = parsed.hostname
+            self.remote_port = parsed.port
+            if self.remote_port is None:
+                self.remote_port = 22
```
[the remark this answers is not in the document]
dthienan-nv: Addressed via assertion

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -228,6 +293,7 @@ def __init__(
                 trtexec_args = [
                     arg for arg in trtexec_args if "--remoteAutoTuningConfig" not in arg
                 ]
+        self.is_safe = "--safe" in trtexec_args
```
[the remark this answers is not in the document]
dthienan-nv: Addressed via assertion

galagam:
@willg-nv @ajrasane Can you take a look please?

ajrasane (changes requested):
## Code Review Findings

### Critical

**1. Existing mocked tests will fail (regression).**
`tests/gpu/onnx/quantization/autotune/test_benchmark.py:196` and `:214` mock `subprocess.run` to return `"[I] Latency: min = ... median = X ms"` and assert `bench.run(...) == X`. The PR replaces `latency_pattern` with `r"\[I\]\s+GPU Compute Time:.*?median\s*=\s*([\d.]+)\s*ms"`, which doesn't match the `"[I] Latency:"` mocks. Running the tests against this PR's `benchmark.py` confirms it:
```
FAILED tests/gpu/onnx/quantization/autotune/test_benchmark.py::test_trtexec_run_returns_parsed_latency
FAILED tests/gpu/onnx/quantization/autotune/test_benchmark.py::test_trtexec_run_accepts_bytes_input
```
Fix: update the mock stdout in those two tests to use the new `[I] GPU Compute Time: min = ..., max = ..., median = X ms` format.

**2. No tests cover any of the new logic.**
133 lines of new code — URL parsing, query-param validation, port/user/host fallbacks, scp/ssh invocation, primary-vs-fallback remote command selection, two distinct latency regexes — and zero tests added or updated. At minimum, please add unit tests with `subprocess.run` mocked for:
- `--remoteAutoTuningConfig` parsing happy-path (both `=URL` and ` URL` arg forms).
- Each `ValueError` branch in the parser (missing value, non-`ssh://`, missing user/host, missing query params, duplicate config arg).
- The `safe_pattern` regex against a realistic `Average over N runs - GPU latency: X ms` line.
- The `trtexec_safe` → `trtexec --safe` fallback path (returncode-based selection of `latency_pattern`).

### Important

**3. `# nosec B603` comments remain unresolved per `SECURITY.md`.**
Per project security guidelines, `# nosec` is forbidden without `@NVIDIA/modelopt-setup-codeowners` approval and explicit PR-description justification. CodeRabbit raised this twice; the author replied "Partially addressed" but the four `# nosec B603` comments at the new `subprocess.run` sites are still in `benchmark.py` (all on the new ssh/scp calls). Either remove them and seek codeowner approval with justification in the PR body, or align with whatever approved pattern the project uses for subprocess invocations.

**4. `FileNotFoundError` handler misattributes the error.**
The outer `except FileNotFoundError` in `run()` catches errors from any of the new subprocess calls (`scp`, `ssh`, `sshpass`) and logs:
```
trtexec binary not found: {self.trtexec_path}
Please ensure TensorRT is installed and trtexec path is correct
```
A missing `sshpass` or `scp` (likely on minimal containers) will produce that misleading message. Either differentiate by logging the actual `e.filename`, or scope the existing handler narrowly to the local trtexec call and add a separate handler for the remote path.

**5. `self.is_safe` is set in two places; the inner assignment is dead.**
`self.is_safe = True` inside the `if "--safe" not in trtexec_args:` block is followed by an unconditional `self.is_safe = "--safe" in trtexec_args` after the block (and the inner `self.trtexec_args.append("--safe")` mutates the same list `trtexec_args` aliases, so the outer assignment ends up True anyway). The inner one isn't a bug but is confusing — please drop it so there's a single source of truth.

**6. Multi-valued query params silently corrupt remote paths.**
```python
self.remote_options = {
    k: v[0] if len(v) == 1 else v for k, v in parse_qs(parsed.query).items()
}
```
If `remote_exec_path` appears more than once in the URL (`?remote_exec_path=a&remote_exec_path=b`), `self.remote_options["remote_exec_path"]` becomes a list, and `os.path.dirname(str([...]))` produces nonsense like `''`. Either reject duplicates (raise `ValueError`), or always take `v[-1]`/`v[0]`.

**7. Fixed remote engine filename → no concurrency safety.**
`self.remote_engine_path = "trtexec_benchmark_model.trt"` is hardcoded and lands in the remote user's home dir. Two autotune runs against the same target will clobber each other's engines mid-benchmark. Suggest deriving from a `tempfile`-style suffix (PID + uuid) or the local `self.engine_path` basename.

**8. `--useCudaGraph` is silently injected into the remote command.**
The remote `trtexec_safe` and `trtexec --safe` invocations always append `--useCudaGraph`, regardless of what the user passed in `trtexec_args`. This was added in commit `110f3d7f` ("more stable when using cudaGraphs"). Worth either documenting the override on `--remoteAutoTuningConfig` or making it conditional on user opt-in — silently flipping a kernel-launch mode for latency measurement is surprising.

### Suggestions

**9. Module-level regex constants should be private.**
`safe_pattern`/`std_pattern` are exported by virtue of being module-level. They're internal implementation. `_SAFE_LATENCY_PATTERN` / `_STD_LATENCY_PATTERN` would scope and signal intent.

**10. Stale comment on warning text.**
The `except ImportError` warning was trimmed in commit `2512076` after CodeRabbit flagged it — please double-check the final string still makes sense in isolation; it now reads as a bare statement rather than an actionable warning.

**11. Document that `sshpass` is now an implicit runtime dependency.**
If `--remoteAutoTuningConfig` includes a password, `sshpass` must be installed on the host running ModelOpt. That's not in any docstring or README; users will discover it via the misleading "trtexec binary not found" message above.

**12. Consider redacting credentials from remote-config error messages.**
The `ValueError` raises that include `arg` or `config_arg_value` substrings can echo a `ssh://user:pass@host` back at the user. Most are caught early enough that the password isn't yet parsed, but `f"Malformed --remoteAutoTuningConfig argument: {arg}"` could echo a password. Strip the password before formatting.

### Positive
- Good iteration on review feedback: query-param validation, `shlex.quote` on every interpolated remote-side token, explicit `ValueError` on missing user/host, `urlparse`-based URL parsing rather than regex.
- Switching from the rolled-up `[I] Latency:` line to `[I] GPU Compute Time:` is the right call — `Latency:` includes H2D/D2H copies, which is noisy for autotune comparisons.
- `ImportError` is now re-raised (was previously swallowed and silently stripped from args), so a TRT < 10.15 environment fails fast.
- Hard fail on multiple `--remoteAutoTuningConfig` args closes off an ambiguous configuration.

## Test Coverage Assessment
- Unit-test coverage for new code: **0**.
- Existing tests: **2 mocked tests will regress** (see Critical #1).
- No GPU-CI run posted on the PR (vetting still pending).
- No regression test for the original bug — a mock that returns the `[I] GPU Compute Time:` format (or simulates the remote SSH path) would lock the fix in place.

## Final Verdict
**Request Changes.**

Blocking items: the test regression (Critical #1), the missing tests for new functionality (Critical #2), and the unresolved `# nosec` situation (Important #3). The remote-execution logic itself is reasonable, and the iteration history shows the author has been responsive to security feedback — but the PR can't land in its current state because it breaks pre-existing tests.

dthienan-nv:
I've addressed all comments.  The following remain unresolved:
```
3. # nosec B603 comments remain unresolved per SECURITY.md. -> This still requires an exception.

7. Fixed remote engine filename → no concurrency safety. -> Added a note, should not use a remote device's GPU concurrently anyways since that would result in inaccurate timing.
```

ajrasane (changes requested):
## Review

Fixes a real bug (safety-runtime + `remoteAutoTuning` returning `inf` latencies) and the refactor into testable helpers (`_parse_remote_autotuning_url`, `_RemoteAutotuningConfig`, `_redact_url_password`, etc.) is clean. New unit tests in `test_trtexec_benchmark.py` (728 lines) and `test_cli_pipeline.py` (211 lines) close the prior "no tests for new logic" gap. A few items below are blocking; the rest are quick-fix follow-ups.

### Critical

1. **All three remote `subprocess.run` calls have no `timeout=`** — `modelopt/onnx/quantization/autotune/benchmark.py` lines 430 (scp), 453 (trtexec_safe ssh), and 471 (fallback ssh). A hung SSH (network blip, host-key prompt because `StrictHostKeyChecking` isn't set, keyboard-interactive auth fallback) will block the entire autotuning workflow forever. CodeRabbit flagged this and marked it resolved against commit 28e4e8e, but it is not in HEAD — please re-add `timeout=` on all three calls and a `subprocess.TimeoutExpired` handler that logs and returns `float("inf")`.

2. **`remote_user` / `remote_ip` not validated against leading `-` (argv smuggling).** The scp destination is built as `f"{self.remote_user}@{self.remote_ip}:..."` and `remote_user` / `remote_ip` come from `urlparse`. `urlparse` does not reject usernames or hostnames that look like options. A URL like `ssh://-oProxyCommand=evil@host/...` lets a crafted username be reinterpreted as an `ssh`/`scp` flag (CVE-2017-1000117 class). Please reject `remote_user`/`remote_ip` that start with `-` in `_parse_remote_autotuning_url`.

### Important

3. **Fallback path uses `_STD_PATTERN` but ran `trtexec --safe`** (line 475). The `--safe` invocation of `trtexec` typically emits the `Average over N runs - GPU latency:` line (i.e., `_SAFE_PATTERN`), not `GPU Compute Time: ... median = ...`. If the safety-proxy target emits the SAFE format, parsing silently returns `inf`. Either try both patterns in the fallback, or switch the fallback to `_SAFE_PATTERN`.

4. **`stderr` is dropped in the consolidated failure log** (lines 478–480): `f\"Failed to run trtexec_safe or trtexec with '--safe'\n {result.stdout}\"`. SSH/sshpass auth failures, host-key prompts, and connection-refused errors are emitted on stderr. Include `result.stderr` so users can diagnose connection failures.

5. **`--useCudaGraph` is silently injected into the remote command** (lines 445, 467) regardless of the user's `trtexec_args`. Document this behavior, and consider gating it (only add if not already present).

6. **No cleanup of the remote engine file.** `remote_engine_path = \"trtexec_benchmark_model.trt\"` is scp'd every run and never removed; repeated sessions accumulate stale files on the target. Add an `ssh ... rm -f <shlex.quote(remote_engine_path)>` in a `finally`.

7. **Hardcoded `remote_engine_path`.** Concurrency is disallowed by design, but unrelated users on the same target box can still clash. Suggest `f\"trtexec_benchmark_model_{os.getpid()}_{uuid.uuid4().hex}.trt\"`.

8. **`self.remote_options` is populated but never read** in `run()`. Either remove it or document why it's kept on the instance.

9. **Backward compatibility of `_STD_PATTERN`.** It now matches `[I] GPU Compute Time: ... median = X ms` instead of `[I] Latency: ... median = X ms`. The remote path is gated at `>=10.15`, but the local std path has no version guard — older `trtexec` may still emit `Latency:`. Either try both patterns or pin the minimum TensorRT version.

### Suggestions

- Add `-oStrictHostKeyChecking=accept-new` to the `ssh`/`scp` invocations so first-connect doesn't hang (becomes critical without timeouts).
- `_ensure_remote_autotuning_flags` appends `--safe` if missing, which means `self.is_safe` ends up coupled to "remote autotuning is configured". Worth a docstring note.
- Docstring typo in `docs/source/guides/9_autotune.rst`: "Only once instance" → "Only one instance".
- `["-P", str(self.remote_port)]` would be more consistent with the `ssh` invocation right below it than `f\"-P{self.remote_port}\"`.

### CI

- `unit-pr-required-check` and `check-dco / wait` are failing on HEAD — must be green before merge.

### Verdict

Request Changes. Primary blockers: missing `timeout=` on all three remote `subprocess.run` calls, and argv-smuggling validation on `remote_user`/`remote_ip`. The fallback latency-pattern mismatch is also worth fixing in the same pass.

dthienan-nv:
All three remote subprocess.run calls have no timeout= 
 - This has been addressed for the remote calls.

remote_user / remote_ip not validated against leading - (argv smuggling).
 - Addressed

Fallback path uses _STD_PATTERN but ran trtexec --safe
 - The pattern is dependent on the executable, not the runtime.

stderr is dropped in the consolidated failure log
 - Addressed

--useCudaGraph is silently injected into the remote command
 - This is documented in 9_autotune.rst

No cleanup of the remote engine file.
 - Addressed

Hardcoded remote_engine_path
 - Concurrency shouldn't be allowed across users either.

self.remote_options is populated but never read
 - Removed

Backward compatibility of _STD_PATTERN
 - False positive, GPU compute time is available in older builds.

Docstring typo ...
- Addressed

Add -oStrictHostKeyChecking=accept-new
- Addressed

ajrasane (changes requested):
## Changes Requested

### Critical (blocking)

1. **CI is red — three of this PR's own unit tests fail on HEAD.** The new `cleanup_remote_engine()` `finally` block in `modelopt/onnx/quantization/autotune/benchmark.py` adds a 4th `subprocess.run` (`ssh … rm -f <remote_engine_path>`) that the new tests don't account for. From the `linux` job log:
   - `test_remote_run_scp_then_ssh_trtexec_safe` → `assert run_mock.call_count == 3` fails with `assert 4 == 3`.
   - `test_remote_run_uses_sshpass_when_password_set` → `_, scp_cmd, ssh_cmd = (c.args[0] for c in run_mock.call_args_list)` raises `ValueError: too many values to unpack (expected 3)`.
   - `test_remote_run_falls_back_to_trtexec_safe_flag` → `assert 'trtexec --safe' in 'rm -f trtexec_benchmark_model.trt'` (the test inspects `call_args_list[-1]`, which is now the cleanup call).

   Fix: extend each test's `side_effect` list with one extra `_make_proc()` for the cleanup, replace `[-1]` with an explicit call index, and unpack 4/5 calls instead of 3.

2. **`# nosec B603` markers don't meet SECURITY.md's exception bar.** Four `# nosec B603` suppressions remain on `subprocess.run` sites in `benchmark.py` (lines 513, 539, 563, 583), plus `# nosec B404` on the import. SECURITY.md requires both an inline `why-necessary / why-safe` comment AND a PR-description justification with `@NVIDIA/modelopt-setup-codeowners` review. Neither is in place. The underlying code is genuinely safe (list-form subprocess, no `shell=True`, `shlex.quote` on every interpolated value, argv-smuggling guards on `user`/`host`) — so the simplest path is to **drop the `# nosec` markers entirely and confirm Bandit is green**. If they must stay, add an inline reason at each site and request `@NVIDIA/modelopt-setup-codeowners` review with an explicit "Security exception" section in the PR description. @kevalmorabia97, please review

### Important

3. **`sshpass -p <password>` leaks the password to other local users.** Lines 496–500 still build `["sshpass", "-p", self.remote_password]`. Any local user can read the password via `/proc/<pid>/cmdline` or `ps -ef` while the subprocess is alive. Switch to `sshpass -e` and pass the password via `subprocess.run(env={**os.environ, "SSHPASS": self.remote_password, …})` so it stays out of argv.

4. **`-oStrictHostKeyChecking=accept-new` only on `scp`, not on the three `ssh` invocations.** The first-connect host-key prompt will block until `network_timeout_seconds` (default 300s) on the cleanup ssh (line ~526), primary trtexec_safe ssh (line ~549), and fallback ssh (line ~569). Add the flag to all three.

5. **`remote_engine_path` is a hardcoded literal and collides across unrelated users on the same target.** `benchmark.py:400` is `self.remote_engine_path = "trtexec_benchmark_model.trt"`. Two users sharing a safety box will overwrite each other's engines mid-benchmark. Derive a unique name: `f"trtexec_benchmark_model_{os.getpid()}_{uuid.uuid4().hex}.trt"`. The cleanup contextmanager already handles removal in `finally`, so no orphan accumulation.

6. **`--useCudaGraph` injection has zero test coverage.** Lines 553 (primary) and 573 (fallback) unconditionally splice `--useCudaGraph` into the remote command, but no test in `test_trtexec_benchmark.py` mentions it. Add a test asserting the flag appears exactly once in both primary and fallback `ssh` argv, and at least one test that a user-supplied `--useCudaGraph` doesn't get duplicated.

7. **Custom remote port is parsed but flow-through is never asserted.** `_REMOTE_URL` uses port `2222`, stored on the instance, but no test asserts `["scp", "-P", "2222", …]` or `["ssh", "-p", "2222", …]` end up in argv. The current `"scp" in scp_cmd`-style membership checks would also pass if the port silently fell back to 22. Tighten one of the existing tests.

8. **`cleanup_remote_engine` has no direct test.** Once Critical #1 is fixed, add explicit assertions that the cleanup `ssh … rm -f <shlex.quote(remote_engine_path)>` argv is produced on (a) success, (b) fallback success, (c) both paths failing, and (d) a raise inside the cleanup is swallowed and only logged at warning level.

9. **`_RemoteAutotuningConfig.options` is populated but unused.** `_parse_remote_autotuning_url` populates `options={"remote_exec_path": …, "remote_lib_path": …}` redundantly with `bin_path`/`lib_path`. Either drop the field (and the corresponding branch in the test for `options`) or document why it's kept on the dataclass.

### Suggestions

10. **Empty-string username is not validated.** `urlparse("ssh://:password@host?…")` returns `username=""`, which passes the `is None` check in `_parse_remote_autotuning_url`. Add `or not parsed.username.strip()` to the guard. Same for `hostname`.

11. **Add a one-line comment next to `self.is_safe = "--safe" in self.trtexec_args`.** This is only correct because `_ensure_remote_autotuning_flags` has already mutated `self.trtexec_args` to include `--safe` in the remote path. A future reader could easily "simplify" the ordering and break it.

12. **Validate port range in `_parse_remote_autotuning_url`.** `urlparse` raises on negatives but values >65535 silently pass. Add `if not 1 <= port <= 65535: raise ValueError(...)`.

13. **Docs typo in `docs/source/guides/9_autotune.rst`:** "Only once instance" → "Only one instance".

### Verdict

Request Changes. Items 1 (CI) and 2 (SECURITY.md `# nosec` policy) are blocking. Items 3–9 are worth addressing in the same revision; the underlying bug fix and refactor are sound — this is hardening, not redesign.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -287,19 +491,120 @@ def run(
...
+                    "-P",
+                    str(self.remote_port),
+                    "-oStrictHostKeyChecking=accept-new",
+                    self.engine_path,
+                    f"{self.remote_user}@{self.remote_ip}:{shlex.quote(self.remote_engine_path)}",
+                ]
+                scp_cmd = ssh_pass + scp_cmd
+                result = subprocess.run(
+                    scp_cmd, capture_output=True, text=True, timeout=self.network_timeout_seconds
+                )  # nosec B603
```
kevalmorabia97: Why do we need ssh + scp here? Are there any workarounds or whether we can make these steps manual for users to do on their own?
I think what we have here may not be secure

dthienan-nv: Context:
This PR expands the capabilities of the QDQ autotune feature by allowing us to target a _remote_  device.  Consequently the main use case for this is QNX-safe with the TensorRT safety runtime.
As such we need a way of pushing the engine files to the remote device for latency measurement, in this PR I'm using scp to push the engine file and then ssh to run trtexec_safe on the remote device.  There is no way to do this locally since we need to execute on the remote hardware in the remote environment for accurate latency measurements; and we cannot port ModelOpt to qnx-safe (as we don't even have python in that environment).
This tuning must be done in the loop as ModelOpt iteratively tries new QDQ placements, thus it is impractical to have the user execute steps as part of this process.

> I think what we have here may not be secure

Practically speaking users who use this feature will be executing in a secure environment with access controls to reach the remote device since they will need access to the host machine or physical access for flashing the device.

kevalmorabia97: Can we require users to first setup password-less ssh so we dont need to deal with password anywhere?

dthienan-nv: This is somewhat feasible, we can remove the password from ssh and scp calls but trtexec fails with the following error if we do not provide a password:

`Error Code: 9: Skipping tactic 0x0000000000000000 due to exception [autotuner.cpp:780: autotuner_t] Remote device requires -autotune:password In compileGraph at /_src/optimizer/myelin/codeGenerator.cpp:1766`

So while we can remove some handling of the password in subsequent calls, it is still visible from the call to the parent ModelOpt process cli args.  Due to this limitation I do not see a benefit to additional password obfuscation since the password must be exposed through the ModelOpt CLI args and then during the initial engine build the password will once more be exposed in the remoteAutoTuningConfig options at the trtexec call.

kevalmorabia97: Can you check with trtexec team if they have a more secure way of handling passwords e.g. via a env variable? Current approach in trtexec to explicitly add the password in the command seems very insecure

I'm concerned if we handle password in modelopt code side, we may need to get security team's guidance / approval to do so

kevalmorabia97:
/ok to test [5da6d33](https://github.com/NVIDIA/Model-Optimizer/pull/1378/commits/5da6d33151d7a3a12eab2038545cae21c69f2178)

kevalmorabia97:
/ok to test [39f0394](https://github.com/NVIDIA/Model-Optimizer/pull/1378/commits/39f0394b79be94e50d3579399f19605e253abd44)

arata-miyamoto:
@dthienan-nv 

### B1 — anchor: `benchmark.py` L412

**Please keep [`_redact_url_password()`](https://github.com/NVIDIA/Model-Optimizer/blob/4420ee48e9a26b56f37e3f7303d6c3b7e1084c21/modelopt/onnx/quantization/autotune/benchmark.py#L158-L167)**

[`27aee61`](https://github.com/NVIDIA/Model-Optimizer/commit/27aee61d9275f1d8e1cc2f87b11609fc3e6ef3d1) dropped it along with sshpass, but users can still put a password in the URL and we still log it — [L412](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L412), [L451](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L451), [L459](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L459) (`Command:` in `_write_log_file()`, which lands on disk).

It also needs to cover `result.stdout`/`result.stderr` at [L455-468](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L455-L468): trtexec echoes the URL in its own `&&&& RUNNING` banner ([raw argv](https://github.com/NVIDIA/TensorRT/blob/release/10.15/samples/trtexec/trtexec.cpp#L226)), so redacting our own command strings isn't enough.

---

### B2 — anchor: `benchmark.py` L525-532

**Forward the measurement settings to `trtexec_safe`**

The remote command at [L525-532](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L525-L532) is `--useCudaGraph --loadEngine=...` only, so we measure at trtexec_safe's defaults — [`warmUp` is 1 iteration](https://github.com/NVIDIA/TensorRT/blob/release/10.15/samples/trtSafeExec/trtSafeExec.cpp#L63). `--warmup_runs` / `--timing_runs` only reach the local build at [L449](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L449), which runs `--skipInference`.

Could we forward them? e.g.

    --warmUp={self.warmup_runs} --iterations={self.timing_runs} --avgRuns=1 --duration=0

---

### B3 — anchor: `benchmark.py` L152

**`_SAFE_PATTERN` reads the first trace block, not a median**

[`_SAFE_PATTERN`](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L152-L155) matches the `=== Trace details ===` lines and `re.search` takes the first one — the mean over the first `avgRuns` executions, right after warm-up. [L574](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L574) reports it as `(median)`.

trtexec_safe also prints `GPU Compute Time: ... median = ...` in its summary, under the [same guard](https://github.com/NVIDIA/TensorRT/blob/release/10.15/samples/trtSafeExec/trtSafeExec.cpp#L1136-L1166) as the trace lines, and [`_STD_PATTERN`](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L156) already matches it. Could we drop `_SAFE_PATTERN` and use `_STD_PATTERN` for both branches?

---

### B4 — anchor: `benchmark.py` L494

**Route the new failure paths through #2078's classifier**

scp failure ([L494](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L494)), remote run failure ([L566](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L566)) and `TimeoutExpired` ([L584](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L584)) all return `float("inf")`, so a dropped link gets recorded as a slow scheme.

#2078 handles this, but it's based on `main` and only covers the `inf` sites that exist there. Whichever of the two lands second, these three should go through [`_benchmark_failed()`](https://github.com/NVIDIA/Model-Optimizer/blob/c045c3de424d2bf1dd85501f12371f9420580867/modelopt/onnx/quantization/autotune/benchmark.py#L393-L409).

---

### B5 — anchor: `benchmark.py` L345

**Raise the default `network_timeout_seconds` and make it configurable**

The 300 s at [L345](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/benchmark.py#L345) is a plain wall clock — it kills the scp even while the transfer is progressing normally, and the candidate then becomes `inf`. You mentioned ~6 min for one model's upload, so the current default is already close to what you've measured. Could we raise it to something like 10 min?

It also can't be changed today: [`workflows.py:110`](https://github.com/NVIDIA/Model-Optimizer/blob/39f0394b79be94e50d3579399f19605e253abd44/modelopt/onnx/quantization/autotune/workflows.py#L110) doesn't pass it and there's no CLI flag. `quantize.py` has a second call site into `init_benchmark_instance()` that needs the same change.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -144,6 +149,184 @@ def _write_log_file(self, file: Path | str | None, content: str) -> None:
             self.logger.warning(f"Failed to save logs to {file}: {e}")
 
 
+_SAFE_PATTERN = (
+    r"\[\d{2}/\d{2}/\d{4}-\d{2}:\d{2}:\d{2}\]\s+\[I\]\s+"
+    r"Average over \d+ runs - GPU latency:\s*([\d.]+)\s*ms"
+)
+_STD_PATTERN = r"\[I\]\s+GPU Compute Time:.*?median\s*=\s*([\d.]+)\s*ms"
```
cjluo-nv, line 156: > _Bot comment._

`_STD_PATTERN` changes the metric for the **local/standard** path from `[I] Latency: … median` (end-to-end host latency) to `[I] GPU Compute Time: … median` (GPU-only). That's a silent behavior change for every existing `--use_trtexec` user — reported baselines and absolute numbers will move, and the GPU tests had to be edited to match. If the intent is to make the local number comparable with `trtexec_safe`'s "GPU latency", please say so in the PR body/CHANGELOG; otherwise keep `Latency:` for the non-safe path and use `GPU Compute Time` only where it's needed.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -170,59 +354,60 @@ def __init__(
...
                 self.logger.warning(
-                    "Remote autotuning is not supported with TensorRT version < 10.15. "
-                    "Removing --remoteAutoTuningConfig from trtexec arguments"
+                    "Remote autotuning is not supported with TensorRT version < 10.15."
                 )
-                trtexec_args = [
-                    arg for arg in trtexec_args if "--remoteAutoTuningConfig" not in arg
-                ]
-        self._base_cmd.extend(trtexec_args)
+                raise
```
cjluo-nv, line 406: > _Bot comment._

Behavior change: previously a TRT < 10.15 environment logged a warning and stripped `--remoteAutoTuningConfig`, continuing with a local benchmark. Now the `ImportError` propagates out of `__init__` (caught by `init_benchmark_instance`, which returns `None` → CLI exit 1). Failing fast is defensible, but it's user-visible and undocumented — please note it in the PR body / autotune guide. Same for the new `ValueError`s from strict URL validation.

Minor: the `network_timeout_seconds` entry in the docstring above is indented outside the `Args:` block, so napoleon will render it as loose text.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -287,19 +474,114 @@ def run(
                 self.logger.error(f"trtexec failed with return code {result.returncode}")
                 self.logger.error(f"stderr: {result.stderr}")
                 return float("inf")
+            latency_pattern = _STD_PATTERN
+            if self.has_remote_config and self.is_safe:
```
cjluo-nv, line 478: > _Bot comment._

`self.is_safe` is always `True` when `has_remote_config` is `True` (the constructor injects `--safe` via `_ensure_remote_autotuning_flags`, and `is_safe` is computed after that). The `and self.is_safe` conjunct is dead — either drop it or drop the forced injection so the flag can actually be false here.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -287,19 +474,114 @@ def run(
...
+                    finally:
+                        # Cleanup remote engine file after benchmarking to avoid disk filling up
+                        cleanup_cmd = [
+                            "ssh",
+                            "-p",
+                            str(self.remote_port),
+                            f"{self.remote_user}@{self.remote_ip}",
+                            f"rm -f {shlex.quote(self.remote_engine_path)}",
+                        ]
+                        cleanup_cmd = cleanup_cmd
```
cjluo-nv: > _Bot comment._

`cleanup_cmd = cleanup_cmd` is a no-op left over from a refactor; same pattern appears twice more for `trtexec_safe_cmd` (lines ~534 and ~553). Please remove all three.

While here: this nested `@contextlib.contextmanager` closure is only used once — a plain `try/finally` around the measurement (or a small `_cleanup_remote_engine()` method) would read simpler. The `ssh` invocations here and below also lack `-oStrictHostKeyChecking=accept-new` / `-oBatchMode=yes`, which only the `scp` call sets; `tools/launcher/core.py::_ControlMasterSession._ssh_opts` sets both and is worth mirroring so a prompt can't stall until the 5-minute timeout.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -287,19 +474,114 @@ def run(
...
-                self.logger.debug(f"trtexec stdout:\n{result.stdout}")
+                    trtexec_safe_cmd = trtexec_safe_cmd
+                    result = subprocess.run(
+                        trtexec_safe_cmd,
+                        capture_output=True,
+                        text=True,
+                        timeout=self.network_timeout_seconds,
+                    )  # nosec B603
+                    latency_pattern = _SAFE_PATTERN
+                    if result.returncode != 0:
```
cjluo-nv, line 541: > _Bot comment._

The fallback triggers on *any* non-zero exit from `trtexec_safe`, not just "binary not found", and it also swaps `latency_pattern` to `_STD_PATTERN`. Two consequences:

1. A genuine engine-load/runtime failure is retried with a different binary, masking the real error (the `trtexec_safe` stderr is discarded — only the fallback's is logged).
2. If `trtexec_safe` succeeds for some candidates and fails for others within the same autotune run, the search compares safe-mode "Average over N runs - GPU latency" against std "GPU Compute Time median" across candidates, i.e. non-comparable numbers driving scheme selection.

Suggest probing once (e.g. first invocation or an `ssh test -x`), caching which binary/pattern to use for the whole run, and logging the `trtexec_safe` stderr when falling back.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -170,59 +354,60 @@ def __init__(
...
+            log=self.logger,
+        )
 
-        if has_remote_config:
+        # Defaults for remote-autotuning fields; overwritten when configured.
+        self.has_remote_config: bool = False
+        self.remote_ip: str | None = None
+        self.remote_port: int = 22
+        self.remote_user: str = "root"
+        self.remote_engine_path: str = "trtexec_benchmark_model.trt"
```
cjluo-nv, line 384: > _Bot comment._

`remote_engine_path` is a fixed relative filename in the remote user's home dir and isn't configurable. Combined with the docs' "only one instance at a time" note this is workable, but two concurrent runs (or a stale run) will silently overwrite each other's engine and produce wrong latencies. Consider including `os.getpid()`/a uuid and/or placing it under a temp dir (`/tmp/…`) so accidental concurrency fails loudly instead of measuring the wrong engine.

docs/source/guides/9_autotune.rst
```diff
@@ -252,6 +252,10 @@ To use remote autotuning during Q/DQ placement optimization, run with ``trtexec`
 * Valid remote autotuning configuration
 * ``--use_trtexec`` must be set (benchmarking uses ``trtexec`` instead of the TensorRT Python API)
 * ``--safe --skipInference`` must be enabled via ``--trtexec_benchmark_args``
+* ssh and scp must be available on the local machine
```
cjluo-nv, line 255: > _Bot comment._

The code now *requires* the config value to be `ssh://user@host[:port]?remote_exec_path=<path-to-trtexec>&remote_lib_path=<lib-dir>` — anything else raises `ValueError` and aborts the run at benchmark init. The guide still only says "Replace `<remote autotuning config>` with an actual remote autotuning configuration string (see `trtexec --help`)". Please document the accepted form (and the fact that `remote_exec_path`'s dirname is reused to locate `trtexec_safe`), plus the new hard failure on TensorRT < 10.15.

tests/unit/onnx/quantization/autotune/test_cli_pipeline.py
```diff
@@ -0,0 +1,211 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv, line 1: > _Bot comment._

This file (+211) tests `run_autotune` argv plumbing and is unrelated to the `trtexec_safe` fix — it's good coverage, but it doubles the reviewable surface of a bugfix PR. Consider splitting it into its own PR.

cjluo-nv (commented):

**Design review (gate fired on 1465 lines; ~1090 of those are new tests):** the problem is real and narrow — with `--remoteAutoTuningConfig --safe --skipInference`, the local `trtexec` run only builds the engine, so no latency line is ever produced and every candidate scores `inf`. The fix (scp the safety engine to the target, run `trtexec_safe` over ssh, parse its latency line) is an incremental extension of the existing `TrtExecBenchmark`, not a new subsystem: no new module/directory, one new dataclass and four small module-level helpers in the file that already owned this logic. Alternatives considered: (a) `paramiko`/`fabric` — not a `modelopt` runtime dependency, and the repo's only other remote-exec code (`tools/launcher/core.py::_ControlMasterSession`) also shells out to `ssh`/`scp` via `subprocess`, so this is consistent with in-repo precedent; (b) having trtexec itself measure on the remote timing server — that's precisely what doesn't work in safe mode, which the PR body explains. So I'm satisfied on the "should this be built" question and reviewed for correctness. Concerns below.

Blocking-ish / worth resolving before merge:

1. **Leftover no-op self-assignments** (`cleanup_cmd = cleanup_cmd`, `trtexec_safe_cmd = trtexec_safe_cmd` ×2) — dead code from a refactor.
2. **The standard (non-remote) latency metric silently changed** from `Latency: … median` (end-to-end, incl. H2D/D2H) to `GPU Compute Time: … median` (GPU-only). Every existing `--use_trtexec` user's baseline/absolute numbers shift. Neither the PR body nor the docs mention this; if it's intentional (to be comparable with `trtexec_safe`'s "GPU latency"), say so explicitly.
3. **Fallback can mix two different metrics across iterations of one search.** Any non-zero exit from `trtexec_safe` (missing binary *or* a genuine load failure) falls back to `trtexec --safe` and switches to `_STD_PATTERN`. If that happens intermittently, the autotuner compares safe-mode "GPU latency" against std "GPU Compute Time median" between candidates. Probe once and cache which binary/metric to use.
4. **Docs don't document the now-mandatory config URL shape.** The constructor now hard-fails (`ValueError`) unless the value is `ssh://user@host[:port]?remote_exec_path=…&remote_lib_path=…`; the guide still just says "see `trtexec --help`". Also, TRT < 10.15 now raises instead of degrading gracefully — worth a docs/PR-body note.
5. **ssh option asymmetry**: only the `scp` call passes `-oStrictHostKeyChecking=accept-new`; the `ssh` calls pass neither that nor `BatchMode=yes`, so an unknown host / key-auth failure can block on a prompt until the 5-minute timeout. `tools/launcher/core.py` sets both — mirror it.
6. **Test scope**: `tests/unit/.../test_cli_pipeline.py` (+211) is unrelated to this bugfix (it exercises `run_autotune` argv plumbing). Fine code, but it inflates the diff; consider landing it separately. Also `network_timeout_seconds` isn't plumbed through `init_benchmark_instance`/CLI, so the documented CLI workflow can't tune it.

Positives: the new unit tests are genuinely good (URL parsing incl. argv-smuggling guards, scp/ssh ordering, fallback path, per-call timeouts, both regex patterns), remote engine cleanup is in a `finally`, and `shlex.quote` is applied on the remote-shell side. No licensing concerns (new files carry the standard NVIDIA/Apache header). No prompt-injection attempts observed in the PR content.

docs/source/guides/9_autotune.rst
dthienan-nv: Referencing trtexec for the auto tuning config is better since it may include additional options beyond what is required by this PR.  We should track the source of truth for this config for those options instead.

modelopt/onnx/quantization/autotune/__main__.py
```diff
@@ -314,6 +316,18 @@ def get_parser() -> argparse.ArgumentParser:
...
+        "--network_timeout_minutes",
+        type=int,
+        default=10,
+        help="Timeout used for network commands when performing remote auto tuning, mainly relevant for scp",
+    )
+    trt_group.add_argument(
+        "--remote_engine_path",
+        type=str,
+        default="trtexec_benchmark_model.trt",
+        help="Path used for storing temporary TensorRT engine files for remote profiling",
```
cjluo-nv: > _Bot comment._

This default undoes the uniqueness fix. `TrtExecBenchmark.__init__` only generates `trtexec_benchmark_model_{uuid4}.trt` when `remote_engine_path is None`, but the CLI always passes the literal `"trtexec_benchmark_model.trt"`, so every invocation through the documented entry point uses the same fixed name in the remote user's home dir — precisely the collision case the uuid was added for.

Suggest `default=None` here (and in `modelopt/onnx/quantization/__main__.py::--autotune_remote_engine_path` and `quantize.py`'s `autotune_remote_engine_path` / `_find_nodes_to_quantize_autotune(remote_engine_path=...)`), letting the benchmark pick the unique name unless the user explicitly overrides it. Worth a test asserting the CLI default reaches `init_benchmark_instance` as `None`.

modelopt/onnx/quantization/__main__.py
```diff
@@ -463,6 +463,18 @@ def get_parser() -> argparse.ArgumentParser:
...
     )
+    autotune_group.add_argument(
+        "--autotune_network_timeout_minutes",
+        type=int,
+        default=10,
+        help="Timeout used for network commands when performing remote auto tuning, mainly relevant for scp",
+    )
+    autotune_group.add_argument(
+        "--autotune_remote_engine_path",
+        type=str,
```
cjluo-nv, line 489: > _Bot comment._

Same as `autotune/__main__.py`: `default="trtexec_benchmark_model.trt"` overrides the per-instance uuid path that `TrtExecBenchmark` generates for `remote_engine_path=None`. Please default to `None` here and in `quantize.quantize(autotune_remote_engine_path=...)` / `_find_nodes_to_quantize_autotune(remote_engine_path=...)` so the collision-avoidance actually takes effect for CLI users.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -287,19 +486,112 @@ def run(
...
+            if result.returncode != 0:
+                self.logger.error(
+                    f"Failed to run trtexec_safe or trtexec with '--safe'\n{result.stdout}\n{result.stderr}"
+                )
+                return float("inf")
+            _latency_pattern = r"\[I\]\s+GPU Compute Time:.*?median\s*=\s*([\d.]+)\s*ms"
+            if not (match := re.search(_latency_pattern, result.stdout, re.IGNORECASE)):
+                # this could be due to creating a degenerate onnx file that can't be engine built.
+                # thus not a hard failure
+                self.logger.warning(f"trtexec stdout:\n{result.stdout}")
```
cjluo-nv: > _Bot comment._

Unredacted credential leak. `result.stdout` here is the local `trtexec` output, and trtexec echoes its full argv in the `&&&& RUNNING …` banner — including `--remoteAutoTuningConfig=ssh://user:pass@host`. Same for `self.logger.error(f"stderr: {result.stderr}")` in the non-zero-returncode branch above.

The log-file path is already wrapped in `_redact_url_password`; please wrap these two direct logger calls too (this is the second half of @arata-miyamoto's B1 — redacting only our own command strings isn't enough because trtexec re-prints the URL).

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -287,19 +486,112 @@ def run(
...
+                    "scp",
+                    "-P",
+                    str(self.remote_port),
+                    "-oStrictHostKeyChecking=accept-new",
+                    self.engine_path,
+                    f"{self.remote_user}@{self.remote_ip}:{shlex.quote(self.remote_engine_path)}",
+                ]
+                result = subprocess.run(
+                    scp_cmd, capture_output=True, text=True, timeout=self.network_timeout_seconds
+                )  # nosec B603
```
cjluo-nv: > _Bot comment._

`# nosec B603` (×4, plus `# nosec B404` on the import) is still unaccompanied. SECURITY.md's exception path requires (a) an inline comment stating *why* the pattern is necessary and *why* it's safe here, and (b) a justification in the PR description with review from `@NVIDIA/modelopt-setup-codeowners`.

The calls themselves are fine (list argv, no `shell=True`, `shlex.quote` on every interpolated remote token, argv-smuggling guards on user/host), and `ort_utils.py` in this same package already carries bare `# nosec B603`, so this is a sign-off question rather than a defect. Concretely: add a one-line why-safe comment at each site and a short "Security exception" section to the PR body, then tag the codeowners. (@kevalmorabia97 flagged this too.)

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -287,19 +486,112 @@ def run(
...
+                        )  # nosec B603
+                    except Exception as e:
+                        self.logger.warning(f"Error during remote engine cleanup: {e}")
+
+            if result.returncode != 0:
+                self.logger.error(
+                    f"Failed to run trtexec_safe or trtexec with '--safe'\n{result.stdout}\n{result.stderr}"
+                )
+                return float("inf")
+            _latency_pattern = r"\[I\]\s+GPU Compute Time:.*?median\s*=\s*([\d.]+)\s*ms"
```
cjluo-nv, line 578: > _Bot comment._

This pattern now applies to the *local* path as well, changing the reported metric for every existing `--use_trtexec` user from `[I] Latency: … median` (end-to-end, includes H2D/D2H) to `[I] GPU Compute Time: … median` (GPU-only). Making it comparable with `trtexec_safe`'s output is a defensible choice, but absolute baselines and previously-recorded numbers will shift, and neither the PR body nor `9_autotune.rst` nor CHANGELOG mentions it. Please call it out explicitly.

Minor while here: `_latency_pattern` is rebuilt on every `run()` — hoist it to a module-level `_LATENCY_PATTERN` (or `re.compile` it once). The new test module's docstring also still refers to `_SAFE_PATTERN`/`_STD_PATTERN`, which no longer exist.

cjluo-nv (commented):

Re-review of the `trtexec_safe` remote-benchmark fix. Design gate is satisfied: this is an incremental extension of the existing `TrtExecBenchmark` (one dataclass + four module-level helpers in the file that already owned the logic), and shelling out to `ssh`/`scp` via `subprocess` matches in-repo precedent (`tools/launcher/core.py`); no new subsystem or second config system is introduced. Licensing is clean (new test file carries the standard NVIDIA/Apache header). No prompt-injection attempts in the PR content.

Most previously-raised items are now genuinely closed: timeouts on all remote calls (+ `TimeoutExpired` handler), argv-smuggling guards on user/host, port-range validation, duplicate/missing query-param rejection, `shlex.quote` on every remote-shell token, remote engine cleanup in `finally`, `sshpass`/password handling removed in favour of key-based auth, `-oStrictHostKeyChecking=accept-new` on all four ssh/scp invocations, `--warmUp/--iterations/--avgRuns/--duration=0` forwarded to the remote run, `_SAFE_PATTERN` folded into a single pattern, `network_timeout` plumbed through both CLIs, and the leftover no-op self-assignments removed. Test coverage went from zero to substantial and the two regressed GPU tests were updated.

Remaining items (details inline):

1. **The CLI defaults defeat the unique remote engine filename.** `TrtExecBenchmark` generates `trtexec_benchmark_model_<uuid>.trt` when `remote_engine_path is None`, but all three CLI/API entry points hardcode `default="trtexec_benchmark_model.trt"`, so every user going through the documented commands lands back on the fixed shared name the uuid was added to fix. Defaults should be `None`.
2. **Credential redaction is incomplete.** `_redact_url_password` covers the log file and the debug command echo, but `logger.error(f"stderr: {result.stderr}")` and `logger.warning(f"trtexec stdout:\n{result.stdout}")` are raw — and trtexec echoes its own argv (including `--remoteAutoTuningConfig=ssh://user:pass@host`) in its `&&&& RUNNING` banner, which is exactly the case @arata-miyamoto's B1 asked to cover.
3. **:speech_balloon: `# nosec` exception is still open.** Author acknowledged "This still requires an exception." Four `# nosec B603` + one `# nosec B404` remain with no inline why-safe comment and no security-exception section in the PR description. There *is* in-package precedent (`ort_utils.py`), so this is a policy/sign-off call for @NVIDIA/modelopt-setup-codeowners rather than a code defect — but SECURITY.md requires both the inline justification and the PR-description request, and neither is present yet.
4. **:speech_balloon: The local/standard latency metric silently changed** from `[I] Latency: … median` (end-to-end) to `[I] GPU Compute Time: … median` (GPU-only) for *every* `--use_trtexec` user, not just the remote-safe path. The author's reply ("GPU compute time is available in older builds") answers availability but not the semantic shift — absolute baselines will move. Please note it in the PR body/CHANGELOG/docs.
5. Docs: the new hard failures (`ValueError` on a malformed config URL, `ImportError` propagating on TRT < 10.15 instead of degrading to a local benchmark) are user-visible behavior changes and aren't mentioned in `9_autotune.rst`. Also `remote_engine_path` isn't documented in `init_benchmark_instance`'s docstring.
6. Nits: `self.is_safe` is now dead in production code (only asserted in tests); the new test module's docstring still describes `_SAFE_PATTERN`/`_STD_PATTERN`, which no longer exist; a failed `scp` leaves a possibly-partial engine on the remote (early `return` happens before the cleanup `finally`).

cjluo-nv (commented):

Most prior feedback is genuinely resolved and test coverage is now strong, but the `# nosec` security exception is still incomplete per `SECURITY.md` and a few small issues remain.

Needs action:
- [ ] Add a "Security exception" section to the PR body justifying the four `# nosec B603` / one `# nosec B404` in `benchmark.py` and request review from `@NVIDIA/modelopt-setup-codeowners` — the inline why-safe comments cover step 1 of `SECURITY.md`, step 2 is still missing (author: "This still requires an exception").
- [ ] Fix the `FileNotFoundError` message in `benchmark.py:624` — `_run_trtexec` raises a fresh exception with no `filename`, so a missing local `trtexec` now logs "None not found". See inline.
- [ ] Add a `CHANGELOG.rst` entry under 0.48.0 for the user-visible changes: local `--use_trtexec` latency now parses `GPU Compute Time` instead of `Latency`, and TRT < 10.15 / malformed config URLs now raise instead of falling back.
- [ ] Align the `network_timeout_seconds=300` default in `TrtExecBenchmark` with the 10-minute CLI default, or document why direct API callers get half the timeout. See inline.
- [ ] Move the `scp` `subprocess.run` inside the `try` so a `TimeoutExpired` there still runs the remote cleanup. See inline.

No action needed:
- :heavy_check_mark: Resolved since the last review: CLI `remote_engine_path` defaults, stdout/stderr redaction, probe-once metric consistency, forwarded measurement flags, remote-engine cleanup, and the docs for the metric change and new hard failures.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -263,43 +503,132 @@ def run(
...
+                # thus not a hard failure
+                self.logger.warning(f"trtexec stdout:\n{_redact_url_password(result.stdout)}")
+                self.logger.error("Could not parse median latency from trtexec output")
                 return float("inf")
             latency = float(match.group(1))
             self.logger.info(f"TrtExec benchmark (median): {latency:.2f} ms")
             return latency
-        except FileNotFoundError:
+        except FileNotFoundError as e:
             self.logger.error(
```
cjluo-nv, line 642: > _Bot comment._

`_run_trtexec` in `ort_utils.py` catches the original `FileNotFoundError` and re-raises a new one built from a plain string, so `e.filename` is `None` on the most common failure path (trtexec not in PATH). This logs `None not found, please ensure system dependencies are installed and in the PATH: \n ssh, scp, trtexec`, which is worse than the message it replaced.

Suggest `f"{e.filename or e} not found..."`, or keep the generic dependency list without interpolating `e.filename`.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -263,43 +503,132 @@ def run(
...
+            if self.has_remote_config:
+                # need to push the model to the device and use trtexec_safe to run
+                scp_cmd = [
+                    "scp",
+                    "-P",
+                    str(self.remote_port),
+                    "-oStrictHostKeyChecking=accept-new",
+                    self.engine_path,
+                    f"{self.remote_user}@{self.remote_ip}:{shlex.quote(self.remote_engine_path)}",
+                ]
```
cjluo-nv, line 549: > _Bot comment._

The `scp` call sits outside the `try`, so a `TimeoutExpired` here propagates straight to the outer handler and the `finally` cleanup never runs — a partially transferred engine can be left on the remote device. Moving this call inside the `try` (the `returncode != 0` check already is) makes cleanup unconditional.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -159,6 +341,8 @@ def __init__(
         timing_runs: int = 10,
         plugin_libraries: list[str] | None = None,
         trtexec_args: list[str] | None = None,
+        network_timeout_seconds: float = 60 * 5,  # 5 minutes
```
cjluo-nv: > _Bot comment._

This default is 5 minutes, but all three CLI entry points default `network_timeout_minutes` to 10. Direct API callers therefore get half the timeout the documented workflow uses, on the operation (`scp` of a large engine) that motivated raising it. Suggest making this `60 * 10` so the two agree, and updating the docstring and `test_network_timeout_default_is_five_minutes`.

kevalmorabia97:
/ok to test [c4315fa](https://github.com/NVIDIA/Model-Optimizer/pull/1378/commits/c4315facbce59b56373fb4ccb451d8b641f4d6ce)

cjluo-nv (commented):

Prior concerns are resolved and the remote path now reads correctly, but two open questions remain before this can merge.

Needs action:
- [ ] Narrow the new `except (ValueError, ImportError): raise` in `autotune/workflows.py:141` — it also escapes for `TensorRTPyBenchmark`'s "TensorRT not available" / "PyTorch CUDA not available" `ImportError`, so a plain missing-TRT install now produces a raw traceback out of `run_autotune()`/`quantize()` instead of the previous "Failed to initialize TensorRT benchmark" + exit 1. Confirm intent or scope the re-raise to the remote-config path, and note it in `CHANGELOG.rst`.
- [ ] :speech_balloon: Security Exception section is now in the PR body and each site has a why-safe comment — still needs an explicit review request to `@NVIDIA/modelopt-setup-codeowners` for the four `# nosec B603` and the `# nosec B404` per `SECURITY.md` step 2.
- [ ] Use `self.has_remote_config` instead of `self._remote_use_trtexec_safe is not None` for latency-pattern selection in `benchmark.py:622` — equivalent today but fragile.
- [ ] Add `--warmUp`/`--iterations` to the forwarded-flags bullet in `docs/source/guides/9_autotune.rst:259`.

No action needed:
- :heavy_check_mark: 5 previous concerns addressed, including the `FileNotFoundError` filename and moving `scp` inside the `try`/`finally`.

kevalmorabia97:
/ok to test [65cd70d](https://github.com/NVIDIA/Model-Optimizer/pull/1378/commits/65cd70d852ee75114ce72aad4d13d62fa46097d6)

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -263,43 +510,143 @@ def run(
...
+                        "-oBatchMode=yes",
+                        "-oStrictHostKeyChecking=accept-new",
+                        "-p",
+                        f"{self.remote_port}",
+                        f"{self.remote_user}@{self.remote_ip}",
+                        f"{ld_path} {shlex.quote(trt_path)} {extra_flags}--useCudaGraph "
+                        f"--warmUp={self.warmup_runs} --iterations={self.timing_runs} "
+                        f"--avgRuns={self.timing_runs} --duration=0 "
+                        f"--loadEngine={shlex.quote(self.remote_engine_path)}",
+                    ]
```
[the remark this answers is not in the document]
dthienan-nv: Dynamic shapes are not permitted in safety thus not used in remote benchmarking.  We should not need to reproduce the safety restrictions here as we want to maintain one source of truth.  Thus assume the user is already familiar with the external restrictions.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -263,43 +510,143 @@ def run(
...
+                    f"{binary_name} failed (exit {result.returncode}):\n"
+                    f"{_redact_url_password(result.stdout)}\n{_redact_url_password(result.stderr)}"
+                )
+                return float("inf")
+            # trtexec_safe / trtexec --safe emit "GPU Compute Time"; local trtexec emits "Latency"
+            _latency_pattern = (
+                r"\[I\]\s+GPU Compute Time:.*?median\s*=\s*([\d.]+)\s*ms"
+                if self._remote_use_trtexec_safe is not None
+                else r"\[I\]\s+Latency:.*?median\s*=\s*([\d.]+)\s*ms"
+            )
```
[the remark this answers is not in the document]
dthienan-nv: Somehow the agents are still wrong. We must key off of latency when using the trtexec binary which can still be remote with --safe and we need to key off of GPU Compute Time when using the trtexec_safe binary.

kevalmorabia97:
please rebase your PR so merge conflicts is fixed and also see if any claude review comment is pending to be addressed or not

PR #1380, opened 2026-05-01
[DeepSeek] Default to top-k calibration with peer-max input amax sync

cjluo-nv:
## Summary

- DeepSeek PTQ (`examples/deepseek/ptq.py`) now defaults to native top-k routing during MoE calibration. The previous all-tokens-to-all-experts path (`CalibMoe`) is preserved behind a new `--calib_all_experts` flag.
- After `mtq.quantize`, `fixup_moe_expert_amax` syncs every expert's `input_quantizer.amax` (w1/w2/w3) to the per-layer global peer max via `dist.all_reduce(MAX)` across EP ranks. `weight_quantizer.amax` stays per-expert; any uncalibrated expert is filled by computing amax over the dequantized FP8 weight.
- `mtq.print_quant_summary` is now also written to `<output_path>/.quant_summary.txt`, mirroring `llm_ptq/hf_ptq.py`.

## Why

Forcing all tokens through every expert doubled calibration time and inflated `input_quantizer.amax` for cold-routing experts with outliers they never see at inference. The new flow matches the inference distribution, runs roughly 2x faster, and mirrors the `layer_sync_moe_local_experts_amax` semantics that mtq runs automatically for `QuantSequentialMLP`-derived MoEs.

## Validation (DeepSeek-V3.2-Exp, MP=8, NVFP4_DEFAULT_CFG)

Compared `_amax_baseline` (CalibMoe) vs `_amax_synced` (new default):
- All 44,544 expert weight amaxes bit-identical.
- Attention, shared experts, gate: identical.
- Expert `w1.input` and `w3.input` (shared MoE block input): identical.
- Expert `w2.input` (post-SiLU gated, expert-specific): synced to layer-wide peer max — 99.3% are larger than baseline (median 11.4x) since peer-max captures the worst-case outlier from any expert in the layer; 0.7% are smaller. This is the same trade-off `set_expert_quantizer_amax` makes for HF MoEs in `unified_export_hf.py`.

## Test plan

- [x] DeepSeek-V3.2-Exp MP8 PTQ with default flags — completes in ~7 min (vs ~27 min with CalibMoe), produces `_amax_synced/` consistent with the comparison above.
- [x] DeepSeek-V3.2-Exp MP8 PTQ with `--calib_all_experts` — produces `_amax_baseline/` identical (other than rounding) to the prior `CalibMoe`-default behavior.
- [x] `.quant_summary.txt` written under `output_path` on rank 0.

## Summary by CodeRabbit

* **New Features**
  * Added a `--calib_all_experts` option to enable an alternate PTQ calibration mode; default remains top-k routing with a post-calibration per-layer peer-max synchronization and a compute fallback for uncalibrated experts.
* **Documentation**
  * Clarified default and alternate calibration behaviors and added note about generation of a `.quant_summary.txt` summary file.

### DIFF
--- a/examples/deepseek/ptq.py
+++ b/examples/deepseek/ptq.py
@@ -290,7 +290,9 @@ def fixup_moe_expert_amax(transformer):
                     continue
                 # DeepSeek stores experts as FP8 with a per-block .scale; dequantize
                 # to bf16 first so we measure the real weight distribution, not bytes.
-                deq = weight_dequant(w, w.scale, torch.bfloat16) if w.element_size() == 1 else w
+                deq = (
+                    weight_dequant(w, w.scale, dtype=torch.bfloat16) if w.element_size() == 1 else w
+                )
                 axis = getattr(wq, "_axis", None)
                 if axis is None:
                     reduce_axis = None
--- a/examples/deepseek/ptq.py
+++ b/examples/deepseek/ptq.py
@@ -61,6 +61,7 @@ from modelopt.torch.quantization.utils import (
     is_quantized_column_parallel_linear,
     is_quantized_parallel_linear,
     is_quantized_row_parallel_linear,
+    reduce_amax,
 )
 from modelopt.torch.utils.dataset_utils import get_dataset_dataloader
 from modelopt.torch.utils.distributed import ParallelState
@@ -81,7 +82,7 @@ import model as deekseep_model  # noqa: E402
 from kernel import act_quant, fp8_gemm  # noqa: E402
 
 
-def monkey_patch_deepseek_model():
+def monkey_patch_deepseek_model(calib_all_experts: bool = False):
     gemm_impl: Literal["bf16", "fp8"] = "bf16"
     block_size = 128
 
@@ -199,6 +200,10 @@ def monkey_patch_deepseek_model():
             self.pe_bmm_quantizer = TensorQuantizer()
 
     class CalibMoe(deekseep_model.MoE):
+        """MoE override that forces every token through every expert during
+        calibration. Slower (~2x forwards) but every expert sees the full token
+        distribution, so per-expert input amaxes are calibrated directly."""
+
         def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
             self._setup()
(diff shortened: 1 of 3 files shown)


examples/deepseek/ptq.py
```diff
@@ -228,10 +230,86 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+                if not _missing(wq.amax):
+                    continue
+                w = lin.weight
+                if w.is_meta:
+                    continue
+                # DeepSeek stores experts as FP8 with a per-block .scale; dequantize
+                # to bf16 first so we measure the real weight distribution, not bytes.
+                deq = weight_dequant(w, w.scale, torch.bfloat16) if w.element_size() == 1 else w
+                axis = getattr(wq, "_axis", None)
+                if axis is None:
```
meenchen, line 297: > _Bot comment._

Bug: `torch.bfloat16` is passed as the third positional argument, which maps to `block_size` (an `int`), not `dtype`. This will fail at runtime in the triton kernel.

```python
# Current (broken):
deq = weight_dequant(w, w.scale, torch.bfloat16) if w.element_size() == 1 else w

# Fix — use the keyword argument, matching the `linear()` function above:
deq = weight_dequant(w, w.scale, dtype=torch.bfloat16) if w.element_size() == 1 else w
```

meenchen (commented):

Bug found in `fixup_moe_expert_amax`: `weight_dequant(w, w.scale, torch.bfloat16)` passes `torch.bfloat16` as the `block_size` positional argument instead of `dtype`. This will cause a runtime error (or incorrect behavior) in the triton kernel. The existing `linear()` function in the same file correctly uses the keyword form `dtype=torch.bfloat16`.

Otherwise the PR is well-structured and reasonable in size (~141 lines). The approach of syncing expert input amaxes to peer max across EP ranks makes sense, and the custom `fixup_moe_expert_amax` is justified since the existing `sync_moe_expert_amax` utility doesn't handle cross-rank `dist.all_reduce` needed for expert parallelism. The `print_quant_summary` with `output_path` uses the newly added signature in `model_quant.py` correctly. No new tests, but this is an example script so integration testing is the primary validation path.

cjluo-nv:
Thanks for the catch — fixed in 3e6b05ec. Confirmed via grep that this was the only call site that passed dtype positionally; the existing `linear()` helper in the same file already uses the keyword form. As noted, the buggy branch never fired in our validation run (every expert got activated, so `_missing(wq.amax)` was always False), but it would have triggered for any setup with cold experts.

meenchen (approved):

The critical bug from the previous review — `weight_dequant(w, w.scale, torch.bfloat16)` passing `torch.bfloat16` as positional `block_size` — has been fixed to use `dtype=torch.bfloat16` as a keyword argument. 

The PR is well-scoped (~143 lines), correctly implements peer-max input amax sync across EP ranks via `dist.all_reduce(MAX)`, preserves the old behavior behind `--calib_all_experts`, and includes good documentation. As an example script, integration testing (documented in the test plan) is the appropriate validation path. The `reduce_amax` import is valid and available from the quantization utils.

PR #1381, opened 2026-05-01
Adds AutoQuant support for VLM / Qwen3.5-Qwen3.6 style models

meenchen:
### What does this PR do?

Type of change: new feature, bug fix, new tests

### Details

- Enables AutoQuant search over fused MoE expert containers by snapshotting/restoring their per-expert quantizers.
- Adds Qwen3.5/3.6 linear-attention grouping rules so fused deployment layers keep compatible quant formats.
- Supports `w4a16_nvfp4` as an AutoQuant search format.
- Preserves disabled AutoQuant layer patterns in generated configs while allowing selected modules like `lm_head` to override default disables.
- Keeps recipe-mode and AutoQuantize VLM paths on the outer CausalLM so Qwen3.5/3.6-MoE `lm_head` remains visible.
- Skips `parent_class`-scoped quant config entries during AutoQuant bare quantizer matching, preventing class-scoped global entries from last-match overriding every selected module.
- Adds temporary hardcoded Qwen/VLM AutoQuant disabled-layer patterns in `hf_ptq.py` with a TODO to refactor into the config system.

### Usage

```bash
  --pyt_ckpt_path <model_path> \
  --qformat fp8,w4a16_nvfp4 \
  --auto_quantize_bits 5.0 \
  --auto_quantize_cost_model active_moe \
  --auto_quantize_checkpoint <autoquant_state.pt> \
  --export_path <output_dir>
```

### Testing

- `/Users/weimingc/miniconda3/envs/modelopt/bin/python -m pytest tests/unit/torch/quantization/test_autoquant.py::test_get_auto_quantize_config_keeps_selected_lm_head_enabled tests/unit/torch/quantization/test_config_validation.py::TestMatchQuantizerCfg::test_parent_class_scoped_entries_are_ignored_for_bare_autoquant_lookup`
- `/Users/weimingc/miniconda3/envs/modelopt/bin/python -m pytest tests/unit/torch/quantization/test_autoquant.py tests/unit/torch/quantization/test_config_validation.py -k "not data_parallel"` (`120 passed, 1 deselected`)
- `/Users/weimingc/miniconda3/envs/modelopt/bin/python -m py_compile examples/llm_ptq/hf_ptq.py modelopt/torch/quantization/algorithms.py modelopt/torch/quantization/_auto_quantize_cost.py tests/unit/torch/quantization/test_autoquant.py tests/unit/torch/quantization/test_config_validation.py`
- Full local affected-file pytest without `-k "not data_parallel"` only failed `test_data_parallel_auto_quantize` because this local sandbox cannot bind a free socket (`PermissionError: Operation not permitted`).
- Ran Qwen3.6 35B AutoQuant e2e with `fp8,w4a16_nvfp4` and exported a checkpoint.
- Verified exported checkpoint loads in vLLM nightly without local patches.

## Summary by CodeRabbit

* **New Features**
  * Added w4a16_nvfp4 quantization format and optional cost-exclusion patterns for AutoQuantize.

* **Improvements**
  * Safer multimodal/VLM handling and AutoQuantize now runs on the full outer model when applicable.
  * Better fused-MoE support, more accurate weight accounting, and refined attention-grouping for improved quantization choices.
  * Dynamic layer-disabling support for targeted disables.

* **Tests**
  * New unit tests covering cost-model exclusions, fused-MoE accounting, and config selection.

* **Documentation**
  * Updated cost-constraint example to show exclusion-pattern usage.

### DIFF
--- a/examples/llm_ptq/hf_ptq.py
+++ b/examples/llm_ptq/hf_ptq.py
@@ -15,7 +15,6 @@
 
 import argparse
 import copy
-import os
 import random
 import time
 import warnings
@@ -104,7 +103,6 @@ QUANT_CFG_CHOICES: dict[str, dict[str, Any]] = {
     "int8_sq": mtq.INT8_SMOOTHQUANT_CFG,
     "int8_wo": mtq.INT8_WEIGHT_ONLY_CFG,
     "fp8": mtq.FP8_DEFAULT_CFG,
-    "fp8_w8a8": mtq.FP8_DEFAULT_CFG,
     "int4_awq": mtq.INT4_AWQ_CFG,
     "w4a8_awq": mtq.W4A8_AWQ_BETA_CFG,
     "nvfp4": mtq.NVFP4_DEFAULT_CFG,
@@ -139,41 +137,25 @@ _KV_CAST_FORMATS = {"fp8_cast", "nvfp4_cast"}
 mto.enable_huggingface_checkpointing()
 
 
-NVFP4_W4A16_CFG = {
-    "quant_cfg": [
-        {"quantizer_name": "*", "enable": False},
-        {
-            "quantizer_name": "*weight_quantizer",
-            "cfg": {
-                "num_bits": (2, 1),
-                "block_sizes": {-1: 16, "type": "dynamic", "scale_bits": (4, 3)},
-            },
-        },
-        {"quantizer_name": "*input_quantizer", "enable": False},
-        *_default_disabled_quantizer_cfg,
-    ],
-    "algorithm": "max",
-}
-
-FP8_W8A16_CFG = {
-    "quant_cfg": [
-        {"quantizer_name": "*", "enable": False},
-        {
-            "quantizer_name": "*weight_quantizer",
-            "cfg": {"num_bits": (4, 3), "axis": None},
-        },
-        {"quantizer_name": "*input_quantizer", "enable": False},
-        *_default_disabled_quantizer_cfg,
-    ],
-    "algorithm": "max",
-}
-
-QUANT_CFG_CHOICES.update(
-    {
-        "nvfp4_w4a16": NVFP4_W4A16_CFG,
-        "fp8_w8a16": FP8_W8A16_CFG,
-    }
+_QWEN36_AUTOQ_DISABLED_LAYERS = (
+    "*shared_expert_gate*",
+    "*linear_attn.in_proj_a*",
+    "*linear_attn.in_proj_b*",
 )
+_VLM_AUTOQ_DISABLED_LAYERS = ("*visual*", "*mtp*", "*vision_tower*")
+
+
+def get_auto_quantize_disabled_layers(model) -> list[str]:
+    """Return layer patterns that should be excluded from AutoQuantize search."""
+    disabled_layers = [
+        entry["quantizer_name"]
+        for entry in _default_disabled_quantizer_cfg
+        if "parent_class" not in entry and entry["quantizer_name"] != "*lm_head*"
+    ]
+    disabled_layers.extend(p for p in _QWEN36_AUTOQ_DISABLED_LAYERS if p not in disabled_layers)
+    if is_multimodal_model(model):
+        disabled_layers.extend(p for p in _VLM_AUTOQ_DISABLED_LAYERS if p not in disabled_layers)
+    return disabled_layers
 
 
 def extract_and_prepare_language_model_from_vl(full_model):
@@ -351,7 +333,6 @@ def auto_quantize(
         qformat
         in [
             "fp8",
-            "fp8_w8a8",
             "int8_sq",
             "int8_wo",
             "int4_awq",
@@ -366,8 +347,6 @@ def auto_quantize(
             "nvfp4_omlp_only",
             "nvfp4_local_hessian",
             "mxfp8",
-            "nvfp4_w4a16",
-            "fp8_w8a16",
         ]
         for qformat in qformat_list
     ), "One or more quantization formats provided are not supported for unified checkpoint export"
@@ -390,44 +369,6 @@ def auto_quantize(
             f"Invalid auto_quantize_method: {auto_quantize_method}. Must be 'gradient' or 'kl_div'"
         )
 
-    # Let AutoQuantize search lm_head, but keep modules out that vLLM either
-    # constructs as BF16-only paths today or has known unsafe fused dispatch for.
-    disabled_layers = [
-        entry["quantizer_name"]
-        for entry in _default_disabled_quantizer_cfg
-        if "parent_class" not in entry and entry["quantizer_name"] != "*lm_head*"
-    ]
-    enable_linear_attn_big3 = os.environ.get("MODELOPT_AUTOQ_ENABLE_LINEAR_ATTN_BIG3") == "1"
-    enable_linear_attn_all = os.environ.get("MODELOPT_AUTOQ_ENABLE_LINEAR_ATTN_ALL") == "1"
-    enable_shared_expert = os.environ.get("MODELOPT_AUTOQ_ENABLE_SHARED_EXPERT") == "1"
-    if enable_linear_attn_all:
-        enable_linear_attn_big3 = True
-    autoq_extra_disabled = [
-        "*shared_expert_gate*",
-        # Keep the GDN a/b projections in BF16 even for "all linear_attn"
-        # searches. Prior healthy NVFP4 controls excluded these small
-        # projections, while low-end full-search checkpoints quantized them.
-        "*linear_attn.in_proj_a*",
-        "*linear_attn.in_proj_b*",
-    ]
-    if not enable_shared_expert:
-        autoq_extra_disabled.append("*mlp.shared_expert*")
-    if not enable_linear_attn_big3:
-        autoq_extra_disabled.extend(
-            [
-                "*linear_attn.in_proj_qkv*",
-                "*linear_attn.in_proj_z*",
-                "*linear_attn.out_proj*",
-            ]
-        )
-    for pat in autoq_extra_disabled:
-        if pat not in disabled_layers:
-            disabled_layers.append(pat)
-    if is_multimodal_model(language_model):
-        for pat in ("*visual*", "*mtp*", "*vision_tower*"):
-            if pat not in disabled_layers:
-                disabled_layers.append(pat)
-
     language_model, _ = mtq.auto_quantize(
         language_model,
         constraints={"effective_bits": args.auto_quantize_bits},
@@ -442,13 +383,11 @@ def auto_quantize(
             len(calib_dataloader), max(auto_quantize_score_size // args.batch_size, 1)
         ),
         verbose=True,
-        disabled_layers=disabled_layers,
+        disabled_layers=get_auto_quantize_disabled_layers(language_model),
         method=auto_quantize_method,
         checkpoint=auto_quantize_checkpoint,
         cost_model=args.auto_quantize_cost_model,
         active_moe_expert_ratio=args.auto_quantize_active_moe_expert_ratio,
-        cost_lower_bound=args.auto_quantize_cost_lower_bound,
-        cost_objective=args.auto_quantize_cost_objective,
     )
 
     calibrate_loop = create_forward_loop(dataloader=calib_dataloader)
@@ -1488,27 +1427,6 @@ def parse_args() -> argparse.Namespace:
         ),
     )
     parser.add_argument(
-        "--auto_quantize_cost_lower_bound",
-        type=float,
-        default=None,
-        help=(
-            "Optional lower bound, as a fraction of the requested effective-bits budget, "
-            "for the auto_quantize LP. Active-MoE cost mode uses a best-effort lower bound "
-            "by default when this is omitted."
-        ),
-    )
-    parser.add_argument(
-        "--auto_quantize_cost_objective",
-        type=str,
-        default="sensitivity",
-        choices=["sensitivity", "active_moe"],
-        help=(
-            "Objective for auto_quantize LP. 'sensitivity' minimizes quantization sensitivity. "
-            "'active_moe' minimizes active routed-MoE cost while the cost model constraint "
-            "still controls the requested budget."
-        ),
-    )
-    parser.add_argument(
         "--moe_calib_experts_ratio",
         type=float,
         default=None,
@@ -1535,17 +1453,12 @@ def parse_args() -> argparse.Namespace:
         parser.error("--auto_quantize_active_moe_expert_ratio must be in the range (0.0, 1.0].")
     if (
         args.auto_quantize_cost_model == "weight"
-        and args.auto_quantize_cost_objective != "active_moe"
         and args.auto_quantize_active_moe_expert_ratio is not None
     ):
         parser.error(
             "--auto_quantize_active_moe_expert_ratio requires "
-            "--auto_quantize_cost_model active_moe or --auto_quantize_cost_objective active_moe."
+            "--auto_quantize_cost_model active_moe."
         )
-    if args.auto_quantize_cost_lower_bound is not None and not (
-        0.0 < args.auto_quantize_cost_lower_bound <= 1.0
-    ):
-        parser.error("--auto_quantize_cost_lower_bound must be in the range (0.0, 1.0].")
 
     if args.specdec_offline_dataset is not None and args.sparsity_fmt != "dense":
         parser.error("--specdec_offline_dataset is only supported with --sparsity_fmt dense (PTQ).")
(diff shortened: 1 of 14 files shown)


examples/llm_ptq/hf_ptq.py
```diff
@@ -140,6 +140,28 @@ def _set_kv_cache_constant_amax(quant_cfg: list) -> None:
 mto.enable_huggingface_checkpointing()
 
 
+# TODO: To be refacored into config system.
+_QWEN36_AUTOQ_DISABLED_LAYERS = (
+    "*shared_expert_gate*",
+    "*linear_attn.in_proj_a*",
+    "*linear_attn.in_proj_b*",
+)
+_VLM_AUTOQ_DISABLED_LAYERS = ("*visual*", "*mtp*", "*vision_tower*")
```
meenchen: @juhi10071998 let's move this to the config systems once ready

cjluo-nv (commented):

The fused-MoE / w4a16_nvfp4 / linear_attn grouping changes look reasonable and include unit tests for the fused-MoE cost path and the `lm_head` override. A few things to address before this lands:

1. **Likely regression in `--recipe` path for Qwen3.5/3.6-MoE VLMs.** The gate in `load_model` flipped from `if args.recipe is None:` to `if args.auto_quantize_bits is None:`. Per the deleted comment, recipe-mode was *exactly* the case that needed `language_model = full_model` so a recipe targeting `*lm_head*` could see the outer CausalLM's `lm_head`. Under the new condition, recipe + VLM now extracts the inner LM and passes it to `mono_quantize`, which is the opposite of what's intended — recipe-driven `lm_head` quantization of Qwen3.5/3.6-MoE VLMs will silently miss `lm_head`. The autoquant path can be opted into without breaking the recipe path, e.g. `if args.recipe is None and args.auto_quantize_bits is None:`.

2. **Duplicated `_get_module_weight_numel`.** The same helper is defined in both `modelopt/torch/quantization/algorithms.py` and `modelopt/torch/quantization/_auto_quantize_cost.py`. Pick one canonical location (the cost module is the more natural home; `algorithms.py` already imports from it) and re-export.

3. **`_match_quantizer_cfg` now skips `parent_class` entries.** This looks like a real bug fix (a `{parent_class: nn.BatchNorm1d, quantizer_name: '*', enable: False}` entry would previously last-match-win against any `weight_quantizer`/`input_quantizer` lookup and disable everything), but it changes behavior for any recipe that relies on `parent_class`-scoped global entries reaching `get_auto_quantize_config` output. Worth either calling out explicitly in the PR body or adding a regression test that pins the new behavior.

4. **Hardcoded VLM/Qwen patterns in `hf_ptq.py`.** Already acknowledged with a TODO; just noting that `*shared_expert_gate*` is broader than the existing `*mlp.shared_expert_gate.*` in `_default_disabled_quantizer_cfg` — fine in practice, but worth keeping in mind when refactoring into the config system.

5. **Test fragility.** `test_get_auto_quantize_config_keeps_selected_lm_head_enabled` relies on `FP8_DEFAULT_CFG` containing `*lm_head*` (via `_default_disabled_quantizer_cfg`) for the assertion to succeed. If that default ever changes, the test will fail confusingly. Consider seeding the search_state's recipe with an explicit `*lm_head*: enable=False` entry so the test is self-contained.

examples/llm_ptq/hf_ptq.py
```diff
@@ -550,12 +568,9 @@ def load_model(args: argparse.Namespace):
...
 
-            # We only quantize the language model for VLMs other than the type supported above.
-            # Recipe mode is the exception: in Qwen3.5/3.6-MoE VLMs, lm_head sits
-            # on the outer CausalLM, not the inner language backbone. A recipe that targets
-            # lm_head must therefore quantize against the full model and explicitly keep visual
-            # and MTP siblings disabled.
-            if args.recipe is None:
+            # AutoQuantize walks the outer CausalLM so lm_head is visible to the
+            # search. Visual/MTP siblings are excluded by disabled-layer patterns.
+            if args.auto_quantize_bits is None:
```

This gate was previously `if args.recipe is None:` with a comment explaining that recipe-mode is the exception that needs `language_model = full_model` so a recipe targeting `*lm_head*` can see the outer CausalLM's `lm_head` on Qwen3.5/3.6-MoE VLMs.

Flipping it to `if args.auto_quantize_bits is None:` reverses that behavior: recipe + VLM now extracts the inner LM and passes it to `mono_quantize`, so a recipe that quantizes `*lm_head*` will silently miss it. AutoQuantize can be opted into without breaking the recipe path, e.g.:

```python
if args.recipe is None and args.auto_quantize_bits is None:
    extracted_lm, extracted_model_type = extract_and_prepare_language_model_from_vl(full_model)
```

modelopt/torch/quantization/algorithms.py
```diff
@@ -54,6 +54,77 @@
...
+
+    Standard QuantLinear modules have a single ``weight`` parameter. Fused
+    experts modules have two 3-D fused parameters (``gate_up_proj`` and
+    ``down_proj``) instead — both contribute to the cost accounting.
+    """
+    if _is_fused_experts_module(module):
+        total = 0
+        for attr in ("gate_up_proj", "down_proj"):
+            param = getattr(module, attr, None)
+            if param is not None:
```
cjluo-nv: > _Bot comment._

This helper is duplicated in `_auto_quantize_cost.py`. Pick one canonical location (the cost module is the natural home — `algorithms.py` already imports from it) and re-export from the other to avoid drift.

modelopt/torch/quantization/algorithms.py
```diff
@@ -1502,6 +1618,9 @@ def _match_quantizer_cfg(quant_cfg, quantizer_attr):
     matched = None
     matched_enable = None
     for entry in quant_cfg:
+        parent_class = entry.get("parent_class") if hasattr(entry, "get") else entry.parent_class
+        if parent_class is not None:
+            continue
```
cjluo-nv, line 1634: > _Bot comment._

Skipping `parent_class`-scoped entries here looks like a genuine bug fix (e.g. `{parent_class: nn.BatchNorm1d, quantizer_name: '*', enable: False}` would previously last-match-win against any `weight_quantizer` lookup and disable everything), but it's a behavior change worth calling out in the PR body and pinning with a regression test.

tests/unit/torch/quantization/test_autoquant.py
```diff
@@ -619,3 +639,32 @@ def test_get_auto_quantize_config(method):
...
+
+    config = mtq.get_auto_quantize_config(search_state)
+    quant_cfg = config["quant_cfg"]
+    quantizer_names = [entry["quantizer_name"] for entry in quant_cfg]
+
+    default_disable_idx = next(
+        idx for idx, entry in enumerate(quant_cfg) if entry["quantizer_name"] == "*lm_head*"
+    )
+    weight_idx = next(
+        idx
```
cjluo-nv, line 712: > _Bot comment._

This assertion implicitly depends on `FP8_DEFAULT_CFG` containing a `*lm_head*` disable via `_default_disabled_quantizer_cfg`. If that default ever changes, the test will fail confusingly. Consider constructing a small explicit recipe (with `*lm_head*: enable=False` in its `quant_cfg`) so the test pins the override behavior independent of the default config contents.

examples/llm_ptq/hf_ptq.py
juhi10071998: thanks @meenchen , yes let me do that, unfortunately I got pulled into a PO day0 model and won't get time until next week

meenchen: no problem, just want to make sure we are aligned with the future refactor.

cjluo-nv (commented):

Re-review of PR #1381 (AutoQuant for Qwen3.5/3.6 + VLM).

**Previous critical comments — addressed:**
- :speech_balloon: cjluo-nv's `--recipe` regression: gate is now `if args.recipe is None and args.auto_quantize_bits is None:` at hf_ptq.py:583 — recipe-mode VLMs again retain the outer CausalLM. ✅
- :speech_balloon: Duplicated `_get_module_weight_numel`: deduped — single canonical helper now lives in `_auto_quantize_cost.py` and is imported into `algorithms.py`. ✅
- :speech_balloon: `_match_quantizer_cfg` `parent_class`-skip behavior change: pinned with `test_parent_class_scoped_entries_are_ignored_for_bare_autoquant_lookup` regression test. ✅
- :speech_balloon: lm_head test fragility: test now appends `{quantizer_name: "*lm_head*", enable: False}` to its own recipe (test_autoquant.py:695), so it no longer relies on `_default_disabled_quantizer_cfg` for the enable-check. The `num_bits == (4, 3)` assertion still implicitly trusts `FP8_DEFAULT_CFG`'s weight-quantizer shape, but that's a minor coupling.

**Outstanding for human sign-off:**
- **Nemotron VL + `--auto_quantize_bits` still hard-fails before search.** `load_model` unconditionally sets `args.calib_with_images = True` for Nemotron VL (hf_ptq.py ~L471-473), and `auto_quantize()` then raises `NotImplementedError("AutoQuantize with image-text calibration is not supported yet.")` at L332-336. CodeRabbit marked this "✅ Addressed in commit ec996da" but the suggested guard (`and args.auto_quantize_bits is None`) is **not** in the current file. The PR's target architecture is Qwen3.5/3.6 (validated end-to-end per the PR body), so this likely doesn't break the shipped use case, but it should either be fixed or explicitly scoped out.
- **Hardcoded VLM/Qwen disable patterns in `hf_ptq.py`.** Acknowledged via TODO and inline reply chain (juhi10071998 to follow up next week). The Qwen-only patterns (`*shared_expert_gate*`, `*linear_attn.in_proj_*`) are appended for every model, but in practice these names don't appear on non-Qwen architectures, so the over-broadening is benign. Worth confirming that's acceptable until the config-system refactor lands.
- The new `get_auto_quantize_disabled_layers` / `get_auto_quantize_cost_excluded_patterns` helpers in `examples/llm_ptq/hf_ptq.py` are public-named but only used locally; minor.

Tests for the new behavior (cost-model exclusion, fused MoE weight accounting, lm_head override, parent_class skip) are present and meaningful. Recommend cjluo-nv take a final look at the Nemotron VL gating before approval.

cjluo-nv (commented):

Re-review: previous critical comments are addressed. Recommending nudge for human sign-off due to remaining surface-area items, not unresolved bugs.

**Previously-flagged criticals — all resolved:**
- :speech_balloon: `--recipe` regression: gate at `hf_ptq.py` L607 is now `if args.recipe is None and args.auto_quantize_bits is None:`. ✅
- :speech_balloon: Duplicated `_get_module_weight_numel`: helper now lives canonically in `_auto_quantize_cost.py` and is imported by `algorithms.py`. ✅
- :speech_balloon: `_match_quantizer_cfg` `parent_class` skip: pinned by `test_parent_class_scoped_entries_are_ignored_for_bare_autoquant_lookup`. ✅
- :speech_balloon: lm_head test fragility: test now appends an explicit `*lm_head*: enable=False` entry to its own recipe rather than relying on `FP8_DEFAULT_CFG` defaults. ✅
- :speech_balloon: Nemotron VL + AutoQuantize hard-fail: `parse_args()` now errors on `--auto_quantize_bits` + `--calib_with_images`, and `load_model` no longer auto-enables image calibration when `args.auto_quantize_bits` is set. New `tests/examples/llm_ptq/test_hf_ptq_args.py` regression-tests both. ✅
- :speech_balloon: Qwen-only patterns over-broadening: `_get_auto_quantize_disabled_layers` now gates `_QWEN36_AUTOQ_DISABLED_LAYERS` behind `_is_qwen_model(model)`, with a unit test that asserts they are absent for Llama. ✅

**Outstanding for human sign-off:**
- Public-API surface continues to widen: new `EXCLUDED_MODULE_NAME_PATTERNS_KEY`, `w4a16_nvfp4` qformat choice in both `model_quant.auto_quantize` docstring and `hf_ptq.py` choice list, new `disabled_layers` in `default_state_dict` (loaded from search checkpoints).
- Hardcoded `_QWEN36_AUTOQ_DISABLED_LAYERS` / `_VLM_AUTOQ_DISABLED_LAYERS` in `hf_ptq.py` still have a `TODO: Refactor into the config system` (juhi10071998 to follow up). Worth confirming the hardcoded list is acceptable for the v1 ship.
- End-to-end Qwen3.6 35B + `fp8,w4a16_nvfp4` and vLLM-loadable export were validated locally per the PR body, but there is no CI coverage for that path; recommend cjluo-nv take a final look before merge.

examples/llm_ptq/hf_ptq.py
```diff
@@ -140,6 +141,59 @@ def _set_kv_cache_constant_amax(quant_cfg: list) -> None:
 mto.enable_huggingface_checkpointing()
 
 
+# TODO: Refactor into the config system.
```
cjluo-nv: can we move model specific code to example_utils.py?

modelopt/torch/quantization/algorithms.py
```diff
@@ -54,6 +55,59 @@
 from .utils import is_quantized_linear
 
 
+def _is_fused_experts_module(module: nn.Module) -> bool:
```
cjluo-nv: why not move it to huggingface.py?

examples/llm_ptq/hf_ptq.py
meenchen: Addressed in 0cc2bef1: moved the Qwen/VLM AutoQuant model-specific helper logic out of hf_ptq.py and into example_utils.py. hf_ptq.py now imports those helpers from example_utils.

modelopt/torch/quantization/algorithms.py
meenchen: Addressed in 0cc2bef1: moved the HF quantized fused-experts wrapper check into plugins/huggingface.py. algorithms.py now calls it through a late import to avoid the existing registration import cycle.

realAsma (commented):
Why do we need the `cost_excluded_patterns`?
It seems to add complexity without real benefits.

meenchen:
> Why do we need the `cost_excluded_patterns`? It seems to add complexity without real benefits.

Without this, AutoQuant takes MTP and Vision tower into the cost function. It will be hard for users to define the effective bits ratio for the LLM part.

modelopt/torch/quantization/algorithms.py
```diff
@@ -1401,6 +1485,11 @@ def _cfg_to_dict(v):
         return v
 
     quant_cfg: list[dict] = [{"quantizer_name": "*", "enable": False}]
+    quant_cfg.extend(
+        {"quantizer_name": pattern, "enable": False}
+        for pattern in _as_list(search_state.get("disabled_layers"))
+    )
+    per_module_entries: list[dict] = []
     _per_module_attrs = ("input_quantizer", "weight_quantizer", "output_quantizer")
```
realAsma: get_auto_quantize_config() still emits only standard quantizer attrs, so fused-expert AutoQuant choices may not eplay correctly from generated configs. Can we fix it?

Yes, I think we should fix it. The issue is specifically in the replay config, not in the live searched model.

During search this PR now snapshots/restores fused experts through the real fused attrs:

- `gate_up_proj_input_quantizer`
- `gate_up_proj_weight_quantizers`
- `down_proj_input_quantizer`
- `down_proj_weight_quantizers`

But `get_auto_quantize_config()` still uses only:

```python
("input_quantizer", "weight_quantizer", "output_quantizer")
```

So if a fused expert wins a recipe, the generated config emits entries like:

```python
{"quantizer_name": "layers.N.mlp.experts.weight_quantizer", ...}
```

That will not match the actual fused-expert quantizers on replay. Their real names are shaped like `layers.N.mlp.experts.gate_up_proj_weight_quantizers.0` / `down_proj_weight_quantizers.0`. `conversion.py` normalizes those to `...gate_up_proj_weight_quantizer` / `...down_proj_weight_quantizer`, not to plain `...weight_quantizer`. Because the generated config starts with `{"quantizer_name": "*", "enable": False}`, the selected fused-expert quantizers can remain disabled/default when someone calls `mtq.get_auto_quantize_config(search_state)` and replays it through `mtq.quantize()` or an export flow.

Suggested fix: persist config-replay quantizer attrs into `candidate_stats` while the model objects are still available, then have `get_auto_quantize_config()` use those attrs, defaulting old search states to the standard trio for backward compatibility. For fused experts, the replay attrs should be something like:

```python
(
    "gate_up_proj_input_quantizer",
    "gate_up_proj_weight_quantizer",   # singular form matches all ModuleList children via conversion normalization
    "down_proj_input_quantizer",
    "down_proj_weight_quantizer",
)
```

A regression test should exercise `get_auto_quantize_config()` for a fused expert hparam and assert the generated `quant_cfg` contains the gate/up and down fused-expert quantizer names, not only `module.weight_quantizer`.

realAsma (approved):
Look great! 

I am not convinced of the benefits of `EXCLUDED_MODULE_NAME_PATTERNS_KEY` compared to the complexity it adds. - However I dont have a strong preference.

modelopt/torch/quantization/algorithms.py
meenchen: Fixed in 613c82fc. This was a real replay-path bug: live AutoQuant search/export used the fused-expert quantizer attrs correctly, but get_auto_quantize_config() could emit plain *.weight_quantizer entries for fused experts. The fix persists replayable quantizer attrs in candidate_stats and emits gate_up_proj/down_proj quantizer names, with a fallback for older search checkpoints. Added regression coverage in test_get_auto_quantize_config_emits_fused_expert_quantizer_names; focused AutoQuant/fused-expert tests and pre-commit passed.

cjluo-nv (approved):

PR #1387, opened 2026-05-04
[Quantization] Fused Triton kernel for NVFP4 FP8 scale sweep search

cjluo-nv:
## Summary

- Replaces the 126-iteration Python sweep in `NVFP4MSECalibrator` with a fused Triton kernel that loads each NVFP4 block once, evaluates all 126 valid FP8 E4M3 scale candidates in registers, and emits the per-block ``best_amax`` directly.
- Triton-backed `TritonNVFP4MSECalibrator` is the **default** for `mse_calibrate(..., fp8_scale_sweep=True)`. Set `MODELOPT_NVFP4_TRITON_SWEEP=0` to fall back to the reference for debugging or numerics comparison.
- Microbenchmark: **~42x speedup** on a B300 over the reference (`NVFP4MSECalibrator`) on a representative LLM weight (`8192x4096`, ~2M NVFP4 blocks): `176.68 ms -> 4.23 ms`.
- End-to-end on Qwen3-8B PTQ (`hf_ptq.py --qformat nvfp4_mse`, calib=128): **~6.7x faster mtq.quantize** with **identical global weight-MSE** as the reference.

## End-to-end Qwen3-8B PTQ (B300)

> Note: ran on **Qwen3-8B** instead of Qwen3.5-9B because the docker's transformers (4.57.3) doesn't yet recognize the `qwen3_5` (multimodal) architecture and the model dir doesn't ship `modeling_*.py` for `trust_remote_code`. Qwen3-8B is the same family / similar size and gives a representative comparison.

Settings: `--calib_size 128 --calib_seq 512`, default `nvfp4_mse` config (`NVFP4_W4A4_WEIGHT_MSE_FP8_SWEEP_CFG`). The first `nvfp4` run was discarded as a warm-up for HF weight loading.

| qformat | `mtq.quantize` time | global weight MSE vs FP16 orig | notes |
|---|---:|---:|---|
| `nvfp4` (no MSE search)             |  **3.46 s** | 6.363e-6 | max-calibration baseline |
| `nvfp4_mse` (reference, slow)       | **43.42 s** | 4.788e-6 | `MODELOPT_NVFP4_TRITON_SWEEP=0` |
| `nvfp4_mse` (Triton, fast — this PR)|  **6.51 s** | 4.788e-6 | default in this PR |

- The Triton path's quantized weights produce **bit-identical global weight MSE** to the reference (4.788e-6 vs 4.788e-6), validating the kernel on a real model.
- Triton `nvfp4_mse` is **6.67x faster** than the reference `nvfp4_mse`, and adds only **~3 s** over the no-MSE baseline (vs ~40 s for the reference) — making the MSE-search option nearly free in practice.
- The MSE search itself reduces weight quantization error by **~25%** vs plain max-calibration (6.363e-6 → 4.788e-6).

Per-layer MSE CSVs are produced by `tools/debugger/compare_mse_qwen.py` (one row per Linear weight) for closer inspection if needed.

## Microbenchmark (B300, 8192x4096 weight, ~2M NVFP4 blocks)

```
reference NVFP4MSECalibrator:   176.68 ms
triton  TritonNVFP4MSECalibrator: 4.23 ms
speedup: 41.8x
```

## Why this works

Each candidate is constructed as `valid_fp8_e4m3_value / 448`. With `block_amax = global_amax * candidate`, the FP8 round-trip on the per-block scale `block_amax / 6` (using `global_amax / 6` as the FP8 amax) is the **identity** — so the kernel can compute `scale = candidate * global_amax / 6.0` inline and skip the FP8 cast. This keeps the kernel runnable on any CUDA + Triton (no `tl.float8e4nv` requirement).

Because every candidate's per-block scale is just a rescaling of the same input block, all 126 candidates can be evaluated against a single `[BLOCKS_PER_PROGRAM, BLOCK_SIZE]` tile held in registers — replacing 126 weight-bandwidth passes with 1.

Two follow-on optimizations close the gap to the compute ceiling:
- `@triton.autotune` over `(BLOCKS_PER_PROGRAM, num_warps)` — the original hand-picked default (BPP=4, num_warps=4) left ~4x on the table; the best B300 config is `BPP=64, num_warps=8`.
- Drop the sign-handling `tl.where`: FP4 quant preserves sign, so `(w - w_q)^2 == (|w| - |w_q|)^2` and the kernel works on `|w|` throughout (one fewer where + negation per element per candidate).

## Files

- `modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py` — new kernel + `nvfp4_fp8_scale_sweep` wrapper, autotuned.
- `modelopt/torch/kernels/quantization/gemm/__init__.py` — wire-in.
- `modelopt/torch/quantization/calib/mse.py` — new `TritonNVFP4MSECalibrator(NVFP4MSECalibrator)`.
- `modelopt/torch/quantization/model_calib.py` — opt-out env var (`MODELOPT_NVFP4_TRITON_SWEEP=0`); Triton path is default.
- `tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py` — 16 GPU tests covering parity (across seeds, block counts, dtypes), input validation, output round-trip, reset, and a wall-clock speedup report.

## Numerics

Bit-identical to the reference for typical block counts (`{4, 64, 1024}` blocks, 3 seeds, fp32/fp16/bf16 — 14/15 microbenchmark tests bit-exact).

On multi-million-block weights an occasional adjacent-candidate tie-break can differ at the fp32-noise level (observed **2 / 2,097,152 blocks** in the speedup test, per-block MSE within ~1e-7 relative). The reference's CUDA `fake_e4m3fy` and the Triton inline math have slightly different op ordering, which lets nearly-tied candidates flip. The speedup test asserts the worst per-block MSE gap is `< 1e-5` relative on differing blocks — both choices are valid argmins; the resulting quantized weights are equally good. The Qwen3-8B end-to-end run confirms this: aggregate weight MSE matches the reference exactly at the displayed precision.

## Test plan

- [x] `pytest tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py -v` — 16/16 pass on B300
- [x] `pytest tests/gpu/torch/quantization/test_nvfp4_static_quantizer_cuda.py -v` — existing NVFP4 tests still pass (9/9)
- [x] End-to-end PTQ on Qwen3-8B with `--qformat nvfp4_mse`: 6.67x speedup, identical weight MSE to reference

## Summary by CodeRabbit

* **New Features**
  - Added a Triton-based fused NVFP4 FP8 scale-sweep for faster per-block scale selection.
  - Exposed a device-aware FP8 scale candidate generator and a GPU sweep API used by the calibrator.
  - Calibrator now uses the Triton fast-path by default with env var opt-out (MODELOPT_NVFP4_TRITON_SWEEP).

* **Bug Fixes / Behavior**
  - One-shot fast-path enforcement in the NVFP4 calibrator; reset restores reuse.

* **Tests**
  - Added comprehensive GPU tests validating parity, input validation, dispatch behavior, one-shot semantics, and speed benchmarks.

### DIFF
--- a/modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
+++ b/modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
@@ -113,7 +113,6 @@ def nvfp4_fp8_scale_sweep(
     x: torch.Tensor,
     global_amax: torch.Tensor,
     block_size: int = 16,
-    candidates: torch.Tensor | None = None,
 ) -> torch.Tensor:
     """Find the per-block FP8 scale that minimizes NVFP4 quantization MSE.
 
@@ -126,8 +125,6 @@ def nvfp4_fp8_scale_sweep(
             ``block_size``; layout is treated as a flat ``[N_BLOCKS, BLOCK_SIZE]``.
         global_amax: Scalar FP32 global amax (``= reduce_amax(per_block_amax)``).
         block_size: NVFP4 block size (typically 16).
-        candidates: Optional precomputed candidate tensor of shape ``[126]`` (must
-            be the FP8 E4M3 valid values divided by 448). Built lazily if omitted.
 
     Returns:
         ``best_amax`` of shape ``[N_BLOCKS]``, fp32, on the same device as ``x``.
@@ -139,13 +136,7 @@ def nvfp4_fp8_scale_sweep(
     if x.numel() % block_size != 0:
         raise ValueError(f"x.numel() ({x.numel()}) is not divisible by block_size ({block_size}).")
 
-    if candidates is None:
-        candidates = fp8_scale_candidates(x.device)
-    candidates = candidates.contiguous().to(device=x.device, dtype=torch.float32)
-    if candidates.ndim != 1 or candidates.numel() == 0:
-        raise ValueError(
-            f"candidates must be a non-empty 1-D tensor; got shape {tuple(candidates.shape)}."
-        )
+    candidates = fp8_scale_candidates(x.device).to(dtype=torch.float32)
 
     n_blocks = x.numel() // block_size
     x_flat = x.contiguous().view(-1)
--- a/modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
+++ b/modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
@@ -24,13 +24,15 @@ The 126 candidates are constructed as ``valid_fp8_e4m3_value / 448`` (see
 the per-block scale is the identity, so the kernel can use
 ``scale = candidate * global_amax / 6.0`` without an explicit FP8 cast — making it
 runnable on any CUDA GPU with Triton (no ``tl.float8e4nv`` requirement).
+
+Tile shape (``BLOCKS_PER_PROGRAM``) and ``num_warps`` are autotuned per ``N_BLOCKS``.
 """
 
 import torch
 import triton
 import triton.language as tl
 
-from .nvfp4_quant import nvfp4_scalar_quant
+from .nvfp4_quant import fp4_round_magnitude
 
 __all__ = ["fp8_scale_candidates", "nvfp4_fp8_scale_sweep"]
 
@@ -43,6 +45,18 @@ def fp8_scale_candidates(device: torch.device | str = "cpu") -> torch.Tensor:
     return fp8_values[valid_mask] / 448.0
 
 
+# Selected from a (BLOCKS_PER_PROGRAM, num_warps) sweep on B300:
+#   BPP=16,nw=2: 6.06 ms   BPP=32,nw=4: 6.06 ms   BPP=64,nw=8: 5.08 ms
+# The smaller-tile entries cover cases where N_BLOCKS is small enough that BPP=64
+# would underfill the SMs.
+_FP8_SWEEP_AUTOTUNE_CONFIGS = [
+    triton.Config({"BLOCKS_PER_PROGRAM": 16}, num_warps=2),
+    triton.Config({"BLOCKS_PER_PROGRAM": 32}, num_warps=4),
+    triton.Config({"BLOCKS_PER_PROGRAM": 64}, num_warps=8),
+]
+
+
+@triton.autotune(configs=_FP8_SWEEP_AUTOTUNE_CONFIGS, key=["N_BLOCKS"])
 @triton.jit
 def _fp8_scale_sweep_kernel(
     x_ptr,  # [N_BLOCKS * BLOCK_SIZE], any float dtype (loaded as fp32)
@@ -59,10 +73,13 @@ def _fp8_scale_sweep_kernel(
     block_idx = block_start + tl.arange(0, BLOCKS_PER_PROGRAM)
     block_mask = block_idx < N_BLOCKS
 
-    # Load weights for this tile: [BLOCKS_PER_PROGRAM, BLOCK_SIZE]
+    # Load weights for this tile and pre-compute their absolute values once.
+    # The squared error is sign-invariant since FP4 quant preserves sign:
+    #   (w - w_q)^2 = (|w| - |w_q|)^2 = (|w| - q_mag * scale)^2
+    # so we never need ``w`` itself again, dropping a tl.where + negation per element.
     elem_offs = block_idx[:, None] * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)[None, :]
     elem_mask = block_mask[:, None]
-    w = tl.load(x_ptr + elem_offs, mask=elem_mask, other=0.0).to(tl.float32)
+    w_abs = tl.abs(tl.load(x_ptr + elem_offs, mask=elem_mask, other=0.0).to(tl.float32))
 
     global_amax = tl.load(global_amax_ptr).to(tl.float32)
 
@@ -70,13 +87,17 @@ def _fp8_scale_sweep_kernel(
     best_idx = tl.zeros([BLOCKS_PER_PROGRAM], dtype=tl.int32)
 
     # Loop over the 126 FP8 candidates (compile-time unrolled).
+    # Scales are guaranteed positive and finite (constructed from a positive candidate
+    # times nonneg global_amax), so the degenerate-scale guard from nvfp4_scalar_quant is
+    # unnecessary apart from the global_amax == 0 case handled below.
     for k in tl.static_range(NUM_CANDIDATES):
         c = tl.load(candidates_ptr + k).to(tl.float32)
-        # block_amax = global_amax * c by construction; the FP8 round on the resulting
-        # scale is the identity for our candidate set, so we can skip the FP8 cast.
         scale = c * global_amax / 6.0
-        w_q = nvfp4_scalar_quant(w, scale, BLOCK_SIZE)
-        diff = w - w_q
+        # Avoid divide-by-zero when global_amax == 0; the resulting err == w_abs² is
+        # the same for every candidate, so any best_idx is fine.
+        scale_safe = tl.where(scale == 0.0, 1.0, scale)
+        q_mag = fp4_round_magnitude(w_abs / scale_safe)
+        diff = w_abs - q_mag * scale
         loss = tl.sum(diff * diff, axis=1)  # [BLOCKS_PER_PROGRAM]
         is_better = loss < best_loss
         best_loss = tl.where(is_better, loss, best_loss)
@@ -93,7 +114,6 @@ def nvfp4_fp8_scale_sweep(
     global_amax: torch.Tensor,
     block_size: int = 16,
     candidates: torch.Tensor | None = None,
-    blocks_per_program: int = 4,
 ) -> torch.Tensor:
     """Find the per-block FP8 scale that minimizes NVFP4 quantization MSE.
 
@@ -108,8 +128,6 @@ def nvfp4_fp8_scale_sweep(
         block_size: NVFP4 block size (typically 16).
         candidates: Optional precomputed candidate tensor of shape ``[126]`` (must
             be the FP8 E4M3 valid values divided by 448). Built lazily if omitted.
-        blocks_per_program: Number of blocks each Triton program handles. Trades
-            launch overhead for register pressure; 4 is a reasonable default.
 
     Returns:
         ``best_amax`` of shape ``[N_BLOCKS]``, fp32, on the same device as ``x``.
@@ -127,7 +145,7 @@ def nvfp4_fp8_scale_sweep(
     global_amax_f32 = global_amax.detach().to(device=x.device, dtype=torch.float32).reshape(1)
     best_amax = torch.empty(n_blocks, dtype=torch.float32, device=x.device)
 
-    grid = (triton.cdiv(n_blocks, blocks_per_program),)
+    grid = lambda meta: (triton.cdiv(n_blocks, meta["BLOCKS_PER_PROGRAM"]),)
     with torch.cuda.device(x.device):
         _fp8_scale_sweep_kernel[grid](
             x_flat,
@@ -137,6 +155,5 @@ def nvfp4_fp8_scale_sweep(
             n_blocks,
             BLOCK_SIZE=block_size,
             NUM_CANDIDATES=int(candidates.numel()),
-            BLOCKS_PER_PROGRAM=blocks_per_program,
         )
     return best_amax
--- a/modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
+++ b/modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
@@ -1,0 +1,142 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Fused Triton kernel for the NVFP4 weight-MSE FP8 scale sweep.
+
+Replaces the 126-iteration Python sweep in :class:`NVFP4MSECalibrator` with a single
+kernel that, for each NVFP4 block, evaluates all 126 valid FP8 E4M3 scale candidates
+and emits the per-block ``best_amax`` directly.
+
+The 126 candidates are constructed as ``valid_fp8_e4m3_value / 448`` (see
+:func:`fp8_scale_candidates`). For these specific candidates, the FP8 round-trip on
+the per-block scale is the identity, so the kernel can use
+``scale = candidate * global_amax / 6.0`` without an explicit FP8 cast — making it
+runnable on any CUDA GPU with Triton (no ``tl.float8e4nv`` requirement).
+"""
+
+import torch
+import triton
+import triton.language as tl
+
+from .nvfp4_quant import nvfp4_scalar_quant
+
+__all__ = ["fp8_scale_candidates", "nvfp4_fp8_scale_sweep"]
+
+
+def fp8_scale_candidates(device: torch.device | str = "cpu") -> torch.Tensor:
+    """Return the 126 valid finite positive FP8 E4M3 scale candidates / 448."""
+    uint8_values = torch.arange(0, 128, dtype=torch.uint8, device=device)
+    fp8_values = uint8_values.view(torch.float8_e4m3fn).float()
+    valid_mask = torch.isfinite(fp8_values) & (fp8_values > 0)
+    return fp8_values[valid_mask] / 448.0
+
+
+@triton.jit
+def _fp8_scale_sweep_kernel(
+    x_ptr,  # [N_BLOCKS * BLOCK_SIZE], any float dtype (loaded as fp32)
+    candidates_ptr,  # [NUM_CANDIDATES] fp32
+    global_amax_ptr,  # scalar fp32
+    best_amax_ptr,  # [N_BLOCKS] fp32 output
+    N_BLOCKS,
+    BLOCK_SIZE: tl.constexpr,
+    NUM_CANDIDATES: tl.constexpr,
+    BLOCKS_PER_PROGRAM: tl.constexpr,
+):
+    pid = tl.program_id(axis=0)
+    block_start = pid * BLOCKS_PER_PROGRAM
+    block_idx = block_start + tl.arange(0, BLOCKS_PER_PROGRAM)
+    block_mask = block_idx < N_BLOCKS
+
+    # Load weights for this tile: [BLOCKS_PER_PROGRAM, BLOCK_SIZE]
+    elem_offs = block_idx[:, None] * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)[None, :]
+    elem_mask = block_mask[:, None]
+    w = tl.load(x_ptr + elem_offs, mask=elem_mask, other=0.0).to(tl.float32)
+
+    global_amax = tl.load(global_amax_ptr).to(tl.float32)
+
+    best_loss = tl.full([BLOCKS_PER_PROGRAM], float("inf"), dtype=tl.float32)
+    best_idx = tl.zeros([BLOCKS_PER_PROGRAM], dtype=tl.int32)
+
+    # Loop over the 126 FP8 candidates (compile-time unrolled).
+    for k in tl.static_range(NUM_CANDIDATES):
+        c = tl.load(candidates_ptr + k).to(tl.float32)
+        # block_amax = global_amax * c by construction; the FP8 round on the resulting
+        # scale is the identity for our candidate set, so we can skip the FP8 cast.
+        scale = c * global_amax / 6.0
+        w_q = nvfp4_scalar_quant(w, scale, BLOCK_SIZE)
+        diff = w - w_q
+        loss = tl.sum(diff * diff, axis=1)  # [BLOCKS_PER_PROGRAM]
+        is_better = loss < best_loss
+        best_loss = tl.where(is_better, loss, best_loss)
+        best_idx = tl.where(is_better, k, best_idx)
+
+    # Map each block's winning candidate index back to its amax = global_amax * c[best].
+    best_c = tl.load(candidates_ptr + best_idx, mask=block_mask, other=0.0).to(tl.float32)
+    best_amax = global_amax * best_c
+    tl.store(best_amax_ptr + block_idx, best_amax, mask=block_mask)
+
+
+def nvfp4_fp8_scale_sweep(
+    x: torch.Tensor,
+    global_amax: torch.Tensor,
+    block_size: int = 16,
+    candidates: torch.Tensor | None = None,
+    blocks_per_program: int = 4,
+) -> torch.Tensor:
+    """Find the per-block FP8 scale that minimizes NVFP4 quantization MSE.
+
+    Equivalent to the 126-step sweep in :class:`NVFP4MSECalibrator`, but fused into
+    a single Triton kernel: every block's weight elements are loaded once, all 126
+    candidates are evaluated in registers, and the running argmin is kept inline.
+
+    Args:
+        x: Weight tensor on CUDA. Total element count must be divisible by
+            ``block_size``; layout is treated as a flat ``[N_BLOCKS, BLOCK_SIZE]``.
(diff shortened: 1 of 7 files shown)


modelopt/torch/quantization/calib/mse.py
```diff
@@ -198,3 +198,82 @@ def _generate_candidates(self, device: torch.device) -> torch.Tensor:
...
+            raise RuntimeError(
+                "TritonNVFP4MSECalibrator only supports a single collect() per cycle; "
+                "call reset() before collecting again."
+            )
+
+        x = x.detach()
+        block_size = x.shape[-1]
+        n_blocks = x.numel() // block_size
+        if self._initial_amax.numel() != n_blocks:
+            raise ValueError(
```
meenchen: > _Bot comment._

Nit: `block_size = x.shape[-1]` assumes the input tensor has already been reshaped to `[n_blocks, block_size]`. This is true for the current `mse_calibrate` weight flow, but could silently produce wrong results if someone uses this calibrator with a differently-shaped tensor. Consider adding a brief assertion or docstring note, e.g.:
```python
assert x.ndim == 2, "Expected x to be [n_blocks, block_size] from the weight quantizer reshape"
```

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
```diff
@@ -0,0 +1,142 @@
...
+import triton
+import triton.language as tl
+
+from .nvfp4_quant import nvfp4_scalar_quant
+
+__all__ = ["fp8_scale_candidates", "nvfp4_fp8_scale_sweep"]
+
+
+def fp8_scale_candidates(device: torch.device | str = "cpu") -> torch.Tensor:
+    """Return the 126 valid finite positive FP8 E4M3 scale candidates / 448."""
```
meenchen: > _Bot comment._

Minor duplication: this function reproduces the same logic as `NVFP4MSECalibrator._generate_candidates()` in `calib/mse.py`. Consider having one call the other (or extracting a shared utility) so the candidate generation stays in sync if the candidate set ever changes.

modelopt/torch/quantization/model_calib.py
```diff
@@ -391,8 +392,13 @@ def mse_calibrate(
                         continue
 
                 if fp8_scale_sweep and is_nvfp4_static:
-                    # Replace calibrator with NVFP4MSECalibrator
-                    module._calibrator = NVFP4MSECalibrator(
+                    # Replace calibrator with the fused Triton sweep kernel by default
+                    # (single-shot collect, ~7-20x faster for the weight-MSE phase).
```
meenchen: > _Bot comment._

The env var check `os.environ.get("MODELOPT_NVFP4_TRITON_SWEEP", "1") != "0"` is evaluated on every weight quantizer in the loop. Since it won't change mid-loop, consider hoisting it above the loop for clarity and minor efficiency.

meenchen (commented):

Clean, well-structured Triton kernel for speeding up the NVFP4 FP8 scale sweep. The implementation correctly reuses the existing `nvfp4_scalar_quant` JIT function from `nvfp4_quant.py`, the math insight about the FP8 round-trip identity for the candidate set is sound, and test coverage is solid (15 GPU tests covering parity, dtypes, round-trip, reset, and speedup).

A few points for the owner:

1. **Unchecked test plan items**: The PR body has two unchecked items — H100/A100 run and end-to-end PTQ on a 70B model. Per project norms, these should be completed before merge.

2. **Minor code duplication**: `fp8_scale_candidates()` in `nvfp4_fp8_sweep.py` duplicates `NVFP4MSECalibrator._generate_candidates()`. Consider having one call the other (or extracting a shared utility) to keep the candidate generation logic in one place.

3. **`local_hessian_calibrate` not using the Triton path**: This function still uses `NVFP4MSECalibrator` directly (not `TritonNVFP4MSECalibrator`), which is correct since it needs a custom `error_func`. Worth adding a comment there noting that the Triton path doesn't support custom error functions, so someone doesn't "helpfully" switch it later.

4. **`collect` assumes `x.shape[-1]` is block_size**: This works for the current MSE weight calibration flow where the tensor is pre-reshaped, but could be fragile if the calibrator is used in a different context. A brief assert or docstring note would help.

meenchen (commented):

All critical issues from the previous review have been addressed:

- ✅ `assert x.is_cuda` → `raise ValueError`
- ✅ `block_size <= 0` validation added
- ✅ Custom `candidates` tensor validation added
- ✅ `reset()` reusability fixed (metadata stashed in `__init__`, tested in `test_reset_allows_recollect`)
- ✅ Env var hoisted above the loop
- ✅ `x.ndim == 2` assertion added in `collect()`
- ✅ Candidate duplication addressed with sync docstring in `_generate_candidates`

The kernel implementation is clean, the math insight about the FP8 round-trip identity is sound, and test coverage is solid (15 GPU tests covering parity, dtypes, round-trip, reset, input validation, and speedup).

**Remaining concern**: The PR body still has two unchecked test plan items — H100/A100 validation and end-to-end PTQ on a 70B model. Per project norms, these should be completed before merge. Nudging for human sign-off on that.

meenchen (approved):

All critical issues from previous reviews have been addressed:

- `assert x.is_cuda` replaced with `raise ValueError`
- `block_size <= 0` validation added before use
- Custom `candidates` tensor validation (non-empty, 1-D) added
- `reset()` reusability fixed — shape/dtype/n_blocks metadata stashed in `__init__`, verified by `test_reset_allows_recollect`
- Env var read hoisted above the weight quantizer loop
- `x.ndim == 2` shape check added in `collect()`
- Candidate duplication addressed with sync docstring

The kernel implementation is correct — the FP8 round-trip identity insight is sound, input validation is thorough, and test coverage is solid (16 GPU tests covering parity across seeds/dtypes/block counts, round-trip, reset, input validation, and speedup reporting). License headers match the canonical `LICENSE_HEADER`.

meenchen (commented):
LGTM. @Fridah-nv could you also help review this PR?

modelopt/torch/quantization/calib/mse.py
```diff
@@ -192,9 +192,100 @@ def _compute_candidate_amax(self, candidates: torch.Tensor) -> torch.Tensor:
...
+
+    @torch.no_grad()
+    def compute_amax(self, verbose: bool = False):
+        """Return the per-block amax computed during ``collect``."""
+        return self._best_amax
+
+    def reset(self):
+        """Reset the stored best amax. Subsequent ``collect`` calls are allowed."""
+        self._best_amax = None
+        super().reset()
```
realAsma: same as https://github.com/NVIDIA/Model-Optimizer/pull/1387/changes#r3192226670

modelopt/torch/quantization/calib/mse.py
```diff
@@ -192,9 +192,100 @@ def _compute_candidate_amax(self, candidates: torch.Tensor) -> torch.Tensor:
...
+            axis=axis,
+            quant_func=quant_func,
+            error_func=error_func,
+        )
+        # Stash shape metadata so collect() can keep working after reset() releases
+        # the (potentially large) _initial_amax buffer.
+        self._initial_amax_shape = tuple(amax.shape)
+        self._initial_amax_dtype = amax.dtype
+        self._n_blocks = int(amax.numel())
+        self._best_amax: torch.Tensor | None = None
```
realAsma: why do we need this? why not inherit from parent class?

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
```diff
@@ -0,0 +1,166 @@
...
+    for k in tl.static_range(NUM_CANDIDATES):
+        c = tl.load(candidates_ptr + k).to(tl.float32)
+        scale = c * global_amax / 6.0
+        # Avoid divide-by-zero when global_amax == 0; the resulting err == w_abs² is
+        # the same for every candidate, so any best_idx is fine.
+        scale_safe = tl.where(scale == 0.0, 1.0, scale)
+        q_mag = fp4_round_magnitude(w_abs / scale_safe)
+        diff = w_abs - q_mag * scale
+        loss = tl.sum(diff * diff, axis=1)  # [BLOCKS_PER_PROGRAM]
+        is_better = loss < best_loss
```
realAsma, line 95: This uses the minimum scale candidate for tie breaking which multiple scales produce the same loss. 

We are discussing a better alternative where we use the median scale for tie breaking.

What is your thought?
Cc @jenchen13  @Fridah-nv

realAsma:
> _Bot comment._

Reviewed current head `bd4fc3a`. I did not find additional blocking correctness issues beyond the active threads.

Remaining items I would resolve before merge:

- The unresolved `collect()` input-validation comment is still valid: `assert x.ndim == 2` should be explicit `ValueError` validation so the check is not stripped under optimized Python, and `x.shape[-1]` should be checked before division.
- The `TritonNVFP4MSECalibrator` inheritance works for the documented one-shot weight-MSE flow, but it is not a clean subtype of `NVFP4MSECalibrator`; a shared helper/base with sibling reference/Triton calibrators would be cleaner if we want this as the default exported path.
- Public CI currently has `regression-pr-required-check` failing. The `regression-tests` step itself completed successfully, but the job was canceled during Codecov upload, so the required check still needs to be rerun or otherwise cleared before merge.

realAsma:
> _Bot comment._

Notes on the active review threads:

- On the `TritonNVFP4MSECalibrator` metadata/reset question: the extra shape/dtype/block-count metadata is needed with the current implementation because `MseCalibrator.reset()` intentionally deletes `_initial_amax`, while this Triton path still needs those values after reset to reshape the fused-kernel result and validate later `collect()` calls. That said, this is not a clean behavioral subtype of `NVFP4MSECalibrator`; if this remains exported and selected by default, I would prefer a small shared NVFP4 sweep helper/base with the reference and Triton calibrators as sibling implementations.
- On the scale tie-breaking question: for this acceleration PR I would keep the behavior reference-compatible. `loss < best_loss` with ascending candidates selects the first/minimum candidate on exact ties, which matches the existing reference path's `torch.argmin` over the same ordered candidate list. Median-scale tie-breaking may be a valid quantization policy change, but I would make it deliberately in both the reference and Triton paths with tests/docs rather than folding it into the default-on kernel replacement.

realAsma:
> _Bot comment._

Additional review notes from the current head `bd4fc3a`:

- `nvfp4_fp8_scale_sweep()` still exposes `candidates` as a public override, but the docstring says it must be the fixed 126-entry FP8 E4M3 candidate set. The wrapper currently only checks non-empty 1-D, and `candidates.numel()` becomes the constexpr for the unrolled `tl.static_range`, so wrong-length input can turn a caller mistake into excessive Triton codegen, while nonpositive or nonfinite entries violate the kernel's scale assumptions and can emit invalid amax values. I would either remove the override or validate the full contract before launch: exactly 126 entries, finite, positive, and ideally equal to `fp8_scale_candidates(...)` if this path is meant to stay reference-compatible. Please add coverage for wrong length and nonpositive/nonfinite candidates if the override remains.
- Small test gap: the new default/opt-out wiring in `mse_calibrate(..., fp8_scale_sweep=True)` is not directly covered. A focused test that the default installs `TritonNVFP4MSECalibrator` for NVFP4 static weights, and that `MODELOPT_NVFP4_TRITON_SWEEP=0` falls back to `NVFP4MSECalibrator`, would protect this public behavior without depending only on the lower-level kernel tests.

modelopt/torch/quantization/calib/mse.py
cjluo-nv: Agreed on the design call-out. The shape/dtype/n_blocks stash in `__init__` is necessary today only because `MseCalibrator.reset()` deletes `_initial_amax`, and the Triton path still needs that metadata to reshape the kernel's flat output and validate later `collect()` calls — so I'd rather not strip it without the larger refactor.

The cleaner shape — a small NVFP4 sweep base/helper with the reference and Triton calibrators as siblings off it — I'd like to do as a follow-up so this PR stays scoped to "port the sweep to Triton." Will file a follow-up issue.

Per-thread items addressed in the latest commit:
- `collect()` now raises `ValueError` instead of `assert`, and validates `x.shape[-1]` before using it.
- Dropped the public `candidates` override in `nvfp4_fp8_scale_sweep` (no internal callers used it; wrong-length input would have silently inflated the unrolled `tl.static_range` codegen).
- Added `test_mse_calibrate_dispatch` covering the default-on + `MODELOPT_NVFP4_TRITON_SWEEP=0` fallback wiring.

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
cjluo-nv: Per your follow-up note: keeping `loss < best_loss` (min-tie-break) here matches the reference path's `torch.argmin` over the same ascending candidate list, so this PR stays reference-compatible.

Median-tie-break is worth doing, but as a deliberate change applied to **both** the reference and Triton paths together with tests/docs — happy to take that as a separate PR. Resolving here.

cjluo-nv:
@realAsma — thanks for the thorough review. Punch-list against your three top-level comments, addressed in commit e57c7a87b:

**Fixed in this PR**

- `TritonNVFP4MSECalibrator.collect()` input validation: replaced `assert x.ndim == 2` with `ValueError` so the contract holds under `python -O`, validate `block_size > 0` before use, and switched `n_blocks = x.numel() // block_size` to `x.shape[0]` so a zero last-dim cannot fault before the shape check.
- `nvfp4_fp8_scale_sweep()` `candidates` contract: dropped the public `candidates` override. The candidate set is fixed (FP8 E4M3 valid values / 448), no internal caller used the override, and a wrong length would have silently inflated the unrolled `tl.static_range` codegen while nonpositive/nonfinite entries violated the kernel's scale assumptions. Removed the now-unreachable validation cases from the kernel input-validation test.
- Wiring test for `mse_calibrate(..., fp8_scale_sweep=True)`: added `test_mse_calibrate_dispatch` in `tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py` that quantizes a `SimpleLinear` under the NVFP4-static + `mse`/`fp8_scale_sweep=True` config and asserts (a) default installs `TritonNVFP4MSECalibrator` and (b) `MODELOPT_NVFP4_TRITON_SWEEP=0` falls back to `NVFP4MSECalibrator` (exact class match, since the Triton class is a subclass).
- Tie-break thread resolved per your follow-up: `loss < best_loss` matches the reference path's `torch.argmin` over the same ascending candidate list, so this PR stays reference-compatible.

**Deferring to a follow-up**

- The `TritonNVFP4MSECalibrator`-as-clean-subtype concern: I agree the right shape is a small NVFP4 sweep base/helper with reference and Triton calibrators as siblings off it. I'd like to do that as a separate PR so this one stays scoped to "port the sweep to Triton" — will file a follow-up issue. The metadata stash in `__init__` is necessary today only because `MseCalibrator.reset()` deletes `_initial_amax` and the Triton path still needs that metadata to reshape the kernel's flat output and validate later `collect()` calls — keeping it as-is until the sibling refactor.
- Median-tie-break: agreed worth doing, but as a deliberate change applied to **both** the reference and Triton paths together with tests/docs, in a separate PR.

**CI**

- `regression-pr-required-check`: the `regression-tests` step itself succeeded; the job was canceled during Codecov upload. Will re-run once CI picks up the new commit.

realAsma:
> _Bot comment._

Thanks for adding the Triton sweep path here. One alternate design that I think is worth considering is to keep `NVFP4MSECalibrator` as the only public calibrator and make Triton an internal fast path for the plain NVFP4 weight-MSE FP8 scale sweep, rather than exposing a separate `TritonNVFP4MSECalibrator` up to `model_calib.py`.

That would keep the higher-level calibration wiring focused on calibration intent instead of kernel selection, while still allowing the fast path when it is valid. Concretely, `mse_calibrate()` could always instantiate `NVFP4MSECalibrator` for NVFP4 static `fp8_scale_sweep`, and `NVFP4MSECalibrator.collect()` could choose the Triton path only when Triton is enabled, `error_func is None`, the input is CUDA, and the shape matches the expected blocked layout. Otherwise it would fall back to the existing reference MSE sweep automatically.

I like this boundary because custom `error_func` users, including local Hessian calibration, continue to get the reference behavior without higher-level special casing; the env opt-out remains an implementation detail; and future changes to the kernel wrapper stay localized inside the NVFP4 calibrator. Tests could then cover default Triton/reference parity, env opt-out fallback, custom `error_func` fallback, and a focused local-Hessian regression.

cjluo-nv:
@realAsma — adopted this design in 95b8a956a. Outcome:

**Refactor**

- Removed `TritonNVFP4MSECalibrator` from the public surface entirely.
- `NVFP4MSECalibrator.collect()` now selects the fused Triton kernel via an internal predicate `_can_use_triton_fast_path(x)` that requires:
  - `error_func is None`,
  - `x.is_cuda`,
  - `x.ndim == 2` and `x.shape[0] == _initial_amax.numel()` (blocked layout matches the per-block amax),
  - the kernel package is importable,
  - `MODELOPT_NVFP4_TRITON_SWEEP \!= "0"`.

  Otherwise it delegates to the parent's reference 126-step sweep. Multi-collect after the fast path raises `RuntimeError` with a message pointing the caller at `reset()`, the env opt-out, or `error_func`.
- Custom `error_func` users (including the local-Hessian path) now naturally get the reference behavior — this fixes a latent correctness issue: the old `TritonNVFP4MSECalibrator` accepted `error_func` "for API parity" but the kernel always used squared error, so a Hessian-weighted loss would have been silently ignored.
- `mse_calibrate()` always instantiates `NVFP4MSECalibrator`; the env var lookup at that layer is gone — kernel selection is now an implementation detail of the calibrator, not a higher-level switch.
- Override `NVFP4MSECalibrator.reset()` to clear only per-cycle state and keep `_initial_amax` (shape `[num_blocks]`, small) so the same instance can be re-used after `reset()`.

**Tests**

- `_force_sweep_path(triton_enabled=…)` context manager pins the dispatch via the env var around each test.
- New dispatch tests assert the predicate behavior for the four cases you listed: default fast path, env opt-out fallback, custom-`error_func` fallback, and CPU input fallback (asserted via the predicate directly, since NVFP4 fake-quant itself is CUDA-only).
- New end-to-end `test_mse_calibrate_end_to_end` runs `mtq.quantize` with the `mse`/`fp8_scale_sweep=True` config under both default and `MODELOPT_NVFP4_TRITON_SWEEP=0`, then asserts bitwise-identical post-calibration model outputs.
- Reuse-after-reset is covered by `test_reset_allows_recollect` (now passes with the new `reset()` semantics).

**Validation on B300**

- Full sweep test file: 21 passed in 8.61s. Triton fast path measured at **40.7×** vs reference (173.03 ms → 4.25 ms on a (8192, 4096) weight, ~2M NVFP4 blocks).
- Integration: `test_quantize` for `NVFP4_WEIGHT_ACT_MSE_CFG`, `NVFP4_WEIGHT_MSE_FP8_SWEEP_CFG`, and `NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG` (which exercises the `error_func` fallback path) plus `test_save_restore` — 8 passed in 2:04.

Pre-existing items that are unaffected by this commit:
- The earlier sibling-base-class follow-up is now obsolete — there are no siblings to share a base.
- The `regression-pr-required-check` Codecov-cancellation will need a re-run once CI picks up `95b8a956a`.

modelopt/torch/quantization/calib/mse.py
```diff
@@ -185,16 +194,91 @@ def __init__(
...
+        self._best_amax_fast: torch.Tensor | None = None
 
     def _compute_candidate_amax(self, candidates: torch.Tensor) -> torch.Tensor:
         if candidates.ndim != 0:  # Called during final compute amax
             candidates = candidates.view_as(self._initial_amax)
         return torch.ones_like(self._initial_amax) * self._global_amax * candidates
 
     def _generate_candidates(self, device: torch.device) -> torch.Tensor:
-        """Generate 126 valid FP8 E4M3 scale candidates."""
+        """Generate 126 valid FP8 E4M3 scale candidates.
```
jenchen13: can the triton kernel be an optional arg? We are seeing bugs in the MSE FP8 sweep path that @Fridah-nv is fixing, and completely replacing the current sweep is risky. We can keep triton as an option (via quant config arg) while we test its confidence on more model types (you've only tested 1 qwen 8b model)

modelopt/torch/quantization/calib/mse.py
```diff
@@ -185,16 +194,91 @@ def __init__(
...
+
+        The kernel produces the final per-block amax in one shot, so it's only usable
+        when the caller wants the standard squared-error sweep on a single CUDA tensor
+        whose layout already matches the per-block amax.
+        """
+        if self._error_func is not None:
+            return False
+        if not x.is_cuda:
+            return False
+        if os.environ.get("MODELOPT_NVFP4_TRITON_SWEEP", "1") == "0":
```
realAsma, line 224: @jenchen13 see here ->

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
realAsma: sounds good.

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
```diff
@@ -0,0 +1,157 @@
...
+    # times nonneg global_amax), so the degenerate-scale guard from nvfp4_scalar_quant is
+    # unnecessary apart from the global_amax == 0 case handled below.
+    for k in tl.static_range(NUM_CANDIDATES):
+        c = tl.load(candidates_ptr + k).to(tl.float32)
+        scale = c * global_amax / 6.0
+        # Avoid divide-by-zero when global_amax == 0; the resulting err == w_abs² is
+        # the same for every candidate, so any best_idx is fine.
+        scale_safe = tl.where(scale == 0.0, 1.0, scale)
+        q_mag = fp4_round_magnitude(w_abs / scale_safe)
+        diff = w_abs - q_mag * scale
```
realAsma: Can we use the same scale here please?
```suggestion
        diff = w_abs - q_mag * scale_safe
```

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
realAsma: +1, valid comment

tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py
```diff
@@ -0,0 +1,403 @@
...
+    with _force_sweep_path(triton_enabled=True):
+        cal = _make_calibrator(per_block_amax, global_amax)
+        cal.collect(x)
+        return cal.compute_amax()
+
+
+@requires_triton
+@pytest.mark.parametrize("seed", [0, 1, 2])
+@pytest.mark.parametrize("num_blocks", [4, 64, 1024])
+def test_parity_random_weights(seed, num_blocks):
```
realAsma: This test currently tests only fp32 type. can we extend this test for bf16 and fp16 as well? This is an important test.

realAsma (approved):
Looks great! 

I have left a few comments. Could you please address them as well?

modelopt/torch/quantization/calib/mse.py
cjluo-nv: @jenchen13 I will make the triton as the default. The two path has unittest covering and make sure they produces the same scales.

tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py
cjluo-nv: Done in a9c8ccf16. `test_parity_random_weights` is now parametrized on `dtype ∈ {fp32, fp16, bf16}` in addition to seed and num_blocks, so the canonical 3 seeds × 3 num_blocks grid is exercised on every supported dtype. Folded the smaller `test_parity_dtypes` into this since it was a strict subset.

Note on validation: the host I had available for this round (single-GPU B200) has unusually slow Triton compile (~9.5s cold per kernel signature vs ~50ms on the B300 used previously), so the full test file timed out at 600s on this host. The kernel itself runs correctly (verified via direct call: 9.55s cold → 0.7ms cached, correct shape/dtype out). Pre-commit + ruff + mypy clean. Will rely on CI for the full sweep validation.

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
cjluo-nv: Done in a9c8ccf16. Equivalent on real inputs (the only case where `scale_safe` differs from `scale` is `global_amax == 0`, but in that case `w_abs` is also zero by construction — `global_amax = max|w|` — so the loss is zero for every candidate either way), but more consistent with the divisor used to compute `q_mag`.

modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py
cjluo-nv: Done in a9c8ccf16. Extracted `fp8_scale_candidates` to a triton-free module `modelopt/torch/kernels/quantization/gemm/_fp8_scale_candidates.py`. Both the kernel wrapper and `NVFP4MSECalibrator._generate_candidates` now import from it, so there's a single source of truth.

PR #1389, opened 2026-05-04
[specdec_bench] Stratify --num_requests across categories

milesial:
When the dataset has a `category` column with >1 distinct categories and ``--num_requests N`` is below the dataset size, take ceil(N / num_categories) rows from each category and round-robin interleave them so any prefix is balanced. Falls back to the existing ``range(N)`` slice when category metadata is absent or there's only one category.

Fixes a sampling bug where SPEED-Bench parquet files are sorted by category, so e.g. ``--num_requests 64`` on throughput_8k pulls 64 high_entropy prompts and zero from low_entropy / mixed.

## Summary by CodeRabbit

* **New Features**
  * SPEEDBench dataset truncation now performs deterministic, balanced stratified sampling across categories (round-robin) when multiple categories exist and a smaller sample size is requested.
  * Falls back to simple prefix selection when no category or only one category is present, preserving prior behavior in that case.

### DIFF
--- a/examples/specdec_bench/specdec_bench/datasets/speed.py
+++ b/examples/specdec_bench/specdec_bench/datasets/speed.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 # mypy: disable-error-code="index"
-import math
 import random
 import re
 from enum import Enum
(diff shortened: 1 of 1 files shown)


h-guo18 (approved):
Please take a look at CodeRabbit’s comment above. Other than that LGTM

kevalmorabia97 (approved):

PR #1398, opened 2026-05-06
[6034518] Fix Conv->Relu->Concat Q/DQ insertion gap

willg-nv:
Models with Conv->Relu->Concat patterns were not receiving Q/DQ nodes on Concat inputs, preventing TRT INT8 Concat fusion. Three fixes:

- Remove Concat from autotuner skip-ops list (Concat is byte-level copy in TRT and can pass INT8 data through)
- Add Conv->[BN->]Add skip rule for TRT Conv+Add+Relu kernel fusion
- Promote tensor-level Q/DQ when uncovered consumers are all Concat

## Summary

  - Remove Concat from autotuner skip-ops list — it was incorrectly blocking Q/DQ insertion at all Relu->Concat boundaries (root cause)
  - Add Concat-aware merge promotion: when uncovered consumers are all Concat, promotes to tensor-level Q/DQ (enables INT8 Concat fusion)
  - Add Conv->[BN->]Add skip rule to prevent breaking TRT Conv+Add+Relu INT8 kernel fusion
  - Add atomic Concat-group mutation to scheme sampling so the search explores all-or-nothing Concat quantization

### Usage

```python
```

## Summary by CodeRabbit

* **Enhancements**
  * Autotuner now performs Concat-group-aware mutations that add or remove entire Concat input groups, improving exploration of quantization schemes in Concat-heavy regions.
  * Insertion-point logic refined to avoid disrupting main Conv/BN/activation paths, skip additional structural/utility ops, and better promote safe INT8 Concat fusion.
  * New configuration option to control minimum Concat-group mutation sampling per region.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1398)

### DIFF
--- a/modelopt/onnx/quantization/autotune/autotuner_base.py
+++ b/modelopt/onnx/quantization/autotune/autotuner_base.py
@@ -929,6 +929,97 @@ class QDQAutotunerBase:
             for p in self.profiled_patterns
         )
 
+    def _sample_concat_group_mutation(
+        self,
+        selected_points: list,
+        all_points: list,
+        region: "Region",
+    ) -> list:
+        """Probabilistically add or remove a full Concat input group as an atomic unit.
+
+        TRT requires ALL inputs of a Concat to be INT8 for INT8 Concat fusion;
+        partial quantization has no benefit. This method treats Concat input sets
+        as atomic groups — randomly choosing to add all inputs of one Concat, or
+        remove all inputs of one Concat, as a single mutation step.
+
+        Called with adaptive probability (min 5%, scaling with budget) during scheme
+        generation to inject Concat-aware samples into the search space without forcing
+        all schemes to have full groups.
+
+        Args:
+            selected_points: Currently selected NodeInputInsertionPoint list
+            all_points: Full set of available NodeInputInsertionPoint list
+            region: The region being profiled
+
+        Returns:
+            Updated list with one Concat group atomically added or removed
+        """
+        # Identify which local node indices are Concat ops
+        node_indices = region.get_nodes(sort=True)
+        concat_local_indices = set()
+        for local_idx, node_idx in enumerate(node_indices):
+            node = self.graph.nodes[node_idx]
+            if node.op == "Concat":
+                concat_local_indices.add(local_idx)
+
+        if not concat_local_indices:
+            return selected_points
+
+        # Build groups: concat_node_index -> all available insertion points for that Concat
+        concat_groups: dict[int, list] = {}
+        for p in all_points:
+            if p.node_index in concat_local_indices:
+                concat_groups.setdefault(p.node_index, []).append(p)
+
+        if not concat_groups:
+            return selected_points
+
+        # Identify fully-present and absent Concat groups in current selection
+        selected_keys = {(p.node_index, p.input_index) for p in selected_points}
+        full_groups = []  # Concat groups fully present (can remove)
+        absent_groups = []  # Concat groups fully absent (can add)
+
+        for concat_idx, group_points in concat_groups.items():
+            group_keys = {(p.node_index, p.input_index) for p in group_points}
+            present = group_keys & selected_keys
+            if len(present) == len(group_keys):
+                full_groups.append(concat_idx)
+            elif len(present) == 0:
+                absent_groups.append(concat_idx)
+            # Partial groups: also eligible for completion (add missing siblings)
+            elif len(present) < len(group_keys):
+                absent_groups.append(concat_idx)
+
+        # Choose action only from feasible options to avoid no-op mutations
+        actions = []
+        if absent_groups:
+            actions.append("add")
+        if full_groups:
+            actions.append("remove")
+        if not actions:
+            return selected_points
+        action = random.choice(actions)
+
+        if action == "add":
+            target = random.choice(absent_groups)
+            points_to_add = []
+            for p in concat_groups[target]:
+                if (p.node_index, p.input_index) not in selected_keys:
+                    points_to_add.append(p)
+            logger.debug(
+                f"Concat group mutation: added {len(points_to_add)} points for Concat node {target}"
+            )
+            return selected_points + points_to_add
+
+        elif action == "remove":
+            target = random.choice(full_groups)
+            group_keys = {(p.node_index, p.input_index) for p in concat_groups[target]}
+            result = [p for p in selected_points if (p.node_index, p.input_index) not in group_keys]
+            logger.debug(
+                f"Concat group mutation: removed {len(group_keys)} points for Concat node {target}"
+            )
+            return result
+
     def _mutate_insertion_points(
         self, base_points, all_points, point_type: str, max_mutations: int
     ) -> list:
@@ -1055,6 +1146,17 @@ class QDQAutotunerBase:
                     point_type,
                     max_mutations,
                 ),
+            )
+
+        # Probabilistically apply Concat-group-aware mutation: atomically add or remove
+        # all inputs of a Concat as a group. Probability adapts to budget:
+        # - Large budget (>=100): 5% → at least 5 samples
+        # - Small budget (<100): min_samples/budget, clamped to [0.05, 0.5]
+        num_schemes = max(len(pattern_schemes.schemes), 1)
+        concat_prob = min(max(self.config.concat_group_min_samples / num_schemes, 0.05), 0.5)
+        if random.random() < concat_prob:
+            scheme.node_inputs = self._sample_concat_group_mutation(
+                scheme.node_inputs, full_insertion_scheme.node_inputs, region
             )
 
         return scheme
--- a/modelopt/onnx/quantization/autotune/autotuner_base.py
+++ b/modelopt/onnx/quantization/autotune/autotuner_base.py
@@ -999,10 +999,7 @@ class QDQAutotunerBase:
             present = group_keys & selected_keys
             if len(present) == len(group_keys):
                 full_groups.append(concat_idx)
-            elif len(present) == 0:
-                absent_groups.append(concat_idx)
-            # Partial groups: also eligible for completion (add missing siblings)
-            elif len(present) < len(group_keys):
+            elif len(present) == 0 or len(present) < len(group_keys):
                 absent_groups.append(concat_idx)
 
         # Choose action only from feasible options to avoid no-op mutations
--- a/modelopt/onnx/quantization/autotune/autotuner_base.py
+++ b/modelopt/onnx/quantization/autotune/autotuner_base.py
@@ -1021,7 +1021,9 @@ class QDQAutotunerBase:
             logger.debug(
                 f"Concat group mutation: added {len(points_to_add)} points for Concat node {target}"
             )
-            return selected_points + points_to_add
+            # Rebuild in all_points order so scheme identity is independent of mutation history
+            result_keys = selected_keys | {(p.node_index, p.input_index) for p in points_to_add}
+            return [p for p in all_points if (p.node_index, p.input_index) in result_keys]
 
         elif action == "remove":
             target = random.choice(full_groups)
--- a/modelopt/onnx/quantization/autotune/autotuner_base.py
+++ b/modelopt/onnx/quantization/autotune/autotuner_base.py
@@ -1014,10 +1014,10 @@ class QDQAutotunerBase:
 
         if action == "add":
             target = random.choice(absent_groups)
-            points_to_add = []
-            for p in complete_concat_groups[target]:
-                if (p.node_index, p.input_index) not in selected_keys:
-                    points_to_add.append(p)
+            points_to_add = [
+                p for p in complete_concat_groups[target]
+                if (p.node_index, p.input_index) not in selected_keys
+            ]
             logger.debug(
                 f"Concat group mutation: added {len(points_to_add)} points for Concat node {target}"
             )
@@ -1031,6 +1031,8 @@ class QDQAutotunerBase:
                 f"Concat group mutation: removed {len(group_keys)} points for Concat node {target}"
             )
             return result
+
+        return selected_points
 
     def _mutate_insertion_points(
         self, base_points, all_points, point_type: str, max_mutations: int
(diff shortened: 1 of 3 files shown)


modelopt/onnx/quantization/autotune/common.py
```diff
@@ -842,6 +842,7 @@ class Config:
     minimum_schemes_to_mutate: int = 10
     maximum_mutations: int = 3
     maximum_generation_attempts: int = 100
+    concat_group_min_samples: int = 5  # Minimum Concat-group-aware mutations per region
```
gcunhase: Why was `5` chosen here? What are the benefits or a larger or smaller number? @willg-nv

modelopt/onnx/quantization/autotune/insertion_points.py
```diff
@@ -369,6 +369,30 @@ def skip_invalid_insertion_points(
...
+                            if (
+                                bn_act
+                                and bn_act.inputs
+                                and bn_act.inputs[0].op in ["Conv", "ConvTranspose"]
+                            ):
+                                conv_node = bn_act.inputs[0]
+                        if conv_node is not None and conv_node.inputs:
+                            conv_act_input = conv_node.inputs[0]
+                            if len(conv_act_input.outputs) == 1:
+                                return True
```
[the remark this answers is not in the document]
gcunhase: @willg-nv is this a reasonable comment?

modelopt/onnx/quantization/autotune/common.py
willg-nv: The purpose of this change is to mutate insertion spec to insert Q/DQ for all concat input. But this probability should not be too high (5%) to avoid impaction on normal mutation prcoess. But when sample per region is too small (eg: 30), the estimated sample count to trigger "all concat input QDQ" would be ~1, this is too small, sometimes cannot be triggered. so `concat_group_min_samples` sets the minimum sample count which could trigger "all concat input QDQ". I expect this case could be covered, but with small probability.

modelopt/onnx/quantization/autotune/insertion_points.py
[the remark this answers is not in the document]
willg-nv: Yes. TRT could fuse for Conv->Add->Relu chain so this case should be skipped.

modelopt/onnx/quantization/autotune/common.py
gcunhase: So is the 5 in `concat_group_min_samples: int = 5` a percentage? If so, can we add that in the comment? Thanks!

willg-nv:
@gcunhase I have resolved one code-quality check issue, please help start a new CI job, thanks!

modelopt/onnx/quantization/autotune/common.py
willg-nv: I have updated the comment to make it clear.

gcunhase (approved):
LGTM, deferring final approval to @ajrasane

modelopt/onnx/quantization/autotune/insertion_points.py
```diff
@@ -369,6 +369,41 @@ def skip_invalid_insertion_points(
...
+                # Conv -> [BN ->] Add -> Relu: skip quantizing the main-path Conv
+                # output feeding Add to preserve TRT Conv+Add+Relu INT8 fusion.
+                # Guards:
+                #   1. The Add output has a single consumer and that consumer is Relu
+                #      (otherwise TRT cannot fuse, and skipping removes a legitimate
+                #      quantization point).
+                #   2. The Conv feeding Add is a "main-path" Conv (its activation input
+                #      has a single consumer), not a downsample/projection Conv (whose
+                #      activation input fans out to multiple consumers).
+                if node.op == "Add":
```
cjluo-nv, line 381: > _Bot comment._

This new rule changes behaviour that an existing test pins down: `TestSkipInvalidInsertionPoints.test_residual_block_add_inputs_allowed` asserts `skip_invalid_insertion_points(graph, "conv2_out", add_node) is False`. In the residual fixture, `add_out` has a single `Relu` consumer and `conv2`'s activation input (`relu1_out`) has a single consumer, so on a real graph this now returns `True`. The test keeps passing only because the fixture uses `MagicMock` tensors, where `len(conv_act_input.outputs)` evaluates to 0.

Please add a real-graph test for both guards (Add→Relu single consumer; main-path vs. downsample Conv) and update that existing assertion so the coverage is meaningful again.

modelopt/onnx/quantization/autotune/insertion_points.py
```diff
@@ -486,18 +526,40 @@ def merge_resolved_insertion_points(
...
-        if all_users == {ip.node_index for ip in qdq_users}:
+        covered_nodes = {ip.node_index for ip in qdq_users}
+
+        if all_users == covered_nodes:
+            # All consumers have Q/DQ — merge to tensor-level
             results.add(
                 ResolvedInsertionPoint(tensor_name=tensor_name, node_index=None, input_index=None)
             )
+        elif covered_nodes and all_users - covered_nodes:
+            # Some consumers lack Q/DQ — check if all uncovered ones are Concat
```
cjluo-nv, line 537: > _Bot comment._

Two things worth confirming here:

1. The promotion fires regardless of whether the Concat's *other* inputs are quantized. If they aren't, TRT gets no INT8 Concat fusion and we've added quantization error on a branch the search never selected — so "no accuracy cost" isn't quite right in the partial case.
2. `merge_resolved_insertion_points` is only called from `insert_qdq_at_tensors` (the autotuner's synthetic export used for latency measurement). The final model is produced via `get_ort_quantization_config()` → `quantize_int8`, which doesn't apply this merge, and `op_types_to_quantize` there is `get_autotuner_quantizable_ops()`, which still has no `"Concat"`. Please confirm the promoted Q/DQ actually materialises in the ORT-quantized output, otherwise measured latency and shipped model diverge.

A unit test on this branch (uncovered consumers all Concat / mixed) would also be good — the sibling cases are already covered in `TestMergeResolvedInsertionPoints`.

modelopt/onnx/quantization/autotune/autotuner_base.py
```diff
@@ -929,6 +929,113 @@ def _is_region_profiled(self, region: Region) -> bool:
...
+        full_groups = []  # Concat groups fully present (can remove)
+        absent_groups = []  # Concat groups fully absent (can add)
+
+        for concat_idx, group_points in complete_concat_groups.items():
+            group_keys = {(p.node_index, p.input_index) for p in group_points}
+            present = group_keys & selected_keys
+            if len(present) == len(group_keys):
+                full_groups.append(concat_idx)
+            elif len(present) == 0 or len(present) < len(group_keys):
+                absent_groups.append(concat_idx)
```
cjluo-nv, line 1003: > _Bot comment._

`elif len(present) == 0 or len(present) < len(group_keys):` is equivalent to a plain `else` (the first branch already covers `present == group_keys`). The list name/comment `absent_groups  # Concat groups fully absent` is also misleading since partially-present groups land here too — which is the right behaviour (completing the group), it just isn't what the comment says. Suggest `else: incomplete_groups.append(concat_idx)`.

Also, the trailing `return selected_points` at the end of the function is unreachable given the `add`/`remove` branches both return.

modelopt/onnx/quantization/autotune/autotuner_base.py
```diff
@@ -1057,6 +1164,17 @@ def _generate_next_insertion_sample(self) -> InsertionScheme:
                 ),
             )
 
+        # Probabilistically apply Concat-group-aware mutation: atomically add or remove
+        # all inputs of a Concat as a group. Probability adapts to budget:
+        # - Large budget (>=100): 5% → at least 5 samples
+        # - Small budget (<100): min_samples/budget, clamped to [0.05, 0.5]
+        num_schemes = max(len(pattern_schemes.schemes), 1)
+        concat_prob = min(max(self.config.concat_group_min_samples / num_schemes, 0.05), 0.5)
```
cjluo-nv, line 1172: > _Bot comment._

`num_schemes` is `len(pattern_schemes.schemes)` — the number of schemes generated so far for this pattern, not the sampling budget. So the probability decays from 0.5 early to 0.05 as the search progresses; the comment above ("Large budget (>=100): 5% → at least 5 samples") describes something else and the "at least 5 samples" guarantee doesn't hold. Please reword the comment (and the "scaling with budget" line in the `_sample_concat_group_mutation` docstring) to match what the code does.

cjluo-nv (commented):

Requesting changes: three behaviour changes to Q/DQ placement land with zero new tests, and one of them silently invalidates an existing test that only passes because of MagicMock semantics.

Needs action:
- [ ] Add unit tests in `tests/unit/onnx/quantization/autotune/test_insertion_points.py` for the Concat promotion in `merge_resolved_insertion_points`, the new `Add` skip rule (both guards), and Concat no longer being in `get_autotuner_skip_ops()`.
- [ ] Fix `test_residual_block_add_inputs_allowed` — on a real graph the new Add rule now returns `True` for `conv2_out`; it only still passes because `len(MagicMock.outputs)` is 0. Rebuild it on a real `gs.Graph` and assert the intended behaviour.
- [ ] Confirm the tuned Concat Q/DQ actually reaches the shipped model: `get_ort_quantization_config()` never calls `merge_resolved_insertion_points`, and `op_types_to_quantize` comes from `get_autotuner_quantizable_ops()`, which still excludes `Concat` (see inline).
- [ ] Simplify `_sample_concat_group_mutation` (dead `elif`/unreachable `return`, "absent" list holds partial groups) and fix the "budget" comments — `num_schemes` is schemes generated so far, not the budget.
- [ ] Run `ruff format`; the `points_to_add` comprehension indentation will fail the pre-commit check.

modelopt/onnx/quantization/autotune/autotuner_base.py
```diff
@@ -532,6 +538,12 @@ def get_ort_quantization_config(
         op_types_to_quantize = get_autotuner_quantizable_ops()
         if self.config.default_quant_type == "fp8":
             op_types_to_quantize &= {"Conv", "Gemm", "MatMul", "Add"}
+
+        # If merge_resolved_insertion_points promoted tensor-level Q/DQ that now
+        # covers Concat consumers, include "Concat" so ORT quantizes those nodes.
```
cjluo-nv, line 543: > _Bot comment._

The export-parity concern is only partially resolved. Including Concat makes ORT quantize all its inputs, but the `no_quantize_inputs` loop below only records inputs with producers (`if inp.inputs`). For `Concat(relu_out, graph_input)` with only `relu_out` selected/promoted, synthetic export leaves `graph_input` untouched, while final ORT export quantizes it and never removes that Q/DQ. The same applies to an unselected initializer input. Preserve uncovered producer-less inputs as well, and add a regression through actual INT8 quantization that compares selected/unselected Concat edges with synthetic export (not just the returned op-type list).

modelopt/onnx/quantization/autotune/autotuner_base.py
```diff
@@ -938,6 +950,112 @@ def _is_region_profiled(self, region: Region) -> bool:
             for p in self.profiled_patterns
         )
```
cjluo-nv, line 951: > _Bot comment._

The new ~100-line sampling path still has no dedicated tests. Please add deterministic tests for completing a partial group, removing a full group, choosing only feasible actions, skipping groups missing a selectable input, translating non-contiguous global node indices, and preserving canonical selection order. Existing insertion-point merge tests do not exercise this helper or its integration into scheme generation.

tests/unit/onnx/quantization/autotune/test_insertion_points.py
```diff
@@ -464,14 +576,74 @@ def test_with_region(self):
...
         graph, _ = _create_residual_graph()
         add_node = graph.nodes[3]
 
+        # Mock: add_out.outputs is a MagicMock (len 0), so guard 1 never matches.
         assert skip_invalid_insertion_points(graph, "conv2_out", add_node) is False
         assert skip_invalid_insertion_points(graph, "input", add_node) is False
 
+    # -- Real-graph tests for Conv→[BN→]Add→Relu skip guards --
+
+    def test_add_relu_main_path_conv_skipped(self):
```
cjluo-nv, line 592: > _Bot comment._

The real residual fixtures address the previous MagicMock concern, but the requested eligibility regression is still absent: nothing here fails if `Concat` is restored to the skip list. Add a real Conv→Relu→Concat collection/resolution test (including preservation of integer/small-tensor filters), and cover Conv→BN→Add→Relu to exercise the new BN-unwrapping branch.

modelopt/onnx/quantization/autotune/common.py
```diff
@@ -842,6 +842,13 @@ class Config:
     minimum_schemes_to_mutate: int = 10
     maximum_mutations: int = 3
     maximum_generation_attempts: int = 100
+    # Minimum number of Concat-group-aware mutations to attempt per region. Controls the
```
cjluo-nv, line 845: > _Bot comment._

The author's clarification and updated sampling-site comment explain the intent, but this configuration comment still promises behavior the implementation does not provide. A Bernoulli draw with a decaying probability guarantees no minimum number of attempts, and ordinary mutations still select individual Concat inputs. Describe this as the numerator controlling an adaptive mutation probability, with `num_schemes` meaning schemes generated so far; state that only the chosen group mutation is atomic.

cjluo-nv (commented):

Requesting changes: prior guard and cleanup fixes are addressed, but final ORT export can still quantize unselected Concat inputs and critical coverage remains missing.

Needs action:
- [ ] Fix uncovered graph-input handling in `get_ort_quantization_config()` and add an actual ORT-output regression test; see inline.
- [ ] Add deterministic Concat-group mutation tests in `test_autotuner.py` for add/remove, partial selections, filtered groups, and canonical ordering.
- [ ] Complete `test_insertion_points.py` coverage for Concat eligibility and the optional BN path.
- [ ] :speech_balloon: Author clarified the sampling knob — revise `common.py` to describe adaptive probability, not a guaranteed minimum or globally all-or-nothing quantization.

No action needed:
- :heavy_check_mark: Resolved prior concerns include real-graph Add guards, merge-branch tests, complete-group filtering, feasible actions, probability clamping, ordering, and helper cleanup.
- The existing mock test rename is justified: its limitation is documented and real-graph assertions now cover the changed behavior.

PR #1405, opened 2026-05-07
Schematize config loading and quantizer config entries

shengliangxu:
### What does this PR do?

This PR makes ModelOpt config loading schema-aware and moves `quant_cfg` entries from `TypedDict` validation to Pydantic validation.

Key changes:

- `load_config()` now returns validated schema instances when a schema is provided through `schema_type=...` or declared with `# modelopt-schema:`.
  - Without a schema, it still returns the raw resolved dict/list.
  - With a schema, it returns the validated schema object, such as `QuantizeConfig` or `list[QuantizerCfgEntry]`.

- `QuantizerCfgEntry` is now a `ModeloptBaseConfig` Pydantic model.
  - The “must specify `cfg`, `enable`, or both” rule is enforced wherever entries are constructed.
  - Enabled entries must provide non-empty `cfg` values when `cfg` is present.
  - Existing mapping-style access like `entry["cfg"]` and `entry.get("enable")` continues to work.

- `RecipeMetadataConfig` is now a `ModeloptBaseConfig`, so recipe metadata uses the same schema validation path.

- Recipe loading now delegates shape validation to `load_config()` instead of manually checking loaded dicts.

- `normalize_quant_cfg_list()` now accepts:
  - `Sequence[QuantizerCfgEntry]`
  - `Sequence[Mapping[str, Any]]`
  - legacy flat mapping configs, with deprecation warnings

- Public quantization config constants remain plain dict/list structures for backward compatibility, even when they are built from schema-validated YAML snippets.

### Behavior changes

- `load_config()` may now return a schema instance instead of a raw `dict`/`list` when a schema is available.
  Callers that checked `isinstance(result, dict)` should use `isinstance(result, Mapping)` or check the expected schema type.

- Normalized `quant_cfg` entries are now `QuantizerCfgEntry` objects internally.
  Mapping-style access is preserved, but direct equality checks against literal dicts should use `entry.model_dump()`.

### Testing

- `python -m pytest tests/unit/recipe/test_loader.py`
- `python -m pytest tests/unit/torch/quantization/test_config_validation.py`
- `ruff check` / `ruff format --check` on touched files
- Commit-time pre-commit hooks passed


cjluo-nv (commented):

This is a cohesive schematization refactor that builds on already-established in-repo patterns (the `$import` system from #1253, `ModeloptBaseConfig` for pydantic configs, `load_config`'s `# modelopt-schema:` declaration). It does **not** introduce a new framework — it migrates `QuantizerCfgEntry` from `TypedDict` to `ModeloptBaseConfig` and lets `load_config()` return validated schema instances. So design-protocol-wise the choice is to extend existing systems, which is the right call.

**User interface answer (asked in instructions):** Yes, this PR changes user-facing interfaces. Two intentional behavior changes are called out in the PR body and warrant a human checking downstream consumers:

1. `load_config()` may now return a schema instance (e.g. `QuantizeConfig`, `list[QuantizerCfgEntry]`) instead of a raw `dict`/`list` whenever a schema is supplied via `schema_type=` or a `# modelopt-schema:` comment. Callers doing `isinstance(result, dict)` must switch to `Mapping`/schema-type checks.
2. Normalized `quant_cfg` entries are now `QuantizerCfgEntry` pydantic instances, not plain dicts. Mapping-style access (`entry["cfg"]`, `entry.get(...)`) still works (good — `ModeloptBaseConfig` now subclasses `MutableMapping`), but literal-dict equality (`entry == {...}`) breaks; tests were updated to use `entry.model_dump()`. The public quant config constants (`FP8_DEFAULT_CFG`, etc.) are deliberately dumped back to plain dicts via `model_dump(exclude_unset=True)` for backward compat — that's a thoughtful choice that I'd want a human to bless.

Other notes:
- `ModeloptBaseConfig.__delitem__` now raises `TypeError` (not `KeyError`), and `__getitem__`/`__setitem__` were switched to `KeyError` to satisfy the `MutableMapping` protocol. Inherited `pop(key)` on an existing key will therefore raise `TypeError`, which is documented but unusual.
- Two tests (`test_error_on_non_dict_non_list_cfg_enable_true`, `test_error_on_cfg_list_with_non_dict_element_enable_true`) lost their `match="non-empty dict"` assertion because pydantic's field-type check now fires before `QuantizerCfgEntry`'s model validator. Acceptable but a weaker check.
- PTQ recipes without a `metadata` or `quantize` section are now hard-rejected (no silent default fallback) — explicit and tested.
- One small concern: `find_quant_cfg_entry_by_path` dropped its `isinstance(entry, dict)` filter, which was load-bearing only for legacy mixed lists. Acceptable given the type signature is now `list[QuantizerCfgEntry]`, but a human should confirm no caller still feeds it a heterogeneous list.

Tests are reasonable and cover the new schema-instance return path, missing-section rejection, and the empty-cfg-with-disable normalization. PR size is moderate (+436/-218, 11 files). Recommending nudge so an owner with architectural context can sign off on the public-API behavior change rather than auto-approving.

shengliangxu:
Thanks for the review. Walking through the four observations:

**1. `pop(key)` on an existing key raises `TypeError`** — intentional. Schema fields are fixed; we can't silently delete a `quantizer_name` from a `QuantizerCfgEntry` (the resulting object would no longer satisfy its own validator), and resetting to default would surprise callers. Raising `TypeError` from `__delitem__` and letting the inherited `MutableMapping.pop` / `clear` / `popitem` mixin chain bubble that up was a deliberate design choice — documented in the class docstring and in the `__delitem__` docstring. `pop(missing_key, default)` still returns the default cleanly because the mixin short-circuits on `KeyError` from `__getitem__` before reaching `__delitem__`. If reviewers want softer semantics, the alternative is to override `pop` / `popitem` / `clear` individually so they raise `KeyError` (or a more specific error) before hitting `__delitem__` — happy to do that if preferred.

**2. Two cfg-shape tests lost their `match="non-empty dict"` assertion** — addressed in d7b6e0a2 ("test(quant): tighten cfg-shape rejection assertions"). Previously the tests just asserted `ValueError`, which is too weak. The new assertion is a regex that accepts either error path (pydantic field-type check or our model validator) but requires the error to implicate the `cfg` field and identify a type/shape problem:

```python
match=r"(?s)cfg.*(non-empty|valid dictionary|valid list)"
```

**3. `find_quant_cfg_entry_by_path` dropped its `isinstance(entry, dict)` filter** — audited all callers. There are five:
- `modelopt/torch/quantization/backends/{fp8_per_tensor,nvfp4}_gemm.py` pass `FP8_DEFAULT_CFG["quant_cfg"]` / `NVFP4_DEFAULT_CFG["quant_cfg"]`, which are all-dict after c43c0f0f6d ("fix(quant): keep shared cfg snippets as dicts in public constants").
- `examples/deepseek/ptq.py` passes `mtq_cfg["quant_cfg"]`, built from the same dict-based public constants.
- `examples/llm_ptq/example_utils.py` passes `quant_cfg["quant_cfg"]` after `copy.deepcopy(quant_cfg)` — homogeneous within one call (all-dict if the input was a dict, all-`QuantizerCfgEntry` if the input was a `QuantizeConfig`). `.get("quantizer_name")` works on both via the `MutableMapping` interface, so no heterogeneous-list breakage.
- The tests pass deliberate homogeneous lists.

No mixed-list caller observed → the dropped filter is safe to leave out.

**4. Two intentional public-API behavior changes** — confirmed in the PR description ("Behavior changes worth calling out"); both are flagged there for downstream consumers.

cjluo-nv (commented):

**Does this PR change user interfaces? — Yes, intentionally and documented.** Two public-API behavior changes are spelled out in the PR body and code comments:

1. `load_config()` now returns a validated schema instance (e.g. `QuantizeConfig`, `RecipeMetadataConfig`, `list[QuantizerCfgEntry]`) instead of a raw `dict` / `list` whenever a schema is supplied via `schema_type=...` or a `# modelopt-schema:` comment. Callers using `isinstance(result, dict)` must switch to `isinstance(result, Mapping)` or check the expected schema type.
2. Normalized `quant_cfg` entries are now `QuantizerCfgEntry` pydantic instances. Mapping-style access (`entry["cfg"]`, `entry.get(...)`) still works because `ModeloptBaseConfig` now subclasses `MutableMapping`, but literal-dict equality (`entry == {...}`) breaks; existing tests were updated to compare via `entry.model_dump()`.
3. Side effect: `ModeloptBaseConfig.__delitem__` now raises `TypeError` (schema fields are fixed), while `__getitem__`/`__setitem__` raise `KeyError` (was `AttributeError`) to satisfy the `MutableMapping` protocol. Inherited `pop(existing_key)` / `clear()` / `popitem()` will therefore raise `TypeError` — unusual but documented in the class docstring. `pop(missing, default)` still returns the default normally.

**Design protocol:** This is *not* a new framework — it extends the existing `$import` / `load_config` system (PR #1253) and the existing `ModeloptBaseConfig` pydantic base, migrating `QuantizerCfgEntry` from `TypedDict` to a pydantic model and `RecipeMetadataConfig` similarly. The author chose to extend rather than introduce a parallel system, which is the right call.

**Re-review state of CodeRabbit comments:**
- `load_config` returning instances when schema is supplied — addressed (overloads added, return type updated, recipe loader simplified).
- `MutableMapping` protocol violations on `__getitem__` / `__setitem__` — addressed (now raise `KeyError`).
- Empty-cfg-with-disable normalization to `None` — addressed and tested.
- Recipe missing `metadata` / `quantize` hard-rejected — addressed and tested.
- Out-of-diff CodeRabbit nits about `tensor_quantizer.py`, `core_utils.py`, `hf_ptq.py`, `example_utils.py`, `vllm_ptq_utils.py` are touching files **not changed in this PR** — they're follow-ups, not blockers for this diff.

**Other notes worth a human eye:**
- Two tests (`test_error_on_non_dict_non_list_cfg_enable_true`, `test_error_on_cfg_list_with_non_dict_element_enable_true`) lost their `match="non-empty dict"` assertion because pydantic's field-type check now fires *before* `QuantizerCfgEntry`'s model validator. Acceptable but a weaker check; comment in test calls this out.
- Public quant config constants (`FP8_DEFAULT_CFG`, `FP8_KV_CFG`, `_default_disabled_quantizer_cfg`) are deliberately dumped back to plain dicts via `model_dump(exclude_unset=True)` for back-compat. Thoughtful, but worth a human nodding at.
- `find_quant_cfg_entry_by_path` dropped its `isinstance(entry, dict)` filter — now relies on the typed `list[QuantizerCfgEntry]` signature; safe given the new normalizer always returns instances.

Tests are reasonable (+~50 new test lines covering the new schema-instance return, missing-section rejection, empty-cfg normalization, list-cfg-with-empty-dicts). PR size is moderate (+436/-218, 11 files in-diff). No licensing concerns. Recommending **nudge** so an owner with architectural context can sign off on the public-API behavior change rather than auto-approving.

cjluo-nv (commented):

**User interfaces changed: yes — and intentionally.** The PR documents two public API behavior changes: (1) `load_config()` now returns validated schema instances (e.g. `QuantizeConfig`, `RecipeMetadataConfig`, `list[QuantizerCfgEntry]`) instead of raw dicts/lists when a schema is supplied via `schema_type=` or a `# modelopt-schema:` comment — callers using `isinstance(result, dict)` must switch to `Mapping` checks; (2) normalized `quant_cfg` entries are now `QuantizerCfgEntry` pydantic instances. Mapping-style access still works (because `ModeloptBaseConfig` now subclasses `MutableMapping`), but literal-dict equality breaks; existing tests were updated accordingly.

**Design protocol**: This is *not* a new framework — it extends the existing `$import` / `load_config` schema mechanism (PR #1253) and the existing `ModeloptBaseConfig` pydantic base. The author chose to extend rather than introduce a parallel system; that's the right call and no design justification is missing.

**Re-review of previous comments**:
- ✅ `__getitem__` / `__setitem__` now raise `KeyError` (MutableMapping protocol) — addressed.
- ✅ Empty-cfg-with-`enable=False` normalised to `cfg=None` — addressed and tested.
- ✅ Recipe missing `metadata`/`quantize` hard-rejected — addressed and tested.
- ⚠️ `__delitem__` raises `TypeError` rather than `KeyError`. The author documented this deliberately ("schema fields are fixed") so inherited `pop(existing)` / `clear()` raise `TypeError`. Unusual but documented; worth a human nodding at.
- ⚠️ `examples/llm_ptq/example_utils.py` `build_quant_cfg` phi4mm path still appends raw dicts (`{"quantizer_name": "*speech*", "enable": False}`) into a config that the rest of the PR is migrating to typed `QuantizerCfgEntry`. Same pattern in `examples/llm_ptq/hf_ptq.py` `mono_quantize` for Nemotron VL. These work because `normalize_quant_cfg_list` accepts both shapes, but they undercut the typing story.
- ⚠️ Tests `test_error_on_non_dict_non_list_cfg_enable_true` / `test_error_on_cfg_list_with_non_dict_element_enable_true` lost their tight `match="non-empty dict"` assertion because pydantic's field-type check now fires before the model validator. Acceptable but a weaker check.
- ⚠️ Outside-diff CodeRabbit nits about `tensor_quantizer.py` (silent-truncation `zip(attributes, self)`), `core_utils.py` (shared-mutable-state on KV append) — valid follow-up concerns, but not regressions introduced by this PR.

**Other notes worth a human eye**:
- Public quant config constants (`FP8_DEFAULT_CFG`, `FP8_KV_CFG`, `_default_disabled_quantizer_cfg`) are deliberately dumped back to plain dicts via `model_dump(exclude_unset=True)` for back-compat — thoughtful, but a deliberate design decision worth blessing.
- `find_quant_cfg_entry_by_path` dropped its `isinstance(entry, dict)` filter — safe given the new typed signature but worth confirming no caller still feeds a heterogeneous list.

PR size is moderate (+440/-218, 11 files). Tests are reasonable and cover the new schema-instance return, missing-section rejection, and empty-cfg normalisation. No licensing concerns. Recommending **nudge** so an owner with architectural context can sign off on the public-API behaviour change rather than auto-approving.

---

- `examples/llm_ptq/example_utils.py:255` — > _Bot comment._

These four appends (and the parallel block in `hf_ptq.py` `mono_quantize` for Nemotron VL) push raw dicts back into `quant_cfg_obj["quant_cfg"]` after the rest of the PR migrates entries to typed `QuantizerCfgEntry`. Functionally fine because `normalize_quant_cfg_list` still accepts dicts, but it undercuts the schematization story for callers that introspect `build_quant_cfg`'s return value. Consider constructing `QuantizerCfgEntry(quantizer_name=..., enable=False)` here for consistency.
- `modelopt/torch/quantization/utils/core_utils.py:939` — > _Bot comment._

`updated_quant_cfg["quant_cfg"] = inner + list(kv_cache_quant_cfg)` references the caller's `kv_cache_quant_cfg` entries by reference. The outer `copy.deepcopy(quant_cfg)` at the top doesn't cover this — `kv_cache_quant_cfg` is the second argument and isn't deep-copied. Now that entries are `QuantizerCfgEntry` pydantic instances, downstream mutation of `entry["cfg"]` (e.g. `_set_kv_cache_constant_amax` in `hf_ptq.py`) can leak back into the shared `mtq.FP8_KV_CFG`/`NVFP4_KV_CFG` constants. Suggest `inner + [copy.deepcopy(e) for e in kv_cache_quant_cfg]`.

modelopt/torch/opt/config.py
```diff
@@ -110,18 +117,49 @@ def __contains__(self, key: str) -> bool:
...
+        class fully satisfies the ``MutableMapping`` protocol (``__delitem__`` is
+        required), without committing to actual deletion semantics.
+        """
+        raise TypeError(
+            f"{type(self).__name__} does not support key deletion; schema fields are "
+            f"fixed (attempted to delete {key!r})."
+        )
 
     def get(self, key: str, default: Any = None) -> Any:
         """Get the value for the given key (can be name or alias) or default if not found."""
```
cjluo-nv, line 159: > _Bot comment._

Choosing `TypeError` for `__delitem__` (rather than `KeyError`) means inherited `pop(existing_key)` / `clear()` / `popitem()` all raise `TypeError`, which is unusual for a `MutableMapping`. The class docstring documents this, but third-party generic mapping code (e.g. anything that catches `KeyError` to detect deletion failure) will see an unexpected exception type. Worth confirming with the team that `TypeError` is the desired contract here vs. `KeyError`.

shengliangxu: Sticking with `TypeError` here, deliberately. Reasoning:

`KeyError` is the conventional signal for *"key not present in the mapping"* — i.e. the lookup failed. That's not what's happening: the key **is** present (it's a declared schema field); we are refusing to perform the operation at all. The semantically closer exception is `TypeError` — "operation not supported on this type" — which is also what `types.MappingProxyType` raises when you attempt mutation on a read-only mapping. We'd be lying about the cause of failure if we raised `KeyError`.

The concrete worry — "third-party code that does `try: del cfg[k] except KeyError` will see an unexpected exception" — is the *intended* behavior, not a bug. That pattern only makes sense when the caller already assumes deletion is supported and is hedging against the key being absent. If deletion isn't supported on this mapping at all, the caller's assumption is wrong and we want them to find out, not silently no-op past the refusal. A swallowed `KeyError` could leave a schema field untouched after a `pop` that the caller believes succeeded, which is a worse failure mode than a loud `TypeError` at the call site.

Documented in the class docstring and the `__delitem__` docstring, and inherited `pop` / `popitem` / `clear` follow naturally — they all chain through `__delitem__` and surface the same `TypeError`, which is consistent. `pop(missing_key, default)` still returns the default cleanly because the `MutableMapping` mixin short-circuits on `KeyError` from `__getitem__` before `__delitem__` is reached.

Happy to revisit if a concrete downstream caller surfaces that needs `KeyError`, but I'd rather not weaken the signal speculatively.

cjluo-nv (commented):

Re-review of the latest commit. The PR extends established in-repo patterns (`$import` / `load_config` from #1253, `ModeloptBaseConfig`) rather than introducing a new system, so the design protocol passes without a separate justification needed.

**Critical previous comments — addressed:**
- `__getitem__` / `__setitem__` now raise `KeyError`, satisfying the `MutableMapping` protocol.
- Empty `cfg` with `enable=False` is normalized to `cfg=None` (prevents silent attribute reset on later re-enable); test added.
- Recipes missing `metadata` or `quantize` are now hard-rejected (no silent default fallback); tests added.
- `load_config()` now returns schema instances when a schema is supplied (via `schema_type=` arg or `# modelopt-schema:` comment), with proper `@overload`s for the typed list / single-instance / untyped cases.
- `normalize_quant_cfg_list` validates entries through `QuantizerCfgEntry`'s pydantic model validators, with passthrough for already-validated instances.

**Documented design choices that warrant a human nod:**
- `ModeloptBaseConfig.__delitem__` raises `TypeError` (not `KeyError`). Author defended this in a reply: schema fields can't be deleted, `TypeError` matches `MappingProxyType`'s "operation not supported" semantics, and documenting it loud-fails on any caller who assumed deletion works. Reasonable, but unusual for a `MutableMapping` and worth confirming with the team.
- Public quant config constants (`FP8_DEFAULT_CFG`, `FP8_KV_CFG`, `_default_disabled_quantizer_cfg`) are deliberately dumped back to plain dicts via `model_dump(exclude_unset=True)` for back-compat. Thoughtful, but a deliberate API surface decision.
- Two tests (`test_error_on_non_dict_non_list_cfg_enable_true` / `test_error_on_cfg_list_with_non_dict_element_enable_true`) lost their tight `match="non-empty dict"` assertion because pydantic's field-type check now fires before the model validator. The replacement regex is acceptable but a weaker check; explicitly noted in the test docstrings.

**Minor — non-blocking:**
- `examples/llm_ptq/example_utils.py` `build_quant_cfg` phi4mm path still appends raw `{"quantizer_name": "*speech*", "enable": False}` dicts. This is consistent with the constants being plain-dict trees post-`model_dump`, so it's not actually a regression — the caller-facing config tree is still dict-shaped — but it does undercut the typing story for any caller introspecting `build_quant_cfg`'s return value.
- CodeRabbit's follow-ups about `tensor_quantizer.py` (silent-truncation `zip(attributes, self)`) and `core_utils.py` (shared-mutable-state on KV append) touch files outside this PR's diff and are pre-existing concerns, not regressions introduced here.

PR size is moderate (+469/-254, 11 files in-diff). Tests are reasonable and cover the new schema-instance return path, missing-section rejection, and empty-cfg normalization. No licensing concerns.

Recommending **nudge** so an owner with architectural context can sign off on the public-API behavior change (load_config return type, QuantizerCfgEntry being pydantic instances, `__delitem__` → `TypeError`) rather than auto-approving.

cjluo-nv (approved):
LGTM. Maybe we can also wait for other reviews before merging

modelopt/torch/opt/config.py
shengliangxu: pop(key) on an existing key raises TypeError — intentional. Schema fields are fixed; we can't silently delete a quantizer_name from a QuantizerCfgEntry (the resulting object would no longer satisfy its own validator), and resetting to default would surprise callers. Raising TypeError from __delitem__ and letting the inherited MutableMapping.pop / clear / popitem mixin chain bubble that up was a deliberate design choice — documented in the class docstring and in the __delitem__ docstring. pop(missing_key, default) still returns the default cleanly because the mixin short-circuits on KeyError from __getitem__ before reaching __delitem__. If reviewers want softer semantics, the alternative is to override pop / popitem / clear individually so they raise KeyError (or a more specific error) before hitting __delitem__ — happy to do that if preferred.

PR #1407, opened 2026-05-07
[Recipes][LLM PTQ] Add nvfp4 MSE+FP8-cast-KV recipes (experts_only / mlp_only) + --recipe in example scripts

cjluo-nv:
## Summary

- Adds two PTQ recipes that combine **experts/MLP-only NVFP4 W4A4** with **MSE FP8 scale-sweep weight calibration** and **FP8 KV cache with `use_constant_amax: true`** (skips KV calibration; matches the `nvfp4_default-fp8_cast_kv` contract):
  - `modelopt_recipes/general/ptq/nvfp4_experts_only_mse-fp8_cast_kv.yaml` — applies to `*mlp.experts*` / `*block_sparse_moe*` only.
  - `modelopt_recipes/general/ptq/nvfp4_mlp_only_mse-fp8_cast_kv.yaml` — applies to all `*mlp*` / `*block_sparse_moe*` (dense MLP + MoE).
- Threads a new `--recipe` flag through `examples/llm_ptq/scripts/parser.sh` and `huggingface_example.sh`. Either `--quant` or `--recipe` is required; passing **both errors out**. Recipe names are not validated in the script — `hf_ptq.py` is the source of truth.
- Drops the bash-side `qformat` whitelist case-statement in `huggingface_example.sh` for the same reason.

## Files

**New recipes (`modelopt_recipes/general/ptq/`):**
- `nvfp4_experts_only_mse-fp8_cast_kv.yaml` — same patterns as `nvfp4_experts_only-fp8_kv.yaml`.
- `nvfp4_mlp_only_mse-fp8_cast_kv.yaml` — same patterns as `nvfp4_mlp_only-fp8_kv.yaml`.

Both differ from their `_kv` siblings by:
- `algorithm: max` → `{ method: mse, fp8_scale_sweep: true, layerwise: false }`
- All targeted **weight quantizers** switch `type: dynamic` → `type: static` (otherwise `mse_calibrate` skips them: only static block-quant weight quantizers are recognized for the FP8 sweep — see `model_calib.py:369-374`).
- Input quantizers stay dynamic.
- KV bmm adds `use_constant_amax: true` (the `_cast_kv` flavor).

**Scripts (`examples/llm_ptq/scripts/`):**
- `parser.sh` — adds `--recipe` long-option, default `RECIPE=""`, validates one-of-{`--quant`, `--recipe`} and not-both.
- `huggingface_example.sh` — when `RECIPE` is set, derives `MODEL_NAME` from the recipe basename, passes `--recipe=…` to `hf_ptq.py` instead of `--qformat=…`, and exits after export with a TRT-LLM deployment hint (recipes can produce arbitrary configs that the script's downstream `run_tensorrt_llm.py` path doesn't know how to handle generically). Drops the `qformat` whitelist; defers to `hf_ptq.py`.

## Behavior

```
# Errors with: "Cannot specify both --quant and --recipe; pick one."
bash huggingface_example.sh --model=... --quant=nvfp4 --recipe=... --tasks=quant

# Errors with usage if neither is given
bash huggingface_example.sh --model=... --tasks=quant

# Both of these are now accepted; --recipe is forwarded verbatim to hf_ptq.py
bash huggingface_example.sh --model=... --quant=nvfp4 --tasks=quant
bash huggingface_example.sh --model=... --recipe=general/ptq/nvfp4_experts_only_mse-fp8_cast_kv --tasks=quant
bash huggingface_example.sh --model=... --recipe=general/ptq/nvfp4_mlp_only_mse-fp8_cast_kv  --tasks=quant
```

## Test plan

- [x] `experts_only_mse-fp8_cast_kv` loads via `modelopt.recipe.load_recipe(...)` and produces the expected algorithm + per-pattern `quant_cfg` (verified in a working env: `algorithm == {'method': 'mse', 'fp8_scale_sweep': True, 'layerwise': False}`; expert weight quantizers `type: static`; KV bmm has `use_constant_amax: True`).
- [x] Parser sanity: 4 flag combinations (both, neither, only `--quant`, only `--recipe`) all behave as designed.

## Note

Pre-commit hook `check-modelopt-recipes` was skipped on both commits because the local conda env has a broken `torchvision` install (`AttributeError: partially initialized module 'torchvision' has no attribute 'extension'`) that prevents `from modelopt.recipe.loader import load_recipe`. The `experts_only` recipe was validated independently by running `tools/precommit/check_modelopt_recipes.py` in a working environment (exits 0); the `mlp_only` one is the same shape with a different glob.

Rebased onto `main` from #1391 (which targeted `chenjiel/nvfp4-fp8-sweep-triton`). The diff is scoped to the recipes + script wiring; no kernel/sweep changes are included here.

## Summary by CodeRabbit

* **New Features**
  * Added recipe-based quantization as an alternative to format-based quantization with a new `--recipe` CLI option.
  * Added two new quantization recipes for targeted layer optimization: one for expert-layer-only quantization and one for MLP-layer-only quantization, both featuring NVFP4 and FP8 KV-cache optimization.

* **Configuration**
  * `--quant` and `--recipe` options are now mutually exclusive; specify one to configure quantization behavior.


examples/llm_ptq/scripts/huggingface_example.sh
```diff
@@ -49,18 +49,7 @@ dense | sparsegpt) ;;
...
-    case $qformat in
-    fp8 | fp8_pc_pt | fp8_pb_wo | int8_wo | int8_sq | int4_awq | w4a8_awq | fp16 | bf16 | nvfp4 | nvfp4_awq | nvfp4_mse | w4a8_nvfp4_fp8 | w4a8_mxfp4_fp8 | nvfp4_experts_only | nvfp4_mlp_only | nvfp4_omlp_only | nvfp4_svdquant | mxfp8 | nvfp4_local_hessian) ;;
-    *)
-        echo "Unknown quant argument: Expected one of: [fp8, fp8_pc_pt, fp8_pb_wo, int8_wo, int8_sq, int4_awq, w4a8_awq, fp16, bf16, nvfp4, nvfp4_awq, nvfp4_mse, w4a8_nvfp4_fp8, w4a8_mxfp4_fp8, nvfp4_experts_only, nvfp4_mlp_only, nvfp4_omlp_only, nvfp4_svdquant, mxfp8, nvfp4_local_hessian]" >&2
-        exit 1
-        ;;
-    esac
-done
-IFS=" "
+# Quant format / recipe validation is delegated to hf_ptq.py.
```
meenchen, line 52: > _Bot comment._

Regression: deleting the `for qformat in $QFORMAT; do … done` loop also drops the implicit binding of the lowercase loop variable `$qformat`, which is still used below at `if [ "$qformat" == "bf16" ] || [ "$qformat" == "fp16" ]`. With the loop removed, `$qformat` is empty and that bf16/fp16 shortcut (which symlinks the source model into `$SAVE_PATH` and marks `MODEL_CONFIG_EXIST=true`) will never trigger — users running `--quant=bf16` or `--quant=fp16` will now fall through to `python hf_ptq.py --qformat=bf16` instead. Either replace `$qformat` with `$QFORMAT` in that check, or add a dedicated `qformat="$QFORMAT"` assignment here.

modelopt_recipes/general/ptq/nvfp4_experts_only_mse-kv_fp8_cast.yaml
```diff
@@ -0,0 +1,48 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
meenchen, line 1: > _Bot comment._

Stylistic nit: sibling recipes in this directory follow the `<numerics>-kv_<kv_fmt>` naming pattern (e.g. `nvfp4_experts_only-kv_fp8.yaml`, `nvfp4_default-kv_fp8_cast.yaml`). This file uses `_mse-fp8_cast_kv` which flips the KV descriptor order. Consider renaming to something like `nvfp4_experts_only_mse-kv_fp8_cast.yaml` for consistency with the existing convention. Same for `nvfp4_mlp_only_mse-fp8_cast_kv.yaml`.

meenchen (commented):

The recipe YAMLs themselves look correct — they reuse existing `nvfp4`, `nvfp4_static`, and `kv_fp8_cast` units, and only differ from the `-kv_fp8` siblings in the ways the PR body describes (weight quantizers switch to `type: static` to be eligible for the MSE FP8 scale sweep; `algorithm` switches to `{method: mse, fp8_scale_sweep: true, layerwise: false}`). The `hf_ptq.py --recipe` flag is already wired up from the prior PR and correctly bypasses the post-hoc `_set_kv_cache_constant_amax` so the recipe is authoritative.

One real regression in `huggingface_example.sh`: deleting the `for qformat in $QFORMAT; do … done` whitelist also removes the implicit binding of the lowercase loop variable `$qformat`, which is still referenced later in the `bf16`/`fp16` shortcut branch (`if [ "$qformat" == "bf16" ] || [ "$qformat" == "fp16" ]`). With the loop gone, that variable is empty, so the symlink-the-model shortcut for pure-bf16/fp16 runs is now dead code. Replace `$qformat` with `$QFORMAT` (or reintroduce a single assignment) to preserve the prior behavior.

Minor stylistic nit: the new files are named `*_mse-fp8_cast_kv.yaml` while the existing convention in `modelopt_recipes/general/ptq/` is `*-kv_fp8_cast` (KV descriptor after the dash, with `kv_` prefix). Not blocking, but inconsistent with siblings.

Tests: no new unit tests, and the PR notes the pre-commit `check-modelopt-recipes` was skipped locally due to a broken env. The recipes are YAML-only and are covered by the standard recipe loader check in CI, so this is low-risk provided CI runs that check — but the author should confirm the pre-commit hook passes in CI on this branch.

examples/llm_ptq/scripts/huggingface_example.sh
meenchen: Do we still need bf16/fp16 path anyway? Maybe we can deprecate them

modelopt_recipes/general/ptq/nvfp4_experts_only_mse-kv_fp8_cast.yaml
meenchen: +1

hychiang-git: agree.

examples/llm_ptq/scripts/huggingface_example.sh
hychiang-git: Not sure if we still have the use cases where we quantize fp32 to fp16.

cjluo-nv: yeah I think we can delete. Let me add this to the PR

meenchen (approved):

Re-review: all critical previous comments have been addressed.

- **Regression with `$qformat` bf16/fp16 shortcut**: Fixed by removing the block entirely (author and meenchen agreed to deprecate; no use case for bf16/fp16 "quantize" today).
- **Naming convention**: Files renamed to `nvfp4_experts_only_mse-kv_fp8_cast.yaml` and `nvfp4_mlp_only_mse-kv_fp8_cast.yaml`, matching the sibling `*-kv_fp8_cast` convention.
- **`mlp_only` description wording**: Updated from "all linear layers" to "MLP/MoE linear layers".
- **Basename quoting**: `$MODEL_PATH` and `$RECIPE` are now quoted in the new basename calls.

Remaining CodeRabbit nit (`QUANT_SPEC_ARGS` as plain string instead of a Bash array) is stylistic; recipe paths are repo-relative without spaces, and the prior script already uses unquoted expansions elsewhere. Not a blocker.

Licensing: new YAML files carry the standard NVIDIA Apache-2.0 header; no licensing concern. Recipes are covered by the existing `check-modelopt-recipes` loader test in CI.

hychiang-git (approved):
LGTM

meenchen (approved):

shengliangxu (approved):
LGTM

PR #1410, opened 2026-05-07
[NVBug 6143871] Fix awq_lite uncalibrated branch leaving input_quantizer disabled

cjluo-nv:
## Summary

`awq_lite.setup()` disables `module.input_quantizer` at the start of search. The calibrated branch re-enables it inside `postprocess()`, but the uncalibrated branch (no cache-pass tokens, e.g. an MoE expert that never gets routed) never did. Worse, for experts that had cache hits but missed the search pass, the per-channel `_amax` left over from `max_calibrate` during cache mode tripped `preprocess_linear_fusion`'s `numel == 1` assertion and prevented `_export_quantized_weight` from emitting a per-tensor `input_scale`.

Result: per-expert `input_scale` was missing in the exported HF checkpoint, and TRT-LLM `CutlassFusedMoE` crashed on load with `KeyError: '<idx>.w1.input_scale'` for any expert that did not see enough calibration tokens (e.g. Qwen3-30B-A3B + `nvfp4_awq` from the bug report).

## Fix

Mirror the calibrated `postprocess()` path in `modelopt/torch/quantization/model_calib.py`: collapse any per-channel `_amax` to scalar (axis=None) and re-enable the `input_quantizer`.

## Test plan

- [x] Added regression test `test_awq_lite_uncalibrated_linear_keeps_input_quantizer_enabled` using `NVFP4_AWQ_LITE_CFG` with a two-branch model where only one branch is exercised; verifies the uncalibrated linear's `input_quantizer` remains enabled after `mtq.quantize`.
- [x] End-to-end pipeline test on a tiny synthetic Qwen3-MoE (8 experts, top-1 routing) confirms all 48 expert `input_scale` keys are present in the exported state_dict (vs. multiple missing pre-fix).
- [x] Manual repro of the bug command (Qwen3-30B-A3B, `--qformat nvfp4_awq`) confirmed the missing `input_scale` keys before the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed AWQ-Lite quantization for uncalibrated modules so export/preprocessing invariants are preserved even when calibration/parameter updates are skipped.

* **Tests**
  * Added regression tests: one verifies uncalibrated linear modules keep their input quantizer enabled after quantization; another verifies weight-only AWQ-Lite cases keep the input quantizer disabled as expected.

### DIFF
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -1316,6 +1316,19 @@ def awq_lite(
                     dtype=w_dtype,
                     device=w_device,
                 )
+                # Mirror the calibrated postprocess path: collapse any
+                # per-channel _amax left over from cache_mode max_calibrate
+                # into a per-tensor scalar (so preprocess_linear_fusion's
+                # numel==1 assertion passes), and re-enable the quantizer
+                # (awq_lite.setup disabled it). Without these, export drops
+                # input_scale and per-expert MoE loaders fail.
+                if module.input_quantizer.amax is not None:
+                    act_amax = module.input_quantizer.amax
+                    module.input_quantizer._amax_for_smoothing = act_amax.cpu()
+                    module.input_quantizer.reset_amax()
+                    module.input_quantizer.axis = None
+                    module.input_quantizer.amax = act_amax.amax()
+                module.input_quantizer.enable()
             else:
                 with enable_weight_access_and_writeback(module, model, name_to_module):
                     postprocess(module, name)
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -1316,19 +1316,21 @@ def awq_lite(
                     dtype=w_dtype,
                     device=w_device,
                 )
-                # Mirror the calibrated postprocess path: collapse any
-                # per-channel _amax left over from cache_mode max_calibrate
-                # into a per-tensor scalar (so preprocess_linear_fusion's
-                # numel==1 assertion passes), and re-enable the quantizer
-                # (awq_lite.setup disabled it). Without these, export drops
-                # input_scale and per-expert MoE loaders fail.
-                if module.input_quantizer.amax is not None:
-                    act_amax = module.input_quantizer.amax
-                    module.input_quantizer._amax_for_smoothing = act_amax.cpu()
-                    module.input_quantizer.reset_amax()
-                    module.input_quantizer.axis = None
-                    module.input_quantizer.amax = act_amax.amax()
-                module.input_quantizer.enable()
+                # Mirror the calibrated postprocess path, gated on
+                # is_input_quantized so weight-only AWQ configs (where
+                # setup() never disabled input_quantizer) stay untouched.
+                # Collapse any per-channel _amax left over from cache_mode
+                # max_calibrate into a per-tensor scalar so
+                # preprocess_linear_fusion's numel==1 assertion passes, and
+                # re-enable the quantizer (awq_lite.setup disabled it).
+                if module.awq_lite.is_input_quantized:
+                    if module.input_quantizer.amax is not None:
+                        act_amax = module.input_quantizer.amax
+                        module.input_quantizer._amax_for_smoothing = act_amax.cpu()
+                        module.input_quantizer.reset_amax()
+                        module.input_quantizer.axis = None
+                        module.input_quantizer.amax = act_amax.amax()
+                    module.input_quantizer.enable()
             else:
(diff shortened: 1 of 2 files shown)


modelopt/torch/quantization/model_calib.py
```diff
@@ -1316,6 +1316,19 @@ def postprocess(module, name):
...
+                # numel==1 assertion passes), and re-enable the quantizer
+                # (awq_lite.setup disabled it). Without these, export drops
+                # input_scale and per-expert MoE loaders fail.
+                if module.input_quantizer.amax is not None:
+                    act_amax = module.input_quantizer.amax
+                    module.input_quantizer._amax_for_smoothing = act_amax.cpu()
+                    module.input_quantizer.reset_amax()
+                    module.input_quantizer.axis = None
+                    module.input_quantizer.amax = act_amax.amax()
+                module.input_quantizer.enable()
```
meenchen: > _Bot comment._

Missing guard on `module.awq_lite.is_input_quantized`. The calibrated branch of `postprocess()` (right above, around line ~1270) only runs this block under `if module.awq_lite.is_input_quantized:`. Without that guard here, weight-only AWQ configs (e.g. `INT4_AWQ_CFG`, which sets `*input_quantizer enable: False`) will have `module.input_quantizer.enable()` called on an input_quantizer that the user's config explicitly disabled and that `setup()` never touched (setup only disables input_quantizer if `is_enabled`). That's a behavior regression for INT4_AWQ / weight-only AWQ on uncalibrated experts — the input_quantizer would start fake-quantizing in forward without an amax.

Suggest wrapping the new block in `if module.awq_lite.is_input_quantized:` to match the calibrated path, and adding a regression test with `INT4_AWQ_CFG` (or any weight-only AWQ config) + an uncalibrated branch that asserts `uncalibrated.input_quantizer.is_enabled is False` after `mtq.quantize`.

meenchen (commented):

Fix is the right shape and mirrors the calibrated postprocess path, but I think it's missing the `if module.awq_lite.is_input_quantized:` guard that the calibrated branch has. For weight-only AWQ configs (e.g. `INT4_AWQ_CFG`, where the config sets `*input_quantizer enable: False`), `is_input_quantized` is False and `setup()` never disabled the input_quantizer. With this PR, an uncalibrated expert under INT4_AWQ will now unconditionally call `module.input_quantizer.enable()`, turning on a quantizer that the user's config had explicitly disabled. Combined with the pre-existing `_enable_pre_quant_scale = True` / `pre_quant_scale = ones(...)` lines a few lines up (which were harmless when the quantizer stayed disabled), this will start applying pre_quant_scale in forward and cause an amax-less input_quantizer to try to quantize.

Also, the test only covers `NVFP4_AWQ_LITE_CFG` (input_quantizer enabled). Please add a regression test for the weight-only AWQ path (`INT4_AWQ_CFG` or similar) with an uncalibrated expert, asserting `input_quantizer.is_enabled == False` after `mtq.quantize` — that would have caught this.

cjluo-nv:
@meenchen good catch — fixed. The whole postprocess block (per-channel-amax collapse + `enable()`) is now gated behind `module.awq_lite.is_input_quantized`, so weight-only AWQ configs (e.g. `INT4_AWQ_CFG`) are untouched. Also added a regression test on `INT4_AWQ_CFG` asserting the uncalibrated linear's `input_quantizer` stays disabled.

@coderabbitai strengthened the existing test with the scalar-amax invariant (`axis is None`, `amax.numel() == 1`) — gated on `amax is not None` because the truly-uncalibrated case A (no cache hits) leaves `amax` for `set_expert_quantizer_amax` to populate during export.

meenchen (approved):

Re-review: both critical comments from the previous review are addressed.

1. **`is_input_quantized` guard**: The new block in `model_calib.py` (lines 1325-1331) wraps the `_amax` collapse and `input_quantizer.enable()` in `if module.awq_lite.is_input_quantized:`, correctly mirroring the calibrated `postprocess()` path. Weight-only AWQ configs (e.g. `INT4_AWQ_CFG`) where `setup()` never disabled the input_quantizer are now left untouched.

2. **Regression test for weight-only AWQ**: `test_awq_lite_uncalibrated_weight_only_keeps_input_quantizer_disabled` has been added, using `INT4_AWQ_CFG` with the two-branch model and asserting `uncalibrated.input_quantizer.is_enabled is False` after `mtq.quantize`. This would have caught the missing-guard bug.

3. **CodeRabbit nitpick on scalar amax invariant**: also addressed (conditionally, with a reasonable comment about the None-amax export path).

Fix is small, targeted, and well-tested.

meenchen (approved):
LGTM

PR #1421, opened 2026-05-08
[Quantization] MSE-calibrate every per-expert weight in fused-experts MoE

cjluo-nv:
### What does this PR do?

Type of change: Bug fix

Two-part fix for transformers 5.x fused-experts containers (Qwen3-MoE / Qwen3.5-MoE / Mixtral / DeepSeek / Kimi-K2.x ...) where weight quantizers live in `nn.ModuleList`s (`gate_up_proj_weight_quantizers`, `down_proj_weight_quantizers`):

1. **Per-expert weight iteration for calibration.** Add `_QuantFusedExperts.iter_weights_for_calibration` that yields per-expert `(weight_slice, quantizer)` pairs for both projections. The base impl uses singular `*_weight_quantizer` and silently skips fused-experts modules, so weight-only calibration paths never reached per-expert quantizers.

2. **`mse_calibrate` refactor.**
   - Add `_bootstrap_uncalibrated_weight_quantizers` after `max_calibrate` to populate `_amax` on quantizers the forward pass didn't reach (dead MoE experts that received no calibration tokens). Runs the existing calibrator on the weight slice surfaced by `iter_weights_for_calibration`.
   - Replace the singular-only `weight_attr_names` discovery + `getattr`-by-name walk with an `iter_weights_for_calibration` walk done inside each parent module's `enable_weight_access_and_writeback` context, so MSE processes every per-expert quantizer (active and dead) and remains FSDP-safe.

Without this, the export-time fallback in `_export_fused_experts` derived separate gate/up amaxes from each half of the fused weight, breaking the `gate==up` `weight_scale_2` invariant on dead experts.

Also includes:
- `_sanitize_generation_config_for_save` in `unified_export_hf` — coerces `do_sample=True` when an upstream `generation_config.json` has `top_k`/`top_p` set, so newer transformers' strict validate doesn't block `save_pretrained`.
- Small companion plumbing in `moe_utils.py`, `tensor_quantizer.py`, and `core_utils.py` to support the per-expert iteration and bootstrap path.

### Usage

```python
from modelopt.recipe import load_config

# Recipe `nvfp4_experts_only_mse-kv_fp8_cast` (already on main) now correctly
# MSE-calibrates every per-expert weight quantizer in fused-experts MoE models.
cfg = load_config("general/ptq/nvfp4_experts_only_mse-kv_fp8_cast")
mtq.quantize(model, cfg, forward_loop=calibration_forward_loop)
```

### Testing

**Original validation — Qwen3.5-122B-A10B with `nvfp4_experts_only_mse-fp8_cast_kv`:**
- **Before:** 1/12288 (layer 38 expert 69) `gate \!= up`; 0 weights MSE-calibrated.
- **After:** 0/12288 mismatches; 24576 weights MSE-calibrated; ~4.2 min.

**End-to-end pipeline validation — Qwen3.5-35B-A3B (40 layers × 256 experts × 2 projections = 20,480 per-expert weight quantizers), TRT-LLM 1.3.0rc13 + transformers 5.6 docker, single B200:**

| | Path A (4-sample calib, deliberately undercalibrated) | Path B (zero forward-pass tokens) |
|---|---|---|
| Per-expert weight quantizers calibrated | 20,480 / 20,480 | 20,480 / 20,480 |
| Missing `_amax` | 0 | 0 |
| All-zero `_amax` | 0 | 0 |
| `mtq.quantize` time | 25–34 s | 23 s |

- **Cross-path diff:** every per-expert weight amax matches **bit-for-bit** between the two paths (`n=20480 exact=20480 diff=0 max_rel=0`). With 8/256 experts routed per token and 4 calib samples, almost all experts are "dead" in Path A. Bootstrap fills them from `max(|weight|)`, MSE searches deterministically from there → identical to Path B which bootstraps everything.
- **Export to HF NVFP4 checkpoint** succeeded (~95 s, 22 GB checkpoint). Resulting `generation_config.json` has `do_sample: true` (upstream had `top_k=20` + `top_p=0.95` which would have failed strict validate).
- **TRT-LLM inference loaded the checkpoint and generated text:** `"Born in north-east France, Soyer trained as a"` → `" tailor. Demonstrating his craft at a young age, at 20 he moved to Paris at the requests of the noble people of Picardy."` (coherent grammar; factually wrong as expected with 4-sample calib, but no NaN/Inf in logits, no scale-mismatch crash). 92 GB GPU memory used.

- Did you write any new necessary tests?: ❌ <\!-- relies on existing recipe-level integration coverage; verified end-to-end on Qwen3.5-122B-A10B and Qwen3.5-35B-A3B + TRT-LLM 1.3.0rc13 -->
- Did you get Claude approval on this PR?: ❌ <\!-- will run \`/claude review\` -->

### Additional Information

Follow-up to PR #1407 (MSE+FP8-cast-KV recipes). The recipe YAML files landed there; this PR fixes the calibration codepath so the MSE recipes actually exercise per-expert weight quantizers in fused-experts MoE containers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed generation configuration validation for HuggingFace model exports.
  * Improved handling of quantization shape mismatches during expert weight export.

* **New Features**
  * Enhanced calibration process with automatic population of missing expert quantizers.
  * Added grouped quantizer synchronization for improved multi-expert quantization.

* **Tests**
  * Added regression tests for fused expert export and calibration correctness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1421)

### DIFF
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -64,7 +64,97 @@ __all__ = [
     "max_calibrate",
     "smoothquant",
     "svdquant",
+    "sync_grouped_weight_global_amax",
 ]
+
+
+# Sibling weight-quantizer name groups whose ``global_amax`` should share an
+# FP8 scale-of-scales. All members of a group sit under the same parent module
+# (e.g. one self-attention or one MLP block) and either consume the same input
+# tensor or get fused at deployment, so a divergent global_amax across siblings
+# would split their FP8 grids and skew the round.
+_GROUPED_WEIGHT_QUANTIZER_PATTERNS: tuple[tuple[str, ...], ...] = (
+    # Standard self-attention (skipped for fused qkv_proj — single weight).
+    ("q_proj", "k_proj", "v_proj"),
+    # Gated MLP, modern naming (Llama / Qwen / Mistral / etc.).
+    ("gate_proj", "up_proj"),
+    # Gated MLP, older Mixtral-style naming.
+    ("w1", "w3"),
+)
+
+
+def _is_calibrated_nvfp4_static_weight_quantizer(q) -> bool:
+    """True for an NVFP4-static weight quantizer that ``max_calibrate`` already
+    populated with a per-block ``_amax`` and that is currently enabled.
+    """
+    return (
+        isinstance(q, TensorQuantizer)
+        and not q._disabled
+        and q.is_nvfp4_static
+        and hasattr(q, "_amax")
+        and q._amax is not None
+    )
+
+
+def _collect_grouped_linears(model: nn.Module) -> list[list[nn.Module]]:
+    """Find groups of Linear-like submodules whose NVFP4-static weight quantizers
+    should share ``global_amax`` (Q/K/V under one attention parent; gate/up under
+    one MLP parent).
+    """
+    groups: list[list[nn.Module]] = []
+    wq_attr = quantizer_attr_names("weight").weight_quantizer
+    for parent in model.modules():
+        for sibling_names in _GROUPED_WEIGHT_QUANTIZER_PATTERNS:
+            members: list[nn.Module] = []
+            for n in sibling_names:
+                child = getattr(parent, n, None)
+                if child is None:
+                    continue
+                wq = getattr(child, wq_attr, None)
+                if _is_calibrated_nvfp4_static_weight_quantizer(wq):
+                    members.append(child)
+            if len(members) >= 2:
+                groups.append(members)
+    return groups
+
+
+@torch.no_grad()
+def sync_grouped_weight_global_amax(model: nn.Module) -> int:
+    """Sync ``global_amax`` across sibling NVFP4-static weight quantizers.
+
+    For each group of siblings (Q/K/V projections under one attention parent;
+    gate/up — a.k.a. ``w1``/``w3`` — under one MLP parent) unifies the
+    NVFP4 ``global_amax`` so the per-block FP8 round picks scales against a
+    consistent FP8 grid across the group during MSE / local-Hessian search.
+
+    Reuses :func:`modelopt.torch.export.quant_utils.preprocess_linear_fusion`
+    (whose ``NVFP4StaticQuantizer`` branch performs the same
+    ``max(stack(global_amax))`` unification at export time). To call it before
+    MSE, this helper first promotes each grouped weight quantizer to
+    :class:`NVFP4StaticQuantizer` with its local ``global_amax`` (=
+    ``reduce_amax(_amax)``); ``preprocess_linear_fusion`` then unifies in
+    place.
+
+    Must be called after ``max_calibrate`` has populated each weight
+    quantizer's ``_amax``. Idempotent. Returns the number of groups synced.
+    """
+    from modelopt.torch.export.quant_utils import preprocess_linear_fusion
+
+    n_groups = 0
+    for group in _collect_grouped_linears(model):
+        # Promote each member's weight quantizer so `preprocess_linear_fusion`
+        # sees post-conversion NVFP4StaticQuantizers (its NVFP4 branch reads
+        # `global_amax`, which only exists post-promotion).
+        wq_attr = quantizer_attr_names("weight").weight_quantizer
+        for child in group:
+            wq = getattr(child, wq_attr)
+            if not isinstance(wq, NVFP4StaticQuantizer):
+                local_global = reduce_amax(wq._amax, axis=None)
+                NVFP4StaticQuantizer.from_tensor_quantizer(wq, global_amax=local_global)
+        preprocess_linear_fusion(group)
+        n_groups += 1
+    return n_groups
+
 
 CalibratorFactory: TypeAlias = Callable[
     [torch.Tensor, int | tuple | list | None, Callable[..., torch.Tensor]], _Calibrator
@@ -349,6 +439,13 @@ def mse_calibrate(
     # Step 1: First get initial amax using max calibration
     max_calibrate(model, forward_loop, distributed_sync)
 
+    # Step 1b: Sync global_amax across sibling NVFP4-static weight quantizers
+    # (q/k/v_proj under one attention block; gate/up — a.k.a. w1/w3 — under one
+    # MLP block) so their FP8 scale-of-scales matches and the per-block FP8
+    # round uses a consistent grid. No-op when there are no sibling groups
+    # (e.g. fused QKV / fused gate_up_proj).
+    sync_grouped_weight_global_amax(model)
+
     # Step 2: Replace calibrators with MseCalibrator for enabled quantizers
     # and identify weight quantizers
     weight_quantizers = []
@@ -360,19 +457,16 @@ def mse_calibrate(
                 # Get the initial amax from max calibration
                 initial_amax = module._amax.clone().detach()
 
-                is_nvfp4_static = (
-                    module.is_static_block_quant
-                    and module._num_bits == (2, 1)
-                    and module._block_sizes is not None
-                    and module._block_sizes.get("scale_bits") == (4, 3)
-                )
+                is_nvfp4_static = module.is_nvfp4_static
 
                 if is_nvfp4_static:
-                    # Compute and set global_amax
-                    global_amax = reduce_amax(initial_amax, axis=None)
-
-                    # Convert to NVFP4StaticQuantizer in-place
-                    NVFP4StaticQuantizer.from_tensor_quantizer(module, global_amax=global_amax)
+                    # If sync_grouped_weight_global_amax already promoted this
+                    # quantizer (it's a sibling in a Q/K/V or gate/up group),
+                    # its global_amax has been unified across the group; just
+                    # leave it. Otherwise convert + set local global_amax.
+                    if not isinstance(module, NVFP4StaticQuantizer):
+                        global_amax = reduce_amax(initial_amax, axis=None)
+                        NVFP4StaticQuantizer.from_tensor_quantizer(module, global_amax=global_amax)
 
                 if fp8_scale_sweep:
                     # Check if backend has a registered custom calibrator factory.
@@ -612,6 +706,11 @@ def local_hessian_calibrate(
     print_rank_0("local_hessian: Running max calibration for all quantizers...")
     max_calibrate(model, forward_loop, distributed_sync)
 
+    # Sync global_amax across sibling NVFP4-static weight quantizers
+    # (q/k/v_proj, gate/up_proj a.k.a. w1/w3) so the FP8 scale-of-scales
+    # is consistent across the group. Idempotent; no-op when fused.
+    sync_grouped_weight_global_amax(model)
+
     # Setup helpers for all quantized linear modules
     name_to_module = dict(model.named_modules())
     weight_quantizers_info = []
@@ -666,14 +765,9 @@ def local_hessian_calibrate(
 
             return xq
 
-        is_nvfp4_static = (
-            weight_quantizer.is_static_block_quant
-            and weight_quantizer._num_bits == (2, 1)
-            and weight_quantizer._block_sizes is not None
-            and weight_quantizer._block_sizes.get("scale_bits") == (4, 3)
-        )
+        is_nvfp4_static = weight_quantizer.is_nvfp4_static
 
-        if is_nvfp4_static:
+        if is_nvfp4_static and not isinstance(weight_quantizer, NVFP4StaticQuantizer):
             global_amax = reduce_amax(initial_amax, axis=None)
             NVFP4StaticQuantizer.from_tensor_quantizer(weight_quantizer, global_amax=global_amax)
 
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -52,7 +52,6 @@ from .utils import (
     promote_nvfp4_static_quantizers,
     quantizer_attr_names,
     reduce_amax,
-    weight_attr_names,
 )
 from .utils.calib_utils import _GPTQ_HELPER_REGISTRY, GPTQHelper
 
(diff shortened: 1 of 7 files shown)


meenchen (commented):

Follow-up bug fix to #1407 for dead-expert MSE calibration in fused-experts MoE. The design is not a new subsystem — the new `sync_grouped_weight_global_amax` composes on the existing `preprocess_linear_fusion` from `quant_utils.py`, and `iter_weights_for_calibration` extends an already-existing base-class hook. The fix targets a real correctness bug (gate/up `weight_scale_2` divergence on experts that received no calibration tokens), backed by end-to-end Qwen3.5-122B numbers (0/12288 vs 1/12288 mismatches).

Reasons I did not approve:

1. **No unit tests for non-trivial new behavior.** The PR body explicitly opts out ("Did you write any new necessary tests?: ❌"). `tests/unit/torch/quantization/plugins/test_fused_experts.py` already has good infrastructure (`_SyntheticFusedExperts`, `TestFusedExpertsCalibration`) that could exercise:
   - `_QuantFusedExperts.iter_weights_for_calibration` yielding `num_experts * 2` pairs.
   - `_bootstrap_uncalibrated_weight_quantizers` populating `_amax` on never-routed experts (simulate by only routing to a subset in `forward_loop`).
   - The new `_export_fused_experts` per-block amax reshape path for NVFP4 (the existing `test_uncalibrated_expert_gate_up_share_amax` covers the per-tensor fallback but not the new `amax.numel() % fused_total == 0` reshape branch).
   - `sync_grouped_weight_global_amax` unifying `global_amax` across a Q/K/V sibling group.

2. **`_sanitize_generation_config_for_save` silently mutates user state.** It flips `do_sample=False → True` on the model's live `generation_config` whenever `top_k`/`top_p` is set, and this mutation is persisted to the exported `generation_config.json`. In practice this is a semantic no-op for greedy decoding, but the change is invisible to the caller. Consider (a) emitting a warning naming the fields being rewritten, or (b) doing the mutation on a copy that's scoped to the `save_pretrained` call only.

3. **`_GROUPED_WEIGHT_QUANTIZER_PATTERNS` is architecture-name heuristic.** The hardcoded tuple covers Llama/Qwen/Mistral/Mixtral but will silently miss any model using different attribute names (`wqkv`, fused `qkv_proj`, DeepSeek naming variants, etc.) — grouped unification would just not run and export would fall back to per-module amax. Worth either documenting this as a known limitation in the docstring or logging when a model has NVFP4-static quantizers but produces zero groups.

Minor: `sync_grouped_weight_global_amax` is added to `__all__` as public API, but given the hardcoded sibling-name heuristic it's really an internal helper; consider dropping it from `__all__` or renaming to `_sync_...`.

No licensing changes. Size is fine (+195/-71 across 6 files). Happy for a human reviewer with MoE export context to make the final call on the testing gap and the generation_config mutation policy.

meenchen (commented):

Re-review of #1421 (dead-expert MSE calibration for fused-experts MoE).

**Previous critical comments — addressed:**
- ✅ `_bootstrap_uncalibrated_weight_quantizers` now runs inside `enable_weight_access_and_writeback(module, model, name_to_module)` and the blanket try/except was removed (addresses CodeRabbit FSDP-safety concern).
- ✅ Unit tests added in `tests/unit/torch/quantization/plugins/test_fused_experts.py`: `test_bootstrap_populates_dead_expert_quantizers` exercises the dead-expert path end-to-end including the `max(|weight|)` invariant, and `test_per_block_amax_reshape_for_fused_export` guards the new NVFP4 per-block amax reshape branch in `_export_fused_experts`. Addresses meenchen's main nudge reason.
- ✅ The new helpers are underscore-prefixed and not exported in `__all__` (addresses the "public API for hardcoded heuristic" concern).
- ✅ `_GROUPED_WEIGHT_QUANTIZER_PATTERNS` limitation is documented in the `_sync_grouped_weight_global_amax` docstring and a warning is emitted when matching fails despite NVFP4-static quantizers existing.

**Still unresolved (minor, non-blocking individually):**
- ❌ `_sanitize_generation_config_for_save` still mutates `model.generation_config.do_sample` in-place with no restore. If `save_pretrained` raises, `_unpatch_revert_weight_conversion` still runs (it's in `finally`) but `do_sample` is never reverted — callers reusing the model object silently get `do_sample=True`. A try/finally-scoped restore (captured-original-value pattern) was suggested twice by CodeRabbit and not applied. Consider wrapping in the existing `try/finally` around `model.save_pretrained`.
- ❌ `_sync_grouped_weight_global_amax` warning is still gated only on `has_nvfp4_static`, so models with only standalone NVFP4 linears (no Q/K/V or gate/up siblings at all) will emit a misleading warning every time `mse_calibrate`/`local_hessian_calibrate` runs. A guard on "candidate siblings exist" would make it actionable.
- ❌ No MSE end-to-end test — the bootstrap path is only exercised via the direct helper, not via `mse_calibrate(..., algorithm="mse")`. If a future refactor stops calling the helper from `mse_calibrate`, the test won't catch it.

**Design check:** not a new subsystem. `_sync_grouped_weight_global_amax` composes on `preprocess_linear_fusion` from `quant_utils.py`; `iter_weights_for_calibration` extends an existing base-class hook; new helpers are small and localized.

No licensing changes. Size OK (+423/-71, 7 files). Core correctness fix looks right and is backed by end-to-end Qwen3.5-122B/35B validation. Flagging for human sign-off on the three minor items above — all three are stylistic/defensive rather than correctness bugs, but were called out explicitly by the previous reviewers and not acted on.

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,127 @@
...
+                ):
+                    members.append(child)
+            if len(members) >= 2:
+                groups.append(members)
+    return groups
+
+
+@torch.no_grad()
+def _bootstrap_uncalibrated_weight_quantizers(model: nn.Module) -> int:
+    """Populate ``_amax`` from weights for quantizers the forward pass didn't reach.
```
jenchen13: @realAsma After PTQ, if there are uncalibrated amaxes I thought we intentionally want to throw an error/warning and tell user to increase number of samples/seq length. this changes the behavior to automatically fill in the amaxes without any logging/warning which seems problematic.

this docstring is also misleading -- it sounds like you are loading amax from the weights but should say that you are recalibrating weights with missing `_amax`

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,127 @@
...
+                NVFP4StaticQuantizer.from_tensor_quantizer(
+                    wq, global_amax=reduce_amax(wq._amax, axis=None)
+                )
+        preprocess_linear_fusion(group)
+        n_groups += 1
+
+    if n_groups == 0:
+        # Surface architectures whose Q/K/V or gate/up siblings don't match the
+        # pattern list — without this, sibling-sync is a silent no-op.
+        has_nvfp4_static = any(
```
jenchen13: why not reuse the `is_nvfp4_static` method from `modelopt/torch/quantization/nn/modules/tensor_quantizer.py` ?

modelopt/torch/quantization/model_calib.py
```diff
@@ -412,52 +525,50 @@ def mse_calibrate(
                     quant_func=partial(_mse_quant_func, quantizer=module),
                 )
 
-    # Identify weight quantizers by checking if they have corresponding weight parameters
+    # Step 3: calibrate weight quantizers via iter_weights_for_calibration.
```
jenchen13, line 506: this comment is too long and specific to fused experts. remove mention of fused experts to make it more generic and in general reduce length of AI comments

modelopt/torch/quantization/model_calib.py
```diff
@@ -412,52 +525,50 @@ def mse_calibrate(
...
+                cal = weight_quantizer._calibrator
+                if cal.compute_amax() is not None:
+                    weight_quantizer.load_calib_amax()
 
-        weight_quantizer.enable_quant()
-        weight_quantizer.disable_calib()
+                weight_quantizer.enable_quant()
+                weight_quantizer.disable_calib()
 
-        # Synchronize ALL CUDA devices before resetting to ensure all async operations complete
```
jenchen13, line 448: why is this comment removed?

cjluo-nv: The old comment described the previous `weight_attr_names` lookup; with the refactor to `parent_module.iter_weights_for_calibration()` the call site is self-describing. Trimmed the replacement block to a single line per your other comment.

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,116 @@
...
+            f"Bootstrapped {n} weight quantizer(s) with no routed calibration tokens; "
+            f"their activation quantizers (if any) remain uncalibrated. "
+            f"Increase calib size/seq len to activate all experts.",
+            stacklevel=2,
+        )
+    return n
+
+
+@torch.no_grad()
+def _sync_grouped_weight_global_amax(model: nn.Module) -> int:
```
meenchen, line 143: This function is also useful for other algorithms like AWQ and GPTQ. We need to sync the amax for fused layers before some algorithm begins. cc @sychen52 on the design of fused modules.

meenchen (approved):
LGTM

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,116 @@
     "svdquant",
 ]
 
+
+# Sibling weight groups that should share an FP8 scale-of-scales after fusion.
+_GROUPED_WEIGHT_QUANTIZER_PATTERNS: tuple[tuple[str, ...], ...] = (
+    ("q_proj", "k_proj", "v_proj"),
+    ("gate_proj", "up_proj"),  # Llama/Qwen/Mistral
+    ("w1", "w3"),  # Mixtral
+)
```
Fridah-nv: We have existing fusion metadata in
   `modelopt/torch/export/quant_utils.preprocess_linear_fusion`, could we avoid adding new hard-coded tuples for fusion?

modelopt/torch/quantization/model_calib.py
```diff
@@ -346,32 +455,23 @@ def mse_calibrate(
     See :class:`MseCalibConfig <modelopt.torch.quantization.config.MseCalibConfig>` for
     details on the remaining arguments.
     """
-    # Step 1: First get initial amax using max calibration
+    # Step 1: max calibrate, bootstrap dead-expert weight quantizers,
+    # unify grouped NVFP4 global_amax so MSE sees a consistent FP8 grid.
     max_calibrate(model, forward_loop, distributed_sync)
+    _bootstrap_uncalibrated_weight_quantizers(model)
```
Fridah-nv, line 452: I wonder why the dead-expert weight quantizers does not break the MAX calibration export path with NVFP4 dynamic quantizer. i.e., does `max_calibrate` need this fix as well?
Does it make more sense to run `_bootstrap_uncalibrated_weight_quantizers` in  `max_calibrate`. (after weight_only_quantize/forward,
 before `promote_nvfp4_static_quantizers`)? Therefore other recipes (AWQ, GPTQ, local-hessian) gets the fix as well.

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,116 @@
...
+            wq = getattr(child, wq_attr)
+            if not isinstance(wq, NVFP4StaticQuantizer):
+                NVFP4StaticQuantizer.from_tensor_quantizer(
+                    wq, global_amax=reduce_amax(wq._amax, axis=None)
+                )
+        preprocess_linear_fusion(group)
+        n_groups += 1
+
+    if n_groups == 0 and any(_is_calibrated_nvfp4_static(m) for m in model.modules()):
+        warnings.warn(
```
Fridah-nv: If I'm understanding correctly, this "no group matched" warning will fire on every model that
   doesn't happen to use those three tuples — including `_QuantFusedExperts` itself, who gate vs. up halves are not in any group. Maybe include `_QuantFusedExperts` in the grouping pass?

modelopt/torch/quantization/model_calib.py
cjluo-nv: no. Max calibrate can be covered in the export time. @Edwardf0t1 added the logics that covers the unquantized experts.

modelopt/torch/quantization/model_calib.py
cjluo-nv: removed the warning. It's not important.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -900,6 +900,24 @@ def forward(self, *args, **kwargs):
         self._down_proj_linear = False
         return super().forward(*args, **kwargs)
 
+    def iter_weights_for_calibration(self):
```
jenchen13, line 903: this name is misleading, you are only iterating fused_expert weights, not all weights
```suggestion
    def iter_fused_expert_weights_for_calibration(self):
```

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,107 @@
...
+    return (
+        isinstance(q, TensorQuantizer)
+        and not q._disabled
+        and q.is_nvfp4_static
+        and getattr(q, "_amax", None) is not None
+    )
+
+
+def _collect_grouped_linears(model: nn.Module) -> list[list[nn.Module]]:
+    """Collect sibling groups (Q/K/V, gate/up) with calibrated NVFP4-static weight quantizers."""
```
jenchen13, line 80: if this is only specific to HF it also belongs in `plugins/huggingface.py`

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,107 @@
...
+            f"Bootstrapped {n} weight quantizer(s) with no routed calibration tokens; "
+            f"their activation quantizers (if any) remain uncalibrated. "
+            f"Increase calib size/seq len to activate all experts.",
+            stacklevel=2,
+        )
+    return n
+
+
+@torch.no_grad()
+def _sync_grouped_weight_global_amax(model: nn.Module) -> int:
```
jenchen13, line 143: can you move any HF specific functions to plugins/huggingface.py?

jenchen13 (approved):

modelopt/torch/quantization/model_calib.py
```diff
@@ -66,6 +65,107 @@
...
+                for child in (getattr(parent, n, None) for n in sibling_names)
+                if child is not None and _is_calibrated_nvfp4_static(getattr(child, wq_attr, None))
+            ]
+            if len(members) >= 2:
+                groups.append(members)
+    return groups
+
+
+@torch.no_grad()
+def _bootstrap_uncalibrated_weight_quantizers(model: nn.Module) -> int:
```
jenchen13, line 102: why do we need this during calibration time and how is this different from what we do at export time `_ensure_weight_quantizer_calibrated` (mte.quant_utils.py)? can we either reuse the existing code inside this helper function or just rely on export to fill in the missing amax?

modelopt/torch/quantization/plugins/huggingface.py
cjluo-nv: This is an override of the base `QuantModule.iter_weights_for_calibration` (defined in `modelopt/torch/quantization/nn/modules/quant_module.py:122`). Callers like `mse_calibrate`, `weight_only_quantize`, and `_bootstrap_uncalibrated_weight_quantizers` invoke `module.iter_weights_for_calibration()` polymorphically — they hit the overridden method by class dispatch, not by checking for fused-experts specifically. Renaming this override would break that dispatch: the base method would still be called by name, but `_QuantFusedExperts` instances would fall back to the base impl (which uses singular `*_weight_quantizer` and silently skips fused-experts modules — the exact bug this PR fixes).

modelopt/torch/quantization/model_calib.py
cjluo-nv: The helpers in this section operate on generic types (`TensorQuantizer`, `QuantModule`, `NVFP4StaticQuantizer`) and have no HF imports or HF-class isinstance checks:

- `_is_calibrated_nvfp4_static` — pure check on `TensorQuantizer.is_nvfp4_static` property; could apply to a Megatron model with NVFP4-static quantizers.
- `_collect_grouped_linears` — uses pattern names that happen to be common HF conventions (`q_proj`, `gate_proj`, `w1`, etc.) but operates on any `nn.Module`. The pattern data is now sourced from `export.layer_utils._GATE_UP_PAIRS` for the gate/up half.
- `_bootstrap_uncalibrated_weight_quantizers` — walks `QuantModule` instances and uses each module's `iter_weights_for_calibration` override polymorphically. Works for any registered fused-experts container (Megatron's would be covered too if registered as `QuantModule`).
- `_sync_grouped_weight_global_amax` — same; operates on the result of `_collect_grouped_linears`.

The HF plugin's role is registering specialized quant modules (like `_QuantFusedExperts` for HF `*MoeExperts` classes), not housing every helper that uses common naming conventions. Moving these would add a layering hop (`model_calib → plugins/huggingface`) for what's actually quantization-level logic invoked from `mse_calibrate`.

modelopt/torch/quantization/model_calib.py
cjluo-nv: Same reasoning as in the reply to your other comment about moving HF-specific functions: `_is_calibrated_nvfp4_static` is a pure `TensorQuantizer` property check with zero HF dependence, and `_collect_grouped_linears` uses pattern names that happen to be HF conventions but operates on generic `nn.Module`. Leaving these in `model_calib.py` keeps the sibling-sync helpers next to their only caller (`_sync_grouped_weight_global_amax` in `mse_calibrate`).

modelopt/torch/quantization/model_calib.py
cjluo-nv: Good question. Two parts:

**Why calibration-time, not export-time:** `_ensure_weight_quantizer_calibrated` runs at export and does a max-style fill-in for missing `_amax`. If we deferred dead-expert handling to export only:
- MSE's Step 3 walk would skip dead experts (they have `_amax=None` at MSE time)
- `_sync_grouped_weight_global_amax` would skip them (the `_is_calibrated_nvfp4_static` check requires `_amax` populated)
- `promote_nvfp4_static_quantizers` (in `max_calibrate`) wouldn't promote them either (it gates on `_amax` populated)
- Dead experts would receive only export-time max calibration — no MSE refinement, no sibling-amax sync, no NVFP4-static promotion

That defeats the PR's main feature ("MSE-calibrate every per-expert weight in fused-experts MoE"); dead experts would just go back to max-only. The bootstrap is timed before MSE/sync/promotion to bring dead experts into all three.

**Reuse of the existing helper:** reasonable as a follow-up DRY win, but they differ in detail today:
- `_ensure_weight_quantizer_calibrated` is per-quantizer; for `NVFP4StaticQuantizer` it uses `reduce_block_amax` + sets `global_amax` directly
- Our bootstrap walks all `QuantModule`s, uses each quantizer's existing calibrator (so any calibrator class works, not just max), runs inside `enable_weight_access_and_writeback` for FSDP/TP safety, and emits a single aggregate warning

Factoring out a shared inner step would save a handful of lines but pulls in another `export → quantization` inline import and switches the bootstrap from "calibrator-driven" to "tensor-driven". Happy to do as a small follow-up alongside moving `preprocess_linear_fusion` (also discussed elsewhere on this PR) — but not blocking the current fix.

PR #1422, opened 2026-05-09
Safetensor metadata mismatch fix in Mcore export

jinhangchoi:
### What does this PR do?

Type of change: Bug fix

In MCore export, shard metadata (`*.json`) and shard weights (`*.safetensors`) are produced from mutable shard maps and can be generated from different views of the same dict. In real Nemotron 3 Ultra PTQ runs, I observed MTP-related drift where metadata and shard contents were not aligned. This is plausible because MTP is stage-local (typically last-stage only), so per-rank shard contents are intentionally asymmetric.

The exact mutation interleaving is hard to prove from this code path alone, but the current implementation reads mutable shard maps across separate write steps, making metadata/weights consistency timing-sensitive. The issue is most visible with PP>1, where staggered per-shard writes widen the timing window between metadata and tensor-file generation.

This PR makes shard serialization deterministic in both paths:
- take a per-shard snapshot once,
- write `.safetensors` from that snapshot,
- write per-shard `.json` from the same snapshot.

Apply this consistently to:
- `save_safetensors_by_layer_index`
- `save_safetensors`
This guarantees shard JSON and shard safetensors cannot diverge due to late dict mutations.

### Usage

N/A

### Testing

- Reproduced in Nemotron 3 Ultra MCore PTQ export under PP>1 (multi-stage pipeline parallelism).

## Summary by CodeRabbit

* **Bug Fixes**
  * Model export now writes each binary shard immediately and uses a frozen per-shard/per-layer snapshot when generating its metadata. This prevents mismatches between shard contents and metadata, improving export consistency and reliability.

### DIFF
--- a/modelopt/torch/export/plugins/mcore_custom.py
+++ b/modelopt/torch/export/plugins/mcore_custom.py
@@ -251,6 +251,10 @@ def save_safetensors(state_dict, save_directory: str | os.PathLike):
         meta_filename = filename + ".json"
         ckpt_filename = filename + ".safetensors"
 
+        # Keep safetensors-first ordering so late shard updates are reflected
+        # in both the shard file and the shard metadata JSON.
+        save_file(tensors, save_directory + "/" + ckpt_filename, metadata={"format": "pt"})
+
         weight_map = {}
         local_total_size = 0
 
@@ -264,7 +268,6 @@ def save_safetensors(state_dict, save_directory: str | os.PathLike):
                 f,
                 indent=4,
             )
-        save_file(tensors, save_directory + "/" + ckpt_filename, metadata={"format": "pt"})
 
     # Barrier to ensure all ranks have written the metadata
     torch.distributed.barrier()
--- a/modelopt/torch/export/plugins/mcore_custom.py
+++ b/modelopt/torch/export/plugins/mcore_custom.py
@@ -251,14 +251,14 @@ def save_safetensors(state_dict, save_directory: str | os.PathLike):
         meta_filename = filename + ".json"
         ckpt_filename = filename + ".safetensors"
 
-        # Keep safetensors-first ordering so late shard updates are reflected
-        # in both the shard file and the shard metadata JSON.
-        save_file(tensors, save_directory + "/" + ckpt_filename, metadata={"format": "pt"})
+        # Freeze a per-shard snapshot so safetensors and JSON are emitted from the same view.
+        frozen_tensors = dict(tensors)
+        save_file(frozen_tensors, save_directory + "/" + ckpt_filename, metadata={"format": "pt"})
 
         weight_map = {}
         local_total_size = 0
 
-        for key, val in tensors.items():
+        for key, val in frozen_tensors.items():
             local_total_size += val.numel() * val.element_size()
             weight_map[key] = ckpt_filename
 
@@ -308,17 +308,17 @@ def save_safetensors_by_layer_index(
         meta_filename = filename + ".json"
         ckpt_filename = filename + ".safetensors"
 
-        # Keep safetensors-first ordering so late layer_state_dict updates are captured
-        # in the layer .safetensors shard and reflected in its per-layer metadata JSON.
+        # Freeze a per-layer snapshot so safetensors and JSON are emitted from the same view.
+        frozen_layer_state_dict = dict(layer_state_dict)
         save_file(
-            layer_state_dict,
+            frozen_layer_state_dict,
             save_directory + "/" + ckpt_filename,
             metadata={"format": "pt"},
         )
 
         weight_map = {}
         layer_total_size = 0
-        for key, val in layer_state_dict.items():
+        for key, val in frozen_layer_state_dict.items():
             tensor_size = val.numel() * val.element_size()
             layer_total_size += tensor_size
             weight_map[key] = ckpt_filename
--- a/modelopt/torch/export/plugins/mcore_custom.py
+++ b/modelopt/torch/export/plugins/mcore_custom.py
@@ -305,6 +305,12 @@ def save_safetensors_by_layer_index(
         meta_filename = filename + ".json"
         ckpt_filename = filename + ".safetensors"
 
+        # Write safetensors first, then build the per-layer meta JSON from the same dict.
+        # Order matters: any late mutations to layer_state_dict (e.g. MTP tensors added after
+        # the dict was first constructed) must be captured by both files.  Writing safetensors
+        # first ensures the JSON is always consistent with what is physically on disk.
+        save_file(layer_state_dict, save_directory + "/" + ckpt_filename, metadata={"format": "pt"})
+
         weight_map = {}
         layer_total_size = 0
         for key, val in layer_state_dict.items():
@@ -318,7 +324,6 @@ def save_safetensors_by_layer_index(
                 f,
                 indent=4,
             )
-        save_file(layer_state_dict, save_directory + "/" + ckpt_filename, metadata={"format": "pt"})
 
     # [TODO]: this global barrier needs to be replaced with something safer
     torch.distributed.barrier()
--- a/modelopt/torch/export/plugins/mcore_custom.py
+++ b/modelopt/torch/export/plugins/mcore_custom.py
@@ -305,11 +305,13 @@ def save_safetensors_by_layer_index(
         meta_filename = filename + ".json"
         ckpt_filename = filename + ".safetensors"
 
-        # Write safetensors first, then build the per-layer meta JSON from the same dict.
-        # Order matters: any late mutations to layer_state_dict (e.g. MTP tensors added after
-        # the dict was first constructed) must be captured by both files.  Writing safetensors
-        # first ensures the JSON is always consistent with what is physically on disk.
-        save_file(layer_state_dict, save_directory + "/" + ckpt_filename, metadata={"format": "pt"})
+        # Keep safetensors-first ordering so late layer_state_dict updates are captured
+        # in the layer .safetensors shard and reflected in its per-layer metadata JSON.
+        save_file(
+            layer_state_dict,
+            save_directory + "/" + ckpt_filename,
+            metadata={"format": "pt"},
+        )
 
         weight_map = {}
         layer_total_size = 0
--- a/tests/unit/torch/export/test_mcore_save_safetensors.py
+++ b/tests/unit/torch/export/test_mcore_save_safetensors.py
@@ -1,0 +1,54 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+
+import torch
+
+from modelopt.torch.export.plugins import mcore_custom
+
+
+def test_save_safetensors_by_layer_index_uses_single_snapshot(monkeypatch, tmp_path):
+    """Shard JSON and safetensors must be produced from the same key snapshot."""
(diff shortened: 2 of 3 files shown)


modelopt/torch/export/plugins/mcore_custom.py
```diff
@@ -305,6 +305,12 @@ def save_safetensors_by_layer_index(
         meta_filename = filename + ".json"
         ckpt_filename = filename + ".safetensors"
 
+        # Write safetensors first, then build the per-layer meta JSON from the same dict.
+        # Order matters: any late mutations to layer_state_dict (e.g. MTP tensors added after
```
jenchen13: can this AI comment be more concise?

jenchen13 (approved):

modelopt/torch/export/plugins/mcore_custom.py
jinhangchoi: Adjusted the comment a little bit

shengliangxu (approved):
LGTM

cjluo-nv (commented):

Small reorder (write safetensors shard before its per-layer JSON) that mitigates a real symptom seen in MCore PTQ on Nemotron, but I'm not fully convinced it addresses the root cause and there are gaps:

- **Root cause story is unclear from the code path I can see.** In `unified_export_megatron.py::save_pretrained`, `_get_mtp_state_dict()` is invoked synchronously *before* `save_safetensors_by_layer_index`, and the result is merged into `layer_state_dicts[num_layers]` before the barrier. There's no obvious place in the current code where MTP state is added between the two writes inside the per-layer loop. If the real bug is that something is still mutating `layer_state_dict` between `json.dump` and `save_file` (or vice versa), then this reorder only flips which side wins — after this PR, a late mutation will cause the JSON to advertise keys that were never written to the `.safetensors` shard, which is arguably worse than the original direction. A snapshot-then-write (e.g. `weight_map`/sizes computed once into local vars, then write safetensors and JSON from those, with no further dict reads) would actually be race-safe. Worth the author confirming what concretely mutates `layer_state_dict` after `save_file` was originally called.
- **Sibling function not updated.** `save_safetensors` (just above, used for the non-per-layer path) has the exact same JSON-then-safetensors ordering. If the ordering matters here, it presumably matters there too — please update both for consistency, or document why only the per-layer path is affected.
- **No test.** The author checked "Did you write any new necessary tests? ❌". A regression test that exercises the MTP-included shard and asserts that every key in the per-layer index JSON is actually present in the corresponding `.safetensors` would lock this down.
- **No CHANGELOG entry** (also unchecked).

jinhangchoi:
- Updated PR description, and add `save_safetensors` update.
- Meanwhile, regression test is a little bit difficult to design. Need some brainstrom.
- As bugfix, I don't think we need CHANGELOG here

cjluo-nv (commented):

Re-review focused on whether cjluo-nv's first two pieces of feedback have been addressed.

- **Feedback #1 — root cause / snapshot-then-write: NOT addressed.** :speech_balloon: Author updated PR description and reordered the writes, but the underlying concern was that a bare reorder only flips which side "wins" a race — if `layer_state_dict` is mutated between `save_file` and the metadata loop, the JSON will now advertise keys that were never written to the `.safetensors` shard (arguably worse than the original direction). The race-safe fix suggested by both cjluo-nv and CodeRabbit was to snapshot the dict once (e.g. `frozen = dict(layer_state_dict)`) and drive both `save_file` and the `weight_map` loop from that frozen view. The current diff at `mcore_custom.py:308-329` still calls `save_file(layer_state_dict, ...)` and then iterates the live `layer_state_dict.items()` — same in `save_safetensors` at lines 254-268 with `tensors`. So if the author's hypothesis about late mutation is correct, the consistency hole is still open; if it's wrong, it'd be good to hear what concretely was observed to drift on Nemotron 3 Ultra. Worth a human reviewer pushing on this before merge.

- **Feedback #2 — sibling `save_safetensors` not updated: addressed.** :speech_balloon: The new commit applies the same reorder to `save_safetensors` (lines 254-256), so both paths are consistent. (The snapshot caveat from #1 applies to both, though.)

For completeness on the other items the author replied to: regression test is still missing (author says it's hard to design), and CHANGELOG was declined as "bugfix doesn't need it" — leaving those as-is for the human owner to weigh in on.

jinhangchoi:
I agree `snapshot-then-write` is the generic race-safe pattern if in-loop mutation is expected. In this PR, the observed MCore issue is better characterized as an ordering/coherence problem: weak ordering guarantees between shard data and shard metadata generation, not a proven same-process mutation between `save_file` and the subsequent metadata loop.

Under PP>1, MTP is stage-local (typically last-stage only), and staggered shard emission makes this ordering sensitivity more visible. `_get_mtp_state_dict()` is merged synchronously before `save_safetensors_by_layer_index`, and we do not currently have an identified concurrent mutator inside the writer loop. The intent of this change is therefore to standardize `safetensors -> shard JSON` ordering across both writer paths to remove the stale-metadata-first failure mode I observed.
If we want strict future-proof determinism regardless of potential mutators, explicit finalize-then-serialize (or snapshot-at-boundary) is the stronger follow-up.

cjluo-nv (commented):

Re-review: sibling-function fix landed, but the core snapshot-then-write concern is still open and worth a human sign-off before merge.

- :speech_balloon: Author replied (PR comment): *"snapshot-then-write is the generic race-safe pattern if in-loop mutation is expected... the observed MCore issue is better characterized as an ordering/coherence problem, not a proven same-process mutation between `save_file` and the metadata loop... explicit finalize-then-serialize is the stronger follow-up."* — still warrants human sign-off because the current diff at `mcore_custom.py:311-329` (and the matching `save_safetensors` at 254-268) still passes the live `layer_state_dict`/`tensors` to `save_file` and then re-iterates the same live dict to build `weight_map`. If the author's "no in-loop mutator" assumption holds, the reorder alone fixes the symptom; if it doesn't, the new ordering means the JSON can advertise keys that aren't in the `.safetensors` shard — arguably worse than the original direction. A two-line `frozen = dict(layer_state_dict)` snapshot would close this regardless of which hypothesis is right, and the author already agrees it's the more defensible pattern.
- :speech_balloon: Sibling `save_safetensors` reorder: addressed in this revision (lines 254-256). Both paths are now consistent.
- :speech_balloon: Author replied: *"regression test is a little bit difficult to design. Need some brainstorm."* — flagging anyway because the hypothesized failure mode (per-layer JSON listing keys not present in the shard) is straightforward to assert in a unit test by mocking `save_file`/mutating the dict between writes; a real distributed Nemotron repro isn't required.
- :speech_balloon: Author replied: *"as bugfix, I don't think we need CHANGELOG here."* — leaving for the human owner to weigh in on per project policy.

jinhangchoi:
I would like to emphasize this is not a hypothetical issue. **It was observed in production export artifacts.**
```python
for key, val in layer_state_dict.items():
    weight_map[key] = ckpt_filename
with open(meta_filename, "w") as f:
    json.dump({"metadata": ..., "weight_map": weight_map}, f, indent=4)
save_file(layer_state_dict, ckpt_filename, metadata=...)   # <-- runs LAST
```
`weight_map` is built from the dict's keys at iteration time, the JSON is committed, then `save_file` writes the shard from whatever the dict contains now. Any late mutation lands in the binary but not in the per-layer JSON — and consequently not in rank 0's aggregated `model.safetensors.index.json`. I saw this in production exports under PP>1: the shards physically contained the keys; the index didn't list them.

The direction in your comment is inverted. By construction of the previous code before fix, the failure is "the JSON lists fewer keys than the shard contains." The shard is the superset, the index is a subset. The comment describes the opposite (JSON listing keys that aren't in the shard), which the previous code before fix can't produce. 

Regarding the regression test, I am supportive of adding a unit test along the lines you described, but to be clear about what it covers. This would be a freeze-invariant test, not a true regression for the production failure. The production failure requires PP>1 plus the MTP load path that mutates `layer_state_dicts[num_layers]` **after** construction. Neither is exercised by a mocked single-process test. It guards the invariant the freeze is there to guarantee.

The actual end-to-end repro is the original MCore export that motivated this fix. I would like to call this out so we don't oversell what the test asserts.

tests/unit/torch/export/test_mcore_save_safetensors.py
```diff
@@ -0,0 +1,54 @@
...
+        # Simulate a late mutation against the original source dict (not the writer snapshot).
+        layer_state_dicts[1][late_key] = torch.ones(1, dtype=torch.float32)
+
+    monkeypatch.setattr(mcore_custom, "save_file", _fake_save_file)
+    monkeypatch.setattr(torch.distributed, "barrier", lambda: None)
+    monkeypatch.setattr(torch.distributed, "get_rank", lambda: 0)
+
+    mcore_custom.save_safetensors_by_layer_index(
+        layer_state_dicts=layer_state_dicts,
+        total_layers=1,
```
cjluo-nv, line 40: > _Bot comment._

`tmp_path` is a `pathlib.Path`, but `save_safetensors_by_layer_index` uses string concatenation: `save_directory + "/" + ckpt_filename`. `Path + str` raises `TypeError`, so this test should fail before reaching the assertions. Please either pass `str(tmp_path)` here, or (better) switch the production code to `os.path.join(save_directory, ckpt_filename)` / `Path(save_directory) / ckpt_filename` so both `str` and `Path` callers work.

cjluo-nv (commented):

Re-review.

Addressed since the previous round:
- :speech_balloon: Snapshot-then-write **applied** in both `save_safetensors` (lines 254-256) and `save_safetensors_by_layer_index` (lines 311-317). `frozen_tensors = dict(tensors)` / `frozen_layer_state_dict = dict(layer_state_dict)` is now the single source for both `save_file` and the `weight_map` loop, so JSON and shard cannot diverge regardless of late mutations. This closes the main concern from cjluo-nv and CodeRabbit.
- :speech_balloon: Sibling `save_safetensors` reorder + freeze: applied.
- :speech_balloon: Author replied that a regression test is hard to design — a unit test was nonetheless added in `tests/unit/torch/export/test_mcore_save_safetensors.py` that mocks `save_file` and mutates the source dict mid-write to assert the JSON does not pick up the late key. This is exactly the freeze-invariant test that was suggested. 👍

Still warrants human sign-off:
- The new test passes `tmp_path` (a `pathlib.Path`) as `save_directory`, but the production code does `save_directory + "/" + ckpt_filename`, which is `Path + str` and raises `TypeError`. The test as written looks like it would fail at the first `save_file` call before reaching the assertions. Either I'm missing something or the test wasn't run — please double-check by either casting to `str(tmp_path)` in the test or switching the source to `os.path.join` / `Path /`. (Same shape issue would also bite `save_safetensors` if it were ever exercised this way.)
- :speech_balloon: Author replied: *"as bugfix, I don't think we need CHANGELOG here."* — deferred to human per project policy.

tests/unit/torch/export/test_mcore_save_safetensors.py
jinhangchoi: Thanks. Fixed & tested it locally.

tests/unit/torch/export/test_mcore_save_safetensors.py
```diff
@@ -0,0 +1,54 @@
...
+# limitations under the License.
+
+import json
+
+import torch
+
+from modelopt.torch.export.plugins import mcore_custom
+
+
+def test_save_safetensors_by_layer_index_uses_single_snapshot(monkeypatch, tmp_path):
```
jenchen13, line 23: if the metadata mismatch only happens with PP > 1 should this belong in tests/gpu_megatron and use PP > 1?

jinhangchoi: Thank you for the pointer. I just noticed `test_unified_export_megatron.py` :) Will try to add a test case there

jinhangchoi: This test itself should be kept for `modelopt-review`'s regression coverage.
Created new test, `test_unified_export_megatron_pp2_mtp_metadata_matches_shards` for PP==2

meenchen (approved):
LGTM

meenchen:
/ok to test [ef71f1d](https://github.com/NVIDIA/Model-Optimizer/pull/1422/commits/ef71f1d6a23b99b521cfc573bce0e3d7bd03f3cb)

meenchen:
/ok to test https://github.com/NVIDIA/Model-Optimizer/commit/799bce3cf985c29a4507c1ebc8e33c56424253a6

jenchen13:
/ok to test [64c73f9](https://github.com/NVIDIA/Model-Optimizer/pull/1422/commits/64c73f90040aae41d9b60c978e0053ebb5a6d3fa)

PR #1417, opened 2026-05-08
EAGLE3 new model support: pipeline configs, triage docs, and Ministral-3 fixes

yeyu-nvidia:
## Summary

EAGLE3 automation triage work (OKR-30): testing the 4-step EAGLE3 offline pipeline against 12 new model architectures, documenting failure modes, and fixing issues found.

### Code fixes (modelopt)

| File | Change |
|------|--------|
| `modelopt/torch/speculative/utils.py` | Extend VLM detection in `load_vlm_or_llm` to check `text_config`/`llm_config` attrs (catches `mistral3` models) |
| `modelopt/torch/speculative/plugins/modeling_fakebase.py` | Add `consolidated.safetensors` fallback for checkpoints with incomplete HF shards |
| `modelopt/torch/export/plugins/hf_spec_configs.py` | Set `use_cache=True` in EAGLE export templates (fixes strict `huggingface_hub` validation) |

### Pipeline infrastructure

- `examples/speculative_decoding/pipeline/eagle3/` — pipeline scripts and configs:
  - `offline_training.sh` — training + export with runtime patches for older container modelopt
  - `dump_offline_data_vllm.sh` — vLLM-based hidden state extraction (with speculators compat patches)
  - `dump_offline_data.sh`, `dump_offline_data_hf.sh` — alternative dump paths
  - 18 quick-fail-check YAMLs for 12 models
  - 4 standalone task1 YAMLs

### Documentation

- `eagle3_triage_chart.md` — model test matrix, triage decision tree, per-model results, failure catalog
- `eagle3_new_model_triage_guide.md` — step-by-step guide for triaging new models

### Model test results (as of 2026-05-27)

| Model | task_0 | task_1 | task_2 | task_3 | Blocker |
|-------|--------|--------|--------|--------|---------|
| Qwen3-8B | - | - | - | - | Reference (existing) |
| Kimi-K2.5 | - | - | - | - | Existing (GB200) |
| **Ministral-3-8B** | SKIP | PASS | PASS | FAIL | `use_cache=null` in export (fixed) |
| Ministral-3-14B | FAIL | - | - | FAIL | vLLM engine init fails |
| Qwen3.5-35B-A3B | TIMEOUT | - | - | - | Data synth too slow |
| gpt-oss-20b | FAIL | - | - | - | Tokenizer `HarmonyError` |
| Step-3.5-Flash | TIMEOUT | - | - | - | Data synth time limit |
[...]

PR #1429, opened 2026-05-11
feat(okr30): add EAGLE3 Claude Code skills for triage, validation, and new-model support

yeyu-nvidia:
## Summary

Adds four user-invocable Claude Code skills for the EAGLE3 offline pipeline (OKR 30 — Claude-assist experience).

- **`/eagle3-triage`**: Diagnose a failed pipeline run. Failure tables for all 4 tasks covering vLLM server startup, hidden state dump (3 backends: TRT-LLM / HF / vLLM), training crashes, and benchmark failures. New-model-specific issue checklist (VLMs, MoE, SWA, custom tokenizers).
- **`/eagle3-validate`**: Verify a completed run end-to-end. Artifact checks per task, AR threshold validation (≥ 2.1), structured validation report.
- **`/eagle3-new-model`**: Guided workflow for adding a new model. Architecture lookup, GB200 GPU/TP calculation, dump backend selection, full YAML template with correct public-launcher script paths.
- **`/eagle3-review-logs`**: Lightweight log reader. Finds `sbatch_*.out` files, reads all task logs, produces pass/fail summary with root causes and next steps.

Skills use public launcher paths (`common/eagle3/`, `common/vllm/`, etc.) and read `sbatch_*.out` files directly — no sandbox-specific tooling required.

## Related

- OKR 31 (triage chart + model configs): #1417

## Test plan

- [ ] Run `/eagle3-triage` against a known-failed experiment and verify it identifies the root cause
- [ ] Run `/eagle3-validate` against a passing experiment and verify AR check works
- [ ] Run `/eagle3-new-model` to generate a config for a new model and verify the YAML is correct
- [ ] Run `/eagle3-review-logs` and verify summary output matches actual log contents

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive EAGLE3 guides: onboarding new models and example config creation, a log-review workflow that generates structured pass/fail reports with root-cause one-liners and suggested fixes, a user-invocable end-to-end validation workflow for pipeline runs, and a triage/troubleshooting guide mapping common failure patterns to concrete remediation steps.


.claude/skills/eagle3-new-model/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+pipeline:
+  allow_to_fail: false
+  skip: false
+  note:
+
+  global_vars:
+    hf_model: /hf-local/<org>/<model>
+```
+
+### task_0 — Data synthesis (`common/vllm/query.sh`)
```
h-guo18: Theses steps should be obvious for agents to infer from existing launcher examples. I personally suggest letting agent infer from existing ones, rather than provide a general guidance here.
Reason: launcher examples are easier to test and maintain than md skills.


cc @ChenhanYu for discussion

.claude/skills/eagle3-new-model/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+| Requires `--trust-remote-code` | Add to task_0 vLLM args (before `--`) |
+| VLM / multimodal | Use `dump_offline_data_hf.sh` for task_1 |
+| Sliding window attention | Use `dump_offline_data_hf.sh` or `_vllm.sh` for task_1 |
+| MoE with large expert hidden dim | Increase `intermediate_size` in eagle_config.json |
+| Non-standard attention (MLA) | Verify `eagle_decoder_type` in the eagle3 recipe YAML |
+| Custom tokenizer (e.g., tiktoken) | Set `TIKTOKEN_RS_CACHE_DIR` env var in task_0 and task_1 |
+| NVFP4 quant model | task_0/task_3 use quant container; task_1/task_2 use BF16 base model — add `hf_model_bf16` global_var |
+| Model needs `trust_remote_code` at benchmark | Add `--trust-remote-code` to task_3 args |
+
+## Step 6 — Test with dry run
```
h-guo18: In my experience, I never told the agent to do dry-run, but it does basically everytime. I think the dry-run command maybe something obvious for agent to read from launcher interface.

.claude/skills/eagle3-new-model/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+
+Preview the resolved config before submitting:
+
+```bash
+uv run launch.py --yaml examples/<Org>/<Model>/hf_offline_eagle3.yaml --dryrun --yes -v
+```
+
+## Step 7 — Update triage chart
```
h-guo18: What's the relationship between the triage chart and pensive-intern's flow chart?

cc @ChenhanYu

.claude/skills/eagle3-new-model/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+| Total / active parameters | Model card |
+| Dense or MoE? | `config.json` → `num_experts`, `num_experts_per_tok` |
+| Attention type (MHA / GQA / MLA / SWA) | Model card |
+| Multimodal? (vision encoder) | Model card |
+| BF16 weight size (GB) | `total_params × 2 bytes` |
+| Special serving flags | vLLM docs, model README (`--trust-remote-code`, parsers) |
+
+## Step 2 — Calculate GPU requirements (OCI-HSG / GB200)
+
+OCI-HSG nodes: **4 GPUs × 192 GB HBM3e = 768 GB per node**
```
h-guo18: These cluster info is obvious for agent to infer. I don't think we need to hard-coded it here.

.claude/skills/eagle3-new-model/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+## Step 3 — Choose the hidden state dump backend
+
+| Backend | Script | When to use |
+|---------|--------|-------------|
+| vLLM | `common/eagle3/dump_offline_data_vllm.sh` | Default; broad coverage via vLLM + speculators |
+| HF | `common/eagle3/dump_offline_data_hf.sh` | VLMs, custom-code models, SWA attention |
+| TRT-LLM | `common/eagle3/dump_offline_data.sh` | Pure-text models with TRT-LLM support (needs `--tp`/`--moe-ep`) |
+
+Use **HF** when the model is a VLM or uses sliding window attention (TRT-LLM does not support these).
+Use **vLLM** for everything else as the default.
```
h-guo18: Curious what's the consideration here

.claude/skills/eagle3-review-logs/SKILL.md
```diff
@@ -0,0 +1,99 @@
+---
+name: eagle3-review-logs
```
h-guo18, line 2: I never used any skills, but my agent does basically the same thing when given the launcher yaml example alone. Any specific reason we need this?

.claude/skills/eagle3-triage/SKILL.md
```diff
@@ -0,0 +1,179 @@
+---
+name: eagle3-triage
```
h-guo18, line 2: Is this somehow functional overlap with pensive-intern? cc @ChenhanYu

.claude/skills/eagle3-review-logs/SKILL.md
yeyu-nvidia: The intent is triage-and-expose, not to replace human judgment: encode the launcher-specific surface (experiment dir layout, the 4-task mapping, acceptance-rate threshold, known per-model gotchas) so the agent reliably localizes and reports a failure for a human to act on. If your agent already does this well from the YAML alone, and pensive-intern covers the same surface, I'm happy to consolidate — drop `eagle3-review-logs` and keep only `eagle3-new-model`. cc @ChenhanYu

.claude/skills/eagle3-triage/SKILL.md
yeyu-nvidia: Possibly — if pensive-intern already covers EAGLE3 triage, there's functional overlap and I'd rather not duplicate. I'll check pensive-intern's scope; if it covers this, I'll drop `eagle3-triage` and keep only `eagle3-new-model` (config generation). cc @ChenhanYu

.claude/skills/eagle3-new-model/SKILL.md
yeyu-nvidia: Good question — I should reconcile these. If pensive-intern's flow chart already covers the triage decision tree, the chart here is redundant and should point at it instead of duplicating. cc @ChenhanYu

.claude/skills/eagle3-new-model/SKILL.md
yeyu-nvidia: Reasonable — agents do infer most of this from existing YAMLs. I'll trim the skill to the non-obvious bit (backend-selection heuristic: HF for VLM/SWA, vLLM as default, TRT-LLM for plain text) and let the agent infer the rest from the example configs rather than spelling out a general step list.

.claude/skills/eagle3-new-model/SKILL.md
yeyu-nvidia: Agreed — dry-run is something the agent reads from the launcher interface anyway. I'll drop the explicit dry-run step.

.claude/skills/eagle3-new-model/SKILL.md
yeyu-nvidia: Agreed — the OCI-HSG cluster sizing shouldn't be hardcoded here. I'll remove it and let the agent infer node/GPU sizing from existing examples + the launcher interface.

.claude/skills/eagle3-new-model/SKILL.md
yeyu-nvidia: No strong reason — this falls in the same bucket as the other "agent can infer this" notes. I'll trim it when I slim the skill down.

.claude/skills/eagle3-triage/SKILL.md
yeyu-nvidia: Checked Pensieve ([docs](https://omniml.gitlab-master-pages.nvidia.com/modelopt-site/pensieve/)) — it's the team's Slack memory/status bot (knowledge graph, weekly digests, JIRA *ticket* triage). It doesn't read EAGLE3 experiment logs, diagnose pipeline task failures, or validate runs, so there's no functional overlap with these skills, which operate on the launcher's `experiments/` output and the 4-task pipeline. So I'd keep `eagle3-triage`/`eagle3-review-logs` rather than consolidate. Happy to reconsider if you were thinking of a different tool.

.claude/skills/eagle3-new-model/SKILL.md
yeyu-nvidia: Checked Pensieve ([docs](https://omniml.gitlab-master-pages.nvidia.com/modelopt-site/pensieve/)) — it's the team's Slack memory/status bot (knowledge graph, weekly digests, JIRA *ticket* triage). It doesn't read EAGLE3 experiment logs, diagnose pipeline task failures, or validate runs, so there's no functional overlap with these skills, which operate on the launcher's `experiments/` output and the 4-task pipeline. So I'd keep `eagle3-triage`/`eagle3-review-logs` rather than consolidate. Happy to reconsider if you were thinking of a different tool.

.claude/skills/eagle3-new-model/SKILL.md
```diff
@@ -0,0 +1,49 @@
...
+
+These are the non-obvious knobs that vary per model:
+
+| Situation | What to change |
+|---|---|
+| Requires `--trust-remote-code` | Add to `task_0` vLLM args (before the `--` separator) and to `task_3` benchmark args |
+| VLM / multimodal | Use `dump_offline_data_hf.sh` for `task_1` |
+| Sliding window attention | Use `dump_offline_data_hf.sh` or `_vllm.sh` for `task_1` |
+| MoE with large expert hidden dim | Increase `intermediate_size` in `eagle_config.json` to match `moe_intermediate_size` |
+| Non-standard attention (MLA) | Verify `eagle_decoder_type` in the eagle3 recipe YAML |
```
h-guo18: Sicne there's no more one-model constraint in serving framework, eagle_decoder_type should be something free to choose, instead of something needs to be verified. I suggest remove this item.

yeyu-nvidia: Agreed — removed the "Non-standard attention (MLA) → verify `eagle_decoder_type`" row in `b626da4e5`. Since the serving framework no longer has a single-model constraint, `eagle_decoder_type` is a free choice, not something to verify here.

.claude/skills/eagle3-new-model/SKILL.md
```diff
@@ -0,0 +1,49 @@
...
+| Custom tokenizer (e.g. tiktoken) | Set `TIKTOKEN_RS_CACHE_DIR` env var in `task_0` and `task_1` |
+| NVFP4 quant model | `task_0`/`task_3` use the quant container; `task_1`/`task_2` train against the BF16 base — add an `hf_model_bf16` global_var |
```
h-guo18: This row doesn't seem clear to me. what is quant container? why do we have to do task1/2 against bf16 base? what is hf_model_bf16?

yeyu-nvidia: Fair — it was unclear, and it describes the NVFP4 flow which is itself unverified (the NVFP4 example was just reverted). Removed it for now. For context, the intent was: an NVFP4-quantized checkpoint is used for serving (task_0 synth / task_3 benchmark) in an NVFP4-capable container, while the hidden-state dump (task_1) and EAGLE3 training (task_2) run against the full-precision BF16 base weights — hence a second `hf_model_bf16` path. I'll re-add a clear, verified version of this once we actually validate an NVFP4 model end-to-end.

.claude/skills/eagle3-review-logs/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+## Step 0 — Find experiment logs
+
+Locate the experiment directory. The default is `experiments/` relative to the launcher root,
+or wherever `--job-dir` was pointed.
+
+```bash
+ls -td experiments/cicd/cicd_* | head -10
+```
+
+Each experiment has one subdirectory per task (0–3). Logs are `sbatch_*.out` files inside:
```
h-guo18: I think the file name is not deterministic. For example, with the launcher's local launch mode, there should not be any sbatch prefix. Instead of hard-coding these file names and pathes, can we make it more general and concise?

yeyu-nvidia: Good point — generalized it. The discovery now matches `*.out`/`*.log` instead of hardcoding `sbatch_*.out`, so it covers both Slurm (`sbatch_*.out`) and local Docker (`*.log`) launches, and I dropped the duplicated `find`:

```bash
find experiments/<exp_id>/ -type f \( -name '*.out' -o -name '*.log' \) | sort | while read -r f; do
  echo "=== $f ==="; tail -200 "$f"; echo
done
```

h-guo18 (commented):
for agent:

Could you run a pass on these skills, and check for: 
1. if there's any hard-code or dependency to existing implementation that may not applied later. If so, please try to eliminate to make the skills as general and easy-to-maintain as possible.
2. If some info is easy to observe or infer for agent during the task context, there should be no need to hard-code into skills. Only leave the high-value info that is otherwise hard to get automatically.

Please list such improvable items and go through with developer for approval.

yeyu-nvidia:
Ran the audit pass on the four skills against both criteria (no implementation-coupling that may break later; only keep high-value info that's hard to infer in-context). Applied in `30b717b81`:

**Removed stale / implementation-coupled references**
- Pointers to `EAGLE3_TRIAGE.md` / the triage chart (review-logs, triage, validate) → redirect to the team's internal triage tracker. (These docs are removed in #1623.)
- The `Qwen/Qwen3.5-35B-A3B` reference example in `eagle3-new-model` (deleted in #1623) → "copy a surviving example, e.g. Qwen3-8B for dense."
- The `speculators.VllmHiddenStatesGenerator` mention → vLLM's native hidden-state extractor.
- The `offline_training.sh` HF-Hub row (script no longer exists).

**De-hardcoded**
- Log discovery now matches `*.out`/`*.log` instead of `sbatch_*.out` in `eagle3-triage` and `eagle3-validate` (same fix you flagged for review-logs — local launch has no sbatch prefix).
- Dropped the baked-in GB200 "192 GB × 4 = 768 GB" sizing and the pinned vLLM version (`≥ v0.15.0`) — the agent infers sizing/containers from the examples.

**Trimmed inferable content**
- `eagle3-new-model`: removed the VLM / sliding-window rows that just restated the backend-selection table.
- `eagle3-triage`: removed rows an agent infers anyway (pip-install→network, generic ImportError); de-emphasized `eagle_decoder_type` (free to choose, not to verify, per your earlier note).

**Deliberately kept** (hard to get otherwise): the backend-selection heuristic, the "known benign patterns" table, the AR `>= 2.1` threshold and the exact `Average Acceptance Length {...}` log string, and the TIKTOKEN / MoE-`intermediate_size` gotchas. Let me know if you'd like any of these trimmed further.

.claude/skills/eagle3-validate/SKILL.md
```diff
@@ -0,0 +1,121 @@
...
+## Step 6 — Suggest next steps
+
+**If PASS:**
+- Record the verified result (and checkpoint path) in the team's internal triage tracker
+- This model is now a candidate to add as a launcher example in a dedicated PR
+
+**If FAIL:**
+- Identify which step or metric failed
+- Suggest running `/eagle3-triage` for diagnosis
+- If AR is close but below threshold, suggest:
```
h-guo18: Low AR could be anything. I suggest remove the following suggestion items, but give suggestions by agents based on specific scenarios.

h-guo18 (approved):

.claude/skills/eagle3-validate/SKILL.md
h-guo18: agent generally come up better suggestion beyond what we can enumerate here, in my experience

yeyu-nvidia: Agreed — removed the canned list. Step 6 now tells the agent to diagnose the specific cause of a low AR (training loss curve, data volume/quality, draft-head capacity, hyperparameters) and suggest fixes targeted to that scenario, rather than a generic checklist.

PR #1431, opened 2026-05-11
Save next_inputs.pt only for last two layers for layerwise save/restore

sugunav14:
### What does this PR do?

Type of change: Optimization (no API or on-disk-format change)

Reduces disk usage of layerwise calibration checkpoints by pruning stale next_inputs.pt files after each layer save. Without this change, every non-final transformer layer's next_inputs.pt accumulates on disk, dominating checkpoint size for long calibration runs (e.g. ~250 MB × ~80 layers ≈ 20 GB for a 70B model). On resume we only ever read the most recent one, so older copies are dead weight.

The fix is small and localized: save only latest layer's input.pt. The disk footprint of activation copies becomes O(1) (1 file) instead of O(N_layers). The static per-layer artifacts (weights.pt, quantizer_state.pt, output_meta.pt) are kept for every layer as before because full_restore needs them

### Usage
No API change required — pruning happens automatically when checkpoint_dir is used with layerwise_calibrate

```python
  mtq.calibrate(
      model,
      algorithm={
          "method": "gptq",
          "use_sequential": True,
          "checkpoint_dir": "/path/to/checkpoint",
      },
      forward_loop=forward_loop,
  )
```

### Testing
- test_full_run_creates_checkpoints — extended to assert layer 0's next_inputs.pt is gone after a 3-layer run while layer 1's remains.
- test_only_last_two_next_inputs_kept (new) — 5-layer walk-through verifying only the most recent next_inputs.pt is retained and all static files survive for every layer.
- test_resume_matches_full_run — uses a calib_func that raises during layer 1 to simulate a real crash; asserts the resumed run produces bit-identical weights to a reference full run.

## Summary by CodeRabbit

* **New Features**
  * More reliable resume: calibration now uses a single root-level resume artifact for restarting interrupted runs.

* **Refactor**
  * Atomic commit behavior and reduced per-layer temporary files to minimize on-disk clutter and partial-state exposure.
  * Resume now fails fast if the root resume artifact is missing.

* **Tests**
  * New and updated tests simulate mid-run crashes and verify atomic checkpointing and correct resume behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1431?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


shengliangxu (approved):
LGTM

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -640,6 +640,12 @@ def full_restore(self, layers: nn.ModuleList, model: nn.Module) -> None:
 
         print_rank_0(f"Checkpoint: restored {self.start_layer} previously calibrated layers")
 
+    def prune_old_next_inputs(self, layer_idx: int) -> None:
```
Fridah-nv:
```suggestion
    def _prune_old_next_inputs(self, layer_idx: int) -> None:
```

Fridah-nv (approved):

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma, line 1: @sugunav14  should we have just one global `next_inputs.pt`  and keep updating it instead of saving  `next_inputs.pt` to each `layer_folder/` and deleting it? (So no per-layer `next_inputs.pt`)

realAsma (commented):
@sugunav14  could you also add a warning to `layerwise`  calibration that this is still an experimental feature and backward compatibility is not guaranteed?

realAsma (approved):

PR #1465, opened 2026-05-12
fix: clamp dynamic NVFP4 FP8 scale export

ShawRong:
### What does this PR do?

Type of change: Bug fix

Clamp the dynamic NVFP4 per-block FP8 scale before casting it to `torch.float8_e4m3fn` during export.

The static quantizer-derived path already clamps the exported FP8 scale, but the dynamic path can still produce scale values larger than the finite FP8 E4M3 range. Since `torch.float8_e4m3fn` has no Inf representation, casting values such as `480` or `1000` produces NaN payloads in the exported scale tensor. This patch applies the same finite-range clamp to the dynamic path so exported dynamic NVFP4 scales remain finite.

### Usage

```python
# No API changes. Dynamic NVFP4 export now avoids NaN FP8 scale entries when
# the computed per-block scale exceeds the finite fp8_e4m3fn range.
```

### Testing

```bash
uv run pytest tests/gpu/torch/quantization/test_qtensor_cuda.py::TestQTensor::test_nvfp4_dynamic_export_fp8_scale_no_nan_when_scale_exceeds_fp8 -q
```

Result: `1 passed`.

Also ran pre-commit on the touched files:

```bash
uv run pre-commit run --files \
  modelopt/torch/quantization/qtensor/nvfp4_tensor.py \
  tests/gpu/torch/quantization/test_qtensor_cuda.py
```

Result: all applicable hooks passed.

### Additional Information

This complements the existing static NVFP4 scale clamp by applying the same safety guard to the dynamic scale export path.

## Bug Fixes
* Fixed NaN values in FP8 scale conversions during dynamic NVFP4 quantization export. Scale values now clamp properly when they exceed the valid FP8 range, preventing overflow and ensuring numerical stability in exported scale bytes.

## Tests
* Added regression test to verify FP8 scale exports no longer contain NaN bytes during dynamic NVFP4 operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1465)


shengliangxu (approved):
LGTM

ShawRong:
Hi @shengliangxu , thanks for the review! It looks like the required CI checks are waiting for maintainer approval. Could you please approve and run the workflows when convenient?

sychen52:
Looks like it is already patched. Only see a test case added in this PR.

ShawRong:
> Looks like it is already patched. Only see a test case added in this PR.
Looks it's covered by PR #1382. And if suitable, I will close this PR.

sychen52 (approved):
I think the unittest is still good to add.

PR #1473, opened 2026-05-12
fix(te-plugin): make _Linear arg indexing robust to TE signature changes

kevalmorabia97:
### What does this PR do?

Type of change: Bug fix

ModelOpt's `te_quantized_linear_fn` and `te_grouped_quantized_linear_fn` read `weight` / `inp` from hard-coded positions in `args`. Two TE signature changes broke this scheme:

- **TE 1.x → 2.0:** dropped the legacy `weight_fp8` slot between `weight` and `inp`. ModelOpt handled this with an `if Version("2.0") <= _TE_VERSION:` branch + a duplicate else branch.
- **TE 2.14 → 2.15:** inserted `weight_workspace` between `weight` and `inp` at the `_Linear.forward` call site ([TE 2.15 linear.py L1663](https://github.com/NVIDIA/TransformerEngine/blob/release_v2.15/transformer_engine/pytorch/module/linear.py#L1663)). Unhandled by ModelOpt — `args[idx + 1]` resolved to `None` (workspace is None outside FP8), which then crashed `TensorQuantizer.forward` on `inputs.numel()` with `AttributeError: 'NoneType' object has no attribute 'numel'`. Surfaced as a regression in Megatron-Bridge after the TE 2.15 bump alongside ModelOpt 0.44.0rc3.
- **TE 2.10:** `_GroupedLinear.forward`'s second positional slot was renamed `m_splits` → `non_tensor_args` (tuple wrapping). ModelOpt had a separate `Version("2.10")` gate for this.

Replace all three version gates with **parameter-name introspection** of the live `_Linear.forward` / `_GroupedLinear.forward` signature. The parameter names (`weight`, `inp`, `m_splits`, `non_tensor_args`) have been stable across TE 1.x, 2.x, and 2.15+; only their relative positions shift. The new code reads the live signature via `inspect.signature(...).parameters`, locates `weight`/`inp` by name, and mutates only those positions in a list copy of `args` — everything between (e.g. TE 2.15's `weight_workspace`) and after passes through verbatim. The dual-branch code in `te_quantized_linear_fn` collapses to a single path.

### Usage

No public API change. PTQ continues to work transparently across all supported TE versions:

```python
# Works on TE 1.x, 2.0-2.14, 2.15.x, and 2.16+ — no version flag needed.
mtq.quantize(model, mtq.NVFP4_DEFAULT_CFG, forward_loop)
```

### Testing

Existing TE plugin tests (`tests/gpu_megatron/torch/quantization/plugins/test_transformer_engine.py`) exercise both the `_forward` (no-grad calibration) and `_apply` (grad-enabled training) paths of `te_quantized_linear_fn` for `te.pytorch.Linear` — they would have caught the original TE 2.15 regression on a CI matrix entry pinned to TE 2.15. Verified trace correctness across:

| TE version | `_Linear.forward` signature | `_te_linear` weight→inp gap | `_GroupedLinear.forward` second slot |
|---|---|---|---|
| 1.x | `(ctx, weight, weight_fp8, inp, …)` | 1 | n/a |
| 2.0–2.14 | `(ctx, weight, inp, bias, …)` | 0 | `m_splits` |
| 2.15.x | `(ctx, weight, weight_workspace, inp, …)` | 1 | `non_tensor_args` |
| 2.16+ (main) | `(ctx, weight, inp, bias, fwd_args)` | 0 | `non_tensor_args` |

- Did you write any new necessary tests?: Existing tests sufficient 

### Additional Information

Triggered by Megatron-Bridge https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3783 failing tests after bumping ModelOpt 0.44.0rc2 → 0.44.0rc3 together with a Megatron-LM bump that pulls TE 2.15. ModelOpt rc2 had the same latent bug — it just wasn't exercised until TE 2.15 became the runtime version.

## Summary by CodeRabbit

* **Refactor**
  * Improved Transformer Engine quantization plugin robustness by using runtime parameter inspection instead of version-based branching, ensuring compatibility across TE versions without requiring manual updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1473)

### DIFF
--- a/modelopt/torch/quantization/plugins/transformer_engine.py
+++ b/modelopt/torch/quantization/plugins/transformer_engine.py
@@ -15,6 +15,7 @@
 
 """Support quantization for Transformer Engine layers."""
 
+import inspect
 import warnings
 
 import torch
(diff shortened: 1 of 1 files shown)


kevalmorabia97:
Thanks for the catch — the analysis is correct. `inspect.signature(te_linear._Linear.forward)` runs *while* `replace_function`'s context is active, so it inspects the `functools.partial` wrapper whose signature collapses to `(*args, **kwargs)`. `names.index("weight")` would raise `ValueError` on the first calibration forward.

Verified locally with a minimal repro:

```python
import inspect
from functools import partial

class _Linear:
    @staticmethod
    def forward(ctx, weight, weight_workspace, inp, bias, non_tensor_args): pass

_Linear._forward = _Linear.forward
_Linear.forward = partial(lambda *a, **kw: None, _Linear, "_forward", "self")

# Buggy: list(inspect.signature(_Linear.forward).parameters) == ['args', 'kwargs']
# Fixed: list(inspect.signature(getattr(_Linear, "_forward", _Linear.forward)).parameters)
#       == ['ctx', 'weight', 'weight_workspace', 'inp', 'bias', 'non_tensor_args']
```

Fix in eeb9b9d3 — uses `getattr(te_linear._Linear, "_forward", te_linear._Linear.forward)` (and the `_GroupedLinear` analogue): when the `forward` patch is active the cached original is at `_forward`; when `apply` is patched instead, `_forward` is absent and `forward` is itself the original, so the fallback works for both paths.

Re: the test concern — agreed that `test_quantize` would have caught this had it run with a real TE install; the most likely explanation is the CI environment doesn't have TE available, so the `pytest.importorskip("transformer_engine")` skips the whole file. A TE-pinned matrix entry is the right follow-up but is out of scope for this PR.

shengliangxu (approved):
LGTM

kinjalpatel27 (approved):
LGTM

PR #1480, opened 2026-05-13
chore: declare contents: read on code_quality workflow

arpitjain099:
`code_quality.yml` runs lint and code-quality checks. Twelve other workflows already declare permissions; this brings the remaining quality-check workflow in line.

## Summary by CodeRabbit

* **Chores**
  * Updated automation workflow permissions to explicitly grant read access to repository contents, ensuring code-quality processes can access necessary files.

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -14,6 +14,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   code-quality:
     runs-on: ubuntu-latest


.github/workflows/code_quality.yml
```diff
@@ -14,6 +14,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
+permissions:
```
kevalmorabia97, line 17: I believe this is already the default so why do we need to explicitly set it here?

arpitjain099: @kevalmorabia97 you're right that the effective scope may already be `contents: read` on this repo today if the default token scope is set to the restrictive value. The reasons to declare it explicitly anyway:

1. It documents the actual scope this workflow needs, so a reviewer does not have to check repo-level token settings to know.
2. It survives repo-default changes. If the default is ever flipped back to permissive (the legacy default which grants `contents: write` across the board), every workflow that did not explicitly cap stays permissive too.
3. It matches the in-repo convention: 12 of the 13 sibling workflows in `.github/workflows/` already declare `permissions:` explicitly (`bandit.yml`, `build-docs.yml`, `ci.yml`, etc.); this PR brings the lone remaining one in line.

Defense-in-depth precedent: the tj-actions/changed-files supply-chain attack (CVE-2025-30066, March 2025) showed how an over-scoped `GITHUB_TOKEN` exfiltrates credentials when a third-party action is compromised. Pinning per workflow caps the blast radius regardless of what the repo default is at any given moment.

Happy to close if you'd prefer to rely on the default.

arpitjain099: The effective default depends on the repo-wide Actions setting under Settings > Actions > General > Workflow permissions. If that's set to "Read and write permissions" (which is GitHub's default for new repos), then `GITHUB_TOKEN` actually defaults to write across most scopes, not read. Setting `contents: read` explicitly here pins the workflow's intent so it stays correct even if the repo-wide default changes, and matches what twelve other workflows in this repo already do (`gpu-tests.yml`, `unit.yml`, etc.).

PR #1483, opened 2026-05-13
[OMNIML-4687] training_support

ChenhanYu:
Draft PR opened by **pensieve-intern** for [OMNIML-4687](https://jirasw.nvidia.com/browse/OMNIML-4687).

Stage `training_support` of Epic `OMNIML-4681`. The agent ran from the SPEC on the ticket description; review every change before marking ready.

_Always-draft is enforced — the bot never auto-merges._


ChenhanYu:
Closing — Qwen3-8B was a demo Epic for v0.33.5 dispatch validation; that demo is now wrapped up. The agent diff here also drifted onto unrelated files (`modelopt/torch/quantization/algorithms.py`, `uv.lock`) which would need scrubbing before this could merge. Future training_support runs will fire against real release Epics on v0.33.38.

PR #1495, opened 2026-05-14
[OMNIML-4730] Support quantized nn.Embedding

ajrasane:
### What does this PR do?

Type of change: new feature

Register `nn.Embedding` in `QuantModuleRegistry` so the embedding table and lookup activations participate in quantization end-to-end:

- New `modelopt/torch/quantization/nn/modules/quant_embedding.py` exposes `weight_quantizer` (embedding table), `output_quantizer` (lookup activations, off by default), and an `input_quantizer` placeholder. Embedding inputs are integer indices that cannot be fake-quantized, so direct `enable()` / `enable_quant()` / `enable_calib()` calls on `input_quantizer` raise; `forward()` never invokes `input_quantizer` at all, so a back-door flip of `_disabled` is a no-op rather than a crash. Wildcard configs (`*input_quantizer`) are accepted silently so the stock deny-all → enable-wildcards → opt-out pattern in `NVFP4_DEFAULT_CFG` and friends still works.
- `default_disabled_quantizers.yaml` installs `parent_class: nn.Embedding, enable: false` so embedding quantization is opt-in and existing model behavior is unchanged.
- `is_quantized_linear` in `core_utils.py` early-returns `False` for `nn.Embedding` so AWQ / SmoothQuant / SVDQuant don't treat it as a GEMM op.
- `_process_quantized_modules` in `unified_export_hf.py` routes quantized `nn.Embedding` modules through `_export_quantized_weight`, so the exported checkpoint contains the packed NVFP4 / FP8 / INT bytes plus `weight_scale*` buffers, exactly like Linear layers.

### Usage

```python
import copy
import torch.nn as nn

# Opt embeddings into the stock NVFP4 config — the YAML default is opt-out.
cfg = copy.deepcopy(mtq.NVFP4_DEFAULT_CFG)
cfg["quant_cfg"].append(
    {
        "parent_class": "nn.Embedding",
        "quantizer_name": "*weight_quantizer",
        "cfg": {"num_bits": (2, 1), "block_sizes": {-1: 16, "type": "dynamic", "scale_bits": (4, 3)}},
    }
)

model = mtq.quantize(model, cfg, forward_loop)
export_hf_checkpoint(model, export_dir="./out")
# out/model.safetensors contains: embedding.weight (uint8, NVFP4-packed),
# embedding.weight_scale (FP8 E4M3 per-block), embedding.weight_scale_2 (FP32).
```

### Testing

- New unit tests `tests/unit/torch/quantization/test_quant_embedding.py` cover: default quantizer state, no-quant identity, per-tensor and per-row weight fake quant against the manual `tensor_quant.fake_tensor_quant` reference, output quantizer activation, locked-mutator raises (parametrized over `enable` / `enable_quant` / `enable_calib`), and the wildcard-then-opt-out pattern. All 8 cases pass under `CUDA_VISIBLE_DEVICES=""`. Export-path coverage (weight packing + tied-weight guard) lives in `tests/gpu/torch/export/test_export_embedding.py` because the export path hits `torch.cuda.empty_cache()`.
- Verified end-to-end on an embedding-only model: `mtq.quantize` with `NVFP4_DEFAULT_CFG` + the embedding opt-in produces `embedding.weight (uint8)`, `embedding.weight_scale (float8_e4m3fn)`, `embedding.weight_scale_2 (float32)` in the exported safetensors, with `"quant_algo": "NVFP4"` in `hf_quant_config.json`.
- End-to-end smoke test on TinyLlama-1.1B-Chat-v1.0 via `hf_ptq.py` (32-sample calibration on the default cnn_dailymail + nemotron-post-training datasets), sweeping the embedding's `weight_quantizer` across FP8 (per-tensor E4M3) and NVFP4 (E2M1, block_size=16, FP8 scales) while varying the Linear-layer format. KV cache is FP8 in all runs. Greedy generation (`max_new_tokens=100`) from the standard CNN/DailyMail preview prompt:

  > `<s>` LONDON, England (Reuters) -- Harry Potter star Daniel Radcliffe gains access to a reported £20 million ($41.1 million) fortune as he turns 18 on Monday, but he insists the money won't cast a spell on him. […] The Londoner has filmed a TV movie called "My Boy Jack," about author Rudyard Kipling and

  - **Baseline FP16 (no quantization):** *"his son, who died at the age of 18. He also has a role in the upcoming "The Prestige" and is set to star in the remake of "The League of Extraordinary Gentlemen." "I'm not going to be a star for the rest of my life," he said. "I'm going to be a star for a while, but I'm not going to be a star for the rest of my life."*
  - **FP8 Linear + FP8 embedding** (`hf_quant_config.json`: `"quant_algo": "FP8"`, `exclude_modules: ["lm_head"]`, so `model.embed_tokens` is FP8): *"his son, who died at the age of 18. He also has a role in the upcoming "The Prestige" and is set to star in the upcoming "The Golden Compass." "I'm not going to be a star forever," he said. "I'm going to be a star for a while, but I'm not going to be a star forever."*  (matches FP16 token-for-token until a single lexical drift)
  - **FP8 Linear + NVFP4 embedding** (`model.embed_tokens: W4A16_NVFP4, group_size=16`; Linear: `FP8`): *"his son, who died in the First World War. He also has a role in the upcoming "The Golden Compass" and is set to star in the BBC's "The Bill." "I'm not going to be a star for the rest of my life," he said. "I'm going to be a star for a while, but I'm not going to be a star for the rest of my life."*  (embedding format swap is visible — first phrase diverges from baseline; rest of the sentence still tracks)
  - **NVFP4 Linear + FP8 embedding** (`model.embed_tokens: FP8`; Linear: `NVFP4, group_size=16`): *"his son, and is also in talks to play the title role in a film version of the stage play "The Importance of Being Earnest." "I've been offered a lot of things," he said. "I've been offered a lot of things, but I've also been offered a lot of things that I've said no to." Radcliffe's first movie, "The Tailor of Gloucester," was released in the UK in"*  (expected drift at W4A4)
  - **NVFP4 Linear + NVFP4 embedding** (`model.embed_tokens: W4A16_NVFP4, group_size=16`; Linear: `NVFP4, group_size=16`): *"his son, who died in a car accident. Radcliffe will also star in the film, which is being shot in London. "It's a very different kind of film," he said. "It's a very personal story, and I've been very lucky to be able to do it." Radcliffe's next film, "The Girl with the Dragon Tattoo," is due to be released in the U.S. In December. He will also"*  (full-NVFP4 path; quantized embedding row is `W4A16_NVFP4` because the embedding's input quantizer is permanently disabled by design)

- Is this change backward compatible?: ✅ — embedding quantizers are opt-in via `parent_class: nn.Embedding, enable: false` in `default_disabled_quantizers.yaml`, so existing model behavior is unchanged.
- Did you get Claude approval on this PR?: ❌ — will run `/claude review` after the PR is up.

## Summary by CodeRabbit

* **New Features**
  * Opt-in quantization for embedding layers: weight quantization enabled by default, optional output quantization; input quantization remains permanently disabled.

* **Bug Fixes**
  * Preserve tied embedding weights during export by skipping packing (emits warning) to avoid breaking weight sharing.

* **Tests**
  * Added unit tests covering embedding quantization behavior, export packing, calibration, and tied-weight scenarios.

* **Documentation**
  * Changelog updated with embedding quantization notes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1495?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cjluo-nv (commented):

Quantized `nn.Embedding` support cleanly mirrors the `_QuantLinear`/`QuantLinearConvBase` pattern (dynamic-attribute weight, `quantize_weight` context, `_register_temp_attribute`), so design-wise it slots in well — no second composition system. The wildcard-tolerance via `_UnsettableInputQuantizer` is unusual but justified: stock recipes apply `*input_quantizer` enables, and the YAML `parent_class: nn.Embedding, *, enable: false` rule is appended last in every preset that uses `_default_disabled_quantizer_cfg`, so the disabled state is restored before forward. Unit tests cover the lock semantics and weight quant against `fake_tensor_quant` reference.

Three things worth a maintainer look before approving:

1. **`output_quantizer` is silently bypassed under `torch.export`.** `_QuantEmbedding.forward` does `if is_torch_export_mode(): return super().forward(...)` — that path never calls `self.output_quantizer(output)`. `QuantLinearConvBase`/`QuantInputBase` both keep the output_quantizer in the export path. If a user opts into `output_quantizer` and then `torch.export`s, they'll lose it without warning. Probably harmless today (output_quantizer is off by default) but it's an inconsistency.

2. **Tied embeddings (`tied_word_embeddings=True`) likely break on export.** `_export_quantized_weight` does `setattr(sub_module, weight_name, nn.Parameter(quantized_weight, ...))`, replacing `embedding.weight` with a new Parameter holding packed uint8 bytes. If `lm_head.weight` was tied to the same Parameter, the tie is severed and `lm_head` keeps a stale float weight; `postprocess_state_dict`'s tied-weight dedup will then drop one of the keys from the safetensors output. The PR description's example uses an embedding-only model, which sidesteps this — but in real LLMs (Llama/Qwen with tied embeddings) this needs at least a guard or explicit warning.

3. **No export-path test.** All new tests are pure forward tests; the new `_process_quantized_modules` branch routing `nn.Embedding` through `_export_quantized_weight` has no coverage. Given (2), an export round-trip test on a tiny tied-embedding model would catch the issue. The PR description says it was verified manually on an embedding-only model — that's exactly the case that doesn't exercise the tying path.

Smaller/optional: the `_UnsettableInputQuantizer.enable*` overrides catch user-facing direct calls, but `set_from_attribute_config({"enable": True})` writes `_disabled` directly via `setattr`, so the only real defense is the runtime check in `forward`. The current docstring already explains this; just confirm the runtime guard is the load-bearing one and the method overrides are belt-and-suspenders.

cjluo-nv (commented):

Re-review: all three critical comments from the previous round are addressed.

1. **output_quantizer under torch.export**: `forward()` now restructured so `return self.output_quantizer(output)` lives outside the if/else — applied in both the export and non-export paths, matching `QuantInputBase.forward`. ✅
2. **Tied-embedding export safety**: `_process_quantized_modules` now scans `model.named_modules()` for any other module whose `weight` is the same Parameter object as the embedding's, and skips packing with a `UserWarning` when a tie is found, preserving `tied_word_embeddings` semantics. ✅
3. **Export-path tests**: New `TestQuantEmbeddingExport` covers both the success path (uint8-packed weight + `weight_scale`/`weight_scale_2` registered, no `input_scale`) and the tied-weight guard (warning + tie preserved). ✅

Design-wise this isn't a new abstraction — `quant_embedding.py` slots in next to `quant_linear.py`/`quant_conv.py`/etc., reusing `QuantModule`, `QuantModuleRegistry`, `_register_temp_attribute`, `_register_dynamic_attribute`, and the `quantize_weight` context manager pattern. Previous reviewer explicitly endorsed the design.

Minor docs drift: PR body still says "`forward()` raises if `_disabled` is flipped via any back door", but the current forward has no such runtime check. That's fine in practice — `nn.Embedding.forward` never invokes `input_quantizer`, so a back-door enable can't cause an integer-index tensor to hit the fake-quant path. Worth tightening the PR description but not a blocker.

modelopt/torch/quantization/nn/modules/quant_embedding.py
```diff
@@ -0,0 +1,157 @@
...
+    def _setup(self):
+        """Register weight, (locked) input, and output quantizers."""
+        self._register_temp_attribute(
+            "weight_quantizer", TensorQuantizer(self.default_quant_desc_weight)
+        )
+        # Build the input quantizer disabled. _UnsettableInputQuantizer's mutators raise,
+        # so we disable it once at construction via direct attribute assignment.
+        input_quantizer = _UnsettableInputQuantizer(self.default_quant_desc_input)
+        input_quantizer._disabled = True
+        self._register_temp_attribute("input_quantizer", input_quantizer)
```
cjluo-nv, line 83: could you help me understand:

why we have an input_quantizer here? Isn't this a weight quantizer only?

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -297,6 +297,11 @@ def is_quantized_linear(module):
     """Check if a module is a quantized linear module."""
     from ..nn import QuantModule, TensorQuantizer
 
+    # Embedding has a 2D weight but is not a GEMM op, so calibration passes that operate
+    # on linear activations (AWQ, SmoothQuant, SVDQuant) must skip it.
+    if isinstance(module, nn.Embedding):
```
cjluo-nv, line 302: why do we need this?

cjluo-nv (commented):
do we plan to run some fake quant evals?

modelopt/torch/quantization/nn/modules/quant_embedding.py
ajrasane: You're right that only the weight is *actually* quantized. The slot exists so the wildcard-recipe pattern keeps working without special-casing `nn.Embedding`:

1. **Stock recipes glob `*input_quantizer`.** `NVFP4_DEFAULT_CFG` / `FP8_DEFAULT_CFG` install a `*input_quantizer` entry. With the slot present, `_UnsettableInputQuantizer.set_from_attribute_config` absorbs the cfg and force-disables it — without it, the wildcard would either raise on `nn.Embedding` or silently miss.
2. **Symmetry with other QuantModules.** Export/calibration code uses `getattr(sub_module, quantizer_attrs.input_quantizer, None)`; keeping the attribute means no `if isinstance(..., nn.Embedding): ...` branches scattered around.
3. **Loud failure on accidental enable.** Direct `enable()` / `enable_quant()` / `enable_calib()` raise instead of silently no-opping a misconfigured recipe.

modelopt/torch/quantization/utils/core_utils.py
ajrasane: Without it, `is_quantized_linear` returns `True` for `QuantEmbedding` — it's a `QuantModule` with a `weight_quantizer`, an `input_quantizer` (the `_UnsettableInputQuantizer`), and a 2-D weight, so it passes every check below.

That would route the embedding into the linear-only calibration passes that consume `is_quantized_linear`: AWQ-Lite / SmoothQuant scale insertion (`model_calib.py:1275`, `1605`), SVDQuant convert (`conversion.py:154`), real-quant compression (`compress.py:73`), etc. Those all assume the module is a GEMM with a quantized activation input — for `nn.Embedding`, the "input" is integer indices, so they'd either crash on the missing input quantizer state or silently insert a meaningless smooth scale.

The early-return keeps `is_quantized_linear` semantically "is this a quantizable GEMM" rather than "does it have linear-shaped quantizer attributes".

cjluo-nv (commented):

Re-review of PR #1495 (quantized `nn.Embedding`). Design is settled — `quant_embedding.py` slots in next to `quant_linear.py`/`quant_conv.py`, reusing existing `QuantModule`/`QuantModuleRegistry`/`_register_temp_attribute`/`_register_dynamic_attribute` patterns; no second composition system introduced. The three critical issues from the prior round are all resolved in the current diff:

- **output_quantizer under torch.export** ✅ — `forward()` now applies `self.output_quantizer(output)` in both branches (lines 145–157 of `quant_embedding.py`).
- **Tied-embedding export safety** ✅ — `_process_quantized_modules` (`unified_export_hf.py` 682–708) scans for any other module whose `weight is sub_module.weight` and skips packing with a `UserWarning` to preserve `tied_word_embeddings`.
- **Export-path tests** ✅ — `TestQuantEmbeddingExport` covers both the success path (uint8-packed weight + `weight_scale*` registered) and the tied-weight guard (warning + tie preserved).

Flagging for human approval rather than auto-approving because:
- :speech_balloon: Maintainer @cjluo-nv asked "do we plan to run some fake quant evals?" — this looks unanswered in the thread. The PR body has a TinyLlama-1.1B greedy-generation smoke test across FP8/NVFP4 combos, but no perplexity/MMLU-style numbers. A maintainer should confirm whether the smoke test is sufficient or whether a quantitative eval is expected before merging.
- This is a public-API behavior change: `nn.Embedding` is newly registered in `QuantModuleRegistry`, `is_quantized_linear` now early-returns `False` for embeddings (affects AWQ/SmoothQuant/SVDQuant routing), and a new `parent_class: nn.Embedding` rule lands in `default_disabled_quantizers.yaml`. Worth a human eye on downstream callers.
- Minor docs drift the bot already noted: PR body still claims "`forward()` raises if `_disabled` is flipped via any back door", but the runtime check was removed; in practice this is fine because `nn.Embedding.forward` never invokes `input_quantizer`, but the description should be tightened.
- CodeRabbit's `warn_rank_0` nit on the tied-embedding warning is reasonable per the project's rank-0-logging convention but not blocking.

modelopt/torch/quantization/nn/modules/quant_embedding.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+_INPUT_QUANTIZER_ERR = (
+    "Cannot configure input_quantizer on a quantized nn.Embedding: the input is integer "
+    "indices and cannot be fake-quantized. Configure weight_quantizer (and optionally "
+    "output_quantizer) instead."
+)
+
+
+class _UnsettableInputQuantizer(TensorQuantizer):
```
realAsma: Should we make this a generic `HardDisabledTensorQuantizer` and move it to `..nn/modules/tensor_quantizer.py` ?

modelopt/torch/quantization/nn/modules/quant_embedding.py
realAsma, line 1: The hard disabled TensorQuantizer might not be preserved during restore;

Could you add a unittest which do quantize -> save -> restore?
See: 
https://github.com/NVIDIA/Model-Optimizer/blob/9ced018ec420da1ee84b8260e35c27ddf05571ee/tests/unit/torch/quantization/test_quantize_cpu.py#L143


you would need to add the restore support here - https://github.com/NVIDIA/Model-Optimizer/blob/9ced018ec420da1ee84b8260e35c27ddf05571ee/modelopt/torch/quantization/conversion.py#L134

realAsma: if you have bandwidth would be great if you could add a registry for TensorQuantizers and avoid manual conditional branching in `restore` method

ajrasane: Added `TestQuantEmbeddingSaveRestore` covering the full `quantize → modelopt_state → restore → load_state_dict` round-trip — asserts the restored `input_quantizer` is still a `HardDisabledTensorQuantizer`, `enable*()` still raises, and forward output matches.

No `conversion.py:134`-style branch was needed: `_QuantEmbedding._setup()` always creates the `HardDisabledTensorQuantizer`, and `replace_quant_module` runs before `restore_quantizer_state`, so the right subclass is already in place by the time properties are restored.

ajrasane: > if you have bandwidth would be great if you could add a registry for TensorQuantizers and avoid manual conditional branching in restore method

Ok, I will do this as a part of a separate PR.

cjluo-nv (commented):

Re-review of PR #1495 (quantized `nn.Embedding`).

**Design:** Settled. `quant_embedding.py` slots in next to `quant_linear.py`/`quant_conv.py`, reusing `QuantModule` / `QuantModuleRegistry` / `_register_temp_attribute` / `_register_dynamic_attribute` / `quantize_weight` context patterns — no second composition system. The new `HardDisabledTensorQuantizer` is a small subclass in the existing `tensor_quantizer.py`, not a new abstraction.

**Previously raised critical issues — all resolved in current diff:**
- :speech_balloon: output_quantizer under torch.export — `forward()` now applies `self.output_quantizer(output)` outside the export/non-export branch, matching `QuantInputBase.forward`. ✅
- :speech_balloon: Tied-embedding export safety — `_process_quantized_modules` (`unified_export_hf.py:679–708`) scans `model.named_modules()` for any other module whose `weight is sub_module.weight` and skips packing with a `UserWarning`, preserving `tied_word_embeddings` semantics. ✅
- :speech_balloon: Export-path tests — `tests/gpu/torch/export/test_export_embedding.py::TestQuantEmbeddingExport` covers both the success path (uint8-packed weight + `weight_scale`/`weight_scale_2` registered, no `input_scale`) and the tied-weight guard (warning + tie preserved). ✅
- :speech_balloon: realAsma's save/restore concern — `TestQuantEmbeddingSaveRestore` added; verifies the restored `input_quantizer` is still a `HardDisabledTensorQuantizer`, that `enable*()` still raises, and that forward output round-trips. The flow is correct because `replace_quant_module` runs before `restore_quantizer_state`, so `_QuantEmbedding._setup()` has already installed the right subclass before saved state is replayed via `set_from_modelopt_state`. ✅

**Remaining items for human sign-off:**
- :speech_balloon: Maintainer asked "do we plan to run some fake quant evals?" — appears unanswered. PR body has a TinyLlama-1.1B greedy-generation smoke test across FP8/NVFP4 combos but no perplexity/MMLU numbers. A maintainer should confirm whether the smoke test is sufficient before merging.
- This is a public-API behavior change worth a human eye on downstream consumers: `nn.Embedding` is newly registered in `QuantModuleRegistry`, `is_quantized_linear` now early-returns `False` for embeddings (changes routing in `model_calib.py`/`compress.py`/SVDQuant `conversion.py`), and a new `parent_class: nn.Embedding` rule lands in `default_disabled_quantizers.yaml`.
- Minor docs drift (non-blocking): PR body still says "`forward()` raises if `_disabled` is flipped via any back door", but the runtime check was removed. Harmless in practice (`nn.Embedding.forward` never invokes `input_quantizer`), but the description should be tightened.
- CodeRabbit's `warn_rank_0` nit on the tied-embedding warning is reasonable per the project's rank-0-logging convention but not blocking.
- realAsma's "TensorQuantizer registry to remove manual conditional branching in restore" was explicitly deferred to a follow-up PR by the author — fine.

realAsma (approved):

cjluo-nv (commented):

All three previously-raised critical issues are fully resolved in the current diff:

- :speech_balloon: **output_quantizer under torch.export** — `forward()` now applies `self.output_quantizer(output)` outside the export/non-export branch, matching `QuantInputBase.forward`. ✅
- :speech_balloon: **Tied-embedding export safety** — `_process_quantized_modules` (`unified_export_hf.py:679–708`) scans `model.named_modules()` for any other module whose `weight is sub_module.weight` and skips packing with a `UserWarning`, preserving `tied_word_embeddings`. ✅
- :speech_balloon: **Export-path tests** — `tests/gpu/torch/export/test_export_embedding.py::TestQuantEmbeddingExport` covers both the success path (uint8-packed weight + `weight_scale`/`weight_scale_2`, no `input_scale`) and the tied-weight guard. ✅
- :speech_balloon: **Save/restore (realAsma)** — `TestQuantEmbeddingSaveRestore` added; verifies `HardDisabledTensorQuantizer` survives `modelopt_state` round-trip and mutators still raise. ✅

Design is settled — `quant_embedding.py` slots in next to `quant_linear.py`/`quant_conv.py`, reusing `QuantModule` / `QuantModuleRegistry` / `_register_temp_attribute` / `_register_dynamic_attribute` / `quantize_weight` patterns; `HardDisabledTensorQuantizer` is a small subclass in the existing `tensor_quantizer.py`, not a new abstraction.

**Why I'm flagging for human approval rather than auto-approving:**

- :speech_balloon: Maintainer @cjluo-nv asked *"do we plan to run some fake quant evals?"* — this thread looks unanswered. The PR body has a TinyLlama-1.1B greedy-generation smoke test across FP8/NVFP4 combos, but no perplexity/MMLU numbers. A maintainer should confirm whether the smoke test is sufficient or whether a quantitative eval is expected before merging.
- This is a **public-API behavior change** worth a human eye on downstream consumers: `nn.Embedding` is newly registered in `QuantModuleRegistry`, `is_quantized_linear` now early-returns `False` for embeddings (changes routing in `model_calib.py` AWQ/SmoothQuant, `compress.py`, SVDQuant `conversion.py`), and a new `parent_class: nn.Embedding` rule lands in `default_disabled_quantizers.yaml`.
- Minor docs drift (non-blocking): PR body still says *"`forward()` raises if `_disabled` is flipped via any back door"*, but that runtime check was removed. Harmless in practice (`nn.Embedding.forward` never invokes `input_quantizer`), but the description should be tightened.
- CodeRabbit's `warn_rank_0` nit on the tied-embedding warning is reasonable per the project's rank-0-logging convention but not blocking.
- realAsma's "TensorQuantizer registry to remove manual conditional branching in restore" was explicitly deferred to a follow-up PR by the author — fine.

Recommendation: get an answer on the eval question and a quick sanity check on the `is_quantized_linear` API change from someone with calibration-pass context, then approve.

cjluo-nv (approved):

PR #1497, opened 2026-05-14
Add active-MoE AutoQuant cost accounting

meenchen:
### What does this PR do?

• Type of change: new feature

  Adds an active_moe cost model for auto_quantize effective-bits search. This lets AutoQuant account for routed MoE expert weights by active decode weight traffic instead of total checkpoint weight
  size, using active_moe_expert_ratio = num_experts_per_tok / num_experts.

  The default behavior is unchanged: cost_model="weight" still counts all quantizable weights equally.

  ### Usage

  model, search_state = mtq.auto_quantize(
      model,
      constraints={"effective_bits": 5.0},
      quantization_formats=[
          mtq.NVFP4_DEFAULT_CFG,
          mtq.FP8_DEFAULT_CFG,
      ],
      data_loader=calib_dataloader,
      forward_step=forward_step,
      loss_func=loss_func,
      cost_model="active_moe",
      # Optional. If omitted, ModelOpt tries to infer this from model.config.
      active_moe_expert_ratio=2 / 64,
  )

  The HF PTQ example also exposes:

  --auto_quantize_cost_model active_moe \
  --auto_quantize_active_moe_expert_ratio 0.03125

  ### Testing

  python -m pytest tests/unit/torch/quantization/test_autoquant.py -q -k 'active_moe or quant_recipe_hparam_cost_weight'
  python -m pytest tests/unit/torch/quantization/test_autoquant.py -q -k 'not data_parallel_auto_quantize'

  Results:

  - 4 passed
  - 58 passed, 1 deselected

## Summary by CodeRabbit

* **New Features**
  * Added active-MoE cost model option for auto-quantization with configurable expert ratio; API and CLI accept cost_model and active_moe_expert_ratio
  * Unified auto-quantize supports new quant format w4a16_nvfp4

* **Bug Fixes**
  * Ensure labels are moved to the logits device for base models without an lm_head
  * CLI enforces valid expert-ratio range and requires active-MoE mode when a ratio is provided

* **Tests**
  * Added unit tests for active-MoE behavior, cost-weighting, ratio handling, and search budget selection

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1497?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


modelopt/torch/quantization/model_quant.py
```diff
@@ -283,6 +331,8 @@ def auto_quantize(
     verbose: bool = False,
     method: str = "gradient",
     checkpoint: str | None = None,
+    cost_model: str = "weight",
+    active_moe_expert_ratio: float | None = None,
```
realAsma: Can we move these arguments to the search constraints?
This should be part of constraints. We will be adding support for more types of constraints in future (Such as per-layer measure latency.

```
constraints: dict[str, float | str] = {"cost_model": weight, "cost": {"active_moe_expert_ratio": 0.5}},
```

modelopt/torch/quantization/model_quant.py
```diff
@@ -311,6 +419,13 @@ def auto_quantize(
                 # For an effective quantization bits of 4.8
                 constraints = {"effective_bits": 4.8}
 
+                # For active-MoE accounting where 2 of 8 routed experts are active per token
+                constraints = {
+                    "effective_bits": 4.8,
+                    "cost_model": "active_moe",
+                    "cost": {"active_moe_expert_ratio": 0.25},
+                }
+
```
juhi10071998, line 323: hi @meenchen , should we group the `cost_model` and `cost` as part of the **constraints** or would it be better if we add a new arg (something like _cost_info_) and add these there?

I am working on auto_quantize YAML and this will help in getting a good YAML structure.

cc- @shengliangxu?

meenchen: I moved this under `constraints`:

```python
constraints = {
    "effective_bits": 4.8,
    "cost_model": "active_moe",
    "cost": {"active_moe_expert_ratio": 0.25},
}
```

I think this should stay in `constraints` rather than a separate `cost_info`, because the cost model defines how the effective-bits constraint is measured. Keeping the ratio in nested `constraints["cost"]` also leaves room for future cost-model-specific knobs, such as latency-related parameters, without adding another top-level API surface.

modelopt/torch/quantization/algorithms.py
```diff
@@ -407,6 +436,53 @@ def sanitize_search_config(self, config: SearchConfig | None) -> SearchConfig:
         )
         return config
 
+    def _get_cost_constraints(self) -> tuple[str, float | None]:
```
juhi10071998: hi @meenchen - is it possible to unify (adding a hook or something) the check in the model_quant.py `_normalize_auto_quantize_constraints` and this function? it seems like there is a duplication of checking the constraints and it might be prone to missing as we scale and add more cost models

examples/llm_ptq/hf_ptq.py
```diff
@@ -1401,6 +1410,27 @@ def parse_args() -> argparse.Namespace:
             "(sensitivity scores, costs, etc.). Only used when auto_quantize_bits is specified."
         ),
     )
+    parser.add_argument(
+        "--auto_quantize_cost_model",
+        type=str,
+        default="weight",
+        choices=["weight", "active_moe"],
```
juhi10071998, line 1407: @meenchen - let's review this again after our today's discussion. If we want to have multiple cost models stacked together, should we take cost_model as a list instead? this way we don't have to add a new combinatorial empty if we were to stack ever

We could make cost_model a list of strategies that compose:

```
  constraints = {
      "effective_bits": 5.0,
      "cost_model": ["weight_plus_activation", "active_moe"],   # both apply
      "cost": {"active_moe_expert_ratio": 0.125, "target_seq_length": 2048},
  }
```

  Each strategy contributes its own multiplier or additive term to cost(module, recipe). They stack without explosion of named variants.

modelopt/torch/quantization/model_quant.py
```diff
@@ -265,10 +265,115 @@ def forward_loop(model) -> None:
...
+        if ratio <= 0.0:
+            continue
+        return min(ratio, 1.0)
+    return None
+
+
+def _normalize_auto_quantize_constraints(
+    model: nn.Module, constraints: dict[str, Any] | None
+) -> dict[str, Any]:
+    constraints = {"effective_bits": 4.8} if constraints is None else dict(constraints)
```
shengliangxu: can we move this 4.8 magic number to some constant and add comments explain the meaning? it's scattered around in multiple places

modelopt/torch/quantization/model_quant.py
```diff
@@ -265,10 +265,115 @@ def forward_loop(model) -> None:
...
+    cost_model = constraints.get("cost_model", "weight")
+    if cost_model not in ("weight", "active_moe"):
```
shengliangxu: We should also move these constants somewhere centralized and add comments explain what are they.

shengliangxu: better define 2 classes and define their behaviour/properties using the class.

examples/llm_ptq/hf_ptq.py
meenchen: Discussed offline. For this PR, let us keep `cost_model` as a single selected strategy rather than making the API compose/stack multiple cost models.

Stacking cost models can be ambiguous because different models may not combine uniformly as simple multipliers or additive terms, so I would avoid encoding that in the current YAML/CLI surface. In the future, we can provide an extension API/hook for users to define their own cost model, and users who need compound behavior can implement it there without us adding combinatorial built-in variants.

modelopt/torch/quantization/model_quant.py
```diff
@@ -265,6 +265,54 @@ def forward_loop(model) -> None:
...
+    return None
+
+
+def _infer_active_moe_expert_ratio(model: nn.Module) -> float | None:
+    num_active_experts = _get_first_numeric_config_attr(model, _ACTIVE_MOE_TOP_K_ATTRS)
+    num_experts = _get_first_numeric_config_attr(model, _ACTIVE_MOE_NUM_EXPERTS_ATTRS)
+    if num_active_experts is None or num_experts is None or num_experts <= 0:
+        return None
+    return min(num_active_experts / num_experts, 1.0)
+
```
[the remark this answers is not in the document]
meenchen: Updated in fac22b7e. The ratio inference now lives in `_auto_quantize_cost.py` and checks top-k and num-experts on the same config object before returning a ratio; the unit test covers the wrapper/text_config mismatch case.

modelopt/torch/quantization/algorithms.py
meenchen: Updated in fac22b7e. I moved shared validation/normalization into `_auto_quantize_cost.normalize_auto_quantize_constraints()` and both the public API and the searcher call it, so adding a cost model has one validation path.

modelopt/torch/quantization/model_quant.py
meenchen: Done in fac22b7e. `DEFAULT_AUTO_QUANTIZE_EFFECTIVE_BITS` is centralized in `_auto_quantize_cost.py` with a compatibility comment; `auto_quantize()` gets the default through the shared normalizer.

modelopt/torch/quantization/model_quant.py
meenchen: Done in fac22b7e. Added `AutoQuantizeCostModel`, `WeightCostModel`, and `ActiveMoECostModel`; the searcher now asks the selected class for module cost weights and total cost denominator.

meenchen: Done in fac22b7e. Constants and cost-model-specific keys moved into `_auto_quantize_cost.py`, with comments/docstrings for the default target and active-MoE behavior.

modelopt/torch/quantization/algorithms.py
```diff
@@ -586,23 +608,27 @@ def initialize_candidate_stats(self):
...
 
-            formats, scores, costs = [], [], []
+            formats, scores, costs, active_costs = [], [], [], []
             prev_score = float("inf")
             for recipe in hparam.choices:
                 formats.append(recipe)
 
                 score = hparam.get_score(recipe)  # type: ignore [arg-type]
                 cost = hparam.get_cost(recipe)  # type: ignore [arg-type]
+                active_cost = hparam.get_cost(recipe, cost_weight=hparam.cost_weight)  # type: ignore [arg-type]
```
cjluo-nv, line 622: > _Bot comment._

`active_cost = hparam.get_cost(recipe, cost_weight=hparam.cost_weight)` is identical to `hparam.get_cost(recipe)` because `get_cost` defaults `cost_weight` to `self.cost_weight` (which is `hparam.cost_weight`). So `active_costs` ends up equal to `costs` for every entry, and the test only checks key presence. Either drop the field, or compute something actually distinct — e.g. the unweighted physical cost via `cost_weight=1.0` for reporting routed-MoE savings vs. total weight size.

modelopt/torch/quantization/algorithms.py
```diff
@@ -1393,6 +1465,8 @@ def _resolve_best_recipe(search_state, constraints, verbose=False):
         )
 
     searcher.candidate_stats = candidate_stats
+    searcher.cost_model = search_state.get("cost_model", COST_MODEL_WEIGHT)
```
cjluo-nv, line 1470: > _Bot comment._

This builds a stub `searcher.config = {"cost_model": ...}` with no `"cost"` key and none of the other default search-config entries. It works today only because `_get_search_lower_bounds` reads just `cost_model`. If anything in `run_search_with_stats` (or anything it calls) starts touching `self.config["cost"]` or other keys, the `get_auto_quantize_config(..., constraints=...)` re-solve path will silently break. Consider initializing through `default_search_config` or persisting/restoring the full config from `search_state`.

modelopt/torch/quantization/model_quant.py
```diff
@@ -514,6 +524,8 @@ def forward_backward_step(model, batch) -> None:
     else:
         raise ValueError(f"Invalid method: {method}. Valid options are 'gradient' or 'kl_div'.")
 
+    constraints = normalize_auto_quantize_constraints(model, constraints)
```

Constraints are now normalized here and again in `_AutoQuantizeBaseSearcher.before_search`. The function is idempotent today but it's an easy place for the two sites to drift. Pick one (probably `before_search`, so external callers using the searcher directly also get normalized constraints) and drop the other.

cjluo-nv (commented):

Adds an `active_moe` cost model for AutoQuantize effective-bits accounting. Overall structure is reasonable: separate `_auto_quantize_cost.py` with a small registry, propagation of `cost_weight` through `QuantRecipeHparam`, checkpoint compat (falls back to `sum(costs[-1])` when `cost_denominator` is missing), and unit tests covering the routed-vs-shared expert weighting, config inference, and search-budget lower-bound ordering.

A few concrete things to look at before merging:

1. **`active_costs` looks like dead code.** In `initialize_candidate_stats`, `hparam.get_cost(recipe)` and `hparam.get_cost(recipe, cost_weight=hparam.cost_weight)` are equivalent because `get_cost`'s default already substitutes `self.cost_weight` when `cost_weight is None`. So `active_costs == costs` for every recipe, and the only test assertion on it is `"active_costs" in stats`. Either compute something genuinely different (e.g. the unweighted physical cost with `cost_weight=1.0`, which would actually be useful for reporting) or drop the field and the corresponding `cost_weight`/state-dict plumbing.

2. **Constraints are normalized twice.** `model_quant.auto_quantize` calls `normalize_auto_quantize_constraints` before `apply_mode`, and `_AutoQuantizeBaseSearcher.before_search` calls it again on `self.constraints`. Idempotent today, but it's an easy footgun if either side changes shape — pick one site.

3. **`_resolve_best_recipe` reconstructs a partial searcher config:** `searcher.config = {"cost_model": searcher.cost_model}` (no `"cost"` key, no other defaults). Today `_get_search_lower_bounds` only reads `cost_model` so this works, but it's fragile — any future code in `run_search_with_stats` that touches `self.config[...]` will break the resolve path. Consider populating from `search_state` more completely, or routing through `default_search_config`.

4. **Minor:** in `modelopt/torch/quantization/model_quant.py` the new `from ._auto_quantize_cost import normalize_auto_quantize_constraints` is placed after `from .algorithms import ...`; isort would put the underscore-prefixed sibling first. Also the deleted blank line above `def auto_quantize(...)` makes the module-level constant `_AUTO_QUANTIZE_SUPPORTED_ALGORITHMS` visually run into the function definition.

modelopt/torch/quantization/model_quant.py
meenchen: Fixed in 34a6baf7. Normalization now happens in `_AutoQuantizeBaseSearcher.before_search()` only, so direct searcher callers and the public API share the same normalization path.

modelopt/torch/quantization/algorithms.py
meenchen: Fixed in 34a6baf7. I dropped `active_costs` from newly generated candidate stats instead of keeping a duplicate of `costs`; the active-MoE test now asserts the duplicate field is absent while preserving `cost_weight` for the actual cost accounting.

modelopt/torch/quantization/algorithms.py
meenchen: Fixed in 34a6baf7. `_resolve_best_recipe()` now starts from `default_search_config` and fills the persisted cost fields (`cost_model`, `cost`, `active_moe_expert_ratio`) instead of using a one-key stub config.

meenchen:
Pushed follow-up fixes for the latest review comments: 34a6baf7 handles the three substantive AutoQuantize cost-model comments, and 3411e4a1 fixes the minor API formatting issue. Validation: compileall passed, `git diff --check` passed, focused active-MoE/get-config tests passed, and `tests/unit/torch/quantization/test_autoquant.py -q -k "not data_parallel_auto_quantize"` passed with 59 passed / 1 deselected.

shengliangxu (approved):
LGTM now.

examples/llm_ptq/hf_ptq.py
```diff
@@ -1434,6 +1464,18 @@ def parse_args() -> argparse.Namespace:
...
+    ):
+        parser.error("--auto_quantize_active_moe_expert_ratio must be in the range (0.0, 1.0].")
+    if (
+        args.auto_quantize_cost_model == "weight"
+        and args.auto_quantize_active_moe_expert_ratio is not None
+    ):
+        parser.error(
+            "--auto_quantize_active_moe_expert_ratio requires "
+            "--auto_quantize_cost_model active_moe."
+        )
```
juhi10071998, line 1470: I think for this PR we should keep this, once I update the Autoquant YAML recipe PR to reflect the cost model, I will remove the CLI arg support for AutoQuant as we discussed yesterday, does that align? @meenchen , @shengliangxu ?

juhi10071998 (approved):
LGTM, left a minor comment

examples/llm_ptq/hf_ptq.py
meenchen: SGTM

cjluo-nv (approved):

Re-review: all important previous review comments are addressed in the current diff.

Critical comments resolved:
- :speech_balloon: CodeRabbit's `_infer_active_moe_expert_ratio` cross-config pairing bug — fixed; logic moved into `_auto_quantize_cost.infer_active_moe_expert_ratio` which iterates configs and checks both top-k and num-experts attributes on the same config object. New test `test_active_moe_ratio_requires_single_config_object` covers the wrapper/`text_config` mismatch.
- :speech_balloon: cjluo-nv's "`active_costs` is dead code" — resolved; the duplicated field was dropped from `initialize_candidate_stats`, and the test now asserts its absence while still validating `cost_weight`.
- :speech_balloon: cjluo-nv's "constraints normalized twice" — resolved; normalization now lives only in `_AutoQuantizeBaseSearcher.before_search`, and `auto_quantize()` in `model_quant.py` no longer pre-normalizes.
- :speech_balloon: cjluo-nv's "`_resolve_best_recipe` partial searcher config" — resolved; it now seeds from `default_search_config` and populates `cost_model`, `cost`, and `active_moe_expert_ratio` from `search_state`.
- :speech_balloon: realAsma / juhi10071998 design feedback (move cost knobs into `constraints`, unify validation) — implemented; `cost_model`/`cost` live under `constraints`, and `normalize_auto_quantize_constraints` is the single validation entry point.
- :speech_balloon: shengliangxu's magic-number / class-refactor asks — resolved via `DEFAULT_AUTO_QUANTIZE_EFFECTIVE_BITS` and the `AutoQuantizeCostModel` / `WeightCostModel` / `ActiveMoECostModel` hierarchy.

Minor remaining nits (non-blocking): `default_state_dict` hardcodes the literal `"weight"` instead of using `COST_MODEL_WEIGHT`, and `_resolve_best_recipe` reaches into `search_state` for cost fields that older checkpoints predating this PR will not contain (the `.get(..., default)` calls do degrade gracefully, but a checkpoint saved before this PR has no `cost_denominator`, so the `or sum(...)` fallback kicks in — that path is exercised but not unit-tested).

Test coverage is solid: routed-vs-shared expert weighting, `num_experts` vs `num_local_experts` config-attr fallback, single-config inference requirement, and lower-bound search-order preference for active-MoE are all covered.

examples/llm_ptq/hf_ptq.py
```diff
@@ -1391,6 +1400,27 @@ def parse_args() -> argparse.Namespace:
...
+            "quantizable weights equally. 'active_moe' scales routed MoE expert weights by "
+            "--auto_quantize_active_moe_expert_ratio, or infers top_k/num_experts from model config."
+        ),
+    )
+    parser.add_argument(
+        "--auto_quantize_active_moe_expert_ratio",
+        type=float,
+        default=None,
+        help=(
+            "Routed MoE expert active ratio for --auto_quantize_cost_model active_moe. "
```
cjluo-nv, line 1419: how is it different from --moe_calib_experts_ratio?

cjluo-nv (changes requested):
left a comment. Need clarification on the added flags

examples/llm_ptq/hf_ptq.py
meenchen: They control different things.

  `--moe_calib_experts_ratio` is a calibration behavior knob: it changes how many routed experts are exercised during calibration to improve coverage. Users may set it high, even 1.0, to calibrate all experts.

  `--auto_quantize_active_moe_expert_ratio` is only for AutoQuant cost accounting. It estimates the active expert weight footprint per token for the effective-bits budget, so it should reflect runtime routing, typically `top_k / num_experts`. It does not affect routing or calibration.

  So they intentionally can be different. For example, we may calibrate all experts with `--moe_calib_experts_ratio=1.0`, but still account active MoE cost as `8 / 128` during AutoQuant search. If omitted, the AutoQuant ratio is inferred from model config when possible.

cjluo-nv: I see. Could you add this clarification to the flag description?

cjluo-nv (approved):

examples/llm_ptq/hf_ptq.py
meenchen: They control different parts of the flow. `--moe_calib_experts_ratio` is a calibration behavior knob: it changes how many routed experts are exercised during calibration to improve coverage. `--auto_quantize_active_moe_expert_ratio` is only for AutoQuant cost accounting: it estimates active expert weight footprint per token for the effective-bits budget, typically `top_k / num_experts`, and does not change routing/calibration.

I updated the flag description to make this explicit and to point users to `--moe_calib_experts_ratio` for calibration expert coverage.

meenchen:
/ok to test [adbf9f0](https://github.com/NVIDIA/Model-Optimizer/pull/1497/commits/adbf9f0a3db61eb82a4bf3cfa557d729dc51dc54)

meenchen:
/ok to test [93c4138](https://github.com/NVIDIA/Model-Optimizer/pull/1497/commits/93c41382d9a8682d7524f0d3af9a575159aad047)

kevalmorabia97 (approved):

kevalmorabia97:
/ok to test [b6b8be2](https://github.com/NVIDIA/Model-Optimizer/commit/b6b8be2b965cedcaacc992023fbdcc96ed3cf7d3)

PR #1498, opened 2026-05-15
Add 7 nvidia/Nemotron-* calibration datasets to SUPPORTED_DATASET_CONFIG

Fridah-nv:
Registers nemotron-{sft-instruction-following-chat-v2, science-v1, competitive-programming-v1, sft-agentic-v2, math-v2, sft-swe-v2, sft-multilingual-v1} so hf_ptq.py's --dataset flag (which enumerates get_supported_datasets() automatically) can select these for PTQ calibration. Splits with heterogeneous parquet schemas that crash streaming CastError mid-iteration are excluded per inline comments.

Adds a parametrized smoke test that skips when HF_TOKEN is unset since the nvidia/Nemotron-* datasets are gated.

### What does this PR do?

Type of change: ? 

### Usage

```python
```

## Summary by CodeRabbit

* **New Features**
  * Standardized chat-style preprocessing (joining message contents) and normalized chat key to "messages" across seven Nemotron dataset variants, with explicit dataset paths and curated splits.
* **Tests**
  * Added registry-shape unit checks for each new dataset entry.
  * Added a gated integration smoke test that fetches sample data when a Hugging Face Hub token is present.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1498?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/modelopt/torch/utils/dataset_utils.py
+++ b/modelopt/torch/utils/dataset_utils.py
@@ -32,6 +32,11 @@ from tqdm import tqdm
 if TYPE_CHECKING:
     from transformers import PreTrainedTokenizerBase
 
+
+def _join_messages_content(sample: dict) -> str:
+    return "\n".join(turn["content"] for turn in sample["messages"])
+
+
 # Use dict to store the config for each dataset.
 # If we want to export more options to user like target languages, we need more standardized approach like dataclass.
 SUPPORTED_DATASET_CONFIG: dict[str, Any] = {
@@ -61,7 +66,7 @@ SUPPORTED_DATASET_CONFIG: dict[str, Any] = {
             "path": "nvidia/Nemotron-Post-Training-Dataset-v2",
             "split": ["stem", "chat", "math", "code"],
         },
-        "preprocess": lambda sample: "\n".join(turn["content"] for turn in sample["messages"]),
+        "preprocess": _join_messages_content,
         "chat_key": "messages",
     },
     "nemotron-post-training-dataset-v1": {
@@ -69,7 +74,7 @@ SUPPORTED_DATASET_CONFIG: dict[str, Any] = {
             "path": "nvidia/Nemotron-Post-Training-Dataset-v1",
             "split": ["stem", "chat", "math", "code", "tool_calling"],
         },
-        "preprocess": lambda sample: "\n".join(turn["content"] for turn in sample["messages"]),
+        "preprocess": _join_messages_content,
         "chat_key": "messages",
     },
     "nemotron-sft-instruction-following-chat-v2": {
@@ -78,7 +83,7 @@ SUPPORTED_DATASET_CONFIG: dict[str, Any] = {
             "path": "nvidia/Nemotron-SFT-Instruction-Following-Chat-v2",
             "split": ["reasoning_off"],
         },
-        "preprocess": lambda sample: "\n".join(turn["content"] for turn in sample["messages"]),
+        "preprocess": _join_messages_content,
         "chat_key": "messages",
     },
     "nemotron-science-v1": {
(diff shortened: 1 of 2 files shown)


cjluo-nv (commented):

Small, focused registry addition that follows the existing pattern for `nemotron-post-training-dataset-v{1,2}` exactly — same `messages` chat key, same `\n`.join preprocess, same shape. No correctness issues spotted; the inline comments explaining which splits are excluded due to heterogeneous parquet schemas are helpful.

The reason for nudging rather than approving: the only test added is gated on `HF_TOKEN` being present in the environment, and these `nvidia/Nemotron-*` datasets are gated on HF Hub. If CI runners don't have `HF_TOKEN` set, the test will silently skip and the registry entries (split names, dataset paths) won't actually be exercised before merge — a typo in any of the 30+ split names wouldn't be caught. Worth a human confirming whether the CI environment has `HF_TOKEN` configured for these tests, or whether a structural-only test (asserting each entry has expected keys / a callable preprocess) would be a useful complement.

Minor nit (non-blocking): the preprocess lambda `lambda sample: "\n".join(turn["content"] for turn in sample["messages"])` is now duplicated across 9 entries. Could be hoisted to a module-level helper, but matches existing style in this file.

modelopt/torch/utils/dataset_utils.py
kevalmorabia97, line 1: Have you tried directly trying these datasets (e.g. `get_dataset_dataloader("nvidia/Nemotron-SFT-Instruction-Following-Chat-v2", ...)`) without adding to this dictionary? We have a fallback logic that auto-detects dataset type and parses it if not present in the dictionary

tests/unit/torch/utils/test_dataset_utils.py
```diff
@@ -689,3 +690,33 @@ def test_dataloader_mixing_hf_and_local_jsonl(self, tmp_path, pad_tokenizer):
...
+)
+def test_get_dataset_samples_new_nemotron(dataset_key):
+    """Smoke-test the 7 newly registered nvidia/Nemotron-* calibration datasets.
+
+    Skipped when ``HF_TOKEN`` is unset because these datasets live behind the HF Hub
+    and the CI runner may not have credentials.
+    """
+    pytest.importorskip("datasets")
+    pytest.importorskip("huggingface_hub")
+    if not os.environ.get("HF_TOKEN"):
```
kevalmorabia97: @claude is this correct check? Does this als work if `hf auth login` is done instead of setting `HF_TOKEN` Env var?

kevalmorabia97:
Commits are not signed hence CI wont run. Please take a look at https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/CONTRIBUTING.md#%EF%B8%8F-signing-your-work

modelopt/torch/utils/dataset_utils.py
Fridah-nv: Tried it on main for nvidia/Nemotron-SFT-Instruction-Following-Chat-v2 — the fallback hits three issues:
 
 1. `get_dataset_samples(path, num_samples=2) → ValueError: Bad split: train. Available splits: ['reasoning_off', 'reasoning_on']. `
 Auto-detect defaults to `split=["train"] (dataset_utils.py:415);` none of the 7 Nemotron datasets has a `train` split.
2. With the correct split but no tokenizer → `ValueError: Dataset '…' has a 'messages' column but no tokenizer with apply_chat_template was provided (dataset_utils.py:256-260)`.
3. With a tokenizer but `apply_chat_template=False` → the fallback still calls `apply_chat_template` unconditionally, which fails on tokenizers without a chat template (e.g. gpt2).

We can think about more of improving the fallback pass. 
As a background, we are evaluating if these dataset should be the default calibration dataset instead. So adding them in this PR

tests/unit/torch/utils/test_dataset_utils.py
Fridah-nv: >  Does this als work if hf auth login is done instead of setting HF_TOKEN Env var?

I did a test locally and it worked

cjluo-nv (approved):

PR #1504, opened 2026-05-15
fix: layerwise mse fix for fuse-experts MoE

Fridah-nv:
### What does this PR do?

Type of change: Bug fix 

  ### Bug fixed

  Layerwise NVFP4-MSE calibration on production fused-experts MoE (GLM-5.1,
  Qwen3.6-35B-A3B) hit three failures invisible on small models:

  1. **GPU OOM during capture.** `_patched_forward`'s `"capture"` branch
     appended `(args, kwargs)` to `collected_inputs` on the layer's compute
     device. With sequential `device_map` and N calibration batches, the
     layer's GPU saturates on `bs × seq × hidden × N` activations before the
     state machine transitions to `"run"`.

  2. **Per-expert NVFP4 quantizers stayed un-promoted.**
update: this is removed, should be fixed in https://github.com/NVIDIA/Model-Optimizer/pull/1421

  3. **Resume from completed checkpoint crashed.** `detect_resume_point`
     returned `(num_layers, manifest)` for completed runs, but
     `layerwise_calibrate` never gated on it — `setup_resume` tried to load
     `next_inputs.pt` for the final layer (never written) and raised
     `FileNotFoundError`.

  4. A secondary failure — CUDA illegal-memory-access from `deepcopy`-ing CUDA
  tensors with non-zero `storage_offset` reconstructed from serialized views
  inside `_export_fused_experts` — also fed back into the checkpoint path.
### Usage

```python
```

### Testing

Ran related unit tested
E2e test with GLM5.1

## Release Notes

* **Bug Fixes**
  * Fixed checkpoint-based calibration to skip re-calibration when a checkpoint has already been completed.

* **Improvements**
  * Enhanced quantizer promotion to support cases where amax data is unavailable.
  * Improved device management and GPU memory efficiency during layerwise calibration.

* **Tests**
  * Added test coverage for checkpoint skipping and device handling behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1504?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -227,12 +227,21 @@ def _patched_forward(self, *args, **kwargs):
...
                 return output
 
             if info.mode == "capture":
-                info.collected_inputs.append((args, kwargs))
+                # Offload to CPU so the per-layer compute device doesn't OOM
+                # while accumulating thousands of batches; "run" moves back.
+                cpu = torch.device("cpu")
+                info.collected_inputs.append(
+                    (_move_to_device(args, cpu), _move_to_device(kwargs, cpu))
+                )
```
cjluo-nv, line 244: > _Bot comment._

Captured inputs are now CPU-resident, but `_patched_forward` only moves them back to the layer's device in the `"run"` branch. The actual calibration forward in `layerwise_calibrate` is `_layer_forward_loop`, which calls `m(*args, **kwargs_input)` while the layer is in `"original"` mode — no device move happens there. With HF accelerate hooks the per-module `pre_forward` silently re-bounces CPU args onto the right GPU, which is presumably why the GLM5.1 e2e run was clean. On a non-accelerate single-GPU model (params on CUDA, args on CPU) `_original_forward` will hit a device mismatch.

Consider either (a) moving tensors back to `get_module_device(layer)` inside `_layer_forward_loop` mirroring the `run`-mode code on line ~232, or (b) adding an explicit comment + assertion that this path requires accelerate-style hooks, so the assumption is at least documented.

tests/unit/torch/quantization/test_layerwise_calibrate.py
```diff
@@ -719,3 +719,56 @@ def test_mtq_quantize_layerwise_raises_for_unsupported_algorithm():
...
+
+    def forward_loop(m):
+        m(torch.randn(2, 8))
+
+    collector._patch_all_layers()
+    try:
+        inputs = collector.get_input_activations(model.layers[0], forward_loop)
+    finally:
+        collector._unpatch_all_layers()
+
```
cjluo-nv, line 741: > _Bot comment._

This asserts the capture-side invariant but doesn't exercise the calibration replay. The OOM fix's correctness for the calib_func path (CPU-stashed inputs eventually reaching a GPU layer's forward) isn't covered — both layerwise e2e tests run on CPU-only models so a CPU-arg / GPU-weight mismatch would never surface. A small GPU smoke test (or even a CPU test that asserts inputs are device-correct at the moment `_layer_forward_loop` invokes the layer) would lock in the invariant the PR description claims.

cjluo-nv (commented):

Targeted fixes for three real failures on production fused-experts MoE (capture-time OOM, NVFP4 promotion gating, completed-checkpoint resume crash). Logic looks right and rationale matches what the diff does. Two things worth a human look before merge:

- **Capture-time CPU offload may regress the non-accelerate GPU path.** `_patched_forward`'s `capture` branch now stashes `(args, kwargs)` on CPU, but `_patched_forward` only moves them back to the layer's device in the `"run"` branch. The actual calibration replay goes through `layerwise_calibrate`'s `_layer_forward_loop`, which calls `m(*args, **kwargs_input)` while `m` is in `"original"` mode (the mode is reset to `"original"` at the end of `get_input_activations`). On HF-accelerate `device_map` models the per-module `_hf_hook.pre_forward` quietly moves CPU args to the layer's device, which is presumably why the GLM5.1 e2e run worked. On a vanilla single-GPU model with no accelerate hooks, the layer's params are on CUDA and the args are on CPU — `_original_forward` will hit a device mismatch. Either confirm only the accelerate path is supported here, or move tensors to `get_module_device(layer)` inside `_layer_forward_loop` (mirroring what `run` mode does) to keep both paths working.
- **Tests don't exercise the affected flow.** `test_collected_inputs_are_cpu_at_capture` only asserts the capture invariant, and `test_layerwise_calibrate_early_returns_on_completed_checkpoint` short-circuits before any replay. The pre-existing layerwise e2e test runs on a CPU-only model, so a CPU-arg / GPU-weight mismatch wouldn't show up there. Worth either a small GPU smoke test or an explicit comment that the CPU-stash invariant relies on accelerate hooks.

Minor observations (non-blocking):
- `promote_nvfp4_static_quantizers` now promotes quantizers with no `_amax`. `from_tensor_quantizer` + `NVFP4StaticQuantizer._fake_quantize` handle `amax is None` by falling back to `super()._fake_quantize`, so pre-MSE behavior is preserved — looks correct.
- `full_restore` now loads weights/qstate to CPU and uses `load_state_dict(..., assign=False)`. Combined with `enable_weight_access_and_writeback` materializing params first, this keeps params on their original device and dodges the `deepcopy` of CUDA-view storage_offset hazard mentioned in the PR body. Reasonable.
- `detect_resume_point` no longer filters out completed runs; the new `start_layer >= num_layers` short-circuit in `layerwise_calibrate` is the right place to handle it. Public-ish helper's contract changed (docstring updated), so any other caller relying on the old "None for completed" semantics would silently see `(num_layers, manifest)` instead — `_CheckpointState.from_folder` is updated, but worth a grep before release.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -620,23 +629,23 @@ def full_restore(self, layers: nn.ModuleList, model: nn.Module) -> None:
...
                 )
                 weights = torch.load(
                     os.path.join(d, "weights.pt"),
-                    map_location=layer_device,
+                    map_location="cpu",
                     weights_only=False,
                 )
                 restore_quantizer_state(layer, dummy_config, {"quantizer_state": qstate})
-                layer.load_state_dict(weights, strict=False, assign=True)
+                layer.load_state_dict(weights, strict=False, assign=False)
```
realAsma, line 648: why are we doing assign = False? 
With this the model will get moved to CPU right?

realAsma: Does this break accelerate weight management with GPU/CPU/Disk offloading?

A simple test could be trying to see if we can still run forward after GPU/CPU/Disk offloading

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -433,6 +442,8 @@ def cache_outputs_for_next_layer_calib(
 
         next_layer = self._decoder_layers[next_idx]
         with persistent_materialization(layer):
+            # Free cached-but-unused GPU memory left over from the previous layer's calibration.
+            torch.cuda.empty_cache()
```
realAsma, line 446: `torch.cuda.empty_cache()` usually does not make your own PyTorch program able to use more memory, because PyTorch could already reuse that cached memory internally.

you could trigger a manual garbage collection to delete any unused tensors
```suggestion
            gc.collect()
            torch.cuda.empty_cache()
```

modelopt/torch/quantization/utils/layerwise_calib.py
Fridah-nv: From torch [doc](https://docs.pytorch.org/docs/2.12/generated/torch.nn.Module.html#torch.nn.Module.load_state_dict):
 - `assign=True`: state-dict tensors replace the module's parameters. Whatever device the loaded tensors live on becomes the module's device. 
  - `assign=False` (PyTorch default): the module's existing parameter tensors stay put; only their data is overwritten via `param.copy_(value_from_dict)`. `.copy_()` handles cross-device transfer automatically, so a CPU source tensor is copied into a CUDA destination tensor and the layer stays on CUDA.

So it's the opposite. I think  the previous `assign=True` only worked because the load itself used `map_location=layer_device`

realAsma: got it, makes sense

modelopt/torch/quantization/model_calib.py
Edwardf0t1, line 272: This comment need to be updated?

Edwardf0t1 (commented):
I think we can add a line in CHANGELOG for this bugfix.

PR #1506, opened 2026-05-16
[OMNIML-3707] Model-specific PTQ recipes bootstrap

shengliangxu:
### What does this PR do?

Type of change: new feature

Replaces the hardcoded model-type branches in `examples/llm_ptq/` with opt-in declarative **model-specific recipes** under `modelopt_recipes/huggingface/<model_type>/ptq/`. Any adjustment specific to a model type or instance must live in that model's recipe — there is no implicit model-specific path anymore. Users select a model's recipe with `--recipe huggingface/<model_type>/ptq/<recipe>`; users on the plain `--qformat` path get only the generic numerics.

What moved out of Python (`examples/llm_ptq/example_utils.py::build_quant_cfg` and `examples/llm_ptq/hf_ptq.py::mono_quantize`):

- **gemma / mpt** `w4a8_awq` → `awq_lite` with `alpha_step=1` (coarser search to avoid TRT-LLM overflow).
- **gemma** `int8_sq` → SmoothQuant `alpha=0.5` (default `1.0` regresses Gemma 7B).
- **phi4mm** → disable `*speech*`, `*audio*`, `*image*`, `*vision*` (quantize only the language model).
- **Nemotron VL** → disable `*vision*`, `*image*`, `*radio*`, `*visual*`, `*encoder*`, `*model_encoder*` (quantize only the decoder).

What stayed in Python:

- MTP dynamic layer exclusion in `hf_ptq.py` (depends on runtime-detected layer indices).
- `is_nemotron_vl(full_model)` detection itself, which still drives the VLM calibration loop and the post-quantize `full_model` update — only the `quant_cfg` adjustment it triggered moved into the Nemotron VL recipe.

`multinode_ptq.py` shares the same `build_quant_cfg` call site and was updated to match the new 2/3-arg signature; multinode users on `--qformat` get the generic numerics (no `--recipe` plumbing in multinode yet, so model-specific recipes are only reachable via `hf_ptq.py`).

Already-YAML recipes that were elsewhere in the tree are relocated into the same `huggingface/<model_type>/ptq/` layout so all model-specific recipes live under one convention:

- **Step3.5-Flash** — moved from `modelopt_recipes/huggingface/step3p5/Step3.5-Flash/` to `huggingface/step3p5/Step3.5-Flash/ptq/` to match the `<model>/ptq/` convention.
- **Qwen3.5 / Qwen3.6** — moved from `modelopt_recipes/models/Qwen3.5-Qwen3.6/w4a16.yaml` to per-model_type folders, anchored on the HuggingFace `model_type` (verified against transformers 5.8.1 + HF model hub `config.json` for `Qwen/Qwen3.6-27B`, `Qwen/Qwen3.6-35B-A3B`, `nvidia/Qwen3.5-397B-A17B-NVFP4`):
  - `huggingface/qwen3_5/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast.yaml` — dense `qwen3_5`
  - `huggingface/qwen3_5_moe/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast.yaml` — `qwen3_5_moe`
  - Both wrappers `$import` the shared `quant_cfg` snippet `huggingface/qwen3_5/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast.quant_cfg.yaml` (one source of truth; the two model_types share the same hybrid linear-attention + softmax-attention architecture so the rules apply identically).

Full recipe layout (`modelopt_recipes/huggingface/`):

```
gemma/ptq/{w4a8_awq,int8_sq}-kv_fp8_cast.yaml
mpt/ptq/w4a8_awq-kv_fp8_cast.yaml
phi4mm/ptq/{disabled_quantizers,nvfp4-kv_fp8_cast}.yaml
nemotron_vl/ptq/{disabled_quantizers,nvfp4-kv_fp8_cast}.yaml
qwen3_5/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast{,.quant_cfg}.yaml
qwen3_5_moe/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast.yaml
step3p5/Step3.5-Flash/ptq/nvfp4-mlp-only.yaml
```

All recipes ship with FP8 KV-cache cast (`kv_fp8_cast`). For phi4mm and nemotron_vl, `disabled_quantizers.yaml` is a multi-document list unit that `$import`s the standard `default_disabled_quantizers` exclusions and appends the model-specific ones — so each recipe imports a single disabled-quantizer slot instead of layering two, with no duplication in YAML. Each `ptq/` folder has a `README.md` describing exactly what is model-specific.

### Usage

```bash
# Gemma W4A8 AWQ with the Gemma-specific algorithm tuning + FP8 KV cache:
  --pyt_ckpt_path google/gemma-7b \
  --recipe huggingface/gemma/ptq/w4a8_awq-kv_fp8_cast \
  --export_path ./out

# Nemotron VL with vision branches excluded automatically:
  --pyt_ckpt_path nvidia/<nemotron-vl-model> \
  --recipe huggingface/nemotron_vl/ptq/nvfp4-kv_fp8_cast \
  --export_path ./out
```

### Testing

- Pre-commit recipe validator (`tools/precommit/check_modelopt_recipes.py`) loads every new recipe via `load_recipe()` — passes for all new YAMLs (gemma/mpt/phi4mm/nemotron_vl recipes + phi4mm/nemotron_vl `disabled_quantizers` snippets + qwen3_5 / qwen3_5_moe recipe wrappers + the shared `w4a16_nvfp4-fp8_attn-kv_fp8_cast.quant_cfg` snippet + Step3.5-Flash relocation).
- For qwen3_5 / qwen3_5_moe specifically, `load_recipe(...)` on both wrappers produces an identical 33-entry resolved `quant_cfg`, confirming the shared snippet is the single source of truth.
- `yamlfmt` + `markdownlint` + `bandit` + license-insertion hooks all pass.
- No tests reference the removed `build_quant_cfg(qformat, ..., model_type, ...)` signature; the only call sites (`hf_ptq.py`, `multinode_ptq.py`) were updated to the new 2/3-arg form.

- Is this change backward compatible?: ❌ — users who relied on **automatic** model-specific quant_cfg behavior via `--qformat` (gemma/mpt AWQ, gemma SmoothQuant, phi4mm exclusions, Nemotron VL exclusions) now need to pass `--recipe huggingface/<model_type>/ptq/<recipe>` to apply the model's recipe. The flag itself is unchanged; only the implicit behavior was removed.
- Did you write any new necessary tests?: ❌ — relies on the existing pre-commit recipe validator that loads each new YAML.

## Summary by CodeRabbit

* **New Features**
  * Added many model-specific PTQ recipes (Gemma, MPT, Nemotron VL, Phi‑4‑Multimodal, Qwen3.5, Qwen3.5‑MoE) and support for AWQ block-size and MoE calibration ratio in quantization options.

* **Documentation**
  * Expanded READMEs and changelog to document recipe locations, layout, and how to opt into model-specific PTQ recipes.

* **Refactor**
  * Model-specific PTQ tweaks moved to opt‑in recipes; default behavior uses generic numerics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1506?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


modelopt_recipes/huggingface/mpt/ptq/w4a8_awq-kv_fp8_cast.yaml
```diff
@@ -0,0 +1,47 @@
...
+quantize:
+  algorithm:
+    method: awq_lite
+    alpha_step: 1
+  quant_cfg:
+    - $import: base_disable_all
+    - quantizer_name: '*weight_quantizer'
+      cfg:
+        - $import: int4_per_block
+        - $import: fp8
```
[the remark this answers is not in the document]
shengliangxu: Disagree — the `cfg: [int4_per_block, fp8]` list form is the documented W4A8 AWQ pattern, not two mutually exclusive numerics on the same quantizer. It resolves to a list of two `QuantizerAttributeConfig` entries (INT4 block storage + FP8 attribute layer) that together implement the W4A8 AWQ scheme: weights are quantized to INT4 blocks but the kernel-facing dtype is FP8.

The canonical preset `modelopt_recipes/configs/ptq/presets/model/w4a8_awq_beta.yaml` (shipped in 0.45) has the **identical** structure, and this recipe only overrides the AWQ algorithm tuning (`alpha_step: 1`) — not the numerics layering. Removing the `fp8` import would break W4A8 AWQ.

Verified by `load_recipe(...).quantize.quant_cfg`: the `*weight_quantizer` entry resolves to `cfg: [{num_bits: 4, block_sizes: {-1: 128, type: static}}, {num_bits: (4, 3)}]` — exactly the same shape as the canonical W4A8 AWQ preset.

modelopt_recipes/huggingface/qwen3_5/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast.yaml
```diff
@@ -0,0 +1,39 @@
...
+# architecture. Shares its `quant_cfg` with the MoE counterpart at
+# `huggingface/qwen3_5_moe/ptq/nvfp4_mlp-fp8_attn-kv_fp8_cast.yaml`; the
+# snippet lives under
+# `huggingface/qwen3_5/ptq/nvfp4_mlp-fp8_attn-kv_fp8_cast.quant_cfg.yaml`.
+
+imports:
+  quant_cfg: huggingface/qwen3_5/ptq/nvfp4_mlp-fp8_attn-kv_fp8_cast.quant_cfg
+
+metadata:
+  recipe_type: ptq
```
cjluo-nv, line 28: > _Bot comment._

Optional: the import alias `quant_cfg` collides with the field name on line 38, making `- $import: quant_cfg` read as if it were importing the field itself. Renaming the alias (e.g. `body`, `shared_quant_cfg`, `qwen3_5_rules`) would make the wrapper easier to scan. Not blocking.

examples/llm_ptq/example_utils.py
```diff
@@ -202,10 +202,8 @@ def calibrate_loop(_model):
 
 
 def build_quant_cfg(
-    qformat,
     quant_cfg,
```
cjluo-nv, line 205: > _Bot comment._

Worth adding a unit test (or extending `tools/precommit/check_modelopt_recipes.py`) that loads each of the four new model-specific recipes and asserts the resolved `quant_cfg`/`algorithm` matches what the pre-PR `build_quant_cfg(qformat, QUANT_CFG_CHOICES[qformat], awq_block_size, model_type)` + the prior `mono_quantize` Nemotron-VL appends produced (the latter modulo the intentional new FP8 KV-cache cast). Right now nothing pins the four migrated cases to their previous numerics, so future edits to either side could silently drift.

cjluo-nv (commented):

Solid migration from Python-side model-type branches into declarative `$import`-composed YAML recipes — design-wise this fits cleanly on top of the existing recipe/`$import` machinery (PR #1253), and the PR body justifies the choice. The new `huggingface/<model_type>/<task>/` convention and the multi-document `disabled_quantizers.yaml` pattern (one `$import` slot, no two-list layering) are both nice. Backward break is loud and well-documented in CHANGELOG.

Two things to address before approval:

1. **`docs/source/guides/10_recipes.rst` is stale** — still references `models/Step3.5-Flash/nvfp4-mlp-only` in the "Model-specific recipes" table and still shows the old `models/` subtree in "Recipe repository layout". Since the file was moved (not symlinked) to `huggingface/step3p5/Step3.5-Flash/ptq/`, anyone copy-pasting from those docs will hit a load failure. Please update both spots and ideally add a short pointer to `modelopt_recipes/huggingface/README.md`.

2. **No equivalence test for the four migrated cases.** The PR description shows `load_recipe()` produces an identical 33-entry `quant_cfg` across the dense/MoE Qwen wrappers, which is great for the qwen3_5 split. But the harder claim is that the new recipes reproduce the old `build_quant_cfg`/`mono_quantize` Python paths for gemma `w4a8_awq` (alpha_step=1), gemma `int8_sq` (alpha=0.5), phi4mm exclusions, and Nemotron VL exclusions. A small pytest that resolves each new recipe and diffs the resulting `quant_cfg`/`algorithm` against `build_quant_cfg(QUANT_CFG_CHOICES[qformat], ...)` plus the Python-side appends from prior `mono_quantize`/`example_utils.build_quant_cfg` (extracted from git history) would lock the migration in. Without it, future drift in either side is silent. (Note also: all four recipes additionally enable FP8 KV-cache cast, which the prior implicit path did not — the changelog mentions this, just confirming it's intentional.)

Minor / non-blocking:

- `examples/llm_ptq/multinode_ptq.py` lost the `qformat`/`model_type` args to `build_quant_cfg`. In practice it never had a `w4a8_awq` choice in `QUANT_CFG_CHOICES`, so there's no actual regression — but worth a one-line note in the PR body that multinode is also affected by the signature change.
- `huggingface/qwen3_5/ptq/nvfp4_mlp-fp8_attn-kv_fp8_cast.yaml` aliases the imported snippet as `quant_cfg`, which collides visually with the field name (`quant_cfg: - $import: quant_cfg`). Renaming the alias (e.g. `body` or `qwen3_5_quant_cfg`) would read better but isn't worth blocking on.

---

- `docs/source/guides/10_recipes.rst:458` — > _Bot comment._

This table row still points at `models/Step3.5-Flash/nvfp4-mlp-only`, but that file was moved by this PR to `huggingface/step3p5/Step3.5-Flash/ptq/nvfp4-mlp-only`. Please update the path here and the corresponding `models/Step3.5-Flash/` entry in the "Recipe repository layout" tree below — anyone copy-pasting from these docs after merge will get a `load_recipe` failure.

modelopt_recipes/huggingface/qwen3_5_moe/ptq/nvfp4_mlp-fp8_attn-kv_fp8_cast.yaml
```diff
@@ -0,0 +1,41 @@
...
+  description: >-
+    NVFP4 MLP / FP8 attention / FP8 KV-cast PTQ recipe for HuggingFace
+    `qwen3_5_moe` models (Qwen3.5-MoE and Qwen3.6-MoE releases): NVFP4 for MoE
+    / shared-expert MLP projection weights and lm_head; FP8 for self-attention
+    and the large linear-attention projections; FP8 KV cache with constant
+    amax.
+quantize:
+  algorithm:
+    method: max
+    layerwise: false
```
juhi10071998: hi Shengliang, how does this recipe for Qwen3.5_Moe differ from the Qwen3.5 we are importing, how are we signaling this is expert related, is there anything we are adding on top? layerwise?

modelopt_recipes/huggingface/README.md
```diff
@@ -0,0 +1,70 @@
...
+these sources before placing a recipe.
+
+### Sharing content across recipes
+
+When the same body is reused by multiple recipes — for example, one
+recipe that applies to several `model_type`s, or several recipes that
+share a sub-block — extract the reused portion into a sibling
+**snippet** file with a `# modelopt-schema:` header and have each
+recipe `$import` it. The recipe wrappers stay thin; the shared body
+lives in one place.
```
juhi10071998, line 62: trying to understand this. If a specific model does not add anything new on top of existing PTQ recipe, what value do we get from placing it under it's own specific modelopt_recipes/huggingface/<model_type> folder?

modelopt_recipes/huggingface/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Name snippet files so they are obviously not runnable recipes, e.g.
+include the field name the snippet represents as a secondary suffix
+(`<recipe>.<field>.yaml`). The snippet lives next to whichever recipe
+is its natural canonical home; other importers reference it by the
+same relative path under `modelopt_recipes/`.
+
```
juhi10071998, line 69: Generally trying to understand this from a user perspective and their ease of use- how would a user know whether they reference a recipe from the `https://github.com/NVIDIA/Model-Optimizer/tree/main/modelopt_recipes/general/ptq` or look for in the specific `modelopt_recipes.huggingface/<model_type>/ptq` folder?

Is the idea that for anything that has any delta {existing path in hf_ptq.py or changes in future model in terms of recipe} that will have it's own specific recipe ptq path? if yes, I think we should see if we can make it very explicit to the user, that would be helpful

modelopt_recipes/huggingface/qwen3_5_moe/ptq/nvfp4_mlp-fp8_attn-kv_fp8_cast.yaml
shengliangxu: The 2 recipes share the same quant_cfg, but still need them separate to support 2 model types:

`qwen3_5` and `qwen3_5_moe`

modelopt_recipes/huggingface/README.md
shengliangxu: so people/agents know for the model_type, there is a model specific recipe

modelopt_recipes/huggingface/README.md
shengliangxu: In general they should look at the model specific folder first, then the general folder. The general recipes are good place to start/try out for a new model.

examples/llm_ptq/example_utils.py
shengliangxu: Thanks for the suggestion. We're not planning to add a parity test for these recipes:

1. The model-specific logic in each recipe is small enough to verify by eye — the algorithm overrides and the disable-pattern appends are short, and the original Python source is still in git history if anyone wants to diff. A dedicated test for this would carry more maintenance overhead than it saves.

2. The 'two-sided drift' risk doesn't really apply going forward: with this PR, the contract is that no model-specific quant_cfg logic is allowed in `hf_ptq.py` / `example_utils.py`. Any future per-model adjustment must live in a YAML recipe under `huggingface/<model_type>/ptq/`. There's no Python branch left for the YAML to drift from.

Happy to revisit if a concrete drift case surfaces.

meenchen (approved):
This is great, I like the refactor

meenchen (changes requested):
Previously the recipe for qwen3.5-qwen3.6 was for w4a16, now it is nvfp4. Could you change it back to w4a16?

shengliangxu:
> Previously the recipe for qwen3.5-qwen3.6 was for w4a16, now it is nvfp4. Could you change it back to w4a16?

Yeah, right, I need to make the filename explicit. The config content is correct.

shengliangxu:
> Previously the recipe for qwen3.5-qwen3.6 was for w4a16, now it is nvfp4. Could you change it back to w4a16?

fixed and also make it more explicit, with nvfp4 still attached to indicate the 4 here is nvfp4 instead of int4 or mxfp4

meenchen (approved):

PR #1508, opened 2026-05-17
Add DATASET_COMBOS for grouped calibration datasets

cjluo-nv:
## Summary

- Add ``DATASET_COMBOS`` to ``modelopt.torch.utils.dataset_utils`` — single ``--dataset`` tokens that fan out to several entries in ``SUPPORTED_DATASET_CONFIG``. The per-entry ``num_samples`` is split evenly across the members inside ``get_dataset_dataloader``.
- Two initial combos:
  - ``cnn_nemotron_v2_mix`` → ``cnn_dailymail`` + ``nemotron-post-training-dataset-v2``. Replaces the hardcoded two-element fallback list in ``hf_ptq.py`` when ``--dataset`` is omitted.
  - ``nemotron-post-training-v3`` → the seven ``nvidia/Nemotron-*`` SFT datasets registered in #1498 (mirroring the upstream [`nemotron-post-training-v3` collection](https://huggingface.co/collections/nvidia/nemotron-post-training-v3)).
- ``get_supported_datasets()`` now appends combo names so they show up in ``--dataset`` help.
- ``hf_ptq.py``'s default ``--calib_size`` bumped from ``512`` to ``1024`` so the ``cnn_nemotron_v2_mix`` combo's even split preserves the previous total sample count (was 512 per-dataset × 2 datasets = 1024; now 1024 split → 512 per-dataset × 2). ``--calib_size`` now denotes the total calibration budget regardless of combo cardinality.
- Reject mixing a combo with one of its member datasets in the same ``--dataset`` list (e.g. ``cnn_dailymail,cnn_nemotron_v2_mix``) — combo would otherwise double-sample the explicit member with a smaller per-member quota.
- Reject combo names in ``get_dataset_samples``; combos are dataloader-only. The error message points callers to ``get_dataset_dataloader``.
- Validate ``DATASET_COMBOS`` at import time: empty member lists, name collisions with ``SUPPORTED_DATASET_CONFIG``, and references to unknown datasets raise ``ValueError`` up front.

## Test plan

End-to-end validated against ``/hf-local/Qwen/Qwen3.5-0.8B`` via ``get_dataset_dataloader`` on the actual streamed data, plus 5 new unit tests in ``TestDatasetCombosExpansion`` (all 44 tests in ``test_dataset_utils.py`` pass with no regressions).

- [x] ``python -c "from modelopt.torch.utils.dataset_utils import DATASET_COMBOS, get_supported_datasets; assert 'cnn_nemotron_v2_mix' in get_supported_datasets() and 'nemotron-post-training-v3' in get_supported_datasets()"``
- [x] ``hf_ptq.py`` with no ``--dataset`` flag still calibrates on cnn_dailymail + nemotron-post-training-dataset-v2 with the same total sample count as before.
- [x] ``--dataset nemotron-post-training-v3 --calib_size 1024`` allocates 146 per member across the seven Nemotron datasets; full 1022-sample dataloader builds without error.
- [x] ``--dataset cnn_dailymail,nemotron-post-training-v3 --calib_size 256,1024`` composes correctly: 256 from cnn_dailymail (as a plain entry) plus the 7-way split from the combo. (The earlier ``cnn_dailymail,cnn_nemotron_v2_mix`` example is rejected by design since ``cnn_dailymail`` is a member of that combo.)
- [x] ``--dataset cnn_dailymail,cnn_nemotron_v2_mix`` raises ``ValueError`` with a clear message.
- [x] ``get_dataset_samples("cnn_nemotron_v2_mix", ...)`` raises ``ValueError`` pointing to ``get_dataset_dataloader``.
- [x] Unit tests: ``pytest tests/unit/torch/utils/test_dataset_utils.py`` — 44 passed.

## Post-validation fix

End-to-end testing surfaced that the original ``nemotron-sft-agentic-v2`` entry kept the two splits (``interactive_agent``, ``tool_calling``) that pyarrow's streaming JSON reader cannot parse, and excluded ``search`` which is the only clean split. Failures reproduce deterministically across cache wipes with ``force_redownload``, so they are content-level defects in the published JSONL files at the pinned revision, not local artifacts:

- ``interactive_agent`` — heterogeneous schema (``Column(.../member_id/type) changed from string to array``) at JSONL row 4.
- ``tool_calling`` — malformed JSON row in a later shard, fails at sample ~885 with ``Missing a closing quotation mark in string``.
- ``search`` — streams cleanly (verified to 2500 samples).

Commit ``10f3cfd`` corrects ``nemotron-sft-agentic-v2`` to use only ``search``, with an updated comment. The CHANGELOG calls this out as a separate bullet so the behavior change on a previously-released dataset entry from #1498 is discoverable.

## Summary by CodeRabbit

* **New Features**
  * Dataset combo support: a single dataset token can expand into multiple registered datasets with even sample splitting; predefined combos added (e.g., cnn_nemotron_v2_mix, nemotron-post-training-v3) and listed as supported.

* **Updates**
  * Default dataset when none specified now uses cnn_nemotron_v2_mix.
  * Calibration size default increased from 512 to 1024.

* **Bug Fixes**
  * nemotron-sft-agentic-v2 now uses only the deterministic "search" split to avoid streaming JSON errors.

* **Tests**
  * Added coverage for combo expansion, splitting, overlap validation, and rejection behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1508?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


meenchen (commented):

Small, cohesive feature (~40 lines). Logic looks correct: `divmod`-based even split with remainder distribution is fine, and `get_supported_datasets()` correctly surfaces combo names in `--help`. A few things worth a human look before merge:

1. **No unit tests for the new combo expansion.** The diff adds a non-trivial branch to `get_dataset_dataloader` (combo-membership lookup, even split, remainder fan-out) with zero coverage. The "test plan" in the PR body is manual end-to-end runs. A simple unit test that mocks `get_dataset_samples` and asserts the expanded `dataset_name` / `num_samples` lists for a combo input (including the remainder case, e.g. `n=10` over 7 members → `[2,2,2,1,1,1,1]`) would be cheap and would also pin combo membership against accidental edits. The existing `tests/unit/torch/utils/test_dataset_utils.py::TestGetDatasetDataloaderBlending` is the natural home.

2. **`--calib_size` default change from `512` → `1024` is broader than the PR body suggests.** The justification ("preserve previous total sample count for the `default` combo") only holds when the user is *also* using the new default-combo path. Anyone running `hf_ptq.py --dataset cnn_dailymail` (or any single dataset) without `--calib_size` now gets 2× the calibration samples they got before — a silent runtime/cost regression for existing single-dataset workflows. Two cleaner options: (a) keep `--calib_size` default at 512 and have the `default` combo *not* split (or absorb the doubling inside the combo), or (b) call out the doubling explicitly in CHANGELOG so users with pinned scripts notice. The CHANGELOG entry today only mentions the doubling in the context of the combo, not as a standalone default change.

3. **`nemotron-sft-agentic-v2` split swap is a piggybacked behavior change.** Changing `split` from `["interactive_agent", "tool_calling"]` to `["search"]` silently alters what users of that single dataset key get, independent of combos. The PR body justifies it (data-quality bugs in the upstream JSONL files at the pinned revision) and the rationale is reasonable, but it's worth a maintainer eyeballing — and ideally a CHANGELOG line of its own, since it's a behavior change on a previously-released dataset entry from #1498.

4. **Minor**: combos are only expanded inside `get_dataset_dataloader`. If someone passes a combo name straight to `get_dataset_samples` (a public function listed in `__all__` and exposed via `get_supported_datasets()`), it falls through to the "treat as HF path" branch and will fail confusingly. Either expand combos in `get_dataset_samples` too, or document that combos are dataloader-only.

None of these are blockers; flagging for owner sign-off because of the silent default change and the absence of tests on new logic.

meenchen (commented):

Re-review of PR #1508. Most previously-flagged critical items are addressed in the new diff:

- ✅ Unit tests for combo expansion: new `TestDatasetCombosExpansion` (5 tests) covers even split, remainder distribution (10/7 → [2,2,2,1,1,1,1]), plain+combo composition, overlap rejection, and combo-name in `get_dataset_samples`.
- ✅ Import-time validation of `DATASET_COMBOS` (`_validate_dataset_combos()` checks empty members, key collisions, and unknown member references).
- ✅ Combo names rejected from `get_dataset_samples` with a pointer to `get_dataset_dataloader`.
- ✅ Mixing a combo with a member dataset (`cnn_dailymail,default`) raises a clear `ValueError`.
- ✅ `nemotron-sft-agentic-v2` split swap now has its own CHANGELOG bullet.

What still warrants a human eyeball before merge:

1. **`--calib_size` default 512 → 1024 still silently doubles single-dataset workflows.** This was the most concrete concern in the previous review (meenchen item 2). The author's PR body justifies it as "`--calib_size` now denotes the total calibration budget regardless of combo cardinality," and the CHANGELOG mentions the bump — but only in the context of the `default` combo. A user running `hf_ptq.py --dataset cnn_dailymail` (single dataset, no `--calib_size`) now silently gets 1024 samples where they previously got 512. That's a 2× cost/runtime regression on existing pinned scripts that is not explicitly called out as a standalone behavior change. Worth either (a) keeping the default at 512 and absorbing the doubling inside the `default` combo, or (b) adding an explicit CHANGELOG line that single-dataset users without `--calib_size` will now get 2× the samples.

2. **Minor / not blockers**: the warning message in `hf_ptq.py` hardcodes the combo members in prose (will drift if `DATASET_COMBOS["default"]` changes); zero-allocated combo members (e.g. `nemotron-sft-mix` with `--calib_size 3` → 4 members get allocation 0) still get passed to `get_dataset_samples(name, 0)`, which triggers a pointless HF `load_dataset` call per zero-member. Only a concern for callers passing very small calib sizes against the 7-member combo; the default 1024/7 → 146 each path is fine.

Deferring to owner to confirm the `--calib_size` default change is intended for single-dataset users too.

examples/llm_ptq/hf_ptq.py
```diff
@@ -1238,7 +1239,7 @@ def parse_args() -> argparse.Namespace:
             "This argument will be parsed and converted as a list of ints."
         ),
         type=str,
-        default="512",
```
meenchen, line 1241: Curious to know, are you seeing accuracy improvement with more samples?

modelopt/torch/utils/dataset_utils.py
```diff
@@ -195,6 +197,42 @@ def _join_messages_content(sample: dict) -> str:
...
+DATASET_COMBOS: dict[str, list[str]] = {
+    "default": ["cnn_dailymail", "nemotron-post-training-dataset-v2"],
+    "nemotron-sft-mix": [
+        "nemotron-sft-instruction-following-chat-v2",
+        "nemotron-science-v1",
+        "nemotron-competitive-programming-v1",
+        "nemotron-sft-agentic-v2",
+        "nemotron-math-v2",
+        "nemotron-sft-swe-v2",
+        "nemotron-sft-multilingual-v1",
```
meenchen: Have you compared the input scale distribution between the default and nemotron-sft-mix? Interesting to know how much different it is

examples/llm_ptq/hf_ptq.py
cjluo-nv: no. Before it's 512 cnn + 512 nemotron. Now it's 1024 total samples (still 512 cnn + 512 nemotron). No behavior change

modelopt/torch/utils/dataset_utils.py
cjluo-nv: Yes I have tried. the nemotron ones will have higher input scales/amax for GLM 5.1

meenchen (approved):

modelopt/torch/utils/dataset_utils.py
```diff
@@ -560,6 +605,34 @@ def get_dataset_dataloader(
...
+    plain_inputs = {n for n in dataset_name if n not in DATASET_COMBOS}
+    for ds_name in dataset_name:
+        if ds_name in DATASET_COMBOS:
+            overlap = plain_inputs & set(DATASET_COMBOS[ds_name])
+            if overlap:
+                raise ValueError(
+                    f"--dataset includes both combo '{ds_name}' and its "
+                    f"member(s) {sorted(overlap)}; remove one to avoid "
+                    "double-sampling."
+                )
```
Fridah-nv, line 620:
```suggestion
                raise ValueError(
                    f"--{dataset_name} includes both combo '{ds_name}' and its "
                    f"member(s) {sorted(overlap)}; remove one to avoid "
                    "double-sampling."
                )
```

Fridah-nv: a nit suggestion

Fridah-nv (commented):
I notice that the current PR does not support per-member knob of num_samples, i.e. specify different number of samples for each dataset under the combo. This would be a useful feature for dataset blend study.
Is the scope of the PR only to add the calib dataset example for GLM5? Overall I haven't fully explored the nemotron v3 dataset blend, not sure if we want to recommend it as default.

cjluo-nv:
> I notice that the current PR does not support per-member knob of num_samples, i.e. specify different number of samples for each dataset under the combo. This would be a useful feature for dataset blend study. Is the scope of the PR only to add the calib dataset example for GLM5? Overall I haven't fully explored the nemotron v3 dataset blend, not sure if we want to recommend it as default.

Yes. If user want this, then they should specify them 1by1 instead of using the combo.
No. the nemotron-v3 dataset is not the default. The default is still cnn_daily + v2

CHANGELOG.rst
```diff
@@ -25,6 +25,8 @@ Changelog
...
+- The ``nemotron-sft-agentic-v2`` registered dataset (added in #1498) now uses only the ``search`` split. The previously configured ``interactive_agent`` and ``tool_calling`` splits contain content-level defects (heterogeneous schema and a malformed JSON row, respectively) that cause pyarrow's streaming JSON reader to fail deterministically.
```
kevalmorabia97, line 29: nemotron-agentic-v1 also has tool_calling split which is correctly structured if you are interested to use it. Its also much larger than nemotron-sft-agentic-v2's tool_calling split

modelopt/torch/utils/dataset_utils.py
```diff
@@ -195,6 +197,42 @@ def _join_messages_content(sample: dict) -> str:
     },
 }
 
+# Named groups of registered datasets, expanded in ``get_dataset_dataloader``.
+# Useful when callers want a single ``--dataset`` token that fans out to several
+# entries; per-dataset ``num_samples`` is split evenly across the members.
+DATASET_COMBOS: dict[str, list[str]] = {
+    "default": ["cnn_dailymail", "nemotron-post-training-dataset-v2"],
```
kevalmorabia97: having default dataloader means if we later change default to something else, people will see regression in their ptq results. Is that okay? Or should we explicitly name it something else?

CHANGELOG.rst
cjluo-nv: maybe next run. RIght now we pick these 7 datasets for GLM5.1

modelopt/torch/utils/dataset_utils.py
cjluo-nv: I see what you mean. Let me actually rename it.

modelopt/torch/utils/dataset_utils.py
shengliangxu: Also remove --, seems to be referring to `--dataset `CLI arg, but meaningless inside this utils file.

PR #1520, opened 2026-05-19
[fix] Fix NVFP4 AWQ GQA prequant fusion

ShawRong:
## Summary

- Enable the existing grouped-head pre-quant-scale fusion path for NVFP4 AWQ export.
- Add a unit regression test to ensure `requantize_resmooth_fused_llm_layers()` calls `fuse_prequant_to_linear(..., fuse_grouped_heads=True)` for `nvfp4_awq` checkpoints.

## Motivation

`fuse_prequant_to_linear()` already supports GQA/MQA grouped-head fusion, but the unified HF export path called it without enabling that mode. For GQA/MQA models, this can leave `o_proj.pre_quant_scale` unfused because hidden-size pre-quant scales do not match KV projection output channels.

Native vLLM real-quant ModelOpt NVFP4 loading does not consume `pre_quant_scale`, so AWQ exports should fold these scales when possible.

## Validation

- `uv run pytest tests/unit/torch/export/test_unified_export_hf.py -q`

## Summary by CodeRabbit

* **New Features**
  * NVFP4-AWQ model export now enables grouped-head prequant fusion for improved quantization handling.

* **Tests**
  * Added test validation for NVFP4-AWQ export behavior with grouped-head prequant fusion.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1520?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/modelopt/torch/export/unified_export_hf.py
+++ b/modelopt/torch/export/unified_export_hf.py
@@ -328,7 +328,7 @@ def requantize_resmooth_fused_llm_layers(model: torch.nn.Module):
     #    the later gate up fusion.
     # Fuse pre_quant_scale to the linear weights if possible
     if quantization_format is not None and "nvfp4_awq" in quantization_format.lower():
-        fuse_prequant_to_linear(model)
+        fuse_prequant_to_linear(model, fuse_grouped_heads=True)
 
     # Pre-process MoE experts
     for name, module in model.named_modules():
--- a/tests/unit/torch/export/test_unified_export_hf.py
+++ b/tests/unit/torch/export/test_unified_export_hf.py
@@ -1,0 +1,48 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from types import SimpleNamespace
+
+import torch
+
+import modelopt.torch.export.unified_export_hf as export_hf
+
+
+class _DummyModel(torch.nn.Module):
+    config = SimpleNamespace(is_encoder_decoder=False)
+    device = torch.device("cpu")
+
+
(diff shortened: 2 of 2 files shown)


cjluo-nv (commented):

Small, focused change that flips on the existing `fuse_grouped_heads` path of `fuse_prequant_to_linear` for `nvfp4_awq` exports, plus a mock-based regression test. Code change itself is one line and looks correct: the grouped-head branch in `fuse_prequant_to_linear` already implements the GQA averaging math, and only `nvfp4_awq` is affected (the `int4_awq`/`w4a8_awq` paths don't go through this branch).

Reasons to nudge for a human look rather than approve:

1. **Behavior change for all NVFP4 AWQ consumers, not just vLLM.** The PR description justifies folding `pre_quant_scale` into `o_proj` weights on the grounds that "Native vLLM real-quant ModelOpt NVFP4 loading does not consume `pre_quant_scale`." But `requantize_resmooth_fused_llm_layers` is the unified HF export path used by other backends (TRT-LLM, etc.) too. For GQA/MQA models, the new path replaces a per-channel `o_proj.pre_quant_scale` with the group-averaged scale folded into `v_proj`, which is a lossy approximation. Worth confirming a human is comfortable with that trade-off for non-vLLM consumers, or that those consumers handle the scaled weights correctly.

2. **Test is mock-only.** `test_nvfp4_awq_export_enables_grouped_head_prequant_fusion` monkeypatches `fuse_prequant_to_linear`, `is_moe`, `collect_shared_input_modules`, and `_fuse_shared_input_modules`, so it just asserts the kwarg is forwarded. It doesn't exercise the actual GQA averaging math in `fuse_prequant_to_linear` or verify end-to-end output equivalence on a small GQA toy model. A real (even tiny) GQA module test would catch regressions in the fusion math, not just the call site.

3. Minor: previous PRs in this area (e.g. PR #1382 fused-MoE fixes) flagged similar "silent change to fusion behavior" concerns; this is on the same surface.

License header on the new test file matches `LICENSE_HEADER` (2026) — no licensing concern.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -328,7 +328,7 @@ def requantize_resmooth_fused_llm_layers(model: torch.nn.Module):
     #    the later gate up fusion.
     # Fuse pre_quant_scale to the linear weights if possible
     if quantization_format is not None and "nvfp4_awq" in quantization_format.lower():
-        fuse_prequant_to_linear(model)
+        fuse_prequant_to_linear(model, fuse_grouped_heads=True)
```
meenchen, line 331: @ShawRong can you make this configurable instead of hardcoding? I would like to keep the original behavior because `fuse_grouped_heads=True` will impact accuracy.

shengliangxu:
Tagging @sychen52 for vis on fusion stuff

sychen52 (approved):

PR #1521, opened 2026-05-19
Support Mixed precision & Static MSE in MCore; Nemotron Super v3 NVFP4 recipe

jenchen13:
### What does this PR do?

Type of change: New recipe + Bug Fixes

 MCore and MSE fixes
  - support mixed precision export in MCore by detecting mixed precision layers in HF Quant Config
  - Restore static quantizer in MCore checkpoint restore as `NVFP4QTensor` (not TensorQuantizer which can call max calibrate. we want to skip max calibrate for static quantizer during restore) --> fixes bug during MCore export for MSE
  - Skip MSE calibration for any non-NVFP4 quantization format if `fp8_scale_sweep=True`
  - Fix dynamic block quantizer detection when `block_sizes` is dict-backed.                                                                                                       
  -  Add a YAML quantization recipe that roughly mirrors Nemotron 3 Super NVFP4 `hf_quant_config.json`
  Export bug fixes     
  - copy .py files properly from original HF ckpt (for reasoning parser etc)                                                                 
                                                                                                                       
                                                                                                                                                                                   
## Super recipe

Mirrors the published nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4 hf_quant_config.json:
 - MoE routed experts (mixer.experts.<N>.{up,down}_proj):       NVFP4 W4A4 weight MSE, group_size 16
 - MoE shared experts (mixer.shared_experts.{up,down}_proj):    FP8 per-tensor
- Mamba mixer linears (mixer.{in,out}_proj):                   FP8 per-tensor
- KV cache:                                                    FP8
rest: not quantized

### Usage

```python
```

### Testing
Tested on Nemotron model

- Is this change backward compatible?: ✅ / ❌ / N/A <!--- If ❌, explain why. -->
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ / ❌ / N/A <!--- Mandatory -->
- Did you write any new necessary tests?: ✅ / ❌ / N/A <!--- Mandatory for new features or examples. -->
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ / ❌ / N/A <!--- Only for new features, API changes, critical bug fixes or backward incompatible changes. -->

### Additional Information
<!-- E.g. related issue. -->


## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Added NVFP4 (4-bit) quantization checkpoint restore and export support for Megatron-Core models
  * Added tokenizer file export capability in model checkpoints
  * Extended quantization support for expert-parallel distributed training
  * Introduced new PTQ recipes for Nemotron-3-Super models with mixed-precision quantization

* **Bug Fixes**
  * Fixed FP8 and FP4 hardware compatibility detection on non-CUDA systems
  * Improved offline Hugging Face Hub access handling with better error messaging
  * Enhanced calibration validation for mixture-of-experts models
  * Fixed amplitude maximum validation for static block quantizers

* **Documentation**
  * Updated expert weight quantization configuration documentation

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1521?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->


jenchen13:
A continuation of https://github.com/NVIDIA/Model-Optimizer/pull/1363

modelopt/torch/export/plugins/hf_checkpoint_utils.py
```diff
@@ -152,3 +152,34 @@ def load_multimodal_components(
...
+_TOKENIZER_FILES = (
+    "tokenizer.json",
+    "tokenizer_config.json",
+    "special_tokens_map.json",
+    "chat_template.jinja",
+    "added_tokens.json",
+    "tokenizer.model",
+    "vocab.json",
+    "merges.txt",
+)
```
realAsma: Is there a way we can do this more elegantly without hard-coding these files?

realAsma: Can we copy everything from the local folder first except safe tensors? We can overwrite config and quantization json files after this step;

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -737,3 +777,48 @@ def sharded_state_dict(self, prefix="", sharded_offsets=(), metadata=None):
...
+
+
+# GPTQ layerwise calibration support
+def get_mcore_decoder_layers(model: torch.nn.Module) -> torch.nn.ModuleList | None:
+    layers = None
+    if hasattr(model, "decoder") and hasattr(model.decoder, "layers"):
+        layers = model.decoder.layers
+    if hasattr(model, "output_layer"):
+        layers.append(model.output_layer)
+    return layers
```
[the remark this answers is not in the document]
jenchen13: fixed by checking for non-null layers before appending to it

examples/specdec_bench/specdec_bench/datasets/speed.py
```diff
@@ -730,11 +730,7 @@ def _load_dataset(self, config_name_or_dataset_path: config_type | str) -> "Data
                 # Strip HF metadata from the schema to avoid Feature parsing errors
                 schema = table.schema
                 if schema.metadata and b"huggingface" in schema.metadata:
-                    new_meta = {
-                        k: v
```
jenchen13, line 734: just a linter change

h-guo18 (approved):
specdec changes LGTM

modelopt/torch/export/plugins/hf_checkpoint_utils.py
jenchen13: fixed

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -287,10 +293,20 @@ def save_pretrained(
...
         # medusa_heads and eagle_module only exist in the last stage.
         if is_last_stage_main_rank:
+            # Baseline: pull every non-safetensors file from the source (tokenizer,
+            # remote_code *.py, README, etc.). modelopt-owned files (config.json,
+            # generation_config.json, hf_quant_config.json, preprocessor_config.json)
+            # are overwritten below.
+            if self._hf_pretrained_model_name is not None and os.path.isdir(
+                self._hf_pretrained_model_name
+            ):
+                copy_non_safetensor_files_from_ckpt(self._hf_pretrained_model_name, save_directory)
```
ChenhanYu: **[IMPORTANT Regression]** This `copy_non_safetensor_files_from_ckpt` call is gated on `os.path.isdir(self._hf_pretrained_model_name)` — only the **local-directory** source path. But the previous `copy_hf_ckpt_remote_code(pretrained_model_name_or_path, save_directory)` call at the tail of `save_pretrained` (formerly around line 357 on main) is removed in this PR.

Net effect: when the source is a **HuggingFace Hub repo ID** (not a local dir), the export no longer copies `*.py` sidecar files. That directly contradicts the PR body's "copy .py files properly from original HF ckpt (for reasoning parser etc)" bullet — Hub-ID exports of `trust_remote_code` models (Nemotron, etc.) will be missing their custom modeling code.

Fix: add a fallback for the non-local case:

```python
if self._hf_pretrained_model_name is not None and os.path.isdir(self._hf_pretrained_model_name):
    copy_non_safetensor_files_from_ckpt(self._hf_pretrained_model_name, save_directory)
elif self._hf_pretrained_model_name is not None:
    # Hub-ID source: bulk copy isn't applicable, but we still need the *.py sidecars.
    copy_hf_ckpt_remote_code(self._hf_pretrained_model_name, save_directory)
```

(`copy_hf_ckpt_remote_code` is still used by `puzzletron/tools/bypassed_training/init_child_from_parent.py` and `tests/_test_utils/torch/puzzletron/utils.py`, so it's not going anywhere.)

modelopt/torch/export/plugins/hf_checkpoint_utils.py
```diff
@@ -123,3 +152,27 @@ def load_multimodal_components(
 
     print(f"Successfully loaded {len(multimodal_state_dict)} multimodal tensors")
     return multimodal_state_dict
+
+
+def copy_non_safetensor_files_from_ckpt(src: str | os.PathLike, dst: str | os.PathLike):
```
ChenhanYu, line 157: **[SUGGESTION]** The current implementation copies **everything** non-safetensors from the source — including `config.json`, `hf_quant_config.json`, `generation_config.json`, `preprocessor_config.json`. The docstring acknowledges this and says "The caller is expected to overwrite the files modelopt owns" — and today `save_pretrained` does, immediately after.

The risk is the load-bearing convention. If a future refactor adds a guarded path in `save_pretrained` that skips writing one of those files under some condition (e.g., a new `is_last_stage_main_rank` sub-branch, or a try/except around `_hf_config.save_pretrained`), the stale source file silently survives — no test failure, no warning, just a quietly-wrong exported checkpoint.

Two safer alternatives:

1. **Filter modelopt-owned files in the helper itself** with an explicit skip-list (preferred, no caller-side discipline required):

```python
_MODELOPT_OWNED_FILES = frozenset({
    "config.json",
    "generation_config.json",
    "hf_quant_config.json",
    "preprocessor_config.json",
})

def copy_non_safetensor_files_from_ckpt(src, dst):
    ...
    for entry in os.listdir(src):
        if entry in _MODELOPT_OWNED_FILES:
            continue
        if entry.endswith(".safetensors") or entry == "model.safetensors.index.json":
            continue
        ...
```

2. Or add a post-condition assert at the end of `save_pretrained` that the modelopt-owned files were rewritten (timestamp / contents check).

Option 1 removes the silent-failure mode entirely without changing today's behavior.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1360,6 +1424,31 @@ def _gather_exclude_modules(self):
...
+
+        all_layer_config_dicts = [None] * torch.distributed.get_world_size()
+        torch.distributed.all_gather_object(all_layer_config_dicts, self.layer_config_dict)
+        combined_layer_config_dict = {}
+        for layer_config_dict in all_layer_config_dicts:
+            if layer_config_dict:
+                combined_layer_config_dict.update(layer_config_dict)
+        return dict(sorted(combined_layer_config_dict.items()))
+
+    def _gather_kv_cache_dtype(self):
```
ChenhanYu, line 1445: **[SUGGESTION]** Returning the first non-`None` `kv_cache_dtype` silently picks one if ranks disagree (programmer error). For a setup bug where one attention block was configured with `fp8` and another with `nvfp4`, the writer rank would emit whichever rank `0` happened to see first into `hf_quant_config.json` with no warning.

Cheap defense:

```python
def _gather_kv_cache_dtype(self):
    local = getattr(self, "kv_cache_dtype", None)
    if not torch.distributed.is_initialized():
        return local
    all_dtypes = [None] * torch.distributed.get_world_size()
    torch.distributed.all_gather_object(all_dtypes, local)
    seen = {dt for dt in all_dtypes if dt is not None}
    if len(seen) > 1:
        raise RuntimeError(f"Inconsistent kv_cache_dtype across ranks: {seen}")
    return seen.pop() if seen else None
```

Same applies to `_gather_layer_config_dict` if a key appears with conflicting values across ranks — current `.update()` silently picks the last one.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -300,14 +316,19 @@ def save_pretrained(
...
-                tokenizer.save_pretrained(save_directory)
-            except (OSError, TypeError, ValueError, ImportError):
-                pass
+            # Fall back to AutoTokenizer for remote (non-local) sources where the
+            # bulk copy above is a no-op.
+            if self._hf_pretrained_model_name is None or not os.path.isdir(
+                self._hf_pretrained_model_name
+            ):
+                try:
+                    tokenizer = transformers.AutoTokenizer.from_pretrained(
```
ChenhanYu, line 330: **[SUGGESTION]** Behavior change worth calling out in the PR body: the previous code unconditionally tried `AutoTokenizer.from_pretrained(...).save_pretrained(save_directory)` and silently swallowed errors. That had a useful side effect — if the source dir had a partial tokenizer (e.g. `tokenizer.json` present but `tokenizer_config.json` missing or stale), AutoTokenizer would often synthesize a valid `tokenizer_config.json` on the save side.

The new code skips this entirely for local-dir sources, trusting whatever the bulk copy produced. Mostly fine for clean source dirs, but it removes the safety net for partial/stale tokenizer files. Not a blocker — just consider keeping the AutoTokenizer call as a "second pass" overwrite even in the local-dir case, since it's idempotent on a clean source and corrective on a stale one.

ChenhanYu:
## Review summary — two asks

### 1. Split the AutoQuant changes into a separate PR

The PR title and body advertise four things: MCore mixed-precision export, static MSE NVFP4 fixes, the Nemotron-3 Super NVFP4 YAML recipe, and `*.py` sidecar copy. But the diff also bundles a logically independent **AutoQuant: EP support + Megatron registration hooks + budget calc fix** workstream:

| File | What it changes |
|---|---|
| `modelopt/torch/quantization/algorithms.py` (+31/-14) | Drops "AutoQuantize does not support EP yet" warnings; adds `expert_model_parallel_group` to score / cost / best-format sync axes. Adds NemotronH MoE-in-MCore-naming regex (`linear_fc1`/`linear_fc2`). Fixes `total_weight_size` budget to come from candidate-stats (avoids a ParallelLinear weight-counting bug). |
| `modelopt/torch/quantization/model_quant.py` (+8) | Lazy-imports `register_megatron_autoquant_support` inside the `auto_quantize` entry. |
| `modelopt/torch/quantization/plugins/megatron.py` (subset of +113) | Adds `register_megatron_autoquant_support()`, `_is_supported_megatron_model`, `_megatron_grad_ckpt_context`, `_is_param_grad_enabled_for_megatron`, `get_mcore_decoder_layers`, and a `LayerActivationCollector.register_decoder_layer_support` registration for GPTQ layerwise calib. |
| `tests/unit/torch/quantization/test_autoquant.py` (+35) | `test_auto_quantize_budget_uses_no_quant_candidate_cost`. |

These are not prerequisites for the Nemotron Super recipe — the recipe is a static hand-authored YAML, not an `auto_quantize` search. And the AutoQuant EP-sync change is non-trivial enough on its own that it deserves a focused review. Splitting also lets the GPTQ `get_mcore_decoder_layers` work (which Claude's review correctly flagged as `[CRITICAL]` — `NoneType` crash + in-place `ModuleList` mutation) move at its own pace without holding back the recipe + export-fix work that the PR title is actually about.

### 2. Export-path concerns — four inline comments below

Claude's review already covered:
- **[CRITICAL]** `get_mcore_decoder_layers` correctness (agree, blocker)
- `_amax_is_invalid` double `.item()` sync (agree)
- `untie_embeddings_and_output_weights` bare `except Exception` (agree — narrow to `ImportError`)

Not in Claude's review, posting inline:

- **[IMPORTANT Regression]** `unified_export_megatron.py:309` — the new `copy_non_safetensor_files_from_ckpt` is gated on `os.path.isdir(...)`, but the previous unconditional `copy_hf_ckpt_remote_code(...)` tail call is deleted. **Hub-ID source exports lose `*.py` sidecar copy** — directly contradicts the PR's "copy .py files properly" bullet.
- **[SUGGESTION]** `hf_checkpoint_utils.py:157` — `copy_non_safetensor_files_from_ckpt` doesn't filter modelopt-owned `*.json` files; relies on the load-bearing convention that `save_pretrained` overwrites them after. A future refactor that guards any of those writes silently leaks a stale source `config.json`. Cheap fix: skip-list in the helper itself.
- **[SUGGESTION]** `unified_export_megatron.py:1440` — `_gather_kv_cache_dtype` returns first non-`None` across ranks; if ranks disagree (programmer error), the wrong one wins silently. Assert single non-None value.
- **[SUGGESTION]** `unified_export_megatron.py:325` — tokenizer auto-repair safety net is removed for local-dir sources. Behavior change worth flagging in the PR body; not a blocker.

Algorithmic correctness of the per-layer config aggregation + `_global_amax` shard handling looks good (matches Claude's read). The Nemotron Super recipe + the static MSE fixes are well-targeted.

modelopt/torch/export/unified_export_megatron.py
jenchen13: fixed

jenchen13:
@ChenhanYu I removed Autoquant & GPTQ changes

modelopt/torch/quantization/model_calib.py
```diff
@@ -491,17 +527,17 @@ def mse_calibrate(
...
+                if not fp8_scale_sweep:
+                    # Create MSE calibrator with quant_func
+                    module._calibrator = MseCalibrator(
+                        amax=initial_amax,
+                        axis=module._calibrator._axis,
+                        step_size=step_size,
+                        start_multiplier=start_multiplier,
+                        stop_multiplier=stop_multiplier,
+                        quant_func=partial(_mse_quant_func, quantizer=module),
+                    )
```
cjluo-nv, line 542: > _Bot comment._

This is a public-API behavior change that the PR doesn't document. Before this PR, calling `mse_calibrate(..., fp8_scale_sweep=True)` on an FP8 (or any non-NVFP4) quantizer would fall through to a default `MseCalibrator` (covered by the now-deleted `test_unregistered_backend_uses_default_mse_calibrator`). After this PR, with `fp8_scale_sweep=True` and no registered backend factory, a non-`is_nvfp4_static` quantizer takes neither branch, so `_calibrator` is left as the max-calibrated state and no MSE search runs.

This is what you want for the new mixed NVFP4+FP8 recipes, but it silently changes behavior for any existing caller passing `fp8_scale_sweep=True` on a homogeneous FP8 model. Please either:
- restore the default MSE fallback for non-NVFP4 quantizers under `fp8_scale_sweep=True`, or
- document the new semantics in the `mse_calibrate` docstring, the `fp8_scale_sweep` docstring on `MseCalibConfig`, and the changelog, so users aren't confused why their FP8 amax stopped moving.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1190,6 +1257,7 @@ def _pack_name_remapping(self, module, prefix, layer_type=None):
             weight_scale_list.append(weight_scale)
             weight_scale_2_list.append(weight_scale_2)
             input_scale_list.append(input_scale)
+            self._record_layer_quant_config(prefix, qformat, block_size)
```
cjluo-nv, line 1260: > _Bot comment._

`self._record_layer_quant_config(prefix, qformat, block_size)` is called inside the `for expert in module:` loop with the same `prefix` every iteration, so it just overwrites the same two `layer_config_dict` keys N times. Move this call out of the loop (same applies to the corresponding line in `_pack_name_remapping_gpt_oss`).

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4-max-calib.yaml
```diff
@@ -0,0 +1,131 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv: > _Bot comment._

Header is `Copyright (c) 2024` while `LICENSE_HEADER` in the repo root is `Copyright (c) 2026` and the sibling `super-nvfp4.yaml` added in this PR uses 2026. Please align.

cjluo-nv (commented):

Design review: this PR extends existing infrastructure rather than introducing new abstractions — the MCore exporter now feeds its per-layer quant metadata through the existing `process_layer_quant_config` / `layer_config_dict` pipeline already used by `model_config_export.py`, and the recipes ride the YAML composition system from #1253. So the architectural choice is fine. However, several substantive issues remain:

1. **Untested on the target model.** PR body says "TODO test in HF and MCore PTQ on Nemotron model" — for a recipe explicitly mirroring a published Nemotron-3-Super NVFP4 config, the equivalence/round-trip check on real hardware is the headline test. Please run it before merging.

2. **Undocumented public-API behavior change in `mse_calibrate(..., fp8_scale_sweep=True)`.** Previously, an FP8 (or any non-NVFP4) quantizer with no registered FP8-sweep backend factory fell through to a default `MseCalibrator`. The PR drops `test_unregistered_backend_uses_default_mse_calibrator` and the new control flow leaves such quantizers untouched (max-calibrated amax preserved). The new mixed-precision tests rely on this, but neither the docstring of `mse_calibrate` / `fp8_scale_sweep`, the `MseCalibConfig` description, nor the PR body / changelog mention that calling `fp8_scale_sweep=True` on a non-mixed FP8 model now silently keeps max amax instead of MSE-searching it. Either restore the fallback or document the change explicitly (changelog + docstring).

3. **PR size.** 25 files / 1468 LOC across MCore export, calibration, restore, recipes, and hf_checkpoint_utils. The four pieces are largely independent and could be split — please consider it for the next round.

4. **Minor:** redundant `_record_layer_quant_config(prefix, qformat, block_size)` calls inside the per-expert loops in `_pack_name_remapping` and `_pack_name_remapping_gpt_oss` (same `prefix` overwritten N times); `super-nvfp4-max-calib.yaml` header is `Copyright (c) 2024` while the rest of the repo and the sibling `super-nvfp4.yaml` use `2026` (LICENSE_HEADER canonical is 2026); empty `Usage` code-fence in PR body.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
```diff
@@ -0,0 +1,134 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
shengliangxu, line 1: please restructure the recipe files:

```
modelopt_recipes/huggingface/nemotron_h/Nemotron-3-Super-120B-A12B/ptq/super-nvfp4.yaml
modelopt_recipes/huggingface/nemotron_h/Nemotron-3-Super-120B-A12B/ptq/super-nvfp4-max-calib.yaml
```

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
```diff
@@ -0,0 +1,134 @@
...
+    - quantizer_name: '*mixer.experts.*input_quantizer'
+      enable: true
+      cfg:
+        block_sizes:
+          -1: 16
+          type: dynamic
+          scale_bits: e4m3
+        num_bits: e2m1
+    # Megatron-Core/PTQ names: decoder.layers.*.mlp.experts.local_experts.*.linear_fc{1,2}.
+    - quantizer_name: '*mlp.experts*weight_quantizer'
```
shengliangxu, line 67: So these configs use megatron path names right?

jenchen13: yes we need both HF and MCore names for quantization

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
jenchen13: I can't move it to a huggingface folder, these are for both HF and MCore. but I can add an NVIDIA/ folder to mimic the HuggingFace model name

ChenhanYu (approved):
Went through the export part. Great that the Autoquant related changes have been separated out.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
shengliangxu: it's still a huggingface model, we shall still put them there.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
shengliangxu: If I understand it correctly how things work with MCore-based quantization, then this does not seem to be correct way in my opinion. But I am not yet 100% sure.

In my opinion, the quantization config shall be just configure the HF format, and our library internally shall convert the HF format based config to MCore converted config.

But let me read through this part of logic and will catch you for a discussion.

Non-blocking for this PR.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
shengliangxu: just that in the yaml file, for an MCore config, add an mcore tag to the name to make it explicit.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
jenchen13: no that's not true, it does pattern matching based on the model structure and in MCore the model has different names than the HF model

We should come up with a way to unify the HF and MCore PTQ API though, I agree.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
jenchen13: hmm these recipes work for both HF and MCore since they have patterns for both ...

shengliangxu: Got it, it's your call to name them but putting them into huggingface is still correct, MCore is the tool we use for quantization, but the model itself is HF. We identify the recipes by the models not by the tools we use for quantization.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
shengliangxu: I understand the MCore models have different module paths, but if I understand it correctly, the MCore models are converted as intermediate models from HF models, which is the original input to the quant pipeline.

So for a config for the quant pipeline, or any other pipeline, it should target the original HF input model rather than the intermediate model format. Internally, we should then convert the original config to a config that makes intermediate modeling work.

Maybe my understanding is not correct. That's why I need to read through this part of logic. Will catch you up once I have a full understanding.

modelopt/torch/quantization/model_calib.py
```diff
@@ -491,17 +527,17 @@ def mse_calibrate(
...
-                # Create MSE calibrator with quant_func
-                module._calibrator = MseCalibrator(
-                    amax=initial_amax,
-                    axis=module._calibrator._axis,
-                    step_size=step_size,
-                    start_multiplier=start_multiplier,
-                    stop_multiplier=stop_multiplier,
-                    quant_func=partial(_mse_quant_func, quantizer=module),
-                )
+                if not fp8_scale_sweep:
```
meenchen, line 533: nit:
```suggestion
                else:
```

modelopt/torch/quantization/model_calib.py
```diff
@@ -246,7 +279,8 @@ def max_calibrate(
         forward_loop: A callable which takes the model as argument and
             forwards calibration data through the model.
         distributed_sync: Whether to sync input_quantizer amax across distributed processes.
-        sync_expert_weight_amax: Whether to sync weight quantizer amax across MoE experts.
+        sync_expert_weight_amax: SequentialMLP only — share one weight amax across all experts
+            in a MoE layer (within-rank sync + EP all-reduce when EP>1).
```
meenchen, line 283: Would this impact accuracy? I think for HF PTQ, experts have separate amax values

jenchen13: HF PTQ doesn't use this EP sync

meenchen: >HF PTQ doesn't use this EP sync

That's true, but I am more curious about how this will impact the accuracy. Have we run anything to measure the impact on the accuracy?

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
jenchen13: There are two options for MCore PTQ model loading: directly from HF model or directly from MCore model.
Oftentimes we choose the 2nd path to avoid the time it takes to convert a HF model to MCore model. For the 2nd path that's why we need names in both HF and MCore convention. 

I understand what you're saying though, it's not immediately obvious to users that they have two ways to pass in model for PTQ.

jenchen13: Ideally we should just have recipes in HF convention and both MCore PTQ paths should work by ModelOpt storing some HF to MCore model name mapping .. that can be improved in the future when we unify HF and MCore PTQ APIs.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
jenchen13: ok I can move to huggingface folder .. just curious do we expect to suport models outside of HF in the future?

meenchen (approved):

modelopt/torch/quantization/model_calib.py
realAsma: We should deprecate this argument one TE MoE supports seperate quantizers per expert.

realAsma (approved):

modelopt/torch/quantization/model_calib.py
jenchen13: `sync_expert_weight_amax` is by default False in `max_calibrate`. It is added for testing purposes only (e.g. to compare against TEGroupedMLP which still shares amax for all experts.

The old behavior of MCore PTQ was to always sync EP experts, which during Nano/Super PTQ experiments we  realized could lead to accuracy degradation. We removed the rank local expert sync in the `layer_sync_moe_local_experts_amax` function, but forgot to remove the cross-EP-rank sync. This PR removes the EP sync so that all experts in an MoE have different amaxes for full correctness.

jenchen13: if you see `_should_sync_amax_across_ep` in line 256 it skips EP sync  for routed experts unless you turn on `sync_expert_weight_amax`

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
shengliangxu: I do want to. But the world seems converging to HF. So I am not sure if we have the chance/need of doing that lol.

modelopt_recipes/models/Nemotron-3-Super-120B-A12B/super-nvfp4.yaml
shengliangxu: Right, that's exactly what I mean. The mapping should be done by modelopt internally, even if we load from a MCore checkpoint, we should still provide the original model's information to create the mapping.

modelopt/torch/quantization/model_calib.py
jenchen13: @realAsma agreed, I have a PR to add separate expert quantizers in TEGroupedMLP but that needs more testing ..

PR #1523, opened 2026-05-21
Add YAML based AutoQuantize recipe (currently only CLI is supported)

juhi10071998:
## What does this PR do?

**Type of change:** New feature.

Extends the recipe system (PR #1423) to support `mtq.auto_quantize`. Users
can now run autoquant via a single `--recipe <name>` flag instead of
combining `--auto_quantize_bits`, `--qformat`, `--auto_quantize_method`,
etc. The recipe carries the full search spec — candidate formats, budget,
scoring method, KV cache scheme — as a typed YAML.

Mirrors the existing PTQ recipe pattern (PR #1423): recipe is authoritative
for the search; CLI flags supply runtime concerns (dataset, calib size,
batch size).

## Usage

```bash
# Before:
python hf_ptq.py --pyt_ckpt_path Qwen/Qwen3-8B \
    --qformat nvfp4,fp8 --auto_quantize_bits 4.8 \
    --auto_quantize_method gradient --kv_cache_qformat fp8_cast \
    --calib_size 512 --export_path ./out

# After:
python hf_ptq.py --pyt_ckpt_path Qwen/Qwen3-8B \
    --recipe general/auto_quantize/nvfp4_fp8_at_4p8bits-kv_fp8_cast \
    --calib_size 512 --export_path ./out
```

Example recipe (`modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_4p8bits-kv_fp8_cast.yaml`):

```yaml
imports:
  nvfp4: configs/ptq/presets/model/nvfp4
  fp8: configs/ptq/presets/model/fp8
  kv_fp8_cast: configs/ptq/units/kv_fp8_cast

metadata:
  recipe_type: auto_quantize
  description: Mixed NVFP4 + FP8 at 4.8 effective bits with FP8 KV cache (cast).

auto_quantize:
  constraints:
    effective_bits: 4.8

  candidate_formats:
    - $import: nvfp4
    - $import: fp8

  kv_cache:
    quant_cfg:
      - $import: kv_fp8_cast

  method: gradient
  num_score_steps: 128

  disabled_layers:
    - "*lm_head*"
```

## Key design points

| # | Decision | Choice |
|---|---|---|
| 1 | `constraints` shape | Mirror upstream `mtq.auto_quantize` nested dict exactly — zero-transformation dispatch via `.model_dump(exclude_none=True)`. Future-compat with PR #1497 (cost models). |
| 2 | KV cache placement | Top-level optional `kv_cache.qformat` field, **not** per-candidate `$import` (avoids duplication when KV is shared across candidates). |
| 3 | CLI override policy | Recipe is strict-authoritative for LP search fields (`effective_bits`, candidates, etc.). CLI may fall back only for orthogonal post-step fields — today only `kv_cache.qformat`. `--auto_quantize_bits + --recipe` errors out explicitly. |
| 4 | `auto_quantize()` helper layout | Helper is a leaf orchestrator — does not know whether inputs came from CLI or recipe. All resolution happens at the dispatch site in `quantize_main`. |

## Testing

**Unit tests** (`tests/unit/recipe/test_loader.py`, 7 tests):
- Built-in recipe loads, type dispatch is correct
- Pydantic defaults applied (`method=gradient`, `num_score_steps=128`, `score_checkpoint=None`)
- `$imported` candidates byte-identical to `mtq.NVFP4_DEFAULT_CFG` / `FP8_DEFAULT_CFG` (single source of truth)
- Schema validation rejects: missing `auto_quantize` section, `<2` candidates, `effective_bits` outside `(0, 16]`
- `kv_cache` field is optional

**Equivalence smoke** on `Qwen/Qwen3-8B` at `--calib_size 512`:

```
                            CLI (--auto_quantize_bits 6.0)   Recipe (effective_bits: 6.0)
quant_algo                  MIXED_PRECISION                  MIXED_PRECISION
kv_cache_quant_algo         FP8                              FP8
Total quantized layers      252                              252
NVFP4 layers                157                              157
FP8 layers                  95                               95
hf_quant_config.json hash   4b0564bf1f613132                 4b0564bf1f613132
```

`hf_quant_config.json` is **byte-identical** between the two paths.

## Backward compatibility

✅ Yes. All four existing flows preserved:

| Flow | Path | Status |
|---|---|---|
| CLI PTQ (`--qformat nvfp4`) | unchanged | ✓ |
| CLI autoquant (`--auto_quantize_bits 4.8`) | dispatch site resolves args inline; helper is pure orchestrator now (no behavior change) | ✓ |
| PTQ recipe (`--recipe general/ptq/...`) | recipe-load gate widened to accept PTQ + AutoQuantize | ✓ |
| AutoQuantize recipe (NEW) | new dispatch branch | ✓ |

One new explicit error: `--auto_quantize_bits + --recipe` (previously would silently honor recipe). Fails fast with a clear message.

## Files changed

- `modelopt/recipe/config.py` — Pydantic schema (`AutoQuantizeConfig`, etc.) + `RecipeType.AUTO_QUANTIZE` enum + dispatch entry
- `examples/llm_ptq/hf_ptq.py` — dispatch site resolves recipe/CLI knobs and passes them to `auto_quantize()` as kw-only kwargs; helper signature is pure value-driven
- `modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_4p8bits-kv_fp8_cast.yaml` — example recipe
- `tests/unit/recipe/test_loader.py` — 7 unit tests

---

### Checklist

- [x] Backward compatible
- [x] Signed commits (`git commit -s -S`)
- [x] Pre-commit clean
- [x] New unit tests added
- [ ] CHANGELOG (n/a — new feature, but maintainer to decide)
- [ ] Claude review (`/claude review`)

## Summary by CodeRabbit

* **New Features**
  * First-class auto-quantize recipe type and a built-in mixed-precision recipe targeting ~4.8 effective bits; calibration can include labels for gradient-based auto-quantize and recipes may override KV-cache qformat.
* **Bug Fixes**
  * Unified recipe- and CLI-driven auto-quantize flows with earlier validation, clearer incompatible-option rejection, and broader batch-size probing to trigger recipe-driven runs; auto-quantize now receives resolved tuning knobs explicitly.
* **Documentation**
  * Clarified help text on budget ownership and KV-cache override behavior.
* **Tests**
  * Added tests for recipe loading, defaults, validation, and error cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1523?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/llm_ptq/hf_ptq.py
+++ b/examples/llm_ptq/hf_ptq.py
@@ -1068,7 +1068,7 @@ def quantize_main(
     # All auto_quantize() knobs are resolved here before calling the helper.
     # Helper is a leaf orchestrator — it does not know whether inputs came from
     # CLI args or a recipe.
-    if isinstance(recipe, ModelOptAutoQuantizeRecipe) or args.auto_quantize_bits:
+    if isinstance(recipe, ModelOptAutoQuantizeRecipe) or args.auto_quantize_bits is not None:
         default_disabled_layers = [
             entry["quantizer_name"]
             for entry in _default_disabled_quantizer_cfg
--- a/examples/llm_ptq/hf_ptq.py
+++ b/examples/llm_ptq/hf_ptq.py
@@ -55,7 +55,12 @@ from transformers import (
 import modelopt.torch.opt as mto
 import modelopt.torch.quantization as mtq
 import modelopt.torch.sparsity as mts
-from modelopt.recipe import ModelOptPTQRecipe, load_recipe
+from modelopt.recipe import (
+    ModelOptAutoQuantizeRecipe,
+    ModelOptPTQRecipe,
+    ModelOptRecipeBase,
+    load_recipe,
+)
 from modelopt.torch.export import (
     export_hf_checkpoint,
     export_hf_vllm_fq_checkpoint,
@@ -208,6 +213,7 @@ def make_calib_dataloader(
     tokenizer: PreTrainedTokenizerBase | None,
     device: torch.device,
     model_type: str | None,
+    recipe: ModelOptRecipeBase | None = None,
 ) -> tuple[DataLoader | _DeviceDataLoader, str | None]:
     calib_dataloader = None
     first_text_speech_dataset = None
@@ -271,8 +277,12 @@ def make_calib_dataloader(
         assert tokenizer is not None and isinstance(
             tokenizer, (PreTrainedTokenizer, PreTrainedTokenizerFast)
         ), "The PreTrainedTokenizer must be set"
-        # Labels are only needed for gradient-based auto_quantize
-        include_labels = (
+        # Labels are only needed for gradient-based auto_quantize (CLI or recipe path).
+        is_autoquant_recipe_gradient = (
+            isinstance(recipe, ModelOptAutoQuantizeRecipe)
+            and recipe.auto_quantize.method == "gradient"
+        )
+        include_labels = is_autoquant_recipe_gradient or (
             args.auto_quantize_bits is not None and args.auto_quantize_method == "gradient"
         )
 
@@ -292,12 +302,21 @@ def auto_quantize(
     args: argparse.Namespace,
     language_model: torch.nn.Module,
     calib_dataloader: DataLoader,
-    auto_quantize_method="gradient",
-    auto_quantize_score_size=128,
-    auto_quantize_checkpoint=None,
     full_model: torch.nn.Module | None = None,
+    *,
+    auto_quantize_method: str,
+    auto_quantize_score_size: int,
+    auto_quantize_checkpoint: str | None,
+    constraints: dict,
+    quantization_formats: list[dict],
+    disabled_layers: list[str],
+    kv_cache_qformat: str,
 ):
-    """Auto search quantization of multiple formats."""
+    """Pure orchestrator: build forward_step/loss_func, call mtq.auto_quantize,
+    run KV cache post-step. All knobs are explicit keyword-only args; the
+    caller (dispatch site in ``quantize_main``) is responsible for resolving
+    them from either CLI args or a recipe before invoking this function.
+    """
 
     if args.calib_with_images:
         raise NotImplementedError(
@@ -305,34 +324,9 @@ def auto_quantize(
             "Please run plain PTQ (e.g., --qformat nvfp4) with --calib_with_images."
         )
 
-    assert not (args.auto_quantize_bits and args.inference_pipeline_parallel > 1), (
+    assert args.inference_pipeline_parallel <= 1, (
         "Auto Quantization is not supported for pipeline parallel size > 1"
     )
-
-    qformat_list = args.qformat.split(",")
-    assert qformat_list, "No quantization formats provided"
-    # Check if all provided quantization formats are supported
-    assert all(
-        qformat
-        in [
-            "fp8",
-            "int8_sq",
-            "int8_wo",
-            "int4_awq",
-            "nvfp4",
-            "nvfp4_awq",
-            "nvfp4_mse",
-            "w4a8_awq",
-            "fp8_pb_wo",
-            "w4a8_mxfp4_fp8",
-            "nvfp4_mlp_only",
-            "nvfp4_experts_only",
-            "nvfp4_omlp_only",
-            "nvfp4_local_hessian",
-            "mxfp8",
-        ]
-        for qformat in qformat_list
-    ), "One or more quantization formats provided are not supported for unified checkpoint export"
 
     # When language_model is a base text model without lm_head (e.g. Gemma4TextModel),
     # use full_model's lm_head to compute logits/loss from hidden states.
@@ -384,45 +378,39 @@ def auto_quantize(
 
     language_model, _ = mtq.auto_quantize(
         language_model,
-        constraints={"effective_bits": args.auto_quantize_bits},
+        constraints=constraints,
         data_loader=calib_dataloader,
         forward_step=forward_step,
         loss_func=loss_func,  # Only used for gradient-based method
         # TRTLLM only support one quantization format or None (do not quantize, internally supported)
-        quantization_formats=[QUANT_CFG_CHOICES[format] for format in qformat_list],
+        quantization_formats=quantization_formats,  # type: ignore[arg-type]
         num_calib_steps=len(calib_dataloader),
         # AutoQuantize scoring is the costly phase; allow smaller sample counts than calibration.
         num_score_steps=min(
             len(calib_dataloader), max(auto_quantize_score_size // args.batch_size, 1)
         ),
         verbose=True,
-        # Disable all default disabled layers such as lm_head, mlp.gate, router etc.
-        disabled_layers=[
-            entry["quantizer_name"]
-            for entry in _default_disabled_quantizer_cfg
-            if "parent_class" not in entry
-        ],
+        disabled_layers=disabled_layers,
         method=auto_quantize_method,
         checkpoint=auto_quantize_checkpoint,
     )
 
     calibrate_loop = create_forward_loop(dataloader=calib_dataloader)
-    # We need to explicitly set up KV cache quantization after auto_quantize
-    enable_quant_kv_cache = args.kv_cache_qformat != "none"
+    enable_quant_kv_cache = kv_cache_qformat != "none"
     print(f"{'Enable' if enable_quant_kv_cache else 'Disable'} KV cache quantization")
     if enable_quant_kv_cache:
         kv_cache_quant_cfg = copy.deepcopy(
-            getattr(mtq, KV_QUANT_CFG_CHOICES[args.kv_cache_qformat])["quant_cfg"]
+            getattr(mtq, KV_QUANT_CFG_CHOICES[kv_cache_qformat])["quant_cfg"]
         )
         kv_cache_quant_cfg = [
             e for e in kv_cache_quant_cfg if e["quantizer_name"] != "*"
         ]  # keep other quantizers from auto_quantize
 
-        if args.kv_cache_qformat in _KV_CAST_FORMATS:
+        if kv_cache_qformat in _KV_CAST_FORMATS:
             _set_kv_cache_constant_amax(kv_cache_quant_cfg)
 
         mtq.set_quantizer_by_cfg(language_model, quant_cfg=kv_cache_quant_cfg)
-        if args.kv_cache_qformat not in _KV_CAST_FORMATS:
+        if kv_cache_qformat not in _KV_CAST_FORMATS:
             # Calibrate only the KV cache quantizers; disable all others.
             with mtq.set_quantizer_by_cfg_context(
                 language_model,
@@ -987,12 +975,20 @@ def quantize_main(
 ):
     # Load the recipe up front so we can detect layerwise calibration before batch-size probing.
     recipe = None
-    if args.recipe is not None and not args.auto_quantize_bits:
+    if args.recipe is not None:
         print(f"Use recipe {args.recipe} for quantization")
         recipe = load_recipe(args.recipe)
-        if not isinstance(recipe, ModelOptPTQRecipe):
+        if not isinstance(recipe, (ModelOptPTQRecipe, ModelOptAutoQuantizeRecipe)):
             raise TypeError(
-                f"Expected PTQ recipe, but got {type(recipe).__name__} from {args.recipe}"
+                f"Expected PTQ or AutoQuantize recipe, but got {type(recipe).__name__} "
+                f"from {args.recipe}"
+            )
+        # Fail-fast on conflicting budget sources: a recipe carries its own
+        # effective_bits, so silently honoring one over the other would be a
+        # reproducibility hazard.
+        if args.auto_quantize_bits is not None:
+            raise ValueError(
+                "Cannot combine --auto_quantize_bits with --recipe; the recipe owns the budget."
             )
 
     def _is_layerwise(obj):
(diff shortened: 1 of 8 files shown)


realAsma (commented):
Proposal looks good; What is the plan to plugin these yaml recipes to `auto_quantize` API?

juhi10071998:
> Proposal looks good; What is the plan to plugin these yaml recipes to `auto_quantize` API?

Thanks for the review Asma!
Currently I am resolving the recipe at the dispatch site (so whether it is CLI/ recipe- they get resolved) and then passed to the [auto_quantize helper](https://github.com/NVIDIA/Model-Optimizer/blob/c9098b63fb5e855958182cd760cecb50e9bef794/examples/llm_ptq/hf_ptq.py#L291) inside hf_ptq.py. some details [here](https://gitlab-master.nvidia.com/juhim/modelopt_framework/-/blob/main/autoquant_yaml.md?ref_type=heads#:~:text=4-,Helper%20is%20recipe%2Dagnostic%20(single%20resolution%20boundary),-auto_quantize()%20helper
) we can review in today's meeting 

I believe the actual `mtq.auto_quantize` API will stay as is currently as I understand, similar to the `quantize` API (the API shouldn't differentiate whether the args came from CLI/ recipe)

tests/unit/recipe/test_loader.py
```diff
@@ -243,6 +245,102 @@ def test_load_recipe_dir_missing_quantize_raises(tmp_path):
...
+    assert candidates[0].model_dump(exclude_unset=True) == mtq.NVFP4_DEFAULT_CFG
+    assert candidates[1].model_dump(exclude_unset=True) == mtq.FP8_DEFAULT_CFG
+
+
+def test_load_recipe_autoquantize_missing_section_raises(tmp_path):
+    """An AutoQuantize recipe missing the ``auto_quantize`` section is rejected."""
+    bad = tmp_path / "bad.yml"
+    bad.write_text("metadata:\n  recipe_type: auto_quantize\n")
+    with pytest.raises(ValueError, match="auto_quantize"):
+        load_recipe(bad)
```
Edwardf0t1, line 310: Should we add a `RecipeType.AUTO_QUANTIZE: "auto_quantize"` in `REQUIRED_SECTION_PER_RECIPE_TYPE` in `modelopt/recipe/loader.py`?

cc @shengliangxu @realAsma

modelopt/recipe/config.py
```diff
@@ -104,6 +106,135 @@ class ModelOptPTQRecipe(ModelOptRecipeBase):
...
+        description="Number of batches for sensitivity scoring.",
+    )
+
+    disabled_layers: list[str] = ModeloptField(
+        default=[],
+        title="Excluded layer patterns",
+        description="Glob patterns; matching layers are excluded from the search.",
+    )
+
+    score_checkpoint: str | None = ModeloptField(
```
shengliangxu: this is not something define the autoquant pipeline, should move out from recipe config.

modelopt/recipe/config.py
```diff
@@ -104,6 +106,135 @@ class ModelOptPTQRecipe(ModelOptRecipeBase):
...
+        default=None,
+        title="Search-state checkpoint path",
+        description="Path to save/restore search state for resume or cheap re-solve.",
+    )
+
+    kv_cache: AutoQuantizeKVCache | None = ModeloptField(
```
shengliangxu: let's not use a hard coded string, let's also make it a quant cfg so user can customize

modelopt/recipe/config.py
juhi10071998: hi Shengliang, this gets attached to this param in mtq.auto_quantize API- https://github.com/NVIDIA/Model-Optimizer/blob/c9098b63fb5e855958182cd760cecb50e9bef794/modelopt/torch/quantization/model_quant.py#L285

I do want to discuss the naming discrepancy in today's meeting since CLI, actual API names don't agree

modelopt/recipe/config.py
juhi10071998: I see, so if I understand correctly, they could be some presets that we have defined? https://github.com/NVIDIA/Model-Optimizer/tree/main/modelopt_recipes/configs/ptq/presets/kv?

modelopt/recipe/config.py
juhi10071998: actually, I think it makes sense what you're saying, let me update this and we can discuss more

tests/unit/recipe/test_loader.py
juhi10071998: agreed, makes sense, let me update that

modelopt/recipe/config.py
shengliangxu: yeah, we do not need to have everything accepted by an API to be in the recipe config, these are not recipe config, but runtime system config, we may expand config to those sometime in the future

modelopt/recipe/config.py
shengliangxu: just use the raw config as what you do for `candidate_formats: list[QuantizeConfig]`

examples/llm_ptq/hf_ptq.py
```diff
@@ -1083,20 +1080,104 @@ def _is_layerwise(obj):
...
     )
 
-    if args.auto_quantize_bits:
-        assert len(args.qformat.split(",")) > 1, (
-            "Auto quantization needs multiple quantization format."
-        )
+    # All auto_quantize() knobs are resolved here before calling the helper.
+    # Helper is a leaf orchestrator — it does not know whether inputs came from
+    # CLI args or a recipe.
+    if isinstance(recipe, ModelOptAutoQuantizeRecipe) or args.auto_quantize_bits is not None:
```
shengliangxu, line 1070: I think auto_quantize may not have that many users, so we may want to just remove the args support for auto quantize, move completely to yaml recipes. How do you guys think @realAsma @meenchen

juhi10071998: I agree, I can remove the auto_quantize specific args that are part of the proposed recipe now, this will help us get rid of the branch when calling the `auto_quantize` orchestrator in hf_ptq.pty

modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_4p8bits-kv_fp8_cast.yaml
realAsma, line 1: Can we add support to specify the per format effective bit cost?

Currently we have a heuristic which uses 4.0 for FP4 and 8 for FP8;

The actual effective bits of FP4 is 4.5.

It will be nice if we can specify the FP4 cost from recipe libraries.

juhi10071998: Thanks Asma, yes that makes sense.
I am thinking of something where we specify this information in the existing recipe libraries (presets), and then we import them in the candidate_formats in auto_quant recipe. 

Something like this-

```
imports:                                                                                                                                                                               
    base_disable_all: configs/ptq/units/base_disable_all                                                                                                                                 
    default_disabled_quantizers: configs/ptq/units/default_disabled_quantizers                                                                                                           
    w4a4_nvfp4_nvfp4: configs/ptq/units/w4a4_nvfp4_nvfp4                                                                                                                                 
    fp8: configs/ptq/presets/model/fp8                                                                                                                                                   
                                                                                                                                                                                         
  auto_quantize:                                                                                                                                                                         
    candidate_formats:                                                                                                                                                                   
      # Inline NVFP4 with custom effective_bits                                                                                                                                          
      - algorithm: max                                                                                                                                                                   
        effective_bits: 4.7         # ← inline value                                                                                                                                     
        quant_cfg:                                                                                                                                                                       
          - $import: base_disable_all                                                                                                                                                    
          - $import: w4a4_nvfp4_nvfp4                                                                                                                                                    
          - $import: default_disabled_quantizers                                                                                                                                         
                                                                                                                                                                                         
      # Use FP8 preset directly                                                                                                                                                          
      - $import: fp8         
```

And then in the `estimate_quant_compression`, we will perform the override and skip the quant compression computation if the user already provides the effective bits in the QuantRecipe

```
def estimate_quant_compression(quant_cfg: QuantizeConfig) -> float:                                                                                                                    
      # NEW: if user supplied effective_bits, use it                                                                                                                                     
      if quant_cfg.effective_bits is not None:                                                                                                                                           
          return quant_cfg.effective_bits / 16.0                                                                                                                                         
      # FALLBACK: existing heuristic (today's behavior)                                                                                                                                  
      cfgs = [e.get("cfg") for e in quant_cfg.quant_cfg if e.get("enable", True)]                                                                                                        
      return min(estimate_quant_compression_for_quantizer(c) for c in cfgs)  
```

cc- @shengliangxu , @Edwardf0t1 

It is still open item whether we want to include the effective bits as part of the QuantConfig or of the AutoQuantize class

juhi10071998: Implemented in [commit](https://github.com/NVIDIA/Model-Optimizer/pull/1523/changes/635227bb1a002dc001fcf172839d098983f011ca). Adds an optional effective_bits: float | None field on QuantizeConfig; when set, estimate_quant_compression() uses it directly, bypassing the heuristic. Backward compatible: existing recipes without the field hit the heuristic exactly as before.                                                                 

For now I added in the QuantizeConfig instead of the auto_quantize config.
 1. Add effective_bits to **QuantizeConfig**— preset YAMLs in configs/ptq/presets/... carry the override; recipes inherit via $import; or override inline through sibling-field merge.

 2. Add an **autoquant-only** override (e.g. a candidate_effective_bits: dict[str, float] map on AutoQuantizeConfig) — strictly autoquant scope.

For now I have picked option1 as users get the inline form as below, and in option2 we may have to add more wrapper class in the auto_quantize config {when the QuantRecipe gets constructed for auto_quantize)

```
candidate_formats:                                                                                                                                                      
      - $import: nvfp4                                                                                                                                                      
             effective_bits: 4.5      # inline override on the candidate itself                                                                                                  
      - $import: fp8     
```

Let me know your thoughts if this aligns with what you had in mind or we can discuss further

cc- @shengliangxu , @Edwardf0t1 , @meenchen

examples/llm_ptq/hf_ptq.py
```diff
@@ -292,48 +302,32 @@ def auto_quantize(
...
-    auto_quantize_checkpoint=None,
     full_model: torch.nn.Module | None = None,
+    *,
+    auto_quantize_method: str,
+    auto_quantize_score_size: int,
+    auto_quantize_checkpoint: str | None,
+    constraints: dict,
+    quantization_formats: list[dict],
+    disabled_layers: list[str],
+    kv_cache_quant_cfg: dict | None,
```
realAsma, line 313: nit: should we pass the auto_quantize recipe to this method instead? Then we might not need to expand this method arguments.

examples/llm_ptq/hf_ptq.py
```diff
@@ -292,48 +302,32 @@ def auto_quantize(
...
 ):
-    """Auto search quantization of multiple formats."""
+    """Pure orchestrator: build forward_step/loss_func, call mtq.auto_quantize,
+    run KV cache post-step. All knobs are explicit keyword-only args; the
+    caller (dispatch site in ``quantize_main``) is responsible for resolving
+    them from either CLI args or a recipe before invoking this function.
+    """
```
realAsma, line 319: nit: This seems a verbose comment to me (not aligned with https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md#-coding-standards(

examples/llm_ptq/hf_ptq.py
```diff
@@ -1067,20 +1064,104 @@ def _is_layerwise(obj):
...
+            # preset YAML's natural shape (mtq.X_CFG dicts don't carry Pydantic-filled
+            # defaults). The payload still passes the full dump to upstream.
+            def _candidate_for_mtq(fmt):
+                strict = fmt.model_dump(exclude_unset=True)
+                for cfg in QUANT_CFG_CHOICES.values():
+                    if cfg == strict:
+                        return cfg
+                return fmt.model_dump()
+
+            auto_quantize(
```
realAsma, line 1109: can we pass aq to `auto_quantize` helper here and avoid expanding the arguments here?

examples/llm_ptq/hf_ptq.py
```diff
@@ -1067,20 +1064,104 @@ def _is_layerwise(obj):
...
+
+            # mtq.auto_quantize labels candidates by upstream identity: dicts that ARE
+            # an mtq.X_CFG object get the constant's name in logs (e.g. NVFP4_DEFAULT_CFG);
+            # all other dicts get "CUSTOM_N" plus a "results may not be optimal" warning.
+            # Recipe candidates come from .model_dump() — equal by value but not identity,
+            # so we'd lose the friendly names. Substitute the canonical object back when
+            # the dump matches a known preset, so logs and the warning line up with CLI.
+            # The match check uses exclude_unset=True so it compares against the
+            # preset YAML's natural shape (mtq.X_CFG dicts don't carry Pydantic-filled
+            # defaults). The payload still passes the full dump to upstream.
```
realAsma, line 1101: verbose comment. Can we simplify and make it concise?

examples/llm_ptq/hf_ptq.py
realAsma: Sure sounds good; I am okay with this plan. We should update the README and show the correct auto_quantize usage if we decide to remove the CLI args.

tests/unit/recipe/test_loader.py
```diff
@@ -243,6 +245,115 @@ def test_load_recipe_dir_missing_quantize_raises(tmp_path):
         load_recipe(tmp_path)
 
 
+# ---------------------------------------------------------------------------
+# load_recipe — AutoQuantize recipes
+# ---------------------------------------------------------------------------
```
realAsma, line 250: redundant comment.
```suggestion
```

PR #1525, opened 2026-05-21
[OMNIML-3708] Drive PTQ example qformat choices from preset YAMLs (hf_ptq, multinode_ptq, megatron_bridge)

shengliangxu:
### What does this PR do?

Type of change: Refactor

Replace the hardcoded `QUANT_CFG_CHOICES` / `KV_QUANT_CFG_CHOICES` dicts in the PTQ
example scripts with a small `_load_preset_cfg_choices()` helper that discovers the
available qformat names by listing `modelopt_recipes/configs/ptq/presets/{model,kv}/`
and **eagerly loads every preset YAML into a plain dict at import** via the existing
`load_config(..., schema_type=QuantizeConfig).model_dump(exclude_unset=True)` path.
The directory listing becomes the source of truth for the `--qformat` /
`--kv_cache_qformat` CLI vocabulary.

> Note: an earlier revision used a lazy, copy-on-access `Mapping`. That was overkill
> for these example scripts — the previous `mtq.*_CFG` module constants were
> themselves eagerly-loaded shared dicts, and every call site that mutates a config
> already deepcopies first — so it is now a plain eager dict. A lazy variant can be
> reintroduced later if import time ever matters.

**Scope.** Three scripts carried the same hardcoded tables and all three are migrated:

- `examples/llm_ptq/hf_ptq.py` — `--qformat` / `--kv_cache_qformat`.
- `examples/llm_ptq/multinode_ptq.py` — `--qformat` / `--kv_cache_qformat`.
- `examples/megatron_bridge/quantize.py` — `--quant_cfg` / `--kv_cache_quant`
  (still also accepts any full `mtq.config.choices` name).

All three scripts share the discovery helper (`load_quant_cfg_choices`), the canonical
alias table (`QFORMAT_ALIASES`), the KV disable sentinel (`KV_CACHE_NONE`), and the
ready-built `QUANT_CFG_CHOICES` / `KV_QUANT_CFG_CHOICES` mappings via the new
**`modelopt.recipe.presets`** module — no copy lives in the example scripts anymore. The
module is a standalone import, so `import modelopt.recipe` stays cheap; only an explicit
`import modelopt.recipe.presets` triggers the eager preset load.

A small alias table preserves previously-supported short CLI names (`int8_sq`,
`nvfp4_awq`, `fp8_pb_wo`, …, plus the Megatron-Bridge `fp8_blockwise`) as deprecation
shims. It is documented as not-for-extension — new formats land as preset YAMLs, and
longer term, configurations should be authored as full recipes (`--recipe`). The alias
logic is fail-fast: an alias pointing at a missing preset raises `ValueError` at import.

Also adds `presets/kv/fp8_cast.yaml` and `presets/kv/nvfp4_cast.yaml`, composed from the
existing `kv_fp8_cast` / `kv_nvfp4_cast` unit fragments. This promotes `fp8_cast` /
`nvfp4_cast` to first-class KV presets and lets us delete the runtime
`_set_kv_cache_constant_amax` helper and all its call sites — `use_constant_amax` is now
authoritative in the YAML. The KV-calibration-skip decision is derived from the config
(`_kv_cfg_uses_constant_amax`), not from hardcoded format names.

**⚠️ CLI surface expansion (owner sign-off requested).** Because the directory listing
is now the CLI vocabulary, each script accepts **every** preset under
`presets/{model,kv}/`, not just its previously curated subset. For `hf_ptq.py` this is
the same surface the prior table covered; for `multinode_ptq.py` and the Megatron-Bridge
script it is broader (e.g. KV `fp8_affine` / `fp8_cast` / `nvfp4_cast` / `nvfp4_rotate`
are now selectable). This is intended ("the directory is the policy"), but please confirm
those two scripts are meant to expose all presets — if a given path has not validated a
format, it should be gated explicitly.

### Usage

```bash
# Old short names still work via the alias shim
python examples/llm_ptq/hf_ptq.py --pyt_ckpt_path <model> --qformat int8_sq --kv_cache_qformat fp8_cast --export_path out/

# Canonical preset basenames work directly
python examples/llm_ptq/hf_ptq.py --pyt_ckpt_path <model> --qformat int8_smoothquant --kv_cache_qformat fp8_cast --export_path out/

# A newly-added preset YAML is valid on the CLI of all three scripts with no code change
python examples/llm_ptq/hf_ptq.py --pyt_ckpt_path <model> --qformat nvfp4_awq_full --export_path out/
```

### Testing

- New: `tests/unit/recipe/test_presets.py` smoke tests for `modelopt.recipe.presets` —
  every discovered model/KV preset loads into a `quant_cfg` dict, the directory listing is
  fully covered, deprecation aliases resolve to their canonical preset, the KV `none`
  sentinel does not collide with a preset, and a stale alias raises. These guard the eager
  import-time load (one bad preset would otherwise break `import modelopt.recipe.presets`
  and every PTQ example).
- Previously verified locally (uv `.venv` py3.13 + `dev-py310-modelopt` conda):
  all previously-supported `--qformat` / `--kv_cache_qformat` names resolve to dicts
  bit-equal to the corresponding `mtq.*_CFG` constants; `fp8_cast` / `nvfp4_cast` carry
  `use_constant_amax: true` while non-cast variants do not; argparse accepts
  `--kv_cache_qformat none` plus all variants; unknown qformats raise at lookup / argparse.
- All pre-commit hooks pass (ruff, mypy, bandit, license, rst, yaml).
- Pre-merge manual checks recommended by review (run in an env with the deps installed):
  `python examples/llm_ptq/hf_ptq.py --help`, `… multinode_ptq.py --help`,
  `… megatron_bridge/quantize.py --help`.

- Is this change backward compatible?: ✅ — all previously-valid CLI values continue to work via the alias table; output configs are bit-equivalent to the prior hardcoded path.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no new deps.
- Did you write any new necessary tests?: ✅ — `tests/examples/llm_ptq/test_example_utils.py` preset-discovery smoke tests.
- Did you get Claude approval on this PR?: ☐

### Additional Information

Out of scope / follow-up: the `_AUTO_QUANTIZE_QFORMATS` table and `_canonical_qformat`
helper in `hf_ptq.py` are intentionally left hardcoded — auto_quantize is being
refactored/reimplemented and they are expected to be removed soon.


examples/llm_ptq/hf_ptq.py
```diff
@@ -446,21 +542,14 @@ def load_model(args: argparse.Namespace):
...
+            f"Quantization format is not supported for low memory mode. Supported formats: {list(QUANT_CFG_CHOICES)}"
         )
         quant_cfg = QUANT_CFG_CHOICES[args.qformat]
-        if args.kv_cache_qformat != "none":
+        if args.kv_cache_qformat != _KV_NONE:
             quant_cfg = mtq.utils.update_quant_cfg_with_kv_cache_quant(
                 quant_cfg,
-                getattr(mtq, KV_QUANT_CFG_CHOICES[args.kv_cache_qformat])["quant_cfg"],
+                KV_QUANT_CFG_CHOICES[args.kv_cache_qformat]["quant_cfg"],
             )
```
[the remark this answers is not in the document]
shengliangxu: Good catch. Fixed in 3222285fd6 by rejecting the combination in `parse_args` (fail-fast, alongside the existing `--low_memory_mode` incompatibility checks) rather than deep in `load_model`:

```python
if args.low_memory_mode and args.recipe is not None:
    parser.error(
        "--low_memory_mode does not yet support --recipe; the low-memory loader still "
        "initializes quantizers from --qformat/--kv_cache_qformat."
    )
```

The low-memory loader pre-instruments quantizers via `init_quantized_weights()` from `--qformat`/`--kv_cache_qformat`, which would diverge from the recipe-driven layout `quantize_main()` applies, so we reject it until the low-memory path consumes the recipe config.

examples/megatron_bridge/quantize.py
```diff
@@ -54,37 +54,104 @@
...
+
+QUANT_CFG_CHOICES: dict[str, dict[str, Any]] = _load_preset_cfg_choices(
+    _QUANT_PRESET_DIR, _QUANT_CFG_ALIASES
+)
+KV_QUANT_CFG_CHOICES: dict[str, dict[str, Any]] = _load_preset_cfg_choices(_KV_QUANT_PRESET_DIR)
+
+# Guard against a future ``none.yaml`` (or alias) colliding with the disable sentinel:
+# argparse would silently allow both, but the runtime branch on ``!= _KV_NONE`` would
+# become ambiguous and the user couldn't reach the real preset.
+assert _KV_NONE not in KV_QUANT_CFG_CHOICES, (
```
cjluo-nv: > _Bot comment._

This helper is byte-identical to `_load_preset_cfg_choices` in `examples/llm_ptq/example_utils.py` (modulo docstring). The PR body justifies this on the grounds that `megatron_bridge` doesn't import `example_utils` — fair, but the helper only depends on `modelopt.recipe.load_config` and `modelopt.torch.opt.config_loader.BUILTIN_CONFIG_ROOT`, which are already modelopt-internal. Worth considering relocating this into `modelopt/recipe/` (or similar) so a third caller doesn't duplicate it a third time. The aliases dict itself is per-script and should stay local.

cjluo-nv (commented):

Design is reasonable and built on existing infrastructure (the `BUILTIN_CONFIG_ROOT` + `load_config` machinery from PR #1423), not a new abstraction — so the design protocol is satisfied. The PR body addresses the obvious "why a new helper" question and explains the eager-load + alias decisions. Behavior changes (cast-format `use_constant_amax` moving from runtime patch to YAML, KV `none` sentinel split out of the dict) look correct, and the tests cover the import-time load surface.

Two reasons this still wants a human eyeball before merging:

1. **CLI surface expansion (the author explicitly asks for owner sign-off in the PR body).** `multinode_ptq.py` and `examples/megatron_bridge/quantize.py` now expose **every** preset under `presets/{model,kv}/` — including ones never previously selectable from these scripts (e.g. `mamba_moe_*`, `nvfp4_svdquant`, `nvfp4_awq_full`/`_clip`, `nvfp4_fp8_mha`, `int4_blockwise_weight_only`, plus KV `fp8_affine`, `fp8_cast`, `nvfp4_cast`, `nvfp4_rotate`). If any of those haven't been validated on the multinode or Megatron-Bridge path, "the directory is the policy" silently advertises an untested combination. An explicit allow-list per script (or a curated subset on the two non-`hf_ptq` scripts) is the safer alternative the author flags.

2. **Duplicated `_load_preset_cfg_choices` helper** in `examples/llm_ptq/example_utils.py` and `examples/megatron_bridge/quantize.py` (~40 lines, identical except for docstring). The body's justification is that `megatron_bridge` is a separate example directory that doesn't import `example_utils` — fair, but the helper depends only on `modelopt.recipe.load_config` + `modelopt.torch.opt.config_loader.BUILTIN_CONFIG_ROOT`, both already public-ish modelopt internals. Worth considering whether this belongs in `modelopt/recipe/` (or similar) so a third caller doesn't duplicate it again.

Minor:
- `hf_ptq.py` `--kv_cache_qformat` help text still names only "fp8_cast, nvfp4_cast" as cast variants; with the broader surface, the help is incomplete (not a blocker — argparse `choices=` still lists everything).
- The eager import-time load of ~30 YAMLs makes a single malformed preset break `import example_utils` for every llm_ptq script. The new tests guard this, but it's a fragility the previous static dict didn't have.

examples/llm_ptq/example_utils.py
```diff
@@ -47,11 +47,105 @@
...
+                f"Alias {alias!r} points at preset {target!r} which is not present "
+                f"under modelopt_recipes/{subdir}/."
+            )
+        choices[alias] = choices[target]
+    return choices
+
+
+QUANT_CFG_CHOICES: dict[str, dict[str, Any]] = _load_preset_cfg_choices(
+    _QFORMAT_PRESET_DIR, _QFORMAT_ALIASES
+)
```
cjluo-nv: > _Bot comment._

Eager import-time load means a single malformed preset YAML now breaks `import example_utils` for every llm_ptq script (hf_ptq, multinode_ptq, …) before the user can even reach `--help`. The new tests guard the in-tree presets, but the failure mode for an out-of-tree contributor adding a broken YAML is much worse than the previous static dict. The lazy-mapping variant the PR body mentions would be a strict improvement here; consider keeping it on the radar even if not done now.

examples/llm_ptq/hf_ptq.py
```diff
@@ -1293,7 +1272,7 @@ def parse_args() -> argparse.Namespace:
         "--kv_cache_qformat",
         required=False,
         default="fp8_cast",
-        choices=KV_QUANT_CFG_CHOICES.keys(),
+        choices=[_KV_NONE, *KV_QUANT_CFG_CHOICES],
```

Help text still hardcodes `(fp8_cast, nvfp4_cast)` as the cast variants. With the broader surface (and the `_kv_cfg_uses_constant_amax` runtime detection), this is no longer the source of truth — any preset whose `*[kv]_bmm_quantizer` cfg sets `use_constant_amax: true` will skip calibration. Worth softening to e.g. "formats whose YAML pins `use_constant_amax` (e.g. fp8_cast, nvfp4_cast)" so future cast-style presets don't drift from the docs.

examples/megatron_bridge/quantize.py
kevalmorabia97, line 1: this script was added few days ago and new so we dont have to worry about bw-breaking or deprecation warnings. Feel free to cleanup directly to match how we do in hf_ptq.py

examples/megatron_bridge/quantize.py
kevalmorabia97: +1 to move the helper to modelopt lib

shengliangxu: I was thinking these logic are temporally so I do not want to move them into core lib, but, on second thought, maybe we do need the mapping as core logic. moved.

examples/llm_ptq/hf_ptq.py
shengliangxu: Fixed in c9e3b9a701. The `--kv_cache_qformat` help no longer names `(fp8_cast, nvfp4_cast)` as *the* cast variants; it now describes the actual runtime rule — formats whose preset pins `use_constant_amax` on the KV bmm quantizer (detected by `_kv_cfg_uses_constant_amax`), e.g. fp8_cast / nvfp4_cast — so future cast-style presets stay in sync with the docs.

examples/megatron_bridge/quantize.py
shengliangxu: Thanks — done in c9e3b9a701. Dropped the `fp8_blockwise` deprecation alias (it only existed for the Megatron-Bridge script, which is new and has no backward-compat surface to preserve). The shared `QFORMAT_ALIASES` in `modelopt.recipe.presets` now matches hf_ptq.py exactly; the canonical `fp8_2d_blockwise_weight_only` (or the `fp8_pb_wo` shim) is the way to select it.

examples/llm_ptq/example_utils.py
shengliangxu: Agreed it stays on the radar. Keeping eager-load for now (deliberate — the previous `mtq.*_CFG` constants were also eagerly-loaded shared dicts), but the logic now lives in `modelopt.recipe.presets`, its docstring explicitly notes a lazy/copy-on-access variant can be reintroduced if load time ever matters, and `tests/unit/recipe/test_presets.py` guards the in-tree presets. The lazy swap would be localized to `load_quant_cfg_choices` if we decide to do it.

cjluo-nv (commented):

Re-review: most prior critical issues are resolved.

**Resolved since last review**
- ✅ `_load_preset_cfg_choices` deduplication: helper now lives in `modelopt/recipe/presets.py` and is shared by all three scripts (per kevalmorabia97's earlier ask).
- ✅ `use_constant_amax` is now derived from the loaded KV cfg (`_kv_cfg_uses_constant_amax`) instead of a hardcoded name set; YAML is authoritative.
- ✅ `--kv_cache_qformat` help text updated to describe the runtime rule (`use_constant_amax` on the KV bmm quantizer) rather than naming `fp8_cast`/`nvfp4_cast` as the source of truth.
- ✅ `--low_memory_mode` + `--recipe` rejected fail-fast in `parse_args`.
- ✅ Auto-quantize qformat validation now canonicalizes via `_canonical_qformat`/`QFORMAT_ALIASES` before checking `_AUTO_QUANTIZE_QFORMATS` (canonical names), so canonical preset basenames and short aliases both pass.
- ✅ Eager-load fragility guarded by `tests/unit/recipe/test_presets.py` (every preset loads, aliases resolve, KV `none` sentinel has no preset clash, stale alias raises).

**Still warrants human sign-off**
- :speech_balloon: PR body explicitly asks owner sign-off on **CLI surface expansion** for `multinode_ptq.py` and `examples/megatron_bridge/quantize.py`. Because `choices=` is now `[KV_CACHE_NONE, *KV_QUANT_CFG_CHOICES]` / `list(QUANT_CFG_CHOICES)`, both scripts now silently advertise every preset under `presets/{model,kv}/` — including ones never previously selectable on these paths (`mamba_moe_*`, `nvfp4_svdquant`, `nvfp4_awq_full/_clip`, `nvfp4_fp8_mha`, `int4_blockwise_weight_only`, plus KV `fp8_affine`/`fp8_cast`/`nvfp4_cast`/`nvfp4_rotate`). If any of these have not been validated on multinode-FSDP2 or Megatron-Bridge, "the directory is the policy" exposes untested combos. An explicit per-script allow-list (or curated subset) is the safer alternative the author calls out.
- Minor (not blocking): hf_ptq's `--kv_cache_qformat` default is still the literal `"fp8_cast"`. If that preset is ever renamed, the parser will inject an invalid default. Hoisting to a constant + asserting membership next to the `KV_CACHE_NONE` collision guard would mirror the existing fail-fast posture.
- Minor (acknowledged): the eager import-time YAML load is now wider blast-radius than the previous static dicts (a single malformed YAML breaks `import modelopt.recipe.presets` for all three scripts). Tests guard the in-tree set; the lazy-mapping variant remains on the radar per the module docstring.

kevalmorabia97 (approved):
MBridge quantize changes LGTM

meenchen (approved):
LGTM

PR #1531, opened 2026-05-21
[OMNIML-4788] specdec_bench: configuration.json provenance + upload_to_s3

ChenhanYu:
> [!WARNING]
> **Breaking on-disk schema change (specdec_bench v1.0.0).** This PR renames the acceptance-rate metric fields across `AcceptanceRate` / `MTBench` / `SpecBench` writers:
>
> | Old (pre-1.0.0) | New (1.0.0) |
> |---|---|
> | `Request_AR` | `Request_AL` |
> | `Category_AR` | `Category_AL` |
> | `Average_AR` | `Average_AL` |
> | — | `Joint_Acceptance_Rate` (new) |
>
> The renamed values were always **acceptance length** (mean tokens generated per inference step), not a rate, and the visualizer reads `*_AL`. Pre-1.0.0 runs in S3 have `*_AR` and no `Joint_AR`; they must be re-run or post-processed before comparing. The visualizer aggregates runs by `specdec_bench` major version so accidental cross-methodology comparison is blocked.

### What does this PR do?

Type of change: new feature

Adds reproducibility provenance to `specdec_bench/configuration.json` and ports `upload_to_s3.py` from `iputterman/specdec_bench@main` (personal-namespace fork) into upstream. This is the first PR in a multi-stage migration off Izzy's fork now that he's left the team. Tracked in [OMNIML-4788](https://jirasw.nvidia.com/browse/OMNIML-4788).

**Provenance fields added to configuration.json** (alongside existing argv / engine_version / gpu / python_version):

- `specdec_bench_version` — methodology semver declared in `specdec_bench/__init__.py`. Bump minor on additive metrics, major on changed metric *definitions*. The visualizer (Phase 4 of the migration) will aggregate runs by major version so plots don't accidentally compare across methodology changes.
- `specdec_bench_sha`, `modelopt_sha`, `modelopt_version`, `nmm_sandbox_sha`, `container_image` — code/runtime provenance. Each prefers an env var set by the harness (`SPECDEC_BENCH_SHA`, `MODELOPT_SHA`, `MODELOPT_VERSION`, `NMM_SANDBOX_SHA`, `CONTAINER_IMAGE`) and falls back to `git rev-parse` / `modelopt.__version__` when running standalone. The env-var preference is necessary because the runtime container has no `.git/` (the launcher packager tarballs source without git metadata) and may not have `modelopt` installed.
- `checkpoint.{path, size_bytes, index_sha256, index_source}` — cheap reproducibility fingerprint that hashes `model.safetensors.index.json` (or `config.json` fallback). Changes whenever any tensor changes.
- `serving_config` — engine-level config dict captured after init via a new `Model.get_serving_config()` method. VLLM dumps `AsyncEngineArgs` + the live `vllm_config.to_dict()`; SGLANG dumps the `engine_kwargs` passed to `sgl.Engine`; TRTLLM left at the base default `{}` for a later iteration.
- `timestamp` — UTC ISO 8601.

**Other changes**
- `upload_to_s3.py` + `specdec_bench/s3_utils.py` ported from iputterman/specdec_bench@main. Recognizes run dirs by sentinel files, refuses to overwrite existing S3 prefixes.
- `_redact_config` allowlists `tokenizer`, `tokenizer_path`, `tokenizer_mode`, `tokenizer_revision` so the model path stops being redacted (latent bug from substring-matching `token` ⊂ `tokenizer`).
- `requirements_speed.txt`: `boto3`, `botocore` added (used by `s3_utils`).

**Out of scope** (deferred to Phase 1b / Phase 2):
- `--sweep_config` driver that emits per-run-dir nesting `<sweep>/<NNN_dataset_c<conc>>/`
- `--s3_upload` flag baked into `run.py` itself
- Launcher auto-injection of the provenance env vars (currently the example YAML sets them statically)
- `container_digest` (enroot integration) and full GPU/driver inventory
- TRTLLM `get_serving_config()`

### Usage

```bash
# Run a smoke benchmark (Qwen3.5-4B + vLLM + MTP draft=3) — example YAML included
uv run launch.py --yaml examples/Qwen/Qwen3.5-4B/specdec_bench_mtp.yaml --yes

# After it lands, upload the run directory to S3:
S3_KEY_ID=team-specdec-workgroup \
S3_SECRET=... \
python upload_to_s3.py /path/to/sweep_dir s3://team-specdec-workgroup/results
```

### Testing

Cluster-tested end-to-end on cw-dfw (Slurm job 11978794, NeMo Run experiment `cicd_1779403623`, ~19 min wall):

- Qwen3.5-4B + vLLM + MTP draft=3 + SPEED-Bench-Internal/qualitative (80 requests)
- `configuration.json` (22 KB) populated all eight new provenance fields
- `Request_AR` mean 3.327 (vs 3.330 on the pre-Phase-1a run — within noise; methodology unchanged)
- `upload_to_s3.py` (real upload, not dry-run) landed [s3://team-specdec-workgroup/results/qwen35_4_mtp_smoke_2026-05-21/specdec_bench_mtp/](https://app.s8k.io/buckets/team-specdec-workgroup/?prefix=results%2Fqwen35_4_mtp_smoke_2026-05-21%2F) where the visualizer at http://10.131.132.205:8080 can pick it up.

  - `configuration.json` only gains fields. `upload_to_s3.py` / `s3_utils.py` are new files. `Model.get_serving_config()` default = `{}` so existing subclasses without an override behave as before.
  - `boto3` / `botocore` are Apache 2.0 (permissive); `upload_to_s3.py` + `s3_utils.py` are ported from a private NVIDIA repo with explicit copyright headers retained.
  - Validated by cluster smoke (see Testing). Will add unit-tests for `dump_env` provenance fields and `upload_to_s3._discover_runs` in a follow-up.
  - Internal-facing tooling.
- Did you get Claude approval on this PR?: ❌ (triggering after open)

### Additional Information

Tracked in JIRA [OMNIML-4788](https://jirasw.nvidia.com/browse/OMNIML-4788). The full multi-phase plan is on that ticket's SPEC block — this PR is Phase 1a.

Cherry-picked alongside the harness change are two example YAMLs (`examples/Qwen/Qwen3.5-4B/specdec_bench.yaml` for the NONE autoregressive baseline, `..._mtp.yaml` for the MTP run) that gave us cluster-test evidence. Can be split out if preferred.

## Summary by CodeRabbit

* **New Features**
  * Added an S3 upload CLI for benchmark results with dry-run and skip-existing options
  * Automatic capture of run configuration, provenance and redacted environment into saved config
  * Models now export serving configuration for reproducible runs
  * New launcher entrypoint and example job configs for Qwen SPEED-Bench runs

* **Documentation**
  * README section describing S3 upload usage and supported local layouts

* **Bug Fixes / Changes**
  * Acceptance-rate metric keys renamed in output (AR -> AL)

* **Tests / CI**
  * New tests for redaction and S3 utilities; CI now runs specdec_bench examples

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1531?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


examples/specdec_bench/specdec_bench/s3_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    for run_dir in run_dirs:
+        run_key = f"{s3_prefix}/{run_dir.name}"
+        print(f"  {run_dir.name} → s3://{bucket}/{run_key}")
+        if s3_prefix_exists(s3, bucket, run_key):
+            raise ValueError(
+                f"S3 destination already exists: s3://{bucket}/{run_key} — refusing to overwrite"
+            )
+    # All clear — upload everything
+    for run_dir in run_dirs:
+        _upload_files(s3, run_dir, bucket, f"{s3_prefix}/{run_dir.name}")
```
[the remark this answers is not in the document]
ChenhanYu: Fixed in 02fc28cb8a — removed upload_directory and s3_credentials_from_args_or_env from s3_utils.py (also dropped the now-unused 'import os'). If Phase 1b's --s3_upload flag wants them back, they can come back at that point. Public surface now matches what upload_to_s3.py actually calls.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,155 @@ def postprocess_gptoss(text):
...
+    safetensors index file (changes whenever any shard's contents change).
+    Falls back to hashing config.json for non-sharded checkpoints.
+    """
+    if model_dir is None:
+        return None
+    try:
+        p = Path(model_dir)
+        if not p.is_dir():
+            return {"path": str(model_dir)}
+        size_bytes = sum(f.stat().st_size for f in p.rglob("*") if f.is_file())
```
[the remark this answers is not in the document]
ChenhanYu: Fixed in 02fc28cb8a — _checkpoint_provenance now sums only the shards listed in model.safetensors.index.json's weight_map + the index file itself. Falls back to rglob only when no index exists (single-file checkpoints / unusual layouts). Multi-100B-with-hundreds-of-tokenizer/cache-files startup tax is gone.

tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench_mtp.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - HF_MODEL_CKPT: <<global_vars.hf_model>>
+      - HF_LOCAL: /hf-local
+      # Provenance env vars consumed by dump_env() in examples/specdec_bench/run.py.
+      # These are static here for demonstration; a launcher auto-injection
+      # (planned for Phase-2 in OMNIML-4788) will compute them at launch time
+      # so the YAML stays SHA-free.
+      - SPECDEC_BENCH_SHA: "62d1686570448e8c4c057592f45c23a5965ec206"
+      - MODELOPT_SHA: "58633f0a6021a0ca9f250ca53e2090d2ebfea182"
+      - MODELOPT_VERSION: "0.45.0dev-108-g58633f0a60"
+      - NMM_SANDBOX_SHA: "079b41847544cf142e25820197294eea94ad216e"
```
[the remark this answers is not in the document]
ChenhanYu: Fixed in 02fc28cb8a — stripped the hardcoded SHAs. Replaced with a comment block in the YAML's environment: section that documents the env-var contract (SPECDEC_BENCH_SHA / MODELOPT_SHA / MODELOPT_VERSION / NMM_SANDBOX_SHA / CONTAINER_IMAGE) and points at the planned Phase-2 launcher auto-injection. Standalone runs fall back to git rev-parse / modelopt.__version__ in dump_env.

tools/launcher/common/specdec_bench/run.sh
```diff
@@ -0,0 +1,46 @@
...
+# Backend-agnostic specdec_bench entrypoint. The caller's YAML supplies --engine
+# (VLLM | SGLANG | TRTLLM | NONE) and the dataset / sweep flags via "args"; this
+# script just sources service_utils.sh, installs the speed-bench deps, and execs
+# examples/specdec_bench/run.py.
+#
+# Required env: HF_MODEL_CKPT
+# Optional env: HF_DRAFT_MODEL_CKPT (consumed by --draft_model_dir if the YAML passes it)
+
+if ! pip install -r modules/Model-Optimizer/examples/specdec_bench/requirements_speed.txt; then
+    report_result "FAIL: specdec_bench: pip install requirements_speed.txt failed"
```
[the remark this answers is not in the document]
ChenhanYu: Fixed in 02fc28cb8a — added 'pip show boto3 || pip show datasets || pip show seaborn' gate so the install only runs in a cold container. Matches the pattern in other common/*/run.sh scripts.

examples/specdec_bench/upload_to_s3.py
```diff
@@ -0,0 +1,175 @@
...
+            if args.skip_existing:
+                print(f"  Skipped: {exc}")
+                skipped += 1
+            else:
+                print(f"  Error: {exc}")
+                errors += 1
+
+    print(f"\nDone: {uploaded} uploaded, {skipped} skipped, {errors} failed.")
+    if errors:
+        sys.exit(1)
```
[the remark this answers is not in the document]
ChenhanYu: Fixed in 02fc28cb8a — broadened the upload loop's except to also catch boto3 ClientError / per-file OSError / anything else, record per-run failures, and keep going so a partial sweep upload still attempts the remaining runs. Sweep-level non-zero exit still flags overall failure. ValueError still distinguishes the prefix-exists case for the --skip-existing path.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,155 @@ def postprocess_gptoss(text):
...
+    specdec_bench_dir = Path(__file__).resolve().parent
+    config["specdec_bench_sha"] = (
+        os.environ.get("SPECDEC_BENCH_SHA") or _git_sha(specdec_bench_dir)
+    )
+    config["modelopt_version"] = (
+        os.environ.get("MODELOPT_VERSION") or _get_modelopt_version()
+    )
+    config["modelopt_sha"] = (
+        os.environ.get("MODELOPT_SHA") or _git_sha(specdec_bench_dir.parents[2])
+    )
```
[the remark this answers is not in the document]
ChenhanYu: Documented in 02fc28cb8a — added a comment block above the fallback noting the in-tree-only assumption and pointing at the env-var path (MODELOPT_SHA) as the supported one for non-in-tree deployments. Not changing the fallback itself: the env-var preference is already in place, and the in-tree fallback covers standalone dev usage.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,155 @@ def postprocess_gptoss(text):
...
+
+def dump_env(args, save_dir, overrides=None):
+    """Write configuration.json to save_dir capturing run args, engine version, and provenance.
+
+    `overrides` is merged in last and is the channel for runtime-only fields
+    (e.g. the live engine's serving_config dict from runner.get_serving_config()).
+    """
+    config = _redact_config(vars(args).copy())
+    if overrides:
+        config.update(_redact_config(overrides))
```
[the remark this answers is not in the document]
ChenhanYu: Fixed in 02fc28cb8a — replaced top-level _redact_config with a recursive _redact_value() that walks dicts/lists/tuples and redacts on key name match at any depth. Verified locally: nested hf_token inside serving_config and aws_secret_access_key inside serving_config.nested both get redacted. tokenizer + tokenizer_revision stay un-redacted via the existing allowlist.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,155 @@ def postprocess_gptoss(text):
...
+
+    config["engine_version"] = _get_engine_version(config.get("engine"))
+    config["gpu"] = _get_gpu_name()
+    config["python_version"] = sys.version
+    config["argv"] = sys.argv[:]
```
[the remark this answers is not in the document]
ChenhanYu: Fixed in 02fc28cb8a — added _redact_argv() and applied it to sys.argv[:] in dump_env. Masks the value after any --<sensitive-name> flag and rewrites --<sensitive-name>=VAL forms. Uses the same _is_sensitive_key predicate as the recursive redactor.

ChenhanYu:
## Round 2 — addressed all 8 items from `@claude review`

Pushed in 02fc28cb8a:

**[IMPORTANT]** Recursive `_redact_value()` now walks dicts/lists at any depth. Nested `serving_config.hf_token` / `runtime_params.aws_secret_access_key` are now redacted. `tokenizer` + `tokenizer_*` stay un-redacted via the existing allowlist.

**[IMPORTANT]** Stripped the hardcoded SHAs from `specdec_bench_mtp.yaml` and replaced with a comment block documenting the env-var contract. The standalone-run fallback in `dump_env` (git rev-parse / `modelopt.__version__`) is the only path exercised until the Phase-2 launcher auto-injection lands.

**[SUGGESTION]** `_redact_argv()` now masks values after sensitive flag names (`--hf_token VALUE` and `--api_key=VALUE` both handled).

**[SUGGESTION]** Comment added near the `_git_sha(parents[2])` fallback noting it's in-tree-only; env var is the supported path for vendored / installed deployments.

**[SUGGESTION]** `_checkpoint_provenance` now sums only the shards listed in `model.safetensors.index.json`'s `weight_map` (plus the index file itself). Falls back to `rglob` only when no index exists.

**[SUGGESTION]** Removed unused `upload_directory` + `s3_credentials_from_args_or_env` (and the now-unused `import os`) from `s3_utils.py`. Public surface matches what's actually wired.

**[SUGGESTION]** `upload_to_s3.py`'s except now catches boto3 `ClientError` / per-file `OSError` / anything else, records per-run failures, and keeps going. `ValueError` still distinguishes the prefix-exists path for `--skip-existing`.

**[SUGGESTION]** `pip install` in `run.sh` now gated by `pip show boto3 || pip show datasets || pip show seaborn` so it only runs in a cold container.

All 8 threads resolved with per-thread replies pointing at the same SHA.

examples/specdec_bench/specdec_bench/__init__.py
```diff
@@ -13,3 +13,18 @@
...
+# comparisons across methodology changes.
+#
+# 1.0.0: rename Request_AR / Category_AR / Average_AR → *_AL across the
+#        SpecBench / AcceptanceRate / MTBench metric writers, AND add
+#        Joint_Acceptance_Rate to the AcceptanceRate metric. The renamed
+#        values were always acceptance LENGTH (mean tokens generated per
+#        inference step), not a rate, and the visualizer reads *_AL.
+#        Pre-1.0.0 runs in S3 have *_AR and no Joint_AR; they must be
+#        re-run or post-processed before comparing.
+__version__ = "1.0.0"
```
cjluo-nv: > _Bot comment._

The `__version__ = "1.0.0"` bump and the documented `Request_AR → Request_AL` / `Average_AR → Average_AL` rename are a breaking on-disk schema change that's not visible in this PR's title ("configuration.json provenance + upload_to_s3"). Any downstream consumer parsing `Request_AR` from older `acceptance_rate.json` / `specbench_results.json` files breaks silently. Please either split this into its own PR titled around the methodology bump, or call it out explicitly at the top of the PR body — and update README/SPECBENCH_PORTING.md if they reference the old field names.

cjluo-nv (commented):

Note for the Slack operator: I don't have shell access — only repo navigation tools (`get_file`, `list_directory`, `search_code`). I can't `pwd` or run commands; I reviewed the PR by reading files in the NVIDIA/Model-Optimizer repo at this PR's branch.

This PR is doing more than the title advertises and lacks tests for non-trivial new logic. Main concerns:

1. **Hidden breaking change buried under "provenance + S3 upload"**. The PR also renames `Request_AR` / `Average_AR` / `Category_AR` → `*_AL` across `acceptance_rate.py`, `mtbench.py`, `specbench.py` and bumps `specdec_bench/__init__.py` to `__version__ = "1.0.0"`. The version-bump comment explicitly says "Pre-1.0.0 runs in S3 have `*_AR` and no `Joint_AR`; they must be re-run or post-processed before comparing." That is a methodology breaking change to the on-disk schema and deserves either (a) a separate PR with that as the title, or (b) a much more prominent callout in the PR body — not a single line in a docstring. Reviewers and downstream consumers (visualizer, any external script grepping `Request_AR`) will miss it.

2. **`specbench.py` inconsistency from the rename.** The Rich table column header is still `"Average AR"` and the MTBench print is still `f"{MTBENCH_TOPICS[i]} Average AR: {category_ar}"` while the underlying dict key is `Category_AL` / `Average_AL`. Either rename the user-visible strings too or keep the dict keys — pick one.

3. **No unit tests for any of the new logic.** Author acknowledges this and defers to a follow-up, but `dump_env` (provenance + redaction + UID derivation), `_redact_argv`, `_redact_value`, `_checkpoint_provenance`, `parse_s3_path`, and `_discover_runs` are all pure-Python with no external deps and trivially unit-testable. The "validated by cluster smoke" claim doesn't exercise the redaction allowlist behavior or the sweep-vs-flat layout discovery in `upload_to_s3._discover_runs`.

4. **Scope: 15 files / +771 across 6 directories bundling**: provenance dump, S3 upload tool, AR→AL rename + version bump, two launcher YAMLs, a launcher `run.sh`, and `_redact_config` allowlist fix. These split cleanly and would be much easier to review (and revert independently) as 2–3 PRs.

Plus several smaller issues — see inline.

examples/specdec_bench/specdec_bench/metrics/specbench.py
```diff
@@ -96,12 +96,12 @@ def _pretty_print_results(self):
         table.add_column("Average AR", justify="right", style="green")
```
cjluo-nv: > _Bot comment._

After the AR→AL rename the column header still reads `"Average AR"` while the values come from `Category_AL`. Also `mtbench.py:55` still prints `"... Average AR: {category_ar}"`. Either rename the user-facing strings to match ("Average AL" / "Acceptance Length") or revert the dict-key rename. Mixed terminology in the same output is more confusing than either consistent choice.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,250 @@ def postprocess_gptoss(text):
...
+        str(getattr(args, "model_dir", "") or ""),
+        str(getattr(args, "speculative_algorithm", "") or ""),
+        str(getattr(args, "concurrency", "") or ""),
+        str(config.get("specdec_bench_sha") or ""),
+    ])
+    config["uid"] = hashlib.sha256(uid_parts.encode()).hexdigest()[:16]
+
+    os.makedirs(save_dir, exist_ok=True)
+    with open(os.path.join(save_dir, "configuration.json"), "w") as f:
+        json.dump(config, f, indent=4, default=str)
```
cjluo-nv, line 307: > _Bot comment._

No unit tests for any of `dump_env`, `_redact_argv`, `_redact_value`, `_checkpoint_provenance`, or the `_SENSITIVE_KEY_ALLOWLIST` behavior. These are pure-Python and trivially testable without GPUs/engines. The PR description's "cluster smoke run populated all eight provenance fields" doesn't cover: (a) `--hf_token=secret` redaction in argv form, (b) nested-dict secret redaction in `serving_config`, (c) the `tokenizer`/`tokenizer_path` allowlist (which is the latent-bug fix the PR body advertises), (d) sharded-vs-unsharded `_checkpoint_provenance` paths, (e) flat-vs-sweep layout discovery in `upload_to_s3._discover_runs`, (f) `parse_s3_path` with edge inputs (`s3://bucket`, `s3://bucket/`, trailing slashes). Please add these before merge — leaving them "for a follow-up" on internal tooling tends to mean never.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,250 @@ def postprocess_gptoss(text):
...
+
+
+def _redact_config(config):
+    return _redact_value(config)
+
+
+def _redact_argv(argv):
+    """Mask values that follow a sensitive flag name (e.g. --hf_token VALUE).
+
+    Conservative: only masks when the previous element looks like a flag whose
```
cjluo-nv, line 232: > _Bot comment._

`_redact_argv` only handles `--long-flag` forms. Short flags (`-t value`) and `=`-less forms after a sensitive `--flag` rely on the `prev_is_sensitive` carry. Worth at minimum a docstring caveat that short flags aren't redacted, and ideally a unit test fixing the supported forms.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,250 @@ def postprocess_gptoss(text):
...
+    # resolves on Hub; standalone runs leave both empty.
+    config["jira_ticket"] = os.environ.get("JIRA_TICKET") or None
+    config["huggingface_model_id"] = os.environ.get("HUGGINGFACE_MODEL_ID") or None
+
+    # Deterministic per-run UID. SHA-256 of the small set of inputs that
+    # together identify this run; same inputs → same UID, so accidental
+    # re-uploads of the same run dedupe in the visualizer.
+    uid_parts = "|".join([
+        str(config["timestamp"]),
+        str(getattr(args, "model_dir", "") or ""),
```
cjluo-nv: > _Bot comment._

The `uid` mixes `config["timestamp"]` (which is recomputed `datetime.now()` on every call) with stable inputs. "Same inputs → same UID" only holds if the run is launched at the exact same UTC ISO second, which it never is — so the UID is effectively just a timestamp hash and the dedupe-on-re-upload rationale in the comment doesn't actually work. Either drop the timestamp from the UID inputs (so identical model_dir + algo + concurrency + sha hashes the same), or rewrite the comment to reflect what the field actually does.

examples/specdec_bench/specdec_bench/s3_utils.py
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""S3 upload utilities for specdec_bench results."""
+
+from pathlib import Path
+
```
cjluo-nv: > _Bot comment._

`S3_DEFAULT_ENDPOINT = "https://pdx.s8k.io"` hardcodes an NVIDIA-internal S3-compatible endpoint into Apache-2.0 OSS code. Default-empty would be cleaner; users can supply via `--endpoint` / `S3_ENDPOINT` env. At minimum, document in `upload_to_s3.py`'s help that the default endpoint is NVIDIA-internal and external users must override.

examples/specdec_bench/specdec_bench/s3_utils.py
```diff
@@ -0,0 +1,75 @@
...
+    """Upload all files under local_dir without any existence check."""
+    for file_path in sorted(local_dir.rglob("*")):
+        if not file_path.is_file():
+            continue
+        rel = file_path.relative_to(local_dir).as_posix()
+        key = f"{s3_prefix}/{rel}"
+        s3.upload_file(str(file_path), bucket, key)
+        print(f"  Uploaded: s3://{bucket}/{key}")
+
+
```
cjluo-nv: > _Bot comment._

If the destination is `s3://bucket` or `s3://bucket/`, `parse_s3_path` returns `prefix=""`, and `upload_run_dir` (called from `upload_to_s3.py` after `f"{s3_prefix_base}/{local_root.name}"`) ends up uploading to keys with an unintended leading `/`. Worth either rejecting empty `s3_prefix_base` upfront in `upload_to_s3.main()` or normalizing in `parse_s3_path`.

examples/specdec_bench/upload_to_s3.py
```diff
@@ -0,0 +1,183 @@
...
+    for local_run_dir, s3_key in queue:
+        print(f"\n{local_run_dir.name} → s3://{bucket}/{s3_key}")
+        try:
+            upload_run_dir(s3, local_run_dir, bucket, s3_key)
+            uploaded += 1
+        except ValueError as exc:
+            # Raised by upload_run_dir when the destination already exists.
+            if args.skip_existing:
+                print(f"  Skipped: {exc}")
+                skipped += 1
```
cjluo-nv, line 265: > _Bot comment._

Catching bare `Exception` and continuing on per-run failure is fine for sweep resilience, but please at least log `traceback.format_exc()` (or print to stderr) — `f"{type(exc).__name__}: {exc}"` swallows the stack on a real boto3 ClientError and makes a failed sweep upload nearly impossible to debug from the printed log alone.

tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench_mtp.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - --save_dir /scratchspace/specdec_bench_mtp
+      - --trust_remote_code
+    environment:
+      - HF_MODEL_CKPT: <<global_vars.hf_model>>
+      - HF_LOCAL: /hf-local
+      # dump_env() in examples/specdec_bench/run.py reads provenance from these
+      # env vars when present; they are intentionally NOT set here. Hardcoded
+      # SHAs in a committed YAML would stale-lie about the build that's
+      # actually running. Set them in the harness (planned for Phase 2 in
+      # OMNIML-4788, where slurm.py / launch.py will compute and inject them
```
cjluo-nv, line 40: > _Bot comment._

The 11-line comment block here describing how Phase 2 of OMNIML-4788 will inject env vars is going to stale-rot the moment the harness lands. Better to keep the YAML lean and put this in the JIRA / a `tools/launcher/common/specdec_bench/README.md`. Also: `JIRA_TICKET` / `HUGGINGFACE_MODEL_ID` (read by `dump_env`) are documented nowhere — neither here nor in the `specdec_bench` README.

examples/specdec_bench/specdec_bench/__init__.py
ChenhanYu: Fixed in de31ccad07 (string consistency) + PR body now leads with a ⚠️ breaking-change callout (see updated description).

Pushing back on the "split into its own PR" ask: the AR→AL rename and the provenance/upload work share the same downstream consumer (the visualizer + the S3 bucket schema), and the visualizer's `Joint_AR` ingest is part of the same Phase 1a deliverable on [OMNIML-4788](https://jirasw.nvidia.com/browse/OMNIML-4788). Splitting would force a two-step migration on the same artifact format. The honest fix is to make the breaking change *visible*, which the body now does. Open to revisiting if reviewer prefers split — let me know.

examples/specdec_bench/specdec_bench/metrics/specbench.py
ChenhanYu: Fixed in de31ccad07.

- `specbench.py:96` column header: `"Average AR"` → `"Average AL"`
- `mtbench.py:55,57`: print strings → `"... Average AL: ..."` / `"Average AL: ..."`
- `acceptance_rate.py:81`: print string → `"Average AL: ..."`

Now consistent with the underlying `*_AL` dict keys. No mixed terminology in the same output.

examples/specdec_bench/specdec_bench/utils.py
ChenhanYu: Added in de31ccad07 under `examples/specdec_bench/tests/`. 41 tests, all pass on CPU. Covers the security-adjacent surface (per scope choice):

- `test_redaction.py` — `_is_sensitive_key` allowlist (your (c)), `_redact_value` nested-dict + list redaction (your (b)), `_redact_argv` long-flag and `--flag=val` forms (your (a) for long-flag form), tuple handling, primitive passthrough.
- `test_upload_to_s3.py` — `parse_s3_path` edge inputs (your (f): `s3://bucket`, `s3://bucket/`, trailing slashes), `_discover_runs` flat vs sweep (your (e)), plus a `test_empty_prefix_no_leading_slash` regression for the leading-`/` concern you raised separately.

**Deferred to follow-up** (acknowledged): `dump_env` end-to-end and `_checkpoint_provenance` sharded-vs-unsharded — these need filesystem fixtures (real safetensors index, fake shards) and a subprocess mock for `_git_sha`. The work is non-trivial to factor cleanly and the load-bearing security surface (redaction) is now covered. Tracking on [OMNIML-4788](https://jirasw.nvidia.com/browse/OMNIML-4788).

Run with:
```bash
cd examples/specdec_bench && python -m pytest tests/ -v
```

examples/specdec_bench/specdec_bench/utils.py
ChenhanYu: Tests added in de31ccad07 pin the supported forms explicitly:

- `test_long_flag_separate_value` — `--hf_token VALUE`
- `test_long_flag_equals_form` — `--api_key=VALUE`
- `test_prev_is_sensitive_carry_resets` — guards the `prev_is_sensitive` carry doesn't leak past one token
- `test_allowlist_keys_not_redacted_in_argv` — `--tokenizer /path` survives the allowlist

Short flags (`-t VALUE`) are intentionally out of scope: no specdec_bench arg today uses a short form for a sensitive value, and adding heuristic short-flag handling would risk false positives. If/when we add a sensitive short flag, the docstring will be updated alongside.

examples/specdec_bench/specdec_bench/utils.py
ChenhanYu: Fixed in de31ccad07.

`config["timestamp"]` removed from `uid_parts`. `uid` is now the per-config fingerprint the visualizer needs for grouping repeats; `timestamp` stays in `config["timestamp"]` for ordering. Comment rewritten to describe what the field actually does — "same inputs → same UID" now holds.

examples/specdec_bench/specdec_bench/s3_utils.py
ChenhanYu: Already addressed in 0b759a29a4 (pushed shortly before your review landed — the file you're commenting on no longer exists).

`s3_utils.py` was merged into `upload_to_s3.py` and all defaults dropped:

```python
# from upload_to_s3.py
parser.add_argument("--endpoint", default=os.environ.get("S3_ENDPOINT", ""), ...)
parser.add_argument("--key-id",   default=os.environ.get("S3_KEY_ID", ""),   ...)
parser.add_argument("--secret",   default=os.environ.get("S3_SECRET", ""),   ...)
```

The header comment in the merged file explicitly calls out "Defaults are empty — no team-specific infrastructure leaks into this public example." External users supply via `--endpoint` / `S3_ENDPOINT`.

examples/specdec_bench/specdec_bench/s3_utils.py
ChenhanYu: Already mitigated in 0b759a29a4 (same s3_utils merge): `_discover_runs` does `base = f"{s3_prefix_base}/{local_root.name}".lstrip("/")` so the leading `/` is stripped before any S3 key is constructed.

Added a `test_empty_prefix_no_leading_slash` regression in de31ccad07 that pins this — `parse_s3_path("s3://bucket")` returns `("bucket", "")` and the resulting S3 keys are `myrun/...`, never `/myrun/...`. Also covered: `s3://bucket/`, `s3://bucket/prefix/` (trailing-slash normalization), and the equivalent through the full `_discover_runs` flat-and-sweep path.

Verified via `python -m pytest tests/test_upload_to_s3.py::TestDiscoverRuns -v`.

examples/specdec_bench/upload_to_s3.py
ChenhanYu: Fixed in de31ccad07.

```python
except Exception as exc:
    ...
    print(f"  Error: {type(exc).__name__}: {exc}")
    traceback.print_exc(file=sys.stderr)   # ← new
    errors += 1
```

Boto3 `ClientError` detail (status code, error code, request id) is now recoverable from the sweep log.

tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench_mtp.yaml
ChenhanYu: Partially addressed in d48b30e2ea: the `JIRA_TICKET` / `HUGGINGFACE_MODEL_ID` *attestation* section in `examples/specdec_bench/README.md` was removed — that flow lives in pensieve-intern (which sets the env vars from team secrets / Hub-resolution checks before launching `run.py`), so surfacing them as user-facing attestation knobs in the public README was actively misleading. `dump_env()` still reads them — they default to `null` for general users, which is the correct behavior.

On the 11-line YAML comment block: keeping it for now as the inline documentation for the env-var contract until Phase 2 launcher auto-injection lands and the block becomes actionable (or removable). Will trim to a 1-line pointer + move details to `tools/launcher/common/specdec_bench/README.md` as part of the Phase 2 PR — tracked on [OMNIML-4788](https://jirasw.nvidia.com/browse/OMNIML-4788).

ChenhanYu:
## Round 3 — addressed @cjluo-nv review

Pushed in `de31ccad07` (+ `d48b30e2ea` for the README cleanup). Replied to all 9 inline threads with per-thread detail; the cross-cutting summary:

**Code fixes**
- **`AR→AL` string consistency** (your Item 2). Rich column header in `SpecBench`, plus `MTBench` / `AcceptanceRate` print strings, now all say `Average AL` to match the underlying `*_AL` dict keys.
- **`uid` timestamp bug** (your inline on `utils.py:314`). `config["timestamp"]` removed from `uid_parts`; `uid` is now the per-config fingerprint the visualizer needs for grouping repeats. Comment rewritten.
- **`traceback.format_exc()` in upload broad-except** (your inline on `upload_to_s3.py:217`). Boto3 `ClientError` detail now recoverable from sweep logs.
- **Ruff format / RUF059** — clears the `code-quality` CI failure.

**Items already addressed in `0b759a29a4` before your review landed** (you commented on the pre-merge `s3_utils.py` — that file no longer exists):
- `S3_DEFAULT_ENDPOINT` NVIDIA-internal hardcoding — defaults are now `""`, no team-specific infrastructure in this Apache-2.0 file.
- `parse_s3_path` empty-prefix → leading `/` — mitigated in `_discover_runs` via `.lstrip("/")`. Added a `test_empty_prefix_no_leading_slash` regression in `de31ccad07` to pin this.

**Tests** (your Item 3, partial-scope per discussion). Added 41 tests under `examples/specdec_bench/tests/` covering the security-adjacent paths:
- `_redact_value` — nested-dict + list redaction, allowlist
- `_redact_argv` — long-flag, `--flag=val`, prev-token carry, allowlist
- `_is_sensitive_key` — sensitive substrings + allowlist
- `parse_s3_path` — edge inputs (your (f))
- `_discover_runs` — flat vs sweep (your (e)) + empty-prefix regression

Deferred to a follow-up: `dump_env` end-to-end and `_checkpoint_provenance` (filesystem + subprocess fixtures, non-trivial to factor cleanly; redaction is the load-bearing security surface).

```bash
cd examples/specdec_bench && python -m pytest tests/ -v
# 41 passed in 1.48s
```

**Pushing back on your Item 1 (split the PR)**: the AR→AL rename and the provenance/upload work share the same downstream consumer (the visualizer + the S3 bucket schema), and the visualizer's `Joint_AR` ingest is part of the same Phase 1a deliverable on [OMNIML-4788](https://jirasw.nvidia.com/browse/OMNIML-4788). Splitting would force a two-step migration on the same artifact format. The honest fix is to make the breaking change *visible*, so the PR body now leads with a ⚠️ breaking-change callout. Open to splitting if you'd still prefer — say the word.

**README updates** (your Item 1 sub-ask): the user-visible field names appear only in `__init__.py`'s version-bump note and the visualizer code (separate repo, tracked on OMNIML-4788). No external doc references `Request_AR` etc.

**Status**: `code-quality` CI failure should clear on this push (verified ruff clean locally). cc @kevalmorabia97 and the codeowner teams — ready for another look.

examples/specdec_bench/specdec_bench/__init__.py
kevalmorabia97: Can we not use separate version and instead use `modelopt.__version__` here?

tests/examples/specdec_bench/__init__.py
kevalmorabia97, line 1: can we move this to `tests/examples/specdec_bench`

examples/specdec_bench/tests/test_redaction.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class TestRedactValue:
+    def test_top_level_secret(self):
+        out = _redact_value({"hf_token": "abc", "model_dir": "/m"})
+        assert out == {"hf_token": REDACTED, "model_dir": "/m"}
+
+    def test_nested_secret_in_dict(self):
+        """Nested hf_token inside serving_config must be redacted — this is
+        cjluo's original IMPORTANT-Compatibility concern."""
```
kevalmorabia97:
```suggestion
        """Nested hf_token inside serving_config must be redacted."""
```

tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - --speculative_algorithm NONE
+      - --tp_size 1
+      - --ep_size 1
+      - --concurrency 1
+      - --num_requests 80
+      - --output_length 4096
+      - --aa_timing
+      - --show_progress
+      - --save_dir /scratchspace/specdec_bench
+      - --trust_remote_code
```
kevalmorabia97: Qwen3.5 should not require trust_remote_code

kevalmorabia97 (changes requested):

tests/examples/specdec_bench/__init__.py
kevalmorabia97: And we need to enable this example test in CICD in appropriate container: https://github.com/NVIDIA/Model-Optimizer/blob/chenhany/specdec-bench-phase-1a-provenance-and-s3-upload/.github/workflows/example_tests.yml

examples/specdec_bench/specdec_bench/__init__.py
ChenhanYu: Done in d2d6527f5d — `specdec_bench/__init__.py` now does `from modelopt import __version__`. configuration.json's `specdec_bench_version` field will track the parent package; the visualizer-aggregation rationale moves to the per-run `specdec_bench_sha` fingerprint instead of a forked semver.

The `*_AL` rename + `Joint_AR` addition still ship in this PR (the breaking-change callout in the PR body still applies); we just no longer mark it with a standalone `1.0.0` bump.

tests/examples/specdec_bench/__init__.py
ChenhanYu: Done in d2d6527f5d — moved with `git mv` to `tests/examples/specdec_bench/`. conftest.py + test_upload_to_s3.py now anchor `_PKG_ROOT` on the repo root via `parents[3] / "examples" / "specdec_bench"`. Verified the path math resolves cleanly and `from specdec_bench.utils import …` still works.

examples/specdec_bench/tests/test_redaction.py
ChenhanYu: Applied in d2d6527f5d.

tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench.yaml
ChenhanYu: Done in d2d6527f5d. Also dropped from `specdec_bench_mtp.yaml` (same model, same reason) for consistency.

tests/examples/specdec_bench/__init__.py
ChenhanYu: Workflow change is staged locally as a separate commit (`4cacc1d08d`) but I can't push it via my PAT — GitHub rejects the push with "refusing to allow a Personal Access Token to create or update workflow `.github/workflows/example_tests.yml` without `workflow` scope". I'll push that one separately under my user creds and update this thread when it lands.

The commit adds `specdec_bench` to the existing `torch` matrix:
```yaml
example: [llm_distill, llm_qat, llm_sparsity, diffusers_sparsity, specdec_bench]
```
Uses the default `nvcr.io/nvidia/pytorch:26.03-py3` container with `[hf,dev-test]` extras (same as the other torch examples). Tests are CPU-only (~1-2s wall) so the GPU runner is overkill — happy to split to a non-GPU job in a follow-up if you'd prefer.

ChenhanYu:
## Round 4 — addressed @kevalmorabia97 review

Pushed in `d2d6527f5d` (code/tests) + local commit `4cacc1d08d` (CI workflow — needs a separate push, my PAT lacks `workflow` scope).

| # | Ask | Status |
|---|---|---|
| 1 | use `modelopt.__version__` | ✅ `d2d6527f5d` — `from modelopt import __version__` in specdec_bench/__init__.py |
| 2 | move tests to `tests/examples/specdec_bench/` | ✅ `d2d6527f5d` — `git mv` preserved history; conftest path math anchored on `parents[3]` |
| 3 | drop cjluo reference from test docstring | ✅ `d2d6527f5d` — applied your suggestion-block verbatim |
| 4 | drop `--trust_remote_code` from Qwen3.5-4B yaml | ✅ `d2d6527f5d` — also dropped from `specdec_bench_mtp.yaml` for consistency |
| 5 | wire into `example_tests.yml` matrix | ⏳ `4cacc1d08d` staged locally — needs a separate push under credentials with `workflow` scope |

CI re-run will fire on `d2d6527f5d`; once `4cacc1d08d` lands the `specdec_bench` matrix entry should pick up on the next PR push.

Re-requesting review.

tests/examples/specdec_bench/__init__.py
ChenhanYu: Pushed via SSH after SAML SSO auth — `4cacc1d08d` now on the PR; `example_tests.yml` matrix now includes `specdec_bench`. The matrix entry will fire on the next pr-gate cycle.

.github/workflows/example_tests.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        example: [llm_distill, llm_qat, llm_sparsity, diffusers_sparsity]
+        example: [llm_distill, llm_qat, llm_sparsity, diffusers_sparsity, specdec_bench]
```
kevalmorabia97, line 39: CI test only installs `requirements.txt` but we have `requirements_speed.txt` which will not be installed hence test may either be skipped or fail with missing package.
https://github.com/NVIDIA/Model-Optimizer/blob/4cacc1d08d028736c662f58bb141baa7178caab5/.github/workflows/_example_tests_runner.yml#L58

kevalmorabia97 (approved):

.github/workflows/example_tests.yml
ChenhanYu: Good catch — fixed in `3997903717`.

Renamed `requirements_speed.txt` → `requirements.txt` so the runner's `find examples/<example> -name requirements.txt` picks it up. Updated the 3 references (`README.md` install instructions, `ImportError` message in `datasets/speed.py`, `tools/launcher/common/specdec_bench/run.sh` install command). Matches the naming convention every other example in `examples/` uses; no workflow-side change needed.

Verified `git grep requirements_speed` returns nothing post-rename.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,250 @@ def postprocess_gptoss(text):
     if "<|channel|>" in final_message:
         final_message = final_message.split("<|channel|>")[0]
     return final_message
+
+
+def _get_engine_version(engine):
```
h-guo18, line 79: nit: why not do this also in `get_serving_config()`?

ChenhanYu: Want to make sure I'm reading the nit right — two interpretations:

1. **"add the engine version to `get_serving_config()`'s returned dict"** — `engine_version` is already captured at the top of `configuration.json` via `dump_env` (line 269: `config["engine_version"] = _get_engine_version(config.get("engine"))`), so adding it inside `serving_config` would just nest the same value. Happy to nest it too if useful for downstream consumers, but it's not adding new info.

2. **"refactor `_get_engine_version` into a per-class method on `Model` subclasses, mirroring `get_serving_config`"** — doable, would look like `VLLMModel.get_engine_version() = vllm.__version__` etc. Asymmetry today: `get_serving_config()` *needs* per-instance state (live `AsyncEngineArgs`, `vllm_config.to_dict()`, etc.) so it has to be a method; `_get_engine_version()` is pure module-import — no instance state — so a free function with engine-name dispatch felt simpler. If you'd prefer the uniform per-class interface, I'll do the refactor in a follow-up.

Which one did you mean? Leaning toward "no change needed" since the version is already in `configuration.json` at the top level, but happy to do either.

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,250 @@ def postprocess_gptoss(text):
...
+    # would make every run unique by construction and defeat the grouping.
+    uid_parts = "|".join(
+        [
+            str(getattr(args, "model_dir", "") or ""),
+            str(getattr(args, "speculative_algorithm", "") or ""),
+            str(getattr(args, "concurrency", "") or ""),
+            str(config.get("specdec_bench_sha") or ""),
+        ]
+    )
+    config["uid"] = hashlib.sha256(uid_parts.encode()).hexdigest()[:16]
```
h-guo18: There are some scenarios that this SHA cannot distinguish multiple runs. E.g.: 
- Compare 2 different drafters of the same algo, on the same target model
- For a same drafter, when we want to compare different setting (draft_steps, temprature, osl, etc)
  
Seems like with current algorithm they will be considreed repetition of the same run.

examples/specdec_bench/upload_to_s3.py
```diff
@@ -0,0 +1,237 @@
...
+_RUN_SENTINELS = ("configuration.json", "timing.json", "acceptance_rate.json")
+
+
+# ── S3 helpers ────────────────────────────────────────────────────────────────
+# Inlined from a former specdec_bench/s3_utils.py so this stays a drop-in
+# single-file tool: a contributor can `wget upload_to_s3.py` and run it without
+# installing the specdec_bench package. Defaults are empty — endpoint, key id,
+# and secret are taken from --endpoint / --key-id / --secret (or the
+# corresponding S3_ENDPOINT / S3_KEY_ID / S3_SECRET env vars). No team-specific
+# infrastructure leaks into this public example.
```
h-guo18: Looks like there are some review history details in this comment. Do we want to keep them in code?

examples/specdec_bench/specdec_bench/utils.py
```diff
@@ -58,3 +74,250 @@ def postprocess_gptoss(text):
...
+    config["modelopt_version"] = os.environ.get("MODELOPT_VERSION") or _get_modelopt_version()
+    # Fallback assumes the in-tree layout examples/specdec_bench/specdec_bench/.
+    # parents[2] reaches the modelopt repo root in that case. When vendored
+    # elsewhere this would `git rev-parse` an unrelated repo; rely on the env
+    # var path instead for non-in-tree deployments.
+    config["modelopt_sha"] = os.environ.get("MODELOPT_SHA") or _git_sha(
+        specdec_bench_dir.parents[2]
+    )
+    config["nmm_sandbox_sha"] = os.environ.get("NMM_SANDBOX_SHA") or None
+    config["container_image"] = os.environ.get("CONTAINER_IMAGE") or None
```
h-guo18, line 292: There are some perf-relevant info (torch version, cuda driver/runtime version, nccl) that is only available in `CONTAINER_IMAGE`. If we want strictly reproduciblity, I think we should enforce there's a CONTAINER_IMAGE provided. 

I think we can check it during upload_to_s3.py instead of here.

examples/specdec_bench/specdec_bench/utils.py
ChenhanYu: Real bug, you're right — fixed in e280082082.

`uid_parts` now includes the full configuration-identity set, so two runs that differ in drafter, engine, draft_length, output_length, tp/ep size, or runtime_params (which is where temperature lives) hash to different uids:

```python
uid_parts = "|".join([
    str(config.get("model_dir") or ""),
    str(config.get("draft_model_dir") or ""),
    str(config.get("engine") or ""),
    str(config.get("speculative_algorithm") or ""),
    str(config.get("draft_length") or ""),
    str(config.get("output_length") or ""),
    str(config.get("concurrency") or ""),
    str(config.get("tp_size") or ""),
    str(config.get("ep_size") or ""),
    str(config.get("specdec_bench_sha") or ""),
    json.dumps(config.get("runtime_params") or {}, sort_keys=True, default=str),
])
```

Reading from `config` (the redacted view of args) rather than raw args means secret-only differences (e.g. two `--hf_token` values) collapse to one uid — which is correct, since they're logically the same configuration.

examples/specdec_bench/upload_to_s3.py
ChenhanYu: Good catch — trimmed in e280082082. The 'Inlined from a former s3_utils.py' and 'No team-specific infrastructure leaks' lines were referencing earlier review-round decisions, not anything load-bearing for the file itself. New comment is just the operational facts:

```python
# ── S3 helpers ────────────────────────────────────────────────────────────────
# Endpoint, key id, and secret default to empty and are taken from --endpoint /
# --key-id / --secret (or the corresponding S3_ENDPOINT / S3_KEY_ID / S3_SECRET
# env vars).
```

examples/specdec_bench/specdec_bench/utils.py
ChenhanYu: Done in e280082082 — moved the enforcement to `upload_to_s3.py` as you suggested.

New `_check_provenance(run_dir)` helper reads each run's `configuration.json` and returns the list of missing required provenance fields. `_REQUIRED_PROVENANCE_FIELDS = ("container_image",)` is the module-level constant — easy to extend if we add more required fields later (e.g. CUDA driver version, NCCL version once we can capture them).

`main()` runs the check after `_discover_runs` and rejects by default with a clear message:
```
Error: 3 run(s) are missing required provenance fields.
Set CONTAINER_IMAGE before running, or pass --allow-incomplete-provenance
to upload anyway.
```

For ad-hoc / dev uploads where you intentionally don't care, the new `--allow-incomplete-provenance` flag lets you bypass with a warning. 5 unit tests under `tests/examples/specdec_bench/test_upload_to_s3.py::TestCheckProvenance` cover the four paths (complete / null / missing config.json / malformed json / empty string).

Left `dump_env` writing `container_image` as `os.environ.get("CONTAINER_IMAGE") or None` unchanged — failing at config-write time would block standalone runs that don't care about S3 upload, which is the wrong UX for the dev path.

examples/specdec_bench/specdec_bench/utils.py
h-guo18: Nevermind. I was thinking engine version should be part of serving_config. It's also fine this way.

ChenhanYu: Got it, thanks — leaving as-is for this PR. Easy to nest into serving_config in a Phase-1b/Phase-2 follow-up if it turns out the visualizer prefers it that way.

examples/specdec_bench/specdec_bench/utils.py
h-guo18: Actually I think there might be still some issue after the fix:
- Dataset info is not in it, so mtbench and speedbench may produce the same sha

Instead of baking the grouping logic in specdecbench, would it be cleaner if: specdec bench save every run it runs, and we leave the grouping logics of runs entirely to the visualizer?

ChenhanYu: Good call — agreed, dropped uid entirely in `64fc7c49c3`.

You're right that baking a grouping opinion into specdec_bench is the wrong layering: the visualizer (Phase 4 of OMNIML-4788) is downstream of this data and should own grouping semantics — it can hash whatever subset of fields makes sense for any given comparison view (per-model, per-algo, per-dataset, ignore-concurrency, etc.).

Also worth noting: uid wasn't actually consumed by anything in this PR — `upload_to_s3.py` dedups by S3 prefix, not by uid — so removing it just deletes dead weight. The raw configuration fields stay in `configuration.json`; the visualizer composes them however it wants.

Bonus: this kills the entire class of "what should be in uid?" questions in one go (cjluo's timestamp-in-uid concern → my draft_model_dir/draft_length/etc. expansion → your dataset addition → ...).

h-guo18 (approved):
LGTM

ChenhanYu:
/ok to test [64fc7c4](https://github.com/NVIDIA/Model-Optimizer/pull/1531/commits/64fc7c49c3f28e95340d0e154b25a8f1fcaeafc1)

ChenhanYu:
Rebased onto current main (which now includes #1559's NGC auth fix). New head: `aa046b9bc2`. 19 commits replayed cleanly, no conflicts.

Posting `/ok to test` above to trigger the NVIDIA-runner workflows on the new SHA — expecting the `megatron / run-test` job to clear this time since `_example_tests_runner.yml` now authenticates against `nvcr.io` via the `NGC_API_KEY` secret.

Approvals on prior SHAs are now stale post-rebase:
- @kevalmorabia97 — last approved `4cacc1d08d`
- @h-guo18 — last approved `64fc7c49c3`

If CI lands green, a re-approval click on `aa046b9bc2` from either of you (Keval is enough to clear the `modelopt-setup-codeowners` team gate) will unblock the merge.

PR #1532, opened 2026-05-22
fix: preserve inlined MTP layers for GLM5

Fridah-nv:
### What does this PR do?

Type of change: Bug fix  

Extends `load_mtp_weights` to detect *inlined* MTP layers — keys `model.layers.{i}.*` for `i in [num_hidden, num_hidden +  num_nextn_predict_layers)` — in addition to the existing `mtp.*` separate-file convention.

  **Bug.** `load_mtp_weights()` only matched the substring `"mtp"` in safetensors keys. GLM-5.1 (`GlmMoeDsaForCausalLM`) stores MTP at `model.layers.78.*` with no `mtp` substring, so detection returned `([], {})`, `_mtp_layer_prefixes` was never set, and MTP tensors were silently dropped from the exported safetensors (had to be re-added manually).

**Detection.**

  1. **Detect** via `config.num_nextn_predict_layers` (the model's own declaration of how many MTP layers exist). 
  2. **Compute** the inlined layer indices: `model.layers.{i}` for `i in range(num_hidden, num_hidden + num_nextn)`.
  3. **Load** matching tensors from the on-disk shards via `safe_open` (walks `model.safetensors.index.json` if present,else falls back to the single shard).
  4. **Split** the loaded tensors by whether `model.state_dict()` has a slot for them:
     - keys present in `model.state_dict()` → `model.load_state_dict(..., strict=False)` (DeepSeek-V3 case: HF instantiates the extra layers).
     - keys absent from `model.state_dict()` → returned as `not_in_state_dict` so the exporter routes them through `extra_state_dict` (GLM-5.1
  case: `GlmMoeDsaModel` in transformers ≥5.7 only builds `num_hidden` decoders, leaving MTP keys orphaned at `from_pretrained` time).
 The returned prefixes flow into the existing plumbing — `_mtp_layer_prefixes` → `quant_cfg` disable +`quantization_config.exclude_modules` — unchanged.

### Usage

```python
```

### Testing
 Verified end-to-end on a mini GLM-5.1 fixture (4 hidden layers + 1 inlined MTP at `model.layers.4`, 7 synthesized MTP tensors mirroring the full GLM-5.1 layout)
To be verified with full model

## Summary by CodeRabbit

* **Improvements**
  * Quantization utilities now stream safetensors and unify loading of multi-token-prediction (MTP) weights from both inline and separate/sharded conventions, reporting detected MTP prefixes and counts of loaded vs orphaned tensors.
* **Tests**
  * Added unit tests and a test import helper covering MTP discovery, loading behaviors (inlined vs standalone/indexed shards), orphan reporting, and non‑MTP checkpoints.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1532?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/llm_ptq/example_utils.py
+++ b/examples/llm_ptq/example_utils.py
@@ -30,6 +30,7 @@ import torch
 import transformers
 from accelerate import infer_auto_device_map, init_empty_weights
 from accelerate.utils import get_max_memory
+from safetensors import safe_open
 from safetensors.torch import load_file
 from transformers import (
     AutoConfig,
@@ -315,98 +316,157 @@ def get_processor(
             return None
 
 
+def _load_inlined_mtp_tensors(model_path: Path, mtp_prefixes: set[str]) -> dict[str, torch.Tensor]:
+    """Stream tensors whose keys start with any ``{prefix}.`` from on-disk shards.
+
+    Walks ``model.safetensors.index.json`` when present, else falls back to a
+    single ``model.safetensors`` file. Returns an empty dict if no matching
+    keys are found.
+    """
+
+    def _matches(key: str) -> bool:
+        return any(key.startswith(p + ".") for p in mtp_prefixes)
+
+    tensors: dict[str, torch.Tensor] = {}
+    index_file = model_path / "model.safetensors.index.json"
+    if index_file.exists():
+        weight_map = json.load(open(index_file))["weight_map"]
+        per_shard: dict[str, list[str]] = {}
+        for key, shard in weight_map.items():
+            if _matches(key):
+                per_shard.setdefault(shard, []).append(key)
+        for shard, keys in per_shard.items():
+            with safe_open(str(model_path / shard), framework="pt", device="cpu") as f:
+                for k in keys:
+                    tensors[k] = f.get_tensor(k)
+    else:
+        single = model_path / "model.safetensors"
+        if single.exists():
+            with safe_open(str(single), framework="pt", device="cpu") as f:
+                # safe_open is not a dict; ``.keys()`` is the public listing API.
+                for k in f.keys():  # noqa: SIM118
+                    if _matches(k):
+                        tensors[k] = f.get_tensor(k)
+    return tensors
+
+
 def load_mtp_weights(
     model: torch.nn.Module, model_path: str
 ) -> tuple[list[str], dict[str, torch.Tensor]]:
     """Load MTP weights from the model checkpoint.
 
-    Some models store additional layers in separate safetensors files with non-standard
-    names (e.g., mtp.safetensors). HuggingFace's from_pretrained() may not load these
-    files even though they're referenced in model.safetensors.index.json.
+    Detects MTP layers under two on-disk conventions:
 
-    This function detects such cases and explicitly loads the missing weights.
+    1. Separate-file: weights live in a non-standard safetensors file (e.g.,
+       ``mtp.safetensors``) with keys prefixed by ``mtp``.
+    2. Inlined: weights live in the main shards under
+       ``model.layers[num_hidden : num_hidden + num_nextn_predict_layers]``
+       (DeepSeek-V3, GLM-5.1 ``GlmMoeDsa``, GLM-4.7).
+
+    Whether the HF model class actually instantiates the MTP module varies:
+    DeepSeek-V3's modeling code adds the extra decoder layers, while
+    ``GlmMoeDsaModel`` in transformers >=5.7 builds only ``num_hidden`` layers
+    and leaves the inlined-MTP keys orphaned. To keep both paths correct, we
+    always read the tensors off disk here and split them: any key that maps
+    to a parameter in ``model.state_dict()`` is loaded into the model;
+    the remainder is returned as ``not_in_state_dict`` for the exporter to
+    merge in via ``extra_state_dict``.
 
     Args:
         model: The loaded model that may be missing weights
         model_path: Path to the model directory
 
     Returns:
-        List of layer prefixes that were loaded from non-standard safetensors files.
-        These layers should typically be excluded from quantization.
-        Empty list if no additional weights were loaded.
-        Dictionary of MTP weights that were not loaded into the model state dict.
+        List of layer prefixes that should be excluded from quantization
+        (e.g., ``"mtp.layers.0"`` or ``"model.layers.78"``). Empty if no MTP
+        layers were detected.
+        Dictionary of MTP weights that have no slot in the model's state
+        dict; ``export_hf_checkpoint`` merges these via ``extra_state_dict``.
     """
-    model_path = Path(model_path)
-    index_file = model_path / "model.safetensors.index.json"
-
-    if not index_file.exists():
-        return [], {}
-
-    # Load the index to find all referenced safetensors files
-    index = json.load(open(index_file))
-    weight_map = index["weight_map"]
-    # Find all files in weight_map whose key or value contains "mtp"
-    mtp_weight_map = {}
-    for k, v in weight_map.items():
-        if "mtp" in k or "mtp" in v:
-            mtp_weight_map.setdefault(v, []).append(k)
-
-    if not mtp_weight_map:
-        return [], {}
-
-    def _extract_layer_prefixes(keys):
-        mtp_layer_prefixes = set()
-        for key in keys:
-            parts = key.split(".")
-            # Capture the top-level MTP module prefix (e.g., "mtp" from "mtp.fc.weight")
-            # so that non-layer MTP weights like mtp.fc, mtp.norm are also excluded
-            if parts:
-                mtp_layer_prefixes.add(parts[0])
-            # Also capture specific layer prefixes (e.g., "mtp.layers.0")
-            for i, part in enumerate(parts):
-                if part == "layers" and i + 1 < len(parts) and parts[i + 1].isdigit():
-                    prefix = ".".join(parts[: i + 2])
-                    mtp_layer_prefixes.add(prefix)
-                    break
-
-        return mtp_layer_prefixes
-
-    # Flatten mtp_weight_map.values() (list of list of str) to a single list of str
-    mtp_keys = [k for keys in mtp_weight_map.values() for k in keys]
-    mtp_layer_prefixes = _extract_layer_prefixes(mtp_keys)
-
-    # Check which non-standard files exist and have missing weights
+    mtp_layer_prefixes: set[str] = set()
+    not_in_state_dict: dict[str, torch.Tensor] = {}
+    model_dir = Path(model_path)
     model_state = model.state_dict()
-    total_loaded = 0
 
-    not_in_state_dict = {}
+    # Inlined-MTP convention: keys ``model.layers.{i}.*`` for
+    # ``i in [num_hidden, num_hidden + num_nextn)``.
+    cfg = model.config
+    num_nextn = int(getattr(cfg, "num_nextn_predict_layers", 0))
+    num_hidden = cfg.num_hidden_layers
+    if num_nextn:
+        inlined_prefixes = {f"model.layers.{i}" for i in range(num_hidden, num_hidden + num_nextn)}
+        inlined_tensors = _load_inlined_mtp_tensors(model_dir, inlined_prefixes)
+        if inlined_tensors:
+            mtp_layer_prefixes |= inlined_prefixes
+            in_state_dict = {k: v for k, v in inlined_tensors.items() if k in model_state}
+            not_in_state_dict |= {k: v for k, v in inlined_tensors.items() if k not in model_state}
+            if in_state_dict:
+                model.load_state_dict(in_state_dict, strict=False)
+            print(
+                f"✓ Detected {len(inlined_tensors)} inlined MTP tensors under "
+                f"{sorted(inlined_prefixes)} "
+                f"(loaded into model: {len(in_state_dict)}, orphaned: {len(not_in_state_dict)})"
+            )
 
-    for filename, mtp_keys in mtp_weight_map.items():
-        filepath = model_path / filename
-        if not filepath.exists():
-            continue
+    index_file = model_dir / "model.safetensors.index.json"
 
-        print(f"Loading {len(mtp_keys)} mtp weights from {filename}...")
-        weights = load_file(str(filepath), device="cpu")
-        weights = {k: v for k, v in weights.items() if k in mtp_keys}
-        # Load the MTP weights to the model state dict
-        in_state_dict = {k: weights[k] for k in weights if k in model_state}
-        not_in_state_dict = not_in_state_dict | {
-            k: weights[k] for k in weights if k not in model_state
-        }
+    if index_file.exists():
+        # Separate-file MTP detection via safetensors index.
+        index = json.load(open(index_file))
+        weight_map = index["weight_map"]
+        # Find all files in weight_map whose key or value contains "mtp"
+        mtp_weight_map: dict[str, list[str]] = {}
+        for k, v in weight_map.items():
+            if "mtp" in k or "mtp" in v:
+                mtp_weight_map.setdefault(v, []).append(k)
 
-        if in_state_dict:
-            model.load_state_dict(in_state_dict, strict=False)
-            total_loaded += len(in_state_dict)
+        if mtp_weight_map:
 
-    if total_loaded > 0:
-        print(
-            f"✓ Successfully loaded {total_loaded} MTP weights, "
-            f"{len(not_in_state_dict)} MTP weights not in model.state_dict"
-        )
+            def _extract_layer_prefixes(keys):
+                prefixes = set()
+                for key in keys:
+                    parts = key.split(".")
+                    # Capture the top-level MTP module prefix (e.g., "mtp" from "mtp.fc.weight")
+                    # so that non-layer MTP weights like mtp.fc, mtp.norm are also excluded
+                    if parts:
+                        prefixes.add(parts[0])
+                    # Also capture specific layer prefixes (e.g., "mtp.layers.0")
+                    for i, part in enumerate(parts):
+                        if part == "layers" and i + 1 < len(parts) and parts[i + 1].isdigit():
+                            prefixes.add(".".join(parts[: i + 2]))
+                            break
+                return prefixes
+
+            mtp_keys_flat = [k for keys in mtp_weight_map.values() for k in keys]
+            mtp_layer_prefixes |= _extract_layer_prefixes(mtp_keys_flat)
+
+            # Load any weights missing from model.state_dict from the non-standard files.
+            total_loaded = 0
+            for filename, mtp_keys in mtp_weight_map.items():
+                filepath = model_dir / filename
+                if not filepath.exists():
+                    continue
+
+                print(f"Loading {len(mtp_keys)} mtp weights from {filename}...")
+                weights = load_file(str(filepath), device="cpu")
+                weights = {k: v for k, v in weights.items() if k in mtp_keys}
+                in_state_dict = {k: weights[k] for k in weights if k in model_state}
+                not_in_state_dict = not_in_state_dict | {
+                    k: weights[k] for k in weights if k not in model_state
+                }
+
+                if in_state_dict:
+                    model.load_state_dict(in_state_dict, strict=False)
+                    total_loaded += len(in_state_dict)
+
+            if total_loaded > 0:
+                print(
+                    f"✓ Successfully loaded {total_loaded} MTP weights, "
+                    f"{len(not_in_state_dict)} MTP weights not in model.state_dict"
+                )
 
     if mtp_layer_prefixes:
-        print(f"✓ Detected MTP layers to exclude from quantization: {mtp_layer_prefixes}")
+        print(f"✓ Detected MTP layers to exclude from quantization: {sorted(mtp_layer_prefixes)}")
 
     return list(mtp_layer_prefixes), not_in_state_dict
 
--- a/examples/llm_ptq/example_utils.py
+++ b/examples/llm_ptq/example_utils.py
@@ -23,6 +23,7 @@ import os
 import shutil
 import sys
 import warnings
+from collections.abc import Iterable
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
+    tensors: dict[str, torch.Tensor] = {}
+    for filename, mtp_keys in mtp_weight_map.items():
+        filepath = model_dir / filename
+        if not filepath.exists():
+            continue
+        print(f"Loading {len(mtp_keys)} mtp weights from {filename}...")
+        loaded = load_file(str(filepath), device="cpu")
+        for k in mtp_keys:
+            if k in loaded:
+                tensors[k] = loaded[k]
```
cjluo-nv: > _Bot comment._

This `See Also` points at `examples/llm_ptq/MTP_DETECTION.md`, but that file isn't in the PR and doesn't exist in the repo. Either add the design doc or drop the reference before merge — dangling doc pointers tend to outlive the person who wrote them.

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
+    if not num_nextn:
+        return []
+    num_hidden = config.num_hidden_layers
+    return [f"model.layers.{i}" for i in range(num_hidden, num_hidden + num_nextn)]
+
+
+def _scan_separate_file_mtp(
+    model_dir: Path,
+) -> tuple[set[str], dict[str, torch.Tensor]]:
+    """Disk → (prefixes, tensors) for the legacy ``mtp.*``-keyed convention.
```
cjluo-nv: > _Bot comment._

If a config sets `num_nextn_predict_layers = None` (some HF configs do this rather than omitting the field), `int(None)` raises `TypeError`. Consider `int(getattr(config, "num_nextn_predict_layers", 0) or 0)` for robustness.

tests/examples/llm_ptq/test_example_utils.py
```diff
@@ -0,0 +1,40 @@
...
+def test_get_inlined_mtp_prefixes_returns_expected_prefixes(num_nextn, num_hidden, expected):
+    """Pure config -> prefix list. Documents the inlined-MTP detection contract."""
+    cfg = SimpleNamespace(num_nextn_predict_layers=num_nextn, num_hidden_layers=num_hidden)
+    assert example_utils.get_inlined_mtp_prefixes(cfg) == expected
+
+
+def test_get_inlined_mtp_prefixes_missing_field_returns_empty():
+    """Configs without num_nextn_predict_layers (non-MTP architectures) yield []."""
+    cfg = SimpleNamespace(num_hidden_layers=32)
+    assert example_utils.get_inlined_mtp_prefixes(cfg) == []
```
cjluo-nv: > _Bot comment._

The PR body mentions a working mini-GLM-5.1 fixture with 7 synthesized MTP tensors. It would be very valuable to land that fixture as a unit test here — `_load_inlined_mtp_tensors` (index walk + single-file fallback), the state-dict-vs-orphan split in `_apply_to_model_state_dict`, and the integrated `load_mtp_weights` round-trip are all untested, and the silent-drop bug being fixed is exactly the kind of regression a fixture test would catch.

cjluo-nv (commented):

Bug fix is reasonable and well-structured: the existing `"mtp" in key` substring detection genuinely misses GLM-5.1's inlined `model.layers.{i}` convention, and splitting into `get_inlined_mtp_prefixes` + `_load_inlined_mtp_tensors` + `_scan_separate_file_mtp` + `_apply_to_model_state_dict` is cleaner than the prior monolithic function. The single-file `model.safetensors` fallback in `_load_inlined_mtp_tensors` is also a nice secondary improvement (legacy path only handled the sharded case).

Flagging for human sign-off because:
- **Test coverage is thin.** The only new test exercises the trivial 4-line `get_inlined_mtp_prefixes` (config → list of strings). The actual on-disk loader (`_load_inlined_mtp_tensors`), the orphan-vs-state-dict split (`_apply_to_model_state_dict`), and the integrated `load_mtp_weights` flow are untested. The PR body says the author verified on "a mini GLM-5.1 fixture (4 hidden layers + 1 inlined MTP at `model.layers.4`, 7 synthesized MTP tensors)" — that fixture would make a solid unit test for the loader path; it's odd not to include it. The bug being fixed (silent tensor drop) is exactly the kind of regression a fixture-based test would catch.
- **End-to-end verification is still pending.** Author explicitly says "To be verified with full model" — i.e. no full GLM-5.1 export run yet.
- **Stale docstring reference.** `load_mtp_weights` now points readers to `examples/llm_ptq/MTP_DETECTION.md` ("See Also") but that file does not exist in the PR or the repo. Either add the doc or drop the reference before merge.
- Minor: `int(getattr(config, "num_nextn_predict_layers", 0))` will raise `TypeError` if the attribute is present but `None` (some HF configs do this); `getattr(..., 0) or 0` would be safer.

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
-        Empty list if no additional weights were loaded.
-        Dictionary of MTP weights that were not loaded into the model state dict.
+        List of layer prefixes that should be excluded from quantization
+        (e.g., ``"mtp.layers.0"`` or ``"model.layers.78"``). Empty if no MTP
+        layers were detected.
+        Dictionary of MTP weights that have no slot in the model's state
+        dict; ``export_hf_checkpoint`` merges these via ``extra_state_dict``.
+
+    See Also:
+        ``examples/llm_ptq/MTP_DETECTION.md`` for the design rationale and
```
Edwardf0t1: Does `examples/llm_ptq/MTP_DETECTION.md` exist?

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
+
+def get_inlined_mtp_prefixes(config: Any) -> list[str]:
+    """Pure: HF config → state-dict-key prefixes for inlined MTP layers.
+
+    Inlined-MTP convention (DeepSeek-V3, GLM-5.1 ``GlmMoeDsa``, GLM-4.7):
+    MTP tensors live under ``model.layers[i]`` for
+    ``i in [num_hidden, num_hidden + num_nextn_predict_layers)``. Returns
+    ``[]`` when the config does not declare any MTP layers.
+    """
+    num_nextn = int(getattr(config, "num_nextn_predict_layers", 0))
```
Edwardf0t1:
```suggestion
    num_nextn = int(getattr(config, "num_nextn_predict_layers", 0) or 0)
```

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
+    for k, v in weight_map.items():
+        if "mtp" in k or "mtp" in v:
+            mtp_weight_map.setdefault(v, []).append(k)
+    if not mtp_weight_map:
+        return set(), {}
+
+    prefixes: set[str] = set()
+    for key in (k for keys in mtp_weight_map.values() for k in keys):
+        parts = key.split(".")
+        # Top-level MTP module prefix (e.g. "mtp" from "mtp.fc.weight") so
```
cjluo-nv: > _Bot comment._

**Simplification: unify the two scanners.** `_load_inlined_mtp_tensors` and `_scan_separate_file_mtp` both walk `model.safetensors.index.json` and stream tensors whose keys match a predicate — they only differ in (a) the predicate (`startswith(prefix_tuple)` vs. `"mtp" in key`) and (b) the I/O API (`safe_open` vs. `load_file`, the latter loading the whole shard into RAM and defeating the point of streaming).

Consider collapsing both into one helper:

```python
def _load_tensors_matching(model_dir: Path, matches: Callable[[str], bool]) -> dict[str, torch.Tensor]:
    """Stream tensors whose keys satisfy `matches` from sharded or single safetensors."""
    index = model_dir / "model.safetensors.index.json"
    if index.exists():
        weight_map = json.loads(index.read_text())["weight_map"]
        per_shard: dict[str, list[str]] = {}
        for k, shard in weight_map.items():
            if matches(k):
                per_shard.setdefault(shard, []).append(k)
        shards = per_shard.items()
    else:
        single = model_dir / "model.safetensors"
        if not single.exists():
            return {}
        with safe_open(str(single), framework="pt", device="cpu") as f:
            return {k: f.get_tensor(k) for k in f.keys() if matches(k)}  # noqa: SIM118
    out: dict[str, torch.Tensor] = {}
    for shard, keys in shards:
        with safe_open(str(model_dir / shard), framework="pt", device="cpu") as f:
            for k in keys:
                out[k] = f.get_tensor(k)
    return out
```

Then:
- Inlined: `_load_tensors_matching(model_dir, lambda k: k.startswith(prefix_tuple))`.
- Separate-file: `_load_tensors_matching(model_dir, lambda k: "mtp" in k)`.

This also fixes the no-index regression CodeRabbit flagged at line 384 for free — the legacy unsharded `mtp.safetensors` (or any single-file checkpoint with `mtp.*` keys) now goes through the same fallback.

Also: `safe_open` is preferable to `load_file` on the separate-file path too — `load_file` reads the whole shard into RAM regardless of how few `mtp.*` keys you actually want.

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
+def _scan_separate_file_mtp(
+    model_dir: Path,
+) -> tuple[set[str], dict[str, torch.Tensor]]:
+    """Disk → (prefixes, tensors) for the legacy ``mtp.*``-keyed convention.
+
+    Walks ``model.safetensors.index.json`` for keys/values containing
+    ``"mtp"``, loads the matching tensors from their referenced shards, and
+    returns both the prefixes (e.g. ``"mtp"``, ``"mtp.layers.0"``) and the
+    raw tensors. Returns empty when no index file exists or no MTP keys
+    are present.
```
cjluo-nv: > _Bot comment._

**Simplification: extract prefix derivation as a pure helper, mirroring `get_inlined_mtp_prefixes`.** Right now `_scan_separate_file_mtp` mixes two concerns — "derive prefixes from keys" (lines 379–390) and "load matching tensors" (lines 392–401). The inlined branch already cleanly separates these (`get_inlined_mtp_prefixes` is pure config→prefixes; `_load_inlined_mtp_tensors` is pure I/O). Applying the same split here:

```python
def _keys_to_mtp_prefixes(keys: Iterable[str]) -> set[str]:
    prefixes: set[str] = set()
    for key in keys:
        parts = key.split(".")
        if parts:
            prefixes.add(parts[0])
        for i, part in enumerate(parts):
            if part == "layers" and i + 1 < len(parts) and parts[i + 1].isdigit():
                prefixes.add(".".join(parts[: i + 2]))
                break
    return prefixes
```

Combined with the unified loader from the other comment, `load_mtp_weights` collapses to:

```python
tensors = _load_tensors_matching(model_dir, lambda k: "mtp" in k)
tensors |= _load_tensors_matching(model_dir, lambda k: k.startswith(inlined_prefix_tuple))
prefixes = _keys_to_mtp_prefixes(t for t in tensors if "mtp" in t) | set(inlined_prefixes_actually_seen)
```

Net: ~3 helpers, each with a single responsibility, instead of 4 with overlapping I/O. Pure helpers (`get_inlined_mtp_prefixes`, `_keys_to_mtp_prefixes`) become trivially unit-testable without fixtures.

examples/llm_ptq/example_utils.py
cjluo-nv: > _Bot comment._

Still vulnerable to `int(None)` raising `TypeError` if `num_nextn_predict_layers` is set to `None` (some HF configs do this rather than omitting the field). The earlier suggestion `int(getattr(config, "num_nextn_predict_layers", 0) or 0)` still applies — the `or 0` collapses both "missing" and "present-but-None" into the same non-MTP short-circuit.

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
-        Dictionary of MTP weights that were not loaded into the model state dict.
+        List of layer prefixes that should be excluded from quantization
+        (e.g., ``"mtp.layers.0"`` or ``"model.layers.78"``). Empty if no MTP
+        layers were detected.
+        Dictionary of MTP weights that have no slot in the model's state
+        dict; ``export_hf_checkpoint`` merges these via ``extra_state_dict``.
+
+    See Also:
+        ``examples/llm_ptq/MTP_DETECTION.md`` for the design rationale and
+        planned migration path to a library-side detector registry.
```
cjluo-nv: > _Bot comment._

Still pointing at `examples/llm_ptq/MTP_DETECTION.md`, which doesn't exist in this PR or the repo (also flagged by Edward and Claude in the previous round). Either land the doc in this PR or drop the `See Also` block — dangling doc references rot fast.

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +317,183 @@ def get_processor(
...
+
+    Walks ``model.safetensors.index.json`` when present, else falls back to a
+    single ``model.safetensors`` file. Returns an empty dict if no matching
+    keys are found.
+    """
+    prefix_tuple = tuple(p + "." for p in mtp_prefixes)
+
+    def _matches(key: str) -> bool:
+        return key.startswith(prefix_tuple)
+
```
cjluo-nv: > _Bot comment._

Two file-handle leaks here and at line 386 (`json.load(open(index_file))`) — CodeRabbit's suggestion to use `with open(index_file) as f: json.load(f)` is straightforward and worth picking up. If the unified `_load_tensors_matching` refactor lands, `json.loads(index.read_text())` is even simpler and removes both occurrences in one pass.

cjluo-nv (commented):

Per Slack request: posted simplification ideas. The inlined- and separate-file MTP paths duplicate the "walk index → stream matching tensors" logic with different I/O APIs and different fallback behavior, and the prefix-extraction concern is mixed into the separate-file scanner. Unifying both around a single `_load_tensors_matching(model_dir, predicate)` helper (using `safe_open` consistently, with single-file fallback) plus a pure `_keys_to_prefixes(keys)` extractor would shrink the diff meaningfully and close the no-index regression on the legacy path that CodeRabbit flagged. Also noting unresolved prior comments: stale `MTP_DETECTION.md` "See Also" pointer, `int(None)` TypeError if `num_nextn_predict_layers` is present-but-None, and the loader/orphan-split paths still have no fixture-based test (the PR body's mini-GLM-5.1 fixture would be a natural unit test).

tests/examples/llm_ptq/test_example_utils.py
cjluo-nv: > _Bot comment._

Test coverage is still limited to the 4-line pure `get_inlined_mtp_prefixes`. The actual bug being fixed (silent tensor drop for inlined MTP) lives in `_load_inlined_mtp_tensors` + `_apply_to_model_state_dict` + the integrated `load_mtp_weights` flow, none of which are exercised here. The PR body explicitly mentions a working mini-GLM-5.1 fixture (4 hidden layers + 1 inlined MTP at `model.layers.4`, 7 synthesized tensors) — landing that as a unit test would directly cover the regression and is exactly the kind of test that prevents future silent drops. Worth adding before merge given that end-to-end verification on a full model is also still pending per the PR description.

examples/llm_ptq/example_utils.py
Fridah-nv: No, just removed

examples/llm_ptq/example_utils.py
Fridah-nv: Done

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +316,127 @@ def get_processor(
...
+    Inlined ``N`` in ``[num_hidden, num_hidden + num_nextn_predict_layers)``;
+    may be orphaned at ``from_pretrained`` time if the HF class only builds
+    ``num_hidden`` decoders.
+
+    Returns ``(prefixes, not_in_state_dict)``: ``prefixes`` populates
+    ``quantization_config.exclude_modules``; ``not_in_state_dict`` is fed to
+    ``export_hf_checkpoint(extra_state_dict=...)``.
+    """
+    model_dir = Path(model_path)
```
cjluo-nv, line 416: > _Bot comment._

Simplification: `key.startswith(())` already returns `False`, so the `if inlined_tuple and ...` guard is redundant — `if key.startswith(inlined_tuple): return True` works whether or not `inlined_tuple` is empty.

Also: the `shard_name` parameter on the predicate looks dead in practice. Every architecture in the docstring's support matrix is already detected via `"mtp" in key` or the inlined-prefix branch; I couldn't construct a real case where the `"mtp" in shard_name` clause is the only thing that matches. Consider dropping the second predicate arg to simplify the helper signature, or add a comment naming the convention it defends against.

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +316,127 @@ def get_processor(
...
-            model.load_state_dict(in_state_dict, strict=False)
-            total_loaded += len(in_state_dict)
+    # Combined predicate covering both conventions in one pass.
+    def predicate(key: str, shard_name: str | None) -> bool:
+        if inlined_tuple and key.startswith(inlined_tuple):
+            return True
+        return "mtp" in key or (shard_name is not None and "mtp" in shard_name)
 
-    if total_loaded > 0:
+    tensors = _load_tensors_matching(model_dir, predicate)
```
cjluo-nv, line 424: > _Bot comment._

Readability: `A | B if tensors else set()` parses correctly (`|` binds tighter than the ternary) but is easy to misread. An early return is cleaner and lets you drop the `if prefixes:` guard on the print below:

```python
if not tensors:
    return [], {}
separate_keys = [k for k in tensors if not k.startswith(inlined_tuple)]
prefixes = inlined_prefixes | _keys_to_prefixes(separate_keys)
```

examples/llm_ptq/example_utils.py
```diff
@@ -315,100 +316,127 @@ def get_processor(
...
+        for i, part in enumerate(parts):
+            if part == "layers" and i + 1 < len(parts) and parts[i + 1].isdigit():
+                prefixes.add(".".join(parts[: i + 2]))
+                break
+    return prefixes
+
+
+def _load_tensors_matching(
+    model_dir: Path, predicate: Callable[[str, str | None], bool]
+) -> dict[str, torch.Tensor]:
```
cjluo-nv, line 349: > _Bot comment._

Minor: `_keys_to_prefixes` is only correct for separate-file keys — if fed an inlined key like `"model.layers.78.eh_proj.weight"` it would emit `"model"` as a top-level prefix, which is not what the exporter wants. The current caller filters via `separate_keys = [k for k in tensors if not k.startswith(inlined_tuple)]` so this is safe today; worth a one-line note in the docstring ("caller must filter inlined keys first") to prevent future misuse.

cjluo-nv (commented):

Re-review pass — the previous round of simplification work has all landed and is in good shape:

- :speech_balloon: Author replied "Done" on the unified scanner — `_load_tensors_matching` + `_keys_to_prefixes` cleanly separate I/O from prefix derivation, and the `safe_open` migration kills the legacy `load_file` whole-shard read.
- :speech_balloon: Author replied "No, just removed" on `MTP_DETECTION.md` — confirmed gone from the docstring.
- :speech_balloon: `int(None)` fix (`or 0`) applied; file handles wrapped in `with`; no-index fallback restored via `model_dir.glob("*.safetensors")`.
- :speech_balloon: Test coverage substantially expanded — all four conventions (inlined-orphaned, inlined-in-state-dict, separate-standalone, separate-indexed) plus the `num_nextn_predict_layers=None` regression now have fixture tests. This addresses the prior "test coverage is thin" concern.

Flagging for human sign-off only because of the additional simplifications the operator asked about — none are correctness issues, all are nits:

1. **`predicate(key, shard_name)` second arg looks dead.** The only consumer of `shard_name` is `"mtp" in shard_name` in `load_mtp_weights.predicate`. For every real MTP convention enumerated in the docstring, the `"mtp" in key` branch already matches (Qwen3-Next, GLM-4.7) or the inlined-prefix branch matches (GLM-5.1, DeepSeek-V3). It's hard to construct a checkpoint where MTP weights would only be detected via shard filename. Consider dropping the `shard_name` parameter from the predicate signature — `_load_tensors_matching` becomes `Callable[[str], bool]` and the predicate body collapses to two clauses. If you want to keep the defensive shard-name fallback, document the case it covers; right now it reads as belt-and-suspenders.

2. **Redundant `if inlined_tuple` guard in `predicate`.** `key.startswith(())` already returns `False`, so `if inlined_tuple and key.startswith(inlined_tuple)` simplifies to `key.startswith(inlined_tuple)`.

3. **Awkward conditional + operator-precedence reading hazard on the prefixes line.** `prefixes = inlined_prefixes | _keys_to_prefixes(separate_keys) if tensors else set()` parses as `(A | B) if tensors else set()` (correct), but the `|`-vs-ternary precedence is non-obvious. Cleaner as an early return:
   ```python
   if not tensors:
       return [], {}
   prefixes = inlined_prefixes | _keys_to_prefixes(separate_keys)
   ```
   This also drops the now-unreachable `if prefixes:` guard around the print.

4. **Minor:** `_keys_to_prefixes` is only safe for separate-file keys (it would emit `"model"` as a prefix if fed an inlined key like `"model.layers.78.eh_proj"`). The current call site filters via `separate_keys = [k for k in tensors if not k.startswith(inlined_tuple)]` so it's fine, but a one-line docstring note ("caller must filter out inlined keys") would prevent future misuse.

5. **Docstring style:** `get_inlined_mtp_prefixes` and `_keys_to_prefixes` open with lowercase verbs ("turn an HF config…", "invert separate-file MTP keys…") — convention is capitalized imperative.

None of these block correctness; all are tighten-up suggestions per the operator's "additional simplification" ask.

cjluo-nv (approved):

Re-review: previous round's findings are addressed.

- :speech_balloon: File-handle leaks at lines 333/415 — both now use `with open(...) as f:` context managers.
- :speech_balloon: No-index fallback (CodeRabbit major) — `_load_tensors_matching` now globs `*.safetensors` independently of the index, restoring the standalone-shard path.
- :speech_balloon: Author replied "No, just removed" — `MTP_DETECTION.md` "See Also" reference is gone.
- :speech_balloon: `int(None)` regression — fixed via `or 0`, plus an explicit `num_nextn_predict_layers=None` test.
- :speech_balloon: Author replied "Done" on unifying the two scanners — `_load_tensors_matching` + `_keys_to_prefixes` cleanly split I/O from prefix derivation; `safe_open` replaces `load_file`.
- :speech_balloon: Test coverage substantially expanded — 5 fixture-based tests across all four conventions (inlined-orphaned, inlined-in-state-dict, separate-standalone, separate-indexed) plus the None regression. The "thin coverage" concern is resolved.
- The prior round's tighten-up nits (dead `shard_name` predicate arg, redundant `if inlined_tuple` guard, ternary precedence on the prefixes line, caller-must-filter docstring note for `_keys_to_prefixes`) are all reflected in the current code. Only remaining items are docstring-capitalization nits, which are non-blocking.

End-to-end full-model verification is still pending per the PR body, but that's an operational gate the author has flagged explicitly and not something a unit-test review can substitute for.

Edwardf0t1 (approved):
LGTM

PR #1536, opened 2026-05-22
Refine static NVFP4 MSE calibration

realAsma:
### What does this PR do?

Type of change: Bug fix

Refines static NVFP4 MSE calibration and forces static NVFP4 amax state to stay FP32 across calibration loading, quantizer promotion, dtype casts, and restore paths.

Main changes:

- Tighten max/MSE calibration bootstrap and static NVFP4 quantizer promotion.
- Keep static NVFP4 `_amax` and `_global_amax` in FP32.
- Update focused GPU/unit coverage for FP8 sweep calibration, promotion, restore, and FP32 amax preservation.

### Usage

```yaml
algorithm:
  method: mse
  fp8_scale_sweep: true
```

### Testing

```bash
pre-commit run --files modelopt/torch/quantization/nn/modules/tensor_quantizer.py tests/gpu/torch/quantization/test_nvfp4_static_quantizer_cuda.py
pytest_pwd tests/gpu/torch/quantization/test_nvfp4_static_quantizer_cuda.py -q
```

- Is this change backward compatible?: Yes
- Did you write any new necessary tests?: Yes
- Did you get Claude approval on this PR?: Yes

### Additional Information

N/A

### DIFF
--- a/modelopt/torch/quantization/calib/mse.py
+++ b/modelopt/torch/quantization/calib/mse.py
@@ -194,7 +194,7 @@ class NVFP4MSECalibrator(MseCalibrator):
         super().__init__(amax=amax, axis=axis, quant_func=quant_func, error_func=error_func)
         self._global_amax = global_amax.to(dtype=torch.float32)
         # Set by collect() after either sweep path; consumed by compute_amax.
-        self._best_amax_fast: torch.Tensor | None = None
+        self._best_amax: torch.Tensor | None = None
 
     def _compute_candidate_amax(self, candidates: torch.Tensor) -> torch.Tensor:
         if candidates.ndim != 0:  # Called during final compute amax
@@ -237,7 +237,7 @@ class NVFP4MSECalibrator(MseCalibrator):
     @torch.no_grad()
     def collect(self, x: torch.Tensor):
         """Collect input statistics and cache the final per-block amax."""
-        if self._best_amax_fast is not None:
+        if self._best_amax is not None:
             raise RuntimeError(
                 "NVFP4MSECalibrator: a previous collect() call produced a final amax; "
                 "multi-collect is not supported. Call reset() to start a fresh cycle."
@@ -248,9 +248,7 @@ class NVFP4MSECalibrator(MseCalibrator):
             best_flat = nvfp4_fp8_scale_sweep(x.detach(), self._global_amax, block_size=x.shape[-1])
             # Store the selected amax in fp32; the fake-quant kernel still returns
             # tensors in the requested output dtype.
-            self._best_amax_fast = best_flat.reshape(self._initial_amax.shape).to(
-                dtype=torch.float32
-            )
+            self._best_amax = best_flat.reshape(self._initial_amax.shape).to(dtype=torch.float32)
             return
 
         self._run_reference_collect(x)
@@ -258,7 +256,7 @@ class NVFP4MSECalibrator(MseCalibrator):
     def _run_reference_collect(self, x: torch.Tensor):
         super().collect(x)
         best_amax = super().compute_amax(verbose=False)
-        self._best_amax_fast = best_amax.to(dtype=torch.float32) if best_amax is not None else None
+        self._best_amax = best_amax.to(dtype=torch.float32) if best_amax is not None else None
         self._losses_sum = None
         # Synchronize before calibrating another weight so reference MSE sweeps do
         # not overlap. _losses_sum stores one fp32 reduced loss per candidate per
@@ -270,7 +268,7 @@ class NVFP4MSECalibrator(MseCalibrator):
     @torch.no_grad()
     def compute_amax(self, verbose: bool = False):
         """Return the cached per-block amax."""
-        return self._best_amax_fast
+        return self._best_amax
 
     def reset(self):
         """Reset per-cycle state. Keep ``_initial_amax`` so the calibrator stays reusable.
@@ -280,7 +278,7 @@ class NVFP4MSECalibrator(MseCalibrator):
         small enough to keep so a follow-up ``collect()`` can run again on the same
         calibrator instance.
         """
-        self._best_amax_fast = None
+        self._best_amax = None
         self._losses_sum = None
         self._candidates = None
         self._amax = None
--- a/modelopt/torch/quantization/calib/mse.py
+++ b/modelopt/torch/quantization/calib/mse.py
@@ -178,9 +178,8 @@ class NVFP4MSECalibrator(MseCalibrator):
     Uses a fused Triton kernel as an internal fast path on the first ``collect`` call
     when (a) ``error_func is None``, (b) the input tensor is on CUDA in the standard
     blocked ``[n_blocks, block_size]`` layout, and (c) Triton + the kernel package are
-    importable. Falls back to the reference 126-step Python sweep otherwise (custom
-    ``error_func`` users, multi-``collect`` activation flows, CPU inputs, or when the
-    fast path is disabled via ``MODELOPT_NVFP4_TRITON_SWEEP=0``).
+    importable. Falls back to the reference 126-step Python sweep otherwise and caches
+    the final amax immediately, so this calibrator is one-shot between resets.
     """
 
     def __init__(
@@ -193,14 +192,16 @@ class NVFP4MSECalibrator(MseCalibrator):
     ):
         """Initialize NVFP4 MSE calibrator with per-block and global amax."""
         super().__init__(amax=amax, axis=axis, quant_func=quant_func, error_func=error_func)
-        self._global_amax = global_amax
-        # Set by the Triton fast path on its (one-shot) collect; consumed by compute_amax.
+        self._global_amax = global_amax.to(dtype=torch.float32)
+        # Set by collect() after either sweep path; consumed by compute_amax.
         self._best_amax_fast: torch.Tensor | None = None
 
     def _compute_candidate_amax(self, candidates: torch.Tensor) -> torch.Tensor:
         if candidates.ndim != 0:  # Called during final compute amax
             candidates = candidates.view_as(self._initial_amax)
-        return torch.ones_like(self._initial_amax) * self._global_amax * candidates
+        return torch.ones_like(self._initial_amax, dtype=torch.float32) * (
+            self._global_amax * candidates
+        )
 
     def _generate_candidates(self, device: torch.device) -> torch.Tensor:
         """Generate the 126 valid FP8 E4M3 scale candidates."""
@@ -235,35 +236,41 @@ class NVFP4MSECalibrator(MseCalibrator):
 
     @torch.no_grad()
     def collect(self, x: torch.Tensor):
-        """Collect input statistics. Uses the Triton fast path when eligible."""
+        """Collect input statistics and cache the final per-block amax."""
         if self._best_amax_fast is not None:
             raise RuntimeError(
-                "NVFP4MSECalibrator: the Triton fast path produced a final amax on a "
-                "previous collect() call; multi-collect after the fast path is not "
-                "supported. Call reset() to start a fresh cycle, set "
-                "MODELOPT_NVFP4_TRITON_SWEEP=0, or pass a non-None error_func to force "
-                "the reference path for activation-style accumulation."
+                "NVFP4MSECalibrator: a previous collect() call produced a final amax; "
+                "multi-collect is not supported. Call reset() to start a fresh cycle."
             )
-        # Fast path is eligible only on the first call, before the reference accumulator
-        # has produced any state.
-        if self._losses_sum is None and self._can_use_triton_fast_path(x):
+        if self._can_use_triton_fast_path(x):
             from modelopt.torch.kernels.quantization.gemm import nvfp4_fp8_scale_sweep
 
             best_flat = nvfp4_fp8_scale_sweep(x.detach(), self._global_amax, block_size=x.shape[-1])
-            # Match the original shape/dtype of the initial amax so downstream
-            # load_calib_amax behaves identically to the reference path.
+            # Store the selected amax in fp32; the fake-quant kernel still returns
+            # tensors in the requested output dtype.
             self._best_amax_fast = best_flat.reshape(self._initial_amax.shape).to(
-                self._initial_amax.dtype
+                dtype=torch.float32
             )
             return
+
+        self._run_reference_collect(x)
+
+    def _run_reference_collect(self, x: torch.Tensor):
         super().collect(x)
+        best_amax = super().compute_amax(verbose=False)
+        self._best_amax_fast = best_amax.to(dtype=torch.float32) if best_amax is not None else None
+        self._losses_sum = None
+        # Synchronize before calibrating another weight so reference MSE sweeps do
+        # not overlap. _losses_sum stores one fp32 reduced loss per candidate per
+        # block. With 16-element NVFP4 blocks and bf16 weights, this is roughly
+        # 128 / 16 * (4 / 2) = 16x the calibrated weight size.
+        if x.is_cuda:
+            torch.cuda.synchronize(x.device)
 
     @torch.no_grad()
     def compute_amax(self, verbose: bool = False):
-        """Return the per-block amax — from the fast path if it ran, else from the reference sweep."""
-        if self._best_amax_fast is not None:
-            return self._best_amax_fast
-        return super().compute_amax(verbose=verbose)
+        """Return the cached per-block amax."""
+        return self._best_amax_fast
 
     def reset(self):
         """Reset per-cycle state. Keep ``_initial_amax`` so the calibrator stays reusable.
--- a/modelopt/torch/quantization/config.py
+++ b/modelopt/torch/quantization/config.py
@@ -722,9 +722,9 @@ class MseCalibConfig(QuantizeAlgorithmConfig):
     Finds a scale s (via amax a, with s = a / q_max) that minimizes the
     reconstruction error of a tensor after uniform Q→DQ:
 
-        s* = argmin_s  E[(X - DQ(Q(X; s)))^2],   X ∈ {weights | activations}
+        s* = argmin_s  E[(W - DQ(Q(W; s)))^2],   W ∈ weights
 
-    When fp8_scale_sweep is enabled, step_size is ignored.
+    When fp8_scale_sweep is enabled for a supported FP8-scale format, step_size is ignored.
     """
 
     method: Literal["mse"] = ModeloptField("mse")
@@ -755,8 +755,8 @@ class MseCalibConfig(QuantizeAlgorithmConfig):
         default=False,
         title="Enable FP8 scale sweep for NVFP4 per-block quantization.",
         description="If True, sweep all 128 FP8 E4M3 scale values instead of using multipliers. "
-        "Only applies to NVFP4 weight quantization. When enabled, num_steps, step_size, "
-        "start_multiplier, and stop_multiplier are ignored.",
+        "Applies to ModelOpt static NVFP4 weight quantizers and registered custom backends with "
+        "FP8 sweep support. Other weight quantizers use the multiplier search.",
     )
 
     distributed_sync: bool | None = ModeloptField(
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -77,7 +77,7 @@ __all__ = [
 def _is_calibrated_nvfp4_static(q) -> bool:
     """True iff ``q`` is an enabled NVFP4-static weight quantizer with ``_amax`` set."""
     return (
-        isinstance(q, TensorQuantizer)
+        isinstance(q, NVFP4StaticQuantizer)
         and not q._disabled
         and q.is_nvfp4_static
         and getattr(q, "_amax", None) is not None
@@ -106,8 +106,12 @@ def _collect_grouped_linears(model: nn.Module) -> list[list[nn.Module]]:
     return groups
 
 
+def _collect_weight_stats(quantizer: nn.Module, weight: torch.Tensor) -> None:
+    quantizer(weight)
+
+
 @torch.no_grad()
-def _bootstrap_uncalibrated_static_weight_quantizers(model: nn.Module) -> int:
+def _bootstrap_uncalibrated_weight_quantizers(model: nn.Module) -> int:
     """Re-run weight calibration on the weight tensor for quantizers missing ``_amax``.
 
     Covers MoE experts that ``max_calibrate`` skipped (no routed tokens) so MSE
@@ -121,19 +125,19 @@ def _bootstrap_uncalibrated_static_weight_quantizers(model: nn.Module) -> int:
             continue
         with enable_weight_access_and_writeback(module, model, name_to_module):
             for weight, q in module.iter_weights_for_calibration():
-                if not isinstance(q, TensorQuantizer) or q._disabled or q._dynamic:
+                if (
+                    not isinstance(q, TensorQuantizer)
+                    or q._disabled
+                    or q._dynamic
+                    or q._calibrator is None
+                ):
                     continue
-                if q._calibrator is None:
+                if weight.is_meta:
                     continue
-                if getattr(q, "_amax", None) is not None and not torch.all(q._amax == 0):
+                amax = q.amax
+                if amax is not None and (amax.is_meta or not torch.all(amax == 0)):
                     continue
-                q.disable_quant()
-                q.enable_calib()
-                q(weight)
-                if q._calibrator.compute_amax() is not None:
-                    q.load_calib_amax()
-                q.enable_quant()
-                q.disable_calib()
+                _run_and_load_max_stats(q, partial(_collect_weight_stats, weight=weight))
                 if hasattr(q._calibrator, "reset"):
                     q._calibrator.reset()
                 n += 1
@@ -160,18 +164,18 @@ def _sync_grouped_weight_global_amax(model: nn.Module) -> int:
     # quant_utils imports back from this module; top-level would cycle.
     from modelopt.torch.export.quant_utils import preprocess_linear_fusion
 
-    wq_attr = quantizer_attr_names("weight").weight_quantizer
     n_groups = 0
     for group in _collect_grouped_linears(model):
-        for child in group:
-            wq = getattr(child, wq_attr)
-            if not isinstance(wq, NVFP4StaticQuantizer):
-                NVFP4StaticQuantizer.from_tensor_quantizer(
-                    wq, global_amax=reduce_amax(wq._amax, axis=None)
-                )
         preprocess_linear_fusion(group)
         n_groups += 1
     return n_groups
+
+
+@torch.no_grad()
+def _promote_nvfp4_static_quantizers_with_global_amax_sync(model: nn.Module) -> None:
+    """Promote static NVFP4 weight quantizers and sync grouped global amax."""
+    promote_nvfp4_static_quantizers(model)
+    _sync_grouped_weight_global_amax(model)
 
 
 CalibratorFactory: TypeAlias = Callable[
@@ -197,6 +201,11 @@ def _register_fp8_sweep_calibrator(backend: str, calibrator_factory: CalibratorF
     _FP8_SWEEP_CALIBRATOR_REGISTRY[backend] = calibrator_factory
 
 
+def _uses_modelopt_fp8_weight_scales(weight_quantizer: TensorQuantizer) -> bool:
+    """Whether the internal ModelOpt FP8-scale MSE sweep applies to this quantizer."""
+    return weight_quantizer.backend is None and weight_quantizer.is_nvfp4_static
+
+
 def weight_only_quantize(model: nn.Module):
     """Just quantize the weights of the model."""
     name_to_module = dict(model.named_modules())
@@ -210,6 +219,16 @@ def weight_only_quantize(model: nn.Module):
                 for weight, weight_quantizer in module.iter_weights_for_calibration():
                     weight_quantizer(weight)
         seen_modules.add(module)
+
+
+def _run_and_load_max_stats(model: nn.Module, forward_loop: ForwardLoop | None = None):
+    """Run max-stat collection and load collected stats without post-processing."""
+    enable_stats_collection(model)
+    if forward_loop is None:
+        weight_only_quantize(model)
+    else:
+        forward_loop(model)
+    finish_stats_collection(model)
 
 
 def _has_expert_parallelism(module: nn.Module) -> bool:
@@ -285,12 +304,7 @@ def max_calibrate(
     See :class:`MaxCalibConfig <modelopt.torch.quantization.config.MaxCalibConfig>` for
     details on the remaining arguments.
     """
-    enable_stats_collection(model)
-    if forward_loop is None:
-        weight_only_quantize(model)
-    else:
-        forward_loop(model)
-    finish_stats_collection(model)
+    _run_and_load_max_stats(model, forward_loop)
 
     # Sync quantizer amax across local experts within each rank (for SequentialMLP)
     for name, module in model.named_modules():
@@ -300,12 +314,14 @@ def max_calibrate(
     # Fail fast on NVFP4 static-block with TP>1 (sharded_state_dict treats _amax as replicated).
     _check_nvfp4_static_tp_supported(model)
 
+    _bootstrap_uncalibrated_weight_quantizers(model)
+
     # Promote eligible static-block NVFP4 weight quantizers to NVFP4StaticQuantizer so
     # the static blockwise fake-quant path is used in forward and export picks up the
     # two-level (per-block + global) scaling. Run before the ``distributed_sync`` early
     # return so single-process callers also get the promotion. No-op for dynamic-block
     # / non-NVFP4 configs.
-    promote_nvfp4_static_quantizers(model)
+    _promote_nvfp4_static_quantizers_with_global_amax_sync(model)
 
     if not distributed_sync:
         return
@@ -451,6 +467,52 @@ def _mse_quant_func(x, amax, quantizer):
     return xq
 
 
+def _make_weight_mse_calibrator(
+    weight_quantizer: TensorQuantizer,
+    step_size: float,
+    start_multiplier: float,
+    stop_multiplier: float,
+    fp8_scale_sweep: bool,
+) -> _Calibrator | None:
+    """Create the MSE calibrator for one eligible weight quantizer."""
+    if (
+        not isinstance(weight_quantizer, TensorQuantizer)
+        or not weight_quantizer.is_enabled
+        or weight_quantizer._dynamic
+        or weight_quantizer._calibrator is None
+        or getattr(weight_quantizer, "_amax", None) is None
+    ):
+        return None
+
+    initial_amax = weight_quantizer._amax.clone().detach()
+    axis = weight_quantizer._calibrator._axis
+    quant_func = partial(_mse_quant_func, quantizer=weight_quantizer)
+
+    if fp8_scale_sweep:
+        backend: str | None = getattr(weight_quantizer, "backend", None)
+        backend_factory = (
+            _FP8_SWEEP_CALIBRATOR_REGISTRY.get(backend) if backend is not None else None
+        )
+        if backend is not None and backend_factory is not None:
+            return backend_factory(initial_amax, axis, quant_func)
+        elif _uses_modelopt_fp8_weight_scales(weight_quantizer):
+            return NVFP4MSECalibrator(
+                amax=initial_amax,
+                axis=axis,
+                global_amax=weight_quantizer.global_amax,
+                quant_func=quant_func,
+            )
+
+    return MseCalibrator(
+        amax=initial_amax,
+        axis=axis,
+        step_size=step_size,
+        start_multiplier=start_multiplier,
+        stop_multiplier=stop_multiplier,
+        quant_func=quant_func,
+    )
+
+
 @torch.no_grad()
 def mse_calibrate(
     model: nn.Module,
@@ -461,11 +523,11 @@ def mse_calibrate(
     stop_multiplier: float = 4.0,
     fp8_scale_sweep: bool = False,
 ):
-    """Calibrate the model using MSE-based amax search.
+    """Calibrate weight quantizers using MSE-based amax search.
 
-    This calibration method first uses max calibration to get initial amax values,
-    then searches for better amax values by minimizing the MSE between original
-    and quantized tensors.
+    This calibration method first uses max calibration to initialize amax values for
+    all quantizers, then searches for better weight amax values by minimizing the MSE
+    between original and quantized weights.
 
     Args:
         model: Model to be calibrated.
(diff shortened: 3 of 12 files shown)


modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -632,10 +632,13 @@ def load_calib_amax(self, *args, **kwargs):
                     err_msg
                     + " Passing 'strict=False' to `load_calib_amax()` will ignore the error."
                 )
+        calib_amax = calib_amax.clone().detach()
         if not hasattr(self, "_amax"):
-            self.register_buffer("_amax", calib_amax.clone().detach())
+            self.register_buffer("_amax", calib_amax)
```
realAsma: revert this; keep the calib_amax.clone().detach()

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -632,10 +632,13 @@ def load_calib_amax(self, *args, **kwargs):
...
+        elif self._amax.shape == calib_amax.shape and self._amax.dtype != calib_amax.dtype:
+            self._buffers["_amax"] = calib_amax.to(device=self._amax.device)
         else:
-            self._amax.data.copy_(calib_amax.clone().detach())
```
realAsma: can we simplify this?

too much branching;

if self._amax exist -> assert that self._amax shape is same as calib_amax;
then do  self._buffers["_amax"] = calib_amax.to(device = calib_amax.device if not hasattr(self, "_amax") else self._amax.device)

modelopt/torch/quantization/model_calib.py
```diff
@@ -423,12 +483,13 @@ def mse_calibrate(
     start_multiplier: float = 0.25,
     stop_multiplier: float = 4.0,
     fp8_scale_sweep: bool = False,
+    apply_mse_nvfp_static_only: bool = False,
```
realAsma: where is this used?

realAsma:
> _:robot: Bot comment._

Regarding CodeRabbit’s stats-lifecycle suggestion at https://github.com/NVIDIA/Model-Optimizer/pull/1536#discussion_r3290222409: I am going to leave this as-is. The helper currently has a simple, linear stats lifecycle, and adding a `try/finally` here makes the common path harder to read without addressing a demonstrated failure in this PR. We can revisit this if we see stale stats after a failed calibration run.

realAsma:
> _:robot: Bot comment._

Auto-replying to realAsma review comments:

- https://github.com/NVIDIA/Model-Optimizer/pull/1536#discussion_r3290192425: Applied in `e980773b` by keeping `calib_amax.clone().detach()` in the final buffer assignment path.
- https://github.com/NVIDIA/Model-Optimizer/pull/1536#discussion_r3290209743: Simplified in `e980773b` to a single buffer assignment with a shape assertion and device preservation.
- https://github.com/NVIDIA/Model-Optimizer/pull/1536#discussion_r3290238530: It is exposed through `MseCalibConfig`; `MseCalibrateModeDescriptor` wires that config to `mse_calibrate`, and the NVFP4 weight-MSE preset sets it so only static NVFP4 weight quantizers get the MSE refinement.

realAsma:
> _:robot: Bot comment._

/claude review

cjluo-nv (commented):

The actual diff goes well beyond the stated "sync on CUDA + remove `_wait_for_cuda_amax`" scope: it adds a new public-ish config field `apply_mse_nvfp_static_only`, removes `promote_nvfp4_static_quantizers` from `utils/core_utils.py`, restructures `mse_calibrate`, narrows `_bootstrap_uncalibrated_weight_quantizers` to NVFP4-static only (renamed accordingly), and changes `load_calib_amax` so that an existing `_amax` buffer is always **replaced** (inheriting `calib_amax`'s dtype) instead of `copy_`'d in place (preserving the buffer's dtype). That last change is model-wide, not NVFP4-specific, and the PR body doesn't call it out — worth a human pass to confirm no downstream caller relies on the buffer dtype being preserved (e.g. an fp16 `_amax` registered earlier silently becomes fp32). Tests look reasonable: new fp32-amax preservation test, fast-path-cache one-shot test, and a static-only-skip test for the registry. Couple of smaller items also worth eyeballing:

- `_promote_nvfp4_static_quantizers_with_global_amax_sync` is now called inside `max_calibrate`, so the bootstrap (formerly only triggered from `mse_calibrate`) now runs for *every* `max_calibrate` consumer (gptq, awq postprocess, smoothquant, etc.). Intended? The `_check_moe_calibration_complete` path runs after promotion — fine since the check looks at `_amax` on input quantizers too, but worth confirming nothing now warns where it didn't before.
- `local_hessian_calibrate` lost its explicit `_sync_grouped_weight_global_amax(model)` call. It's now subsumed by `max_calibrate` → promotion path, but local_hessian also has its own per-module promotion later. Double-promotion is a no-op (covered by `from_tensor_quantizer`), just noting the redundancy.
- `_compute_candidate_amax` now multiplies by `torch.ones_like(self._initial_amax, dtype=torch.float32)` — the `ones_like` is purely for broadcasting and a plain `(self._global_amax.to(torch.float32) * candidates).expand_as(self._initial_amax)` (or simpler, just relying on broadcast) would be cheaper; minor.

modelopt_recipes/configs/ptq/presets/model/nvfp4_w4a4_weight_mse_fp8_sweep.yaml
```diff
@@ -25,6 +25,7 @@ imports:
 algorithm:
   method: mse
   fp8_scale_sweep: true
+  apply_mse_nvfp_static_only: true
```
cjluo-nv: should we apply this to all MSE configs?

realAsma:
> _:robot: Bot comment._

Regarding https://github.com/NVIDIA/Model-Optimizer/pull/1536#pullrequestreview-4347784616: leaving this unchanged per branch-owner review. The current `mse_calibrate` flow intentionally installs the selected MSE calibrator on the weight quantizer before loading the chosen stats, and the extra restore/reset change is not needed here. I reverted the local experimental change.

realAsma:
> _:robot: Bot comment._

Regarding the NVFP4 reference MSE sync comment: updated the source comment to clarify that the CPU-GPU sync is there to prevent reference MSE calibration for another weight from running in parallel. The comment now also documents the retained `_losses_sum` memory: one fp32 reduced loss per candidate per block, about `126 * num_blocks * 4` bytes for NVFP4.

realAsma:
> _:robot: Bot comment._

Correction to my previous note about the NVFP4 reference MSE sync comment: I updated the memory wording to express `_losses_sum` relative to the calibrated weight. With 16-element NVFP4 blocks and bf16 weights, `_losses_sum` is roughly `128 / 16 * (4 / 2) = 16x` the calibrated weight size, and the sync prevents reference MSE calibration for another weight from overlapping that allocation.

realAsma:
> _:robot: Bot comment._

Regarding https://github.com/NVIDIA/Model-Optimizer/pull/1536#pullrequestreview-4347972933: Thanks for the careful pass. The broader scope is intentional: the fp32 amax preservation, static-NVFP4-only MSE gate, max-calibration promotion, and grouped global-amax sync are tied together so static NVFP4 weights are finalized consistently before downstream calibrators/export. I agree the PR body should make that scope clearer, especially the `load_calib_amax` dtype behavior and the fact that `max_calibrate` now performs static NVFP4 promotion. The local-hessian promotion path is redundant but no-op after `max_calibrate`. I left the `ones_like` expression as-is because it keeps the shape/dtype broadcasting explicit. I also pushed `fc8a95d0` to clarify the NVFP4 reference MSE sync comment and its `_losses_sum` memory cost.

modelopt/torch/quantization/calib/mse.py
```diff
@@ -235,35 +237,40 @@ def _can_use_triton_fast_path(self, x: torch.Tensor) -> bool:
...
+                dtype=torch.float32
             )
             return
+
+        self._run_reference_collect(x)
+
+    def _run_reference_collect(self, x: torch.Tensor):
         super().collect(x)
+        if x.is_cuda:
+            torch.cuda.synchronize(x.device)
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Kept the sync and clarified why it is intentional in `fc8a95d0`. The sync prevents reference MSE calibration for another weight from overlapping the retained `_losses_sum` allocation.

modelopt/torch/quantization/model_calib.py
```diff
@@ -204,6 +211,16 @@ def weight_only_quantize(model: nn.Module):
...
 
 
+def _run_and_load_max_stats(model: nn.Module, forward_loop: ForwardLoop | None = None):
+    """Run max-stat collection and load collected stats without post-processing."""
+    enable_stats_collection(model)
+    if forward_loop is None:
+        weight_only_quantize(model)
+    else:
+        forward_loop(model)
+    finish_stats_collection(model)
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Leaving this unchanged. `_run_and_load_max_stats` has a short linear stats lifecycle, and adding a `try/finally` here would make the common path harder to read without addressing a demonstrated stale-stats issue in this PR.

tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py
```diff
@@ -270,6 +292,7 @@ def test_mse_calibrate_end_to_end(monkeypatch):
 
     import modelopt.torch.quantization as mtq
     from modelopt.torch.quantization.extensions import get_cuda_ext_mx
+    from modelopt.torch.quantization.nn import TensorQuantizer
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Fixed in `e980773b` by moving `TensorQuantizer` to module scope.

tests/unit/torch/quantization/test_mse_calibrator.py
```diff
@@ -640,3 +666,46 @@ def test_unregistered_backend_uses_default_mse_calibrator(self):
...
+
+        _promote_nvfp4_static_quantizers_with_global_amax_sync(model)
+
+        assert isinstance(model.weight_quantizer, NVFP4StaticQuantizer)
+        assert torch.equal(model.weight_quantizer.global_amax, torch.tensor(5.0))
+
+    def test_grouped_static_nvfp4_quantizers_share_global_amax(self):
+        from modelopt.torch.quantization.model_calib import (
+            _promote_nvfp4_static_quantizers_with_global_amax_sync,
+        )
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Fixed in `e980773b` by moving `_promote_nvfp4_static_quantizers_with_global_amax_sync` to module scope.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py

Applied in `e980773b` by keeping `calib_amax.clone().detach()` in the final buffer assignment path.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py

Simplified in `e980773b` to a single buffer assignment with a shape assertion and device preservation.

modelopt/torch/quantization/model_calib.py

It is exposed through `MseCalibConfig`; `MseCalibrateModeDescriptor` passes that config into `mse_calibrate`, and the NVFP4 weight-MSE preset enables it so only static NVFP4 weight quantizers get this MSE refinement.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -632,10 +632,12 @@ def load_calib_amax(self, *args, **kwargs):
...
-        if not hasattr(self, "_amax"):
-            self.register_buffer("_amax", calib_amax.clone().detach())
-        else:
-            self._amax.data.copy_(calib_amax.clone().detach())
+        if hasattr(self, "_amax"):
+            assert self._amax.shape == calib_amax.shape, (
+                "Changing shape when loading calibration amax is not allowed."
+            )
+        device = self._amax.device if hasattr(self, "_amax") else calib_amax.device
+        self._buffers["_amax"] = calib_amax.clone().detach().to(device=device)
```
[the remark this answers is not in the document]

Leaving this unchanged. Direct `_buffers` assignment is still a registered module buffer update, and the single path keeps the `load_calib_amax` logic aligned with the simplification requested earlier.

realAsma:
Is fp32 MSE scale preserved after save/restore?

realAsma:
> _:robot: Bot comment._

Regarding https://github.com/NVIDIA/Model-Optimizer/pull/1536#issuecomment-4522056254: yes. The MSE amax is stored in `_amax` as fp32 by `load_calib_amax`, and the quantizer save/restore path records buffer dtype in `_pytorch_state_metadata` before recreating buffers on restore. I also verified a local `TensorQuantizer` round trip: original `_amax`, saved state, metadata, and restored `_amax` all stayed `torch.float32`.

modelopt/torch/quantization/config.py
```diff
@@ -759,6 +759,12 @@ class MseCalibConfig(QuantizeAlgorithmConfig):
         "start_multiplier, and stop_multiplier are ignored.",
     )
 
+    apply_mse_nvfp_static_only: bool | None = ModeloptField(
+        default=False,
```
Fridah-nv: Is this a typo? Should be `apply_mse_nvfp4_static_only`?

Fridah-nv:
At first glance, I find the functionality of the two flag `fp8_scale_sweep` and `apply_mse_nvfp_static_only` a bit confusing. Talked with claude for some better naming ideas. 
Do we agree that for NVFP4 we always recommend using FP8 sweep? If so, the  the `fp8_scale_sweep=false, apply_mse_nvfp_static_only=true` is redundant.

Here's the full suggestion, makes sense on my side:

**Suggestion: collapse `fp8_scale_sweep` + `apply_mse_nvfp_static_only` into a single `sweep_mode` enum**

  The two booleans give four combinations but only three are sensible — the `fp8_scale_sweep=false, apply_mse_nvfp_static_only=true` cell silently runs 
  multiplier-MSE on NVFP4-static weights, which nobody would intentionally pick. An enum eliminates that footgun and makes intent readable at the call site.

  ```yaml
  algorithm:
    method: mse
    sweep_mode: fp8_grid_nvfp4_only   # multiplier | fp8_grid | fp8_grid_nvfp4_only
  ```

  | `sweep_mode`            | NVFP4-static weight | Other weight                       | Replaces today's flags |
  |-------------------------|---------------------|------------------------------------|------------------------|
  | `multiplier` (default)  | multiplier MSE      | multiplier MSE                     | `false / false`        |
  | `fp8_grid`              | FP8 grid sweep      | FP8 grid via registry, else multiplier MSE | `true / false`         |
  | `fp8_grid_nvfp4_only`   | FP8 grid sweep      | **skipped** (keeps max-cal amax)   | `true / true`          |

  The preset becomes one line:

  ```yaml
  algorithm:
    method: mse
    sweep_mode: fp8_grid_nvfp4_only
  ```

  Wins:

  - Three legible modes instead of four boolean combos with one degenerate cell.
  - `fp8_grid` keeps the backend-generic FP8 sweep registry as a valid extension point (matches how `fp8_scale_sweep` works today).
  - The `_only` suffix makes the scope semantics explicit — clearer than `apply_mse_nvfp_static_only` and fixes the `nvfp` → `nvfp4` typo on the way.
  - Forward-compatible: a future `fp8_grid_*_only` mode (e.g. `fp8_grid_mxfp8_only`) is easy to add without another boolean.

  If a full enum migration is heavier than worth it for this PR, the minimal-change fallback is just to rename `apply_mse_nvfp_static_only` → 
  `nvfp4_static_only` (typo + drop the redundant `apply_mse_` prefix) and leave `fp8_scale_sweep` alone since it's genuinely backend-generic via the 
  registry.

modelopt/torch/quantization/calib/mse.py
```diff
@@ -194,13 +193,16 @@ def __init__(
...
         self._best_amax_fast: torch.Tensor | None = None
 
     def _compute_candidate_amax(self, candidates: torch.Tensor) -> torch.Tensor:
+        candidates = candidates.to(dtype=torch.float32)
         if candidates.ndim != 0:  # Called during final compute amax
             candidates = candidates.view_as(self._initial_amax)
-        return torch.ones_like(self._initial_amax) * self._global_amax * candidates
+        return torch.ones_like(self._initial_amax, dtype=torch.float32) * (
+            self._global_amax.to(dtype=torch.float32) * candidates
+        )
```
Fridah-nv, line 204: A question here, in terms of performance, is using `global_amax * candidates.view_as(self._initial_amax).contiguous()` better than using `torch.ones_like()` here?

modelopt/torch/quantization/calib/mse.py
```diff
@@ -235,35 +237,44 @@ def _can_use_triton_fast_path(self, x: torch.Tensor) -> bool:
...
     def compute_amax(self, verbose: bool = False):
-        """Return the per-block amax — from the fast path if it ran, else from the reference sweep."""
-        if self._best_amax_fast is not None:
-            return self._best_amax_fast
-        return super().compute_amax(verbose=verbose)
+        """Return the cached per-block amax."""
+        best_amax_fast = getattr(self, "_best_amax_fast", None)
+        if best_amax_fast is None:
+            return None
+        return best_amax_fast
```
Fridah-nv: In ` __init__`, we have `self._best_amax_fast: torch.Tensor | None = None`. The attribute `_best_amax_fast` is always set, do we need the check `if best_amax_fast is None:` here?

modelopt/torch/quantization/model_calib.py
```diff
@@ -204,6 +211,16 @@ def weight_only_quantize(model: nn.Module):
         seen_modules.add(module)
 
 
+def _run_and_load_max_stats(model: nn.Module, forward_loop: ForwardLoop | None = None):
+    """Run max-stat collection and load collected stats without post-processing."""
```
Fridah-nv, line 225: nit: can we rename the param name from `model` to `module`?

modelopt/torch/quantization/calib/mse.py
```diff
@@ -235,35 +237,44 @@ def _can_use_triton_fast_path(self, x: torch.Tensor) -> bool:
...
+    def _run_reference_collect(self, x: torch.Tensor):
         super().collect(x)
+        best_amax = super().compute_amax(verbose=False)
+        self._best_amax_fast = best_amax.to(dtype=torch.float32) if best_amax is not None else None
+        self._losses_sum = None
+        # Synchronize before calibrating another weight so reference MSE sweeps do
+        # not overlap. _losses_sum stores one fp32 reduced loss per candidate per
+        # block. With 16-element NVFP4 blocks and bf16 weights, this is roughly
+        # 128 / 16 * (4 / 2) = 16x the calibrated weight size.
+        if x.is_cuda:
```
Fridah-nv, line 265: I'm trying to understand the optimization here. Is it a memory (cache) optimization rather than for correctness?

Claude's explanation on this:
```
  # Drain the loss-reduction kernels so the caching allocator can recycle
  # _losses_sum (≈16× the bf16 weight size — 126 fp32 losses per 16-elem
  # block) before the next weight in the MSE calibration loop allocates
  # its own copy.
```
Saying this is for `reference MSE sweeps do not overlap` might be confusing

tests/gpu/torch/quantization/test_nvfp4_static_quantizer_cuda.py
```diff
@@ -200,13 +224,8 @@ def quant_func(x, amax):
...
 
-    def test_multiple_collections(self, device, monkeypatch):
-        """Test that multiple collections accumulate correctly.
-
-        Multi-collect is reference-path-only — the Triton fast path is one-shot
-        and refuses a second ``collect()`` until ``reset()``. Forcing the env var
-        keeps this exercising the accumulator.
-        """
+    def test_reference_path_reset_allows_recollect(self, device, monkeypatch):
+        """Test that reference fallback is one-shot until reset."""
```
Fridah-nv, line 340: nit: can we merge this test with `tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py:142:test_reset_allows_recollect` and parameterize them

tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py
```diff
@@ -198,7 +221,7 @@ def test_dispatch_fast_path_default():
 
 @requires_triton
 def test_dispatch_env_optout_falls_back():
-    """``MODELOPT_NVFP4_TRITON_SWEEP=0`` forces the reference 126-step sweep."""
+    """``MODELOPT_NVFP4_TRITON_SWEEP=0`` forces the cached reference sweep."""
```
Fridah-nv: The main part of this test seems already in `test_nvfp4_static_quantizer_cuda.py:198 test_collect_and_compute_amax `, consider merging them

realAsma:
> _:robot: Bot comment._

@Fridah-nv addressed in `419657d2`.

I went with the smaller config surface: removed `apply_mse_nvfp_static_only` and kept `fp8_scale_sweep` as the single user-facing switch. The implementation now uses the FP8 scale sweep where the quantizer/backend supports it: registered custom backends use their sweep factory, and internal ModelOpt static NVFP4 weights use `NVFP4MSECalibrator` because they have E4M3 block scales. Other weight quantizers continue to use the regular multiplier MSE search.

modelopt/torch/quantization/model_calib.py
```diff
@@ -113,21 +112,16 @@ def _bootstrap_uncalibrated_weight_quantizers(model: nn.Module) -> int:
             continue
         with enable_weight_access_and_writeback(module, model, name_to_module):
             for weight, q in module.iter_weights_for_calibration():
-                if not isinstance(q, TensorQuantizer) or q._disabled or q._dynamic:
-                    continue
-                if q._calibrator is None:
+                if (
+                    not isinstance(q, TensorQuantizer)
```
cjluo-nv, line 129: > _Bot comment._

Bootstrap was narrowed from "any uncalibrated weight quantizer" to "NVFP4-static only" and the unconditional bootstrap step at the top of `mse_calibrate` was removed. For non-NVFP4 MoE configs (INT8/FP8 weight quantization) where some experts get no routed tokens, those weight quantizers will now stay at `_amax=None` after `max_calibrate`, `_make_weight_mse_calibrator` returns `None` for them (the `_amax is None` guard), and they get silently skipped by `mse_calibrate`. The renamed test `test_max_calibrate_populates_dead_static_nvfp4_expert_quantizers` no longer covers the non-NVFP4 case either. Please confirm this regression is intended (and ideally call it out in the PR body), or restore the non-NVFP4 dead-expert handling.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
cjluo-nv: > _Bot comment._

This change replaces `self._buffers["_amax"]` rather than `copy_`'ing in-place. That preserves the calibrator's dtype (fp32) — which is the goal here — but it also changes the *identity* of the `_amax` tensor for every quantizer model-wide, not just NVFP4. Any external caller that grabs `q._amax` and then calls `q.load_calib_amax()` will silently see the pre-load tensor afterwards. The PR self-marks back-compat but doesn't call this out. Worth either documenting in the PR body / a one-line code comment, or auditing the in-tree callers (`amax.setter`, calibration-restore paths, exporters) to confirm none rely on identity.

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -940,26 +940,3 @@ def update_quant_cfg_with_kv_cache_quant(
         quant_cfg["algorithm"] = "max"
```
cjluo-nv, line 945: > _Bot comment._

`promote_nvfp4_static_quantizers` was importable as `from modelopt.torch.quantization.utils import promote_nvfp4_static_quantizers` (used by `tests/gpu/torch/quantization/test_gptq.py` before this PR — that import was also removed in the diff). Removing it outright will `ImportError` for any external script using the same path while the PR claims backward compatibility. Consider keeping a thin shim that warns and forwards to `_promote_nvfp4_static_quantizers_with_global_amax_sync`, or note the removal in the CHANGELOG.

cjluo-nv (commented):

A lot of progress on this re-review — the fp32-amax preservation, test-import scope, FP8-sweep config simplification (collapsing `apply_mse_nvfp_static_only` into `fp8_scale_sweep`), and the `load_calib_amax` simplification have all landed. Below are the points still worth a human pass before approval; ones the author has responded to are marked with :speech_balloon:.

- :speech_balloon: **Non-NVFP4 dead-expert MoE MSE regression** — `_bootstrap_uncalibrated_weight_quantizers` was narrowed to `_bootstrap_uncalibrated_static_weight_quantizers` (NVFP4-only) and the unconditional bootstrap step at the top of `mse_calibrate` was removed. Author's response (to cjluo) confirms scope expansion is intentional and tied to NVFP4 finalization, but doesn't directly address what happens to INT8/FP8 MoE configs with dead experts under `mse`: those weight quantizers will now stay at `_amax=None`, `_make_weight_mse_calibrator` silently returns `None`, and the renamed `test_max_calibrate_populates_dead_static_nvfp4_expert_quantizers` no longer exercises the non-NVFP4 path. Worth confirming this is intended (and not an unrelated MoE recipe regression).

- :speech_balloon: **`load_calib_amax` buffer identity change is model-wide** — cjluo flagged that replacing `self._buffers["_amax"]` instead of `self._amax.data.copy_()` changes the *identity* of the buffer for every quantizer, not just NVFP4. Author kept the new behavior to preserve fp32 dtype and added a regression test for the dtype, but the identity change for non-NVFP4 callers (anyone holding a stale `_amax` reference across `load_calib_amax`) hasn't been audited — the PR body still doesn't call this out as a behavior change. A second pair of eyes on downstream callers would be reassuring.

- **`promote_nvfp4_static_quantizers` removed from `utils/core_utils.py` with no deprecation shim** — the PR self-marks "backward compatible", but `from modelopt.torch.quantization.utils import promote_nvfp4_static_quantizers` (used by at least `tests/gpu/torch/quantization/test_gptq.py` before this PR) will now `ImportError` for any external caller. Either keep a thin deprecation shim that calls `_promote_nvfp4_static_quantizers_with_global_amax_sync`, or note the removal in CHANGELOG and uncheck the back-compat box.

- **In-function imports in test helper** — `tests/unit/torch/quantization/test_mse_calibrator.py::TestRegisterFP8SweepCalibrator._quantize_and_calibrate` still does `import modelopt.torch.quantization as mtq` / `mse_calibrate` / `register_quant_backend` inside the method. Earlier in-function imports in the same file and in `test_nvfp4_fp8_sweep_kernel.py` were moved to module scope; this one was missed. Style-only, not a blocker.

- **Minor**: the `lambda q: q(weight)` in `mse_calibrate` captures the loop `weight`; safe today because `_run_and_load_max_stats` invokes synchronously, but `lambda q, w=weight: q(w)` would harden it against any future deferral. The `getattr(self, "_best_amax_fast", None)` in `NVFP4MSECalibrator.compute_amax` is also redundant given `__init__` sets the attribute — a plain `if self._best_amax_fast is None` would do.

---

- `tests/unit/torch/quantization/test_mse_calibrator.py:562` — > _Bot comment._

Style nit: `mtq`, `mse_calibrate`, and `register_quant_backend` are still imported inside this method. Earlier in-function imports in this file and in `test_nvfp4_fp8_sweep_kernel.py` were moved to module scope in `e980773b`; this helper looks like it was missed. Not a blocker.

modelopt/torch/quantization/model_calib.py
```diff
@@ -414,6 +427,53 @@ def _mse_quant_func(x, amax, quantizer):
...
+        or weight_quantizer._calibrator is None
+        or getattr(weight_quantizer, "_amax", None) is None
+    ):
+        return None
+
+    initial_amax = weight_quantizer._amax.clone().detach()
+    axis = weight_quantizer._calibrator._axis
+    quant_func = partial(_mse_quant_func, quantizer=weight_quantizer)
+
+    if fp8_scale_sweep:
```
realAsma, line 491: nit:

consolidate the branching a bit
```
    if fp8_scale_sweep and backend is not None and backend_factory is not None:
     ....
    elif  fp8_scale_sweep and _uses_modelopt_fp8_weight_scales(weight_quantizer):
```

modelopt/torch/quantization/calib/mse.py
realAsma: I need to check this. This however seems trivial overhead to me. `torch.ones_like(self._initial_amax, dtype=torch.float32)` sizes are at max weight_size / 8.

modelopt/torch/quantization/calib/mse.py
realAsma: Is it a memory (cache) optimization -> yes.

realAsma:
> _:robot: Bot comment._

Pushed `e3da774a` with the review-fix batch.

Addressed the bot-tagged feedback from the latest review:
- https://github.com/NVIDIA/Model-Optimizer/pull/1536#discussion_r3305022071: `max_calibrate()` now bootstraps missing weight amax for skipped weight quantizers before static NVFP4 promotion, so `mse_calibrate()` gets the same non-NVFP4 dead-expert coverage through the shared max-calibration path.
- https://github.com/NVIDIA/Model-Optimizer/pull/1536#discussion_r3305022078: `load_calib_amax()` now raises `ValueError` for shape mismatches under optimized Python and has a comment explaining the intentional buffer replacement for dtype preservation.
- https://github.com/NVIDIA/Model-Optimizer/pull/1536#discussion_r3305022084: `promote_nvfp4_static_quantizers()` is restored as the public static-NVFP4 promotion implementation, and the calibration-side helper delegates to it before grouped global-amax sync.
- https://github.com/NVIDIA/Model-Optimizer/pull/1536#pullrequestreview-4365229392: remaining `test_mse_calibrator.py` helper imports were moved to module scope.

Validation before push:
- `git diff --check`
- `pre-commit run --files modelopt/torch/quantization/model_calib.py modelopt/torch/quantization/nn/modules/tensor_quantizer.py modelopt/torch/quantization/utils/core_utils.py tests/unit/torch/quantization/test_mse_calibrator.py`
- `pytest_pwd tests/unit/torch/quantization/test_mse_calibrator.py::TestRegisterFP8SweepCalibrator tests/unit/torch/quantization/test_mse_calibrator.py::TestStaticNVFP4Promotion -q`

modelopt/torch/quantization/calib/mse.py
realAsma: addressed, thanks!

modelopt/torch/quantization/model_calib.py
realAsma: I am not sure about this. I prefer the name model over module. module is more generic in the context of python (such as a python module).

modelopt_recipes/configs/ptq/presets/model/nvfp4_w4a4_weight_mse_fp8_sweep.yaml
realAsma: removed this flag; no longer applicable

tests/gpu/torch/quantization/test_nvfp4_fp8_sweep_kernel.py
realAsma: done

tests/gpu/torch/quantization/test_nvfp4_static_quantizer_cuda.py
realAsma: I kept these separate because they cover different paths. `test_reset_allows_recollect` exercises the Triton fast path, while `test_reference_path_reset_allows_recollect`
  forces the reference fallback with `MODELOPT_NVFP4_TRITON_SWEEP=0` and verifies the same one-shot/reset behavior there. Parameterizing across the two files would make the
  dispatch setup less local without removing much meaningful duplication.

cjluo-nv (commented):

Re-review summary: most prior concerns are addressed, but one model-wide behavior change still merits human sign-off.

Addressed since the previous round:
- :speech_balloon: Non-NVFP4 dead-expert MoE bootstrap (cjluo) — `e3da774a` widens `_bootstrap_uncalibrated_weight_quantizers` back to all quantizers and runs it inside `max_calibrate` (default `bootstrap_weight_quantizers=True`), with a new `test_max_calibrate_bootstraps_non_nvfp4_dead_weight_quantizer` test. Looks resolved.
- :speech_balloon: `promote_nvfp4_static_quantizers` removal (cjluo) — `e3da774a` restores it in `utils/core_utils.py` as the public helper; `from modelopt.torch.quantization.utils import promote_nvfp4_static_quantizers` works again via the star re-export.
- :speech_balloon: `load_calib_amax` `assert` → `raise ValueError` and new-buffer comment (claude) — applied.
- :speech_balloon: In-method test imports (CodeRabbit) — moved to module scope in `test_mse_calibrator.py` and `test_nvfp4_fp8_sweep_kernel.py`.
- :speech_balloon: `fp8_scale_sweep` + `apply_mse_nvfp_static_only` two-flag confusion (Fridah-nv) — `419657d2` collapses to a single `fp8_scale_sweep` switch, with dispatch handled in `_make_weight_mse_calibrator` via `_uses_modelopt_fp8_weight_scales` and the registered backend factory.
- Lambda capture of loop `weight` in `mse_calibrate` (claude) — implicitly fixed by switching to `partial(_collect_weight_stats, weight=weight)`.

Still warrants a human pass before approval:
- :speech_balloon: `load_calib_amax` now *replaces* `self._buffers["_amax"]` instead of `copy_`'ing in-place (cjluo). Author kept the new behavior to preserve fp32 dtype and added an inline comment + a regression test, but this is a model-wide identity change for every quantizer (not just NVFP4). Any caller holding a stale `q._amax` reference across `load_calib_amax()` will now silently see the pre-load tensor. The PR body still doesn't call this out as a behavior change. Quick audit of in-tree callers of `load_calib_amax` (export paths, calibration restore, `amax.setter`) would be reassuring before sign-off.
- `_run_and_load_max_stats` has no `try/finally` around `enable_stats_collection` / `finish_stats_collection` (CodeRabbit). Author explicitly declined; reasonable given the linear lifecycle, but worth a final sanity check that no new caller (e.g. `max_calibrate` consumers) can leave quantizers stuck in `_if_calib=True` after a forward exception.

No design-review concerns: this PR refactors existing calibration code rather than introducing a new subsystem. Test coverage looks solid (fp32-amax preservation, one-shot collect/reset, dead-expert bootstrap for NVFP4 *and* non-NVFP4, registered FP8-sweep dispatch, grouped global-amax sync).

modelopt/torch/quantization/calib/mse.py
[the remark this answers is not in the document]
jenchen13: +1 always multiplying by torch.ones seems unnecessary

modelopt/torch/quantization/calib/mse.py
```diff
@@ -235,35 +237,41 @@ def _can_use_triton_fast_path(self, x: torch.Tensor) -> bool:
...
             return
+
+        self._run_reference_collect(x)
+
+    def _run_reference_collect(self, x: torch.Tensor):
         super().collect(x)
+        best_amax = super().compute_amax(verbose=False)
+        self._best_amax_fast = best_amax.to(dtype=torch.float32) if best_amax is not None else None
+        self._losses_sum = None
+        # Synchronize before calibrating another weight so reference MSE sweeps do
```
jenchen13, line 261: nit: make comment shorter?

modelopt/torch/quantization/model_calib.py
```diff
@@ -69,7 +69,7 @@
 def _is_calibrated_nvfp4_static(q) -> bool:
     """True iff ``q`` is an enabled NVFP4-static weight quantizer with ``_amax`` set."""
     return (
-        isinstance(q, TensorQuantizer)
+        isinstance(q, NVFP4StaticQuantizer)
         and not q._disabled
         and q.is_nvfp4_static
```
jenchen13, line 82: do we need to check both `is_nvfp4_static` and `NVFP4StaticQuantizer` ?

realAsma:
> _:robot: Bot comment._

Addressing the latest modelopt review note about the model-wide `load_calib_amax` buffer identity change: pushed `9e36ea34`.

`TensorQuantizer.load_calib_amax()` is now restored to the main-branch register-or-`copy_` behavior, so non-NVFP4 quantizers no longer replace `_amax` and retain the prior buffer identity/dtype semantics.

The fp32 preservation requirement is now isolated to `NVFP4StaticQuantizer.load_calib_amax()`: it upcasts an existing lower-precision `_amax` to fp32 before delegating to the base implementation. That means buffer replacement is scoped to NVFP4 static quantizers, and only when dtype migration is required for the NVFP4 MSE fp32-amax path.

Focused validation passed:
`pytest_pwd tests/gpu/torch/quantization/test_nvfp4_static_quantizer_cuda.py::TestNVFP4StaticQuantizer::test_load_calib_amax_preserves_fp32_result_dtype`

modelopt/torch/quantization/model_calib.py
```diff
@@ -414,6 +429,52 @@ def _mse_quant_func(x, amax, quantizer):
...
+            return backend_factory(initial_amax, axis, quant_func)
+        elif _uses_modelopt_fp8_weight_scales(weight_quantizer):
+            return NVFP4MSECalibrator(
+                amax=initial_amax,
+                axis=axis,
+                global_amax=weight_quantizer.global_amax,
+                quant_func=quant_func,
+            )
+
+    return MseCalibrator(
```
jenchen13, line 510: we should skip initializing MseCalibrator for non-NVFP4 static quantizers .. is this code even necessary?

realAsma: We should skip initializing MseCalibrator for non-NVFP4 static quantizers
We skip MSE calibration for non-NVFP4 if fp8_sweep is enabled; otherwise we support sweeping MSE scales. This design is general and support broadly.

modelopt/torch/quantization/calib/mse.py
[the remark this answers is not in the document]
realAsma: This is a small tensor; size is weight_sizein bf16 /8; There is no benefit I see in using view_as; I generally prefer contiguous tensors.

modelopt/torch/quantization/model_calib.py
realAsma: I think keeping both checks is intentional here. is_nvfp4_static verifies the quantizer config is actually static NVFP4, while NVFP4StaticQuantizer confirms it has been promoted and has the static path/global_amax behavior. This helper is only for calibrated promoted static NVFP4 quantizers used in grouped global-amax sync.

modelopt/torch/quantization/calib/mse.py
realAsma: Do you see any redundancy? I made this comment explicitly so the reason for cuda sync is clear.

cjluo-nv (approved):

modelopt/torch/quantization/calib/mse.py
```diff
@@ -194,13 +193,16 @@ def __init__(
         """Initialize NVFP4 MSE calibrator with per-block and global amax."""
         super().__init__(amax=amax, axis=axis, quant_func=quant_func, error_func=error_func)
         self._global_amax = global_amax
-        # Set by the Triton fast path on its (one-shot) collect; consumed by compute_amax.
+        # Set by collect() after either sweep path; consumed by compute_amax.
         self._best_amax_fast: torch.Tensor | None = None
 
     def _compute_candidate_amax(self, candidates: torch.Tensor) -> torch.Tensor:
+        candidates = candidates.to(dtype=torch.float32)
```
jenchen13: why do we cast the candidates to FP32 here instead of have `_generate_candidates` or `fp8_scale_candidates()` return scale candidates in FP32?

modelopt/torch/quantization/calib/mse.py
```diff
@@ -235,35 +237,41 @@ def _can_use_triton_fast_path(self, x: torch.Tensor) -> bool:
...
+                "NVFP4MSECalibrator: a previous collect() call produced a final amax; "
+                "multi-collect is not supported. Call reset() to start a fresh cycle."
             )
-        # Fast path is eligible only on the first call, before the reference accumulator
-        # has produced any state.
-        if self._losses_sum is None and self._can_use_triton_fast_path(x):
+        if self._can_use_triton_fast_path(x):
             from modelopt.torch.kernels.quantization.gemm import nvfp4_fp8_scale_sweep
 
             best_flat = nvfp4_fp8_scale_sweep(x.detach(), self._global_amax, block_size=x.shape[-1])
```
jenchen13, line 248: does `_global_amax` also need to be FP32 since it is the input into the triton FP8 scale sweep kernel? why not set _global_amax to FP32 in the constructor L195 instead of in line 204?

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -1358,6 +1358,19 @@ class NVFP4StaticQuantizer(TensorQuantizer):
     Uses _global_amax and inherited _amax for per-block amax values.
     """
 
+    def _preserve_amax_in_fp32(self):
+        amax = getattr(self, "_amax", None)
+        if amax is not None:
+            self._amax = amax.to(dtype=torch.float32)
+        global_amax = getattr(self, "_global_amax", None)
+        if global_amax is not None:
```
jenchen13, line 1367: if global_amax is already FP32 here then we don't need to cast it so many times in `mse.py`

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -1358,6 +1358,19 @@ class NVFP4StaticQuantizer(TensorQuantizer):
...
+            self._global_amax = global_amax.to(dtype=torch.float32)
+
+    def _amax_setter_helper(self, value):
+        value = value.to(dtype=torch.float32)
+        self._preserve_amax_in_fp32()
```
jenchen13: there's too much repetition in this file. if the amax is already stored in fp32 we do not need to re-cast it to FP32 before setting a new value. just cast the new value to FP32 then set. we also should not need to call `_preserve_amax_in_fp32` so many times, it should only be called once during loading (not even necessary to cast during saving since if amax is FP32 and any transformations on it should keep amax in FP32)

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -1368,14 +1381,18 @@ def from_tensor_quantizer(
             tq: The TensorQuantizer to convert.
             global_amax: Optional global amax value to set on the quantizer.
         """
-        if isinstance(tq, cls):
+
+        def _preserve_and_set_global_amax(tq):
```
jenchen13, line 1385: this seems duplicate with `_amax_setter_helper`

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -1393,10 +1410,30 @@ def global_amax(self, value):
             return
         if not isinstance(value, torch.Tensor):
             value = torch.tensor(value)
-        if not hasattr(self, "_global_amax") or self._global_amax is None:
+        value = value.to(dtype=torch.float32)
```
jenchen13: use _amax_setter_helper, prevent duplication

jenchen13 (commented):
can we identify the minimal amount of FP32 casting needed for TensorQuantizer amax? right now it seems to be repeated in 10+ places for safety reasons which should not be necessary if the quantization algorithms respect the initial dtypes

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
realAsma: I am just being defensive - seems worth it to protect against mutation outside of mtq.quantize API.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
realAsma: no `_amax_setter_helper`  only considers `_amax`, not `_global_amax`

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
realAsma: same as https://github.com/NVIDIA/Model-Optimizer/pull/1536/changes#r3320845867

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
realAsma: explained here - https://github.com/NVIDIA/Model-Optimizer/pull/1536/changes#r3320843695

Fridah-nv (approved):

modelopt/torch/quantization/calib/mse.py
realAsma: done

modelopt/torch/quantization/calib/mse.py
realAsma: done

modelopt/torch/quantization/calib/mse.py
```diff
@@ -193,14 +192,16 @@ def __init__(
     ):
         """Initialize NVFP4 MSE calibrator with per-block and global amax."""
         super().__init__(amax=amax, axis=axis, quant_func=quant_func, error_func=error_func)
-        self._global_amax = global_amax
-        # Set by the Triton fast path on its (one-shot) collect; consumed by compute_amax.
+        self._global_amax = global_amax.to(dtype=torch.float32)
+        # Set by collect() after either sweep path; consumed by compute_amax.
```
jenchen13, line 196: why is this still called `_best_amax_fast` if it can be collected from either sweep path? can rename to `self._best_amax`

modelopt/torch/quantization/calib/mse.py
```diff
@@ -235,35 +236,41 @@ def _can_use_triton_fast_path(self, x: torch.Tensor) -> bool:
...
+        self._losses_sum = None
+        # Synchronize before calibrating another weight so reference MSE sweeps do
+        # not overlap. _losses_sum stores one fp32 reduced loss per candidate per
+        # block. With 16-element NVFP4 blocks and bf16 weights, this is roughly
+        # 128 / 16 * (4 / 2) = 16x the calibrated weight size.
+        if x.is_cuda:
+            torch.cuda.synchronize(x.device)
 
     @torch.no_grad()
     def compute_amax(self, verbose: bool = False):
```
jenchen13, line 269: can this be removed now, it only returns `self._best_amax_fast`

modelopt/torch/quantization/model_calib.py
jenchen13: This needs a `if not fp8_scale_sweep:` before it like in the original line 533 in main

modelopt/torch/quantization/model_calib.py
```diff
@@ -475,122 +537,44 @@ def mse_calibrate(
...
-            if module._calibrator is not None and not module._dynamic and hasattr(module, "_amax"):
-                initial_amax = module._amax.clone().detach()
-
-                # Promote standalone NVFP4-static quantizers; grouped siblings
-                # already promoted by _sync_grouped_weight_global_amax above.
-                if module.is_nvfp4_static and not isinstance(module, NVFP4StaticQuantizer):
-                    global_amax = reduce_amax(initial_amax, axis=None)
-                    NVFP4StaticQuantizer.from_tensor_quantizer(module, global_amax=global_amax)
-
-                is_nvfp4_static = isinstance(module, NVFP4StaticQuantizer)
```
jenchen13, line 504: can you check that you didn't remove some important things from here, e.g. `is_nvfp4_static` line and checking `if fp8_scale_sweep and is_nvfp4_static:` in line 522

modelopt/torch/quantization/model_calib.py
realAsma: thanks for catching

modelopt/torch/quantization/model_calib.py
realAsma: The code looks good to me other than the issue you flagged here - https://github.com/NVIDIA/Model-Optimizer/pull/1536/changes#r3313862718

realAsma: please let me know if you find anything.

modelopt/torch/quantization/calib/mse.py
realAsma: this is an override of base-class  (_Calibrator)'s `compute_amax` method to return `self._best_amax_fast` - hence this is needed

modelopt/torch/quantization/calib/mse.py
realAsma: good catch, let me do that

jenchen13 (approved):

PR #1541, opened 2026-05-25
[6/n] Replace skip-softmax calibration formula

kaix-nv:
### What does this PR do?

Type of change: ? 
Replace the old calibration formula from `t = a * exp(b * S) / L` to `t = 1 - exp(-a * (S / (1 - S))^b / seq_k^c`. The old calibration breaks at short context. We can clamp at runtime or add `1 - exp(-...)` wrapper around t because `1 − exp(−c/L^α) \in (0, 1)`  strictly, so we don't need to hardcode where to cutoff.

### Testing

The calibration curve are shown below.

calibration_qwen3_30b_a3b
calibration_qwen3_8b
calibration_llama_3_1_8b

## Summary by CodeRabbit

* **New Features**
  * Dynamic skip-softmax sparse-attention threshold: t = 1 - exp(-a*(S/(1-S))^b / L^c), supporting runtime sparsity changes without recalibration and exposing a calibrated length exponent.

* **Refactor**
  * Consolidated calibrated-threshold logic and removed legacy log-space fitting flow.

* **Bug Fixes**
  * Kernel skip semantics aligned (non-strict comparison); padding-aware skip decision and consistent measurement block sizing.

* **Documentation**
  * Updated calibration docs and threshold-mode tables to match the new model.

* **Tests**
  * Updated and added tests to validate dynamic calibration and new parameters.

### DIFF
--- a/examples/diffusers/README.md
+++ b/examples/diffusers/README.md
@@ -314,7 +314,7 @@ By following these steps, your PEFT LoRA model should be efficiently quantized u
 
 ## Sparse Attention (Skip-Softmax)
 
-Skip-softmax sparse attention skips KV tiles whose attention scores are negligible during the softmax computation, reducing FLOPs without retraining. An exponential model (`scale_factor = a * exp(b * target_sparsity)`) is calibrated once, then the target sparsity can be adjusted at runtime without recalibration.
+Skip-softmax sparse attention skips KV tiles whose attention scores are negligible during the softmax computation, reducing FLOPs without retraining. A dynamic threshold model (`t = 1 - exp(-a * (S/(1-S))^b / L^c)`) is calibrated once, then the target sparsity can be adjusted at runtime without recalibration.
 
 ### Getting Started
 
--- a/examples/diffusers/sparsity/README.md
+++ b/examples/diffusers/sparsity/README.md
@@ -20,11 +20,10 @@ reducing FLOPs without retraining.
 Two modes are supported:
 - **Fixed threshold** — pass a BLASST lambda threshold directly. No calibration
   needed. Good for quick testing and sweeps.
-- **Calibrated threshold** — an exponential model
-  (`scale_factor = a * exp(b * target_sparsity)`) is calibrated once via the
-  Triton calibration kernel, then the target sparsity can be adjusted at runtime
-  without recalibration. Log-space fitting (`fit_logspace=True`) is recommended
-  for diffusion models where scale_factors span many orders of magnitude.
+- **Calibrated threshold** — the dynamic threshold model
+  (`t = 1 - exp(-a * (S/(1-S))^b / L^c)`) is calibrated once via the Triton
+  calibration kernel, then the target sparsity can be adjusted at runtime
+  without recalibration.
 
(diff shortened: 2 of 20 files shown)


rohansjoshi (approved):
LGTM

cjluo-nv (commented):

Formula swap from `t = a*exp(b*S)/L` to a bounded envelope `t = 1 - exp(-a*(S/(1-S))^b / L^c)`, plus a closed-form lstsq fit replacing scipy's `curve_fit`. The math, refactor, and unit-test updates look clean and consistent across `flash_skip_softmax`, `triton_skip_softmax`, the diffusers/LTX kernel backends, the vLLM plugin, and the export/restore plumbing. The new synthetic-data test recovers (a, b, c) to ~2% — much tighter than the old `r_squared > 0.8` assertions.

Flagging for human sign-off:

- **Backward compatibility with existing exported checkpoints.** The PR body leaves the back-compat box blank. The export schema's `formula` string changed and `c` is read with a `1.0` default, but `a` and `b` have different semantics under the new model. An old `softmax_skip` checkpoint loaded under this code would be silently reinterpreted: e.g. an old `(a=7.93, b=8.61)` plugged into `a*(S/(1-S))^b` at `S=0.5` gives `scale=7.93` (still small) but at `S=0.7` gives `scale ≈ 7.93 * (0.7/0.3)^8.61 ≈ huge` → threshold saturates near 1. The vLLM path has a `0 < threshold < 1` guard that falls back to dense, but the flash/triton paths just clip and run with a near-1 threshold (effectively disabling sparsity, no warning). Worth either (a) adding a formula-string check on load that refuses or warns on the old `a * exp(b * target_sparsity)` string, or (b) explicitly noting in the PR body that previously exported checkpoints must be recalibrated.

- **Identifiability of `c` when `L` doesn't vary.** The lstsq design matrix has columns `[1, logit(S), -log(L)]`. If a calibration phase happens to see a single sequence length (e.g. a user-supplied `forward_loop` for diffusion that always uses the same resolution, or a degenerate RULER bin), `log(L)` is constant and collinear with the intercept, so `c` is unidentified — `np.linalg.lstsq` will silently return an arbitrary split between `log(a)` and `c`. The synthetic test passes because it injects three lengths. Consider detecting `np.unique(log_L).size < 2` and either falling back to `c=1.0` (the legacy form) or warning that `c` is not identifiable. The current `_rank` return from lstsq is discarded; checking `_rank < 3` would catch this generically.

- **Decode-phase length variation.** Related to the point above: in `create_decode_calibration_forward_loop`, `seq_k` reported by the kernel is the KV length at the *last* decode step. If RULER bins are tight or `num_decode_tokens` is small, the L spread per phase may be narrow enough that the c-fit is poorly conditioned in practice. Worth a sanity check in the calibrator output (e.g. log L's coefficient of variation, or the lstsq rank).

No design-review concern: this is a formula replacement inside an existing system, not a new abstraction. The 20-file fan-out is almost entirely doc/test/string updates that mechanically follow the math change.

shengliangxu (approved):
LGTM

PR #1550, opened 2026-05-27
Support per expert weight quantizer in TEGroupedMLP

jenchen13:
### What does this PR do?

Type of change: New feature / breaking change

Change TEGroupedMLP to use per-expert weight quantizers by default for MoEs, matching SequentialMLP implementation for quantization. Currently TEGroupedMLP by default has one weight/input quantizer per layer which is different from SequentialMLP's one quantizer per expert implementation.

This is a breaking change for 0.47. If you are using Transformer Engine for PTQ/QAD in Megatron-Core (with `--export-default-te-spec` in Megatron-LM), checkpoints produced with 0.46 in TE will not be compatible with Model-Optimizer version 0.47.

### Usage

```python
```

### Testing
Tested on Nemotron Nano 3 PTQ & QAD that the accuracy & loss curves are identical

## Summary by CodeRabbit

* **New Features**
  * Added `GroupedQuantizer` support (and `AnyQuantizer` typing) for container-style quantizers.
  * Transformer Engine grouped MoE/TEGroupedMLP now uses independent per-expert weight quantizers with per-expert `amax` across calibration, restore, and export.
  * Opt-in grouped-linear `torch.compile` path via `MODELOPT_TEGROUPED_COMPILE_WEIGHT_LOOP=1` (eager remains the default).
* **Bug Fixes**
  * Improved grouped-quantizer calibration completeness checks and per-expert restore behavior (preserve compatible `amax` shapes).
  * Enhanced export/restore logic for grouped quantizers (Megatron/vLLM and sharded expert state).
* **Breaking Changes**
  * Pre-0.47 TEGroupedMLP quantized checkpoints are incompatible due to per-expert `amax` layout changes.
* **Tests**
  * Added coverage for per-expert `amax` shapes, grouped-quantizer behavior, and compiled-vs-eager parity.


modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -27,12 +29,17 @@
 
 from modelopt.torch.quantization.utils import replace_function
 
-from ..nn import QuantModuleRegistry
+from ..nn import QuantModuleRegistry, SequentialQuantizer
```
realAsma: Should we create a new quantizer class say `GroupedQuantizer` which is ` Pytorch [nn.ModuleList](https://docs.pytorch.org/docs/2.12/generated/torch.nn.ModuleList.html) instead of [nn.Sequantial]?

Sequential container assumes it will be a called as  seq(x) which does `seq[1](seq[0](x))`.

ModuleList simply groups quantizers. but does not define an inherent forward path.

modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -137,8 +144,10 @@ def _setup(self):
         # Remove self.weight after setup.
         delattr(self, "weight")
 
-        # TODO: GroupedLinear supports weights split by `num_gemms`, to support quantization
-        # with static parameters beyond per-tensor, we need to support a unique quantizer for each gemm.
+        self._per_expert_weight_quantizer = _per_expert_weight_quantizer_enabled()
+        if self._per_expert_weight_quantizer:
+            for i in range(self.num_gemms):
+                self.add_module(f"weight_quantizer_{i}", copy.deepcopy(self.weight_quantizer))
```
realAsma: We dont need deep copy here - it creates unnecessary  complications.
```suggestion
                self.add_module(f"weight_quantizer_{i}", TensorQuantizer())
```

modelopt/torch/quantization/plugins/transformer_engine.py
realAsma, line 1: Design looks good to me overall, left a comment here - https://github.com/NVIDIA/Model-Optimizer/pull/1550/changes#r3468773327

modelopt/torch/quantization/plugins/transformer_engine.py
jenchen13: this implementation doesn't use SequentialQuantizer for the expert quantizers, it simply checks in row 174 `if isinstance(wq_i, SequentialQuantizer)` to decide how to call the quantizer.

We can add a GroupedQuantizer class if that would make things easier, but the current implementation doesn't really need it IMO

modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -137,8 +144,10 @@ def _setup(self):
         # Remove self.weight after setup.
         delattr(self, "weight")
 
-        # TODO: GroupedLinear supports weights split by `num_gemms`, to support quantization
-        # with static parameters beyond per-tensor, we need to support a unique quantizer for each gemm.
+        self._per_expert_weight_quantizer = _per_expert_weight_quantizer_enabled()
+        if self._per_expert_weight_quantizer:
+            for i in range(self.num_gemms):
+                self.add_module(f"weight_quantizer_{i}", TensorQuantizer())
```
jenchen13: @realAsma previously I did a `copy.deepcopy(self.weight_quantizer)` but you suggested to remove it .. how does this new quantizer inherit the quant format now?

realAsma: We just setup and initialize the tensor quantizers here. (called here https://github.com/NVIDIA/Model-Optimizer/blob/838b2053df5d4ebd7f78cadd68c2c6a404a7937e/modelopt/torch/quantization/conversion.py#L67)

The actual config will be  loaded here - https://github.com/NVIDIA/Model-Optimizer/blob/838b2053df5d4ebd7f78cadd68c2c6a404a7937e/modelopt/torch/quantization/conversion.py#L68

realAsma: BB: Can you add more context if needed here?

realAsma: > _:robot: Bot comment._

Prepared a one-line context comment here explaining that the per-GEMM quantizers keep independent calibration state for each grouped weight.

Validation: `git diff --check` passed. I did not commit or push because this comment did not explicitly approve publishing; ready to push once approved.

modelopt/torch/quantization/plugins/transformer_engine.py
jenchen13: I added a GroupedQuantizer class and changed the TEGroupedMLP implementation to use it

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -1641,3 +1642,71 @@ def convert_to_single_quantizer(model, indx: int = 0):
         ) in original_sequential_quantizers.items():
             for name, sequential_quantizer in sequential_quantizers_list:
                 setattr(parent_module, name, sequential_quantizer)
+
+
+class GroupedQuantizer(nn.ModuleList):
```
realAsma: nit: `SequentialQuantizer` and `GroupedQuantizer` share sever overlapping pieces (such delegation, attribute setter etc). Can we create a shared base-class which implements these overlapping pieces and subclass `SequentialQuantizer` and `GroupedQuantizer` from it to avoid code level redundancy?

realAsma: RB: Can you create a plan on how to do this and share in this thread?

🐝 I’m reviewing the overlapping behavior and will share a concrete refactoring plan here shortly.

Plan:

1. Add a private `_QuantizerContainerBase` mixin before the two containers. It will own the common representative-property reads, property writes broadcast to every member, delegated lifecycle-method calls, attribute-config fan-out, and the common `get_modelopt_state()` shape (using a subclass-provided state-key).
2. Make `SequentialQuantizer` inherit from the mixin plus `nn.Sequential`, and `GroupedQuantizer` from the mixin plus `nn.ModuleList`. Keep PyTorch container initialization in each subclass so module registration/state-dict keys remain unchanged.
3. Keep only semantic differences in the subclasses: accepted member types, sequential chaining vs. grouped representative `forward`, `disable_rotate` delegation for sequential only, delegated-method return behavior, and `convert_to_single_quantizer` for sequential only.
4. Add parameterized CPU tests for the shared contract (representative reads, broadcast writes/methods, scalar/list config application, and exact modelopt-state output), plus regression tests that sequential chaining, grouped first-member forward, nested `SequentialQuantizer` members, and checkpoint/state-dict layout are unchanged.

This keeps the refactor local to `tensor_quantizer.py` and its focused tests; downstream `(SequentialQuantizer, GroupedQuantizer)` handling does not need to change.

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -697,8 +704,13 @@ def _load_from_state_dict(self, state_dict, prefix, *args, **kwargs):
             return super()._load_from_state_dict(filtered_state_dict, prefix, *args, **kwargs)
 
         def _process_quantizer_amax(self, k, v, quantizer_state_dict):
-            assert v.numel() == 1, "TEGroupedLinear only supports per-tensor quantization"
-            quantizer_state_dict[k] = v.view(-1)
+            # Per-expert quantizers have independent checkpoint keys. Preserve their native
+            # scalar, channel, or block shape instead of flattening them through the legacy
+            # single-quantizer path.
+            if re.match(r"weight_quantizer\.\d+\..+_amax$", k):
+                quantizer_state_dict[k] = v
```
realAsma, line 729: Should we do 
```suggestion
                quantizer_state_dict[k] = v.view(-1) if v.numel() == 1 else v
```

modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -137,8 +148,27 @@ def _setup(self):
         # Remove self.weight after setup.
         delattr(self, "weight")
 
-        # TODO: GroupedLinear supports weights split by `num_gemms`, to support quantization
-        # with static parameters beyond per-tensor, we need to support a unique quantizer for each gemm.
+        # Each fused expert gets its own weight quantizer (independent amax). Storing them in a
+        # GroupedQuantizer (an nn.ModuleList) surfaces them as ``weight_quantizer.{i}``, which the
+        # fused-experts name normalizer maps to ``*weight_quantizer`` so the stock configs apply.
+        self.weight_quantizer = GroupedQuantizer(
+            *(copy.deepcopy(self.weight_quantizer) for _ in range(self.num_gemms))
```
realAsma: We dont need this deep copy here. 

```suggestion
            TensorQuantizer() for _ in range(self.num_gemms))
```

The TensorQuantizers are always initialized with default args; they are configured after `_setup` -> so we can safely initialize new TensorQuantizers here. deepcopy seems unwarranted to me.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
realAsma: @jenchen13  could you take a look at this comment?

modelopt/torch/quantization/plugins/transformer_engine.py
realAsma: RB: Can you please explain concisely "The TensorQuantizers are always initialized with default args; they are configured after _setup -> so we can safely initialize new TensorQuantizers here. deepcopy seems unwarranted to me." with code pointers ?

realAsma: > _:robot: Bot comment._

🐝 I’m tracing the initialization and post-`_setup` configuration paths now; I’ll reply here shortly with exact code pointers.

realAsma: > _:robot: Bot comment._

Yes—the key is the conversion order:

1. [`convert_to_quantized_model()`](https://github.com/NVIDIA/Model-Optimizer/blob/8f9a4db327c9ac8c8dfdfa619465d43d3d08ef41/modelopt/torch/quantization/conversion.py#L62-L69) first calls `replace_quant_module()`, then `set_quantizer_by_cfg()`.
2. Conversion invokes the module's [`_setup()`](https://github.com/NVIDIA/Model-Optimizer/blob/8f9a4db327c9ac8c8dfdfa619465d43d3d08ef41/modelopt/torch/opt/dynamic.py#L637-L654). The base setup creates `weight_quantizer` with the standard 8-bit per-tensor descriptor ([`custom.py:113-117`](https://github.com/NVIDIA/Model-Optimizer/blob/8f9a4db327c9ac8c8dfdfa619465d43d3d08ef41/modelopt/torch/quantization/plugins/custom.py#L113-L117)); that descriptor is just `QuantizerAttributeConfig(num_bits=8)` ([`tensor_quant.py:279`](https://github.com/NVIDIA/Model-Optimizer/blob/8f9a4db327c9ac8c8dfdfa619465d43d3d08ef41/modelopt/torch/quantization/tensor_quant.py#L279)), equivalent to `TensorQuantizer()`'s schema defaults.
3. `_QuantTEGroupedLinear._setup()` builds the group immediately after that base setup ([lines 147-156](https://github.com/NVIDIA/Model-Optimizer/blob/8f9a4db327c9ac8c8dfdfa619465d43d3d08ef41/modelopt/torch/quantization/plugins/transformer_engine.py#L147-L156)), still before user config is applied.
4. Afterwards, `set_quantizer_by_cfg()` configures the registered `weight_quantizer.0`, `.1`, … children; the fused-expert name normalization makes them match stock `*weight_quantizer` rules ([`conversion.py:319-355`](https://github.com/NVIDIA/Model-Optimizer/blob/8f9a4db327c9ac8c8dfdfa619465d43d3d08ef41/modelopt/torch/quantization/conversion.py#L319-L355)).

So `deepcopy` is only cloning a pristine default quantizer (including an empty/default calibrator). Fresh `TensorQuantizer()` instances give each expert independent state and are configured identically in the next phase. `deepcopy` would matter only if setup had already attached non-default config/state, which this path does not.

modelopt/torch/quantization/model_calib.py
```diff
@@ -205,7 +211,7 @@ def _has_expert_parallelism(module: nn.Module) -> bool:
 
 
 def _iter_leaf_quantizers(quantizer):
-    if isinstance(quantizer, SequentialQuantizer):
+    if isinstance(quantizer, (SequentialQuantizer, GroupedQuantizer)):
```
realAsma, line 219: nit:
Create QuantizerContainerBase base class and use that here to avoid explicitly listing like this.

see https://github.com/NVIDIA/Model-Optimizer/pull/1550/changes#r3553373837

modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -137,8 +148,27 @@ def _setup(self):
...
+
+        # Compile only the per-expert quantizer loop. The surrounding TE grouped GEMM remains
+        # eager, and the opt-in flag leaves the default execution path unchanged.
+        if os.getenv(_COMPILE_TEGROUPED_WEIGHT_LOOP_ENV, "0") == "1":
+            quantizers = tuple(self.weight_quantizer)
+
+            def quantize_weights(*weights):
+                return tuple(quantizer(weight) for quantizer, weight in zip(quantizers, weights))
+
+            self._compiled_weight_quantizer_loop = torch.compile(
```
realAsma, line 166: nice!

I have not tried torch.compile + quantize before. 

Does FP4 quantizer + compile work? Does it give any perf improvements?

In my understanding, torch.compile in case of graph-breaks fall backs to using broken graphs - we might not see any performance benefits in this case.

realAsma: BB: Can you quantize a toy model (such as a single `nn.Linear`) and check if torch.compile traces the TensorQuantizer correctly?  Are there any breaks? Use NVFP4_DEFAULT_CFG to quantize. Please report your findings here and notify me in slack.

realAsma: > _:robot: Bot comment._

🐝 I’m on it. I’ll run a focused `NVFP4_DEFAULT_CFG` toy-model compile probe, check Dynamo’s graph-break diagnostics, and report the result here.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -757,6 +762,234 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+
+@pytest.mark.parametrize("compile_enabled", [False, True])
+def test_te_grouped_compiled_weight_quantizer_loop(
+    distributed_setup_size_1, monkeypatch, compile_enabled
+):
+    """The opt-in flag controls compilation and preserves per-expert backward."""
+    compile_kwargs = []
+    compiled_calls = []
+
+    def fake_compile(fn, **kwargs):
```
realAsma, line 791: Why are we doing fake_compile here? 

Can we do real torch.compile to make sure unittests actually test with torch.compile? We can optionally patch the original torch.compile to count/trace that it was called.

realAsma: RB: Here is an example of unittest that is not fully correct - it is monkey patching, testing only the interface - not the actual end to end things. You should flag such things in your review in future.

modelopt/torch/quantization/plugins/transformer_engine.py
realAsma: > _:robot: Bot comment._

Tested at PR head `8f9a4db3` with a CUDA `nn.Linear(16, 16)` quantized using `NVFP4_DEFAULT_CFG`.

- Both the input and weight `TensorQuantizer` trace as **one Dynamo graph with zero graph breaks**.
- `fullgraph=True` succeeds, and compiled output exactly matches eager output (`max_abs_diff=0`).
- I also tested the tuple/closure pattern used here (`tuple(quantizer(weight) for ...)`) with the default Inductor backend: one graph, zero breaks, exact eager parity across repeated calls.

So the quantizer-only compiled loop in this PR traces cleanly. One separate caveat: compiling the *entire* converted `QuantLinear` currently raises an `InternalTorchDynamoError` while Dynamo inspects ModelOpt's dynamic `weight` callback (`_FoldedCallback` missing `_callbacks`), before reaching the quantizer. That does not affect this PR's narrower quantizer-loop compile scope.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
realAsma: > _:robot: Bot comment._

🐝 Acknowledged — I’ll carry this forward as a review criterion for tests that mock the behavior they are meant to validate.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -757,6 +762,234 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+                assert torch.allclose(te_amax, seq_amax, atol=1e-5, rtol=1e-5), (
+                    f"TEGrouped expert {i} amax != Sequential expert {i} amax for {linear_name}"
+                )
+                expert_amaxes.append(te_amax.reshape(-1)[0])
+
+            stacked = torch.stack(expert_amaxes)
+            if (stacked.max() - stacked.min()).item() > 1e-5:
+                saw_per_expert_divergence = True
+
+    assert saw_per_expert_divergence, (
```
realAsma: This is not a good test IMO. This can fail if magically all the expert weights were initialized the same.

realAsma: RB: Do you agree with my comment? Wdyt?

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
realAsma: > _:robot: Bot comment._

Agreed. I’ve recorded this as a durable review criterion. Here, `fake_compile` verifies the opt-in flag, compile kwargs, call-through, and backward on the original Python function, but it cannot prove that Inductor captures or executes the quantizer loop. In future reviews I’ll flag this distinction: mocked tests are useful for focused wiring coverage, while the claimed compiled/backend behavior needs a real end-to-end execution test. If tracing is needed, the test should wrap and delegate to the original `torch.compile` rather than replace it.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
realAsma: > _:robot: Bot comment._

🐝 Looking into this now.

realAsma: > _:robot: Bot comment._

Yes, I agree with the underlying concern. The fixed seed makes this deterministic today, so it is not conventionally flaky, but `saw_per_expert_divergence` tests an incidental property of the fixture rather than the independence contract: a correct implementation would fail if the experts happened to produce equal amaxes. I would initialize each expert's weights to deliberately distinct known magnitudes, then assert the corresponding grouped/sequential amax values (and, if useful, that the quantizer objects are distinct). That makes divergence guaranteed by the test setup and the failure causal.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
realAsma: BB: Can you create a PR to update the ModelOpt coding contribution guidelines with this https://github.com/NVIDIA/Model-Optimizer/pull/1550/changes#r3553521349

realAsma: > _:robot: Bot comment._

🐝 On it — I’ll turn that guidance into a focused contribution-guidelines PR and report back here with the link.

realAsma: > _:robot: Bot comment._

Done — opened #1952 to add this to `CONTRIBUTING.md`: mocks may cover focused wiring, but tests claiming backend/runtime behavior must run the real implementation; call tracing should wrap and delegate to the original function. Commit `1804ca12bb`; `git diff --check` and all pre-commit checks for `CONTRIBUTING.md` pass.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
jenchen13: This seems like a large change for 3 classes which aren't that similar. SequentialQuantizer and GroupedQuantizer are compositions of TensorQuantizer (they own one or more TensorQuantizers), not inheritors. If your concern is listing out all three classes when type checking, I can create a custom type that is a tuple union of the three types.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
jenchen13: the fake compile is to test that the compilation is actually called. compilation is hardware dependent and can be flaky in CI, so testing the plumbing through monkey patching is what this test does. to test true compilation, that can be done in a separate test but it might be flaky in CI and might not be enabled by default

jenchen13:
@coderabbitai resume

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1119,48 +1107,82 @@ def _grouped_mlp_slicing(self, module, prefix, parallel_config=None):
...
+                module.weight = getattr(module, weight_key)
+                if per_expert_wq:
+                    assert isinstance(grouped_wq, GroupedQuantizer)
+                    module.weight_quantizer = grouped_wq[min(local_id, len(grouped_wq) - 1)]
+                    # Dynamic-NVFP4 per-expert quantizers carry no stored amax, but
+                    # weight_scale_2 derivation asserts one. Max-calibration weight amax
+                    # is exactly max(|W|), so compute it from this expert's weight.
+                    _wq = module.weight_quantizer
+                    if getattr(_wq, "_amax", None) is None and getattr(_wq, "is_enabled", False):
+                        _wq.amax = module.weight.detach().abs().max().float()
```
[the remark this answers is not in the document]
jenchen13: i think this is fine because we usually export in a single job

hychiang-git (approved):
Looks good to me. Approve to unblock.

modelopt/torch/quantization/plugins/megatron.py
jenchen13: the static dict should only be flattened if it is not a per-expert quantizer, which is why this plain `= v` is inside the `if re.match..` statement

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
realAsma: If your concern is listing out all three classes when type checking -> No this was not my concern. My concern was that we were duplicating many methods across SequentialQuantizer and GroupedQuantizer.

realAsma: > SequentialQuantizer and GroupedQuantizer are compositions of TensorQuantizer (they own one or more TensorQuantizers), not inheritors

I did not suggest this. I suggested to create a common base class for SequentialQuantizer and GroupedQuantizer `_QuantizerContainerBase`

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -1840,3 +1841,67 @@ def convert_to_single_quantizer(model, indx: int = 0):
...
+    """
+
+    _delegated_properties = ["fake_quant", "is_enabled", "amax"]
+    _delegated_methods = [
+        "reset_amax",
+        "disable",
+        "enable",
+        "load_calib_amax",
+        "load_calib_bias",
+    ]
```
realAsma: @jenchen13  these are duplicated between Sequential Quantizer and this class which could be avoided by the common base class I mentioned in https://github.com/NVIDIA/Model-Optimizer/pull/1550/changes#r3553373837

modelopt/torch/quantization/plugins/transformer_engine.py
realAsma: @jenchen13 can you please address this?

CHANGELOG.rst
kevalmorabia97, line 1: This should go in 0.47 Changelog section

modelopt/torch/quantization/plugins/transformer_engine.py
jenchen13: I can to TensorQuantizer(self.default_quant_desc_weight) to preserve the axis information. otherwise the axis information will be dropped

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
realAsma: RB:
Jenny says "for the other comment about base classes, it really would be an optimization and is not necessary right now. Changing SequentialQuantizer will also require extensive tests to make sure it didn't break"

I do not agree with this. I am not asking to change the interface  of SequentialQuantizer -> but to consolidate SequentialQuantizer and GroupedQuantizer to avoid coercion redundancy.

realAsma: RB: Can you propose a plan how to do this consolidation without breaking SequentialQuantizer.

🐝 Acknowledged — I’ll outline a behavior-preserving consolidation plan.

Yes—this can be a small behavior-preserving consolidation, not an interface change.

1. Add a private `_QuantizerContainerBase` mixin (before the PyTorch container in the MRO) for the shared delegated-property read, broadcast `__setattr__`, lifecycle fan-out, and `set_from_attribute_config` logic. Let a small hook preserve each class’s existing delegated-method return shape.
2. Keep `SequentialQuantizer(nn.Sequential)` initialization and forward chaining untouched; it retains `disable_rotate`, its last-member return behavior, `get_modelopt_state`, and `convert_to_single_quantizer`.
3. Keep `GroupedQuantizer(nn.ModuleList)` initialization, accepted member types, and representative `forward` untouched; it retains its list return from broadcast lifecycle calls.
4. Add parameterized tests for the shared delegation contract, plus regressions for SequentialQuantizer chaining, module/state-dict keys, checkpoint state, nested sequential members, and GroupedQuantizer indexing/representative-forward behavior.

That removes the duplicated container mechanics while explicitly locking down SequentialQuantizer’s current semantics.

🐝 I’m looking into the consolidation plan and the `SequentialQuantizer` compatibility constraints.

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
jenchen13: I implemented the QuantizerContainerBase

realAsma: Can we move these attributes to the `QuantizerContainerBase`? Currently the same properties are repeated across Sequential and Grouped Quantizer. 
In addition `"disable_rotate",` is missing from `GroupedQuantizer`. We can avoid accidentally dropping such things by moving them to their unified shared class and not repeat here.

realAsma (approved):
Looks great! Left a comment here - https://github.com/NVIDIA/Model-Optimizer/pull/1550/changes#r3730695649

Could you please address that before merging?

CHANGELOG.rst
jenchen13: thanks for catching that!

jenchen13:
/ok to test [76e4924](https://github.com/NVIDIA/Model-Optimizer/pull/1550/commits/76e492493e00510ab1bb0136487c0ee935ebd02b)

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -761,6 +767,483 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+        assert not compiled_calls
+    assert all(
+        torch.isfinite(getattr(module, f"weight{i}").grad).all()
+        for module in grouped_modules
+        for i in range(module.num_gemms)
+    )
+    destroy_model_parallel()
+
+
+@pytest.mark.timeout(90)  # real torch.compile: cap runaway inductor recompiles without flaking CI
```
kevalmorabia97: default timeout is 120s. We dont need to downgrade to 90s here

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -761,6 +767,483 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+            assert (stacked.max() - stacked.min()).item() > 1e-4, (
+                f"{linear_name}: per-expert amax did not diverge despite distinct set magnitudes"
+            )
+
+
+@pytest.mark.parametrize("quant_cfg", [mtq.FP8_DEFAULT_CFG, mtq.NVFP4_DEFAULT_CFG])
+def test_te_grouped_vs_sequential_default_amax(dist_workers_size_4, quant_cfg):
+    dist_workers_size_4.run(
+        partial(_test_te_grouped_vs_sequential_default_amax_helper, 1, 2, quant_cfg)
+    )
```
kevalmorabia97, line 1079: Why does this test need to have 4-gpus? Can we not run in 2-gpu?

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -761,6 +767,483 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+    # (e.g. 128 experts in Nemotron Nano). Just sanity-check both errors are finite.
+    assert err_te > 0 and err_seq > 0
+    assert math.isfinite(err_te) and math.isfinite(err_seq)
+
+
+@pytest.mark.parametrize("quant_cfg", [mtq.FP8_DEFAULT_CFG, mtq.NVFP4_DEFAULT_CFG])
+def test_te_grouped_vs_sequential_default_loss(dist_workers_size_4, quant_cfg):
+    dist_workers_size_4.run(
+        partial(_test_te_grouped_vs_sequential_default_loss_helper, 1, 2, quant_cfg)
+    )
```
kevalmorabia97: same comment on 4-gpu requirement

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
kevalmorabia97, line 1: Most new tests have 2 or 4 gpu requirement hence they would not be run per PR. 2-gpu ones will run nightly or on-demand. Can you trigger manually so we can make sure 2-gpu tests pass before merging this.

You can specify your branch name and trigger workflow here: https://github.com/NVIDIA/Model-Optimizer/actions/workflows/gpu_tests.yml

jenchen13: I have moved tests that don't require EP to EP=1 and the one that does require EP to EP=2

jenchen13:
/ok to test [07f2e41](https://github.com/NVIDIA/Model-Optimizer/pull/1550/commits/07f2e41cdcbc5438bdcd377277038ce7f9c76d98)

ChenhanYu (approved):

PR #1556, opened 2026-05-28
Fix puzzletron tests and make FFN pruning check robust

kevalmorabia97:
### What does this PR do?

Type of change: Bug fix + test robustness

Fixes puzzletron GPU tests and adds one mbridge plugin fix. Two-part change:

**Part 1: Test refactor — drop non-portable FFN check, relax lm_loss tolerance**

The previous puzzletron test compared exact-equality FFN pruning scores per layer, which started failing in CI after the transformers 5.7 → 5.9 bump. I tried multiple progressively-looser variants (set-membership in top-K, fp32 forward with TF32 disabled, deterministic algorithms) before concluding the check is **not portable across GPU vendors on this scale of test model**:

- The small test models (hidden=256) have FFN channel scores of similar magnitude.
- cuBLAS picks different fused-kernel paths on different GPU architectures (Ada vs Blackwell).
- Even at fp32 with TF32 off, the resulting numerical drift is enough to reshuffle the channel importance ranking — expected top-K channels land at ranks 200–500 on the other GPU.

Changes:
- Refactor `_assert_*` helpers to `_check_*` that return error lists; the multiprocess job collects all failures and fails once at the end (instead of fail-fast). One CI run now surfaces every mismatch.
- Drop `_check_score_pruning_activations` and `EXPECTED_FFN_PRUNING_VALUES`. Keep a single structural check that the `rank_N.pth` file is produced.
- Widen `_check_lm_loss` tolerance to **0.15** (covers the empirical ~0.1 cross-GPU lm_loss drift). Still catches order-of-magnitude failures (NaN, exploded loss). Hybrid mamba+MoE models (Nemotron-3-Nano) remain excluded — their run-to-run noise is ~0.06 even on the same GPU.
- Update `EXPECTED_LM_LOSS` for current behavior.

Structural checks (replacement library, mip solutions, subblock stats) and `lm_loss` together still validate that the pipeline runs end-to-end and produces a sensible model.

**Part 2: mbridge plugin — register modelopt. as an allowed target prefix**

Megatron-Bridge gates Hydra-style class instantiation behind an allowed-prefix registry. Register `"modelopt."` so modelopt classes can be referenced via `_target_` in heterogeneous puzzletron configs without rejection.

### Testing

Verified locally on RTX 6000 Ada across transformers `4.57.0`, `5.6.0`, `5.9.0` and in 1-GPU + 2-GPU modes — all 9 puzzletron model tests pass cleanly (Qwen3-VL has known pre-existing cross-major-version lm_loss drift on 4.57). CI runs transformers 5.9.

- Did you write any new necessary tests?: N/A (test + plugin change)


Separius (commented):
LGTM, I had the same issue and this is a reasonable fix (maybe it makes sense to also check the most important channel is also correctly identified, not just the least important one, being in topK)

kevalmorabia97:
> LGTM, I had the same issue and this is a reasonable fix (maybe it makes sense to also check the most important channel is also correctly identified, not just the least important one, being in topK)

Good suggestion. Added

Separius (approved):

PR #1558, opened 2026-05-28
[5924759] Fix fp16 ONNX INT8 entropy calibration on numpy >= 2.0

ajrasane:
### What does this PR do?

Type of change: Bug fix

INT8 entropy calibration of fp16 ONNX models (e.g. ConvNext / EfficientViT / YOLOv8 backbones quantized via `python -m modelopt.onnx.quantization --quantize_mode=int8`) used to fail during histogram collection with:

```
ValueError: Too many bins for data range. Cannot create 128 finite-sized bins.
```

`_collect_value` in `modelopt/onnx/quantization/ort_patching.py` derives `threshold = max(abs(min), abs(max))` from the activation tensor and passes `range=(-threshold, threshold)` to `np.histogram(...)`. When the model is fp16 and a calibrated activation has a small range (≲ 1e-5), both endpoints inherit fp16 dtype. Under numpy 2.0's NEP-50 strict promotion, the resulting fp16 `linspace` collapses consecutive 128-bin edges to the same value and numpy refuses to build the histogram. numpy 1.x silently used higher-precision intermediate dtype, masking the issue.

The fix casts the range endpoints to Python `float` so numpy computes bin edges in float64 regardless of input dtype. Applied at both call sites: `_collect_value` and the single-node variant `_collect_value_histogram_collector_single_node_calibration`.

### Usage

```bash
# The affected workflow — INT8 entropy calibration of any fp16 ONNX model:
    --quantize_mode=int8 \
    --onnx_path=model.fp16.onnx \
    --calibration_data_path=calib.npy
```

No API change.

### Testing

- Added `test_collect_value_fp16_narrow_range` in `tests/gpu/onnx/test_ort_patching.py` that calls `_collect_value` with a fp16 tensor (mostly zeros + one ~1e-5 value) and asserts the histogram is built without raising and all bin edges are distinct. Fails on the buggy code, passes after the fix.
- Reproduced the original failure on numpy 2.2.6 before the fix.
- Full `tests/gpu/onnx/test_ort_patching.py` suite (31 tests) passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed INT8 entropy calibration for fp16 ONNX models on NumPy >= 2.0 so histogram bin-edge computation no longer fails for extremely small fp16 activation ranges.

* **Tests**
  * Added unit test validating histogram bin-edge correctness and count integrity during INT8 calibration with narrow fp16 activation ranges.


cjluo-nv (approved):

Small, well-scoped bug fix for fp16 ONNX INT8 entropy calibration on numpy >= 2.0. The root-cause analysis (NEP-50 strict promotion causing fp16 linspace bin-edge collapse) is correct, the fix (casting `threshold` to Python `float` so `np.histogram` computes edges in float64) is minimal and applied at both relevant call sites in `ort_patching.py`. The new `test_collect_value_fp16_narrow_range` reproduces the failure mode and asserts no bin-edge collapse. CHANGELOG is updated. No API change, no licensing impact.

modelopt/onnx/quantization/ort_patching.py
```diff
@@ -1111,8 +1115,12 @@ def _collect_value_histogram_collector_single_node_calibration(histogram_collect
                 threshold,
             )
         else:
+            # Cast range endpoints to Python float so numpy computes bin edges in
+            # float64. A fp16 threshold here can underflow the 128-bin linspace
+            # and trip "Too many bins for data range" on numpy >= 2.0.
+            range_max = float(threshold)
             hist, hist_edges = np.histogram(
-                data_arr, histogram_collector.num_bins, range=(-threshold, threshold)
+                data_arr, histogram_collector.num_bins, range=(-range_max, range_max)
```
gcunhase, line 1123: This code is a duplicate from `_collect_value()`. Can we merge this into a separate util function?

gcunhase (approved):
LGTM after minor comment is addressed.

tests/gpu/onnx/test_ort_patching.py
```diff
@@ -153,6 +153,21 @@ def test_collect_value(self, mock_histogram_collector, sample_tensor_data):
...
 
+    def test_collect_value_fp16_narrow_range(self, mock_histogram_collector):
+        # fp16 activations with a small range (threshold ~1e-5) used to raise
+        # "Too many bins for data range" on numpy >= 2.0, because the fp16 range
+        # produced a fp16 linspace where consecutive bin edges rounded together.
+        activations = np.zeros(1000, dtype=np.float16)
+        activations[0] = np.float16(1e-5)
+        name_to_arr = {"narrow_fp16_tensor": [activations]}
+
+        _collect_value(mock_histogram_collector, name_to_arr)
```
vishalpandya1990, line 164: Does it requiring test for _collect_value_histogram_collector_single_node_calibration as well?

vishalpandya1990 (approved):

PR #1560, opened 2026-05-28
[OMNIML-3994] Make sure all weight quantizers have _amax

sychen52:
### What does this PR do?
  - max_calibrate now always runs weight_only_quantize before the optional forward_loop, so every weight quantizer has _amax populated regardless of MoE routing.
  - awq_lite search loop explicitly deletes _amax around the alpha sweep to keep using the dynamic-amax path; no restore needed (postprocess overwrites for enabled modules, disabled modules early-return).
  - Removed three band-aids:
    - _bootstrap_uncalibrated_weight_quantizers (mse_calibrate)
    - per-module max_calibrate fallback in awq_lite postprocess
    - _ensure_weight_quantizer_calibrated + helpers (export/quant_utils)
  - Renamed test_bootstrap_populates_dead_expert_quantizers -> test_max_calibrate_populates_dead_expert_quantizers; deleted the GPU lazy-calibration test that no longer has a behavior to exercise.
Type of change:  refactor

### Usage
same as before.

### Testing
run unittests locally

## Summary by CodeRabbit

* **Refactor**
  * Simplified the weight quantization calibration process by consolidating logic and removing redundant internal functions.

* **Tests**
  * Removed a redundant quantization format-specific export test.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1560?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -265,11 +265,7 @@ def max_calibrate(
     """
     # Always run weight calibration on the weight tensor directly so every weight
     # quantizer gets ``_amax``, regardless of MoE routing. Downstream algorithms
-    # (MSE, AWQ, export) then no longer need to patch in a missing ``_amax``. Weight
-    # quantizers also exercised by ``forward_loop`` see the same weight twice;
-    # MaxCalibrator's reduction is idempotent (max of identical values), so the extra
-    # pass is a no-op for their stats. Disabled weight quantizers (e.g. AWQ's, which
-    # call this with weight quantizers disabled) are skipped by ``weight_only_quantize``.
+    # (MSE, AWQ, export) then no longer need to patch in a missing ``_amax``.
     enable_stats_collection(model)
     weight_only_quantize(model)
     if forward_loop is not None:
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -111,47 +111,6 @@ def _collect_weight_stats(quantizer: nn.Module, weight: torch.Tensor) -> None:
 
 
 @torch.no_grad()
-def _bootstrap_uncalibrated_weight_quantizers(model: nn.Module) -> int:
-    """Re-run weight calibration on the weight tensor for quantizers missing ``_amax``.
-
-    Covers MoE experts that ``max_calibrate`` skipped (no routed tokens) so MSE
-    doesn't drop them and break the gate==up ``weight_scale_2`` export invariant.
-    Activation quantizers on those modules remain uncalibrated; emits a warning.
-    """
-    name_to_module = dict(model.named_modules())
-    n = 0
-    for module in name_to_module.values():
-        if not isinstance(module, QuantModule):
-            continue
-        with enable_weight_access_and_writeback(module, model, name_to_module):
-            for weight, q in module.iter_weights_for_calibration():
-                if (
-                    not isinstance(q, TensorQuantizer)
-                    or q._disabled
-                    or q._dynamic
-                    or q._calibrator is None
-                ):
-                    continue
-                if weight.is_meta:
-                    continue
-                amax = q.amax
-                if amax is not None and (amax.is_meta or not torch.all(amax == 0)):
-                    continue
-                _run_and_load_max_stats(q, partial(_collect_weight_stats, weight=weight))
-                if hasattr(q._calibrator, "reset"):
-                    q._calibrator.reset()
-                n += 1
-    if n > 0:
-        warnings.warn(
-            f"Bootstrapped {n} weight quantizer(s) with no routed calibration tokens; "
-            f"their activation quantizers (if any) remain uncalibrated. "
-            f"Increase calib size/seq len to activate all experts.",
-            stacklevel=2,
-        )
-    return n
-
-
-@torch.no_grad()
 def _sync_grouped_weight_global_amax(model: nn.Module) -> int:
     """Unify NVFP4 ``global_amax`` across Q/K/V and gate/up sibling weight quantizers.
 
@@ -304,7 +263,18 @@ def max_calibrate(
     See :class:`MaxCalibConfig <modelopt.torch.quantization.config.MaxCalibConfig>` for
     details on the remaining arguments.
     """
-    _run_and_load_max_stats(model, forward_loop)
+    # Always run weight calibration on the weight tensor directly so every weight
+    # quantizer gets ``_amax``, regardless of MoE routing. Downstream algorithms
+    # (MSE, AWQ, export) then no longer need to patch in a missing ``_amax``. Weight
+    # quantizers also exercised by ``forward_loop`` see the same weight twice;
+    # MaxCalibrator's reduction is idempotent (max of identical values), so the extra
+    # pass is a no-op for their stats. Disabled weight quantizers (e.g. AWQ's, which
+    # call this with weight quantizers disabled) are skipped by ``weight_only_quantize``.
+    enable_stats_collection(model)
+    weight_only_quantize(model)
+    if forward_loop is not None:
+        forward_loop(model)
+    finish_stats_collection(model)
 
     # Sync quantizer amax across local experts within each rank (for SequentialMLP)
     for name, module in model.named_modules():
@@ -313,8 +283,6 @@ def max_calibrate(
 
     # Fail fast on NVFP4 static-block with TP>1 (sharded_state_dict treats _amax as replicated).
     _check_nvfp4_static_tp_supported(model)
-
-    _bootstrap_uncalibrated_weight_quantizers(model)
 
     # Promote eligible static-block NVFP4 weight quantizers to NVFP4StaticQuantizer so
     # the static blockwise fake-quant path is used in forward and export picks up the
--- a/modelopt/torch/export/quant_utils.py
+++ b/modelopt/torch/export/quant_utils.py
@@ -41,7 +41,6 @@ from modelopt.torch.quantization.qtensor import (
 from modelopt.torch.quantization.utils import (
     QuantizerAttrNames,
     quantizer_attr_names,
-    reduce_block_amax,
     representative_weight_quantizer,
     weight_attr_names,
 )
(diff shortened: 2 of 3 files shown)


cjluo-nv (commented):

Nice simplification — removing the three band-aid paths in favor of populating `_amax` once in `max_calibrate` is the right direction, and the test rename + assertion update on `_TinyMoEModel` look correct. Flagging for human sign-off on a few points:

- **AWQ lite behavior change is the most fragile piece and lacks a dedicated test.** The new `delattr(self.weight_quantizer, "_amax")` inside `awq_lite`'s `forward` (model_calib.py L1209-1215) is the load-bearing change that keeps the per-alpha dynamic-amax path working now that `weight_only_quantize` populates `_amax` up front. The PR rationale ("postprocess overwrites for enabled modules, disabled modules early-return") is plausible — `apply_pre_quant_scale_and_smooth` → `_apply_weight_pre_quant_scale` repopulates for enabled modules, and the new `assert module.awq_lite.is_enabled` plus inline `max_calibrate` fallback handles disabled ones — but there is no new test exercising AWQ with dead/uncalibrated experts. The existing renamed test only covers plain `max_calibrate`. Worth either adding an AWQ-MoE-with-dead-experts test or explicitly listing which existing GPU test exercises this combination.

- **Export-time safety net removed without a deprecation path.** `_ensure_weight_quantizer_calibrated` previously warned and re-derived `_amax` from the weight if export saw an uncalibrated NVFP4 quantizer. With this PR, the same situation will dereference `weight_quantizer._amax.float() / 448.0` (quant_utils.py L292, L324) and crash with `AttributeError: 'NoneType' object has no attribute 'float'`. Anyone loading a partial checkpoint and exporting without re-running `mtq.quantize`/`max_calibrate` will hit this. The PR claims "backward compatible" — that's true for the documented happy path, but the failure mode for misuse changes from "warn + recover" to "hard crash". Worth confirming this is intentional and ideally raising a clearer error than `AttributeError` in `get_weight_scaling_factor[_2]` when `_amax is None`.

- **Deleted `test_export_nvfp4_static_weight_dynamic_vs_static_match` was checking two things**, not just lazy calibration: (a) dynamic vs static NVFP4 export produce matching weight/scales, and (b) lazy fill from weights when `_amax` is cleared. (b) is dead, agreed — but (a) is still a useful invariant. Consider keeping a trimmed version that exercises (a) without the manual `delattr` setup.

- **`weight_only_quantize` now runs on every `max_calibrate` call**, including the recursive ones in `_apply_weight_pre_quant_scale`, `smoothquant.postprocess`, the awq_lite disabled-fallback, and `awq` SequentialQuantizer recalibration. For these, the lambda already calibrates from `module.weight`, so we now do that work twice on the same tensor. Functionally a no-op (idempotent), but the PR doesn't note the duplicated work and large-model calibration time may tick up slightly.

None of these are blockers — the structural cleanup is good and the diff is net -200 lines. Asking a human to confirm the AWQ test coverage and the export-error UX before approval.

---

- `modelopt/torch/export/quant_utils.py:292` — > _Bot comment._

With `_ensure_weight_quantizer_calibrated` gone, this `weight_quantizer._amax.float() / 448.0` (and the analogous one in `get_weight_scaling_factor_2`) will raise `AttributeError: 'NoneType' object has no attribute 'float'` instead of warning + recomputing if a user exports a checkpoint that wasn't fully calibrated. Worth at least adding an explicit `if weight_quantizer._amax is None: raise RuntimeError("Weight quantizer ... has no _amax; run max_calibrate before export")` so the failure points at the cause, not at a generic NoneType deref.
- `tests/gpu/torch/export/test_export_weight_gpu.py:124` — > _Bot comment._

The deleted test was checking two things: (a) dynamic vs static NVFP4 export produces matching weight + scales, and (b) lazy fill from `_ensure_weight_quantizer_calibrated`. (b) is dead, agreed. But (a) — that the static-quantizer export path matches the dynamic one when both are properly calibrated — is still a useful invariant and isn't covered elsewhere as far as I can see. Consider keeping a trimmed version of the test that drops the `delattr` setup but still asserts the dynamic-vs-static export equivalence.

modelopt/torch/quantization/model_calib.py
```diff
@@ -1248,6 +1206,13 @@ def forward(self, input, *args, **kwargs):
                     max_calibrate(self.input_quantizer, lambda quantizer: quantizer(input), False)
             return out_actual
 
+        # Remove the stored ``_amax`` (set by ``weight_only_quantize`` in
+        # ``max_calibrate`` from the unscaled weight) so the dynamic path
+        # inside ``_get_amax`` recomputes ``max(|weight * pre_quant_scale|)``
+        # per alpha.
+        if hasattr(self.weight_quantizer, "_amax"):
+            delattr(self.weight_quantizer, "_amax")
```
cjluo-nv: > _Bot comment._

This `delattr` is the load-bearing change that keeps awq_lite working now that `_amax` is populated up front by `weight_only_quantize`. Two requests:

1. Could you point at (or add) an AWQ test that exercises the dead-expert case? The renamed `test_max_calibrate_populates_dead_expert_quantizers` only covers plain max_calibrate; if `awq_lite.forward` is ever changed to skip this delete (e.g. someone refactors and forgets the cleanup), per-alpha amax will be silently wrong and the only signal is final accuracy regression.
2. Consider deleting once outside the loop with a brief comment that for cache_mode the delete is unnecessary because `weight_quantizer` is `disable()`'d (the early-return above), to make the invariant easier to follow.

modelopt/torch/quantization/model_calib.py
```diff
@@ -252,9 +211,11 @@ def max_calibrate(
     details on the remaining arguments.
     """
     enable_stats_collection(model)
-    if forward_loop is None:
-        weight_only_quantize(model)
-    else:
+    # Always run weight calibration on the weight tensor directly. This
+    # will avoid uncalibrated MoE experts to have ``_amax = None``. Now
+    # downstream algorithm (MSE, AWQ, export) do not need to patch it.
+    weight_only_quantize(model)
```
cjluo-nv, line 270: > _Bot comment._

Worth a one-line note here that running `weight_only_quantize` *and* the user-provided `forward_loop` means weight quantizers that the forward_loop already exercises will get their amax accumulated from the weight tensor twice. It's idempotent (max calibration on the same weight tensor is the same answer), so functionally fine — just unobvious for readers, and it means the previous `if forward_loop is None` short-circuit was a perf optimization for the common case.

sychen52: Done.

cjluo-nv (commented):

Re-review: the structural cleanup is the right direction and the unreachable `else` in `awq_lite.postprocess` plus the docstring nit on `max_calibrate` look addressed (`:speech_balloon:` author replied "Done" on the duplicated-calibration note, and the `else` arm has been replaced with an `assert module.awq_lite.is_enabled`). However, several substantive points from cjluo-nv's previous human review still appear unresolved — flagging for human sign-off:

- **AWQ dead-expert test still missing.** The `delattr(self.weight_quantizer, "_amax")` inside `awq_lite.forward` (model_calib.py L1209-1215) is the load-bearing change that keeps the per-alpha dynamic-amax path working now that `weight_only_quantize` populates `_amax` up front. The renamed `test_max_calibrate_populates_dead_expert_quantizers` only exercises plain `max_calibrate`, not `awq_lite` with dead/uncalibrated experts. If anyone refactors `forward` and forgets the `delattr`, per-alpha amax will be silently wrong with no test to catch it. Author hasn't responded to this point.

- **Export-time `AttributeError` UX not fixed.** With `_ensure_weight_quantizer_calibrated` gone, `weight_quantizer._amax.float() / 448.0` in `get_weight_scaling_factor[_2]` (quant_utils.py:288, :324 in the post-diff file) will now raise `AttributeError: 'NoneType' object has no attribute 'float'` instead of the previous warn+recover for a checkpoint exported without re-running `max_calibrate`. The previous review asked for an explicit `if weight_quantizer._amax is None: raise RuntimeError("... run max_calibrate before export")` so the failure points at the cause rather than a generic NoneType deref. Not addressed in the current diff.

- **Dropped dynamic-vs-static NVFP4 export equivalence assertion.** The deleted `test_export_nvfp4_static_weight_dynamic_vs_static_match` was checking two things: (a) dynamic vs static NVFP4 export produce matching weight/scales; (b) lazy fill from weights when `_amax` is cleared. (b) is dead code now, agreed — but (a) is still a useful invariant and isn't covered elsewhere as far as I can see. Previous review suggested keeping a trimmed version that drops the `delattr` setup but still asserts (a). Not addressed.

Net direction is good (-200 lines, simpler invariant), but please confirm test coverage for AWQ dead experts and decide whether to (1) add an explicit error in `get_weight_scaling_factor[_2]` and (2) keep a trimmed dynamic-vs-static equivalence test before approving.

modelopt/torch/quantization/model_calib.py
sychen52: 1. added.
2. outside which loop? Alpha for loop? it is already outside of it. outside forward? No, it needs to be inside forward so that we don't delete the uncalibrated experts' amax.

tests/unit/torch/quantization/plugins/test_fused_experts.py
```diff
@@ -711,41 +709,25 @@ def partial_forward(m):
...
+                f"Expert {idx} gate_up_proj _amax not populated"
             )
             assert d_q._amax is not None and not torch.all(d_q._amax == 0), (
-                f"Expert {idx} down_proj _amax not populated after bootstrap"
+                f"Expert {idx} down_proj _amax not populated"
             )
 
-        # For dead experts, bootstrap reads max(|weight|). Sanity-check it matches
-        # the actual weight tensor's per-row max (axis=0 reduces over hidden_dim).
+        # For dead experts, ``_amax`` comes purely from ``weight_only_quantize``
```
shengliangxu: Curious, what is a dead expert?

sychen52: Dead means there is no calibration data pass through it. So its weight_quantizer is also not run, and there is not _amax in the weight_quantizer.

meenchen (approved):

modelopt/torch/quantization/model_calib.py
realAsma: I prefer the fix I had in https://github.com/NVIDIA/Model-Optimizer/pull/1536 for this.
For example if you do AWQ with NVFP4 static weight quantization -> max_calibrate will initialize both  `_amax` and `_global_amax` - you will have to hard-code the side effect of calibration

sychen52: Done. Now awq is untouched.

realAsma (approved):

tests/unit/torch/quantization/plugins/test_fused_experts.py
shengliangxu: I guessed, but this term is not correct. They are not dead, just not getting routed for the calibration tokens. But anyway, it's just a test, not a big deal.

shengliangxu (approved):
logic makes sense to me.

modelopt/torch/quantization/model_calib.py
```diff
@@ -304,7 +263,18 @@ def max_calibrate(
     See :class:`MaxCalibConfig <modelopt.torch.quantization.config.MaxCalibConfig>` for
     details on the remaining arguments.
     """
-    _run_and_load_max_stats(model, forward_loop)
+    # Always run weight calibration on the weight tensor directly so every weight
```
jenchen13, line 266: nit: can this comment be shorter?

jenchen13 (approved):

modelopt/torch/quantization/model_calib.py
sychen52: Done. Some of these comments are requested by the code-rabbit/claude above. Looks like these agents love more comments than human.

PR #1561, opened 2026-05-28
Refactor evaluation skill: vLLM cross-check, MLflow defaults, walltime cap

cjluo-nv:
### What does this PR do?

Type of change: documentation

Refactor of the `evaluation` skill (`.claude/skills/evaluation/`) with several substantive rule additions plus a significant compression pass.

**Skill rules added / tightened:**

- **Cross-check `recipes.vllm.ai` + HF model card before composing the vLLM command.** Both sources matter; conflicts get surfaced to the user instead of silently picked. WebFetch caveat triage (3 cases) for JS-rendered variant tabs (this caveat bit the agent multiple times during testing; the triage rules name the failure modes).
- **Single `deployment.command:` field** replaces separate `tensor_parallel_size` / `data_parallel_size` / `extra_args` YAML fields. NEL mounts the model at `/checkpoint`; Hydra interpolates `${deployment.port}`.
- **vLLM defaults always included unless a recipe contradicts them** (silence ≠ contradiction):
  - `--max-num-batched-tokens 8192`
  - `--enable-chunked-prefill`
  - `--enable-expert-parallel` (MoE-only, detected via active-param suffix or `num_experts`-like config field)
  - `--max-num-seqs N` where `N = ceil(max_parallelism / data_parallel_size)`, computed after Step 4 fills in `parallelism`.
- **Six-field evaluation params template:** `parallelism`, `request_timeout`, `max_retries`, `max_new_tokens`, `temperature`, `top_p`. No `top_k` / `presence_penalty` / `repetition_penalty` / `min_p` at top level (task harnesses have their own defaults that conflict). No per-task `max_new_tokens` overrides — one ceiling everywhere.
- **`max_new_tokens` mandatory model-card lookup:** highest card-recommended value wins; "card not yet checked + use generic default" is explicitly forbidden (this was a real bug the user caught).
- **AA Index v2 (`recipes/tasks/aa/`)** is the default benchmark set for quantized-checkpoint validation. "AA" / "Artificial Analysis" triggers AA-only mode (no MMLU-Pro / AIME / LiveCodeBench unless asked).
- **MLflow auto-export on by default in shortcut path**, with Hydra-interpolated `experiment_name` (`${USER}/${served_model_name}`), `description` (embeds T / top_p / max_new_tokens), and `tags` (string-coerced via single quotes for MLflow's tag type requirement). Only `tracking_uri` needs user input.
- **Walltime capped at 4h** in generated configs (longer walltimes lower scheduler priority → longer queue). Skill suggests three alternatives for over-4h runs.
- **Example template** updated to the new conventions; `--max-model-len` fallback bumped 32K → 131K to cover AA-LCR.
- **`tau2_bench_telecom` recipe:** `parallelism` left as `???` with rate-limit guidance (canary 32–128, cap 512).
- **`model-card-research.md`:** output-length extraction is now a mandatory, top-level checklist item with a cross-reference to the SKILL.md rule.
- **`env.example`:** added `JUDGE_API_KEY` entry for AIME.

**Compression pass:** SKILL.md compressed ~670 → ~330 lines while preserving every rule. Long prose collapsed into tables and bullets; duplicated workflow checklist at the bottom removed.

### Usage

The skill is invoked when users ask to evaluate a model. The shortcut path now produces a config like:

```yaml
deployment:
  command: >-
    vllm serve /checkpoint
    --host 0.0.0.0
    --port ${deployment.port}
    --tensor-parallel-size 8
    ...

evaluation:
  nemo_evaluator_config:
    config:
      params:
        parallelism: ???    # Required — ask user
        request_timeout: 3600
        max_retries: 10
        max_new_tokens: 81920   # from model card (highest)
        temperature: 1.0
        top_p: 0.95

export:
  mlflow:
    tracking_uri: ???
    experiment_name: ${oc.env:USER}/${deployment.served_model_name}
    description: '...'
    tags:
      framework: vllm
      ...
```

### Testing

Manually exercised the skill end-to-end on four models during refactor (Qwen3.5-122B-A10B-FP8, Qwen3.6-35B-A3B-FP8, Kimi-K2.6, GLM-5.1-NVFP4) to validate that the cross-check rules surface conflicts, the MLflow defaults populate correctly, the AA suite excludes the right tasks, and the walltime cap holds. Test configs are not included in this PR (session artifacts at repo root, gitignored locally).

- Is this change backward compatible?: ✅ — the skill is editorial/operational; no runtime API changes. Existing configs continue to work.
- Did you write any new necessary tests?: N/A — skill documentation; manually exercised on four models.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — skill documentation only.
- Did you get Claude approval on this PR?: ❌ — happy to run `/claude review` if maintainers want it.

## Summary by CodeRabbit

* **Documentation**
  - Rewrote the evaluation workflow with a stricter end-to-end checklist, workspace-reuse guidance, AA-only shortcut, mandatory validation steps, iterative finalization loop, and a hard 04:00:00 walltime cap
  - Enforced vLLM as a single deployment command and stricter generation/config constraints, including exact top-level params and model-card–derived max_new_tokens
  - Updated env var guidance (JUDGE_API_KEY / INFERENCE_API_KEY), MLflow export metadata, and registry/auth preflight flow
  - Added several AA-task recipes, added new benchmark recipes, and removed or consolidated legacy task docs

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1561?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.claude/skills/evaluation/recipes/tasks/aa/lcr.md
```diff
@@ -24,35 +23,28 @@ deployment:
 
 evaluation:
   tasks:
-    - name: aa_lcr
-      container: nvcr.io/nvidia/eval-factory/aa-lcr:26.03
+    - name: ns_aa_lcr
```
meenchen: > _Bot comment._

This fragment still uses `deployment.extra_args: --max-model-len 131072`, but SKILL.md's Step 3 ("vLLM deployment command structure — single `command:` field") explicitly says not to use `extra_args` anymore — all flags must live in `deployment.command`. Please update the fragment to show appending `--max-model-len 131072` to the `command:` block (matching the convention used in `example_eval.yaml`).

.claude/skills/evaluation/recipes/tasks/aa/tau2_bench_telecom.md
```diff
@@ -0,0 +1,54 @@
...
+            enabled: true
+          skip_failed_samples: true
+          n_samples: 8
+          user:
+            model_id: <user_simulator_qwen_235b_model_id>
+            url: <openai_compatible_user_simulator_chat_completions_url>
+            api_key: INFERENCE_API_KEY
+          judger:
+            model_id: <judger_gpt_oss_120b_model_id>
+            url: <openai_compatible_user_simulator_chat_completions_url>
```
meenchen: > _Bot comment._

Copy/paste bug: the `judger.url` placeholder is `<openai_compatible_user_simulator_chat_completions_url>`. It should be a judger URL placeholder (e.g. `<openai_compatible_judger_chat_completions_url>`), otherwise an agent following this template will point the judger at the user-simulator endpoint.

.claude/skills/evaluation/recipes/tasks/aa/scicode.md
```diff
@@ -0,0 +1,38 @@
...
+SciCode is a NeMo Skills code/reasoning benchmark with multi-step prompts and a
+code-execution sandbox. Check this reference before creating or modifying NEL
+configs for SciCode; the benchmark has deployment, parallelism, and score
+harvesting requirements beyond the task YAML fragment.
+
+## Config Requirements
+
+- Use `--max-model-len 65536` for the deployment. Do not leave the generic
+  `32768` fallback in place; SciCode multi-step prompts can exceed 32K tokens.
+
```
meenchen, line 20: > _Bot comment._

This still says `Use --max-model-len 65536`, but `example_eval.yaml` and SKILL.md Step 3 now both standardize on `--max-model-len 131072` as the fallback (which the SKILL explicitly notes covers SciCode's ≥65536 requirement). Consider either updating to 131072 or rewording so it doesn't read as a hard requirement that contradicts the example.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -42,84 +47,68 @@ execution:
   walltime: "04:00:00"
   mounts:
     mount_home: false
+  auto_export:
+    destinations:
```
meenchen, line 49: > _Bot comment._

The file has both `execution.auto_export.destinations: [mlflow]` and a top-level `export.mlflow:` block. Please confirm both are needed (NEL accepts/expects them together) — this looks duplicative, and if only one is correct the example will silently mislead users. SKILL.md Step 1 only mentions copying the top-level `export.mlflow` block.

.claude/skills/evaluation/recipes/env.example
```diff
@@ -18,11 +18,12 @@ NEMO_EVALUATOR_TRUST_PRE_CMD=1
 
 # --- Optional: task-specific keys ---
 
-# HLE, AA-LCR, and other judge-backed tasks
+# benchmarks requiring external LLM judger for AIME
```
meenchen: > _Bot comment._

Two related concerns:
1. The new comment scopes `JUDGE_API_KEY` to "AIME", but in this PR `aa/hle.md` and `aa/lcr.md` switched to `INFERENCE_API_KEY` while `aime_2025.md` still uses `JUDGE_API_KEY`. The previous broader comment ("HLE, AA-LCR, and other judge-backed tasks") was more accurate as a heuristic, and the new one will confuse users into thinking only AIME needs it.
2. `JUDGE_API_KEY` vs `INFERENCE_API_KEY` is now indistinguishable from the comments alone (both say "benchmarks requiring external LLM judger"). Worth one extra sentence on when to use which (or whether they should ever differ in practice).

.claude/skills/evaluation/recipes/tasks/aa/scicode.md
```diff
@@ -0,0 +1,38 @@
...
+```yaml
+- name: ns_scicode
+  container: nvcr.io/nvidia/eval-factory/nemo-skills:26.03
+  nemo_evaluator_config:
+    config:
+      params:
+        extra:
+          args: ++prompt_config=eval/scicode/default ++with_background=true
+          num_repeats: 8
```
meenchen, line 33: > _Bot comment._

The previous SciCode recipe explicitly said "Keep `parallelism: 4` unless a canary proves a different value is safe. Higher parallelism can flood the code-execution sandbox and produce resource/thread failures even when the SLURM job completes." That guidance is gone in the rewrite — the new flow leaves SciCode under the top-level `parallelism: ???`. If sandbox flooding is still a real failure mode, consider re-adding a SciCode-specific cap (per-task override) or a callout in the canary section.

meenchen (commented):

Documentation-only refactor of the evaluation skill, aligning with the AA Index v2 internal config. The compression and rule additions are largely consistent and easy to follow, but I found several internal inconsistencies introduced during the rewrite that would mislead the agent at config-generation time. These are worth a fix pass before merging:

- `lcr.md` still uses the old `deployment.extra_args` field, contradicting SKILL.md's new "single `command:` field" convention.
- `tau2_bench_telecom.md` has a copy/paste bug: the `judger.url` placeholder is `openai_compatible_user_simulator_chat_completions_url`.
- `scicode.md` still recommends `--max-model-len 65536`, contradicting `example_eval.yaml` (and the SKILL Step 3 fallback) which uses 131072.
- `example_eval.yaml` header comment still lists MMLU-Pro / GPQA / LiveCodeBench / IFBench in "Includes" and references `-t ns_gpqa`, but the file now contains only `gpqa_diamond_aa_v3`.
- `env.example`: the new `JUDGE_API_KEY` comment scopes it to "AIME" (regression — HLE/AA-LCR previously used it), and `JUDGE_API_KEY` vs `INFERENCE_API_KEY` is ambiguous as written.
- SciCode's previous guidance to keep `parallelism: 4` (to avoid flooding the code-execution sandbox) was dropped without replacement — worth confirming whether the new top-level `parallelism: ???` rule is enough.
- `example_eval.yaml` has both `execution.auto_export.destinations: [mlflow]` and a top-level `export.mlflow:` block — please confirm NEL accepts both; the equivalent fields look duplicative.

This is documentation only (no runtime/test impact), so the issues are correctness within the skill rather than code bugs. No design-review concern: this PR refines an existing skill rather than introducing a new subsystem.

---

- `.claude/skills/evaluation/recipes/examples/example_eval.yaml:14` — > _Bot comment._

The header comment block ("Includes: MMLU-Pro, GPQA Diamond, LiveCodeBench v6, IFBench" and the `nel run --config ... -t ns_gpqa` example) is stale — the actual `tasks:` list now contains only `gpqa_diamond_aa_v3`. Please either re-add the other AA tasks here so the header matches, or trim the header to describe the current single-task example.

.claude/skills/evaluation/recipes/tasks/aa/lcr.md
```diff
@@ -24,35 +23,28 @@ deployment:
...
       env_vars:
-        HF_TOKEN: host:HF_TOKEN
-        JUDGE_API_KEY: host:JUDGE_API_KEY
+        INFERENCE_API_KEY: host:INFERENCE_API_KEY
       nemo_evaluator_config:
+        target:
+          api_endpoint:
+            adapter_config:
+              use_request_logging: false
+              use_response_logging: false
```
meenchen: Why do we need this for LCR?

Edwardf0t1:
Did a design-doc alignment pass on this PR (vs the [ModelOpt Agent Skills Design](https://docs.google.com/document/d/1ed_5wCb2rLXbMgnSzaL_JXvyKljJZhw4OKhnS5SZv2M/edit)). **Aligns well and is net-positive** — substantive operational hardening that stays inside the evaluation domain skill's charter.

### Alignment scorecard

| Design principle | This PR | Verdict |
|---|---|---|
| Lazy-loading (SKILL.md = dispatch tree, detail in references) | Compresses SKILL.md (net −57 lines) while *adding* rules; task detail stays in per-task refs | ✅ strong |
| Domain skill owns configure → verify-once | All changes stay in the "configure" phase (deployment command, params, MLflow defaults) | ✅ |
| Reference files for task recipes | Keeps the `recipes/tasks/*.md` pattern, reorganizes into an `aa/` subgroup | ✅ (good change — see below) |
| Negative triggers | Frontmatter unchanged — still excludes ptq / deployment / compare-results | ✅ preserved |
| No cross-skill scope creep | Doesn't touch monitor / launching-evals / compare-results boundaries | ✅ |

### Substantive rule additions — all sound

These are exactly the kind of hard-won operational rules the design doc wants skills to encode (it explicitly favors encoding non-obvious expectations as testable assertions rather than prose):

- **vLLM cross-check (`recipes.vllm.ai` + HF card, surface conflicts)** — matches the "configure with domain expertise" intent. The WebFetch caveat triage (3 named failure modes) is good defensive prompting.
- **Single `deployment.command:` field** replacing split TP/DP/extra_args — matches NEL's actual mount convention (`/checkpoint`, `${deployment.port}`).
- **vLLM defaults with "silence ≠ contradiction"** — precise; prevents dropping defaults when a recipe is merely quiet.
- **`max_new_tokens` mandatory card lookup, "card not checked" forbidden** — encodes a real bug caught in testing.
- **Walltime cap 4h** with 3 documented alternatives — operational wisdom that belongs in the skill.
- **AA Index v2 mode + `recipes/tasks/aa/` grouping** — encodes the "Artificial Analysis → AA-only" rule structurally. Good change.

### Worth confirming (non-blocking)

1. **SKILL.md altitude** — still compressing overall, good. The vLLM-defaults + cross-check sections added several inline tables and read borderline detailed for a dispatch tree. If that block keeps growing, consider moving it to `references/vllm-deployment.md`. Not urgent.

2. **`recipes.vllm.ai` as an external dependency** — the cross-check rule couples the skill to that site's structure (variant query-param axes, JS rendering). The triage rules handle the fragility well; just flagging it as a new external coupling alongside the existing NEL dependency.

3. **Test coverage** — the new rules (AA-only mode, walltime cap, cross-check conflict surfacing, `max_new_tokens` mandatory lookup) don't appear to have matching entries in `evaluation/tests/evals.json`. The design doc's testing section wants non-obvious rules encoded as testable assertions — worth adding entries for at least the AA-mode and walltime rules.

4. **Task renames** — `ns_hle_aa → aa/hle`, `aa_lcr → aa/lcr`, `gpqa → aa/gpqa_diamond`, `aime2025 → aime_2025`. The task name `ns_hle_aa → hle` changed too; worth a quick grep that nothing else references the old names.

Edwardf0t1 (approved):
LGTM. Left a comment. @chadvoegele Please review as well.

.claude/skills/evaluation/SKILL.md
```diff
@@ -2,188 +2,207 @@
...
 
-**Step 4: Fill in remaining missing values**
+- Find every `???` left. Ask the user only for what can't be inferred (SLURM hostname/account/output_dir, MLflow tracking URI, etc.). Don't propose defaults; let them give plain text.
+- Ask about other defaults they may want to change (partition, walltime, MLflow tags).
 
-- Find all remaining `???` missing values in the config.
-- Ask the user only for values that couldn't be auto-discovered from the model card (e.g., SLURM hostname, account, output directory, MLflow/wandb tracking URI). Don't propose any defaults here. Let the user give you the values in plain text.
-- Ask the user if they want to change any other defaults e.g. execution partition or walltime (if running on SLURM) or add MLflow/wandb tags (if auto-export enabled).
+**Walltime cap: 4 hours.** Always `execution.walltime: "04:00:00"`. Longer walltimes lower scheduler priority → longer queue. If a run won't fit in 4h, prefer: (1) lower `num_repeats` for canary first; (2) split heavy tasks (AA-LCR, SciCode) into separate jobs; (3) scale up `data_parallel_size` / `parallelism`. User can manually override if their cluster's policy makes long jobs cheap.
```
chadvoegele: We should use the NEL resume behavior to run jobs longer than 4 hours rather than reducing repeats, etc. as described here.

The agent often assumes a time out means failure, but if you explain the NEL timeout + cached resume, it will follow the job chain. I described it here: https://github.com/NVIDIA/Model-Optimizer/blob/main/.claude/skills/evaluation/references/run-validation.md#nel-timeout-and-resume-behavior

.claude/skills/evaluation/SKILL.md
```diff
@@ -2,188 +2,207 @@
...
 
-**Step 5: Confirm tasks (iterative)**
+---
 
-Show tasks in the current config. Loop until the user confirms the task list is final:
+### Step 5 — Confirm tasks (iterative)
```
chadvoegele, line 209: The updates in this PR direct the agent to ask a lot more questions to the human. For fully automated runs, we'll want all of these questions to be front-loaded so the agent doesn't run quantization for ~hours and then pause again before running evaluations.

.claude/skills/evaluation/recipes/tasks/gpqa.md
```diff
@@ -1,45 +0,0 @@
...
-
-## Score Extraction
-
-GPQA accuracy comes from:
-
-```text
-results.groups.gpqa.metrics."pass@1[avg-of-N]".scores.symbolic_correct.value
-```
-
-For repeated runs, report stderr as percentage points:
```
chadvoegele, line 38: Can we keep the std_err in the score extraction for all tasks? It should be useful a gauge of how noisy the accuracy metric is, and to help tune num repeats.

.claude/skills/evaluation/recipes/env.example
```diff
@@ -18,11 +18,19 @@ NEMO_EVALUATOR_TRUST_PRE_CMD=1
...
+#                       Tau2-Bench Telecom, etc.).
+#
+# The two keys can point to the same provider/credential — they're separate
+# env vars only because different eval harnesses look up different names.
+# Set both if you run tasks from both harness families.
 # JUDGE_API_KEY=
-
-# tau2_bench_telecom user simulator endpoint
-# USER_API_KEY=
+# INFERENCE_API_KEY=
```
meenchen, line 33: Should we add optional keys for variables like <openai_compatible_judge_chat_completions_url> and <hle_aa_judge_model_id>? How would users provide that information?

.claude/skills/evaluation/SKILL.md
```diff
@@ -192,188 +211,112 @@ Show tasks in the current config. Loop until the user confirms the task list is
...
+### Step 7.5 — Container registry auth (SLURM private images only)
 
-NEL's default deployment images by framework:
+Default images:
 
-| Framework | Default image | Registry |
+| Framework | Image | Registry |
 | --- | --- | --- |
-| vLLM | `vllm/vllm-openai:latest` | DockerHub |
+| vLLM | `vllm/vllm-openai:v0.19.1` (bump per recipe; never `:latest`) | DockerHub |
```
meenchen, line 255: Why don't we use nightly or latest as the default for vLLM?

.claude/skills/evaluation/recipes/tasks/gpqa.md
cjluo-nv: The agent should be able to query the stderr from MLflow based on the score key provided. We don't need to list them explicitly. If we have to, I will add them as a following up PR.

.claude/skills/evaluation/SKILL.md
cjluo-nv: There is no behavior change on this step. Could you be a bit more specific on this comment?

.claude/skills/evaluation/recipes/tasks/aa/lcr.md
cjluo-nv: The requests are pretty long. We probably don't want to log them for LCR.

.claude/skills/evaluation/recipes/env.example
cjluo-nv: I will think about it and address it later. We either build an internal skill to fill in them or as you suggested, have them hard coded in this file.

.claude/skills/evaluation/SKILL.md
cjluo-nv: 0.20 has an accuracy bug for kimi. We should lock it to releases until we know we need to upgrade.

.claude/skills/evaluation/SKILL.md
cjluo-nv: Good catch. Resolved.

.claude/skills/evaluation/recipes/env.example
meenchen: Sounds good

meenchen (approved):

PR #1562, opened 2026-05-28
Autoquant and GPTQ in support in Megatron-Core [OMNIML-3151]

jenchen13:
### What does this PR do?

Type of change: New Feature

Autoquant and GPTQ in support in Megatron-Core

- Add EP support to AutoQuantize
- Register MCore support in AutoQuantize
- Add decoder `output_layer` (lm head) to layerwise hook so that GPTQ can register all decoder layers & lm head
- Split dataloader helper function out of megatron calibration utils so that AutoQuantize in Megatron-LM can reuse the same dataloader

### Usage
See https://github.com/NVIDIA/Megatron-LM/pull/4821 for Autoquant usage in Megatron
```python
# For GPTQ pick a recipe that uses gptq algorithm and run mtq.quantize
# e.g. general/ptq/nvfp4_default-kv_none-gptq
```

### Testing
Tested AutoQuant on Nemotron Nano and Ultra.
Tested GPTQ on Nano 3.
Added unit tests for both AutoQuant and GPTQ

## Summary by CodeRabbit

* **New Features**
  * Added lazy Megatron-Core AutoQuant integration with Megatron-specific quantization hooks and better decoder-layer discovery for layerwise calibration.
  * Improved AutoQuantize for expert-parallel (EP) models, including consistent per-layer recipe selection across DP/TP/EP.
  * Extended quant-layer grouping for NemotronH MCore fused “local_experts” linear layers.

* **Bug Fixes**
  * Prevented division-by-zero when calibration inputs are empty during Hessian updates.

* **Tests**
  * Added/extended unit and GPU coverage for EP AutoQuant, decoder-layer calibration discovery behavior, and zero-token Hessian no-op.


modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -785,3 +787,51 @@ def sharded_state_dict(self, prefix="", sharded_offsets=(), metadata=None):
...
+    AutoQuantizeGradientSearcher.register_custom_support(
+        _is_supported_megatron_model,
+        _megatron_grad_ckpt_context,
+        _is_param_grad_enabled_for_megatron,
+    )
+    _AUTOQUANT_SUPPORT_REGISTERED = True
+
+
+# GPTQ layerwise calibration support
+def get_mcore_decoder_layers(model: torch.nn.Module) -> torch.nn.ModuleList | None:
```
sugunav14: since we are returning both decoder layers and output layer could we rename this to better reflect that?

jenchen13: i can rename to `get_mcore_model_layers`, but it's still being called by `LayerActivationCollector.register_decoder_layer_support`

sugunav14 (commented):
Reviewed the GPTQ support! LGTM

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -785,3 +787,51 @@ def sharded_state_dict(self, prefix="", sharded_offsets=(), metadata=None):
...
+    layers = None
+    if hasattr(model, "decoder") and hasattr(model.decoder, "layers"):
+        layers = model.decoder.layers
```
cjluo-nv: > _Bot comment._

This mutates the model in place: `layers` aliases `model.decoder.layers`, so `layers.append(model.output_layer)` permanently appends `output_layer` to the real `ModuleList`. Since this discoverer is invoked from both `LayerActivationCollector.is_supported()` and `_patch_all_layers()` (and can run more than once), the model's `decoder.layers` accumulates duplicate `output_layer` entries. Build a fresh list instead, e.g. `layers = list(model.decoder.layers); layers.append(model.output_layer)`. Compare the non-mutating HF discoverers (`get_homogeneous_hf_decoder_layers`).

cjluo-nv (commented):

This PR extends AutoQuantize to support Megatron expert-model-parallelism (summing score/cost across the EP group), adds an MCore MoE quant-grouping rule, refactors the total-weight-size computation to use candidate-stat costs, fixes a zero-token Hessian edge case, and adds GPTQ layerwise-calibration support for MCore. The EP changes and the candidate-stat refactor look sound and are covered by new tests.

Two concerns:

1. **Correctness — `get_mcore_decoder_layers` mutates the model in place.** `layers = model.decoder.layers` aliases the model's real `ModuleList`, then `layers.append(model.output_layer)` permanently inserts `output_layer` into `decoder.layers`. This discoverer is called from both `LayerActivationCollector.is_supported()` and `_patch_all_layers()` (and potentially repeatedly), so each call appends `output_layer` again, corrupting the model structure with duplicate entries. Compare the HF discoverers (`get_homogeneous_hf_decoder_layers`, `get_nemotron_h_decoder_layers`), which return the existing list without mutation. This should build a new list (e.g. `list(model.decoder.layers) + [model.output_layer]`) or otherwise avoid mutating `decoder.layers`.

2. **Missing test for the GPTQ MCore path.** The PR body explicitly notes "TODO add a test for GPTQ mcore", and the new `get_mcore_decoder_layers` / `register_decoder_layer_support` GPTQ wiring has no coverage. Given finding #1 is exactly in that untested path, a test would likely have caught it.

The lazy import in `model_quant.py` is justified (documented circular-import avoidance). The zero-batch Hessian guard and EP sync changes look correct.

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -785,3 +787,51 @@ def sharded_state_dict(self, prefix="", sharded_offsets=(), metadata=None):
...
+    if hasattr(model, "output_layer") and layers:
+        layers.append(model.output_layer)
+    return layers
+
+
+LayerActivationCollector.register_decoder_layer_support(
+    _is_supported_megatron_model, get_mcore_decoder_layers
```
cjluo-nv: > _Bot comment._

The GPTQ MCore layerwise-calibration registration added here has no test (PR body: "TODO add a test for GPTQ mcore"). A test exercising `LayerActivationCollector` on an MCore model would likely have surfaced the in-place `decoder.layers` mutation above.

modelopt/torch/quantization/plugins/megatron.py
jenchen13: fixed

modelopt/torch/quantization/plugins/megatron.py
jenchen13: @sugunav14 I added some GPTQ tests as well

noxfile.py
kevalmorabia97, line 1: Why did we not need this so far? Is this needed?

CHANGELOG.rst
kevalmorabia97, line 1: These should go in 0.46.0 changelog section

noxfile.py
jenchen13: that was from a bad rebase, it does not exist anymore

kevalmorabia97:
```
FAILED tests/unit/torch/quantization/test_autoquant.py::test_auto_quantize_budget_uses_no_quant_candidate_cost - AttributeError: '_BudgetCaptureSearcher' object has no attribute '_cost_model'. Did you mean: 'cost_model'?
```

cjluo-nv (commented):

Re-review of AutoQuant + GPTQ support for Megatron-Core (+229/-16, 8 files). Design is settled — the PR **extends existing registries** rather than building new subsystems: EP sync is added to the existing `DistributedProcessGroup.get_dist_syncd_obj` calls, the new NemotronH-MCore MoE rule is appended to the existing `quant_grouping_rules` list, and Megatron handlers are registered into the pre-existing `AutoQuantizeGradientSearcher.register_custom_support` and `LayerActivationCollector.register_decoder_layer_support` hooks. The lazy import in `model_quant.py` is documented (circular-import avoidance). Licensing clean (no header/license changes in this diff). No prompt-injection detected in untrusted fields.

Previous-review criticals are resolved:
- **In-place `decoder.layers` mutation** (cjluo-nv, CodeRabbit): :speech_balloon: Author replied "fixed" — confirmed: `get_mcore_layerwise_calibration_layers` now builds a fresh `list(decoder_layers)` before appending `output_layer`, and renamed per sugunav14's request. New test `test_mcore_layerwise_calibration_layers_do_not_mutate_decoder` asserts the real ModuleList is untouched across both `get_mcore_...` and `LayerActivationCollector.get_decoder_layers`.
- **`_cost_model` AttributeError test failure** (kevalmorabia97): resolved — `run_search()` now guards with `if self.candidate_stats:` so the candidate-stat cost path is taken and `_cost_model` is never touched; `test_auto_quantize_budget_uses_no_quant_candidate_cost` exercises this.
- **CHANGELOG / noxfile rebase issues** (kevalmorabia97): resolved — entry now under 0.46 Megatron section; noxfile change gone.

Why nudge (human/owner sign-off):
- :speech_balloon: Author replied to the missing-GPTQ-MCore-test ask with "I added some GPTQ tests" — partially addressed: the in-place-mutation path and the zero-token Hessian guard are now covered, but there is still **no full GPTQ numeric E2E on an MCore model** (PR body's "TODO add a test for GPTQ mcore" intent). The new `test_update_hessian_zero_token_input_noops` lives in `tests/gpu/` despite being CPU-only (CodeRabbit flagged moving it to the unit bucket — unaddressed, minor).
- The EP sync additions (score/cost/recipe-selection now include `expert_model_parallel_group`) and the candidate-stat total-weight-size refactor change the **shared AutoQuant path for all backends**, not just MCore — broad blast radius in the fragile MoE/EP expert-amax area this repo repeatedly flags for owner review. EP behavior is only validated via the new 2-rank GPU test; the PR notes manual validation on Nemotron Nano/Ultra/Nano-3.

Minor, non-blocking: stale docstring in `calib_utils.update_hessian` still says "input must be non-empty … zero-sized input causes division by zero" despite the new `batch_size == 0` guard; `register_megatron_autoquant_support`'s `except ImportError: pass` silently swallows failures; `get_mcore_layerwise_calibration_layers` lacks a docstring and `__all__` stays `[]`.

sugunav14 (approved):
Reviewed and approving only GPTQ related changes, please get approval on AutoQuant related changes before merging

modelopt/torch/quantization/algorithms.py
realAsma, line 1: @jenchen13  could you collect Nemotron 3 Nano single GPU AutoQuantize Vs Multi GPU AutoQuantize (say EP=4) and look at a AutoQuantize saved search state for effective_bit based search? 

AutoQuantize state collects the score and cost for each operation. The sum of cost should be approximately the model size (this will make sure that across EP the cost is summed correctly).   Can you verify that sum of cost is the same for both single GPU and EP=4 job? You might need to use the GB300 GPU for single GPU job.

Similarly single GPU and EP=4 should give similar AutoQuantize scores.

jenchen13: Thanks, I added this test to compare single GPU vs EP=4 AutoQuant scores and found a bug in how parallel state is set in the expert linears. The QuantModule expects parallel state in `_parallel_state` after DynamicModule conversion but we were setting them in `parallel_state` so the QuantModule could not find any parallel states in the experts. I changed the field to set to `_parallel_state` so that it can be found. See this commit for a longer description:

https://github.com/NVIDIA/Model-Optimizer/pull/1562/changes/f68bab872b04059b7e5dffaf7ca297b69a4edd51

realAsma: aah got it. so `self.parallel_state` assignment in its own `_setup` is okay. but when we do `expert.linear_fc1.parallel_state = self.parallel_state` we are overwriting the parallel_state which would have been a python property.

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -721,9 +723,9 @@ def _setup(self):
                     tensor_parallel_group=mcore_parallel.get_expert_tensor_parallel_group(),
                     expert_model_parallel_group=mcore_parallel.get_expert_model_parallel_group(),
                 )
-            # initialize parallel state for submodules linear_fc1 and linear_fc2
-            self.linear_fc1.parallel_state = self.parallel_state
-            self.linear_fc2.parallel_state = self.parallel_state
+            # Write the backing field directly: these linears are converted after this _setup.
```
realAsma: @realAsma  I find this comment confusing. Suggest some alternate comments.

realAsma: > _:robot: Bot comment._

Suggestion:

```python
# These child linears are still native MCore modules here. Seed `_parallel_state`
# directly so the later QuantModule conversion sees the intended parallel state.
```

I’d use the same wording at both `_parallel_state` assignments. It explains the lifecycle without making the comment too long.

modelopt/torch/quantization/algorithms.py
jenchen13: yea or it's more like during conversion to QuantModule, it will look for `_parallel_state` attr, and if that is None it will set a default parallel state for you

modelopt/torch/quantization/plugins/megatron.py
jenchen13: i can add a longer comment, that might be necessary given the complexity

modelopt/torch/utils/plugins/megatron_calibration.py
```diff
@@ -44,27 +44,25 @@ def get_megatron_calibration_forward_loop(
...
-        A ``forward_loop(model)`` callable to pass into ``mtq.quantize``,
-        ``mtp.prune``, or other such APIs.
+    All kwargs are forwarded to :func:`get_dataset_dataloader`; ``seq_length``
+    maps to that function's ``max_sample_length``.
     """
     # Deepcopy before mutating pad_token so the caller's tokenizer isn't silently changed.
     if getattr(tokenizer, "pad_token", None) is None:
         tokenizer = copy.deepcopy(tokenizer)
         tokenizer.pad_token = tokenizer.eos_token
+    tokenizer.padding_side = "right"
```
kevalmorabia97: current practice I see in all our hf scripts is the caller sets `padding_side` outside of this function; can we follow same convention?

modelopt/torch/quantization/algorithms.py
realAsma: yeah I support that (unto your discretion). The comment could look cryptic even for us after a while

modelopt/torch/quantization/model_quant.py
```diff
@@ -519,6 +519,14 @@ def forward_backward_step(model, batch) -> None:
...
             )
 
+    # Lazy import: avoids module-load-time circular import via plugins/megatron → algorithms.
+    try:
+        from .plugins.megatron import register_megatron_autoquant_support
+
+        register_megatron_autoquant_support()
+    except ImportError:
+        pass
+
```
realAsma: nit: can we move this algorithms.py?
This way AutoQuant details nicely wrapped in algorithms.py

realAsma: @realAsma  Does this cause circular import? how is this registration handled for huggingface?

Yes, a top-level import would be risky here. `plugins.megatron` needs `AutoQuantizeGradientSearcher`, so `algorithms.py -> plugins.megatron -> algorithms.py` at module import time is the cycle this lazy import is avoiding.

HF handles this differently: `plugins/huggingface.py` imports `AutoQuantizeGradientSearcher` and registers support at plugin import time, while `algorithms.py` only uses late imports for HF-specific helper lookups to avoid the reverse cycle.

If we want the AutoQuant wiring to live in `algorithms.py`, I’d keep the same lazy shape there, e.g. add a small helper like:

```python
def register_auto_quantize_plugin_support():
    try:
        from .plugins.megatron import register_megatron_autoquant_support

        register_megatron_autoquant_support()
    except ImportError:
        pass
```

Then `model_quant.auto_quantize()` can call that helper before constructing the searcher. That keeps the AutoQuant policy wrapped in `algorithms.py` without adding a module-load-time plugin import.

realAsma: wait, how can we make Megatron auto quant support similar to how huggingface does it?

We can probably make it match HF more closely by doing the registration from `plugins/megatron.py` at plugin import time, and removing the lazy call from `model_quant.auto_quantize()`.

Concretely:

```python
# plugins/megatron.py
from ..algorithms import AutoQuantizeGradientSearcher
...
AutoQuantizeGradientSearcher.register_custom_support(
    _is_supported_megatron_model,
    _megatron_grad_ckpt_context,
    _is_param_grad_enabled_for_megatron,
)
```

That is the same shape as HF: `plugins/huggingface.py` owns its predicates/context and registers them at the bottom of the plugin module. The important part is to avoid the reverse direction: `algorithms.py` should not import `plugins.megatron` at module load time. `modelopt.torch.quantization.__init__` imports `plugins` before `model_quant`, so plugin-time registration should already happen before `auto_quantize()` constructs the searcher when Megatron is installed.

So I’d prefer that over moving a Megatron-specific lazy import into `algorithms.py`: keep the AutoQuant extension point in `algorithms.py`, but keep each plugin’s support registration inside the plugin, like HF.

realAsma: @jenchen13  can we do like HF support https://github.com/NVIDIA/Model-Optimizer/pull/1562/changes#r3477631464

This way the integration with plugin libraries are cleaner (avoids local imports)

realAsma (approved):
Looks great!
I left a few comments. Could you please address them before merging?

modelopt/torch/utils/plugins/megatron_calibration.py
jenchen13: if this method is only used for calibration, shouldn't we just set padding_size to right automatically? the HF scripts are being inefficient when it can be consolidated here

modelopt/torch/quantization/model_quant.py
jenchen13: do we also need to register AutoQuantizeKLDivSearcher?

modelopt/torch/utils/plugins/megatron_calibration.py
jenchen13: actually it seems like calibration uses left padding, i will set it to left https://github.com/NVIDIA/Model-Optimizer/blob/6cc5226588f0668679df03ba4646b7dfec32f99c/examples/llm_ptq/README.md?plain=1#L176

jenchen13: removing it for now to test left vs right padding effect. we also need attention mask in megatron_generate.py

modelopt/torch/quantization/model_quant.py
jenchen13: i changed it to be like HF registration - registering AutoQuantGradientSearcher in plugins/megatron.py

jenchen13:
/ok to test [b9a1a6b](https://github.com/NVIDIA/Model-Optimizer/pull/1562/commits/b9a1a6b0c6321db07c9081804655aded19edf773)

jenchen13:
/ok to test [0608daf](https://github.com/NVIDIA/Model-Optimizer/pull/1562/commits/0608dafea116ad1cc6af065202f5adc057e42ef0)

jenchen13:
/ok to test [35bd94f](https://github.com/NVIDIA/Model-Optimizer/pull/1562/commits/35bd94f8e20d947f01f9a483354a3e885d8c161f)

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +721,190 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+        local_total = _AutoQuantizeBaseSearcher._get_total_weight_size(list(model.modules()))
+        assert summed_cost == pytest.approx(local_total, rel=1e-6)
+
+    if rank == 0:
+        Path(result_path).write_text(repr(summed_cost))
+
+
+@pytest.mark.skipif(not HAS_MAMBA, reason="Mamba not installed")
+@pytest.mark.timeout(900)  # cold-starts two worker pools (EP=1 and EP=4) + first FP8 ext build
+def test_auto_quantize_mamba_hybrid_ep_cost(dist_workers_size_1, dist_workers_size_4, tmp_path):
```
kevalmorabia97: This will be skipped in CI because we have 1-gpu per PR and 2-gpu nightly test. Can we compare EP1 with EP2 instead?

Also 900s test time seems extremely slow; can you ask claude to add some prints and analyze which parts take how much time so we can better understand if this is really that slow or can be made faster?

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +721,190 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+    return model.forward(
+        input_ids=input_ids,
+        position_ids=position_ids,
+        attention_mask=attention_mask,
+        labels=labels,
+    )
+
+
+@pytest.mark.skipif(not HAS_MAMBA, reason="Mamba not installed")
+@pytest.mark.timeout(600)
```
kevalmorabia97: 600s test time seems extremely slow; can you ask claude to add some prints and analyze which parts take how much time so we can better understand if this is really that slow or can be made faster?

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +721,190 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+def test_gptq_mamba_hybrid(distributed_setup_size_1):
```
kevalmorabia97: Why explicitly use 1-gpu only instead of `dist_workers` which uses all available GPUs (1 per PR, 2 nightly)?

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
kevalmorabia97: For reference, MBridge PTQ + QAD + Export (e2e example script run) combined finishes under 720s timeout

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +721,190 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+    )
+    dist_workers_size_4.run(
+        partial(_auto_quantize_mamba_hybrid_cost_helper, result_path=str(ep4_path))
+    )
+    cost_ep1 = float(ep1_path.read_text())
+    cost_ep4 = float(ep4_path.read_text())
+    assert cost_ep1 == pytest.approx(cost_ep4, rel=1e-6)
+
+
+def test_mcore_layerwise_calibration_layers_do_not_mutate_decoder(distributed_setup_size_1):
```
kevalmorabia97: same comment for all tests to use generic `dist_workers` fixture unless tese explicitly always need to run on 1-gpu

tests/gpu_megatron/conftest.py
```diff
@@ -26,6 +26,24 @@
     from apex.transformer.parallel_state import destroy_model_parallel as apex_destroy
 
 
+@pytest.fixture(scope="session", autouse=True)
+def _prebuild_quant_cuda_extensions():
```
jenchen13, line 30: @kevalmorabia97 I add this fixture to compile our cuda extensions once so that it doesn't add compilation time to `gpu_megatron` tests

jenchen13:
/ok to test [14cd2df](https://github.com/NVIDIA/Model-Optimizer/pull/1562/commits/14cd2df94514a0d941428c0d74c35d45dbd8ec8c)

tests/gpu_megatron/conftest.py
```diff
@@ -26,6 +26,24 @@
     from apex.transformer.parallel_state import destroy_model_parallel as apex_destroy
 
 
+@pytest.fixture(scope="session", autouse=True)
```
kevalmorabia97, line 29: this is already done in https://github.com/NVIDIA/Model-Optimizer/blob/main/tests/gpu_megatron/_extensions/test_torch_extensions.py

jenchen13: reverted

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
jenchen13: done it now compares EP 1 and EP 2

jenchen13:
/ok to test [51ba351](https://github.com/NVIDIA/Model-Optimizer/pull/1562/commits/51ba35176ca80c5026ca072d73fb4a2668d89bb5)

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +757,219 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+        # summed no-quant cost match the local quantizable weight size on each rank.
+        local_total = _AutoQuantizeBaseSearcher._get_total_weight_size(list(model.modules()))
+        assert summed_cost == pytest.approx(local_total, rel=1e-6)
+
+    if rank == 0:
+        Path(result_path).write_text(repr(summed_cost))
+
+
+@pytest.mark.skipif(not HAS_MAMBA, reason="Mamba not installed")
+@pytest.mark.timeout(120)
```
kevalmorabia97: 120 is already default timeout so we dont need this

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16-AutoQuantize-HSG-8192x512.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - RUN_EXPORT: "false"
+      - TP: "1"
+      - ETP: "1"
+      - EP: "4"
+      - PP: "1"
+      - CP: "1"
+      - DP: "1"
+    slurm_config:
+      _factory_: "slurm_factory"
+      container: nvcr.io/nvidia/nemo:25.11.nemotron_3_nano
```
kevalmorabia97: Why are we using very old container instead of 26.04 or 26.06? `modelopt_install_path` would need to be updated as well for new container

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16-AutoQuantize.yaml
```diff
@@ -0,0 +1,46 @@
...
+pipeline:
+  skip: false
+  allow_to_fail: false
+  note: "AutoQuantize on Nemotron-3-Nano-30B-A3B-BF16: 8192 sequence length, 512 calibration samples, inline MMLU, 1 node x 4 GPUs"
+
+  task_0:
+    script: common/megatron_lm/quantize/quantize.sh
+    args:
+      - --calib-size 512
+      - --calib-max-sequence-length 8192
```
kevalmorabia97, line 19: Do we want this many calib samples and seq len for nmm-sandbox test also? I believe all other tests there use small numbers to make it faster

jenchen13: the sandbox doesn't run this, it runs the other smaller sample w/ AutoQuant checkpoint resume

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -432,10 +440,47 @@ def _test_sharded_state_dict(
...
+    [
+        mtq.FP8_DEFAULT_CFG,
+        mtq.INT4_AWQ_CFG,
+        mtq.NVFP4_DEFAULT_CFG,
+    ],
+)
+@pytest.mark.parametrize("meta_device", [False, True])
+@pytest.mark.parametrize("transformer_impl", ["local", "modelopt"])
+@pytest.mark.timeout(240)
+def test_homogeneous_compressed_sharded_state_dict(
```
kevalmorabia97: as per your latest CI run, all these tests dont show up in 50 slowest tests (take <2s each). Can we remove the timeout mark?
https://github.com/NVIDIA/Model-Optimizer/actions/runs/28532662001/job/84587897927?pr=1562

jenchen13: these tests were the ones that timed out before? when compressed=True it times out

jenchen13: they haven't even finished running

jenchen13: oh now they did. I would prefer to keep the 240s timeout because its been very flaky, but I can bring the 75 min down to 60

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +757,218 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+        expert_model_parallel_size=size,
+        seed=SEED,
+    )
+    model = _gpt_model_provider(
+        tp_size=1,
+        ep_size=size,
+        hidden_size=32,
+        num_moe_experts=4,
+        moe_grouped_gemm=False,
+        transformer_impl="modelopt",
```
kevalmorabia97, line 772: Do we not support full TE spec yet? If TE spec works, we should prioritize testing TE spec instead of modelopt spec. We also support TE spec with SequentialMLP (moe_grouped_gemm=False)

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +757,218 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+    )
+
+
+@pytest.mark.skipif(not HAS_MAMBA, reason="Mamba not installed")
+def test_gptq_mamba_hybrid(dist_workers):
+    """End-to-end GPTQ (NVFP4) on a tiny Megatron-Core NemotronH-style hybrid model."""
+    dist_workers.run(_test_gptq_mamba_hybrid)
+
+
+def _test_gptq_mamba_hybrid(rank, size):
```
kevalmorabia97, line 815: are we using data-parallel here? size is not set to any other parallelism. If so can you update test name to include `_dp` in it

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +757,218 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+        num_attention_heads=4,
+        ffn_hidden_size=64,
+        mamba_state_dim=16,
+        mamba_head_dim=8,
+        num_moe_experts=4,
+        moe_grouped_gemm=False,
+        moe_ffn_hidden_size=32,
+        moe_shared_expert_intermediate_size=16,
+        transformer_impl="modelopt",
+        expert_model_parallel_size=expert_model_parallel_size,
```
kevalmorabia97: can you move this ep size argument below tp_size so its easier to read

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +757,218 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+
+    if rank == 0:
+        Path(result_path).write_text(repr(summed_cost))
+
+
+@pytest.mark.skipif(not HAS_MAMBA, reason="Mamba not installed")
+def test_auto_quantize_mamba_hybrid_ep_cost(dist_workers, tmp_path):
+    """AutoQuantize cost must match for EP=1 and EP=2 when two GPUs are available."""
+    ep1_path = tmp_path / "ep1_cost.txt"
+    dist_workers.run(
```
kevalmorabia97, line 913: If 2 gpus are available, the first run here will still use 2 gpus resulting in DP=2 (is that okay)?

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +757,218 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+            expert_model_parallel_size=2,
+            result_path=str(ep2_path),
+        )
+    )
+    cost_ep1 = float(ep1_path.read_text())
+    cost_ep2 = float(ep2_path.read_text())
+    assert cost_ep1 == pytest.approx(cost_ep2, rel=1e-6)
+
+
+def _test_mcore_layerwise_calibration_layers_do_not_mutate_decoder(rank, size):
```
kevalmorabia97, line 936:
```suggestion
def _test_mcore_layerwise_calibration_layers_do_not_mutate_decoder_dp(rank, size):
```

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -712,6 +757,218 @@ def test_te_grouped_vs_sequential_quantize(dist_workers_size_4, quant_cfg):
...
+    assert discovered_layers is not None
+    assert len(discovered_layers) == decoder_len + 1
+    assert discovered_layers[-1] is output_layer
+    assert len(decoder_layers) == decoder_len
+    assert all(layer is not output_layer for layer in decoder_layers)
+
+
+def test_mcore_layerwise_calibration_layers_do_not_mutate_decoder(dist_workers):
+    if not HAS_TE:
+        pytest.skip("Transformer Engine is not installed")
```
kevalmorabia97: Not needed. All other tests also use modelopt or te spec and both assert TE requirement. So either add for all tests or remove for all (cleaner, same approach used everywhere)

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16-AutoQuantize-HSG-8192x512.yaml
jenchen13: thanks for catching that, fixed

tests/gpu_megatron/conftest.py
```diff
@@ -26,6 +26,22 @@
...
+    First-use JIT compilation can take minutes in CI, so build the base, FP8, and MX
+    extensions during session setup and let tests fall back to on-demand JIT if needed.
+    """
+    if torch.cuda.is_available():
+        with contextlib.suppress(Exception):  # best-effort; tests fall back to on-demand JIT
+            from modelopt.torch.quantization import extensions as quant_ext
+
+            quant_ext.get_cuda_ext(raise_if_failed=False)
+            quant_ext.get_cuda_ext_fp8(raise_if_failed=False)
+            quant_ext.get_cuda_ext_mx(raise_if_failed=False)
```
kevalmorabia97:
```suggestion
    import modelopt.torch.quantization.extensions as ext
    
    ext.precompile()
```

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
kevalmorabia97: can you check with claude for a better way to apply the timeout only for compress=True?

kevalmorabia97: Perhaps we can set to 120? Currently its 60 and I dont see it timeout often on any other PRs

kevalmorabia97: We can increase total to 75mins as we are anyways getting close to 60min timeout

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
jenchen13: TE spec I have a separate PR for this, would prefer to address that later because it requires changing a lot of tests https://github.com/NVIDIA/Model-Optimizer/pull/1550

jenchen13: "modelopt" is TE spec btw. "local" is local https://github.com/NVIDIA/Model-Optimizer/blob/main/tests/_test_utils/torch/megatron/models.py#L145

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
jenchen13: not intentionally, I would prefer to just test on 1 gpu

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
jenchen13: that's fine it only tests EP

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16-AutoQuantize-HSG-8192x512.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - TP: "1"
+      - ETP: "1"
+      - EP: "4"
+      - PP: "1"
+      - CP: "1"
+      - DP: "1"
+    slurm_config:
+      _factory_: "slurm_factory"
+      container: nvcr.io/nvidia/nemo:26.04
+      modelopt_install_path: /opt/TensorRT-Model-Optimizer/modelopt
```
kevalmorabia97:
```suggestion
      modelopt_install_path: /opt/venv/lib/python3.12/site-packages/modelopt
```

kevalmorabia97 (approved):
LGTM. Please also wait for 2-gpu tests to pass: https://github.com/NVIDIA/Model-Optimizer/actions/runs/28538480157

jenchen13:
/ok to test [2e5100c](https://github.com/NVIDIA/Model-Optimizer/pull/1562/commits/2e5100c70885362f69c0343fc62b421c9a4d7fa2)

modelopt/torch/quantization/algorithms.py
```diff
@@ -905,9 +916,14 @@ def run_search(self):
...
-            self.model.named_modules(), self._is_auto_quantize_module, self.config["cost"]
-        )
+        if self.candidate_stats:
+            total_weight_size = self._get_total_weight_size_from_candidate_stats(
+                self.candidate_stats
+            )
+        else:
+            total_weight_size = self._cost_model.total_weight_size(
+                self.model.named_modules(), self._is_auto_quantize_module, self.config["cost"]
+            )
```
realAsma: @meenchen  does this break cost_model interface support?

I don't think this breaks the `cost_model` interface. Fresh searches still instantiate/use the selected `_cost_model` to generate each candidate's `costs`; this branch only changes the denominator used after `candidate_stats` already exist.

That matters for EP because the no-quant entry in `candidate_stats` has already gone through the same per-layer cost path and EP synchronization as the quantized candidates, so summing those no-quant costs gives the full quantizable weight size for the active cost model. The fallback to `_cost_model.total_weight_size(...)` is still used when there are no candidate stats yet.

One useful follow-up would be to make that intent explicit in a short comment, because this does look like bypassing `cost_model` at first glance.

realAsma: > this branch only changes the denominator used after candidate_stats already exist.

What do you mean? How is ActiveMoE Cost denominator handled here?

Sorry, I should have phrased that more concretely.

For `active_moe`, the denominator here is still in the ActiveMoE cost space, not raw parameter count. During `insert_hparams_after_merge_rules()`, each hparam gets `cost_weight = self._cost_model.module_cost_weight(...)`; for routed MoE modules that is `active_moe_expert_ratio`, for shared/non-routed modules it is `1.0`, and excluded modules get `0.0`. Then `initialize_candidate_stats()` stores `hparam.get_cost(recipe)` for each recipe, and `get_cost()` multiplies module weight size by that `cost_weight` before applying the recipe compression and EP/TP sync.

So for ActiveMoE, the no-quant candidate cost is effectively:

```text
sum(routed_expert_weight_numel * active_moe_expert_ratio) + sum(shared/non_routed_weight_numel)
```

with the same exclusions and EP aggregation as the candidate costs used by the LP. This branch sums those no-quant entries to set the denominator, so the numerator and denominator stay in the same ActiveMoE-weighted cost units. The existing `test_auto_quantize_active_moe_cost_model` also asserts `cost_denominator == sum(stats["costs"][-1] ...)` and checks routed stats have `cost_weight == 0.25` while shared stats stay `1.0`.

modelopt/torch/quantization/algorithms.py
```diff
@@ -905,9 +916,14 @@ def run_search(self):
         )
 
         compression = self._get_formatted_weight_compression_constraint()
-        total_weight_size = self._cost_model.total_weight_size(
-            self.model.named_modules(), self._is_auto_quantize_module, self.config["cost"]
-        )
+        if self.candidate_stats:
```
realAsma: @jenchen13  why do we need the else branch here? `self.candidate_stats` is always set in `before_search` which gets called before `run_search`

realAsma: The else branch is never used.

modelopt/torch/quantization/algorithms.py
realAsma:
```suggestion
       total_weight_size = self._get_total_weight_size_from_candidate_stats(
                self.candidate_stats
            )
```

realAsma: Does this work?

modelopt/torch/quantization/algorithms.py
jenchen13: hmm I think i kept it for backward compatibility

jenchen13:
/ok to test [054e353](https://github.com/NVIDIA/Model-Optimizer/pull/1562/commits/054e35349b13ea3298ef517277b9abdaf49f440f)

tests/unit/torch/quantization/test_autoquant.py
```diff
@@ -203,24 +195,19 @@ def test_active_moe_cost_model_counts_fused_experts_without_weight():
...
     fused_experts.gate_up_proj = torch.nn.Parameter(torch.empty(2, 3, 5))
     fused_experts.down_proj = torch.nn.Parameter(torch.empty(2, 5, 3))
-    visual = torch.nn.Linear(4, 16)
     cost_model = get_auto_quantize_cost_model("active_moe")
+    cost_constraints = {
+        "active_moe_expert_ratio": 0.25,
+        EXCLUDED_MODULE_NAME_PATTERNS_KEY: ["*visual*"],
+    }
 
-    total_weight_size = cost_model.total_weight_size(
```
jenchen13, line 209: @meenchen can you please take a look at this? I had to remove `total_weight_size()` function since it is not used in auto quantize anymore

modelopt/torch/quantization/_auto_quantize_cost.py
```diff
@@ -156,19 +156,6 @@ def module_cost_weight(
             return 0.0
         return 1.0
 
-    def total_weight_size(
```
jenchen13, line 159: @meenchen see this too

tests/unit/torch/quantization/test_autoquant.py
meenchen: LGTM.

PR #1564, opened 2026-05-28
[OMNIML-4788] specdec_bench/Qwen3.5-4B: throughput_32k benchmark + S3 upload step

ChenhanYu:
### What does this PR do?

Type of change: enhancement (follow-up to [#1531](https://github.com/NVIDIA/Model-Optimizer/pull/1531)).

Extends the merged Qwen3.5-4B SPEED-Bench launcher YAMLs from a single-task qualitative-only smoke into a **3-task pipeline** that also covers long-context throughput and verifies the S3-upload path end-to-end. Two commits, cleanly cherry-picked from #1531's late branch state — they were authored after the merge-commit was resolved against an earlier rebased head and so didn't ride along with that merge.

### Pipeline shape (both YAMLs)

| Task | Split | Save dir |
|---|---|---|
| `task_0` | qualitative (existing quality / acceptance-rate signal) | `/scratchspace/specdec_bench{,_mtp}/qualitative` |
| `task_1` | **throughput_32k** (new — long-context throughput) | `/scratchspace/specdec_bench{,_mtp}/throughput_32k` |
| `task_2` | **upload to S3 in sweep layout** | `s3://team-specdec-workgroup/results/specdec_bench{,_mtp}/<split>/` |

### New artifacts

* `tools/launcher/common/specdec_bench/upload_to_s3.sh` — thin wrapper around `examples/specdec_bench/upload_to_s3.py` so it can be invoked as a launcher task. Installs `boto3` from `requirements.txt` on cold containers; warm pipelines pick it up from the prior `run.sh`.
* `tools/launcher/common/specdec_bench/runtime_params_throughput_32k.yaml` — pins `engine_args.max_model_len = 40,960` (32K input + 4K output + 4K headroom) so vLLM doesn't silently auto-cap `max_model_len` below the 36K minimum needed for `throughput_32k` prompts on single-GPU runs.

### Why max_model_len matters

Without an explicit `max_model_len`, vLLM auto-derives it from the model config (Qwen3.5-4B = 128K) **and from the GPU-memory budget**. On a single GPU the second factor can cap effective `max_model_len` well below 36K, silently truncating 32K-token prompts and producing wrong throughput numbers. The qualitative split is not affected (its prompts top out around 8K, well under any auto-derivation floor) so only `task_1` carries the override.

### S3 credentials

`upload_to_s3.sh` reads `S3_ENDPOINT` / `S3_KEY_ID` / `S3_SECRET` from the runtime environment (not hardcoded). `--skip-existing` + `--allow-incomplete-provenance` are passed by default so re-runs land alongside the prior upload, and runs lacking `CONTAINER_IMAGE` (Phase-2 harness work in OMNIML-4788 will populate it) still upload.

### Testing

Cluster smoke on cw_dfw via:

```
uv run slurm.py --yaml modules/Model-Optimizer/tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench.yaml --yes
```

is currently in-flight (jobs `12257378/79/80`, PD). Will update this PR with timing/AR numbers + S3 upload confirmation once it lands.

- Backward compatible: ✅ (additive — task_0 keeps the prior qualitative behavior, just with `/qualitative` suffix in `save_dir`)
- New PIP dep: ✅ no (boto3 already in `examples/specdec_bench/requirements.txt` from #1531)
- New tests: N/A (launcher YAML + shell wrapper; covered by cluster smoke)
- Changelog: N/A (internal-facing tooling)

## Summary by CodeRabbit

* **New Features**
  * Added a 32K-context runtime configuration (higher max model length) to enable long-context throughput benchmarking and avoid silent prompt truncation.
  * Added a launcher helper to upload benchmark results to S3 with incremental/retry-friendly options and pass/fail reporting.

* **Chores**
  * Split Qwen3.5-4B benchmark into separate qualitative and 32K throughput tasks and added coordinated S3 upload.
  * Applied the same multi-task pipeline layout and clearer output organization to the MTP speculative-decoding benchmark.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/Model-Optimizer/pull/1564?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tools/launcher/common/specdec_bench/runtime_params_throughput_32k.yaml
```diff
@@ -0,0 +1,14 @@
...
+# 32K-token prompts under tight GPU-memory budgets — vLLM's default
+# auto-derivation from the model config + gpu_memory_utilization can
+# cap max_model_len lower than 36K on a single GPU.
+#
+# Used by tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench{,_mtp}.yaml
+# task_1; the qualitative task uses the engine default (no override
+# needed, since its prompts top out at ~8K).
+
+engine_args:
+  max_model_len: 40960
```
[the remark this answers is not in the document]
ChenhanYu: Addressed in `80309cae01` (this PR). `examples/specdec_bench/specdec_bench/models/vllm.py:VLLMModel.__init__` now forwards any kwarg whose name matches an `AsyncEngineArgs` dataclass field — except the dozen kwargs `VLLMModel` consumes itself (collected into `_VLLM_CONSUMED_KWARGS` to prevent `got multiple values for keyword argument` crashes). So `engine_args.max_model_len: 40960` from `runtime_params_throughput_32k.yaml` now actually reaches `AsyncEngineArgs`.

Unit tests in `tests/examples/specdec_bench/test_vllm_kwargs_forwarding.py`:
- `max_model_len` is forwarded
- multiple engine_args fields (`max_model_len` + `dtype` + `gpu_memory_utilization`) all pass through
- consumed kwargs (e.g. `tensor_parallel_size`) are NOT double-forwarded (would crash `AsyncEngineArgs`)
- unknown kwargs are silently dropped (matches prior behaviour)
- a pin test asserts the test's copy of `_VLLM_CONSUMED_KWARGS` matches the module's definition

Cluster re-validation: planning to capture `dump_env`'s `serving_config.vllm_config.max_model_len` from the throughput_32k run dir on the next smoke to confirm it reads `40960`.

tools/launcher/examples/Qwen/Qwen3.5-4B/specdec_bench.yaml
```diff
@@ -45,3 +55,44 @@ pipeline:
...
+      - /scratchspace/specdec_bench
+      - s3://team-specdec-workgroup/results
+      - --skip-existing
+      - --allow-incomplete-provenance
+    slurm_config:
+      _factory_: "slurm_factory"
+      nodes: 1
+      ntasks_per_node: 1
+      gpus_per_node: 1
+      container: vllm/vllm-openai:qwen3_5-cu130
```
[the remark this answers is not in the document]
ChenhanYu: Addressed in `b90e30afc8` (this PR). Both `specdec_bench.yaml` and `specdec_bench_mtp.yaml` task_2 now use `gpus_per_node: 0`. Kept the same `vllm/vllm-openai:qwen3_5-cu130` image so the already-pulled cached image is reused across tasks — Slurm will still schedule the 0-GPU job on the default `batch` partition (no separate cpu partition needed; GPU nodes accept 0-GPU jobs and have spare CPU slots). Switching to a dedicated lightweight `python+boto3` image would shave the image-pull cost further but adds a maintenance surface, deferring that to a follow-up if image-pull becomes the bottleneck.

yeyu-nvidia (approved):

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -76,6 +95,20 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
...
+        # `runtime_params.engine_args` (e.g. `max_model_len`, `dtype`,
+        # `gpu_memory_utilization`) but that VLLMModel doesn't explicitly
+        # consume. Without this forwarding, `runtime_params.engine_args`
+        # values are silently dropped — the same bug the
+        # `runtime_params_throughput_32k.yaml::max_model_len: 40960` override
+        # was meant to fix. See PR #1564 review.
+        _engine_arg_fields = {f.name for f in dataclasses.fields(AsyncEngineArgs)}
+        forwarded_engine_kwargs = {
+            k: v for k, v in kwargs.items()
+            if k in _engine_arg_fields and k not in _VLLM_CONSUMED_KWARGS
```
h-guo18: I think there's a issue here. If some args in _VLLM_CONSUMED_KWARGS are also passed in engine args in its original keyword, then there will still be a error of `got multiple values for keyword argument`?

h-guo18: e.g. if prefix_cache or moe_expert_parallel_size is in engine args, it will not be de-duplicated here, therefore causing the error above.

h-guo18: I think a simpler way here is to fully respect the style of existing code: Only pass in what we need to AsyncEngineArgs. I think here we only need to append max_model_len in the arg list.

ChenhanYu: Good catch — you are right on all three points. Replaced the entire `_VLLM_CONSUMED_KWARGS` / `forwarded_engine_kwargs` machinery with a single explicit `max_model_len=kwargs.get("max_model_len")` in the `AsyncEngineArgs` call (commit c121ab7). Tests updated to match: removed the forwarding-logic tests, added three targeted tests covering the explicit path, the None-default case, and the absence of duplicate-kwarg errors.

tests/examples/specdec_bench/test_vllm_kwargs_forwarding.py
```diff
@@ -0,0 +1,124 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
h-guo18: These tests seems not necessary to keep since we do not have kwargs filters now. I would suggest remove this file for conciseness, but harmless to keep. Other changes lgtm.

h-guo18 (approved):

ChenhanYu:
/ok to test https://github.com/NVIDIA/Model-Optimizer/commit/56f46c8d67f9ff6e0f1df3ac653ad1b4cafde82a

tests/examples/specdec_bench/test_vllm_kwargs_forwarding.py
ChenhanYu: removed

PR #1569, opened 2026-05-29
[6078291][OMNIML-3716] Add ViT FP8 + Torch-TRT example, wire softmax_quantizer in _QuantAttention

ajrasane:
### What does this PR do?

Type of change: new feature + bug fix

Adds a Torch-TensorRT deployment path for HuggingFace ViT and closes the
modelopt-side gap that prevented `*softmax_quantizer` from being applied
on the standard attention forward path.

* **New ViT PTQ recipes** under `modelopt_recipes/huggingface/vit/ptq/`:
  * `fp8.yaml` — W8A8 per-tensor FP8 E4M3 on encoder Linear weights/inputs;
    attention Q/K/V BMMs + softmax output at FP8; per-block LayerNorm output
    at FP8 (one shared Q/DQ feeds Q/K/V + MLP); patch-embed `nn.Conv2d`,
    `classifier`, and the final `vit.layernorm` left FP16. Uses max
    calibration.
  * The recipe is self-contained (no `$import` of shared snippets) and
    use the "specific-enable" style: narrow `parent_class` + path scoping
    on every enable rule, so no `enable: false` carve-outs are needed.

* **New example** under `examples/torch_trt/`:
  * `torch_tensorrt_ptq.py` — single-model pipeline (load HF model,
    calibrate from `zh-plus/tiny-imagenet`, `mtq.quantize`,
    `torch_tensorrt.compile`, verify the compiled-model argmax matches the
    fake-quant argmax). Defaults to `google/vit-large-patch16-224`; pass
    `--model_id` and `--recipe` to target any model + recipe combination.
    `--no_pretrained` + `--model_kwargs` shrink the model for fast tests.
  * `README.md` documenting the flow, the shipped recipes, hardware
    requirements, and CLI usage.
  * `requirements.txt`.

* **Bug fix in `modelopt/torch/quantization/plugins/huggingface.py`** — inside
  `_QuantAttention._quantized_attention`, the non-kitchen branch now
  temporarily replaces `torch.nn.functional.softmax` (via the existing
  `replace_function` context manager) with a wrapper that pipes the softmax
  output through `self.softmax_quantizer`. Previously the slot was created
  on every registered attention class but only consumed by the optional
  Kitchen MXFP8 flash-attention path, so FP8 / NVFP4 recipes that enabled
  `*softmax_quantizer` saw it stay uncalibrated (`amax=None`) and emitted
  no Q/DQ around the softmax output during ONNX / Torch-TRT export. With
  this fix the `softmax_quantizer` is calibrated alongside the rest of
  the model, and both the modelopt ONNX exporter and `torch_tensorrt.compile`
  pick up the Q/DQ pair. The patch short-circuits to the unwrapped call
  when the quantizer is disabled (zero-overhead) and has no effect on SDPA
  paths that fuse softmax inside a C++ kernel.

* **New e2e integration test** at
  `tests/examples/torch_trt/test_torch_tensorrt_ptq.py` — mirrors the
  `torch_onnx` test pattern: invokes the example through
  `run_example_command`, parametrizes over the two precision modes (fp8,
  nvfp4), uses a 1-layer ViT config (`--no_pretrained` + `--model_kwargs`)
  so each parametrized case completes in under a minute. `importorskip` on
  `torch_tensorrt` so the test is automatically skipped on hosts without
  the package.

### Usage

```bash
# FP8 (Hopper / Ada) — default model is google/vit-large-patch16-224
python examples/torch_trt/torch_tensorrt_ptq.py \
    --precision fp8 \
    --calib_samples 128 \
    --batch_size 1

# Custom model + custom recipe
python examples/torch_trt/torch_tensorrt_ptq.py \
    --model_id <huggingface/model-id> \
    --recipe <recipe-path-relative-to-modelopt_recipes-or-absolute-yaml>
```

### Testing

* Recipes load via `modelopt.recipe.load_recipe()` and pass
  `QuantizeConfig` schema validation.
* Run `pytest tests/examples/torch_trt/test_torch_tensorrt_ptq.py` →
  1 parametrized case passes on RTX 6000 Ada (fp8).
* End-to-end on `google/vit-base-patch16-224`: `mtq.quantize` with the new
  FP8 recipe followed by `torch_tensorrt.compile(ir="dynamo")` produces a
  TRT engine whose argmax matches the FP16 baseline.
* ONNX exported from the torch path now contains Q/DQ on **12 / 12**
  softmax outputs (was 0 / 12 before this PR's `_QuantAttention` fix),
  matching the ONNX-CLI output's quantization layout.

  Both FP8 paths land within 0.13 pp Top-1 of the FP16 baseline; Top-5 is
  within 0.02 pp across all three.
* ImageNet-1k validation accuracy via the new `torch_tensorrt_accuracy.py`
  (full 50000 samples, batch=1, **every model Torch-TensorRT-compiled —
  including the baseline** — so the comparison is apples-to-apples) for the
  example's default `google/vit-large-patch16-224`:

  | Model (Torch-TRT) | Top-1 | Top-5 | Δ Top-1 vs baseline |
  |---|---:|---:|---:|
  | Baseline (FP16) | 81.99% | 96.01% | — |
  | FP8 | 82.01% | 96.05% | +0.02 pp |

  FP8 is within noise of the FP16 TRT baseline and NVFP4 W4A4 costs only
  −0.13 pp Top-1 / −0.05 pp Top-5. Absolute Top-1 sits below the model card's
  ~85.5% because evaluation uses the HF `AutoImageProcessor` default
  preprocessing (direct 224×224 resize, no resize-then-center-crop), applied
  identically to all three models — so the deltas are the comparison signal.

- Did you write any new necessary tests?: ✅ — new e2e integration test under `tests/examples/torch_trt/`.

## Summary by CodeRabbit

* **New Features**
  * Added Torch‑TensorRT FP8/NVFP4 deployment examples and end‑to‑end scripts for HuggingFace ViT, plus ViT-specific PTQ recipes and ImageNet-1k vs FP16 accuracy reporting.

* **Bug Fixes**
  * Fixed softmax quantization and export/compilation edge cases (softmax calibration during export, IO casting for empty tensors, routed expert weight syncing, importer key handling).

* **Documentation**
  * Added comprehensive example README with setup, usage, recipes, evaluation, and hardware guidance.

* **Requirements**
  * Pinned minimum versions for example dependencies.

* **Tests**
  * Added tests validating the Torch‑TensorRT quantization examples for fp8.


modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -134,6 +134,17 @@ def _quantized_attention(
         key_states = self.k_bmm_quantizer(key_states)
         value_states = self.v_bmm_quantizer(value_states)
         if not self.use_kitchen:
+            if self.softmax_quantizer.is_enabled:
```
cjluo-nv: > _Bot comment._

This new path is unreachable when `kitchen` is installed and `softmax_quantizer.is_enabled` is True with FP8 (E4M3) — i.e. exactly the configuration the two new ViT recipes set up. A few lines above:

```python
if kitchen is not None and self.kitchen_attn_fn is None:
    self._init_kitchen_attn_fn()
```

and `_init_kitchen_attn_fn` does:

```python
if self.softmax_quantizer.is_mxfp(8):
    ...
else:
    raise NotImplementedError(f"softmax_quantizer not supported: {self.softmax_quantizer}")
```

So on a host that has `kitchen` installed, calibrating the new `huggingface/vit/ptq/fp8.yaml` (or `nvfp4.yaml`, which also enables FP8 on `*softmax_quantizer`) raises `NotImplementedError` before this branch ever runs. The FP8/NVFP4 recipes need the kitchen path to fall through to the new wrapper-softmax path when the quantizer isn't MXFP8 — e.g. drop the `raise` and only set `use_kitchen=True` for the MXFP8 case, or skip `_init_kitchen_attn_fn` entirely when the quantizer is not MXFP8. As written, the new path is gated on `not self.use_kitchen` which is always False whenever kitchen init succeeded.

tests/examples/torch_trt/test_torch_tensorrt_ptq.py
```diff
@@ -0,0 +1,59 @@
...
+# vit.layernorm / classifier skip rules.
+_TINY_VIT_KWARGS = {
+    "num_hidden_layers": 1,
+    "hidden_size": 64,
+    "intermediate_size": 128,
+    "num_attention_heads": 2,
+}
+
+
+@pytest.mark.parametrize("precision", _PRECISIONS)
```
cjluo-nv: > _Bot comment._

Only `torch_tensorrt` is `importorskip`'d, but the new fix has a hidden coupling to whether `kitchen` is importable (see comment on `huggingface.py`). On a CI host that has both `torch_tensorrt` and `kitchen` installed, this test will fail at `mtq.quantize` time with `NotImplementedError: softmax_quantizer not supported: ...` for both parametrized cases. Worth either fixing the underlying kitchen-vs-FP8 ordering bug or adding `if kitchen is not None: pytest.skip(...)` so the regression mode is at least exercised in isolation.

examples/torch_trt/torch_tensorrt_ptq.py
```diff
@@ -0,0 +1,298 @@
...
+    # + [1,N,H] concat that runs once per forward, so falling back to
+    # PyTorch costs essentially nothing.
+    with export_torch_mode():
+        trt_model = torch_tensorrt.compile(
+            model,
+            ir="dynamo",
+            arg_inputs=[example_input],
+            min_block_size=1,
+            truncate_double=True,
+            torch_executed_ops={torch.ops.aten.cat.default},
```
cjluo-nv: > _Bot comment._

Minor: this `import torch_tensorrt` inside `compile_with_torch_tensorrt` is a local import without an explanatory comment. It's defensible here (the `--skip_trt` flag exists specifically so users without torch_tensorrt can still run quantize-only), but a one-line comment noting that — or moving it next to `--skip_trt` handling — would make the intent clearer.

cjluo-nv (commented):

Bug-fix portion has a regression risk on hosts that have the optional `kitchen` package installed: `_init_kitchen_attn_fn` runs *before* the new non-kitchen FP8 softmax path and raises `NotImplementedError` for any non-MXFP8 softmax quantizer. The new ViT FP8/NVFP4 recipes both enable plain E4M3 FP8 on `*softmax_quantizer`, so on a kitchen-equipped GPU they will fail before the new patched-softmax code path is reached. The new test uses `importorskip("torch_tensorrt")` but does not skip when `kitchen` is also installed, so this gap isn't caught in CI. Otherwise the design is reasonable (this is "new example + tuned recipes + 11-line plugin fix" rather than a new subsystem; the deterministic complexity gate fired only because of directory count) and the recipes/example look correct.

examples/torch_trt/README.md
```diff
@@ -0,0 +1,103 @@
...
+`torch_tensorrt.compile(ir="dynamo")` converts those Q/DQ nodes into native
+TensorRT precision layers.
+
+## Setup
+
+```bash
+# From the NVIDIA TensorRT docker image (recommended):
+docker run --gpus all -it --rm -v $(pwd):/workspace -w /workspace nvcr.io/nvidia/tensorrt:26.02-py3 bash
+
+pip install -U "nvidia-modelopt[torch]"
```
kevalmorabia97:
```suggestion
pip install -U "nvidia-modelopt"
```

examples/torch_trt/requirements.txt
```diff
@@ -0,0 +1,3 @@
+datasets>=2.14.4
+torch-tensorrt>=2.4.0
+transformers>=4.40
```
kevalmorabia97:
```suggestion
transformers>=4.56
```

examples/torch_trt/README.md
kevalmorabia97, line 1: directory missing in https://github.com/NVIDIA/Model-Optimizer/blob/main/.github/CODEOWNERS

kevalmorabia97 (approved):
approving as requirements.txt codeowner

tests/examples/torch_trt/test_torch_tensorrt_ptq.py
kevalmorabia97, line 1: Needs to be added to https://github.com/NVIDIA/Model-Optimizer/blob/main/.github/workflows/example_tests.yml to enable in cicd

kevalmorabia97 (changes requested):

modelopt/torch/quantization/plugins/huggingface.py
cjluo-nv: > _Bot comment._

Still unaddressed from the previous review. On a host with `kitchen` installed, `_init_kitchen_attn_fn` runs above this branch and raises `NotImplementedError` for any non-MXFP8 softmax quantizer — i.e. exactly the FP8 E4M3 configuration the two new ViT recipes enable. The new `if not self.use_kitchen:` guard is therefore dead code in the configuration this PR ships, and `mtq.quantize` will fail before reaching the wrapper-softmax path. Suggest restructuring `_init_kitchen_attn_fn` so non-MXFP8 quantizers fall through (don't set `use_kitchen=True` and drop the `raise`) instead of erroring out — that lets this new branch take over for FP8/NVFP4.

tests/examples/torch_trt/test_torch_tensorrt_ptq.py
cjluo-nv: > _Bot comment._

Still no guard on `kitchen`. With only `torch_tensorrt` `importorskip`'d, a CI host that happens to have both packages installed will fail both parametrized cases at `mtq.quantize` time with `NotImplementedError` due to the kitchen-vs-FP8 ordering bug in `huggingface.py`. Either fix the underlying bug (preferred) or add `if kitchen is not None: pytest.skip(...)` so this regression mode is exercised in isolation rather than silently masked.

examples/torch_trt/README.md
cjluo-nv: > _Bot comment._

Suggested change from previous review (`pip install -U "nvidia-modelopt"`) was not applied — this still reads `"nvidia-modelopt[torch]"`.

examples/torch_trt/requirements.txt
cjluo-nv: > _Bot comment._

Suggested bump to `transformers>=4.56` from previous review was not applied; still pinned at `>=4.40`.

cjluo-nv (commented):

Re-review — several previously-flagged issues still unresolved in the latest diff.

**Critical, still present:**
- **Kitchen-vs-FP8 ordering bug in `_QuantAttention._quantized_attention`** (cjluo-nv's earlier review). On hosts with the `kitchen` package installed, `_init_kitchen_attn_fn` runs *before* the new non-kitchen softmax path and raises `NotImplementedError` whenever `softmax_quantizer.is_enabled` is True and the quantizer is not MXFP8. Both new ViT recipes enable plain E4M3 FP8 on `*softmax_quantizer`, so calibrating either recipe on a kitchen-equipped GPU will fail at `mtq.quantize` time before the new wrapper-softmax branch is reachable. The new branch's `if not self.use_kitchen` guard is dead code in exactly the configuration this PR ships. Fix in `_init_kitchen_attn_fn` (don't set `use_kitchen=True` and don't raise for non-MXFP8 — just fall through to the new wrapper path), or skip kitchen init entirely when the quantizer isn't MXFP8.
- **CI gap reinforces the above**: the new test only `importorskip`s `torch_tensorrt`, not `kitchen`, so the regression mode isn't exercised in isolation either. Either fix the underlying ordering bug or add a `if kitchen is not None: pytest.skip(...)` guard.
- **`tests/examples/torch_trt/` is not wired into `.github/workflows/example_tests.yml`** (kevalmorabia97). I confirmed the workflow file still has the matrix `[llm_distill, llm_qat, llm_sparsity, diffusers_sparsity]` with no `torch_trt` entry, so the new e2e test won't run in CI even on PRs that touch it. The "tested in CI" claim in the PR body is currently not true on `main`.

**Minor, still unaddressed (verified against current diff):**
- `examples/torch_trt/README.md:19` still has `pip install -U "nvidia-modelopt[torch]"`; reviewer asked for `"nvidia-modelopt"`.
- `examples/torch_trt/requirements.txt:3` still pins `transformers>=4.40`; reviewer asked for `>=4.56`.
- `examples/torch_trt/` is missing from `.github/CODEOWNERS` — confirmed not added in this PR.
- The "argmax must match" claim in the test docstring (lines 44-46) is still informational only — `fq_match` / `trt_match` in `torch_tensorrt_ptq.py` (lines 277-294) are printed but never gate exit status, so the test passes whenever the pipeline runs to completion regardless of correctness. Either enforce the check (carefully — small `--no_pretrained` NVFP4 runs can flip argmax legitimately) or correct the docstring.

**Design note:** The deterministic complexity gate fired on directory count (5 dirs), but cjluo-nv correctly characterized this as "new example + tuned recipes + 11-line plugin fix" rather than a new subsystem. Recipes reuse the existing YAML recipe loader (`modelopt.recipe.load_recipe` / `ModelOptPTQRecipe`) and don't introduce a new abstraction. Design is fine; the blockers above are correctness/CI-coverage, not architecture.

---

- `.github/workflows/example_tests.yml:38` — > _Bot comment._

Still missing `torch_trt` from the example matrix (kevalmorabia97 raised this previously). Without an entry here, `tests/examples/torch_trt/test_torch_tensorrt_ptq.py` is not run in CI on PRs that touch `examples/torch_trt/` or `modelopt/torch/quantization/plugins/huggingface.py`, so the PR-body claim of 2 parametrized cases passing is hand-verified-only. Add `torch_trt` to the appropriate matrix (looks like the `onnx` group, since both run under `nvcr.io/nvidia/tensorrt:26.02-py3`).
- `.github/CODEOWNERS:49` — > _Bot comment._

Still missing `/examples/torch_trt` (kevalmorabia97). Add an entry alongside the other example owners so PRs touching this directory get the right reviewers.

examples/torch_trt/torch_tensorrt_ptq.py
```diff
@@ -0,0 +1,298 @@
...
+        return
+
+    wrapped = ViTLogitsWrapper(model).to(device).eval()
+    trt_model = compile_with_torch_tensorrt(wrapped, example_input)
+
+    print("\n=== Torch-TensorRT compiled ===")
+    with torch.no_grad():
+        trt_pred = trt_model(example_input).argmax(dim=-1)
+    trt_match = (trt_pred == baseline_pred).all().item()
+    print(f"TRT argmax class: {trt_pred.tolist()} (matches baseline: {trt_match})")
```
cjluo-nv, line 283: > _Bot comment._

`fq_match` / `trt_match` are computed and printed but never gate the exit status, so the example always exits 0 if the pipeline runs to completion. The new test's docstring (`test_torch_tensorrt_ptq.py:44-46`) claims the CLI "exits non-zero ... if the compiled-model argmax doesn't match the fake-quant argmax" — that's currently false. Either enforce the check (with care: tiny `--no_pretrained` ViT under NVFP4 can legitimately flip argmax on random input, so a hard gate may be flaky for that path) or update the test docstring to say the comparison is informational.

cjluo-nv (commented):

Re-review: all previously-flagged critical issues are resolved.

- **Kitchen-vs-FP8 ordering bug** fixed in `_init_kitchen_attn_fn`: the MXFP8 check is now folded into the early-return guard, so non-MXFP8 (e.g. plain FP8 E4M3) softmax quantizers fall through with `use_kitchen=False` and the new wrapper-softmax path in `_quantized_attention` actually runs. The new `if not self.use_kitchen` branch is now reachable on kitchen-equipped hosts, which is exactly the regression mode the previous review flagged.
- **CI wiring**: `torch_trt` is now in the `onnx` matrix in `.github/workflows/example_tests.yml`, and `/examples/torch_trt` is in `.github/CODEOWNERS` under the onnx codeowners team.
- **README/requirements nits**: install line is now `pip install -U "nvidia-modelopt"` and `transformers>=4.56`.
- **Test docstring**: now accurately describes the argmax comparison as informational (a tiny random-weight ViT under NVFP4 can legitimately flip argmax), so the test–behavior contract is consistent.
- **`torch_tensorrt_accuracy.py` baseline label**: the BF16/FP16 mismatch is gone — `--baseline` help text is generic, and the printed tag derives from `str(dtype)` so it always matches.

Design is appropriate: this is "new example + tuned recipes + 11-line plugin fix" reusing the existing recipe loader (`modelopt.recipe.load_recipe` / `ModelOptPTQRecipe`), not a new subsystem. The deterministic complexity gate fired only on directory count.

examples/torch_trt/README.md
```diff
@@ -0,0 +1,126 @@
+# ModelOpt + Torch-TensorRT Deployment
+
+End-to-end examples that quantize a PyTorch model with NVIDIA ModelOpt and
```
cjluo-nv: Could you discuss how this is different from ONNX PTQ in this readme?

examples/torch_trt/torch_tensorrt_ptq.py
ajrasane: Went with the "informational only" route: the test docstring now states explicitly that the baseline/fake-quant/TRT argmax comparison is informational and not gated (a tiny random-weight ViT under NVFP4 can legitimately flip argmax). See test_torch_tensorrt_ptq.py:43-46. Resolving.

tests/examples/torch_trt/test_torch_tensorrt_ptq.py
ajrasane: The underlying kitchen-vs-FP8 ordering bug was fixed in c41c51b56 so this regression mode no longer raises; the later duplicate of this comment is already resolved. Resolving this one too.

examples/torch_trt/README.md
ajrasane: Added `/examples/torch_trt @NVIDIA/modelopt-onnx-codeowners` to .github/CODEOWNERS (line 56). Resolving.

examples/torch_trt/README.md
ajrasane: Added a "How this differs from the ONNX examples" section in c0004d1f1 — a 3-way comparison table contrasting this Torch-TensorRT flow with torch_onnx and onnx_ptq (starting point, where quantization happens, export step, intermediate artifact, and runtime). Resolving.

cjluo-nv (commented):

Re-review: all previously-flagged critical issues are resolved in the latest diff.

- :speech_balloon: Kitchen-vs-FP8 ordering bug — fixed in `_init_kitchen_attn_fn` (`modelopt/torch/quantization/plugins/huggingface.py:85-89`): the early-return guard now folds in `not self.softmax_quantizer.is_mxfp(8)`, so plain FP8 E4M3 (the configuration both new ViT recipes ship) falls through with `use_kitchen=False` and the new wrapper-softmax path in `_quantized_attention` actually runs on kitchen-equipped hosts. The dead-code concern from the prior review no longer applies.
- :speech_balloon: CI wiring — `torch_trt` is now in the `onnx` matrix in `.github/workflows/example_tests.yml:101`, and `/examples/torch_trt` is in `.github/CODEOWNERS:56` under the onnx codeowners team. PR-body claim of "tested in CI" is now true.
- :speech_balloon: README install line + `transformers` pin updated to `pip install -U "nvidia-modelopt"` and `transformers>=4.56`.
- :speech_balloon: Test docstring (`tests/examples/torch_trt/test_torch_tensorrt_ptq.py:43-46`) now correctly states the baseline/fake-quant/TRT argmax comparison is informational — consistent with the script not gating on it.
- :speech_balloon: `torch_tensorrt_accuracy.py` baseline label now derives from `str(dtype)`, so the BF16/FP16 mismatch is gone.

Design is appropriate: this is "new example + tuned recipes + 11-line plugin fix" reusing the existing recipe loader (`modelopt.recipe.load_recipe` / `ModelOptPTQRecipe`), not a new subsystem. The deterministic complexity gate fired only on directory count.

Nudging for a human reviewer to sign off — the PR spans 7 directories and the design-review gate fired, so per the re-review guidance the architectural-scope check warrants a human.

cjluo-nv (approved):

kevalmorabia97 (approved):
New torch_trt test failing: https://github.com/NVIDIA/Model-Optimizer/actions/runs/27324305879/job/80722605185

CHANGELOG.rst
```diff
@@ -65,6 +65,7 @@ Changelog
...
+- Add Torch-TensorRT FP8 / NVFP4 deployment example for HuggingFace ViT (``examples/torch_trt/``): ``torch_tensorrt_ptq.py`` covers ``mtq.quantize`` → ``torch_tensorrt.compile(ir="dynamo")``, and ``torch_tensorrt_accuracy.py`` reports the compiled model's ImageNet-1k top-1/top-5 accuracy via the ``onnx_ptq`` ``evaluate`` harness (the unquantized baseline is Torch-TensorRT-compiled too, for an apples-to-apples comparison). Ships two ViT-tuned PTQ recipes under ``modelopt_recipes/huggingface/vit/ptq/`` (``fp8.yaml``, ``nvfp4.yaml``) composed from the shared ``modelopt_recipes/configs/`` units: FP8 quantizes the encoder Linears, patch-embed ``nn.Conv2d``, ``classifier``, and per-block LayerNorm inputs plus the attention Q/K/V BMMs and softmax; NVFP4 runs W4A4 on the encoder ``nn.Linear`` weights/inputs while holding the patch-embed ``nn.Conv2d``, ``classifier``, and attention BMMs/softmax at FP8. Verified on ``google/vit-base-patch16-224`` (ImageNet-1k 50k validation): FP8 stays within 0.13 pp Top-1 of the FP16 baseline.
```
kevalmorabia97: changelog entries would go to 0.46 section

PR #1571, opened 2026-05-29
feat: Layerwise calibration: nested config + QDQ-from-prev-layer flag + checkpoint I/O knobs

Fridah-nv:
### What does this PR do?

Type of change: new feature  

Groups all layerwise-calibration options under a nested `LayerwiseConfig` and adds three new behavior knobs to it. All changes are backward compatible.
#### 1. Nested `layerwise` config

  `QuantizeAlgorithmConfig.layerwise` changes from `bool` to a Pydantic submodel:

  ```python
  class LayerwiseConfig(ModeloptBaseConfig):
      enable: bool = False
      get_qdq_activations_from_prev_layer: bool = False
      checkpoint_dir: str | None = None
      save_every: int = 1
      save_quantizers_only: bool = False
  ```

  Backward compatibility:
  - `layerwise: True/False` still accepted (emits `DeprecationWarning`).
  - Flat `layerwise_checkpoint_dir` silently migrated into `layerwise.checkpoint_dir`.
  - Legacy `use_sequential` alias preserved (and resolved during flat-key migration so it can't be dropped).
  - Conflicting flat+nested `checkpoint_dir` values raise.
  - All 7 shipped PTQ recipes (`modelopt_recipes/general/ptq/*.yaml`, `huggingface/qwen3_5*/ptq/*.yaml`) migrated to the
  canonical nested shape — no semantic change.

  #### 2. `get_qdq_activations_from_prev_layer` — correct GPTQ vs max-calib semantics

  Controls what layer N's calibration sees:
  - **True** (GPTQ default): activations carry the quantize-dequantize error of layers 0..N-1 — GPTQ's Hessian-compensation 
  goal.
  - **False** (max/mse/local_hessian default): full-precision activations, matching the non-layerwise pass exactly.

  The False branch wraps the next-layer input capture forward with the existing `set_quantizer_by_cfg_context` deny-all idiom
  (`{"quantizer_name": "*", "enable": False}`).

  GPTQ's per-algorithm default is enforced via `@model_validator(mode="after")` that reads `LayerwiseConfig.model_fields_set` —
   works for every input shape (empty constructor, bool, partial dict, full dict) and lets explicit user values override.

  #### 3. `save_every` — gate the large activation-cache writes

  `save_every: int = 1` (`ge=1`). With N > 1, the per-layer `next_inputs.pt` (cached activation tensors, the largest checkpoint
   artifact for most models) is only written for the boundary layer of each N-layer window. Per-layer 
  weight/quantizer/output_meta files are still written every layer (resume needs them to replay skip layers correctly).
  Interrupting mid-window re-calibrates that window on resume.

  ####  **UPDATE: drop this in current PR and moved to https://github.com/NVIDIA/Model-Optimizer/pull/1640**
  #### 4. `save_quantizers_only` — algorithm-aware weight-blob skipping

  `save_quantizers_only: bool = False`. When True, skip `weights.pt` entirely and persist just the per-quantizer `state_dict`
  slice (carries `_amax`) to a new `quantizer_buffers.pt`. On resume, `full_restore` reloads only the quantizer slice and
  trusts that algorithm semantics didn't mutate `layer.weight`.

  Safety is enforced by a **whitelist**: `_supports_save_quantizers_only: ClassVar[bool] = False` on `QuantizeAlgorithmConfig`,
   overridden to `True` only on `MaxCalibConfig`, `MseCalibConfig`, `LocalHessianCalibConfig` (audited — these only touch
  `_amax`). Weight-mutating algorithms (GPTQ folds Hessian updates, AWQ/SmoothQuant fold pre-quant scales) reject the flag at
  config-construction time so in-place weight updates can't be silently lost on resume.

  ### Usage

  ```python

  # GPTQ — `get_qdq_activations_from_prev_layer` defaults to True (Hessian semantics).
  # save_every reduces activation-cache I/O.
  mtq.quantize(
      model,
      {
          "quant_cfg": [...],
          "algorithm": {
              "method": "gptq",
              "layerwise": {
                  "enable": True,
                  "checkpoint_dir": "/path/to/ckpts",
                  "save_every": 4,
              },
          },
      },
      forward_loop=forward_loop,
  )

  # Max-calibration — `get_qdq_activations_from_prev_layer` defaults to False (FP from prior layers).
  # save_quantizers_only skips the weights blob since max only updates _amax.
  mtq.quantize(
      model,
      {
          "quant_cfg": [...],
          "algorithm": {
              "method": "max",
              "layerwise": {
                  "enable": True,
                  "checkpoint_dir": "/path/to/ckpts",
                  "save_quantizers_only": True,
              },
          },
      },
      forward_loop=forward_loop,
  )
  ```

  ### Testing

  New / updated unit tests in `tests/unit/torch/quantization/`:

  - **`test_config_validation.py`** — `TestLayerwiseNestedConfig` covers nested-form acceptance, bool-form
  `DeprecationWarning`, flat `layerwise_checkpoint_dir` migration, conflicting flat+nested checkpoint_dir, `use_sequential`
  alias survival under migration, per-algorithm qdq defaults (parametrized Max/GPTQ), `save_every` `ge=1` validation, and the 
  `save_quantizers_only` whitelist — parametrized rejection on `[GPTQ, AWQLite, SmoothQuant]` and acceptance on `[Max, Mse, 
  LocalHessian]`.

  - **`test_layerwise_calibrate.py`** —
    - `test_layerwise_no_qdq_matches_sequential_amax` — behavioral equivalence: layerwise + `qdq=False` produces the same 
  per-quantizer `_amax` as the non-layerwise (sequential) max-calibration flow (verified via `torch.testing.assert_close`).
    - `test_layerwise_save_every_writes_next_inputs_only_at_window_boundaries` — window-save layout (all layer dirs present, 
  `next_inputs.pt` only at boundaries).
    - `test_layerwise_save_quantizers_only_resume_matches_one_shot_amax` — end-to-end resume: full run → manifest rewound →
  fresh model resumes → final `_amax` matches the one-shot baseline; also pins the on-disk shape (no `weights.pt`,
  `quantizer_buffers.pt` present).

## End-to-end correctness verification

  Ran 4 PTQ jobs on **Qwen3-8B** with NVFP4 W4A16 quant_cfg and `--calib_size 16`, one GPU
  each on 4 GPUs 

  | Run | Algorithm | Layerwise config | Purpose |
  |---|---|---|---|
  | **A** | GPTQ | `enable=true, get_qdq_activations_from_prev_layer=true, save_every=5` | New nested form + the new`save_every` knob |
  | **B** | MSE | `enable=true, get_qdq_activations_from_prev_layer=false, save_quantizers_only=true` | New nested form  + the new `save_quantizers_only` knob |
  | **C** | GPTQ | Legacy flat form: `layerwise: true, layerwise_checkpoint_dir: ...` | Backward-compat baseline for A (exercises the migration validator) |
  | **D** | MSE | `enable=false` | Non-layerwise baseline for B |

 Across two pairwise comparisons (A vs C ; B vs D), all 905 tensors in the exported HF checkpoints are bit-identical with hf_quant_config.json matching exactly — confirming the new layerwise knobs preserve correctness and the flat-form backward-compat path is intact.

  Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md)
  and your commits are signed (`git commit -s -S`).

  Make sure you read and follow the [Security Best 
  Practices](https://github.com/NVIDIA/Model-Optimizer/blob/main/SECURITY.md#security-coding-practices-for-contributors) (e.g.
  avoiding hardcoded `trust_remote_code=True`, `torch.load(..., weights_only=False)`, `pickle`, etc.).

  - Is this change backward compatible?: ✅ — `layerwise: True/False` still accepted with a `DeprecationWarning`; flat
  `layerwise_checkpoint_dir` silently migrated; `use_sequential` alias preserved. The two new knobs default to no-op behavior
  (`save_every=1`, `save_quantizers_only=False`).
  — no new dependencies.
  - Did you write any new necessary tests?: ✅ — see Testing section.
  - Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?:  ✅ 
  ready for review.
  - Did you get Claude approval on this PR?: ✅ — `/claude review` consulted iteratively; review findings (GPTQ default
  survival, save_quantizers_only whitelist scope, docstring accuracy, dead `layer` param) addressed in-PR.

  ### Additional Information

  Notes on design choices that came out of internal review:
  - GPTQ's `qdq=True` default uses a `model_validator(mode="after")` + `model_fields_set` check rather than a `default_factory`
   — a `default_factory` is only fired when the field is absent, so any user-supplied dict (the natural way to enable
  layerwise) would silently lose the GPTQ default.
  - `save_quantizers_only` is enforced as a whitelist (`_supports_save_quantizers_only`) rather than a per-algorithm blacklist,
   which keeps future weight-mutating algorithms safe by default.
  - `set_quantizer_by_cfg_context` is reused for the qdq-disable scope instead of a bespoke helper, keeping `model_calib.py`
  aligned with the existing "deny-all" idiom documented at `conversion.py:240`.
  - Pre-validation recipe helpers in `examples/llm_ptq/example_utils.py` (`needs_checkpoint_path_update` /
  `resolve_checkpoint_dir`) accept both flat and nested shapes since they run before Pydantic validation;
  `resolve_checkpoint_dir` now also returns the resolved path so the caller doesn't re-derive it.

## Summary by CodeRabbit

* **New Features**
  * Layerwise calibration now uses a nested config (enable, get_qdq_activations_from_prev_layer, checkpoint_dir, save_every, save_quantizers_only); checkpointing supports quantizer-only saves and windowed commits with resume support.
* **Behavior**
  * GPTQ calibration defaults get_qdq_activations_from_prev_layer=True when unspecified; save_every controls when next-inputs/manifests are written and must be positive.
* **Deprecations**
  * Legacy flat-style layerwise keys are still accepted but emit DeprecationWarning and are auto-migrated (conflicts detected).
* **Examples/UX**
  * Tools auto-detect legacy vs nested checkpoint layouts and report the resolved path.
* **Tests**
  * Expanded coverage for nested configs, validation, checkpoint/resume semantics, and windowed saves.


modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -630,55 +669,101 @@ def full_restore(self, layers: nn.ModuleList, model: nn.Module) -> None:
...
-            self.checkpoint_dir,
-            layer_idx,
-            weights,
-            qstate,
-            _move_to_device(output_meta, _cpu),
-            _move_to_device(next_layer_inputs, _cpu) if next_layer_inputs is not None else None,
-            self.num_layers,
-        )
+        window_start = self._last_saved_layer + 1
+        for i in range(window_start, layer_idx + 1):
```
cjluo-nv: > _Bot comment._

**Critical: window save reads from `_SkipLayer` stand-ins.**

When `save_every > 1`, this loop iterates `range(window_start, layer_idx + 1)` and calls `layers[i].state_dict()` / `quantizer_state(layers[i])` / `getattr(layers[i]._layerwise_calib, 'output_meta', None)` for each `i` in the window. But by the time `save()` is reached for the boundary layer:

- `cache_outputs_for_next_layer_calib(layer)` → `get_input_activations(next_layer)` → `_set_layer_states(layer_idx + 1)` has already run, which swaps `transformer_layers[layer_idx - 1]` to a `_SkipLayer`.
- All earlier in-window layers (`layer_idx - 2`, `layer_idx - 3`, …) were swapped to `_SkipLayer` by previous iterations.

`_SkipLayer` holds `_original` via `object.__setattr__`, so it is not a registered submodule. Consequently:

- `_SkipLayer.state_dict()` returns essentially `{}` — `weights.pt` for non-boundary layers contains no real weights.
- `quantizer_state(_SkipLayer)` finds no `TensorQuantizer`s — `quantizer_state.pt` is empty.
- `_SkipLayer._layerwise_calib.output_meta` is `None` (the SkipLayer has its own fresh `_LayerCalibState`), so the placeholder `torch.zeros(1)` meta is written, which then mis-shapes `_zeros_from_meta` outputs after resume.
- Same problem for `save_quantizers_only=True`: `get_quantizer_state_dict(_SkipLayer)` returns `{}`.

The new `test_layerwise_save_every_writes_next_inputs_only_at_window_boundaries` only checks file presence; it doesn't read the saved tensors back, and there is no resume test for `save_every > 1`, so the corruption is silent. Please add a test that resumes from a `save_every > 1` checkpoint and asserts amax/parameter equivalence with a one-shot run (the same shape as `test_layerwise_save_quantizers_only_resume_matches_one_shot_amax`, but with `save_every` set on a model where the boundary lands mid-stream).

Suggested fix: snapshot each layer's `(weights | quantizer_buffers, qstate, output_meta)` to CPU immediately after calibrating it (or before `_set_layer_states` swaps it), buffer those snapshots in `_CheckpointState`, and only at window boundaries flush the buffered snapshots plus `next_inputs.pt` to disk. That keeps the I/O-batching benefit without reading state from `_SkipLayer`s.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -572,7 +602,13 @@ def from_folder(cls, checkpoint_dir: str | None, num_layers: int) -> _Checkpoint
...
                 f"Checkpoint: resuming layerwise calibration from layer {start}/{num_layers}"
             )
-        return cls(checkpoint_dir, num_layers, start_layer=start)
+        return cls(
+            checkpoint_dir,
+            num_layers,
+            start_layer=start,
+            save_every=save_every,
+            save_quantizers_only=save_quantizers_only,
+        )
```
cjluo-nv, line 613: > _Bot comment._

Cross-version compatibility: when resuming an existing checkpoint produced with a different `save_every` / `save_quantizers_only` value than the user passes on resume, `_last_saved_layer = start_layer - 1` plus the new `save_every` won't replay correctly if the previous run's final boundary didn't coincide with the new window. Consider persisting `save_every` and `save_quantizers_only` in `manifest.json` and either reusing them or hard-erroring on mismatch (similar to the existing `num_layers` mismatch check). Today there's no signal that the on-disk layout might be inconsistent with the new config.

modelopt/torch/quantization/config.py
```diff
@@ -996,6 +1113,21 @@ class GPTQCalibConfig(QuantizeAlgorithmConfig):
...
+    def _gptq_qdq_default(self):
+        """Inject ``get_qdq_activations_from_prev_layer=True`` unless the user set it.
+
+        GPTQ's Hessian correctness depends on prior-layer QDQ activations, so the
+        default differs from the base class. Uses ``model_fields_set`` to detect
+        whether the user explicitly set the field — covers every input shape
+        (empty constructor, bool, dict) without a per-shape special case.
+        """
+        if "get_qdq_activations_from_prev_layer" not in self.layerwise.model_fields_set:
+            self.layerwise = self.layerwise.model_copy(
```
cjluo-nv, line 1151: > _Bot comment._

Nit: `self.layerwise = self.layerwise.model_copy(update=...)` reassigns the `layerwise` attribute after model construction. With Pydantic v2's default `model_config`, this re-runs validation on the assignment only if `validate_assignment=True`. `ModeloptBaseConfig` may or may not enable that — worth a unit test that constructs `GPTQCalibConfig(layerwise={"enable": True, "save_quantizers_only": True})` and asserts the whitelist validator still rejects it (i.e. that `_validate_save_quantizers_only_supported` and `_gptq_qdq_default` interleave the way you want). The `model_validator(mode='after')` order on a subclass also only runs after the parent's, so `_validate_save_quantizers_only_supported` already saw the pre-update `layerwise` — that's fine here, but worth pinning with a test.

cjluo-nv (commented):

The nested `LayerwiseConfig` refactor, `get_qdq_activations_from_prev_layer` flag, and `save_quantizers_only` whitelist look well-designed and the back-compat shims (bool coercion + flat→nested migration with the `use_sequential` rescue) are thorough. Tests around config validation and the qdq=False-matches-sequential-amax behavioral check are good.

However, **`save_every > 1` looks broken**: by the time `_CheckpointState.save()` is invoked for the boundary layer, the earlier layers in the window have already been replaced by `_SkipLayer` instances in `transformer_layers` (via `_set_layer_states(layer_idx+1)` swapping `layer_idx-1` to a dummy during `cache_outputs_for_next_layer_calib`). Reading `layers[i].state_dict()` / `quantizer_state(layers[i])` / `layers[i]._layerwise_calib.output_meta` on a `_SkipLayer` returns empty/None state because `_original` is attached via `object.__setattr__` and is not a registered submodule. The new `test_layerwise_save_every_writes_next_inputs_only_at_window_boundaries` only checks file existence, so the silent corruption isn't caught — there is no resume test for `save_every > 1`. See inline comment for details and the suggested fix (snapshot to memory immediately after each layer is calibrated, then flush at boundaries; or save real layers eagerly and only window-gate `next_inputs.pt`).

Minor: changelog says "TODO before marking ready for review" in the PR body but the changelog entry is already added — please confirm the description is up to date before merge.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -630,55 +669,101 @@ def full_restore(self, layers: nn.ModuleList, model: nn.Module) -> None:
...
+
+            _save_layer(
+                self.checkpoint_dir,
+                i,
+                weights,
+                qstate,
+                quantizer_buffers,
+                _move_to_device(output_meta, _cpu),
+                _move_to_device(next_layer_inputs, _cpu)
+                if is_last_in_window and next_layer_inputs is not None
```
cjluo-nv: > _Bot comment._

**Critical (carried over, still unresolved):** the window-save loop iterates `layers[i]` for `i` in `[window_start, layer_idx]`. By the time `save()` runs, `cache_outputs_for_next_layer_calib(layer)` has already invoked `_set_layer_states(layer_idx + 1)`, which swaps every layer in `0..layer_idx-1` to a `_SkipLayer` (the swap at `layer_idx - 1` happens *now*; earlier swaps happened in prior iterations). For all `i < layer_idx`:

- `layers[i].state_dict()` returns ~`{}` — `_original` is attached via `object.__setattr__`, so it's not a registered submodule. `weights.pt` for non-boundary layers in the window contains no real weights.
- `quantizer_state(layers[i])` finds no `TensorQuantizer` descendants for the same reason. `quantizer_state.pt` is empty.
- `getattr(layers[i]._layerwise_calib, 'output_meta', None)` reads the `_SkipLayer`'s own fresh `_LayerCalibState`, which has `output_meta=None`, so the placeholder `torch.zeros(1)` meta is written and `_zeros_from_meta` will produce a wrong-shape dummy on resume.
- Same problem for `save_quantizers_only=True`: `get_quantizer_state_dict(layers[i])` returns `{}`.

The new `test_layerwise_save_every_writes_next_inputs_only_at_window_boundaries` only asserts file existence, and `test_layerwise_save_quantizers_only_resume_matches_one_shot_amax` uses default `save_every=1`, so the corruption is silent.

Fix sketch (matches the prior comment): snapshot `(weights | quantizer_buffers, qstate, output_meta)` to CPU immediately after each layer is calibrated (before `_set_layer_states` swaps it), buffer the snapshots in `_CheckpointState`, and only at window boundaries flush the buffered snapshots plus `next_inputs.pt`. Or, in this loop, unwrap `layer_i = layers[i]._original if isinstance(layers[i], _SkipLayer) else layers[i]` before reading state — but the snapshot approach is more robust against future state-machine changes.

Please also add a regression test that resumes from a `save_every > 1` checkpoint and asserts amax/parameter equivalence with a one-shot run.

modelopt/torch/quantization/config.py
```diff
@@ -633,9 +633,88 @@ def validate_calibrator(cls, v, info: ValidationInfo):
...
+            "Passing the layerwise field as a bool is deprecated; use a dict, "
+            "e.g. `{'enable': True}`.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return {"enable": value}
+    if value is None:
+        return {}
+    if isinstance(value, LayerwiseConfig):
+        return value.model_dump()
```
cjluo-nv: > _Bot comment._

**Major (carried over from CodeRabbit, still unresolved):** for `isinstance(value, LayerwiseConfig)`, this returns `value.model_dump()`, which expands all default fields. After re-parse, `self.layerwise.model_fields_set` will contain `"get_qdq_activations_from_prev_layer"` even when the caller only set `enable=True`, and `GPTQCalibConfig._gptq_qdq_default` will then incorrectly skip injecting the GPTQ-correct `True` default.

Reproducer: `GPTQCalibConfig(layerwise=LayerwiseConfig(enable=True)).layerwise.get_qdq_activations_from_prev_layer` returns `False` (should be `True`). The existing parametrized `test_gptq_qdq_default_respects_user_explicit_value` covers dict/bool/None inputs but not a `LayerwiseConfig` instance, so the regression is missed.

Suggested fix: `return value.model_dump(exclude_unset=True)` — or just `return value` and rely on Pydantic's nested-model coercion to preserve `model_fields_set`.

cjluo-nv (commented):

Re-review: the previously flagged **critical** `save_every > 1` corruption bug is still present in `modelopt/torch/quantization/utils/layerwise_calib.py`. The new test only checks file existence, not file contents, so the corruption remains silent. Other previous comments (manifest update during partial-window crash, `_coerce_layerwise_input` `model_dump()` losing `model_fields_set`) are also unaddressed. Marking `comment` rather than `nudge` because the critical bug from the previous review is unresolved.

Status of previous comments:
- **(critical, UNRESOLVED)** `save_every > 1` reads from `_SkipLayer` stand-ins → silent on-disk corruption. Still no resume test for `save_every > 1`.
- **(critical, UNRESOLVED)** Manifest is advanced per-`_save_layer` call, so a mid-window crash leaves `manifest=K` with no `next_inputs.pt` for layer `K`, breaking resume permanently.
- **(major, UNRESOLVED)** `_coerce_layerwise_input` calls `.model_dump()` on a `LayerwiseConfig` instance, which expands defaults and breaks `_gptq_qdq_default`'s reliance on `model_fields_set`.
- **(minor)** Cross-version `save_every` / `save_quantizers_only` mismatch on resume — still no manifest-side guard.
- **(minor)** `__all__` declaration in `config.py` — still missing.
- **(minor)** Symmetric `DeprecationWarning` for flat `layerwise_checkpoint_dir` migration — still silent.
- **(minor)** `resolve_checkpoint_dir` flat+nested redundancy — still no guard.

Resolved/no-longer-flagging:
- Changelog entry is present and the "TODO before marking ready" note appears to have been dropped from the PR body.

Design-review pass: the PR refactors an existing in-repo subsystem (layerwise calibration + checkpoint format) rather than introducing a new one, so the design protocol's "is there an existing tool that already does this?" question doesn't apply — the existing tool *is* what's being refactored. The nested-config + whitelist + per-algorithm default approach is reasonable; correctness, not design, is the issue here.

---

- `modelopt/torch/quantization/utils/layerwise_calib.py:522` — > _Bot comment._

**Critical (carried over, still unresolved):** `_save_layer` calls `_write_manifest(checkpoint_dir, idx, num_layers)` at the end of every individual call, but during a `save_every > 1` window save the *non-boundary* layers in the window are written without a `next_inputs.pt`. If the process is killed (or `torch.save` raises) between the first `_save_layer` and the boundary `_save_layer` within a window, the manifest advances to a layer that has no `next_inputs.pt`. On resume, `setup_resume` then raises `FileNotFoundError("Cannot resume: next_inputs.pt missing for layer N")` and the checkpoint is permanently broken — the PR description's claim that "interrupting mid-window re-calibrates that window on resume" is not what the code does.

Fix: drop the `_write_manifest` call from `_save_layer` and call it once at the end of `_CheckpointState.save()` (with `last_completed_layer = layer_idx`) so the manifest only advances after the boundary `next_inputs.pt` is on disk. A regression test that monkeypatches `_write_manifest` to raise mid-window would lock this in.

modelopt/torch/quantization/utils/layerwise_calib.py
Fridah-nv: Fixed.

modelopt/torch/quantization/config.py
Fridah-nv: Confirmed bug, fixed.

modelopt/torch/quantization/config.py
```diff
@@ -152,9 +152,9 @@
...
 import warnings
 from collections.abc import Mapping, Sequence
-from typing import Any, Literal
+from typing import Any, ClassVar, Literal
 
-from pydantic import AliasChoices, ValidationInfo, field_validator, model_validator
+from pydantic import AliasChoices, Field, ValidationInfo, field_validator, model_validator
 
 from modelopt.torch.opt.config import ModeloptBaseConfig, ModeloptField
 from modelopt.torch.opt.config_loader import load_config
```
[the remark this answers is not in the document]
Fridah-nv: Out of scope for current PR

modelopt/torch/quantization/config.py
```diff
@@ -633,9 +633,88 @@ def validate_calibrator(cls, v, info: ValidationInfo):
...
+            "e.g. `{'enable': True}`.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return {"enable": value}
+    if value is None:
+        return {}
+    if isinstance(value, LayerwiseConfig):
+        return value.model_dump()
+    return value
```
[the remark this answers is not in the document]
Fridah-nv: Fixed

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -630,55 +669,101 @@ def full_restore(self, layers: nn.ModuleList, model: nn.Module) -> None:
...
+                else None,
+                self.num_layers,
+            )
```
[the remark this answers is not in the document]
Fridah-nv: Fixed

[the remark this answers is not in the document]
Fridah-nv: Fixed alongside the _SkipLayer snapshot fix.

examples/llm_ptq/example_utils.py
```diff
@@ -850,22 +850,37 @@ def copy_custom_model_files(source_path: str, export_path: str, trust_remote_cod
...
+        return None
+    flat = algorithm.get("layerwise_checkpoint_dir")
+    if flat is not None:
+        return "flat", flat
+    nested = algorithm.get("layerwise") or {}
+    ckpt = nested.get("checkpoint_dir") if isinstance(nested, dict) else None
+    return ("nested", ckpt) if ckpt is not None else None
+
+
+def needs_checkpoint_path_update(quant_cfg: dict) -> bool:
```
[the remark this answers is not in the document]
Fridah-nv: Applied. `resolve_checkpoint_dir` now writes the resolved path back to both shapes when both are present in the input (the Pydantic migration validator already accepts matching values), so the post-resolve dict stays internally consistent and the subsequent validator doesn't see a stale value.

modelopt/torch/quantization/utils/layerwise_calib.py
Fridah-nv: Good catch, applied. _write_manifest now persists save_every and save_quantizers_only next to num_layers, and _CheckpointState.from_folder checks all three on resume, raising "Use a fresh checkpoint directory" on any mismatch

modelopt/torch/quantization/config.py
```diff
@@ -656,34 +735,66 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
...
+        raw_layerwise = data.pop("layerwise", data.pop("use_sequential", None))
+        layerwise = _coerce_layerwise_input(raw_layerwise)
+        existing = layerwise.get("checkpoint_dir")
+        if existing is not None and existing != flat_dir:
+            raise ValueError(
+                f"Conflicting checkpoint_dir: layerwise_checkpoint_dir={flat_dir!r} "
+                f"differs from layerwise.checkpoint_dir={existing!r}. Set only one."
+            )
+        data["layerwise"] = {**layerwise, "checkpoint_dir": flat_dir}
+        return data
```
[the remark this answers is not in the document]
Fridah-nv: Appiled

cjluo-nv (commented):

Re-review: all previously-flagged critical issues are resolved with regression tests.

- **`save_every > 1` `_SkipLayer` corruption (critical, was unresolved)** — Fixed. `_CheckpointState.save()` now reads from `layers[layer_idx]` (the just-calibrated layer) and is called every iteration before the next-layer capture forward swaps prior layers to `_SkipLayer`. Per-layer files (weights/quantizer state/output_meta) are written eagerly; only `next_inputs.pt` + manifest are deferred to window boundaries.
- **Mid-window crash advancing manifest (critical, was unresolved)** — Fixed. `_save_layer_files` no longer writes the manifest; `_CheckpointState.save()` writes it once at boundary/final, so a torch.save crash mid-window leaves the manifest pointing at the previous boundary. Covered by `test_layerwise_save_every_mid_window_crash_recovers_at_prev_boundary` (monkeypatches torch.save to crash on layer 2 mid-window, asserts manifest stays at layer 1).
- **`_coerce_layerwise_input` `.model_dump()` expanding defaults (critical/major, was unresolved)** — Fixed. Now uses `model_dump(exclude_unset=True)` so `LayerwiseConfig` instance inputs preserve `model_fields_set`, and GPTQ's `_gptq_qdq_default` correctly injects `True` for fields the user didn't explicitly set. Test parametrization extended to cover `LayerwiseConfig(enable=True)` + `LayerwiseConfig(enable=True, get_qdq_activations_from_prev_layer=False)`.
- **Save_every>1 resume parity test (was missing)** — Added. `test_layerwise_checkpoint_resume_matches_one_shot_amax` parametrized over `(save_every=2, save_quantizers_only=False)` and `(save_every=1, save_quantizers_only=True)` rewinds the manifest and asserts amax equivalence with a one-shot baseline — i.e. it would fail under the previous corruption.
- **Cross-version `save_every`/`save_quantizers_only` mismatch (minor)** — Addressed. Manifest now persists both keys; `from_folder` raises on any mismatch. Covered by `test_layerwise_checkpoint_mismatch_save_every_raises`.
- **Symmetric DeprecationWarning for flat `layerwise_checkpoint_dir` (minor)** — Addressed; warning emitted in `_migrate_layerwise_checkpoint_dir`.
- **`resolve_checkpoint_dir` flat+nested redundancy (minor)** — Addressed; resolver writes back to both shapes when both present, keeping the dict internally consistent.
- **`__all__` declaration (minor)** — Author marked out of scope; acceptable.

Design-review pass: this PR refactors an existing in-repo subsystem (layerwise calibration + checkpoint format) rather than introducing a new one, so the "is there an existing tool that already does this?" question doesn't apply — the existing tool is what's being refactored. The nested-config + whitelist + per-algorithm default approach is reasonable.

Complex PR: spans 8 directories (≥ 5). Looping in a human for approval.

modelopt/torch/quantization/model_calib.py
```diff
@@ -1811,12 +1828,25 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+                    nullcontext()
+                    if qdq_from_prev
+                    else set_quantizer_by_cfg_context(
+                        layer, [{"quantizer_name": "*", "enable": False}]
+                    )
+                )
+                with capture_ctx:
+                    next_inputs = input_getter.cache_outputs_for_next_layer_calib(
+                        layer, forward_loop
+                    )
```
realAsma: This is not correct. even if we disable quantizers, some algorithms such as GPTQ modify the weights.

can we do input caching  before calling calib_func if qdq_from_prev is false?

Fridah-nv: Yes, this is a real gap. Updated the code.

cjluo-nv (commented):

Re-review: all four previously-flagged **critical** issues are resolved with meaningful regression tests.

- :speech_balloon: **`save_every > 1` `_SkipLayer` corruption (critical, was unresolved)** — Fixed. `_CheckpointState.save()` now reads from `layers[layer_idx]` (the just-calibrated layer) before the next iteration's capture forward swaps prior layers to `_SkipLayer`. Per-layer files (weights/quantizer state/output_meta) are written eagerly; only `next_inputs.pt` + manifest are deferred to window boundaries. Covered by the parametrized `test_layerwise_checkpoint_resume_matches_one_shot_amax(save_every=2)` parity test against a one-shot baseline.
- :speech_balloon: **Mid-window crash advancing manifest (critical, was unresolved)** — Fixed. `_save_layer_files` no longer writes the manifest; `_CheckpointState.save()` writes it once at boundary/final, so a `torch.save` crash mid-window leaves the manifest pointing at the previous boundary. Covered by `test_layerwise_save_every_mid_window_crash_recovers_at_prev_boundary` (monkeypatches `torch.save` to crash on layer 2 mid-window, asserts manifest stays at layer 1).
- :speech_balloon: **`_coerce_layerwise_input` `.model_dump()` expanding defaults (critical, was unresolved)** — Fixed. Now uses `model_dump(exclude_unset=True)` so `LayerwiseConfig` instance inputs preserve `model_fields_set`, and GPTQ's `_gptq_qdq_default` correctly injects `True` for fields the user didn't explicitly set. Test parametrization extended to cover `LayerwiseConfig(enable=True)` and `LayerwiseConfig(enable=True, get_qdq_activations_from_prev_layer=False)`.
- :speech_balloon: **`qdq_from_prev=False` weight-mutation leak (was raised by realAsma)** — Fixed. Next-layer input capture moved to *before* `calib_func` when `qdq_from_prev=False`, so weight-mutating algorithms (GPTQ/AWQ/SmoothQuant folds) can't leak into downstream captures. Covered by `test_layerwise_no_qdq_captures_inputs_before_calib_func_mutates_weights` (destructive vs identity calib_func produce identical captured inputs).
- :speech_balloon: **Cross-version `save_every` / `save_quantizers_only` mismatch (minor)** — Addressed. Manifest now persists both keys; `from_folder` raises on any mismatch. Covered by `test_layerwise_checkpoint_mismatch_save_every_raises`.
- :speech_balloon: **Symmetric `DeprecationWarning` for flat `layerwise_checkpoint_dir` (minor)** — Addressed; warning emitted in `_migrate_layerwise_checkpoint_dir`.
- :speech_balloon: **`resolve_checkpoint_dir` flat+nested redundancy (minor)** — Addressed; resolver writes the resolved path back to both shapes when both are present so the migration validator sees consistent values.
- :speech_balloon: **`__all__` declaration in `config.py` (minor)** — Author marked out of scope; acceptable.

Design-review pass: this PR refactors an existing in-repo subsystem (layerwise calibration + checkpoint format) rather than introducing a new one, so the "is there an existing tool that already does this?" question doesn't apply — the existing tool is what's being refactored. The nested-config + whitelist + per-algorithm default approach is reasonable.

Looping in a human for final sign-off because (a) complexity gate fired (16 files, +853/-135, 8 directories), and (b) author's reported end-to-end verification (Qwen3-8B, 4 PTQ runs, 905 tensors bit-identical across legacy-vs-nested and nested-vs-baseline pairs) is a strong correctness signal but lives outside CI — a human reviewer with the runtime context should confirm the bit-equivalence claim is sufficient.

realAsma:
@Fridah-nv  can you take a look at https://github.com/NVIDIA/Model-Optimizer/pull/1592?

I have a improvement for layerwise (Avoid write back with FSDP2 / CPU/Disk offload unless needed)

modelopt/torch/quantization/model_calib.py
```diff
@@ -1804,19 +1818,35 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+                    layer, [{"quantizer_name": "*", "enable": False}]
+                ):
+                    next_inputs = input_getter.cache_outputs_for_next_layer_calib(
+                        layer, forward_loop
+                    )
+                # cache_outputs left this layer in "run" mode with an empty
+                # deque; reset so calib_func's replay hits the real forward.
+                layer._layerwise_calib.mode = "original"
+
             with persistent_materialization(layer):
```
realAsma, line 1864: Can we still move the next_inputs = input_getter.cache_outputs_for_next_layer_calib to under this context?

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -489,26 +498,31 @@ def _layer_dir(checkpoint_dir: str, idx: int) -> str:
...
     d = _layer_dir(checkpoint_dir, idx)
     if os.path.isdir(d):
         shutil.rmtree(d)
     os.makedirs(d)
-    torch.save(weights, os.path.join(d, "weights.pt"))
+    if weights is not None:
+        torch.save(weights, os.path.join(d, "weights.pt"))
+    elif quantizer_buffers is not None:
+        torch.save(quantizer_buffers, os.path.join(d, "quantizer_buffers.pt"))
     torch.save(qstate, os.path.join(d, "quantizer_state.pt"))
```
cjluo-nv, line 519: > _Bot comment._

Minor follow-up: `_save_layer_files` still accepts a `quantizer_buffers: dict | None` parameter and an `elif quantizer_buffers is not None: torch.save(... 'quantizer_buffers.pt')` branch, but `save()` now always passes `None` for it (the `save_quantizers_only` feature was split out to PR #1640 per the PR description), and `full_restore()` only loads `weights.pt`. The dead parameter + branch + docstring reference can be removed in a follow-up so the checkpoint format stays single-sourced. Not a merge blocker.

cjluo-nv (commented):

Re-review after extensive iteration. The four previously-flagged criticals all appear addressed with regression tests:

- :speech_balloon: **`save_every > 1` `_SkipLayer` corruption** — Fixed. `_CheckpointState.save()` now reads from `layers[layer_idx]` per-call, before the next iteration's capture forward swaps prior layers to `_SkipLayer`. Per-layer files written eagerly; only `next_inputs.pt` + manifest deferred to boundaries.
- :speech_balloon: **Mid-window crash advancing manifest** — Fixed. `_save_layer_files` no longer writes the manifest; `save()` writes it once at boundary/final. Covered by `test_layerwise_save_every_mid_window_crash_recovers_at_prev_boundary` (monkeypatches `torch.save` to crash on layer 2 mid-window, asserts manifest stays at layer 1).
- :speech_balloon: **`_coerce_layerwise_input` `.model_dump()` expanding defaults** — Fixed. Now uses `model_dump(exclude_unset=True)`. Tests cover `LayerwiseConfig(enable=True)` preserving `model_fields_set` so GPTQ's `_gptq_qdq_default` injects correctly.
- :speech_balloon: **realAsma's `qdq_from_prev=False` weight-mutation leak** — Fixed. Next-layer input capture moved to *before* `calib_func` when `qdq_from_prev=False`. Covered by `test_layerwise_no_qdq_captures_inputs_before_calib_func_mutates_weights` (destructive vs identity calib_func produce identical captured inputs).
- :speech_balloon: Cross-version `save_every` mismatch, symmetric `DeprecationWarning` for flat key, `resolve_checkpoint_dir` redundancy guard — all addressed.
- `__all__` declaration in `config.py` — author marked out of scope; acceptable.

Looping in a human for final sign-off because:
1. Complexity gate fired (16 files, +746/-130, 8 directories).
2. The author's strongest correctness signal — bit-identical 905-tensor exports across `(legacy-flat, nested)` and `(layerwise-MSE, baseline-MSE)` Qwen3-8B runs — lives outside CI; a human reviewer with the runtime context should confirm that's sufficient before merge.
3. Prior review iteration discovered 3 critical bugs that the unit tests didn't catch on the first pass; given the new opt-in surface (`save_every`, `qdq_from_prev`), one more set of human eyes is warranted.

Minor non-blocking observation: `_save_layer_files` still carries a dead `quantizer_buffers` parameter / branch (the `save_quantizers_only` feature was split out to PR #1640 per the PR body, but the helper signature kept the slot). Worth a follow-up cleanup but not a merge blocker.

tests/unit/torch/quantization/test_layerwise_calibrate.py
```diff
@@ -713,9 +714,307 @@ def test_mtq_quantize_layerwise_raises_for_unsupported_algorithm():
...
+    torch.manual_seed(0)
+    model_seq = _SimpleTransformerModel(n_layers=3, dim=16)
+    model_lw = copy.deepcopy(model_seq)
+    calib_data = [torch.randint(0, 32, (2, 8)) for _ in range(2)]
+
+    def fwd(m):
+        for batch in calib_data:
+            m(batch)
+
+    mtq.quantize(model_seq, _int8_layerwise_config({"method": "max"}), forward_loop=fwd)
```
realAsma: Is this correct?
why are we passing layserwise config here?

tests/unit/torch/quantization/test_layerwise_calibrate.py
```diff
@@ -713,9 +714,307 @@ def test_mtq_quantize_layerwise_raises_for_unsupported_algorithm():
...
+        torch.testing.assert_close(
+            actual[name],
+            expected[name],
+            rtol=1e-5,
+            atol=1e-6,
+            msg=f"{label}: amax mismatch at {name}",
+        )
+
+
+def test_layerwise_no_qdq_matches_sequential_amax(monkeypatch):
```
realAsma, line 750: @Fridah-nv can you double check this test?

tests/unit/torch/quantization/test_layerwise_calibrate.py
realAsma: > _:robot: Bot comment._

This is correct, but the helper name is confusing. `_int8_layerwise_config` just starts from the shipped INT8 config so the test uses the same real `quant_cfg`, then replaces `cfg["algorithm"]` with the dict passed in.

In this first call we pass only `{"method": "max"}`, so layerwise is not enabled and it exercises the normal sequential max-calibration baseline. The second call adds `layerwise.enable=True` plus `get_qdq_activations_from_prev_layer=False`, so the test compares that layerwise path against the sequential baseline under the same quantizer config.

We should rename the helper or update its docstring to make that clearer.

realAsma (approved):
LGTM! @sugunav14  could you take a look?

please address https://github.com/NVIDIA/Model-Optimizer/pull/1571/changes#r3365862663

tests/unit/torch/quantization/test_layerwise_calibrate.py
Fridah-nv: The `_int8_layerwise_config` name is mis-leading, it's just "an INT8 config with this algorithm block." updating the name and description.

modelopt/torch/quantization/utils/layerwise_calib.py
Fridah-nv: Removed, thanks for flagging

PR #1578, opened 2026-06-01
Refactor local_hessian onto shared MSE flow + fused-MoE expert support

Fridah-nv:
### What does this PR do?

Type of change: Bug fix + new feature (fused-MoE coverage) 

**Dusted off and refactored `local_hessian_calibrate`** to align with the new MSE calibration flow, fix latent
  drift, extend coverage to fused-MoE experts, and decouple module-specific handling behind a clean extension point.

  Core changes (`modelopt/torch/quantization/model_calib.py`):
  - Extracted a shared `_mse_calibrate_weights` helper now used by both `mse_calibrate` and `local_hessian_calibrate` 
  - Replaced the monolithic `LocalHessianHelper` + bespoke per-weight loop with a small `_LocalHessianAccumulator` (lazy fp32 buffer, freed after building the error func), removing ~200 lines of duplicated scale-search logic and all manual `cuda.synchronize`/`empty_cache` bookkeeping. The `XᵀX` GEMM accumulates in fp32 to avoid bf16/fp16  precision loss.
  - Removed dead NVFP4-static promotion (now handled inside `max_calibrate`).
  - **Fused-MoE expert support:** per-expert Hessians captured from each expert's routed activations, keyed by
  `id(weight_quantizer)` so dense and per-expert paths share one calibration loop. Never-routed experts / non-eager kernels / `cin` not divisible by `block_size` / registered backends fall back to plain MSE (with an eager, module-named warning).

  Decoupling (zero module-type-specific code in `model_calib.py`):
  - Added `QuantModule.register_calibration_input_hooks(callback)` — the activation-side counterpart to `iter_weights_for_calibration`. Base default is a no-op; `QuantLinearConvBase` pairs the weight quantizer with the
  forward input (linear only), and `_QuantFusedExperts` (in `plugins/huggingface.py`) owns the per-expert pairing via `_current_expert_idx`. Any future module type gains local-Hessian support by implementing this one method.

### Usage

```python
```

### Testing
- new `tests/unit/torch/quantization/test_local_hessian.py` (accumulator math/shape/dtype, dense end-to-end, backend-skip, block-size guard) and a per-expert MoE test in `test_fused_experts.py`. 
 - Behavior-preserving check: refactored branch produces **bit-identical** dense weight scales to `main` (216/216 tensors) on Qwen3-8B with the fp32 accumulation neutralized to isolate the structural change.

## Summary by CodeRabbit

* **New Features**
  * Introduced Hessian-weighted calibration method for enhanced weight quantization refinement in mixed-precision models.
  * Enhanced MSE calibration to support custom per-quantizer error functions for specialized calibration workflows.

* **Tests**
  * Added comprehensive unit tests for local Hessian calibration validation across dense models and fused-expert architectures.
  * Included tests for custom error function integration and fallback behavior.

### DIFF
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -804,6 +804,15 @@ def local_hessian_calibrate(
         error_func_for=lambda q: error_funcs.get(id(q)),
     )
 
+    # Free the per-block Hessians (pinned by error_func closures) and the sweep's cached
+    # allocations so export starts from a defragmented allocator.
+    error_funcs.clear()
+    for module in name_to_module.values():
+        if isinstance(module, TensorQuantizer) and isinstance(module._calibrator, MseCalibrator):
+            module._calibrator._error_func = None
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
+
     if debug:
         model._local_hessian_accumulators = accumulators
 
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -465,10 +465,10 @@ def _make_weight_mse_calibrator(
         )
         if backend is not None and backend_factory is not None:
             if error_func is not None:
-                # Registered backends can't take a custom error_func; skip Hessian refinement.
+                # Registered backend factories don't accept a custom error_func.
                 warnings.warn(
-                    f"local_hessian: backend '{backend}' does not support a custom error "
-                    "function; skipping Hessian-weighted calibration for this quantizer."
+                    f"backend '{backend}' does not support a custom error function; skipping "
+                    "error-function-weighted MSE calibration for this quantizer."
                 )
                 return None
             return backend_factory(initial_amax, axis, quant_func)
@@ -670,6 +670,80 @@ def _warn_local_hessian_fallback(name, weight, weight_quantizer, block_size, war
     _warn_if_block_size_mismatch(weight_quantizer, block_size, name)
 
 
+def _is_quant_fused_experts(module: nn.Module) -> bool:
+    """Whether ``module`` is a converted HF fused-MoE-experts wrapper with per-expert quantizers."""
+    return hasattr(module, "_current_expert_idx") and hasattr(
+        module, "gate_up_proj_weight_quantizers"
+    )
+
+
+def _register_local_hessian_input_hooks(model, name_to_module, capture, block_size, warned):
+    """Register forward hooks feeding each weight's input activations to ``capture``.
+
+    Local-Hessian-specific (kept here rather than as a general ``QuantModule`` API): dense
+    quantized linears hook the layer input; HF fused-MoE experts hook the shared input quantizers,
+    keyed by the active expert (``_current_expert_idx``). Weights without a hook (conv,
+    SequentialQuantizer, non-eager experts) fall back to plain MSE. Returns removable handles.
+    """
+    handles: list = []
+
+    def _make_expert_hook(expert_module, weight_name, quantizers, enabled):
+        def _expert_hook(_input_quantizer, args):
+            if not args:
+                return
+            idx = expert_module._current_expert_idx
+            if idx in enabled:
+                # Read the weight fresh (valid under accelerate/FSDP re-materialization).
+                capture(quantizers[idx], getattr(expert_module, weight_name)[idx], args[0])
+
+        return _expert_hook
+
+    for name, module in name_to_module.items():
+        if is_quantized_linear(module) and isinstance(module.weight_quantizer, TensorQuantizer):
+            with enable_weight_access_and_writeback(module, model, name_to_module):
+                # ``weight`` may be absent (e.g. TE GroupedLinear exposes weight0..N, not weight);
+                # such modules have no single 2-D weight to pair and fall back to plain MSE.
+                weight = getattr(module, "weight", None)
+                if weight is None or weight.dim() != 2 or not module.weight_quantizer.is_enabled:
+                    continue
+                _warn_local_hessian_fallback(
+                    name, weight, module.weight_quantizer, block_size, warned
+                )
+
+            def _dense_hook(linear, args):
+                if args:
+                    capture(linear.weight_quantizer, linear.weight, args[0])
+
+            handles.append(module.register_forward_pre_hook(_dense_hook))
+        elif _is_quant_fused_experts(module):
+            with enable_weight_access_and_writeback(module, model, name_to_module):
+                for weight_name, quantizers_name, input_q_name in (
+                    (
+                        "gate_up_proj",
+                        "gate_up_proj_weight_quantizers",
+                        "gate_up_proj_input_quantizer",
+                    ),
+                    ("down_proj", "down_proj_weight_quantizers", "down_proj_input_quantizer"),
+                ):
+                    weight = getattr(module, weight_name, None)
+                    quantizers = getattr(module, quantizers_name, None)
+                    input_quantizer = getattr(module, input_q_name, None)
+                    if weight is None or quantizers is None or input_quantizer is None:
+                        continue
+                    _warn_local_hessian_fallback(
+                        f"{name}.{weight_name}", weight[0], quantizers[0], block_size, warned
+                    )
+                    # Snapshot which experts are enabled now, before the caching forward silences
+                    # all weight quantizers — so we don't capture (and discard) disabled experts.
+                    enabled = {i for i, q in enumerate(quantizers) if q.is_enabled}
+                    handles.append(
+                        input_quantizer.register_forward_pre_hook(
+                            _make_expert_hook(module, weight_name, quantizers, enabled)
+                        )
+                    )
+    return handles
+
+
 @torch.no_grad()
 def local_hessian_calibrate(
     model: nn.Module,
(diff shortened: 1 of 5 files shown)


cjluo-nv (commented):

Solid refactor that consolidates `local_hessian_calibrate` onto the shared MSE flow and extends Hessian capture to HF fused-MoE experts via a new `register_calibration_input_hooks` extension point on `QuantModule`. Code is well-structured, tests are meaningful (accumulator math, dense end-to-end, conv fallback, block-size mismatch, registered-backend skip, per-expert MoE capture + refinement), and bit-identical preservation on Qwen3-8B is claimed.

Design check passed: the new extension point is a small activation-side counterpart to the existing `iter_weights_for_calibration`, not a parallel subsystem; this is the right shape rather than introducing a new registry/loader. The PR explicitly documents fallback paths (never-routed experts, non-eager kernels, `cin % block_size != 0`, registered backends) and makes them warn rather than silently miscalibrate.

Nudging (not approving) for two reasons that warrant a human signoff:

1. **Distributed correctness gap is acknowledged but not fixed.** The new code emits a `warn_rank_0` and a TODO that "the per-block Hessian is not synced across TP/DP ranks ... refined amaxes can diverge under tensor/data parallelism. Treat local_hessian as single-rank for now." The previous (`LocalHessianHelper`) implementation had the same gap (also TODO'd), so this is no regression — but the change still ships a feature whose multi-rank behavior is known-divergent. A maintainer should confirm this is acceptable for the intended use cases (and whether `local_hessian_calibrate` should hard-error rather than warn under TP/DP > 1, as `max_calibrate` does for NVFP4-static + TP).

2. **Bit-equivalence claim is human-verified outside CI.** The PR body says the refactor produces "bit-identical dense weight scales to main (216/216 tensors) on Qwen3-8B with the fp32 accumulation neutralized to isolate the structural change." This is exactly the kind of claim that needs human confirmation since it can't be re-run in CI; please have the original reviewer/runner sign off that the comparison was done with `fp32 accumulation neutralized` (i.e. bf16 GEMM as in the old code) and that the deliberate fp32-accumulation switch in the new path is the only intended numerical change.

Minor (non-blocking) observations:
- `from modelopt.torch.utils.distributed import ...` is split across three statements; the two new ones (`is_initialized as dist_is_initialized`, `size as dist_size`) could be folded into the existing line.
- `_warn_local_hessian_fallback` is only called when `captures` is non-empty, which means a `cin % block_size != 0` warning is suppressed for non-pairable modules (where it wouldn't matter anyway). Consider a brief comment so a future reader doesn't think the warning is missing.
- `SequentialQuantizer` weight quantizers are silently skipped (matches `mse_calibrate`), but this is invisible to users running `local_hessian` on a SequentialQuantizer config — they'll see no fallback warning since `register_calibration_input_hooks` does pair them. Worth a one-line comment in `local_hessian_calibrate` or a sibling warning path.

modelopt/torch/quantization/model_calib.py
```diff
@@ -622,221 +743,82 @@ def local_hessian_calibrate(
...
+    for weight_quantizer in silenced_weight_quantizers:
+        weight_quantizer.disable_quant()
+    print_rank_0("local_hessian: Caching activations and computing local Hessian...")
+    try:
+        forward_loop(model)
+    finally:
+        for weight_quantizer in silenced_weight_quantizers:
+            weight_quantizer.enable_quant()
+        for handle in handles:
+            handle.remove()
```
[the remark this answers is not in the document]
Fridah-nv: input_quantizers's states are expected to be preserved in this algo, intended behavior

modelopt/torch/quantization/model_calib.py
```diff
@@ -581,6 +622,84 @@ def mse_calibrate(
...
+
+    def build_error_func(self) -> Callable[[torch.Tensor, torch.Tensor], torch.Tensor] | None:
+        """Hessian-weighted error function (``None`` if no samples); frees the raw buffer."""
+        if self.hessian_per_block is None or self.num_samples == 0:
+            return None
+        cout = self.cout
+        bs = self.block_size
+        hessian = self.hessian_per_block / self.num_samples
+        self.hessian_per_block = None
+
```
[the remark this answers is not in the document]
Fridah-nv: Fixed — build_error_func(keep_buffer=debug) retains the raw per-block Hessian on the accumulator under debug=True (freed as before in normal runs), so model._local_hessian_accumulators exposes the ΣXᵀX buffers + num_samples for inspection. Docstring now matches.

modelopt/torch/quantization/nn/modules/quant_module.py
```diff
@@ -247,6 +259,18 @@ def _setup(self):
...
+        """Pair the weight quantizer with the forward input (linear only; conv falls back)."""
+        weight = getattr(self, "weight", None)
+        if weight is None or weight.dim() != 2 or not self.weight_quantizer.is_enabled:
+            return []
+
+        def _pre_hook(module, args):
+            if args:
+                callback(module.weight_quantizer, module.weight, args[0])
+
+        return [self.register_forward_pre_hook(_pre_hook)]
```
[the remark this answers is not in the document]
Fridah-nv: Done — register_calibration_input_hooks now requires isinstance(self.weight_quantizer, TensorQuantizer), so seq-quant/conv weights register no hook.

tests/unit/torch/quantization/plugins/test_fused_experts.py
```diff
@@ -656,6 +656,66 @@ def forward_loop(m):
 
         self._cleanup_registry(expert_type)
 
+    def test_local_hessian_per_expert_capture_and_refinement(self):
+        """The plugin's extension point pairs each per-expert weight quantizer with its routed
+        input, and local_hessian uses that to refine every expert's weight amax."""
+        import modelopt.torch.quantization as mtq
+        from modelopt.torch.quantization.model_calib import local_hessian_calibrate
+
```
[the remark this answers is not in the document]
Fridah-nv: Done

modelopt/torch/quantization/model_calib.py
```diff
@@ -622,221 +743,82 @@ def local_hessian_calibrate(
...
+                # Only TensorQuantizer weights are calibrated (matches mse_calibrate), so only
+                # those are silenced.
+                if (
+                    isinstance(weight_quantizer, TensorQuantizer)
+                    and weight_quantizer.is_enabled
+                    and weight_quantizer._if_quant
+                ):
+                    silenced_weight_quantizers.append(weight_quantizer)
+                if captures:
+                    _warn_local_hessian_fallback(name, weight, weight_quantizer, block_size, warned)
```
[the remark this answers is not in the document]
Fridah-nv: Good catch — restored. SequentialQuantizer weight quantizers are now silenced during the capture forward (we iterate their TensorQuantizer leaves and toggle  _if_quant), so downstream Hessians are built from full-precision weights as on main. They remain unrefined (only TensorQuantizer weights are calibrated, matching mse_calibrate) and emit a one-time "unsupported, left at max-cal" warning.

modelopt/torch/quantization/model_calib.py
```diff
@@ -622,221 +743,82 @@ def local_hessian_calibrate(
...
+        fp8_scale_sweep=fp8_scale_sweep,
+        error_func_for=lambda q: error_funcs.get(id(q)),
+    )
 
-    # Cleanup and free memory
-    LocalHessianHelper.cache_mode = False
-    for name, module in all_patched_modules:
-        module.hessian_helper.cleanup()
+    if debug:
+        model._local_hessian_accumulators = accumulators
```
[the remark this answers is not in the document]
Fridah-nv: Fix already applied

tests/unit/torch/quantization/test_local_hessian.py
```diff
@@ -0,0 +1,245 @@
...
+        for h in handles:
+            h.remove()
+        assert captured and captured[0] == ((32, 16), 16)  # cin from activation matches weight
+
+        conv = SimpleConv()
+        mtq.quantize(conv, INT8_WEIGHT_CFG, forward_loop=lambda m: m(SimpleConv.get_input()))
+        assert conv.net[0].register_calibration_input_hooks(lambda *a: None) == []  # 4-D weight
+
+    def test_sequential_quantizer_weight_not_hooked(self):
+        from modelopt.torch.quantization.nn import SequentialQuantizer
```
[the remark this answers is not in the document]
Fridah-nv: Done

cjluo-nv:
Can we test doing this on a model and runs some simple evals and compare that with the base MSE without hessian?

Fridah-nv:
> Can we test doing this on a model and runs some simple evals and compare that with the base MSE without hessian?

I plan to do a larger scale experiments after this PR lands, I'd like to merge it first to make sure basic correctness and code quality.

modelopt/torch/quantization/model_calib.py
```diff
@@ -494,26 +500,34 @@ def _make_weight_mse_calibrator(
             _FP8_SWEEP_CALIBRATOR_REGISTRY.get(backend) if backend is not None else None
         )
         if backend is not None and backend_factory is not None:
+            if error_func is not None:
+                # Registered backends can't take a custom error_func; skip Hessian refinement.
+                warnings.warn(
+                    f"local_hessian: backend '{backend}' does not support a custom error "
+                    "function; skipping Hessian-weighted calibration for this quantizer."
+                )
```
realAsma: Can we make this more generic? 

We might have some other algorithm other than local hessian to use mse_calibration + error function.

modelopt/torch/quantization/model_calib.py
```diff
@@ -622,221 +750,107 @@ def local_hessian_calibrate(
...
+            captures = module.register_calibration_input_hooks(capture)
+            handles.extend(captures)
+            for weight, weight_quantizer in module.iter_weights_for_calibration():
+                # Silence weight fake-quant (incl. SequentialQuantizer leaves) so the capture
+                # forward uses full-precision weights and downstream Hessians aren't corrupted.
+                leaves = (
+                    list(weight_quantizer)
+                    if isinstance(weight_quantizer, SequentialQuantizer)
+                    else [weight_quantizer]
+                )
```
realAsma: we not need this special handling for Sequential Quantizer in my understanding.  we can call quantizer.disable() on SequentialQuantizer instance.

Can we simplify the code here?

modelopt/torch/quantization/nn/modules/quant_module.py
```diff
@@ -247,6 +259,27 @@ def _setup(self):
...
+            or not isinstance(self.weight_quantizer, TensorQuantizer)
+            or not self.weight_quantizer.is_enabled
+        ):
+            return []
+
+        def _pre_hook(module, args):
+            if args:
+                callback(module.weight_quantizer, module.weight, args[0])
+
+        return [self.register_forward_pre_hook(_pre_hook)]
```
realAsma: This seems to be creating a general infrastructure  to support local_hessian calibration. 

Can we for now move this to model_calib.py and make it specific to local_hessian?

If we need this infra for more future algorithms, we can always move it to modelopt/torch/quantization/nn/modules/quant_module.py.

modelopt/torch/quantization/model_calib.py
```diff
@@ -622,221 +750,107 @@ def local_hessian_calibrate(
...
+    for weight_quantizer in silenced_weight_quantizers:
+        weight_quantizer.disable_quant()
+    print_rank_0("local_hessian: Caching activations and computing local Hessian...")
+    try:
+        forward_loop(model)
+    finally:
+        for weight_quantizer in silenced_weight_quantizers:
+            weight_quantizer.enable_quant()
+        for handle in handles:
+            handle.remove()
```
realAsma: Can we have a context manager like this - https://github.com/NVIDIA/Model-Optimizer/blob/4eaf90a6ca69732f1d85b07496f937a44a1def33/modelopt/torch/quantization/model_calib.py#L1261
```suggestion
    with set_quantizer_by_cfg_context(
        model,
        [{"quantizer_name": "*weight_quantizer", "enable": False}],
    ):
        forward_loop(model)
        
   for handle in handles:
            handle.remove()
```

realAsma (commented):
Overall this PR is great.

  My main concern is that this adds a general calibration hook API to `QuantModule` for a need that is currently very specific to
  `local_hessian`. That feels like more infrastructure than we need right now, and it also broadens the quantization module contract before we
  know other algorithms will actually use it.

  Can we keep this surgical for now by moving the input-capture handling into `model_calib.py` and making it explicitly local-Hessian-specific?
  If we later see the same pattern needed by other calibration algorithms, we can promote it into `quant_module.py` as shared infrastructure
  then.

modelopt/torch/quantization/model_calib.py
Fridah-nv: Done

modelopt/torch/quantization/model_calib.py
Fridah-nv: Done, thanks for your suggestion

modelopt/torch/quantization/nn/modules/quant_module.py
Fridah-nv: Done, moved the capture to  a local-Hessian-specific helper `_register_local_hessian_input_hooks` in `model_calib.py`, it handles both dense and fused-MoE experts. In the future if we want to generalize local-hessian further or other algos needs activation capture, we can bring it back

modelopt/torch/quantization/model_calib.py
```diff
@@ -545,6 +586,164 @@ def mse_calibrate(
...
+                    continue
+                _warn_local_hessian_fallback(
+                    name, weight, module.weight_quantizer, block_size, warned
+                )
+
+            def _dense_hook(linear, args):
+                if args:
+                    capture(linear.weight_quantizer, linear.weight, args[0])
+
+            handles.append(module.register_forward_pre_hook(_dense_hook))
```
realAsma, line 717: nit:
On expert path we are adding hooks to input_quantizer. Here we are adding hook to linear layer. Should we add hook to input_quantizer here as well?

modelopt/torch/quantization/model_calib.py
```diff
@@ -586,221 +788,73 @@ def local_hessian_calibrate(
...
+    try:
+        with set_quantizer_by_cfg_context(
+            model, [{"quantizer_name": "*weight_quantizer", "enable": False}]
+        ):
+            forward_loop(model)
+    finally:
+        for handle in handles:
+            handle.remove()
+
+    # TODO(fridah-nv): the per-block Hessian is not synced across TP/DP ranks (max_calibrate's
```
realAsma, line 824: @Fridah-nv  
We can de-prioritize TP sync and support DP sync first. DP sync should be trivial. This can be done in a seperate future PR.

realAsma (approved):
Looks great!!

modelopt/torch/quantization/model_calib.py
Fridah-nv: Good question. Hooking to `input_quantizer` on the dense path causes extra complexity, because inside the hook when PyTorch passes the input_quantizer back as the first arg, it cannot access the weight(The input_quantizer has no reference back to its parent linear), have to use a closure/factory to snapshot the weight. 
For FusedMoE we have to hook on `input_quantizer` as a hook for the batch of experts. The behavior is identical for the both path.
In the future when we start to force a module-agnostic algorithm, this kind of difference can be hid within the modules descriptors.

PR #1583, opened 2026-06-01
Add parallelism sizing reference to evaluation skill

cjluo-nv:
### What does this PR do?

Type of change: documentation

Adds `.claude/skills/evaluation/references/parallelism.md`, a reference for sizing both the **GPU topology** and the **request concurrency** of NEL evaluation runs, and wires the main `SKILL.md` to it. Pure docs for the agent-facing evaluation skill — no code or runtime behavior changes.

The new reference covers:

- **GPU topology (TP / DP / PP):** decision procedure (smallest-TP-then-max-DP), the TP-up triggers, and the TP/DP-split tradeoff for a fixed world size (e.g. all `TP×DP=8` factorizations on an 8-GPU node).
- **Expert parallelism (EP):** `--enable-expert-parallel` is a boolean and `EP = TP × DP` (no direct EP-size flag); the DP-attention + EP-MoE dataflow (per-layer dispatch/combine all-to-all); when to enable vs not.
- **Concurrency (`parallelism` / `--max-num-seqs`):** the request-count-vs-serving-capacity ceiling, KV-driven `--max-num-seqs`, and empirical tuning from vLLM startup logs + preemption.
- **Gotcha + worked examples:** bit-width read from `config.json` (not the model name) sets the topology; includes the FP8-vs-4-bit Kimi example.

`SKILL.md` gains pointers to the reference from the deployment-command, expert-parallel, evaluation-params, Step 4, and canary sections.

### Usage

N/A — documentation only (agent skill guidance).

### Testing

`pre-commit run` passes on both files (markdownlint included).

- Did you update Changelog?: N/A (skill docs, not a library feature)

### Additional Information

Commits are signed (`git commit -s -S`).

## Summary by CodeRabbit

* **Documentation**
  * Expanded deployment guidance with clearer GPU topology, MoE/EP semantics, and bit‑width caveats.
  * Refined concurrency guidance: explicit rules for sizing top-level parallelism and computing max concurrent sequences, plus canary-run tuning to watch preemption and KV utilization.
  * Added task-level guidance for long‑context/judge‑bound jobs with recomputation advice and worked examples.

### DIFF
--- a/.claude/skills/evaluation/references/parallelism.md
+++ b/.claude/skills/evaluation/references/parallelism.md
@@ -1,0 +1,273 @@
+# Parallelism: GPU topology (TP / DP / PP / EP) and concurrency (`parallelism` / `--max-num-seqs`)
+
+Two layers of decisions, made in order:
+
+1. **GPU topology** — how the model is laid out across the GPUs you have
+   (`--tensor-parallel-size`, `--data-parallel-size`, `--pipeline-parallel-size`,
+   `--enable-expert-parallel`). Decide this first; it determines how many
+   independent replicas exist.
+2. **Concurrency** — how many requests are in flight (`parallelism`) and how many
+   sequences each replica decodes at once (`--max-num-seqs`). Sized on top of the
+   topology.
+
+Both layers affect **throughput only** — never scores.
+
+---
+
+## Layer 1 — GPU topology: TP / DP / PP
+
+| Dim | What it shards | What it buys | Cost |
+| --- | --- | --- | --- |
+| **TP** (tensor parallel) | Each layer's weights + KV across GPUs **within one replica** | Lets a model that doesn't fit on one GPU run; splits KV so longer context fits | All-reduce **every layer** → latency + needs fast interconnect (NVLink). Keep **within one node**. |
+| **DP** (data parallel) | Nothing — **replicates** the (TP-sharded) model | Throughput: N independent replicas serve N× the concurrent requests | N× the weight memory (one full copy per replica) |
+| **PP** (pipeline parallel) | Contiguous **layer ranges** across GPUs | Fits a model too big for intra-node TP; cheaper cross-node than TP | Pipeline bubbles → lower utilization. Mostly for very large / multi-node (see `multi-node.md`). |
+
+**Decision procedure (single node, G GPUs):**
+
+1. **TP = the smallest value that makes the model fit with KV headroom.**
+   Estimate weight memory ≈ `params × bytes_per_param`:
+   - NVFP4 ≈ 0.5–0.6 B/param (incl. scales) · FP8 ≈ 1 B · BF16/FP16 ≈ 2 B.
+   - Need `weights/TP + KV cache + activations + CUDA-graph/overhead` to fit in
+     `GPU_mem × gpu_memory_utilization`. If it fits on one GPU → **TP = 1**.
+   - Constraints: TP must **divide `num_attention_heads`** (and ideally
+     `num_key_value_heads` for GQA, else KV heads get replicated and waste memory);
+     use a **power of 2**; never span nodes with TP.
+   - Smaller TP = less communication = higher efficiency. Don't over-shard "to be
+     safe" — it slows decode.
+2. **DP = floor(G / (TP × PP))** — use the leftover GPUs as replicas. For
+   throughput-bound evals (the common case), **maximize DP**: a model that fits on
+   one GPU should run `TP=1, DP=G`, not `TP=G, DP=1`.
+3. **PP only if** the model can't fit even at the largest sensible intra-node TP,
+   or you're going multi-node — then see `multi-node.md`.
+
+DP is the lever that grows **serving capacity** (Layer 2): more replicas → more
+concurrent sequences.
+
+> **Gotcha — bit-width sets the topology, not the model name.** The weight estimate
+> hinges on `bytes_per_param`, so **read the actual precision from `config.json`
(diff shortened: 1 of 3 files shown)


meenchen (approved):

Pure documentation addition for the agent-facing evaluation skill. Adds a new `references/parallelism.md` (consistent in style and cross-referencing with sibling refs like `multi-node.md` and `quantization-benchmarks.md`) and wires `SKILL.md` to it from the relevant sections (deployment defaults, EP flag, evaluation params, Step 4, canary). Content is internally consistent (EP=TP×DP, `max-num-seqs=ceil(parallelism/(DP×num_instances))`, agrees with the existing `multi-node.md` definitions of `num_instances` vs `data_parallel_size`). No code/runtime behavior change, no licensing concerns, tests N/A.

.claude/skills/evaluation/references/parallelism.md
```diff
@@ -0,0 +1,273 @@
+# Parallelism: GPU topology (TP / DP / PP / EP) and concurrency (`parallelism` / `--max-num-seqs`)
```
chadvoegele: Do you think the model already knows about this background info from pre-training? Is it useful to repeat it here?

.claude/skills/evaluation/references/parallelism.md
```diff
@@ -0,0 +1,273 @@
...
+   `quantization-benchmarks.md`). You can't have more in flight than exist.
+2. **Sustainable serving capacity** = `max-num-seqs × DP × num_instances`, bounded
+   by KV-cache memory (below).
+
+| Situation | Set `parallelism` to | Why |
+| --- | --- | --- |
+| `total_requests ≤ serving_capacity` (small run) | `total_requests` (round up a little for uneven DP routing) | All requests dispatch at once → one wave → finishes in ~one generation-time. Higher is wasted. |
+| `total_requests ≫ serving_capacity` (large run) | `serving_capacity` (largest the GPUs sustain) | Throughput-bound: keep every decode slot full until the queue drains. Request count no longer matters. |
+
+So "set it higher" is right **only up to the request count**; past that you just
```
chadvoegele: I think we mention it somewhere else, but maximizing the throughput isn't necessarily the correct goal because we can overwhelm dependent services in the evaluation, like user simulator, code execution envs, etc.

.claude/skills/evaluation/references/parallelism.md
```diff
@@ -0,0 +1,273 @@
...
+concurrent sequences.
+
+> **Gotcha — bit-width sets the topology, not the model name.** The weight estimate
+> hinges on `bytes_per_param`, so **read the actual precision from `config.json`
+> (`quantization_config` / `quant_algo` / dtype) before sizing** — do not infer it
+> from the org/handle. Two checkpoints of the *same architecture at the same
+> bit-width* have the same footprint → the same TP/DP/EP, regardless of vendor or
+> quant scheme (INT4 vs NVFP4 differ only in kernel/quant-method flags, which vLLM
+> auto-detects — and a negligible effective-bit difference). The split only changes
+> when the bit-width changes the *size* (see the Kimi example below).
```
meenchen: We sometimes quantize kv cache using the arguments in vllm serve. Could you also add it to the note?

.claude/skills/evaluation/references/parallelism.md
cjluo-nv: I think it's better to be explicit here

.claude/skills/evaluation/references/parallelism.md
cjluo-nv: for tau2, we will apply a specific cap in the tau2 yaml description.

cjluo-nv: actually I think you are right. I removed this from the latest commit.

.claude/skills/evaluation/references/parallelism.md
cjluo-nv: Good point — added. The "Factors that relax the KV limit" note now covers serve-time KV-cache quantization via vLLM's `--kv-cache-dtype fp8` (`fp8_e4m3` / `fp8_e5m2`) in `deployment.command`, in addition to the checkpoint's `kv_cache_scheme` — noting it ~halves KV vs bf16/fp16 (~2× sustainable concurrency/context) and to verify model/recipe support. Done in 0839159d.

.claude/skills/evaluation/references/parallelism.md
cjluo-nv: Good point — trimmed. I compressed the doc ~52% (345→165 lines, commit e4368c12): cut the generic TP/DP/PP/EP background the model already knows and merged the overlapping sections, keeping only the non-obvious decision rules, formulas, gotchas (e.g. bit-width sets the topology), and the worked examples. Happy to cut further if any of what's left still reads as restated background.

.claude/skills/evaluation/references/parallelism.md
cjluo-nv: Agreed, and addressed directly. The doc no longer treats throughput-max as the goal:

- "Balanced sizing — bigger is NOT always faster" (commit 386b4980) spells out that over-parallelizing *regresses* — preemption thrash, prefill/decode contention, and the latency→timeout→retry cascade — and says to err low for long context.
- "Suites — set parallelism per task" (commit 8ac3342f) calls out exactly your point: judge / user-simulator / sandbox tasks bottleneck *before* the served model, so cap them to the dependent endpoint (429s), not the GPU. AA-LCR and tau2 ship `parallelism: ???` to force a conscious, lower value.

meenchen (approved):

PR #1586, opened 2026-06-01
Fix MLflow auto-export in evaluation skill template

cjluo-nv:
### What does this PR do?

Type of change: documentation

Fixes two issues in the evaluation skill that caused completed eval runs to silently **not** upload to MLflow (found while running a real GPQA eval — the run finished successfully but never appeared in MLflow):

1. **Missing auto-export trigger.** NEL only auto-exports when `execution.auto_export.destinations` is set; the `export.mlflow` block alone just *configures* the exporter. The skill's shortcut said to "copy the `export.mlflow` block" without calling out the trigger, so a config could end up with the block but no upload.

2. **Interpolated export values break at submit time.** With auto-export enabled, NEL resolves the `export.mlflow` block at submit time in a scope that does **not** include `deployment` / `evaluation`, so `${deployment.served_model_name}` / `${evaluation.*}` cross-references fail hard with `Interpolation key '...' not found`. (`example_eval.yaml` shipped this interpolated pattern *with* auto-export, so it would hit the error.)

Changes:

- **`example_eval.yaml`**: mark the `auto_export.destinations` trigger as REQUIRED; switch the `export.mlflow` block to **literal** values (`CHANGEME-served-model-name` placeholders) with a comment explaining why cross-refs break. `${oc.env:USER}` (env interpolation) is retained since it resolves fine.
- **`SKILL.md`**: rewrite the MLflow shortcut step to require **both** the trigger and literal export values, and warn against `${deployment.*}` / `${evaluation.*}` cross-references.

### Usage

N/A — documentation/skill-template only.

### Testing

`pre-commit run` passes on both files (markdownlint + YAML format). Verified the literal-export approach uploads correctly by exporting a real run to MLflow.

- Did you update Changelog?: N/A (skill docs)

### Additional Information

Separate from #1583 (parallelism reference); both touch the evaluation skill but are independent.

## Summary by CodeRabbit

* **Documentation**
  * Clarified MLflow auto-export: auto-export must be explicitly enabled for uploads and the MLflow export block is ignored otherwise.
  * Example config now uses literal experiment name, description, and tags (no cross-scope interpolation) and shows concrete sampling values.
  * Warned to keep sampling params consistent between tags and runtime params and to set the tracking URI.


meenchen (approved):

Docs-only fix to the evaluation skill template. The two changes are consistent and well-explained: (1) `auto_export.destinations` is annotated as the required trigger, and (2) the `export.mlflow` block switches from `${deployment.*}` / `${evaluation.*}` cross-refs (which fail at NEL submit time, where that scope isn't available) to `CHANGEME-…` literals plus `${oc.env:USER}` (env-var interpolation, which still resolves). SKILL.md Step 2 shortcut is updated to match, including the drift-warning to keep the literal `temperature` / `top_p` / `max_new_tokens` tags in sync with the top-level `params`. Author reports verifying upload with a real run. No licensing, code, or architecture impact.

PR #1590, opened 2026-06-01
Fix: use python3 in debugger server.sh

cjluo-nv:
### What does this PR do?

Type of change: Bug fix

`tools/debugger/server.sh` invokes `python -c "..."` inside `check_modelopt_local` (and `pip install -e .[dev]` in the fallback path). Containers that only ship `python3` (no `python` shim) cause the check to fail with `python: command not found`, exit 127. The server interprets this as "modelopt not editable-installed", runs `pip install`, the second check fails the same way, and the server aborts before it ever listens for commands.

Switches both the inline import check and the install fallback to `python3` / `python3 -m pip`.

### Usage

```bash
bash tools/debugger/server.sh
# now starts cleanly on python3-only images
```

### Testing

Verified inside a container where `which python` returns nothing and `python3` is `/usr/bin/python3` (Python 3.12). With the old script, `server.sh` aborted at `check_modelopt_local`. With this fix, the check passes against an existing editable install and the server proceeds to wait for the client handshake.

- Is this change backward compatible?: ✅ — `python3` is present on every image that previously had `python`.
- Did you write any new necessary tests?: N/A — internal tooling shell script.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — internal debugger tooling, not user-facing.

## Summary by CodeRabbit

* **Chores**
  * Updated debugger tooling to ensure consistent use of Python 3 for package validation and installation processes.

### DIFF
--- a/tools/debugger/server.sh
+++ b/tools/debugger/server.sh
@@ -99,7 +99,7 @@ mkdir -p "$CMD_DIR" "$RESULT_DIR"
 
 # Ensure modelopt is editable-installed from WORKDIR
 check_modelopt_local() {
-    python -c "
+    python3 -c "
 import modelopt, os, sys
 actual = os.path.realpath(modelopt.__path__[0])
 expected = os.path.realpath('$WORKDIR')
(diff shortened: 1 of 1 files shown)


meenchen (approved):

Trivial, correct fix: switches `python` → `python3` and `pip` → `python3 -m pip` in `tools/debugger/server.sh` so the editable-install check works in containers that only ship `python3`. Internal debugger tooling, no tests warranted, fully backward-compatible.

PR #1592, opened 2026-06-01
Refine layerwise non-mutating calibration

realAsma:
Stacked on #1571.

Summary:
- Add calib_mutates_weights gating for non-mutating layerwise calibration.
- Skip layer weight checkpoint/writeback for quantizer-state-only calibration.
- Keep FSDP2/Accelerate writeback conditional and improve layerwise progress reporting.
- Make `_zeros_from_meta` allocate skip placeholders on `torch.device("meta")` so skip-mode preserves structure without real-device tensor allocation.

Testing:
- Pre-commit hooks passed during commit.


modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -486,6 +504,11 @@ def fsdp2_weight_access_and_writeback_context(module: nn.Module, root_model: nn.
     assert not hasattr(module, "_hf_hook"), "We dont support FSDP2 with HF accelerate hooks"
     fsdp_module = _get_enclosing_fsdp_module(module, root_model)
     assert fsdp_module is not None, "Module is not wrapped by FSDP"
+    if not writeback:
+        with _fsdp2_unshard_context(fsdp_module):
+            yield
+        return
+
```
realAsma, line 511: @sugunav14  here is an easy perf improvement for layerwise FSDP2

Fridah-nv: @realAsma Claude claims a correctness issue on this branch, it makes sense to me, please check from your side:

  ### Bug: FSDP2 `writeback=False` path never all-gathers weights (non-mutating layerwise calibration computes on sharded shards)

  **Where:** `_fsdp2_unshard_context` (`core_utils.py:480`) reached via `persistent_materialization(layer, writeback=False)`.

  **Root cause — a collision between two layers of the same context stack.**

  `persistent_materialization` enters its context managers in this order:

  ```python
  with (
      _disable_fsdp_unshard_reshard(layer),                              # ① enters FIRST
      enable_weight_access_and_writeback(layer, layer, writeback=False), # ② enters SECOND
      temporarily_remove_accelerate_hook(layer),
  ):
  ```

  - **①** monkeypatches the **class** method `FSDPParamGroup.unshard` to a no-op (to stop FSDP from re-sharding weights between
  calibration batches). This patch is global and is now active.
  - **②** with `writeback=False` routes into `_fsdp2_unshard_context`, whose gather step is:

    ```python
    if was_sharded:
        fsdp_module.unshard()   # core_utils.py:480
    ```

    But `FSDPModule.unshard()` delegates the actual all-gather to `fsdp_param_group.unshard()` — which is exactly the method **①**
  just patched to a no-op. So **no all-gather happens; the params stay sharded `DTensor`s**, and the calibration/capture forward
  runs on partial shards → wrong `_amax` (or a shape/dtype error).

  The layer's own FSDP pre-forward hook can't save it either: that hook also calls `FSDPParamGroup.unshard`, which is still the
  no-op.

  **Why `writeback=True` (GPTQ/AWQ/SmoothQuant) is unaffected:** that branch of `fsdp2_weight_access_and_writeback_context` gathers
  via `param.redistribute(... Replicate ...)` — a DTensor collective that doesn't touch the patched `unshard()`. So only the
  `writeback=False` (max/mse/local_hessian) path, i.e. the non-mutating optimization this PR adds, is broken on FSDP2.

  `_fsdp2_unshard_context` is internally self-contradictory under this caller: line 480 calls `unshard()` (needs it to work) while
  line 482 calls `_disable_fsdp_unshard_reshard` (disables it). It assumes it runs *before* anyone disables unshard, but its only
  production caller disables it first.

  **Impact / trigger:** FSDP2 layerwise calibration with `calib_mutates_weights=False`. The default (`True`) takes the redistribute
  path and works, so the bug only fires when a user opts into the non-mutating feature.

  **Test that should catch this** (likely unrun — it's a multi-GPU GPU test on a draft PR):
  `tests/gpu/torch/quantization/test_fsdp2.py:264`

  ```python
  with persistent_materialization(layer, writeback=False):
      assert not isinstance(layer[0].weight, DTensor)   # fails: still a DTensor
      layer(inputs)
  ```

  **Suggested fix (don't double-disable):**
  1. Have `_fsdp2_unshard_context` call the **unpatched** unshard (`_disable_fsdp_unshard_reshard` already captures `orig_unshard`;
  expose it), so the gather works while per-forward reshard stays suppressed; **or**
  2. Skip `persistent_materialization`'s outer `_disable_fsdp_unshard_reshard` on the FSDP `writeback=False` path, since
  `_fsdp2_unshard_context` already suppresses reshard internally *after* a real unshard.

modelopt/torch/quantization/config.py
```diff
@@ -794,12 +795,12 @@ def validate_layerwise_checkpoint_dir(self):
         return self
 
     @model_validator(mode="after")
-    def _validate_save_quantizers_only_supported(self):
-        """Enforce the ``_supports_save_quantizers_only`` whitelist."""
-        if self.layerwise.save_quantizers_only and not self._supports_save_quantizers_only:
+    def _validate_non_mutating_layerwise_supported(self):
+        """Enforce the ``calib_mutates_weights=False`` whitelist."""
+        if not self.layerwise.calib_mutates_weights and not self._supports_save_quantizers_only:
```
Fridah-nv, line 800: nit: we can rename `_supports_save_quantizers_only` to names like `_calib_is_amax_only` to align with new vocabulary, otherwise a reader has to learn that `_supports_save_quantizers_only=True` means "amax-only algorithm, so `calib_mutates_weights=False` is  allowed"

modelopt/torch/quantization/model_calib.py
```diff
@@ -1833,33 +1833,35 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+                    next_inputs = input_getter.cache_outputs_for_next_layer_calib(
+                        layer, forward_loop
+                    )
+                elif is_last:
+                    next_inputs = None
 
-            if ckpt:
-                ckpt.save(layer_idx, model, transformer_layers, next_inputs)
+                if ckpt:
+                    ckpt.save(layer_idx, model, transformer_layers, next_inputs)
```
Fridah-nv, line 1864: could you explain why we move `ckpt.save` inside the `persistent_materialization` context?

realAsma: We should materialize a layer once and do all the operations (get inputs, calibrate, save); outside of this context; the layer might be sharded or moved to disk in which case we need to materialize again. The point is to avoid redundant materialization.

modelopt/torch/quantization/utils/core_utils.py
realAsma: This is possible. can we move `_disable_fsdp_unshard_reshard(layer)`  to    `_disable_fsdp_reshard(layer)` ??

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -557,18 +584,23 @@ def enable_weight_access_and_writeback(module, root_model, name_to_module: dict
...
-      so per-forward hooks skip materialization/offloading.
+    - **Accelerate**: materializes weights, sets ``hook.offload = False``,
+      and bypasses the layer's top-level accelerate hook while the weights are
+      materialized.
     """
-    with _disable_fsdp_unshard_reshard(layer), enable_weight_access_and_writeback(layer, layer):
+    with (
+        _disable_fsdp_unshard_reshard(layer),
+        enable_weight_access_and_writeback(layer, layer, writeback=writeback),
+        temporarily_remove_accelerate_hook(layer),
```
realAsma, line 602: I added this during an experiment which saved the zeros_from_meta in meta device instead of actual device. 

However zeros_from_meta in meta device was breaking a test and I did not check that in - but I think we should modify that test instead.

modelopt/torch/quantization/utils/core_utils.py
realAsma: > _:robot: Bot comment._

Fixed in `9bfec17479`.

`persistent_materialization()` now enters `enable_weight_access_and_writeback(..., writeback=writeback)` before `_disable_fsdp_unshard_reshard(...)`, so the `writeback=False` FSDP2 path performs the real one-time unshard before per-forward unshard/reshard is suppressed.

I also added regression coverage for the actual public flow: `mtq.quantize(...)` with layerwise `calib_mutates_weights=False` on FSDP2 now asserts that the layer is materialized inside `persistent_materialization` (params are no longer `DTensor`s), and still checks the quantized FSDP2 output against the non-FSDP reference.

Validated with:

```bash
pytest_pwd tests/unit/torch/quantization/test_layerwise_calibrate.py -q
pytest_pwd CUDA_VISIBLE_DEVICES=2,3 tests/gpu/torch/quantization/test_fsdp2.py -k "layerwise_calibrate_fsdp2 or persistent_materialization" -q
```

Fridah-nv:
Move the changes in current PR to https://github.com/NVIDIA/Model-Optimizer/pull/1640 as the base branch changed

PR #1594, opened 2026-06-01
Add Alpamayo-1 example

rohansjoshi:
### What does this PR do?

Type of change: ? New example

Adds example for Alpamayo-1 quantization with ModelOpt (FP8, NVFP4, AutoQuant)

### Usage

```
python quantize.py --ckpt nvidia/Alpamayo-R1-10B --output-dir ./alpamayo-r1-fp8 --quantize fp8
```

## Summary by CodeRabbit

* **New Features**
  * Added Alpamayo 1 vision-language-action model quantization example supporting FP8, NVFP4, and mixed-precision optimization modes
  * Introduced CLI quantization tool with calibration loop and checkpoint export capabilities for both fake-quantized and real-quantized formats

* **Documentation**
  * Added comprehensive guide documenting the Alpamayo quantization example, model details, and usage instructions


examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,688 @@
...
+    num_traj_token = 48
+    hist_traj_placeholder = (
+        f"<|traj_history_start|>{'<|traj_history|>' * num_traj_token}<|traj_history_end|>"
+    )
+
+    return [
+        {
+            "role": "system",
+            "content": [
+                {
```
cjluo-nv, line 65: > _Bot comment._

These tokens contain literal backslashes (`<\|traj_history_start\|>`, etc.) — `\|` is not an escape sequence in Python, so the string literal keeps the backslash. Qwen-family special tokens are `<|traj_history_start|>` (no backslash); passing the backslash form to `apply_chat_template` will tokenize them as ordinary text and silently break calibration. Please drop the backslashes (and on 3.12+ this also avoids the `SyntaxWarning: invalid escape sequence`).

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,688 @@
...
+
+def to_device(
+    data: Any,
+    device: str | torch.device | None = None,
+    dtype: torch.dtype | None = None,
+) -> Any:
+    """Recursively cast data into the specified device, dtype."""
+    if isinstance(data, torch.Tensor):
+        data = data.to(
+            device=device,
```
cjluo-nv, line 114: > _Bot comment._

This duplicates `mto.enable_huggingface_checkpointing()` and pulls in three private symbols (`_LIBRARY_CLASSES_FOR_PATCHING`, `_PATCHED_CLASSES`, `patch_pretrained_methods`) just to skip the `_from_config` patch. Please use the public API, or — if the `_from_config` patch genuinely breaks AlpamayoR1 — add a public option to skip it rather than reaching into module internals from an example. Also, calling this at top-level import is a surprising side effect; prefer calling it from `main()`.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,688 @@
...
+    )
+    prompt_cache.crop(prefill_seq_len)
+    last_hidden = expert_out.last_hidden_state[:, -n_diffusion_tokens:]
+    v_pred = self.action_out_proj(last_hidden).view(b, *self.action_space.get_action_space_dims())
+
+    return {"v_pred": v_pred.to(torch.float32), "v_target": v_target}
+
+
+def patch_teacher_forced_flow_loss_forward() -> None:
+    """Attach teacher_forced_flow_loss_forward to AlpamayoR1 if missing.
```
cjluo-nv, line 247: > _Bot comment._

Module-level monkey-patching of an upstream class is fragile: the body assumes specific attrs/methods on `AlpamayoR1` (`action_in_proj` signature, `expert_non_causal_attention`, `fuse_traj_tokens`, `action_space.traj_to_action`, `action_space.get_action_space_dims`, `vlm.model.rope_deltas`, etc.). If the public `nvlabs/alpamayo` drifts, this will silently produce wrong gradients/loss for AutoQuantize. Please (a) pin an Alpamayo SHA in the README setup, and (b) add explicit `hasattr` assertions before patching so a drift fails loudly.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,688 @@
...
+    *,
+    clip_ids: list[str],
+    processor,
+    t0_us: int,
+    top_p: float,
+    temperature: float,
+    max_generation_length: int,
+    calibration_traj_samples: int,
+    device: str,
+):
```
cjluo-nv, line 270: > _Bot comment._

Docstring says "Tries common column names; falls back to index if needed", but the code only handles a literal `key` column and will `KeyError` on anything else. Either implement the fallback (e.g. try `clip_id`/`clipid`/index) or rewrite the docstring to match the actual behavior.

cjluo-nv (commented):

New ~700-line example. A few real concerns:

1. **Likely bug — backslash-escaped special tokens.** In `create_message`, the f-strings use `<\|traj_history_start\|>`, `<\|traj_history\|>`, `<\|traj_history_end\|>`, `<\|cot_start\|>`. In Python `\|` is not a recognized escape sequence, so the literal string contains a backslash before each pipe. Qwen-family special tokens are `<|...|>` without backslashes — passing the backslash form to `apply_chat_template` will not match the tokenizer's added special tokens and these will tokenize as ordinary text, silently corrupting calibration. Looks like a copy-paste from a markdown-escaped source. (Bonus: these will start emitting `SyntaxWarning: invalid escape sequence` on newer Python.)

2. **Reaches into private API instead of using the public one.** `enable_huggingface_checkpointing_patch()` re-implements `mto.enable_huggingface_checkpointing()` by importing the private `_LIBRARY_CLASSES_FOR_PATCHING`, `_PATCHED_CLASSES`, and `patch_pretrained_methods` from `modelopt.torch.opt.plugins.huggingface`, just to skip the `_from_config` patch. If the only goal is to skip `_from_config`, please make that an option on the public API (or document why this divergence is needed) instead of duplicating the loop with private symbols. Also, the file does this at import time as a top-level side effect, which is surprising.

3. **Monkey-patching `AlpamayoR1.teacher_forced_flow_loss_forward`** at module import time is acknowledged in a docstring but it's a strong coupling: the body is a verbatim copy of an internal training-fork method, and any drift in `nvlabs/alpamayo` (e.g. `action_in_proj` signature, `expert_non_causal_attention`, `action_space` API) will silently produce wrong calibration. At minimum: assert that the public `AlpamayoR1` exposes the methods/attrs this body relies on, and pin a known-good Alpamayo SHA in the README setup section instead of `git clone` of `main`.

4. **`read_clip_ids_from_parquet` doesn't match its docstring.** The docstring says it "tries common column names; falls back to index if needed", but the implementation hard-codes `cols_lower["key"]` and will `KeyError` on any parquet without a literal `key` column. Either implement the documented fallback or simplify the docstring.

5. **Binary parquet checked into the repo.** A `.parquet` calibration set is being added to `examples/alpamayo/`. The repo doesn't appear to use git-LFS (`.gitattributes` is missing). Even if this file is small, distributing calibration data through the example tree is unusual for this repo — most examples download from HF on demand. Please confirm the file size is small and that there are no licensing/redistribution restrictions on the AV calibration clips list, or fetch it at runtime instead.

6. **No tests / no CHANGELOG entry.** The PR template's test and changelog checkboxes are still unfilled. New examples in this repo (`vlm_ptq`, etc.) typically don't get unit tests, but at least a CHANGELOG entry under "New Features" is expected.

7. **Minor:** `args.parquet` default is the bare filename, joined to `script_dir`; if the user passes an absolute path it works (pathlib handles that), but a plain relative path in the user's CWD will be silently resolved against the script's directory, which is surprising. Consider documenting this or using `Path(args.parquet)` as-is when absolute/exists, else fall back to `script_dir / args.parquet`.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,688 @@
...
+from alpamayo_r1.models.token_utils import to_special_token
+from tqdm import tqdm
+from transformers import AutoProcessor, AutoTokenizer
+
+import modelopt.torch.quantization as mtq
+from modelopt.torch.export import export_hf_checkpoint
+from modelopt.torch.export.quant_utils import get_quant_config
+from modelopt.torch.opt.plugins.huggingface import (
+    _LIBRARY_CLASSES_FOR_PATCHING,
+    _PATCHED_CLASSES,
```
cjluo-nv: > _Bot comment._

`assert torch.ops.tensorrt.quantize_op.default` inside a try/except that only logs a warning on failure: this won't actually stop a broken install, and the assert form is also unusual here. Consider a plain attribute check + clearer error message, or remove this block (mtq.quantize will fail loudly later anyway).

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,688 @@
...
+            inputs = processor.apply_chat_template(
+                messages,
+                tokenize=True,
+                add_generation_prompt=False,
+                continue_final_message=True,
+                return_dict=True,
+                return_tensors="pt",
+            )
+            model_inputs = {
+                "tokenized_data": inputs,
```
cjluo-nv, line 441: > _Bot comment._

`disabled_layers` here uses fnmatch wildcards (per `auto_quantize` docs), but the literal module names appended below (`_name`) are not wildcards. The corresponding exclusion in `quantize_model` uses `f"{_name}.*"` — please be consistent. If you want exact-name matching here, prepend/append `*` or document the difference; otherwise an unwrapped name like `vlm.foo.bar` may not match the way you expect against quantizer names like `vlm.foo.bar.weight_quantizer`.

examples/alpamayo/README.md
```diff
@@ -0,0 +1,72 @@
...
+AV clips (`0417_16rows_train_set_for_calibration_25.10.parquet`).
+
+## Setup
+
+Clone Alpamayo and install it into the current environment so `alpamayo_r1` is
+importable:
+
+```bash
+git clone https://github.com/nvlabs/alpamayo
+pip install ./alpamayo
```
cjluo-nv, line 23: > _Bot comment._

Setup says `git clone https://github.com/nvlabs/alpamayo` without pinning a commit. Given that this example monkey-patches an internal method onto `AlpamayoR1`, please pin a known-good SHA so future upstream changes don't silently break calibration.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,652 @@
...
+        print("================== auto_quantize_model summary ==================")
+        mtq.print_quant_summary(model)
+
+    return model
+
+
+def main():
+    ap = argparse.ArgumentParser(description="Quantize AlpamayoR1 and export as HF checkpoint")
+    ap.add_argument(
+        "--ckpt",
```
cjluo-nv, line 508: > _Bot comment._

These two `print(...item()...)` calls run on every AutoQuantize forward/loss step and each `.item()` forces a CPU↔GPU sync (4 syncs per `forward_step`, 2 per `loss_func`). They were flagged in the previous review and don't appear to have been addressed in this commit. Please gate them behind `args.debug` (and ideally a rank check) and only compute `.item()` inside the conditional. Note that `quantization_args.debug` is also hardcoded to `True` at L573, so even adding the gate won't help until that's wired to the CLI.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,652 @@
...
+    )
+    ap.add_argument("--t0_us", type=int, default=5_100_000)
+    ap.add_argument("--top_p", type=float, default=0.98)
+    ap.add_argument("--temperature", type=float, default=0.6)
+    ap.add_argument("--max_generation_length", type=int, default=256)
+    ap.add_argument("--num_traj_samples", type=int, default=6)
+    ap.add_argument(
+        "--limit", type=int, default=16, help="How many clip_ids to use for calibration"
+    )
+    ap.add_argument(
```
cjluo-nv, line 551: > _Bot comment._

Inconsistent with the matching exclusion in `quantize_model` (L389 uses `f"{_name}.*"`). `auto_quantize`'s `disabled_layers` are fnmatch-matched against quantizer names like `vlm.foo.bar.weight_quantizer`, so a bare module name like `vlm.action_in_proj.linear` will not match and that Linear *will* be included in the search — exactly the not-multiple-of-16 layers you intended to keep in high precision. Please use the same `f"{_name}*"` or `f"*{_name}*"` pattern (and consider also wrapping `*lm_head*` consistently). This was raised in the previous review.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,652 @@
...
+
+    The public OSS AlpamayoR1 (github.com/nvlabs/alpamayo) does not define this
+    method; it exists only on the internal training fork. The body ported above
+    is the calibration path used by auto_quantize_model.
+    """
+    if not hasattr(AlpamayoR1, "teacher_forced_flow_loss_forward"):
+        AlpamayoR1.teacher_forced_flow_loss_forward = _teacher_forced_flow_loss_forward
+
+
+patch_teacher_forced_flow_loss_forward()
```
cjluo-nv, line 257: > _Bot comment._

Adding `hasattr(AlpamayoR1, "teacher_forced_flow_loss_forward")` is a good guard against double-patching, but the previous-review concern was the inverse: the *body* of `_teacher_forced_flow_loss_forward` reaches into `self.action_in_proj`, `self.config.expert_non_causal_attention`, `self.fuse_traj_tokens`, `self.action_space.traj_to_action`, `self.action_space.get_action_space_dims`, `self.vlm.model.rope_deltas`, etc. If upstream `nvlabs/alpamayo` renames any of these, this will silently produce wrong calibration. Please add explicit `hasattr` assertions for those attrs at patch time so a drift fails loudly, and consider moving the call into `main()` instead of running it as a top-level import side effect.

modelopt/torch/opt/plugins/huggingface.py
```diff
@@ -163,10 +168,11 @@ def enable_huggingface_checkpointing():
         model = AutoModelForCausalLM.from_pretrained(model_path).cuda()
 
     """
+    skip_methods = set(skip_methods or ())
     for name, (classes, methods_list) in _LIBRARY_CLASSES_FOR_PATCHING.items():
         for cls, patch_methods in zip(classes, methods_list):
             if cls in _PATCHED_CLASSES:
                 continue
```
cjluo-nv, line 169: > _Bot comment._

`skip_methods` is silently ignored if a name doesn't match a registered method — a typo (e.g. `{"_from_cfg"}` instead of `{"_from_config"}`) turns the whole escape hatch into a no-op with no signal. Please validate against the union of registered method names (first element of each entry across `_LIBRARY_CLASSES_FOR_PATCHING`) and raise `ValueError` on unknown names, or at minimum emit a warning.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,652 @@
...
+    # Patch PreTrainedModel.from_pretrained / save_pretrained so ModelOpt state is saved with the
+    # checkpoint (and restored when AlpamayoR1.from_pretrained later loads the quantized weights).
+    # We skip the `_from_config` patch: AlpamayoR1.__init__ builds its action expert via
+    # AutoModel.from_config (which routes through the patched _from_config), and the restore hook
+    # on that path interferes with that sub-module construction.
+    mto.enable_huggingface_checkpointing(skip_methods={"_from_config"})
+
+    device = "cuda"
+    print(f"Loading model from {args.ckpt!r} ...")
+    model = AlpamayoR1.from_pretrained(args.ckpt, dtype=torch.float16).to(
```
cjluo-nv, line 574: > _Bot comment._

Minor: `debug=True` is hardcoded here, so the `if args.debug:` guards in `quantize_model` / `auto_quantize_model` always fire and there's no way to turn the debug prints off from the CLI. Either expose this as `--debug` on the argparser or drop the conditional and the field entirely.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,652 @@
...
+            Required only when `calibration_forward_loop` is not provided.
+        calibration_forward_loop: Optional callable taking `model` and running
+            calibration forward passes. Use this for non-text modules whose
+            forward signature is not compatible with dataset_utils batches.
+
+    Returns:
+        Quantized model
+    """
+    # Create calibration forward loop. For standard text models we can build
+    # it from tokenizer-based data, but vision modules often need custom args.
```
cjluo-nv, line 354: > _Bot comment._

Docstring was updated to match the implementation, which is a fine resolution. One tiny robustness improvement: `cols_lower["key"]` will still `KeyError` if the parquet's column happens to be cased oddly (already handled — `cols_lower` is lowercased) or simply isn't named `key`. A one-line `KeyError` → `ValueError("expected a 'key' column, got: {list(df.columns)}")` would make the failure mode much friendlier. Non-blocking.

cjluo-nv (commented):

Several critical points from the previous round are addressed (backslash special tokens fixed, private-API duplication replaced with a public `skip_methods` arg, README now pins an Alpamayo SHA, CHANGELOG entry added, parquet docstring rewritten to match the code), but a few correctness/UX items raised earlier are still open in this commit:

- :speech_balloon: `forward_step` / `loss_func` debug `print(...item()...)` calls (raised by CodeRabbit and not addressed) — they still run unconditionally on every AutoQuantize iteration and force a per-batch CPU↔GPU sync. `args.debug` is even hardcoded to `True` at L573, so there's no way to turn them off without editing the script.
- :speech_balloon: `auto_quantize_model` `disabled_layers` (raised by cjluo-nv) still appends bare module names (`_name`) while `quantize_model` uses `f"{_name}.*"`. `auto_quantize`'s `disabled_layers` are fnmatch-matched against quantizer names like `vlm.foo.bar.weight_quantizer`, so the bare-name entries here will not match — the not-multiple-of-16 Linears that were the motivation for this exclusion are still being included in the AutoQuantize search. This is a real bug, not a style nit.
- :speech_balloon: Module-level `patch_teacher_forced_flow_loss_forward()` at import (cjluo-nv): a `hasattr`-on-the-class guard was added, but the request was for `hasattr` checks on the attributes the patch *body* depends on (`action_in_proj`, `expert_non_causal_attention`, `fuse_traj_tokens`, `action_space.traj_to_action`, `vlm.model.rope_deltas`, …) so upstream drift fails loudly instead of silently producing wrong calibration. Still not added, and the call is still a top-level side effect.
- New (PR-internal) issue: `enable_huggingface_checkpointing(skip_methods=...)` silently ignores unknown method names — a typo in `{"_from_config"}` would turn this escape hatch into a no-op without warning. Worth either validating against the registered method names or at least warning.
- No tests / no `example_tests.yml` entry for the new example, and a binary `.parquet` calibration set is still committed in-tree (no `.gitattributes` / LFS in this repo). The other VLM examples fetch calibration data on demand; please confirm size + redistribution rights, or fetch at runtime.

examples/alpamayo/README.md
```diff
@@ -0,0 +1,72 @@
...
+research. It takes multi-camera video and egomotion history as input and
+produces a Chain-of-Causation reasoning trace plus a future driving trajectory.
+See the paper, [*Alpamayo-R1: Bridging Reasoning and Action Prediction for
+Generalizable Autonomous Driving in the Long
+Tail*](https://arxiv.org/abs/2511.00088), and the
+[nvlabs/alpamayo](https://github.com/nvlabs/alpamayo) repository for details.
+
+This example produces FP8, NVFP4, and mixed-precision quantized checkpoints of
+Alpamayo using ModelOpt. Quantization calibration runs on a small dataset of 16
+AV clips (`0417_16rows_train_set_for_calibration_25.10.parquet`).
```
meenchen, line 14: Is this a public dataset? I would prefer not to keep this in our repo.

examples/alpamayo/quantize.py
```diff
@@ -0,0 +1,652 @@
...
+        default=6.5,
+        help="Effective-bits budget for AutoQuantize (only used when --quantize auto)",
+    )
+    ap.add_argument(
+        "--parquet",
+        type=str,
+        default="0417_16rows_train_set_for_calibration_25.10.parquet",
+        help="Parquet file with clip_ids for calibration",
+    )
+    ap.add_argument("--t0_us", type=int, default=5_100_000)
```
meenchen: Could you add a description for this argument?

examples/alpamayo/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+AutoQuantize chooses a per-layer format using a **gradient-based sensitivity
+score**: it backpropagates a loss through the model and estimates how much each
+candidate format perturbs that loss, then picks the cheapest assignment that
+stays within the bit budget. Here the loss is the flow-matching objective — an
+MSE between the action expert's predicted velocity field `v_pred` and the
+target `v_target = x_1 - x_0` from a teacher-forced forward pass on the
+calibration clips. Layers the loss is sensitive to keep more bits (FP8); the
+rest go to NVFP4.
```
meenchen, line 72: Is there any inference runtime available for the quantized checkpoint?

examples/alpamayo/README.md
rohansjoshi: The parquet file contains clip IDs (which can be publicly shared), the actual driving clips are on HF in the publicly available data nvidia/PhysicalAI-Autonomous-Vehicles

examples/alpamayo/README.md
rohansjoshi: Not that I'm aware of, there are just some PyTorch eval scripts lying around in different repos. I have validated accuracy of the FP8, NVFP4, and AutoQuant (6.5 bpe) checkpoints.

examples/alpamayo/quantize.py
rohansjoshi: Upstream `nvlabs/alpamayo` is unlikely to change and in any case a specific commit to use is specified in the README

examples/alpamayo/quantize.py
rohansjoshi: Dropping the field

examples/alpamayo/quantize.py
rohansjoshi: Invalid, and cost of syncs from item() calls is small

examples/alpamayo/quantize.py
rohansjoshi: Invalid

modelopt/torch/opt/plugins/huggingface.py
rohansjoshi: This is only used for Alpamayo, and won't be a no-op in practice, since patching or skipping patching certain methods is needed to load the model in the first place

cjluo-nv (commented):

Most critical issues from earlier rounds are now addressed: the backslash-escape bug in special tokens is fixed, the private-API duplication has been replaced by a real public `skip_methods` parameter on `enable_huggingface_checkpointing`, the README pins an Alpamayo SHA (`4cda35d`), a CHANGELOG entry was added, and the `read_clip_ids_from_parquet` docstring was rewritten to match the code. A few items still warrant a human reviewer's sign-off:

- :speech_balloon: Author marked the per-batch `.item()` debug prints in `forward_step` / `loss_func` (lines ~462–475) as "Invalid" without rationale. These still run unconditionally and force CPU↔GPU syncs every AutoQuantize step. Impact is bounded (16 clips × num_score_steps), but the rebuttal isn't substantiated — worth a quick human glance to confirm this is intended.
- :speech_balloon: Author marked the `auto_quantize_model` `disabled_layers` bare-name comment as "Invalid". I checked `_AutoQuantizeBaseSearcher.insert_hparams_after_merge_rules` — `disabled_layers` are fnmatched against `name` from `model.named_modules()`, i.e. module names like `vlm.foo.bar`, **not** quantizer names. So the bare names here are actually correct and the author's "Invalid" is right; the prior reviewer's note was mistaken. Just flagging so it doesn't get re-litigated.
- :speech_balloon: Module-level monkey-patch of `AlpamayoR1.teacher_forced_flow_loss_forward` (L257): author argues the pinned SHA in the README is sufficient defense against drift. The body still reaches into `action_in_proj`, `expert_non_causal_attention`, `fuse_traj_tokens`, `action_space.traj_to_action`, `vlm.model.rope_deltas`, etc. with no `hasattr` guards. Reasonable position given the SHA pin, but worth a human sanity check.
- :speech_balloon: meenchen asked whether the in-tree `0417_16rows_train_set_for_calibration_25.10.parquet` is a public dataset and expressed a preference not to keep it in the repo. Author responded that the clip IDs are publicly shareable (clips themselves are on HF). meenchen hasn't explicitly accepted that resolution; this is the most important remaining sign-off item — the file is binary, no `.gitattributes`/LFS in the repo, and other VLM examples fetch calibration data on demand rather than committing it.
- `enable_huggingface_checkpointing(skip_methods=...)` silently ignores unknown method names (a typo turns the escape hatch into a no-op). Author replied "only used for Alpamayo, won't be a no-op in practice", but this is now a public API surface and a one-line `ValueError` (or warning) on unknown names would be cheap insurance. Non-blocking.
- No tests / no `example_tests.yml` entry. Consistent with how other VLM examples are structured in this repo, so probably fine.

Licensing: new Python files carry the standard NVIDIA Apache-2.0 header; no third-party code copied; no `LICENSE`/`NOTICE` changes. The parquet question above is a redistribution / repo-policy concern rather than a license concern.

cjluo-nv:
### On the `skip_methods` argument

I'd push back on adding `skip_methods` to `enable_huggingface_checkpointing` as the fix here. It works, but it's a fairly leaky/blunt escape hatch:

- **It's process-global.** `skip_methods={"_from_config"}` disables `_from_config` restore for *every* class for the whole run, just to work around one model's constructor. Any legitimate `AutoModel.from_config(...)`-based restore elsewhere silently stops working.
- **It leaks internals into the public API.** Callers now have to know that the private `_from_config` method is the thing to skip.
- **It interacts badly with the `_PATCHED_CLASSES` cache.** Once enabled, a second `enable_huggingface_checkpointing(skip_methods=...)` call is a silent no-op (classes already in `_PATCHED_CLASSES`), so the skip set is effectively "first call wins" — a subtle footgun.

The underlying issue: `AlpamayoR1.__init__` builds its action expert via `AutoModel.from_config(...)`, which re-enters the patched `_from_config` restore machinery *during the parent's `__init__`*. That nested re-entry is what conflicts.

Two alternatives I'd suggest considering instead:

**1. No library change — reorder in the example.** The base checkpoint has no `modelopt_state.pth`, so the restore hook does nothing useful on the initial load; only `save_pretrained` needs patching here. So:

```python
model = AlpamayoR1.from_pretrained(args.ckpt, ...)   # load clean, no patches active
mto.enable_huggingface_checkpointing()               # enable AFTER load
# quantize -> save_pretrained writes modelopt_state.pth
```

Does `quantize.py` ever rely on restore-*during*-load? Reloading the quantized checkpoint looks like a separate deployment step. If load-time restore isn't needed in this script, `skip_methods` can be dropped entirely.

**2. If restore-on-load *is* needed — fix the re-entrancy in the library instead.** Add a thread-local "restore in progress" guard in `_patch_model_init_for_modelopt` so only the *outermost* `from_pretrained`/`_from_config` triggers a restore, and nested sub-module construction skips it. That fixes the problem for *any* composite/VLM model that builds sub-models via `AutoModel.from_config` — not just Alpamayo — without exposing a per-caller knob or requiring knowledge of internal method names.

Happy to help draft the re-entrancy guard if option 2 sounds right.

---

Minor, unrelated: this PR commits a binary `.parquet` calibration file into the repo — worth double-checking whether that should be hosted/downloaded rather than checked into git, for repo-bloat reasons.

jingyu-ml (approved):
LGTM, Could you add requirements and a few simple test cases to verify that the example scripts work?
https://github.com/NVIDIA/Model-Optimizer/tree/main/tests/examples

PR #1591, opened 2026-06-01
eval skill: parameterize external judge/user-sim endpoints via .env

cjluo-nv:
### What does this PR do?

Type of change: documentation

Several AA tasks call an external **judge / user-simulator / scoring endpoint** whose `model_id` + `url` vary per user/site (HLE, AA-LCR, Tau2 today — and the guidance is written as a general pattern for any future such benchmark). Previously each recipe hardcoded `<...>` placeholders that a user had to hand-edit in every config. This makes them reusable **without committing any internal infrastructure**:

- **`recipes/env.example`**: add placeholders — `NS_JUDGE_URL`, `HLE_JUDGE_MODEL_ID`, `LCR_JUDGE_MODEL_ID`, `TAU2_USER_MODEL_ID`, `TAU2_JUDGER_MODEL_ID`, `TAU2_ENDPOINT_URL` — with the **recommended model named** (GPT-4o / Qwen3 235B / gpt-oss-120B) but only **generic hosts** (`https://<your-inference-host>/v1`). No internal hostnames or gateway model-routing strings are committed; real values live in the user's gitignored `.env`.
- **`recipes/tasks/aa/{hle,lcr,tau2_bench_telecom}.md`**: carry `<VAR>` literal placeholders (named after the `.env` keys) that the skill **substitutes as literal values** from the user's `.env`. These are *config, not secrets*, so they are **not** exported — which avoids the `${oc.env:...}` footgun (it silently fails unless the var was exported with `set -a`). Only `api_key` (`INFERENCE_API_KEY`) stays an exported env var read by the harness.
- **`SKILL.md`** (Step 5): instructs literal substitution from `.env`, framed as a general pattern for any external-endpoint task.

The `/v1`-base (nemo-skills) vs full `/v1/chat/completions` (tau2-bench) URL distinction is documented.

### Usage

N/A — documentation / skill-template only.

### Testing

`pre-commit run` passes (markdownlint). Verified no internal hostnames / gateway model IDs are present in any committed file.

### Before your PR is "*Ready for review*"

- Is this change backward compatible?: ✅
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A
- Did you write any new necessary tests?: N/A (documentation)
- Did you update Changelog?: N/A (skill docs)
- Did you get Claude approval on this PR?: ❌ (pending)

### Additional Information

[...]

PR #1493, opened 2026-05-14
Agent Skills Updates From Live Trials

chadvoegele:
### What does this PR do?

Type of change: bug fix

### Usage

Ask Claude Code:
```
Quantize `mistralai/Mistral-Medium-3.5-128B` to NVFP4 using the ModelOpt NVFP4 experts-only recipe.

Run on $cluster

Evaluate the resulting quantized checkpoint on:
- GPQA Diamond AA v3
- SciCode AA v2

Complete the quantization and evaluation workflow end to end. Prompt when you require user input, otherwise keep going.
```

### Testing
I'm running the full loop with the above prompt, and iterating on skills to resolve undesired agent behavior.

### Before your PR is "*Ready for review*"

Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md) and your commits are signed (`git commit -s -S`).

Make sure you read and follow the [Security Best Practices](https://github.com/NVIDIA/Model-Optimizer/blob/main/SECURITY.md#security-coding-practices-for-contributors) (e.g. avoiding hardcoded `trust_remote_code=True`, `torch.load(..., weights_only=False)`, `pickle`, etc.).

- Is this change backward compatible?: ✅
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅
- Did you write any new necessary tests?: N/A
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅
- Did you get Claude approval on this PR?: TODO

### Additional Information
See trials log for details.

## Release Notes

* **New Features**
  * Added SLURM Quality of Service (QoS) configuration support for job submission
  * Introduced 8 new evaluation task recipes (AIME 2025, GPQA, IFBench, LiveCodeBench, SciCode, AA-LCR, HLE-AA, MMMU-Pro, tau2_bench)
  * Enhanced job monitoring with continuous polling-based tracking

* **Documentation**
  * Restructured evaluation workflow with explicit dry-run, canary, and full-run validation stages
  * Expanded PTQ validation with mandatory pre-deployment verification gates
  * Updated remote cluster selection and quantization detection guidance

* **Tests**
  * Updated evaluation test expectations to reflect refined workflow stages

[...]

PR #1595, opened 2026-06-01
[skills] Fix eval/deploy defaults for NVFP4-on-Blackwell + nemo-skills evals

Edwardf0t1:
### What does this PR do?

Type of change: documentation (agent skills under `.claude/skills/**`)

Fixes five agent-skill defaults/docs that each caused a **real failure** during a
Qwen3.5-9B NVFP4 PTQ → deploy → Artificial-Analysis eval run on B300/GB300:

1. **`example_eval.yaml` missing `--served-model-name`** — vLLM registered the
   model under the path `/checkpoint`, so every eval request 404'd
   (`The model '<served_model_name>' does not exist`). Added
   `--served-model-name ${deployment.served_model_name}` to the template command
   and documented it as a required convention in the evaluation SKILL.
2. **nemo-skills `DUMMY_API_KEY`** — `ns_*` tasks hard-fail
   (`api_key_env_var=DUMMY_API_KEY but the value is not set`) unless the api-key
   env var is set *inside the container*; a shell `export` does **not** reach the
   SLURM container. Added `DUMMY_API_KEY: lit:dummy` to the example's
   `evaluation.env_vars` and corrected the Step 5 known-issue + Step 8 note.
3. **NEL env-var value prefixes** — documented the required
   `host:` / `lit:` / `runtime:` prefixes (a bare value hard-errors:
   *"Env var value '…' must have an explicit prefix"*).
4. **`execution.gres` / `gpus_per_node`** — must match the cluster node GPU count
   and QOS, or `sbatch` rejects the job (`Requested node configuration is not
   available` on the eval side, `QOSMinGRES` on the PTQ side). Documented in the
   evaluation SKILL (Step 4) and the PTQ launcher guide.
5. **NVFP4 on Blackwell needs `cu130-nightly`** — `v0.19.1` and any `cu129`
   (CUDA 12.9) vLLM build lack sm_103 FP4 kernels; the server loads the checkpoint
   then dies at engine init with `CUDA error: no kernel image is available`
   (affects `flashinfer` and `cutlass` NVFP4 backends; `marlin` separately fails
   on non-64-divisible layer dims). Documented in the deployment + evaluation
   SKILLs, including `--mm-encoder-attn-backend TRITON_ATTN` for multimodal models
   on sm_103 and the raw-markdown `recipes.vllm.ai` fallback for hardware variants.

Closely-related bonus (same launcher-guide file): `EXTRA_PIP_DEPS` must avoid
shell `>`/`<` metacharacters (use `==` pins) — they are mangled in the launcher's
unquoted sbatch `export` and silently dropped.

### Usage

N/A — agent-skill docs and a config template only; no library/API/code change.

### Testing

- `example_eval.yaml` validated with `yaml.safe_load`; confirmed the command
  includes `--served-model-name` and `evaluation.env_vars` includes
  `DUMMY_API_KEY: lit:dummy`.
- Each fix was empirically validated end-to-end in this session: with all five
  applied, a Qwen3.5-9B NVFP4 checkpoint deploys on B300 (cu130-nightly) and the
  AA suite runs via NEL (GPQA scored; SciCode/IFBench/MMMU-Pro/HLE running).
- Trailing-whitespace / EOF-newline checked on all changed files (`pre-commit`
  is not installed in this environment).

- Is this change backward compatible?: ✅ (docs/templates only)
- Did you update Changelog?: N/A (agent-skill docs, not a library feature/API/critical-bug change)
- Did you get Claude approval on this PR?: ❌ (will run `/claude review`)

### Additional Information

Skills touched: `evaluation` (`SKILL.md`, `recipes/examples/example_eval.yaml`),
`deployment` (`SKILL.md`), `ptq` (`references/launcher-guide.md`).

## Summary by CodeRabbit

* **Documentation**
  * Clarified NVFP4-on-Blackwell requirements: use CUDA‑13 vLLM images (cu130 or cu130-nightly) and add the multimodal encoder flag for sm_103 FP4 support.
  * Added image-selection verification guidance and explicit vLLM serve options (served-model-name, host/port) for reliable endpoints.
  * Improved SLURM guidance: prefer checkpoint_path, set per-node GPU counts (gres), and align parallel sizes.
  * Documented eval env var handling (lit-style DUMMY_API_KEY) and warned about unquoted pip-deps injection risks.

### DIFF
--- a/.claude/skills/deployment/SKILL.md
+++ b/.claude/skills/deployment/SKILL.md
@@ -125,6 +125,18 @@ python -m vllm.entrypoints.openai.api_server \
 
 For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 
+> **NVFP4 on Blackwell needs the CUDA-13 vLLM build.** On B200/B300/GB200/GB300
+> (compute capability sm_100/sm_103), use `vllm/vllm-openai:cu130-nightly-<arch>`
+> (`-x86_64`, or `-aarch64` on Grace). The common `v0.19.1` / any `cu129`
+> (CUDA 12.9) build has **no sm_103 FP4 kernels** — vLLM loads the checkpoint
+> then dies at engine init with `CUDA error: no kernel image is available for
+> execution on the device` (affects the `flashinfer` and `cutlass` NVFP4
+> backends; `marlin` separately fails on non-64-divisible layer dims). Verify the
+> image via `recipes.vllm.ai/<org>/<model>?hardware=b300` (JS-rendered — fetch the
+> raw markdown at `github.com/vllm-project/recipes/blob/main/<org>/<model>.md`).
+> For multimodal models on sm_103, also pass `--mm-encoder-attn-backend
+> TRITON_ATTN` (the default CuTe ViT flash-attn asserts "Only SM 10.x and 11.x").
+
 #### SGLang
 
 ```bash
--- a/.claude/skills/deployment/SKILL.md
+++ b/.claude/skills/deployment/SKILL.md
@@ -125,17 +125,18 @@ python -m vllm.entrypoints.openai.api_server \
 
 For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 
-> **NVFP4 on Blackwell needs the CUDA-13 vLLM build.** On B200/B300/GB200/GB300
-> (compute capability sm_100/sm_103), use `vllm/vllm-openai:cu130-nightly-<arch>`
-> (`-x86_64`, or `-aarch64` on Grace). The common `v0.19.1` / any `cu129`
-> (CUDA 12.9) build has **no sm_103 FP4 kernels** — vLLM loads the checkpoint
+> **NVFP4 on Blackwell B300/GB300 (sm_103): append `-cu130` to the image tag**
+> (e.g. `vllm/vllm-openai:v0.19.1-cu130` — release tags are multi-arch). The
+> default cu12 build has **no sm_103 FP4 kernel**, so vLLM loads the checkpoint
 > then dies at engine init with `CUDA error: no kernel image is available for
 > execution on the device` (affects the `flashinfer` and `cutlass` NVFP4
-> backends; `marlin` separately fails on non-64-divisible layer dims). Verify the
-> image via `recipes.vllm.ai/<org>/<model>?hardware=b300` (JS-rendered — fetch the
-> raw markdown at `github.com/vllm-project/recipes/blob/main/<org>/<model>.md`).
-> For multimodal models on sm_103, also pass `--mm-encoder-attn-backend
-> TRITON_ATTN` (the default CuTe ViT flash-attn asserts "Only SM 10.x and 11.x").
+> backends; `marlin` separately fails on non-64-divisible layer dims). If a
+> pinned release predates the model's arch, use `cu130-nightly-<arch>` instead
+> (Qwen3.5-9B's `qwen3_5` needed it). Cross-check via
+> `recipes.vllm.ai/<org>/<model>?hardware=b300` (JS-rendered — fetch the raw
+> markdown at `github.com/vllm-project/recipes/blob/main/<org>/<model>.md`). For
+> multimodal models on sm_103, also pass `--mm-encoder-attn-backend TRITON_ATTN`
+> (the default CuTe ViT flash-attn asserts "Only SM 10.x and 11.x").
 
 #### SGLang
 
--- a/.claude/skills/evaluation/SKILL.md
+++ b/.claude/skills/evaluation/SKILL.md
@@ -136,11 +136,13 @@ deployment:
     <... rest of cross-checked flags ...>
 ```
 
-Conventions: always start `vllm serve /checkpoint` (NEL mounts here); always `--host 0.0.0.0 --port ${deployment.port}`; use folded scalar (`>-`) for one flag per line. Example fallback `--max-model-len 131072` covers AA-LCR (~120K + 16K gen) and SciCode (≥ 65536) — prefer `config.json` / recipe value.
+Conventions: always start `vllm serve /checkpoint` (NEL mounts here); always `--served-model-name ${deployment.served_model_name}` (**required** — without it vLLM registers the model under the path `/checkpoint`, and every eval request 404s with "The model `<served_model_name>` does not exist."); always `--host 0.0.0.0 --port ${deployment.port}`; use folded scalar (`>-`) for one flag per line. Example fallback `--max-model-len 131072` covers AA-LCR (~120K + 16K gen) and SciCode (≥ 65536) — prefer `config.json` / recipe value.
 
 For how to choose `--tensor-parallel-size` / `--data-parallel-size` / `--pipeline-parallel-size` (and EP) from the model size and your GPU count, read `references/parallelism.md` — cross-check the layout against `recipes.vllm.ai`, then adapt to the GPUs you actually have via the fit math there.
 
 **Image / vLLM version.** Default `image: vllm/vllm-openai:v0.19.1` (pinned for reproducibility). If `recipes.vllm.ai` states a higher minimum version for the chosen variant (e.g. "vLLM >= 0.20.0"), bump the image tag accordingly (e.g. `v0.20.0`) — do **not** stay on `0.19.1` when the recipe explicitly requires newer. Do **not** use `:latest` (drifts across re-runs, breaks reproducibility). The version is part of the cross-check: surface to the user when bumping.
+
+> **NVFP4 on Blackwell needs the CUDA-13 build.** Serving an NVFP4 checkpoint on Blackwell (B200/B300/GB200/GB300, compute capability sm_100/sm_103) requires `vllm/vllm-openai:cu130-nightly-<arch>` (`-x86_64`, or `-aarch64` on Grace). The pinned `v0.19.1` and **all** `cu129` (CUDA 12.9) builds lack sm_103 FP4 kernels — the server loads the checkpoint then dies at engine init with `CUDA error: no kernel image is available for execution on the device` (true for the `flashinfer` *and* `cutlass` NVFP4 backends; `marlin` separately fails on non-64-divisible layer dims). This is the vLLM-recipe-recommended Blackwell image — confirm via `recipes.vllm.ai/<org>/<model>?hardware=b300` (and since that page is JS-rendered, fetch the raw markdown at `github.com/vllm-project/recipes/blob/main/<org>/<model>.md`). For **multimodal** models on sm_103, also add `--mm-encoder-attn-backend TRITON_ATTN` — the default CuTe ViT flash-attn kernel asserts "Only SM 10.x and 11.x are supported" on sm_103.
 
 #### vLLM-backend defaults — always include unless the recipe *contradicts*
 
@@ -196,6 +198,7 @@ Reasoning models: prefer reasoning mode (highest scores). For lower variance / c
 - Find every `???` left. Ask the user only for what can't be inferred (SLURM hostname/account/output_dir, MLflow tracking URI, etc.). Don't propose defaults; let them give plain text.
 - **`parallelism`** — size it yourself from the run shape (total requests = `dataset_size × repeats` vs GPU serving capacity), and set `--max-num-seqs` to match. Read `references/parallelism.md` for the decision rule and worked examples; only ask the user if a non-GPU cap (e.g. judge rate limit) is unknown.
 - Ask about other defaults they may want to change (partition, walltime, MLflow tags).
+- **`execution.gres`** — NEL defaults to `gpu:8`. Set it to the cluster's per-node GPU count (and what the QOS permits), and match `--data-parallel-size`/`--tensor-parallel-size` to it. A mismatch makes `sbatch` reject the job with *"Requested node configuration is not available"* (e.g. `gpu:8` on 4-GPU GB300 nodes → set `gres: gpu:4`). Confirm the node GPU count with `sinfo -o '%P %G'` on the target cluster.
 
 **Walltime cap: 4 hours.** Always `execution.walltime: "04:00:00"`. The cluster does not schedule jobs longer than 4h — this is a hard limit, not a preference.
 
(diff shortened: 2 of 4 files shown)


.claude/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -43,6 +43,9 @@ execution:
   account: ???
   output_dir: ???
   walltime: "04:00:00"
+  # gres defaults to gpu:8. Set it to the cluster's per-node GPU count (and what
+  # the QOS allows) or sbatch fails "Requested node configuration is not available".
+  # gres: gpu:4   # e.g. GB300 nodes (4 GPUs); also drop --data-parallel-size to match
```
kaix-nv: replaced `also drop --data-parallel-size to match` with `match  --data-parallel-size/--tensor-parallel-size to it`?

kaix-nv:
Do we want to add a warning msg for Blackwell-only context? Native FP4 GEMM kernels exist only on sm_100/sm_103. On Hopper/Ada/Ampere there's no native FP4 path and NVFP4 falls back to the marlin weight-dequant backend (slower).

kaix-nv (approved):

.claude/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -43,6 +50,9 @@ execution:
   account: ???
   output_dir: ???
   walltime: "04:00:00"
+  # gres defaults to gpu:8. Set it to the cluster's per-node GPU count (and what
```
cjluo-nv: If you install @chadvoegele 's cluster-reference skill, this should be well taken care of

.claude/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -55,13 +65,24 @@ deployment:
   hf_model_handle:
   served_model_name: ???
   image: vllm/vllm-openai:v0.19.1
+  # NVFP4 on Blackwell (B200/B300/GB200/GB300, sm_100/sm_103): the v0.19.1 and
+  # any cu129 builds have NO sm_103 FP4 kernels (deploy dies with CUDA
+  # "no kernel image is available"). Use the CUDA-13 nightly instead:
+  #   image: vllm/vllm-openai:cu130-nightly-x86_64    # or -aarch64 on Grace
+  # (vLLM-recipe-recommended Blackwell image; see recipes.vllm.ai ?hardware=b300).
+  # For multimodal models on sm_103 also add `--mm-encoder-attn-backend TRITON_ATTN`.
```
cjluo-nv: Is TRITON_ATTN strictly required?

.claude/skills/evaluation/SKILL.md
```diff
@@ -136,12 +138,14 @@ deployment:
...
 
+> **NVFP4 on Blackwell needs the CUDA-13 build.** Serving an NVFP4 checkpoint on Blackwell (B200/B300/GB200/GB300, compute capability sm_100/sm_103) requires `vllm/vllm-openai:cu130-nightly-<arch>` (`-x86_64`, or `-aarch64` on Grace). The pinned `v0.19.1` and **all** `cu129` (CUDA 12.9) builds lack sm_103 FP4 kernels — the server loads the checkpoint then dies at engine init with `CUDA error: no kernel image is available for execution on the device` (true for the `flashinfer` *and* `cutlass` NVFP4 backends; `marlin` separately fails on non-64-divisible layer dims). This is the vLLM-recipe-recommended Blackwell image — confirm via `recipes.vllm.ai/<org>/<model>?hardware=b300` (and since that page is JS-rendered, fetch the raw markdown at `github.com/vllm-project/recipes/blob/main/<org>/<model>.md`). For **multimodal** models on sm_103, also add `--mm-encoder-attn-backend TRITON_ATTN` — the default CuTe ViT flash-attn kernel asserts "Only SM 10.x and 11.x are supported" on sm_103.
```
cjluo-nv: this is a duplicated statement, which exists in the examples.

Also we should recommend vllm/vllm-openai:v0.19.0-cu130 to match the pinned docker. If this release cannot support the model, we then also can update the docker image to a later release.

.claude/skills/deployment/SKILL.md
```diff
@@ -125,6 +125,18 @@ python -m vllm.entrypoints.openai.api_server \
 
 For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 
+> **NVFP4 on Blackwell needs the CUDA-13 vLLM build.** On B200/B300/GB200/GB300
+> (compute capability sm_100/sm_103), use `vllm/vllm-openai:cu130-nightly-<arch>`
+> (`-x86_64`, or `-aarch64` on Grace). The common `v0.19.1` / any `cu129`
+> (CUDA 12.9) build has **no sm_103 FP4 kernels** — vLLM loads the checkpoint
+> then dies at engine init with `CUDA error: no kernel image is available for
```
chadvoegele, line 131: This `CUDA error: no kernel image is available for execution on the device`, description and fix seem like a good candidate for the debugging playbook (/ shared memory) that I attempted in [1493](https://github.com/NVIDIA/Model-Optimizer/pull/1493/commits). Perhaps we can restart the conversation on how to best develop shared memory for the agent?

- https://github.com/NVIDIA/Model-Optimizer/pull/1493/commits
- https://github.com/NVIDIA/Model-Optimizer/pull/1493/changes/6c72cac3474e32d4c955937e4d1b90312a1c1af4

.claude/skills/evaluation/recipes/examples/example_eval.yaml
chadvoegele: DP/TP settings are now covered extensively in https://github.com/NVIDIA/Model-Optimizer/blob/main/.claude/skills/evaluation/references/parallelism.md. Can we refer to that?

.claude/skills/evaluation/SKILL.md
```diff
@@ -136,12 +138,14 @@ deployment:
     <... rest of cross-checked flags ...>
 ```
 
-Conventions: always start `vllm serve /checkpoint` (NEL mounts here); always `--host 0.0.0.0 --port ${deployment.port}`; use folded scalar (`>-`) for one flag per line. Example fallback `--max-model-len 131072` covers AA-LCR (~120K + 16K gen) and SciCode (≥ 65536) — prefer `config.json` / recipe value.
+Conventions: always start `vllm serve /checkpoint` (NEL mounts here); always `--served-model-name ${deployment.served_model_name}` (**required** — without it vLLM registers the model under the path `/checkpoint`, and every eval request 404s with "The model `<served_model_name>` does not exist."); always `--host 0.0.0.0 --port ${deployment.port}`; use folded scalar (`>-`) for one flag per line. Example fallback `--max-model-len 131072` covers AA-LCR (~120K + 16K gen) and SciCode (≥ 65536) — prefer `config.json` / recipe value.
```
chadvoegele: Can we de-duplicate? The `served_model_name` and `/checkpoint` was repeated already a few times.

Edwardf0t1:
Thanks for the reviews! Rebased onto latest `main` (resolved the `SKILL.md` conflict with #1591, keeping both the external-judge guidance and the nemo-skills `DUMMY_API_KEY`/env-prefix notes) and addressed the comments:

- **Image rec (@cjluo-nv):** now recommend the pinned **`vllm/vllm-openai:v0.19.1-cu130`** (matches the default `v0.19.1`), bumping to `cu130-nightly-<arch>` only if a pinned release lacks the model's arch. Heads-up: I did **not** test `v0.19.x-cu130` — Qwen3.5-9B's `qwen3_5` was verified on **`cu130-nightly-x86_64` (vLLM 0.19.2rc1.dev134)**; `v0.19.2-cu130` doesn't exist as a tag, and a pinned 0.19.0/0.19.1 may predate `qwen3_5`. Noted that caveat in the docs.
- **De-dup (@chadvoegele, @cjluo-nv):** trimmed the Step 3 Blackwell note in `SKILL.md` and pointed to `example_eval.yaml` for the full version.
- **DP/TP (@chadvoegele, @kaix-nv):** gres comment now says "match `--tensor-parallel-size`/`--data-parallel-size` to it" and refers to `references/parallelism.md`.
- **`TRITON_ATTN` (@cjluo-nv):** not strictly required in general — it's a workaround for the ViT CuTe flash-attn `Only SM 10.x/11.x` assert on the images we tested on sm_103; flagged it as "may be unneeded on newer builds".
- **transformers pin (@Copilot):** clarified the `EXTRA_PIP_DEPS` version is model-specific (e.g. `qwen3_5` needs 5.5.0), not the repo's library pin.

Not done here (broader, happy to follow up): @chadvoegele's suggestion to fold the "no kernel image" finding into a shared debugging playbook (#1493), and @cjluo-nv's cluster-reference-skill point for the gres/node handling.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
Edwardf0t1: Done — reworded to `match --tensor-parallel-size/--data-parallel-size to it`, and added a pointer to `references/parallelism.md` for the sizing rule.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
Edwardf0t1: Agreed — the gres/node handling is a good fit for @chadvoegele's cluster-reference skill. I kept this PR self-contained rather than take that dependency now, but happy to migrate the gres guidance there as a follow-up.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
Edwardf0t1: Not in general. `--mm-encoder-attn-backend TRITON_ATTN` is a workaround for the ViT CuTe flash-attn assert (`Only SM 10.x and 11.x are supported`) we hit on **sm_103** with the images we tested — text-only paths don't need it. I updated the note to mark it optional ("may be unneeded on newer builds").

.claude/skills/evaluation/SKILL.md
Edwardf0t1: Trimmed this — it now points to `example_eval.yaml` for the full note (de-dup). On the image: switched the recommendation to the pinned **`vllm/vllm-openai:v0.19.1-cu130`** to match the default image. One caveat worth recording: we only verified `qwen3_5` on `cu130-nightly-x86_64` (vLLM `0.19.2rc1.dev134`); `v0.19.2-cu130` doesn't exist as a tag and a pinned `0.19.x` release may predate `qwen3_5`, so I documented the nightly as the fallback "if the pinned release lacks the arch."

.claude/skills/deployment/SKILL.md
Edwardf0t1: +1 — this is a good playbook candidate. I didn't fold it into shared memory here to keep the PR scoped, but happy to restart the #1493 discussion and migrate the "no kernel image → cu130" finding into the debugging playbook.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
Edwardf0t1: Done — the gres comment now refers to `references/parallelism.md` for DP/TP sizing instead of inlining it.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
cjluo-nv: Could you take a look at https://github.com/NVIDIA/Model-Optimizer/pull/1599/changes? That should solve this issue and aligns better with the internal NEL guide

.claude/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -55,13 +65,30 @@ deployment:
   hf_model_handle:
   served_model_name: ???
   image: vllm/vllm-openai:v0.19.1
+  # NVFP4 on Blackwell (B200/B300/GB200/GB300, sm_100/sm_103): the v0.19.1 and
+  # any cu129 builds have NO sm_103 FP4 kernels (deploy dies with CUDA
+  # "no kernel image is available"). Use a CUDA-13 build — prefer the pinned
+  # release matching this image, bump to nightly only if it lacks the arch:
```
cjluo-nv: will there be a case that the release docker lack the arch?

.claude/skills/evaluation/SKILL.md
```diff
@@ -136,12 +138,14 @@ deployment:
...
 
 **Image / vLLM version.** Default `image: vllm/vllm-openai:v0.19.1` (pinned for reproducibility). If `recipes.vllm.ai` states a higher minimum version for the chosen variant (e.g. "vLLM >= 0.20.0"), bump the image tag accordingly (e.g. `v0.20.0`) — do **not** stay on `0.19.1` when the recipe explicitly requires newer. Do **not** use `:latest` (drifts across re-runs, breaks reproducibility). The version is part of the cross-check: surface to the user when bumping.
 
+> **NVFP4 on Blackwell needs a CUDA-13 vLLM build.** On B200/B300/GB200/GB300 (sm_100/sm_103) the pinned `v0.19.1` and all `cu129` builds lack sm_103 FP4 kernels — engine init dies with `CUDA error: no kernel image is available for execution on the device`. Use `vllm/vllm-openai:v0.19.1-cu130` (pinned, matches the default image), and bump to `cu130-nightly-<arch>` only if it lacks the model's arch (Qwen3.5-9B's `qwen3_5` needed the nightly). Multimodal on sm_103 also needs `--mm-encoder-attn-backend TRITON_ATTN` (ViT flash-attn workaround). Full note + the `recipes.vllm.ai ?hardware=b300` lookup are in `recipes/examples/example_eval.yaml`.
```
cjluo-nv: usually vllm docker release is multi arch. I think we can just say: if the GPU is B300 or GB300, append -cu130 on the docker image tag.

.claude/skills/evaluation/SKILL.md
```diff
@@ -196,6 +200,7 @@ Reasoning models: prefer reasoning mode (highest scores). For lower variance / c
 - Find every `???` left. Ask the user only for what can't be inferred (SLURM hostname/account/output_dir, MLflow tracking URI, etc.). Don't propose defaults; let them give plain text.
 - **`parallelism`** — size it yourself from the run shape (total requests = `dataset_size × repeats` vs GPU serving capacity), and set `--max-num-seqs` to match. Read `references/parallelism.md` for the decision rule and worked examples; only ask the user if a non-GPU cap (e.g. judge rate limit) is unknown.
 - Ask about other defaults they may want to change (partition, walltime, MLflow tags).
+- **`execution.gres`** — NEL defaults to `gpu:8`. Set it to the cluster's per-node GPU count (and what the QOS permits), and match `--data-parallel-size`/`--tensor-parallel-size` to it. A mismatch makes `sbatch` reject the job with *"Requested node configuration is not available"* (e.g. `gpu:8` on 4-GPU GB300 nodes → set `gres: gpu:4`). Confirm the node GPU count with `sinfo -o '%P %G'` on the target cluster.
```
cjluo-nv: hope this should be taken care of by https://github.com/NVIDIA/Model-Optimizer/pull/1599/changes as well.

cjluo-nv (commented):
Could you compact the added skill context as well? The agent should be able to help you on it.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
Edwardf0t1: Yes — that's exactly what we hit: `qwen3_5` (Qwen3.5-9B) wasn't in any pinned `0.19.x` release, only the `cu130-nightly` (vLLM `0.19.2rc1.dev134`). So for brand-new archs the pinned release can lack it. I kept it to a single fallback line; reworded the main guidance to your "append `-cu130`" framing.

.claude/skills/evaluation/SKILL.md
Edwardf0t1: Done — reworded to exactly that: "**B300/GB300 (sm_103): append `-cu130` to the image tag**" (noting release tags are multi-arch, e.g. `v0.19.1-cu130`), with a one-line nightly fallback only for archs a pinned release predates (qwen3_5). Trimmed the verbose version. Same wording applied in the deployment skill.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
Edwardf0t1: Reviewed #1599 — agreed it's the better solution for the cluster/gres path: `internal/slurm/<cluster>` pre-fills gres/hostname/partition so this gotcha disappears entirely when the internal package is present. I've **deferred to it** and reduced my note here to a single fallback line for the `slurm/default` (no-internal-package) case — which #1599's fallback branch doesn't currently mention for gres. So the two are complementary. 

One concern for #1599 is the disclosure of internal package in public skills. Left a comment there.

.claude/skills/evaluation/SKILL.md
Edwardf0t1: Done — deferred to #1599. This bullet is now a one-liner: "auto-set if you used a predefined `internal/slurm/<cluster>` config (above); on the `slurm/default` fallback it's gpu:8, set to the node count…". It only covers the gap #1599's manual-fallback branch leaves (that branch lists hostname/account/output_dir but not gres).

.claude/skills/evaluation/SKILL.md
Edwardf0t1: Now fully single-sourced: the `--served-model-name` rationale (vLLM serves it as `/checkpoint` → eval 404s) lives **only** in `example_eval.yaml` (1 line next to the command). The SKILL Step-3 Conventions line is reduced to naming the required flag + "see `example_eval.yaml` for why" — no repeated explanation. (commit `96408be101`)

PR #1599, opened 2026-06-02
eval skill: auto-detect predefined per-cluster execution configs

cjluo-nv:
### What does this PR do?

Type of change: documentation

Some NEL installs ship ready-made per-cluster execution configs as an `internal/slurm/<cluster>` group (via the optional `nemo_evaluator_launcher_internal` package). When present, the matching one pre-fills the cluster's `hostname` / `partition` / `gres` (and node-exclusivity), so the user only sets `account` / `output_dir` / `walltime` instead of hand-entering the hostname.

- **`SKILL.md` Step 4** — adds a "check FIRST" step: run a discovery snippet that lists the available `cluster → hostname` pairs **from the installed package at runtime**; on a hostname match, use `defaults: - execution: internal/slurm/<cluster>` (replacing `slurm/default`) and drop the now-redundant `execution.hostname`. If the package isn't installed or nothing matches, fall back to `slurm/default` and fill the fields manually.
- **`example_eval.yaml`** — a short, name-free comment on the `defaults` block pointing to that check.

**Discovery-based by design (no internal data committed):** cluster names, hostnames, and accounts are read from the install at runtime and are **not** hardcoded here. The only internal references in the repo are the package name `nemo_evaluator_launcher_internal` and the generic `internal/slurm/<cluster>` group pattern. External users without the package degrade gracefully (import fails → `slurm/default`).

### Usage

N/A — documentation / skill guidance only.

### Testing

Ran the discovery snippet verbatim (lists `cluster → hostname` from the installed package) and confirmed `internal/slurm/<cluster>` resolves via a `--dry-run`. `pre-commit run` passes (markdownlint + YAML format). Grep-verified no cluster names/hostnames are hardcoded in the changed files.

- Is this change backward compatible?: ✅ (falls back to `slurm/default`)
- Did you update Changelog?: N/A (skill docs)

## Summary by CodeRabbit

* **Documentation**
  * Added Step 4 guidance for discovering and using optional internal per-cluster SLURM execution configs: how to list available internal configs, select the matching cluster by switching the default execution, remove redundant hostname entries, and verify with a --dry-run.
  * Clarified that slurm/default is cluster-agnostic, recommend using internal per-cluster configs when present, and explained fallback behavior to manual hostname/account/output_dir entry.

### DIFF
--- a/.claude/skills/evaluation/SKILL.md
+++ b/.claude/skills/evaluation/SKILL.md
@@ -193,6 +193,17 @@ Reasoning models: prefer reasoning mode (highest scores). For lower variance / c
 
 ### Step 4 — Fill remaining ??? values
 
+**Predefined cluster execution config (check this FIRST — it removes several `???`s).** Some NEL installs ship ready-made per-cluster execution configs as an `internal/slurm/<cluster>` group (via the optional `nemo_evaluator_launcher_internal` package). When present, prefer the matching one over `slurm/default` — it pre-fills the cluster's `hostname` / `partition` / `gres` (and node-exclusivity), so the user only sets `account`, `output_dir`, and `walltime`. Discover what's available (names/hostnames come from the install at runtime — nothing cluster-specific is hardcoded here):
+
+```bash
+python3 -c 'import nemo_evaluator_launcher_internal' 2>/dev/null && \
+PKG=$(python3 -c 'import nemo_evaluator_launcher_internal as m,os;print(os.path.dirname(m.__file__))') && \
+for f in "$PKG"/configs/execution/internal/slurm/*.yaml; do \
+  echo "$(basename "$f" .yaml) -> $(grep -E '^hostname:' "$f" | awk '{print $2}')"; done
+```
+
+If a listed hostname matches the target SLURM cluster, set `defaults: - execution: internal/slurm/<cluster>` (replacing `slurm/default`), and remove any now-redundant `execution.hostname` (keep `account` / `output_dir` / `walltime` overrides). Confirm it resolves with a `--dry-run`. If the package isn't installed or nothing matches, keep `slurm/default` and fill `hostname` / `account` / `output_dir` manually.
+
 - Find every `???` left. Ask the user only for what can't be inferred (SLURM hostname/account/output_dir, MLflow tracking URI, etc.). Don't propose defaults; let them give plain text.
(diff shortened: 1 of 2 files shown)


meenchen (approved):

Docs-only change (+13/-0) adding a "check FIRST" discovery snippet for optional `nemo_evaluator_launcher_internal` per-cluster execution configs, plus a pointer comment in `example_eval.yaml`. No cluster-specific data is hardcoded — discovery is at runtime and falls back gracefully to `slurm/default` when the package isn't installed. No correctness, licensing, or test concerns.

.claude/skills/evaluation/SKILL.md
```diff
@@ -193,6 +193,17 @@ Reasoning models: prefer reasoning mode (highest scores). For lower variance / c
...
 ### Step 4 — Fill remaining ??? values
 
+**Predefined per-cluster execution config (check FIRST).** Some installs ship `internal/slurm/<cluster>` execution groups (optional `nemo_evaluator_launcher_internal` pkg) that pre-fill hostname/partition/gres — leaving only account/output_dir/walltime. Discover at runtime (nothing cluster-specific hardcoded):
+
+```bash
+python3 -c 'import nemo_evaluator_launcher_internal' 2>/dev/null && \
+PKG=$(python3 -c 'import nemo_evaluator_launcher_internal as m,os;print(os.path.dirname(m.__file__))') && \
+for f in "$PKG"/configs/execution/internal/slurm/*.yaml; do \
+  echo "$(basename "$f" .yaml) -> $(grep -E '^hostname:' "$f" | awk '{print $2}')"; done
+```
```
Edwardf0t1, line 203: This discovery snippet seems brittle, the agent should be able to help you do a double check. Also, do we want to disclose `internal` packages in public skills?

My agent's feedback:

  - It filesystem-globs "$PKG"/configs/execution/internal/slurm/*.yaml and greps ^hostname:. That assumes a fixed on-disk layout and column-0, unquoted hostname:. If the package stores configs elsewhere, uses a namespace package (__file__ can be None → the os.path.dirname line crashes), or templates/indents hostname, the loop silently prints nothing or blank hostnames — which is indistinguishable from "no configs." Hydra resolves internal/slurm/<cluster> via its search path, which isn't guaranteed to equal dirname(__file__)/configs/..., so globbing files is a proxy for "what Hydra can actually load." Preferring a Hydra-native enumeration (or nel-provided listing) would be more robust than globbing.
  - Make failure modes explicit for a deterministic agent: today import … 2>/dev/null && … produces no output when the package is missing or when the glob matches nothing. Add || echo "no internal configs — use slurm/default" and an empty-match branch, so the agent can
  tell the three states apart.

chadvoegele (approved):

PR #1600, opened 2026-06-02
Add Quantization Aware Distillation (QAD) to Megatron-Bridge example

kevalmorabia97:
### What does this PR do?

Type of change: new example

**Note:** This is **part 2 of 4** (builds on #1589):

- **Part 1 (#1589):** Megatron-Bridge `quantize.py` + `export.py` support and tests.
- **Part 2 (this PR):** extend `distill.py` for quantization-aware distillation (QAD) — load a quantized Megatron checkpoint as the student.
- **Part 3:** https://github.com/NVIDIA/Model-Optimizer/pull/1601
- **Part 4:** repeat the NVFP4 + QAD experiments on a non-Nemotron model.

Extends `examples/megatron_bridge/distill.py` to initialize the student from a **Megatron checkpoint** (a quantized checkpoint from `quantize.py`, or a pruned one) via `--student_megatron_path`, enabling **Quantization Aware Distillation (QAD)**:

- `--student_hf_path` still builds the student architecture; `--student_megatron_path` supplies the (optionally quantized) weights.
- For a quantized checkpoint, the ModelOpt quantize mode + base weights are restored onto the **plain student before the knowledge-distillation conversion** (`restore_sharded_modelopt_state` is a no-op once a model is already converted), so the distilled checkpoint stays exportable as a quantized model with `export.py`.

**Upstream dependency / workaround:** `DistillationProvider.provide()` has no seam to transform the student before the KD conversion, so this patches `provide()` at the class level (via an `id()`-keyed registry, because the provider proxies instance-attribute assignment to its teacher once the teacher is set). A companion Megatron-Bridge PR adds a first-class `DistillationProvider.student_pre_conversion_hook`; from nemo:26.06 onwards the workaround should be removed and replaced with that hook (a removal note in `distill.py` documents exactly how).

### Usage

```bash
# 1) PTQ -> quantized Megatron checkpoint (part 1)
torchrun --nproc_per_node 2 quantize.py \
    --hf_model_name_or_path Qwen/Qwen3-8B --quant_cfg fp8 --tp_size 2 \
    --export_megatron_path /tmp/Qwen3-8B-FP8-megatron

# 2) QAD: distill the quantized student from the unquantized teacher
torchrun --nproc_per_node 8 distill.py \
    --teacher_hf_path Qwen/Qwen3-8B \
    --student_hf_path Qwen/Qwen3-8B \
    --student_megatron_path /tmp/Qwen3-8B-FP8-megatron \
    --data_paths 1.0 tokenized/data_text_document \
    --train_iters 1000 --output_dir /output/qwen3_8b_qad

# 3) export the distilled quantized checkpoint (part 1)
torchrun --nproc_per_node 1 export.py \
    --hf_model_name_or_path Qwen/Qwen3-8B \
    --megatron_path /output/qwen3_8b_qad/checkpoints \
    --export_unified_hf_path /tmp/qwen3_8b_qad_fp8_hf
```

### Testing

`tests/examples/megatron_bridge/test_qad.py` (validated on a 2-GPU NeMo `26.04` container): quantize a tiny Qwen3 at TP=2 → QAD distill from the quantized student → `export.py` to a unified HF checkpoint, asserting `hf_quant_config.json` is written (proves the quantize mode survived QAD). Includes a commented-out vLLM deployment check, validated locally (full flow passes; vLLM loads the export as `quantization=modelopt`). Existing normal/Puzzletron distillation tests still pass.

- Is this change backward compatible?: N/A (new example feature; default behavior unchanged when `--student_megatron_path` is not set)
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A (no new dependencies)

### Additional Information

Depends on a companion Megatron-Bridge PR adding `DistillationProvider.student_pre_conversion_hook` (the upstream replacement for the class-level `provide()` workaround). The Nemotron-3 tutorial NVFP4 + QAD experiments ship in part 3.

## Summary by CodeRabbit

* **New Features**
  * Quantization Aware Distillation (QAD) workflow to recover accuracy of quantized Megatron students and distill from quantized checkpoints.
  * CLI option to initialize a distillation student from a Megatron checkpoint and a structure-only load path for bridging.

* **Documentation**
  * Expanded runnable quantize → QAD → export guidance and best-practice tips.

* **Tests**
  * End-to-end test validating quantize → QAD → export artifacts.

* **Chores / UX**
  * Clearer rank-aware messages, improved tokenizer padding handling, and more consistent export behavior (fixed export dtype).


AAnoosheh (approved):

examples/megatron_bridge/distill.py
```diff
@@ -34,6 +35,10 @@
 from megatron.bridge.recipes.utils.optimizer_utils import (
     distributed_fused_adam_with_cosine_annealing,
 )
+
+# _load_model_weights_from_checkpoint is the (private) helper bridge.load_megatron_model uses to load
+# a (quantized) Megatron checkpoint into an already-built model; reused here to initialize the student.
+from megatron.bridge.training.checkpointing import _load_model_weights_from_checkpoint
```
jenchen13: why can't you just use `bridge.load_megatron_model`?

examples/megatron_bridge/distill.py
```diff
@@ -90,6 +101,102 @@ def _patched_to_cfg_dict(self):
...
+    ckpt_root = Path(megatron_path)
+    iter_dirs = [d for d in ckpt_root.iterdir() if d.is_dir() and d.name.startswith("iter_")]
+    ckpt_dir = (
+        max(iter_dirs, key=lambda d: int(d.name.removeprefix("iter_"))) if iter_dirs else ckpt_root
+    )
+
+    if restore_modelopt_state:
+        load_modelopt_state([student_model], str(ckpt_root))
+    print_rank_0(f"Loading student weights from Megatron checkpoint {ckpt_dir}")
+    # strict=False because the bridge loader strips Transformer-Engine extra-state from the loaded
```
jenchen13: if the mbridge loader strips extra-state from the sharded checkpoint, how are the amax values restored?

examples/megatron_bridge/distill.py
kevalmorabia97: Because of some checks in it which doesnt cause load_modelopt_state to be invoked at the righ time in case of quantized ckpt, we need a temporary workaround. @AAnoosheh is looking into the correct fix in megatron bridge side. Alternatively as we discussed in Nemo-ModelOpt meeting, ideally we will have mbridge natively support quantized ckpt then we wont need this workaround

examples/megatron_bridge/distill.py
kevalmorabia97: Its restored from `load_modelopt_state` call above

examples/megatron_bridge/distill.py
```diff
@@ -60,7 +64,7 @@
 def _patched_to_cfg_dict(self):
     """Patched DistillationProvider.to_cfg_dict method for heterogeneous teacher and student models.
 
-    TODO: Upstream this patch to Megatron-Bridge.
+    Can be removed from nemo:26.02.01 container onwards.
```
jenchen13: does this need a TODO?

examples/megatron_bridge/distill.py
```diff
@@ -90,6 +94,71 @@ def _patched_to_cfg_dict(self):
 DistillationProvider.to_cfg_dict = _patched_to_cfg_dict
 
 
+# Megatron-Bridge does not (yet) expose a hook to initialize the student before the
+# knowledge-distillation conversion, so we patch ``DistillationProvider.provide`` to do it. Replace
```
jenchen13, line 98: also add TODO here to replace

examples/megatron_bridge/export.py
```diff
@@ -69,20 +71,25 @@ def get_args() -> argparse.Namespace:
...
-        default="bfloat16",
-        choices=list(_DTYPE_MAP),
-        help="Data type for the exported weights.",
+        "--num_layers_in_first_pipeline_stage",
+        type=int,
+        default=None,
+        help="Number of layers in the first pipeline stage (Uneven Pipeline Parallelism)",
     )
     parser.add_argument(
-        "--export_extra_modules",
```
jenchen13, line 83: keep `export_extra_modules` ? for spec dec exports

examples/megatron_bridge/quantize.py
```diff
@@ -28,12 +28,14 @@
 sharded and can be re-sharded on load (e.g. `export.py` reloads it at TP=1 for the HF export).
 
 Example usage to quantize Qwen3-8B to FP8 on 2 GPUs (Tensor Parallelism = 2):
-    1024 samples from nemotron-post-training-dataset-v2 are used for calibration.
```
jenchen13, line 31: explain cnn_nemotron_v2_mix is default dataset?

examples/megatron_bridge/quantize.py
```diff
@@ -155,18 +161,21 @@ def get_args() -> argparse.Namespace:
...
+            f"HF Dataset name or local path for calibration (supported options: {', '.join(get_supported_datasets())}. "
+            "You can also pass any other dataset and see if auto-detection for your dataset works."
+        ),
     )
     parser.add_argument(
         "--calib_num_samples", type=int, default=1024, help="Number of samples for calibration"
     )
     parser.add_argument("--calib_batch_size", type=int, default=1, help="Calibration batch size")
-    parser.add_argument("--seq_length", type=int, default=4096, help="Calibration sequence length")
+    parser.add_argument("--seq_length", type=int, default=512, help="Calibration sequence length")
```
jenchen13: why is calib seq length only 512?

examples/megatron_bridge/README.md
```diff
@@ -65,31 +65,37 @@ This section shows how to quantize a HuggingFace model using ModelOpt in the Meg
 
 `quantize.py` supports the following formats via `--quant_cfg` (e.g. `fp8`, `nvfp4`, `int8_sq`, `int4_awq`, `w4a8_awq`, ...). You can also pass any full config name exposed by ModelOpt (e.g. `FP8_DEFAULT_CFG`) or a YAML `--recipe` (e.g. `general/ptq/fp8_default-kv_fp8`, authoritative for quant_cfg + algorithm + KV-cache). KV-cache quantization can be enabled on top via `--kv_cache_quant` (e.g. `fp8`, `nvfp4`).
 
-**Step 1 — quantize** Qwen3-8B to FP8 on 2 GPUs (Tensor Parallelism = 2) using 1024 samples from [`nemotron-post-training-dataset-v2`](https://huggingface.co/datasets/nvidia/Nemotron-Post-Training-Dataset-v2) for calibration:
+**Step 1 — quantize** Qwen3-8B to FP8 on 2 GPUs (Tensor Parallelism = 2) using 1024 samples from default dataset for calibration (sequence length = 512):
```
jenchen13: add details for what default dataset is

examples/megatron_bridge/README.md
```diff
@@ -65,31 +65,37 @@ This section shows how to quantize a HuggingFace model using ModelOpt in the Meg
...
 
 ```bash
 torchrun --nproc_per_node 2 quantize.py \
     --hf_model_name_or_path Qwen/Qwen3-8B \
     --quant_cfg fp8 \
     --tp_size 2 \
+    --calib_batch_size 16 \
+    --seq_length 512 \
```
jenchen13: same comment here 512 is pretty low

examples/megatron_bridge/export.py
kevalmorabia97: reordered

examples/megatron_bridge/quantize.py
kevalmorabia97: Matched `hf_ptq` defaults

examples/megatron_bridge/README.md
kevalmorabia97: @cjluo-nv @realAsma is cnn_nemotron_v2_mix dataset, 512 seq len, 1024 samples default in hf_ptq best in practice? Or do we use different e.g. for MoE models?

cjluo-nv (commented):

The QAD addition itself is small and well-scoped — `--student_megatron_path` plus a documented, time-boxed monkey-patch on `DistillationProvider.provide` (with a clear "remove once upstream `student_pre_conversion_hook` lands" note) and an end-to-end test (`test_qad.py`) that validates the full quantize → QAD → export flow including `hf_quant_config.json`. The design question ("why a custom patch instead of an existing seam?") is explicitly addressed in the PR body — there is no upstream hook today, and an instance attribute would leak onto the teacher via DistillationProvider's attribute proxying, so the `id()`-keyed registry is justified. A few items worth a human eyeballing before merge:

1. **Stale comment vs. behavior change in `quantize.py`** — the `try/finally` that reset `unwrapped_model.calibration_mode = False` on exception was dropped, but the surrounding comment still says "Reset it in a finally so a failure mid-calibration doesn't leave the flag set for the subsequent compress / save calls." Either restore the `finally` or update the comment; right now they contradict each other and a calibration crash will leave `calibration_mode=True` for any later code paths.

2. **Silent CLI / behavior change in `export.py`** — `--dtype` (with `bfloat16`/`float16`/`float32` choices) was removed and the export dtype is hard-coded to `torch.bfloat16`. This narrows the public CLI of the example and isn't called out in the PR description (the CodeRabbit summary mentions it only as "more consistent export behavior (fixed export dtype)"). Worth confirming this is intentional and that no downstream tutorial/test passes `--dtype`.

3. **Re-implementation drift risk in `_distill_provide_with_megatron_student`** — the patched `provide()` re-does the upstream KD setup (build student → fetch teacher via `provide_distributed_model` → `setup_distillation_config` → `mtd.convert(..., mode=[("kd_loss", ...)])` → `adjust_distillation_model_for_mcore`). If upstream `DistillationProvider.provide` evolves (e.g. extra setup steps, different `mtd.convert` args), this fork will silently diverge only on the QAD path. The PR commits to ripping it out at nemo:26.06; a CI guard or version assertion that fails loudly when running against a newer mbridge would make that bound enforceable rather than aspirational.

4. **Stacked-PR scope** — most of the +366/-121 diff (quantize.py polish, export.py dtype/uneven-PP changes, `load_mbridge_model_from_hf` `load_weights` plumbing, dataset_utils `warn`→`warn_rank_0` sweep, `print_args` sort) carries over from the stacked base #1589. Reviewer should confirm #1589 is landing first, otherwise the actual reviewable surface here (distill.py, `load_modelopt_megatron_checkpoint`, README QAD section, test_qad.py) is much smaller and items 1–2 belong on #1589 instead.

5. **Coverage gap** — `test_qad.py` exercises the happy path; the defensive `RuntimeError("...the provider was likely copied/wrapped...")` branch in the patched `provide()` is unreachable from the test. Not a blocker, just noting that if a future change wraps the provider, the failure will only show up in real runs.

cjluo-nv (commented):

QAD addition itself is small, well-scoped, and previous critical feedback is mostly addressed. The class-level `DistillationProvider.provide` monkey-patch is time-boxed against the upstream `student_pre_conversion_hook` (26.06), the id()-keyed registry is now defensive (raises if a registered id doesn't match instead of silently falling back), and the inlined load path has dropped the `has_modelopt_state` parameter shadow. The end-to-end test (`test_qad.py`) covers quantize → QAD → export and asserts `hf_quant_config.json` is written. A few items still warrant a human look before merge:

- :speech_balloon: cjluo-nv flagged the silent removal of `--dtype` (with `bfloat16`/`float16`/`float32` choices) from `export.py` — the script now hard-codes `torch.bfloat16` for both `pipeline_dtype` and the `export_mcore_gpt_to_hf(dtype=...)` call. The PR body still doesn't mention this CLI narrowing (only the CodeRabbit summary alludes to "more consistent export behavior"). Worth a confirmation that no downstream tutorial/test passes `--dtype` and that float16/float32 export was deliberately dropped.
- :speech_balloon: cjluo-nv also flagged the stacked-PR scope: most of the +367/−124 (quantize.py polish, export.py uneven-PP/dtype changes, mbridge `load_weights` plumbing, dataset_utils `warn`→`warn_rank_0` sweep, `print_args` sort, `gradient_accumulation_fusion=False` move, `gc.collect()` after quantize) carries over from the stacked base #1589. Reviewer should confirm #1589 lands first; otherwise, several review concerns (e.g. the `--dtype` removal, the dropped `try/finally` around `calibration_mode`) actually belong on #1589.
- The patched `provide()` re-implements the upstream KD setup (`provide_distributed_model` → `setup_distillation_config` → `mtd.convert(..., mode=[("kd_loss", ...)])` → `adjust_distillation_model_for_mcore`). If upstream `DistillationProvider.provide` evolves before the 26.06 hook lands, the QAD path will silently diverge from the vanilla path. A version pin / assertion against mbridge would make the "remove at 26.06" bound enforceable rather than aspirational.
- Coverage gap: the defensive `RuntimeError("...the provider was likely copied/wrapped...")` branch in the patched `provide()` is unreachable from `test_qad.py`. Not a blocker, just noting.

Design review: the PR body explicitly justifies the class-level monkey-patch (no upstream seam today; instance-attribute approach leaks onto teacher via DistillationProvider's attribute proxying), so the design question is settled.

examples/megatron_bridge/quantize.py
realAsma: Can we have the 4096 seq length? hf_ptq does not have a performant backend. 4K might go OOM. MBRidge does not have this limitation. I would recommend either 2048 or 4096.

examples/megatron_bridge/README.md
realAsma: Can we have the 4096 seq length? hf_ptq does not have a performant backend. 4K might go OOM on hf_ptq. MBRidge does not have this limitation. I would recommend either 2048 or 4096.

examples/megatron_bridge/distill.py
realAsma, line 1: Should we name this file train.py and support both QAT and QAD similar to HF  llm_qat folder?

realAsma:
Can we add a link to this from https://github.com/NVIDIA/Model-Optimizer/tree/main/examples/llm_qat as performant QAT/QAD backend?

examples/megatron_bridge/distill.py
kevalmorabia97: My understanding is we dont have a good story for using QAT and instead recommend QAD. But if needed, I can change it in a follow-up PR to also support QAD

examples/megatron_bridge/README.md
kevalmorabia97: Most sequences in cnn_dailymail and nemotron-pt-v2 are under 1-2k length. Should I use the new sequence packing option as well so we see real tokens for the 4k seq len? Then we can reduce num sample default to 512 instead of 1024 as well.

Could also be done as a follow-up if we want to do more experiments before changing defaults

examples/megatron_bridge/README.md
```diff
@@ -65,31 +69,45 @@ This section shows how to quantize a HuggingFace model using ModelOpt in the Meg
 
 `quantize.py` supports the following formats via `--quant_cfg` (e.g. `fp8`, `nvfp4`, `int8_sq`, `int4_awq`, `w4a8_awq`, ...). You can also pass any full config name exposed by ModelOpt (e.g. `FP8_DEFAULT_CFG`) or a YAML `--recipe` (e.g. `general/ptq/fp8_default-kv_fp8`, authoritative for quant_cfg + algorithm + KV-cache). KV-cache quantization can be enabled on top via `--kv_cache_quant` (e.g. `fp8`, `nvfp4`).
 
-**Step 1 — quantize** Qwen3-8B to FP8 on 2 GPUs (Tensor Parallelism = 2) using 1024 samples from [`nemotron-post-training-dataset-v2`](https://huggingface.co/datasets/nvidia/Nemotron-Post-Training-Dataset-v2) for calibration:
+**Step 1 — quantize** Qwen3-8B to FP8 on 2 GPUs (Tensor Parallelism = 2) using 1024 samples from default dataset (Mix of [`cnn_dailymail`](https://huggingface.co/datasets/abisee/cnn_dailymail) and [`nemotron-post-training-dataset-v2`](https://huggingface.co/datasets/nvidia/Nemotron-Post-Training-Dataset-v2)) for calibration (sequence length = 16):
```
meenchen: Should we use nvfp4 as an example? We usually don't QAD FP8 checkpoints.

meenchen (approved):

examples/megatron_bridge/generate_vllm.py
```diff
@@ -0,0 +1,71 @@
...
+# Disable the FlashInfer MoE kernels (FP8/NVFP4), which JIT-compile and autotune on the first
+# run and can hang or error. Falls back to a stable path -- ideal for a quick sanity check.
+# Set these before importing vllm; `setdefault` lets a shell-provided value override.
+os.environ.setdefault("VLLM_USE_FLASHINFER_MOE_FP8", "0")
+os.environ.setdefault("VLLM_USE_FLASHINFER_MOE_FP4", "0")
+
+import vllm
+
+DEFAULT_PROMPTS = [
+    "Hello!",
```
jenchen13, line 35: this can be addressed in later PR but: these prompts are rather short, would be good to have some more long generation/agentic task

examples/megatron_bridge/quantize.py
jenchen13: yea we need longer seq length than 512, like 2048 or 4096. and also suggest a longer seq length for higher quality calibration

examples/megatron_bridge/README.md
jenchen13: are examples in nemotron-post-training really under 2k seq length? nemotron is always post trained with long seq length like 256k

ChenhanYu (approved):

examples/megatron_bridge/README.md
kevalmorabia97: Set default seq len 4K for now

jenchen13 (approved):
LGTM!

PR #1607, opened 2026-06-02
feat: EAGLE3 LoRA co-training improvements

yeyu-nvidia:
## Summary
- Add `eagle_base_lora_start_layer` config for layer-restricted LoRA injection (e.g., apply LoRA only to layers 17+ for last-half experiments)
- Use explicit module name matching for reliable layer filtering when LoRA is injected into the inner base model (PEFT's `layers_to_transform` doesn't work in this context)
- Fix optimizer state mismatch on checkpoint resume: always include LoRA params in the optimizer from init rather than dynamically adding via `add_param_group` at warmup step, which breaks checkpoint compatibility
- Add `eval_lora.sh` script for export + LoRA merge + lm_eval evaluation pipeline

## Test plan
- [x] All-layers MLP LoRA co-training on Qwen3-8B (checkpoint-410k, epoch 1.67)
- [x] Last-half MLP LoRA co-training on Qwen3-8B (checkpoint-220k, epoch 0.90)
- [x] Verified checkpoint resume works correctly after optimizer fix
- [x] Eval pipeline (export + merge + lm_eval + vLLM AR benchmark) passes on both variants

## Summary by CodeRabbit

* **New Features**
  * Config option to restrict which base-model layers receive LoRA adaptation.
  * New evaluation script for LoRA-adapted EAGLE3 models with optional baseline comparison.

* **Changes**
  * LoRA parameters are now always trainable after injection.
  * Warmup flow now activates LoRA co-training when available.
  * Model-merge tool adds a CLI flag to control trusting remote model code.

### DIFF
--- a/examples/speculative_decoding/eagle_utils.py
+++ b/examples/speculative_decoding/eagle_utils.py
@@ -191,29 +191,6 @@ class LoRAWarmupCallback(TrainerCallback):
             raw_model = model.module if hasattr(model, "module") else model
             if hasattr(raw_model, "_lora_cotraining_active"):
                 raw_model._lora_cotraining_active = True
-                # Unfreeze LoRA parameters
-                lora_params = []
-                for name, param in raw_model._base_model.named_parameters():
-                    if "lora_" in name:
-                        param.requires_grad = True
-                        lora_params.append(param)
-
-                # Add LoRA params to optimizer — they were excluded at creation time
-                # because requires_grad was False during warmup.
-                optimizer = kwargs.get("optimizer")
-                if optimizer is not None and lora_params:
-                    existing_ids = {id(p) for g in optimizer.param_groups for p in g["params"]}
-                    new_params = [p for p in lora_params if id(p) not in existing_ids]
-                    if new_params:
-                        optimizer.add_param_group(
-                            {
-                                "params": new_params,
-                                "lr": optimizer.param_groups[0]["lr"],
-                                "weight_decay": 0.0,
-                            }
-                        )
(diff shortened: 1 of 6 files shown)


h-guo18 (approved):
Not familiar to lora cotraining, but this PR seems non-breaking for existing functionalities. Approve to unblock

yeyu-nvidia:
## On-cluster validation

The new `eagle_base_lora_start_layer` validation logic was tested on a GPU node (OCI-HSG, 4×GPU, TRT-LLM 1.3 container) in addition to the GitHub unit suite:

```
tests/unit/torch/speculative/plugins/test_hf_speculative_lora.py
8 passed, 4 warnings in 29.78s
```

This covers the 5 pre-existing LoRA co-training tests plus the 3 added in this PR:
- `test_start_layer_restricts_injection` — LoRA adapters land only on layers ≥ `start_layer`
- `test_start_layer_requires_explicit_target_modules` — `ValueError` when `start_layer` is set without explicit `target_modules`
- `test_start_layer_too_high_raises` — `ValueError` when the start-layer filter matches no modules (prevents the silent zero-adapter no-op flagged in review)

### Review feedback addressed
- **Security (blocking pre-merge check):** `trust_remote_code` is now caller-controlled via `TRUST_REMOTE_CODE` (default `False`) in `eval_lora.sh`, and a `--trust_remote_code` flag (default off) in `merge_lora.py`.
- **Correctness (Claude IMPORTANT):** layer-selective injection now raises a clear `ValueError` instead of silently injecting zero adapters.
- **Nitpick:** `eval_lora.sh` validates `HF_MODEL_CKPT` / `EAGLE_CKPT` up front.

PR #1609, opened 2026-06-02
Nemotron Ultra & Super launcher examples

jenchen13:
### What does this PR do?

Type of change: New example

New launcher example for Nemotron Super with PTQ + Export + VLLM smoke test on small GPQA-style dataset

### Usage

```python
# Usage:
#   source .env-slurm
#   cd tools/launcher
#   uv run launch.py --yaml examples/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16/megatron_lm_ptq.yaml --yes
```

## Summary by CodeRabbit

* **New Features**
  * Added checkpoint export capability for quantized models to Hugging Face format.
  * Introduced complete quantization pipelines with conditional MMLU evaluation and model export stages.

* **Bug Fixes**
  * Fixed num_shards calculation to prevent invalid minimum values.

* **Documentation**
  * Updated vLLM version requirements for optimal NVFP4 model performance.
  * Enhanced quantization pipeline documentation with improved output paths and conditional execution details.

* **Chores**
  * Updated Megatron-LM module to latest version.
  * Added sample dataset for model evaluation testing.


ChenhanYu (commented):
Reviewed the full diff plus `query.sh`/`quantize.sh` context. Overall a clean, well-scoped example — factoring export out of `quantize.sh` into a reusable `export.sh` with `RUN_MMLU`/`RUN_EXPORT` toggles is the right call for a 120B model that needs different parallelism per stage. Inter-task path handoff (`/cicd/megatron-lm/...`), the task_2 `--`-separator arg wiring, and the `EXPORT_DIR` tag that task_2 reads all check out. The `query.py` `max(1, ...)` fix is correct and necessary — with the 8-row smoke set and `--num-shards 1`, the old `min(16, 8//100)` collapsed to 0 shards.

A few things to address before merge:

**1. (also flagged by CodeRabbit) `export.sh` header defaults are wrong.** Lines 36-37 document `/scratchspace/...` defaults, but the code uses `/cicd/...` (lines 41 and 49). Update the comment to match.

**2. `EXPORT_DIR` diverges between the two scripts.** `export.sh` writes to `/cicd/export/...` while `quantize.sh` still writes to `/scratchspace/export/...`, and the QUANT_CFG->tag logic differs (`export.sh` strips `.yaml`/`.yml`, `quantize.sh` does not). In this pipeline `RUN_EXPORT=false` so quantize's export path is dead, but a future caller running `quantize.sh` with inline export would get a different dir and tag than the standalone path. Worth unifying the tag logic and the `/cicd` vs `/scratchspace` choice so the two paths can't drift.

**3. MMLU `--fraction` bumped 0.01 -> 0.05 in the shared `quantize.sh`.** This is a 5x longer MMLU eval for every example that still runs MMLU inline, not just Nemotron, and it isn't mentioned in the PR description. Please confirm it's intentional for all callers; if it's only meant for this model it shouldn't be in the shared default.

**Minor / process:**
- The smoke test only validates that the model serves and emits text — responses are dumped to `/cicd/vllm/...jsonl` and never graded, so the `gpqa_smoke.jsonl` answer keys aren't checked by anything. Fine for a smoke test; just won't catch accuracy regressions.
- PR template checkboxes are unfilled (notably tests + Changelog) and the Testing section is empty — a line on how this was validated would help, since CONTRIBUTING marks tests mandatory for new examples.
- A one-liner on what the Megatron-LM submodule bump (`86bf476` -> `c69697d`) pulls in would help reviewers.

Items 1-3 are the blockers; the rest is polish. Nice work.

tools/launcher/common/megatron_lm/export/export.sh
kevalmorabia97, line 1: General question on the design of all scripts in this directory. Why do we need yet another export/quantize.sh on top of M-LM's export/quantize.sh?

jenchen13: I think this calls the scripts in Megatron-LM which are under `modules`? that's a good question though why can't we just call the scripts in `modules/Megatron-LM` instead of wrapping them again? @ChenhanYu do you know

kevalmorabia97 (approved):
Approving to unblock

tools/launcher/common/megatron_lm/export/export.sh
jenchen13: we can address this in a future PR, thanks!

PR #1610, opened 2026-06-02
[Eval skill] Add reasoning adapter_config to eval template

cjluo-nv:
### What does this PR do?

Type of change: documentation

Adds a reasoning-aware adapter (interceptor) config to the `evaluation` skill's
default eval template and documents how to set `use_reasoning` per model type.

- **`recipes/examples/example_eval.yaml`** — adds an `adapter_config` block under
  `evaluation.nemo_evaluator_config.target.api_endpoint` with request/response
  logging, `use_reasoning`, `log_failed_requests`, and a
  `params_to_add.chat_template_kwargs` thinking block, with inline comments on
  which lines to flip for instruct vs. reasoning vs. hybrid models.
- **`SKILL.md`** — new "Reasoning adapter config (`use_reasoning`)" subsection at
  the end of Step 3:
  - Instruct (non-reasoning) models → `use_reasoning: false` (no CoT trace to
    strip; drop the `chat_template_kwargs` thinking block).
  - Reasoning models → `use_reasoning: true`, especially when the deployment
    command sets `--reasoning-parser`.
  - Hybrid models (can run reasoning on or off) → always force thinking on via
    `chat_template_kwargs` and keep `use_reasoning: true` (highest scores).

### Usage

```bash
nel run --config recipes/examples/example_eval.yaml \
  -o deployment.checkpoint_path=/path/to/checkpoint \
  -o deployment.served_model_name=my-model
```

### Testing

Docs/template-only change. Ran `pre-commit` on both files (yamlfmt + markdownlint
pass).

### Additional Information

Note: there is a pre-existing inconsistency between Step 7 (which references
`nemo_evaluator_config.config.target.api_endpoint.adapter_config`) and the
template's `nemo_evaluator_config.target.api_endpoint` placement. This PR follows
the template's existing path; reconciling Step 7 is left as a follow-up.

## Summary by CodeRabbit

* **Documentation**
  * Added a guide for configuring adapter reasoning/chain-of-thought behavior by model type (instruct, reasoning, hybrid), how reasoning traces can be stripped before scoring, model-family template keyword variants, and guidance to verify template kwargs and select appropriate reasoning-effort settings.

* **Chores**
  * Added an example configuration demonstrating enabling the reasoning adapter with caching, limited request/response logging, failed-request logging, forced “thinking” flags for hybrid models, and a reasoning-effort option.


.claude/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -84,6 +84,32 @@ evaluation:
...
+          params_to_add:
+            # Forces thinking on for hybrid models that can run both modes.
+            # Drop this block for pure-instruct models. The thinking-toggle key
+            # is family-specific — keep only the one your model uses:
+            #   enable_thinking -> Qwen3.5/3.6, GLM 5.1
+            #   thinking        -> Kimi K2.6, DeepSeek V3.2/V4
+            # (DeepSeek models use a Python encoder, not Jinja, so an unused key
+            # can error rather than be ignored — don't emit both.)
+            chat_template_kwargs:
+              thinking: true
```
meenchen, line 108: > _Bot comment._

The example ships both `thinking: true` and `enable_thinking: true` simultaneously, but the comment block immediately above says "keep only the one your model uses" and "don't emit both" (DeepSeek's Python encoder can error on a stray kwarg). A user who runs the template unedited gets a config the docs explicitly warn against. Consider commenting out one of them by default (e.g. leave `enable_thinking: true` active and `# thinking: true` commented, since the docs note enable_thinking covers the Qwen3.x / GLM 5.1 case), so the default state matches the safer guidance.

.claude/skills/evaluation/SKILL.md
```diff
@@ -189,6 +189,27 @@ For quantized checkpoints, read `references/quantization-benchmarks.md` for sens
 
 Reasoning models: prefer reasoning mode (highest scores). For lower variance / cost / apples-to-apples vs non-reasoning baselines, also consider a non-reasoning companion run.
 
+#### Reasoning adapter config (`use_reasoning`)
```
meenchen, line 192: > _Bot comment._

Step 7 (around line 218 in the unchanged section) currently says interceptors live at `nemo_evaluator_config.config.target.api_endpoint.adapter_config` (NOT under `target.api_endpoint.adapter_config`) — but this PR's example, the existing `recipes/tasks/aa/lcr.md` fragment, and `examples/pruning/minitron/.../nemo_evaluator.yaml` all use the latter (`target` as a sibling of `config`). The PR body acknowledges this and defers the fix; consider just correcting Step 7 in this same PR — it's two lines and avoids leaving the skill self-contradictory.

meenchen (commented):

Small docs/yaml-only addition (+47/-0) of an `adapter_config` block to the eval skill's example template plus a "Reasoning adapter config" subsection in SKILL.md. The placement matches existing practice in the repo (Nemotron pruning eval yaml and `aa/lcr.md` task fragment both use `nemo_evaluator_config.target.api_endpoint.adapter_config`), so the PR is internally consistent with the template — Step 7's stated `.config.target...` path is the actually-incorrect one, and the PR explicitly leaves reconciling it as a follow-up. Two things to flag for the owner: the shipped yaml emits both `thinking: true` and `enable_thinking: true` despite its own warning to "keep only one" (footgun if a user copies verbatim), and the Step 7 contradiction probably should be fixed here rather than punted.

.claude/skills/evaluation/SKILL.md
```diff
@@ -189,6 +189,27 @@ For quantized checkpoints, read `references/quantization-benchmarks.md` for sens
...
+
+The `adapter_config` block in `example_eval.yaml` controls request/response
+logging and reasoning handling. `use_reasoning: true` strips the model's
+reasoning/CoT trace before scoring (grade only the final answer). Set per type:
+
+1. **Instruct → `use_reasoning: false`** and drop the `chat_template_kwargs`
```
chadvoegele, line 198: Can you consolidate with https://github.com/NVIDIA/Model-Optimizer/blob/88fd7ff95816975dd3edd1cf830584887572f7eb/.claude/skills/evaluation/SKILL.md?plain=1#L194 & https://github.com/NVIDIA/Model-Optimizer/blob/88fd7ff95816975dd3edd1cf830584887572f7eb/.claude/skills/evaluation/references/model-card-research.md?plain=1#L18

.claude/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -84,6 +84,32 @@ evaluation:
     target:
       api_endpoint:
         api_key_name: DUMMY_API_KEY
+        # Adapter (interceptor) config. `use_reasoning` strips reasoning/CoT
+        # traces before scoring — see SKILL.md "Reasoning adapter config" for
+        # when to enable it (reasoning models only; instruct models -> false)
+        # and how to force thinking on via chat_template_kwargs.
```
meenchen, line 90: Do you know how this interacts with reasoning parsers of the inference runtime? like `--reasoning-parser` in vllm

cjluo-nv: it does not. Though they should be in sync.

.claude/skills/evaluation/recipes/examples/example_eval.yaml
cjluo-nv: this should be fine. chat template will only pick the actual one that works.

.claude/skills/evaluation/SKILL.md
cjluo-nv: Should be consolidated now.

chadvoegele (approved):

meenchen (approved):

PR #1616, opened 2026-06-03
Speed up slow unit/gpu/example tests

kevalmorabia97:
### What does this PR do?

Type of change: test infrastructure / test speedups + CI stabilization

Make the test suite faster, `tests/unit` hermetic, and the CI lanes stable, without losing coverage. Most changes are mechanical test/infra edits; the buckets below cover the diff broadly.

**Unit tests — hermetic (no HF Hub):** toy local datasets/configs + the local tiny tokenizer (with a checked-in chat template) replace Hub assets; `tests/unit/conftest.py` enforces offline mode. Genuinely-HF tests moved to `tests/gpu*` (e.g. the new `tests/gpu/torch/utils/test_dataset_utils.py`). `CONTRIBUTING.md` documents the hermetic-unit-test expectation.

**Unit-test speedups (no coverage loss):** speculative (disable CPU torch.compile), calibrator (fewer histogram bins), ONNX conv/dynamo (smaller shapes + representative subset), Ruler/sparse-attention (local tokenizer), data-parallel autoquant (world size 4→2). Shared `tiny_tokenizer` fixture. The distributed test helper now uses a private `spawn` context instead of mutating the global start method (avoids cross-test contamination).

**Rarely-used autonas/fastnas tests:** heavy parametrize cases marked `@pytest.mark.manual`, one representative kept per test (fastnas preferred); lighter sibling tests still cover core behavior. The legacy FSDP1 NAS distributed test is also dropped: FastNAS/AutoNAS aren't used with either FSDP1 or FSDP2, and FSDP1 is superseded by the newer FSDP2 API — so we keep a single FSDP2 case as a sanity check and drop FSDP1, leaving the suite leaner.

**gpu_megatron:** deduplicate distributed worker pools by world_size within a module (saves a redundant pool spin-up in multi-pool files; module-scoped, no cross-module reuse).

**Example tests:** reduce per-test work via args that default to current behavior (tests pass the fast values) — torch_onnx TRT optimization level, diffusers calibration/inference steps, eagle `sample_size`, megatron_bridge iters/calib, llm_sparsity data slice, export safetensors-structure `calib_size`. Also enable the recently added `gpt-oss` example tests in CI.

**Per-test timeouts:** `pytest-timeout` with a default per-directory timeout (60s unit / 300s gpu+example) enforced in `tests/conftest.py` (`timeout_func_only` in `pyproject.toml`), so a new test cannot silently exceed the budget — an unmapped test dir crashes collection. A few inherently slow tests carry explicit higher per-test overrides (CUDA-compile, autotune, dflash).

**CUDA kernel pre-compilation:** a dedicated `tests/gpu/_extensions` test JIT-builds the conv3d implicit-GEMM kernel up front (collected before the functional tests in the same process) so the one-time build cost no longer lands on — and time out — the first functional test that uses it. Mirrored into the `llm_ptq`/`vlm_ptq` example lanes.

**Test relocation & optional-dependency guards:** vLLM sparsity plugin test moved to `tests/gpu_vllm` (drops the in-test `importorskip`); diffusers-dependent unit test guarded with `importorskip("diffusers")` for partial-install lanes; `gpt_oss` example test dir renamed to `gpt-oss` to match the CI matrix.

**Diffusers test models:** shared model-path constants in `tests/_test_utils/examples/models.py` consolidated/renamed and point at tiny `hf-internal-testing` test pipes (SDXL/SD3/FLUX) so cachify/quantize/export tests run on toy weights; `local_id`s normalized.

**Shared dataset utils:** `examples/llm_sparsity/.../hf_pts.py` now uses `get_dataset_dataloader` (drops the bespoke cnn_dailymail-only `get_calib_dataloader`; supports any registered/HF/JSONL dataset, includes attention_mask); `data_prep.py` gains `--max_samples`.

**CI workflows:** container image bumps (pytorch 26.04→26.05, TRT-LLM rc16→rc17) and tightened lane timeouts (unit 30→15 min, gpu lanes trimmed, onnx example lane 45 min).

**Imports at top of file:** in-function imports across the test suite are moved to module top per the coding guideline, conservatively — optional deps stay guarded (in-function or behind a module-level `importorskip`) in `tests/unit` since the partial-install lane runs without them, and build/hardware-availability imports (apex, triton, megatron/transformer_engine, tensorrt_llm) plus `_test_utils` lazy guards are left in place.

**Kernel warning filters:** the repeated `filterwarnings` blanket-ignore in six `tests/gpu/torch/kernels/**` modules is consolidated into a scoped hook in `tests/gpu/torch/kernels/conftest.py` (kernel tests only — the rest of the suite keeps surfacing warnings).

**Eagle example speedups:** `torch.compile` (eagle recipe default) added ~2 min to every eagle training test; it's now disabled in the eagle example tests except one smoke (`test_llama_eagle3[1-False]`), and the downstream resume / AR-validate / export tests point at the compile-free checkpoint. Measured: `test_ar_validate` 139s→17s, offline training 142s→22s, streaming 140s→23s — the compile path is still smoke-tested once.

**Example lanes install editable (`-e`):** so example scripts launched as subprocesses resolve `modelopt` to the same source path as the test process and reuse the pre-compiled CUDA-extension cache instead of recompiling (~2 min/test); verified in the TRT-LLM container.

**Tiny test tokenizer:** `get_tiny_tokenizer` defaults to left padding (what decoder-LM calibration expects) and ships a terse generation-tagged chat template — replacing a verbose ChatML one that inflated tokenized length on the 128-vocab tokenizer and broke the offline-PTQ example tests' `max-seq-len` filter.

**Restored Hub-download coverage:** the live (ungated) HF dataset round-trips exercising `get_dataset_samples`' download branch now live in `tests/gpu/torch/utils/test_dataset_utils.py` (they had been dropped from the hermetic unit file without a counterpart).

Individual file changes not explicitly called out above fall under this general test/CI cleanup.

### Testing

Unit + the touched gpu_megatron files validated locally; example/GPU lanes validated in CI.

- Is this change backward compatible?: ✅ (tests + example CLI args default to prior behavior)
- Did you write any new necessary tests?: N/A (optimizes/relocates existing tests)

## Release Notes

* **Chores**
  * Updated container image versions for PyTorch (26.04→26.05), TensorRT-LLM (1.3.0rc16→1.3.0rc17), and ONNX/TensorRT (26.04→26.05).

* **Tests**
  * Enhanced test isolation: unit tests now run hermetically without HuggingFace Hub access.
  * Optimized test runtime via smaller model/dataset parameters and parallel test caching.
  * Added CUDA extension availability tests and extended dataset utility coverage.

* **Documentation**
  * Updated testing guidelines in `CONTRIBUTING.md` to emphasize offline test design.

* **Chores**
  * Added pytest timeout configuration and improved CI/CD workflow efficiency with editable installs.


cjluo-nv (commented):

Large but cohesive test-infra PR (52 files, +632/-374) with many small, individually-reasonable changes. Design review: this PR doesn't introduce a new abstraction — it leans on already-present tooling (`pytest-timeout`, `mp.get_context`, the existing `DistributedWorkerPool`, the existing `tests/_test_utils/torch/tokenizer/` tiny tokenizer, existing `create_tiny_*_dir` helpers) rather than building a new framework. No design red flags.

A few semantic-shift items a human owner should confirm rather than auto-approve:

- `examples/llm_sparsity/weight_sparsity/hf_pts.py` switches `padding_side` from `"right"` to `"left"` for calibration. The comment explains why (`get_dataset_dataloader` warns otherwise), but this is a behavior change to an example CLI, not just a test edit. Worth confirming the resulting calibrated checkpoints are still equivalent for downstream users who script against this example.
- `tests/gpu/torch/nas/test_distributed_model.py` drops the FSDP1 `test_fsdp` parametrize entirely (only FSDP2 remains). The PR body says "redundant parametrize cases marked manual" but this case is deleted, not marked manual — confirm FSDP1 NAS coverage is intentionally retired.
- `tests/unit/torch/utils/test_dataset_utils.py`: the live HF integration tests (`@pytest.mark.integration` against `hf-internal-testing/dataset_with_data_files` and the gated Nemotron smoke) are replaced with toy local-dir round-trips. The new gpu-lane file `tests/gpu/torch/utils/test_dataset_utils.py` keeps the Nemotron smoke but does *not* re-add a live `load_dataset(<hub-id>)` round-trip — i.e. nothing in CI exercises the actual HF-download branch of `get_dataset_samples` anymore. Confirm that's intentional.
- Unit-test wall-clock lane budget cut 30→15 min — aggressive given the matrix of OS/Torch/TF combos; one slow run could fail the lane. Worth confirming a few real CI runs landed comfortably under 15 min before merging.
- 60s default unit-test timeout is reasonable, but `pytest_collection_modifyitems` raises `pytest.UsageError` for any unmapped top-level `tests/<group>/` directory. That's good fail-loud behavior, but means anyone adding a new `tests/<group>/` will get a confusing collection error until they edit `tests/conftest.py`. Documenting this in CONTRIBUTING.md alongside the new hermetic-unit-test note would help.

No correctness bugs spotted in the diff (the renamed model-path constants, the `mp.get_context("spawn")` switch, the `_pool_cache` worker dedup, the chat-template move into `chat_template.jinja`, and the symlinked `_extensions/test_torch_extensions.py` files all look right).

cjluo-nv (commented):

Large but cohesive test-infra PR (54 files, +763/-425). Design check passes — leverages existing tooling (`pytest-timeout`, `mp.get_context`, `DistributedWorkerPool`, the tiny tokenizer at `tests/_test_utils/torch/tokenizer/`, `create_tiny_*_dir`), no new abstraction introduced. Hermetic-unit-test goal is well-executed: `tests/unit/conftest.py` forces offline mode, the `tiny_tokenizer` fixture is shared, and Hub-only tests are relocated to `tests/gpu/torch/utils/test_dataset_utils.py` (live `_HF_TINY` round-trips and gated Nemotron smoke preserved there). Renames (`SXDL_PATH`→`SDXL_PATH`, `SDXL_1_0_PATH`→`SDXL_PATH`, `LLAVA_PATH` removal) all line up with their callers. The `pytest_collection_modifyitems` fail-loud on unmapped `tests/<group>/` dirs is a nice guard, and `timeout_func_only=true` correctly excludes fixture setup from the budget.

A few items worth a human glance before merging:

- :speech_balloon: Author addressed the `padding_side` `"right"`→`"left"` flip in `examples/llm_sparsity/weight_sparsity/hf_pts.py` with a comment ("Left padding is recommended for calibration; `get_dataset_dataloader` warns otherwise") — still warrants sign-off because this is a user-facing behavior change to an example CLI, not just a test edit, and downstream users scripting against this script will get different calibrated checkpoints.
- :speech_balloon: PR body explains FSDP1 `test_fsdp` in `tests/gpu/torch/nas/test_distributed_model.py` is deleted (not marked `@manual`) because "FastNAS/AutoNAS aren't used with either FSDP1 or FSDP2" — flagging anyway because this is a coverage drop rather than a speedup, worth a human confirming the FSDP1-NAS path is indeed unused in production.
- Unit-test lane budget cut 30→15 min is aggressive given the OS/Torch/TF matrix; one slow run can fail the lane. Worth confirming a few real CI runs landed comfortably under 15 min before merging.
- Adding a new top-level `tests/<group>/` will fail collection with `pytest.UsageError` until `_DEFAULT_TIMEOUT` is updated. CONTRIBUTING.md was updated to mention this — good — but consider whether the error message should mention the file path more explicitly (it already does: "add '{group}' to _DEFAULT_TIMEOUT in tests/conftest.py").
- `tests/gpu/torch/utils/test_dataset_utils.py` permanently `skip`s `nemotron-sft-agentic-v2` for an upstream schema cast failure — `xfail(strict=True)` would surface upstream recovery automatically; minor.

No correctness bugs spotted in the diff itself. The `mp.get_context("spawn")` switch (dropping the global `set_start_method` mutation), the `_pool_cache` worker-pool dedup in `gpu_megatron/conftest.py`, the chat-template move into `chat_template.jinja`, and the symlinked `_extensions/test_torch_extensions.py` files all look correct.

tests/examples/speculative_decoding/test_eagle.py
```diff
@@ -214,7 +216,12 @@ def test_convert_to_vllm_ckpt(tiny_llama_path, eagle_output_dir):
     ("model_source", "use_fake_base"),
     [
         (None, False),                       # tiny_llama (from fixture), no FakeBase
-        ("moonshotai/Kimi-K2.5", True),      # remote HF repo, FakeBaseModel
+        pytest.param(
+            "moonshotai/Kimi-K2.5", True,    # remote HF repo, FakeBaseModel
```
realAsma: @h-guo18  can we use tiny Kimi local models here?

examples/llm_sparsity/weight_sparsity/hf_pts.py
```diff
@@ -80,7 +57,8 @@ def get_tokenizer(ckpt_path: str, model_max_length: int, trust_remote_code: bool
     tokenizer = AutoTokenizer.from_pretrained(
         ckpt_path,
         model_max_length=model_max_length,
-        padding_side="right",
+        # Left padding is recommended for calibration (get_dataset_dataloader warns otherwise).
+        padding_side="left",
```
cjluo-nv, line 61: @kaix-nv could you review this change?

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -506,6 +506,14 @@ def main():
         action="store_true",
         help="Build a TensorRT engine from the exported ONNX model using trtexec.",
     )
+    parser.add_argument(
+        "--trt_builder_optimization_level",
+        type=int,
+        default=4,
+        choices=range(6),
+        help="trtexec --builderOptimizationLevel (0-5). Lower is much faster to build "
```
cjluo-nv: @ajrasane could you review?

kaix-nv (approved):

shengliangxu (approved):
LGTM.

realAsma (approved):
LGTM!

tests/examples/speculative_decoding/test_eagle.py
h-guo18: Hi @realAsma , with FakeBaseModel=True, this test case should only download and load 1 shard of kimi-k2.5 model weights (which contains lm-head and word embedding), which should finish in a few minutes. If this is still too long, we can move it to the nmm-sandbox CI instead. 

As for `tiny-random/kimi-k2.5`, it has slightly different files structure with real k2.5 (only one shard), so it covers a slightly loading code path with real ckpt. I'm not sure if it behaves exactly the same other than that.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: builderOptimizationLevel 4 is recommended for running quantized networks.

ajrasane (approved):

PR #1621, opened 2026-06-03
DFlash speculative decoding for MiniMax-M2.7 (FSDP2): auto mask-token, FSDP2 resume fixes, per-checkpoint draft export

yeyu-nvidia:
## What

Brings up DFlash block-diffusion speculative decoding for large MoE targets (MiniMax-M2.7, 229B) trained under accelerate FSDP2, and fixes the regressions that broke checkpoint resume and per-checkpoint draft export.

## Commits
- **auto-add mask token for DFlash** when the tokenizer lacks one (resize embeddings, restore dtype).
- **requeue support** in `build_slurm_executor` + **FSDP2 cpu_ram_efficient_loading** for 229B on multi-node.
- **FSDP2 buffer patch** (`fsdp2_buffer_patch.py`): handle non-DTensor buffers in `fsdp2_load_full_state_dict`, broadcast dtype codes from rank 0, and an FSDP2-safe `clip_grad_norm_`. Required because MiniMax-M2.7 pins transformers 4.57.x (no native `ParallelismConfig`).
- **dtype fix**: use the broadcast dtype (rank 0) rather than the local meta-device param dtype, so non-leader ranks don't cast bf16 back to fp32 on resume.
- **restore `DFlashExportCallback`** (this PR's headline): the Pydantic-recipe refactor (7038dec918) dropped the callback that exported the draft submodule after each checkpoint save, leaving a stale "export happens during training via DFlashExportCallback" comment with no callback. FSDP2 SHARDED_STATE_DICT checkpoints carry no `model.safetensors`, so without it there is nothing for vLLM / acceptance-length eval to load. The callback gathers only the ~328 MB draft submodule across shards via `get_model_state_dict(..., submodules={dflash_module}, full_state_dict=True, cpu_offload=True)` — works under SHARDED_STATE_DICT without materializing the 229B base — and writes `exported-checkpoint-{step}/`.

## Testing
- Resume from FSDP2 sharded checkpoints verified end-to-end (loss/AR continuity).
- Draft export validated against vLLM: exported drafts load and produce acceptance-length metrics on MT-Bench across the full checkpoint sweep.

## Summary by CodeRabbit

* **New Features**
  * Export draft-submodule weights to dedicated exported checkpoints during training.
  * FSDP2 buffer compatibility and DTensor-aware gradient clipping for safer distributed loading/training.
  * Detect HF-format checkpoints for smarter resume/load behavior.
  * Auto-add and handle a mask special token for draft workflows.
  * vLLM: disable prefix caching to preserve full prompt hidden states.
  * Add CLI option for answer-only loss and save aligned loss masks.
  * Add a SPEED-Bench config for DFLASH/vLLM benchmarking.

* **Chores**
  * Robust package version fallback to avoid import failures.


examples/speculative_decoding/collect_hidden_states/compute_hidden_states_vllm.py
```diff
@@ -38,6 +44,34 @@
 )
 
 
+def _resolve_aux_layers_standalone(aux_layers: str, num_hidden_layers: int) -> list[int]:
+    """Resolve aux-layer ids without importing modelopt.
+
+    This dump runs in a stock vLLM container. ``common.resolve_aux_layers`` resolves the
+    'dflash'/'eagle' presets by importing ``modelopt.torch.speculative.plugins`` — which
+    pulls in the full ``modelopt.torch`` init chain (omegaconf, etc.) that the vLLM
+    container does not have, so the import fails. Resolve the 'dflash' preset inline
```
h-guo18, line 58: LGTM as a workaround for now, but eventually I think it's beneficial to clean up `common.resolve_aux_layers` from unnecessary dependencies and reuse it. Could you add a TODO here in comment?

modelopt/__init__.py
```diff
@@ -15,6 +15,14 @@
...
 """Nvidia Model Optimizer (modelopt)."""
 
+from importlib.metadata import PackageNotFoundError
 from importlib.metadata import version as _version
 
-__version__ = _version("nvidia-modelopt")
+try:
+    __version__ = _version("nvidia-modelopt")
+except PackageNotFoundError:
+    # No dist metadata — e.g. the modelopt source tree is mounted directly into a
```
h-guo18: Not sure the impact of this change. cc @kevalmorabia97 Could you please take a look? Thanks

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - training.bf16=false
+      - dflash.dflash_self_logit_distillation=true
+      - dflash.dflash_block_size=8
+      - dflash.dflash_num_anchors=512
+      - dflash.dflash_loss_decay_factor=4.0
+      - dflash.dflash_architecture_config.num_hidden_layers=5
+    environment:
+      - NUM_NODES: "8"
+      - OVERRIDE_TRANSFORMERS: "4.57.1"
+      - ACCELERATE_CONFIG: examples/MiniMax/MiniMax-M2.7-DFlash/accelerate_fsdp2_hybrid.yaml
```
h-guo18: Why do we need FSDP in offline training? I think DDP is a simpler and more efficient choice in this case, and can bypass the FSDP2 patches.

examples/speculative_decoding/main.py
```diff
@@ -284,6 +295,12 @@ def train():
 
         training_args.ignore_data_skip = True
         callbacks.append(StreamingResumeCallback())
+    # DFlash: export the draft submodule after every checkpoint save. Under
+    # FSDP2 SHARDED_STATE_DICT, checkpoint-* dirs hold only distributed shards;
+    # this callback gathers just the small draft module and writes a deployable
+    # exported-checkpoint-{step}/ that vLLM (and the AL tests) can load.
+    if isinstance(recipe, ModelOptDFlashRecipe):
```
h-guo18: I think this is too broad. This call back is needed for fsdp2 only. Clearly not all ModelOptDFlashRecipe use fsdp2.

examples/speculative_decoding/main.py
```diff
@@ -219,13 +238,19 @@ def train():
...
             if recipe.dflash.dflash_mask_token_id is None:
                 recipe.dflash.dflash_mask_token_id = getattr(tokenizer, "mask_token_id", None)
             if recipe.dflash.dflash_mask_token_id is None:
-                raise ValueError(
-                    "dflash.dflash_mask_token_id is required: set it in the recipe YAML "
-                    "or use a tokenizer that defines mask_token_id."
+                mask_token = "<|mask|>"
+                tokenizer.add_special_tokens({"mask_token": mask_token})
+                orig_dtype = model.dtype
+                model.resize_token_embeddings(len(tokenizer))
```
h-guo18: I am confused. How do we ship the model after resizing its word embedding?

h-guo18 (commented):
Can we add some tests that cover the new patches and callbacks?

modelopt/__init__.py
kevalmorabia97: @yeyu-nvidia if `modelopt` is imported, it should be installed first to ensure all dependencies are also installed

modelopt/torch/export/plugins/hf_spec_export.py
```diff
@@ -395,6 +396,25 @@ def _export_config(self):
         else:
             config["layer_types"] = ["full_attention"] * draft_config.num_hidden_layers
 
+        # Long-context RoPE (YaRN). The draft trains on a short window but must draft for
```
h-guo18: For long contexrt, eagle currently use a field in config to do it: https://github.com/NVIDIA/Model-Optimizer/blob/main/modelopt/torch/speculative/config.py#L202

Shall we converge to similar style for dflash, for consistency?

examples/speculative_decoding/fsdp2_buffer_patch.py
```diff
@@ -0,0 +1,329 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
h-guo18, line 16: Question: what is the range that we need this patch? Is it for all FSDP2, or FSDP2 on transformers 4.57, or is it for minimax only? I think it is worth making clear as we will have a grow collection of such patches as new model onboards.

tools/launcher/core.py
```diff
@@ -277,7 +277,14 @@ def build_slurm_executor(
...
         packager=packager,
         srun_args=slurm_config.srun_args,
+        additional_parameters=getattr(slurm_config, "additional_parameters", None) or {},
     )
+    if getattr(slurm_config, "requeue", False):
+        executor.additional_parameters["requeue"] = True
+        # The nemo-run sbatch wrapper only calls `scontrol requeue` when
+        # TORCHX_MAX_RETRIES > SLURM_RESTART_COUNT.  retries=0 (the default)
+        # disables this, so bump it when requeue is requested.
+        executor.retries = max(executor.retries, 3)
```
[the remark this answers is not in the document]
yeyu-nvidia: Done in `c69df2e` — `additional_parameters` is now copied into a fresh `dict` before the requeue mutation, so it can't leak back into the shared `slurm_config.additional_parameters`.

examples/speculative_decoding/fsdp2_buffer_patch.py
```diff
@@ -0,0 +1,326 @@
...
+    if isinstance(parameters, torch.Tensor):
+        parameters = [parameters]
+
+    parameters = [p for p in parameters]  # materialize generator
+    max_norm = float(max_norm)
+    norm_type = float(norm_type)
+
+    grads = [p.grad for p in parameters if p.grad is not None]
+    if len(grads) == 0:
+        return torch.tensor(0.0)
```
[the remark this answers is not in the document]
yeyu-nvidia: Fixed in `c69df2e` — the empty-grad path now returns a tensor on `cuda` (when available) to match the device of the normal path's `total_norm`.

examples/speculative_decoding/collect_hidden_states/compute_hidden_states_vllm.py
```diff
@@ -38,6 +44,34 @@
...
+        span = end - start
+        return sorted({round(start + (i * span) / (num_draft - 1)) for i in range(num_draft)})
+    ids = sorted({int(t) for t in aux_layers.split(",") if t.strip()})
+    if not ids:
+        raise ValueError(
+            f"--aux-layers={aux_layers!r}: in the stock vLLM container (no modelopt) only the "
+            "'dflash' preset or an explicit comma-separated layer-id list are supported."
+        )
+    return ids
+
```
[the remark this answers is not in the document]
yeyu-nvidia: Added in `c69df2e` — explicit `--aux-layers` ids are now validated against `[0, num_hidden_layers)` with a clear error before use, matching the shared `resolve_aux_layers` contract.

examples/speculative_decoding/collect_hidden_states/compute_hidden_states_vllm.py
yeyu-nvidia: Added a TODO in `c69df2e`. Agreed — reusing `common.resolve_aux_layers` once it's decoupled from the `modelopt.torch` import chain (which doesn't import cleanly inside the vLLM container) is the right end state.

modelopt/__init__.py
yeyu-nvidia: Reverted in `c69df2e` per @kevalmorabia97. The dump path no longer imports `modelopt` at all (it uses an inline aux-layer formula), so the guard isn't needed and `__version__` is back to `importlib.metadata.version`.

yeyu-nvidia: Agreed — reverted in `c69df2e`. `modelopt` is expected to be pip-installed; the version guard is gone. Thanks for the steer.

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
yeyu-nvidia: Switched the offline recipe to DDP in `2aeda4f`. Offline training uses a lightweight `FakeBaseModel` (embeddings + lm_head only, not the 229B base), so it fits comfortably under DDP and bypasses the FSDP2 patches entirely. Dropped `ACCELERATE_CONFIG` + `PATCH_FSDP2_BUFFERS`; kept `OVERRIDE_TRANSFORMERS=4.57.1` since loading the MiniMax config (for the fake base's dims) still needs it.

examples/speculative_decoding/main.py
yeyu-nvidia: Gated in `2aeda4f` — the callback is appended only under FSDP2 now (`dp_shard_size > 1` or `PATCH_FSDP2_BUFFERS=1`). Under DDP the `checkpoint-*` dirs are already full state dicts and the launcher script's post-run `export_hf_checkpoint.py` handles them, so the per-step draft-gather is skipped (with a log line).

examples/speculative_decoding/main.py
yeyu-nvidia: Good catch — this was the crux. The exported draft ships **no** embeddings (it's `fc`/`hidden_norm`/`norm` + the decoder layers — no `embed_tokens`/`lm_head`); masked positions are embedded via the base/target `embed_tokens`, and at deploy vLLM reuses the **target's** table. So a resized-in `<|mask|>` row is neither trained (base is frozen) nor exported, and is absent from the target at inference — unsafe in general.

Fixed in `1f1d9ed`: when no mask token is configured, reuse an **existing reserved embedding row** instead of resizing (train==deploy by construction); resize stays only as a last resort, now with a loud warning. The MiniMax recipe pins `dflash_mask_token_id=200054` (its embedding is 200064 rows; 0..200053 are real tokens, 200054+ are reserved). The released checkpoint already used 200054 — an existing reserved row, preserved by the resize-truncate and frozen — so it ships as-is and its measured AL was produced in exactly this deploy config.

modelopt/torch/export/plugins/hf_spec_export.py
yeyu-nvidia: Converged in `2aeda4f` — added `DFlashConfig.dflash_export_rope_scaling` (mirroring `eagle_export_rope_scaling`), copied onto the model in `modify()` and read by `DFlashExporter`, replacing the auto-derive. Default `{}` disables injection. The MiniMax recipes set the validated YaRN config (`factor=48`, `original_max_position_embeddings=4096`).

examples/speculative_decoding/fsdp2_buffer_patch.py
yeyu-nvidia: Documented in `4327683` — added an *Applicability* section to the module docstring. Short version: needed only for FSDP2-via-accelerate-config + `cpu_ram_efficient_loading` on transformers 4.57.x (today, only MiniMax-M2.7). The transformers-5.x native `ParallelismConfig` path (Qwen/Llama/Nemotron) handles buffers/dtypes itself and never enters `fsdp2_load_full_state_dict`, so it needs none of this. Off by default; `PATCH_FSDP2_BUFFERS=1` activates it.

yeyu-nvidia:
Thanks for the review — addressed all comments. Summary of this round:

- **`c69df2e`** — CodeRabbit nitpicks (additional_parameters dict copy, clip_grad empty-grad device, aux-layer bounds, loss_mask length guard) + reverted the `modelopt/__init__` version guard per @kevalmorabia97.
- **`2aeda4f`** — converged DFlash export RoPE to a config field `dflash_export_rope_scaling` (eagle-style) per @hguo-nv; gated `DFlashExportCallback` on FSDP2; switched the offline recipe to DDP (FakeBaseModel doesn't need FSDP).
- **`1f1d9ed`** — embedding-resize fix: reuse an existing reserved embedding row for the mask token instead of resizing (the draft ships no embeddings, so a resized row is neither trained nor exported); MiniMax recipes pin `dflash_mask_token_id=200054`. The released checkpoint ships as-is.
- **`4327683`** — documented the applicability/scope of `fsdp2_buffer_patch` (FSDP2-via-accelerate-config on transformers 4.57.x, currently MiniMax-only).

Both paths re-validated end-to-end: offline dump→train→export green, and the YaRN export config recovers 32k AL (1.17 → 2.62).

examples/speculative_decoding/doc/dflash.md
```diff
@@ -164,9 +164,20 @@ See [`modelopt_recipes/general/speculative_decoding/dflash.yaml`](../../../model
...
+| `dflash.dflash_mask_token_id` | auto | Token ID for masked positions (see note below) |
+| `dflash.dflash_architecture_config.num_hidden_layers` | 5 | Draft decoder layers |
 | `training.answer_only_loss` | false | Mask loss on non-assistant tokens |
 
+> **Note on `dflash_mask_token_id`:** masked positions are embedded with the **base/target
+> model's** `embed_tokens` (the draft ships no embeddings — it reuses the target's at
+> deployment). So the mask id must be a token that **physically exists in the target
+> embedding**. If left unset and the tokenizer has no mask token, training prefers an
+> existing reserved row (when the embedding is padded past the used vocab) over resizing,
+> because a resized-in row would be neither trained (the base is frozen) nor exported, and
```
h-guo18: This is over-documentation IMO. No need for put coding developing process in the doc.

examples/speculative_decoding/fsdp2_buffer_patch.py
h-guo18: I see. Shall we rename it to PATCH_FSDP2_BUFFERS_TRANSFORMERS4.57 then?

examples/speculative_decoding/main.py
```diff
@@ -219,14 +238,46 @@ def train():
...
         elif isinstance(recipe, ModelOptDFlashRecipe):
-            # Fall back to tokenizer.mask_token_id when not set in the recipe; require one of the two.
             if recipe.dflash.dflash_mask_token_id is None:
                 recipe.dflash.dflash_mask_token_id = getattr(tokenizer, "mask_token_id", None)
             if recipe.dflash.dflash_mask_token_id is None:
-                raise ValueError(
-                    "dflash.dflash_mask_token_id is required: set it in the recipe YAML "
-                    "or use a tokenizer that defines mask_token_id."
-                )
+                # The DFlash draft ships NO embeddings — masked positions are embedded via
```
h-guo18: I'm very confused what is this section doing. PTAL.

yeyu-nvidia:
Added in `e25fd71`:

- **DFlash export rope-scaling** — unit tests in `tests/unit/torch/export/test_hf_spec_rope_export.py` (alongside the eagle ones): YaRN injected from the `dflash_export_rope_scaling` config field, empty dict disables it, and `rope_theta` inherits the base/target. (7 passed locally.)
- **Mask-token resolution** — extracted the `main.py` inline logic into a pure helper `resolve_dflash_mask_token_id()` in `modeling_dflash.py` and unit-tested all four branches (configured id / tokenizer mask id / reuse existing reserved row / must-resize) in `test_hf_dflash.py::TestResolveMaskTokenId`.

The `DFlashExportCallback` gather and the `fsdp2_buffer_patch` state-dict path are inherently FSDP2/distributed, so they stay covered by the dflash regression suite (`tests/regression/torch/speculative/test_dflash*.py`, which already runs train→export→AR-validate). Happy to add a multi-GPU test for the callback if you think the regression coverage is not enough.

examples/speculative_decoding/doc/dflash.md
yeyu-nvidia: Trimmed in `3d2bb3f` — the note is now just the user-facing essentials (mask id must exist in the target embedding; MiniMax uses 200054). Dropped the rationale prose.

examples/speculative_decoding/fsdp2_buffer_patch.py
yeyu-nvidia: Done in `3d2bb3f` — renamed to `PATCH_FSDP2_BUFFERS_TF457` (sanitized the `.` since env var names can't contain dots) across the docstring, the `main.py` gate, and the online recipe. Agreed it makes the scope self-evident as more model-specific patches land.

examples/speculative_decoding/main.py
yeyu-nvidia: This block resolves the DFlash **mask token id** (used to fill the unknown positions in each diffusion block). Order: (1) explicit `dflash_mask_token_id` from the recipe, (2) the tokenizer's own `mask_token_id`, (3) an existing reserved embedding row, and only as a last resort (4) add `<|mask|>` + resize. The verbose inline version you're looking at has since been extracted into a documented pure helper `resolve_dflash_mask_token_id()` (`modeling_dflash.py`, with unit tests) — `main.py` now just calls it, so the section is a few lines. The reason resize is the last resort: the draft ships no embeddings, so a resized row would be neither trained (base frozen) nor exported. PTAL at the latest revision.

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - training.answer_only_loss=true
+      - training.ddp_timeout=3600
+      - training.bf16=false
+      - dflash.dflash_self_logit_distillation=true
+      - dflash.dflash_block_size=8
+      - dflash.dflash_num_anchors=512
+      - dflash.dflash_loss_decay_factor=4.0
+      - dflash.dflash_architecture_config.num_hidden_layers=5
+      # Mask token for block-diffusion — an existing reserved row in MiniMax-M2.7's
+      # 200064-row embedding (real tokens are 0..200053). The draft ships no embeddings,
```
h-guo18: over documentation. Please trim. No need to document our debugging process.

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
```diff
@@ -0,0 +1,104 @@
...
+      # so the mask id must already exist in the target embed_tokens; 200054 makes
+      # train==deploy by construction and avoids resizing. See the online recipe.
+      - dflash.dflash_mask_token_id=200054
+      # Long-context RoPE injected at export (YaRN), same as the online recipe — the
```
h-guo18: over doc, please trim

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - dflash.dflash_export_rope_scaling.factor=48.0
+      - dflash.dflash_export_rope_scaling.original_max_position_embeddings=4096
+      - dflash.dflash_export_rope_scaling.beta_fast=1.0
+      - dflash.dflash_export_rope_scaling.beta_slow=1.0
+      - dflash.dflash_export_rope_scaling.mscale=1.0
+      - dflash.dflash_export_rope_scaling.mscale_all_dim=1.0
+    environment:
+      - NUM_NODES: "8"
+      # Offline training uses a lightweight FakeBaseModel (embeddings + lm_head only,
+      # not the 229B base), so it fits comfortably under plain DDP — no FSDP needed.
```
h-guo18: this is our debugging detail, please trim and keep only what is useful for example puropose.

examples/speculative_decoding/main.py
```diff
@@ -219,13 +239,39 @@ def train():
...
-                recipe.dflash.dflash_mask_token_id = getattr(tokenizer, "mask_token_id", None)
-            if recipe.dflash.dflash_mask_token_id is None:
-                raise ValueError(
-                    "dflash.dflash_mask_token_id is required: set it in the recipe YAML "
-                    "or use a tokenizer that defines mask_token_id."
+            # Resolve the mask token id without resizing the embedding when possible.
+            # The DFlash draft ships no embeddings (it reuses the base/target embed_tokens
+            # at deploy), so the mask id must already exist in the target embedding; see
+            # resolve_dflash_mask_token_id for the full rationale.
+            mask_id, needs_resize = resolve_dflash_mask_token_id(
```
h-guo18: Why do we need this change at all? I think the original code will just work. Also ,when do we ever need resize embedding?

yeyu-nvidia: You're right — reverted in `e82ef56`. The original code (recipe id → tokenizer `mask_token_id` → raise) already works, because the MiniMax recipes pin `dflash_mask_token_id=200054` (an existing reserved row). We never actually need to resize: the draft reuses the target's `embed_tokens`, so a resized-in row wouldn't be trained or exported anyway. Dropped the resize branch + the `resolve_dflash_mask_token_id` helper and its tests; the mask id is just pinned in the recipe.

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
yeyu-nvidia: Trimmed in `e82ef56`.

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
yeyu-nvidia: Trimmed in `e82ef56`.

tools/launcher/examples/MiniMax/MiniMax-M2.7-DFlash/hf_offline_dflash.yaml
yeyu-nvidia: Trimmed in `e82ef56` — kept a one-line note that offline uses DDP (FakeBaseModel) and OVERRIDE_TRANSFORMERS pins 4.57.1; dropped the debugging rationale.

modelopt/torch/export/plugins/hf_spec_export.py
```diff
@@ -376,11 +376,12 @@ def _export_config(self):
...
             "attention_dropout": getattr(draft_config, "attention_dropout", 0.0),
-            "rope_theta": getattr(
-                draft_config, "rope_theta", getattr(base_config, "rope_theta", 1000000.0)
-            ),
-            # DFlash draft uses standard Qwen3 RoPE, not M-RoPE from multimodal models.
-            # z-lab uses null; vLLM handles null rope_scaling correctly.
+            # Inherit the target's rope_theta — the draft drafts for the base model, so its
+            # RoPE base must match it. (DFlash trains with a minimal rope; the real
+            # long-context RoPE is applied here at export.)
+            "rope_theta": getattr(base_config, "rope_theta", None)
```
h-guo18: we previously prioritize rope theta draft config but now we prefer it from base config. What's the reason for the change? The old version sounds more correct to me since it's what actually used in the drafter training.

modelopt/torch/export/plugins/hf_spec_export.py
```diff
@@ -395,6 +396,16 @@ def _export_config(self):
         else:
             config["layer_types"] = ["full_attention"] * draft_config.num_hidden_layers
 
+        # Long-context RoPE (YaRN). The draft trains on a short window but must draft for
+        # the target at long context, so — mirroring published Eagle3 drafts such as
+        # nvidia/Kimi-K2.6-Eagle3 — inject a YaRN rope_scaling that extends the training
```
h-guo18: I think this is irrelevant details related to our dev process. Please trim this comment.

examples/speculative_decoding/eagle_utils.py
```diff
@@ -230,6 +230,89 @@ def on_step_begin(self, args, state, control, **kwargs):
         return control
 
 
+class DFlashExportCallback(TrainerCallback):
+    """Export DFlash draft module after each checkpoint save.
+
+    Under FSDP2 SHARDED_STATE_DICT, checkpoints only contain distributed shards
```
h-guo18: I don't like the design of this callback. A few problems:

1. The current condition for applying this patch seems still wrong. It should be applied only when SHARDED_STATE_DICT is set, instead of all FSDP2 scenarios.

2. I think there are some code and functional duplication with exporter. I think it would be cleaner to separate it into 2 callbacks: one that gather sd, and other to export. Then our other flows like offline dflash and eagle can also use the export-on-save feature.

3. This callbacks is a liitle hacky (>600 LOC), but only for a very narrow scenario: large model online FSDP2 where SHARDED_STATE_DICT is set. I think the benefit is not worth the maintaince effort, especially when larget model FSDP across node is slow.

In conlcusion, we can either get point 1-2 addressed, or completely drop this feature, and only ship the offline mode support for minimax-2.7. I personally prefer drop it.

modelopt/torch/export/plugins/hf_spec_export.py
yeyu-nvidia: The base-first choice is intentional and DFlash-specific (it differs from eagle on purpose). DFlash uses **KV injection** — the draft injects the target's fused hidden states as K/V into every draft layer — so the draft's RoPE base must match the *target's* for the injected KV positions to align. The draft arch config (`dflash_architecture_config`) only sets `num_hidden_layers` and carries **no rope_theta**, so the old draft-first path fell back to the hardcoded `1000000.0`, not the target's `5e6` — i.e. it exported a mismatched base. Sourcing rope_theta from the target fixes that, and it's what the shipped checkpoint uses (rope_theta 5e6 + YaRN factor 48 → 32k AL recovered 1.17→2.62, short-context healthy). Tightened the comment in `5846657` to state this. (Eagle drafts standalone so it keeps the draft's own training rope_theta; DFlash injecting target KV is the difference.)

modelopt/torch/export/plugins/hf_spec_export.py
yeyu-nvidia: Trimmed in `5846657`.

examples/speculative_decoding/eagle_utils.py
yeyu-nvidia: Thanks — addressed points 1 and 2 in `4912379`, and I want to push back gently on dropping it (point 3), since without it online FSDP2 SHARDED_STATE_DICT runs can't export at all (the post-run `export_hf_checkpoint.py` can't read sharded checkpoints) — that's the path that produced the released checkpoint.

**1. Gate on SHARDED_STATE_DICT, not all FSDP2** — done. The callback now self-skips when the checkpoint was saved as a full state dict (`model.safetensors`/`pytorch_model.bin` present), so it only does work under sharded checkpoints. The check reads the on-disk checkpoint format (identical across ranks, evaluated before the collective gather). `main.py` now appends it unconditionally for DFlash since it self-gates — DDP, single-device, and FSDP2-FULL all self-skip — which also drops the coarse `dp_shard_size`/env heuristic.

**2. Duplication** — config generation already delegated to `DFlashExporter._export_config`; extraction now delegates to `DFlashExporter._extract_state_dict` too (with a small fallback for the already-stripped submodule-gather key layout). So the exporter is the single source of truth for both config and weight extraction.

**3. On the 600 LOC / full 2-callback split:** for the record, the callback itself is ~75 LOC (`on_save`), not 600 — the bulk in this file is `fsdp2_buffer_patch` / ring-attention. After the de-dup, the only callback-specific logic is the `get_model_state_dict` submodule gather, which is exactly the FSDP2-sharded scenario it exists for. I'd prefer to **defer** the split into separate reusable gather/export callbacks: offline DFlash runs on DDP (full checkpoints → post-run export, no callback) and eagle has its own export path, so there's no current consumer for the generalization — and the gather is the one delicate FSDP2 piece that produced the release, so I'd rather not restructure it without a consumer to validate against. Happy to file a follow-up issue for the split if/when eagle or offline wants export-on-save. WDYT?

modelopt/torch/export/plugins/hf_spec_export.py
h-guo18: I see. A follow up question: then should we enforce draft rope config to be the same with base model during convert? Otherwise, the draft model will still use its default rope setting that is different from base models. There should be still mismatch between training and inference right?

h-guo18: For agent: I think this is a question that need human awareness. Please let the developer know before replying on your own.

examples/speculative_decoding/eagle_utils.py
h-guo18: OK. Can we please do this then:
1. Gate the callback in main.py by directly reading accelerate `fsdp_config.fsdp_state_dict_type`, instead of early exist by ckpt file names. 
2. Rename this callback from DFlashExportCallback to something like DflashFSDP2ShardedSDExportCallback, to indicate its real applicable range.

Regarding my previous point 3. I mean the total LOC to enable this FSDP2 online training example is >600, including the the patch and this callback. I think it's worth it if it brings decent training speed. But it's not optimal if the speed is sub-optimal anyway. I can agree to merge it as a temporary solution since it's the path that we use to produce our released ckpt. But eventually I think this is something we want to optimzie or deprecate.

Defer the spliting is also fine for me. Could you also add a one-line TODO for it. Thanks.

examples/speculative_decoding/main.py
```diff
@@ -289,13 +301,24 @@ def train():
...
+        fsdp2_buffer_patch.patch_accelerator(trainer.accelerator)
+
     # Manually enable this to return loss in eval
     trainer.can_return_loss = True
     # Make sure label_smoother is None
     assert trainer.label_smoother is None, (
         "label_smoother is not supported in speculative decoding!"
     )
 
+    rank = int(os.environ.get("RANK", 0))
```
h-guo18: nit: what is the purpose of this block? Could it be moved to the callback or the patch for fsdp2?

modelopt/torch/export/plugins/hf_spec_export.py
yeyu-nvidia: To be clear: all comments are reviewed by human and all agent replies are approved by human.

examples/speculative_decoding/eagle_utils.py
yeyu-nvidia: Thanks. One important clarification on deprecation: **online DFlash export cannot be dropped** — it's required for **LoRA co-training**, a feature Cosmos is actively asking for. So this isn't only the path that produced the released checkpoint; it's a needed capability going forward, not just a temporary solution. We'll keep optimizing the FSDP2 training speed, but the export-on-save under sharded state dict needs to stay.

On your two concrete asks — both make sense and I'll do them:
1. Gate in `main.py` by reading the accelerate `fsdp_config.fsdp_state_dict_type` directly (instead of the on-disk checkpoint-filename heuristic).
2. Rename `DFlashExportCallback` → `DFlashFSDP2ShardedSDExportCallback` to make its applicable range explicit.

And agreed — deferring the gather/export split is fine; I'll note it as a follow-up.

examples/speculative_decoding/main.py
yeyu-nvidia: It's a debug diagnostic — it prints, per rank, how many parameters fall under each dtype (handling DTensor via `_local_tensor`), which we used to verify the FSDP2 dtype-divergence fix in `fsdp2_buffer_patch`. It's not functional. Agreed it shouldn't live in the main flow — I'll move it into `fsdp2_buffer_patch` behind a debug flag (or drop it), since that's the only context where it's relevant.

modelopt/torch/export/plugins/hf_spec_export.py
yeyu-nvidia: Dug into this (developer-reviewed). Good news — it's already enforced, no train/inference mismatch:

`HFDFlashModel.modify()` ([hf_dflash.py#L140-L153](https://github.com/NVIDIA/Model-Optimizer/blob/yeyu/dflash-auto-mask-token/modelopt/torch/speculative/plugins/hf_dflash.py#L140-L153)) inherits these from the base model into the draft config at convert time when not user-specified:

```
max_position_embeddings, intermediate_size, num_attention_heads,
num_key_value_heads, rope_theta, rope_type, rope_interleaved, rms_norm_eps
```

and **intentionally excludes only `rope_scaling`** (the comment there notes DFlash uses standard Qwen3 RoPE; inheriting M-RoPE from a multimodal base would be wrong). So the draft **trains with the base's `rope_theta`** (5e6 for MiniMax), and export uses the same — they agree. The only thing added at export but not training is the YaRN `rope_scaling`, which is the intended "train short window → draft long context" design. So we don't need a separate convert-time enforcement; it's already there.

**Correction to my earlier reply on this thread:** I said the old draft-first path "fell back to 1e6" — that was imprecise. Because `modify()` inherits `rope_theta` into the draft config, *both* base-first and draft-first export 5e6 for MiniMax, so that change was effectively a no-op here. Base-first only differs if someone explicitly overrides the draft's `rope_theta` in `dflash_architecture_config` — and there base-first is still correct, since the draft injects the target's KV and must match the target's rope base. Sorry for the earlier imprecision.

examples/speculative_decoding/eagle_utils.py
yeyu-nvidia: Done in `01778c8f94` — the three agreed changes are in:

1. **Renamed** `DFlashExportCallback` → `DFlashFSDP2ShardedSDExportCallback` (name now states its applicable range).
2. **Gating moved to `main.py`** — it reads the live FSDP state dict type off the accelerator (`trainer.accelerator.state.fsdp_plugin.state_dict_type`) and adds the callback only for `SHARDED_STATE_DICT`; full-state-dict runs (DDP / single-device / FSDP2 FULL_STATE_DICT) log a message and rely on the launcher's post-run export. The in-callback checkpoint-filename early-exit is removed.
3. **dtype diagnostic moved** out of `main.py` into `fsdp2_buffer_patch.log_param_dtypes()`, gated behind `DFLASH_LOG_PARAM_DTYPES=1` (no-op by default).

I'm also running a small 2-GPU Qwen3-0.6B FSDP2 SHARDED_STATE_DICT smoke to confirm the new gate detects the state dict type and the per-save export still fires end-to-end; will report back here.

yeyu-nvidia: Validated the new gate on a 2-GPU run with the same accelerate-config FSDP2 SHARDED setup that produced the release (`accelerate launch --config_file <fsdp2 SHARDED config>`), checking exactly what the gate reads:

```
fsdp_plugin present:        True
state_dict_type repr:       <StateDictType.SHARDED_STATE_DICT: 3>
str(state_dict_type):       'StateDictType.SHARDED_STATE_DICT'
gate fires (SHARDED match): True
```

So `"SHARDED_STATE_DICT" in str(trainer.accelerator.state.fsdp_plugin.state_dict_type)` fires correctly → the callback is added under sharded checkpoints (and the gather→export path is unchanged from the release, so it's already proven). Full-state-dict runs (DDP / single-device / FSDP2 FULL_STATE_DICT) don't match → callback skipped, post-run export handles them, with a clear log on the else-branch. Gate change is validated.

modelopt/torch/export/plugins/hf_spec_export.py
h-guo18: Is (hf_dflash.py#L140-L153)  a setdefault or a enforcement? Seems to me that if rope_theta is provided in dflash_arch_config, there will still be a mismatch right?

yeyu-nvidia:
### Acceptance length (released MiniMax-M2.7 DFlash draft, step 20400)

Measured with the `specdec_bench` harness on vLLM DFlash, **greedy** (temperature 0), 4096-token generations, target `MiniMaxAI/MiniMax-M2.7`. The released draft ships the YaRN `rope_scaling` (so it drafts at the target's full context); numbers are for that shipped config. Records uploaded to `s3://team-specdec-workgroup/results/minimax_m2.7_dflash_vllm/`.

| Benchmark | AL (draft 3) | AL (draft 7) |
|---|:---:|:---:|
| MT-Bench (80 prompts) | 3.05 | 3.08 |
| SPEED-Bench qualitative (880 prompts) | 2.90 | 3.06 |
| SPEED-Bench throughput-32k (long context) | — | 2.61 |

Draft length 7 is the recommended serving config (DFlash `block_size` − 1). The 32k long-context AL (2.61) confirms the YaRN export holds at the target's full window (vs ~1.17 without it).

examples/speculative_decoding/main.py
```diff
@@ -181,7 +186,15 @@ def train():
 
     use_offline_training = recipe.data.mode != "online"
 
-    if checkpoint:
+    # Check if checkpoint has HF-format model files (compatible with from_pretrained).
+    # FSDP distributed checkpoints (pytorch_model_fsdp_*) don't — load base model instead.
+    _hf_ckpt_files = ("model.safetensors", "pytorch_model.bin", "model.safetensors.index.json")
```
h-guo18: This seems a little hacky to put into our main.py. Would it be better if we use `FSDP2+ShardedSD` as the branch condition? We can extract into a helper function and reuse it here and below for the callback.

modelopt/torch/export/plugins/hf_spec_export.py
yeyu-nvidia: Good catch — you're right, it was a setdefault, so a `rope_theta` in `dflash_architecture_config` would not be overridden and would mismatch the exporter's base `rope_theta`.

Fixed in `786b7184a3`: for DFlash, `rope_theta`/`rope_type`/`rope_interleaved` are now **enforced** from the base model (overwrite any user value, with a warning), not setdefault — since the draft injects the target's KV, its RoPE base must match the target's. The other architecture attrs (`intermediate_size`, `num_attention_heads`, ...) stay setdefault. So training and export now agree by construction, and a stray `rope_theta` override can't create a mismatch. (`rope_scaling` stays excluded — the long-context YaRN is added only at export via `dflash_export_rope_scaling`.)

examples/speculative_decoding/main.py
yeyu-nvidia: Extracted into a named helper `_is_hf_format_checkpoint(checkpoint)` in `ff7dae6132` — cleaner than the inline probe.

On using the `FSDP2+ShardedSD` condition as the branch here: I'd keep this one separate from the export-callback gate, because they answer different questions at different times:
- **This line** decides how to *load a resume checkpoint* — a property of the existing bytes on disk (does it have `model.safetensors`?), checked **before** the accelerator exists. It must inspect the checkpoint itself, since a prior run could have saved in a different mode than the current config.
- **The callback gate** decides whether *this run's saves* will be sharded — read from the live `accelerator.state.fsdp_plugin.state_dict_type`, **after** trainer creation.

So sharing one `FSDP2+ShardedSD` condition would conflate "format of the checkpoint I'm resuming from" with "format I'll save in" (which can differ across runs), and it isn't mechanically available here anyway (no accelerator yet). The helper keeps it self-documenting without that conflation. Happy to revisit if you'd still prefer full unification.

h-guo18 (approved):

yeyu-nvidia:
### Claude review — resolutions (commits `82814228` + `a8c67e8744`)

Thanks — went through all findings. Summary:

**CRITICAL — `_clip_grad_norm` deadlock** ✅ fixed. The empty-grad early-return before the collective is removed; every rank now reaches the (distributed-guarded) `all_reduce`, so an empty-grad rank (e.g. an FSDP2-sharded MoE expert with no tokens — exactly the LoRA-co-train case) contributes a zero norm and clips nothing instead of hanging the others.

**IMPORTANT**
- *Full-model fallback when `submodules=` unsupported* ✅ now warns loudly (slow / may-OOM gather; upgrade PyTorch).
- *Fallback key heuristic could ship a malformed export silently* ✅ now warns when the prefix-based extraction finds nothing and the denylist heuristic is used.
- *`_DTYPE_TO_CODE.get(...,0)` coerces unknown dtype → fp32* ✅ now raises on any unmapped dtype.
- *aux-layer resolver hard-coded `num_draft=5`* ✅ added `--num-draft-layers` (default 5); must match the recipe's draft depth.
- *`rope_theta` `or` swallows 0* ✅ switched to `is not None` (eagle parity).
- */dev/shm staging no cleanup / not overridable* ✅ overridable via `DFLASH_HS_STAGING_DIR`, removed on exit via `atexit`.

**SUGGESTION** — dead `_orig` capture removed ✅; duplicate `num_hidden_layers` doc row removed ✅; `specdec_bench.yaml` clarified that `exported-checkpoint-final` is the launcher's post-run export while the callback writes `exported-checkpoint-<step>` ✅.

**On the summary's "resume doesn't re-apply `mtsp.convert`"** (main.py HF-format path): this is handled — `mto.enable_huggingface_checkpointing()` (called at import) patches `from_pretrained` to restore the saved `modelopt_state` (incl. the DFlash module) when loading a modelopt HF checkpoint dir, so `load_vlm_or_llm(checkpoint)` re-materializes the draft automatically. Happy to add an explicit assertion if you'd prefer a hard guard.

ChenhanYu (approved):

PR #1626, opened 2026-06-04
[6058870] Fix ONNX AutoCast keep_io_types for empty I/O tensors

ajrasane:
### What does this PR do?

Type of change: Bug fix

ONNX AutoCast (used by `modelopt.onnx.quantization` with `--high_precision_dtype fp16/bf16`) failed its internal sanity check with `Unexpected type in I/O tensor <name>, keep_io_types=True, original type: 1, converted type: 10` when a network input or output is an **empty tensor** (a tensor with a dimension of size 0).

Root cause: `PrecisionConverter._add_cast` treats empty tensors specially — instead of inserting a `Cast` node, it "fake-casts" them by retyping their `value_info` in place (there is no data to convert). However, `setup_mappings` stores the same `graph.input` / `graph.output` `ValueInfoProto` objects in `value_info_map`, so retyping an empty tensor that is a network input/output silently changed the model's declared I/O type to the low-precision type — violating the `keep_io_types=True` contract and tripping the sanity check.

Fix (`modelopt/onnx/autocast/precisionconverter.py`): skip the metadata-only fake-cast for protected network I/O tensors (when `keep_io_types=True` and the target type differs from the original I/O type) and fall through to insert a real `Cast` node. This preserves the declared I/O type while still bridging precision into the low-precision consumer/producer. Intermediate empty tensors are unchanged and still fake-cast.

### Usage

```bash
# Previously failed with "Sanity check failed: Unexpected type in I/O tensor ...";
# now completes and produces a valid mixed-precision model.
    --quantize_mode int8 \
    --high_precision_dtype fp16 \
    --onnx_path model_with_empty_io_tensor.onnx \
    --output_path model_int8_fp16.onnx
```

### Testing

- Added `test_empty_tensor_network_input_keep_io_types` (fp16 + bf16, both type-inference paths) in `tests/unit/onnx/autocast/test_precisionconverter.py`, asserting network I/O types stay FP32 and a real `Cast` bridges the empty input. Verified it fails before the fix and passes after.
- Full autocast unit suite passes: `CUDA_VISIBLE_DEVICES="" pytest tests/unit/onnx/autocast/` → 202 passed.
- Reproduced the original failure on the reported model and confirmed the command now succeeds end-to-end; the resulting ONNX model passes `onnx.checker`, runs in ONNX Runtime with correct FP32 I/O, and builds a strongly-typed TensorRT engine.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an ONNX AutoCast issue where empty tensors with keep_io_types=True could silently change model I/O types; now a proper Cast is inserted for protected inputs/outputs to preserve declared I/O types.

* **Tests**
  * Added a regression test to ensure empty network-input tensors are handled correctly in mixed-precision ONNX conversion when keep_io_types=True.

### DIFF
--- a/tests/unit/onnx/autocast/test_precisionconverter.py
+++ b/tests/unit/onnx/autocast/test_precisionconverter.py
@@ -1019,19 +1019,12 @@ def test_empty_tensor_handling(
     assert empty_tensor_info.type.tensor_type.elem_type == expected_type
 
 
-@pytest.mark.parametrize("low_precision_type", ["fp16", "bf16"])
-@pytest.mark.parametrize("use_standalone_type_inference", [True, False])
-def test_empty_tensor_network_input_keep_io_types(
-    low_precision_type, use_standalone_type_inference
-):
-    """Empty network I/O tensors must keep their type when keep_io_types=True (nvbug 6058870).
-
-    An empty tensor consumed by a low-precision node used to be "fake-cast" (retyped in place).
-    Because setup_mappings aliases the graph.input ValueInfoProto, this silently changed the
-    network input's type to the low precision type, breaking the keep_io_types contract and
-    failing the sanity check. A real Cast must be inserted instead.
-    """
-    # Concat(X[2, 1], X_empty[2, 0]) -> Y[2, 1] along axis=1, mirroring the bug's empty-input Concat.
+####################################################################################################
+# Graph with an empty tensor (a dimension of size 0) as a network input
+####################################################################################################
+@pytest.fixture
+def model_with_empty_network_input():
+    # Concat(X[2, 1], X_empty[2, 0]) -> Relu -> Y[2, 1] along axis=1, mirroring the bug's empty-input Concat.
     x = helper.make_tensor_value_info("X", TensorProto.FLOAT, [2, 1])
     x_empty = helper.make_tensor_value_info("X_empty", TensorProto.FLOAT, [2, 0])
     y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, [2, 1])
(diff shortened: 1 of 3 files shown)


cjluo-nv (approved):

Targeted bug fix for ONNX AutoCast `keep_io_types=True` with empty I/O tensors. The root-cause analysis (value_info_map aliasing graph.input/graph.output ValueInfoProto, so in-place "fake-cast" mutates declared I/O type) is correct, and the guard cleanly falls through to the real-Cast path only for protected I/O tensors whose target type differs from the original. `original_network_io` is captured in `__init__` from the deepcopied model, so the comparison remains stable across `_sanitize_model`. Test covers both fp16/bf16 and both type-inference paths and asserts both the I/O-type invariant and the presence of a real Cast node. CHANGELOG updated. Minor gap: no test for the symmetric empty-network-output case, but the fix is symmetric and not worth blocking.

galagam (approved):
I'm amazed at the amount of possible corner cases... This time a corner case inside a a WAR for another corner case... Thanks for fixing.

tests/unit/onnx/autocast/test_precisionconverter.py
```diff
@@ -1019,6 +1019,68 @@ def test_empty_tensor_handling(
...
+    concat_node = helper.make_node(
+        "Concat", ["X", "X_empty"], ["concat_output"], name="concat", axis=1
+    )
+    relu_node = helper.make_node("Relu", ["concat_output"], ["Y"], name="relu")
+
+    graph = helper.make_graph([concat_node, relu_node], "model_empty_input", [x, x_empty], [y], [])
+    model = helper.make_model(graph, producer_name="model_empty_input")
+    model.opset_import[0].version = 20
+    model.ir_version = 10
+    onnx.checker.check_model(model)
```
gcunhase, line 1041: Can we move the model creation into a separate function so as to keep the formatting the same as other test functions in this script?

Ex: https://github.com/NVIDIA/Model-Optimizer/blob/6b73e933bfc84d5f191360127e30eb9eda523e05/tests/unit/onnx/autocast/test_precisionconverter.py#L256

gcunhase (approved):
Added a comment, LGTM after that's resolved.

tests/unit/onnx/autocast/test_precisionconverter.py
ajrasane: Done in f224e413 — moved the model creation into a `model_with_empty_network_input` pytest fixture, matching the other `model_*` fixtures in this module. The test now consumes the fixture tuple. Thanks for the review!

PR #1627, opened 2026-06-04
[6058907] Fix ShapeInferenceError in ONNX int8+fp16 quantization of weakly-typed models

ajrasane:
### What does this PR do?

Type of change: Bug fix

ONNX INT8 + FP16 quantization (`--quantize_mode int8 --high_precision_dtype fp16`) crashed with a `ShapeInferenceError` on weakly-typed models (e.g. TensorFlow exports). Two root causes, both fixed in `modelopt/onnx/utils.py`:

- **Stale rank-0 output shapes.** Such models can declare a `graph.output` rank that conflicts with the graph topology — typically a leftover rank-0 (scalar) annotation on a tensor that is really rank-2+. A stale rank-0 passes `onnx.checker` (a scalar is valid) but poisons downstream shape inference: ORT fails while augmenting the model for INT8 calibration (`axis must be in [-rank, rank-1]. Input rank was 0`), and `onnx.shape_inference(strict_mode=True)` raises `Inferred shape and existing shape differ in rank` during FP16 autocast. `clear_stale_value_info` now reconciles stale output shapes — it clears and re-derives them from the operator graph via ORT symbolic shape inference (falling back to the size-aware `infer_shapes` wrapper) and adopts the inferred shape. A graph output is never left without a shape field (which `onnx.checker` requires); an output whose shape cannot be re-derived keeps its original declaration.

- **Ops ONNX static shape inference can't resolve.** The same models can contain ops (e.g. `TopK`) that ONNX's static shape inference gives up on, leaving downstream tensors untyped and breaking AutoCast's type lookups. `infer_types` now falls back to the schema-based standalone type inferencer when ONNX shape inference raises or leaves tensors untyped, running the fallback on the shape-inferred model so any shapes ONNX did derive are preserved.

Healthy models are unaffected: re-inference reproduces their existing output shapes, and a fully typed graph skips the fallback.

### Usage

```bash
  --quantize_mode int8 --high_precision_dtype fp16 \
  --onnx_path model.onnx --output_path model_int8_fp16.onnx
```

### Testing

- Added CPU-only unit tests in `tests/unit/onnx/test_onnx_utils.py`: stale rank-0 output reconciliation, preservation of a valid output shape, and the type-inference fallback on a `TopK`-overflow model. Run: `CUDA_VISIBLE_DEVICES="" pytest tests/unit/onnx/test_onnx_utils.py`.
- Verified end-to-end on a weakly-typed object-detection model that previously failed: the command now completes and produces a valid quantized FP16 model (`onnx.checker` passes, ORT loads it, QuantizeLinear/DequantizeLinear nodes inserted, FP16 initializers present, all graph-output ranks correct).
- Full `tests/unit/onnx` suite: no new failures versus the base branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mixed-precision/FP16 conversions to prevent stale or missing shape/type metadata. Output tensor rank/shape declarations are reconciled after clearing stale info, and stricter ONNX shape/type inference now falls back to standalone type inference when certain ops can’t be resolved.

* **Tests**
  * Expanded ONNX utility and autocast test coverage for fixing rank-0 output shapes, preserving valid static and dynamic `dim_param` output declarations, and verifying strict-mode fallback for unresolved `TopK` during FP16 conversion.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,10 @@ Changelog
   - ``hf_ptq.py`` also unwraps ``ModelOutput`` dataclasses from ``.generate()`` so the preview decode works on diffusion models. Non-tied models see no behavioral change.
 - Add **context-parallel (CP)** and **data-parallel (DP)** support to the shared Megatron-Core inference/calibration utilities. Under CP, ``get_megatron_calibration_forward_loop`` and ``megatron_mmlu`` partition each sequence across CP ranks (zigzag load-balanced), ``megatron_prefill`` accepts a CP-partitioned ``position_ids`` and lets the CP-aware causal attention build the mask, and MMLU gathers per-rank logits back to the full sequence for last-token scoring. Under DP, calibration shards the dataset across data-parallel ranks (``DistributedSampler``; amax is max-reduced across the DP group inside ``mtq``) and ``megatron_mmlu`` shards whole batches across DP ranks and all-reduces the per-subject counts. DP is implicit (``world_size / (tp * pp * cp)``); ``examples/megatron_bridge/quantize.py`` gains a ``--cp_size`` flag.
 
+**Bug Fixes**
+
+- Fix ``ShapeInferenceError`` during ONNX INT8 + FP16 quantization (``--high_precision_dtype fp16``) of weakly-typed models (e.g. TensorFlow exports) that carry stale rank-0 ``graph.output`` shapes or ops such as ``TopK`` that ONNX's static shape inference cannot resolve. ``clear_stale_value_info`` now reconciles stale output shapes via symbolic shape inference (keeping every output's shape field populated), and AutoCast runs ONNX shape inference in strict mode and falls back to schema-based standalone type inference when it fails, so unresolved ops no longer leave tensors untyped.
+
 0.45 (2026-07-02)
 ^^^^^^^^^^^^^^^^^
 
--- a/modelopt/onnx/utils.py
+++ b/modelopt/onnx/utils.py
@@ -1205,19 +1205,32 @@ def infer_types(
     When use_standalone_type_inference is True, uses a standalone type inference implementation
     that only infers types. Otherwise, uses ONNX's infer_shapes which infers both types and shapes.
 
+    ONNX's ``infer_shapes`` can fail on weakly-typed models -- with ``strict_mode=True`` it raises
+    on an op it cannot resolve (e.g. a ``TopK`` whose axis it resolves to a stale dimension)
+    instead of silently leaving that node's outputs untyped. On any shape-inference failure this
+    falls back to the standalone type inferencer, which derives types from operator schemas
+    regardless of shapes, so downstream type lookups (e.g. in AutoCast) do not fail. Callers that
+    need a fully typed graph should pass ``strict_mode=True`` so incomplete inference surfaces as
+    an exception that triggers the fallback.
+
     Args:
         model: ONNX model to infer types/shapes for.
         use_standalone_type_inference: If True, use standalone type inference (_infer_types_only).
                                        If False, use ONNX's shape inference (infer_shapes).
-        **kwargs: Additional arguments passed to infer_shapes when not using standalone type inference.
+        **kwargs: Additional arguments passed to infer_shapes when not using standalone type
+            inference (e.g. ``strict_mode``, ``check_type``, ``data_prop``).
 
     Returns:
         onnx.ModelProto: Model with inferred types (and shapes if not using standalone type inference).
     """
     if use_standalone_type_inference:
         return _infer_types_only(model)
-    else:
+
+    try:
         return infer_shapes(model, **kwargs)
+    except Exception as e:
+        logger.debug("ONNX shape inference failed (%s); using standalone type inference.", e)
+        return _infer_types_only(model)
 
 
 def onnx_type_str_to_enum(dtype: str) -> int:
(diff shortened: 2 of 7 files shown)


cjluo-nv (commented):

Targeted bug fix with three new tests. Two concerns worth a human look before merging:

1. **Dynamic-dim collateral damage in `_reconcile_stale_output_shapes`.** The helper runs unconditionally on every model that goes through `_preprocess_onnx` → `clear_stale_value_info` (not just weakly-typed exports), and it calls ORT's `SymbolicShapeInference.infer_shapes(model, auto_merge=True, guess_output_rank=True)`. With those flags, ORT routinely rewrites user-named `dim_param`s (e.g. `batch`, `seq_len`) into internal `unk__N` symbols, or commits a *guessed* rank when it can't fully propagate. The PR description says "re-inference reproduces their existing output shapes" and "healthy models are unaffected", but the bytes-level `_sig` comparison will treat a `batch` → `unk__0` rename as a *change* and overwrite the declared shape, which can break downstream consumers that look up output dims by name (export pipelines, TRT profile shapes, etc.). The added tests only cover fully static shapes (`[3,4]` / `[3,5]` / `[1,3]`), so this case is untested. Worth either (a) skipping the overwrite when only `dim_param` names differ but ranks/values agree, (b) only invoking the reconciler when the declared shape is actually inconsistent with the topology (the original motivation), or (c) at minimum adding a dynamic-batch regression test.

2. **`strict_mode=True` is now the default for `infer_types` callers in `autocast/convert.py`.** The fallback to `_infer_types_only` on any exception is a behavior broadening — previously-tolerated weakly-typed graphs now silently route through the standalone path instead. The standalone inferencer is a reasonable fallback, but this is a change in the AutoCast hot path with no new AutoCast-level test (the new `test_infer_types_falls_back_to_standalone_when_onnx_fails` exercises `infer_types` directly, not `convert_to_mixed_precision` / `convert_to_f16`). Worth confirming the existing AutoCast test suite still passes against a representative weakly-typed model and not just the synthetic TopK fixture.

Fix itself is well-scoped and the docstrings/comments are clear; flagging mainly so an ONNX-pipeline owner can sign off on the dynamic-shape behavior change.

ajrasane:
Thanks for the careful review — addressed both in the latest revision:

**1. Dynamic-dim collateral damage.** The reconciler now overwrites a declared output shape only when it's genuinely stale (a rank mismatch or a conflicting *concrete* dim); outputs that merely differ in symbolic `dim_param` names (e.g. a re-derived `unk__0` vs a declared `batch`) keep their declaration, so healthy models with dynamic dims are left untouched. I also dropped `guess_output_rank=True` — it's unnecessary for the fix (symbolic inference still derives the affected output ranks without it) and was the source of the guessed-rank risk you noted. Added `test_clear_stale_value_info_preserves_dynamic_dim_names` as a regression test (declares `["my_batch", 4]` where inference derives `["batch", 4]` and asserts the declared dim_param survives).

**2. `strict_mode=True` / AutoCast coverage.** Added `test_convert_to_f16_falls_back_on_unresolvable_op`, which exercises the full `convert_to_f16` path (used by INT8 + `--high_precision_dtype fp16`) on a weakly-typed model whose TopK defeats strict ONNX shape inference, asserting it converts via the standalone type-inference fallback rather than crashing. The existing AutoCast suite (`tests/unit/onnx/autocast/`) still passes, and the original weakly-typed model is verified end-to-end.

cjluo-nv (commented):

Both critical concerns from the previous review have been addressed:

1. **Dynamic-dim collateral damage** (cjluo-nv): `_is_stale` now only flags rank mismatches or conflicting *concrete* dims; symbolic `dim_param` renames are intentionally ignored. `guess_output_rank=True` was also dropped. The new `test_clear_stale_value_info_preserves_dynamic_dim_names` test declares `[my_batch, 4]` while inference would derive `[batch, 4]` and asserts `my_batch` is preserved — exactly the regression case raised before.

2. **`strict_mode=True` AutoCast hot-path**: The new `test_convert_to_f16_falls_back_on_unresolvable_op` exercises the full `convert_to_f16` entry point (not just `infer_types`) on a weakly-typed TopK model, confirming the standalone fallback fires end-to-end.

The local `SymbolicShapeInference` import inside `_reconcile_stale_output_shapes` is justified (optional onnxruntime dep with a graceful fallback to `infer_shapes`). Fix is well-scoped, no design-review concerns (additive fallback paths only, no new subsystem).

Complex PR: spans 5 directories (≥ 5). Looping in a human for approval.

modelopt/onnx/utils.py
```diff
@@ -1862,14 +1875,119 @@ def change_casts_to_fp16(model: onnx.ModelProto, target_op_types: list[str]) ->
     return model
 
 
+def _reconcile_stale_output_shapes(model: onnx.ModelProto) -> int:
+    """Re-derive stale ``graph.output`` shapes from the operator graph.
```
galagam, line 1879: @ajrasane There's similar logic in `PrecisionConverter::_clear_types_and_shapes_recursive`. In which we don't check what's stale and what's not, just clear everything since we're going to infer types (and optionally shapes) from the graph. 
Perhaps we can extract it to utils + add the fallback to standalone type inference on shape inference expcetion?

https://github.com/NVIDIA/Model-Optimizer/blob/cf2d65ade1ea891d3e150eefc52e59ef3f0071db/modelopt/onnx/autocast/precisionconverter.py#L287-L345

ajrasane: Thanks @galagam — good call, addressed in 73746724c.

**1. Extracted the clear logic to utils.** `_clear_types_and_shapes_recursive` is now a module-level `clear_types_and_shapes_recursive(graph, clear_shapes=...)` in `autocast/utils.py`, right next to `walk_subgraphs_recursive` (its only dependency). `PrecisionConverter` calls it with `clear_shapes=not self.use_standalone_type_inference`. I kept it in the autocast utils rather than the general `onnx/utils.py` because the latter can only reach `walk_subgraphs_recursive` through the lazy import it already does in `_infer_types_only` — moving it there would force the same circular-import dance.

**2. Fallback to standalone type inference on shape-inference exception** is centralized in `onnx_utils.infer_types` (the try/except in this PR). The AutoCast precision-conversion path already routes through it via `infer_types(..., strict_mode=True)`, so it gets the fallback for free.

One thing I deliberately did *not* fold into the wholesale clear: the output-shape reconciliation (`_reconcile_stale_output_shapes`). `clear_types_and_shapes_recursive` only renames concrete dims to `unk` and sets `elem_type=UNDEFINED` — it never changes a tensor's **rank**. The bug here is a stale **rank-0** output on a tensor that's really rank-2+, so clear-and-reinfer leaves the rank-0 in place and it re-poisons inference. We also can't just `ClearField("shape")` the output: a graph output with no shape fails `onnx.checker`, and the type-only fallback won't repopulate it. So reconciliation re-derives a present, correct-rank shape (ORT symbolic inference → `infer_shapes`) and overwrites only genuinely-stale declarations.

cjluo-nv (commented):

Re-review of the ONNX INT8+FP16 ShapeInferenceError fix (7 files, +338/-74). All critical concerns from the prior review have been addressed:

- :speech_balloon: **Dynamic-dim collateral damage (cjluo-nv, critical):** Resolved. `_is_stale` now overwrites a declared output shape only on a rank mismatch or a conflicting *concrete* dim; symbolic `dim_param` renames are intentionally ignored, and `guess_output_rank=True` was dropped. New `test_clear_stale_value_info_preserves_dynamic_dim_names` declares `["my_batch", 4]` vs inferred `["batch", 4]` and asserts the declaration survives — exactly the regression case raised.
- :speech_balloon: **strict_mode=True AutoCast hot path (cjluo-nv, critical):** Resolved. `test_convert_to_f16_falls_back_on_unresolvable_op` exercises the full `convert_to_f16` entry point on a weakly-typed TopK model, confirming the standalone fallback fires end-to-end (not just `infer_types` in isolation).
- :speech_balloon: **Duplicate clear logic (galagam, important):** Resolved. `_clear_types_and_shapes_recursive` was extracted to a module-level `clear_types_and_shapes_recursive(graph, clear_shapes=...)` in `autocast/utils.py`; `PrecisionConverter` now calls it with `clear_shapes=not self.use_standalone_type_inference`, and the shape-inference fallback is centralized in `infer_types`. Author gave a clear rationale for keeping `_reconcile_stale_output_shapes` separate (clear-and-reinfer never repairs a stale rank). I confirmed the old method definition and call site are gone in the PR branch and replaced by the shared helper.

Remaining minor item: the coderabbit nitpick to assert strict ONNX inference fails *before* `convert_to_f16` (making the fallback precondition explicit) was not adopted; the author's success-via-fallback assertion is a reasonable alternative.

Flagging for human sign-off because this changes a shared AutoCast code path (strict_mode default + new output-shape reconciler running on every model through `clear_stale_value_info`, not just weakly-typed exports) and the end-to-end / object-detection validation was done manually on a private model not in CI. New tests are CPU-only and synthetic; an ONNX-pipeline owner should confirm the existing AutoCast suite + a representative dynamic-shape/real model still behave correctly.

cjluo-nv (commented):

Re-review of the ONNX INT8+FP16 ShapeInferenceError fix (7 files, +318/-74). All prior critical comments are resolved; flagging for human ONNX-pipeline sign-off because this touches a shared AutoCast hot path and E2E validation was manual/private.

**Prior comments — status:**
- :speech_balloon: **Dynamic-dim collateral damage (cjluo-nv, critical):** Resolved. `_is_stale` now overwrites a declared output shape only on a rank mismatch or a conflicting *concrete* dim; symbolic `dim_param` renames are ignored and `guess_output_rank=True` was dropped. New `test_clear_stale_value_info_preserves_dynamic_dim_names` declares `["my_batch", 4]` vs inferred `["batch", 4]` and asserts the declaration survives — exactly the regression raised. Still warrants human sign-off because the reconciler runs on *every* model through `clear_stale_value_info`, not just weakly-typed exports, and CI coverage is synthetic CPU-only.
- :speech_balloon: **strict_mode=True AutoCast hot path (cjluo-nv, critical):** Resolved. `test_convert_to_f16_falls_back_on_unresolvable_op` exercises the full `convert_to_f16` entry point on a weakly-typed TopK model, confirming the standalone fallback fires end-to-end — still flagging because the strict_mode default broadening reroutes previously-tolerated graphs and only a synthetic fixture exercises it.
- :speech_balloon: **Duplicate clear logic (galagam, important):** Resolved. `_clear_types_and_shapes_recursive` was extracted to module-level `clear_types_and_shapes_recursive(graph, clear_shapes=...)` in `autocast/utils.py`; old method and call site removed (verified). Author gave a clear rationale for keeping `_reconcile_stale_output_shapes` separate (clear-and-reinfer never repairs a stale *rank*).
- **CodeRabbit nitpick (assert strict inference fails before convert):** Minor, not adopted; success-via-fallback assertion is a reasonable equivalent.

**Verified — not a regression (re CodeRabbit's "clear_shapes ignored for main-graph value_info"):** I diffed the extracted helper against the original. The main-graph `value_info` branch cleared shapes *unconditionally* in the original too; the author preserved that with `clear_shape=True`. CodeRabbit's suggested patch would actually *change* behavior. The only wart is that the docstring's `clear_shapes` contract now reads as if it gates all shape clearing when it doesn't for main-graph value_info — non-blocking.

**Answer to "how to simplify while keeping functionality":** A few options, all optional: (1) `_reconcile_stale_output_shapes` does declared→clear→infer→compare with `_sig` bytes-comparison and a separate `_is_stale`; since `_is_stale` already encodes the overwrite policy, the trailing `_sig(decl) != _sig(new_shape)` change-count could be derived directly from whether `_is_stale` fired, dropping the `_sig` helper. (2) The two inference attempts (SymbolicShapeInference then `infer_shapes`) duplicate the output-shape extraction dict-comprehension — extract a small `_outputs_with_shapes(model)` helper. (3) The reconciler currently runs on every model; gating it on "only when at least one declared output has a rank that conflicts with topology" (the original motivation) would both simplify the common-case path and shrink blast radius — but that's a behavior trade-off the owner should weigh, not a free simplification.

No licensing concerns (standard NVIDIA header). No prompt-injection in the PR content.

CHANGELOG.rst
```diff
@@ -73,37 +16,60 @@ Changelog
...
+- Refactor ``llm_qat`` example with unified YAML-based configuration and flexible dataset blending.
+  ``ModelOptArgParser`` adds ``--config`` YAML support with CLI overrides and auto-generates ``ARGUMENTS.md`` from dataclass definitions.
+  Dataset blending (``configs/dataset/blend.yaml``) supports HuggingFace datasets, local JSON/JSONL/Parquet files, and weighted multi-source blends.
+  The legacy FSDP1 accelerate config is removed; ``llm_qat`` now documents FSDP2, DeepSpeed, and DDP backends.
 
 **Bug Fixes**
 
+- Fix ``ShapeInferenceError`` during ONNX INT8 + FP16 quantization (``--high_precision_dtype fp16``) of weakly-typed models (e.g. TensorFlow exports) that carry stale rank-0 ``graph.output`` shapes or ops such as ``TopK`` that ONNX's static shape inference cannot resolve. ``clear_stale_value_info`` now reconciles stale output shapes via symbolic shape inference (keeping every output's shape field populated), and AutoCast runs ONNX shape inference in strict mode and falls back to schema-based standalone type inference when it fails, so unresolved ops no longer leave tensors untyped.
```
kevalmorabia97: goes in 0.46 section

ajrasane:
Follow-up on simplifying `_reconcile_stale_output_shapes` — three options were considered:

**(1) Drop the `_sig` byte-comparison helper — done (`80c4ed0e0`).** `_is_stale` already encodes the overwrite policy, so the change-count is now derived directly from whether `_is_stale` fired; the trailing `_sig(decl) != _sig(new_shape)` comparison and the `_sig` helper are gone.

**(2) Extract a shared `_outputs_with_shapes(model)` helper — done (`80c4ed0e0`).** The two inference attempts (`SymbolicShapeInference` then `infer_shapes`) no longer duplicate the output-shape extraction dict-comprehension.

**(3) Gate the reconciler on "only when a declared output rank conflicts with topology" — not adopted.** This is a behavior trade-off, not a free simplification:

- It's circular: deciding whether a declared rank conflicts with the topology requires deriving the topology rank, which is exactly the shape inference the gate would skip. The only inference-free proxy — "some output is declared rank-0" — narrows the fix to the rank-0 symptom and would silently miss other stale-rank cases (e.g. a declared rank-3 output that is really rank-2), reintroducing the same `ShapeInferenceError`.
- It adds a predicate rather than removing one, and the path is cold (once per quantize, once per reference-output run), so the common-case win is negligible.

> 🤖 _Generated by Claude (AI agent)._

galagam (approved):
Looks good, thanks for addressing the comments.

PR #1628, opened 2026-06-04
[6058841] Fix inconsistent tensor types on control-flow (If/Loop/Scan) subgraphs during FP16/BF16 conversion

ajrasane:
### What does this PR do?

Type of change: Bug fix

Fixes [6058841] — `python -m modelopt.onnx.quantization --high_precision_dtype fp16` crashed with *"Inconsistent type on If node"* on models containing control-flow `If`/`Loop`/`Scan` subgraphs.

**Root cause.** The FP16/BF16 `PrecisionConverter` blindly converted *every* subgraph initializer to the parent control-flow node's precision, without the activation-bracketing casts it uses in the main graph. This left inconsistent tensor types that crash ONNX shape inference / TensorRT strongly-typed parsing:

- A `Gemm` inside an `If` branch reading an outer-scope activation (fp32) ended up with fp16 weights → `B has inconsistent type tensor(float16)`.
- `Resize` `scales` (which must stay fp32 per the ONNX spec) was converted to fp16 → `ParseData type mismatch ... Expected:float Actual:float16`.

**Fix.**
- A subgraph node is converted to low precision **only when all of its float inputs are subgraph initializers** eligible for low precision. Any node consuming a float activation / outer-scope tensor, or an input that must stay high precision (e.g. `Resize` `scales`), stays high precision — so each node's inputs share a single precision.
- Float **outer-scope captures** and **low→high precision boundaries** inside subgraphs are reconciled with `Cast` nodes; a captured tensor's preserved subgraph `value_info` is synced to its real main-graph precision; control-flow node **outputs** are treated as high precision (a low-precision `If` branch may still convert eligible nodes *inside* its body, but the parent control-flow node's outputs stay high precision).
- `Constant`→`Cast` folding in `remove_redundant_casts` now refreshes the constant's `value_info` so a same-type-constrained consumer (e.g. `Greater`) isn't left with a stale, conflicting type. (Pre-existing main-graph bug surfaced once the `If` models completed conversion.)

**Known limitation.** A low-precision `Loop`/`Scan` whose body carries a **float loop-carried or scan-input state variable** is not yet reconciled here — its body's formal inputs are treated as high precision while the main graph casts the parent's inputs to low precision. This PR fully covers `If` branches (both precisions) and high-precision `Loop`/`Scan` bodies (outer-scope captures reconciled); the low-precision `Loop`/`Scan` body case is tracked as a follow-up.

### Usage

```bash
  --quantize_mode int8 --high_precision_dtype fp16 \
  --onnx_path model.onnx \
  --output_path model_strongType_int8+fp16.onnx
```

### Testing

Validated on both reported models:

| graph pattern | convert | strict `infer_shapes(check_type=True)` | ORT load | numerics vs FP32 |
|---|---|---|---|---|
| `If` branch with `Gemm` reading an outer-scope input | ✅ | ✅ | ✅ | bit-exact (Gemms kept fp32) |
| `If` branch with `Resize` (`scales` initializer) | ✅ | ✅ | ✅ | max abs err 9e-5 (fp16 tol) |

Also verified with `keep_io_types=False`. Added 5 regression tests in `tests/unit/onnx/autocast/test_precisionconverter.py` — `If` Gemm-with-outer-scope-input, `If` Resize-scales-stay-FP32, chained-`If` capture, high-precision `Loop`-body capture, and `Constant`→`Cast` fold `value_info` refresh — that fail without the fix and pass with it. Full `tests/unit/onnx/autocast/` suite passes (214). `pre-commit` clean.

- Did you get Claude approval on this PR?: ❌ (draft)

### Additional Information

Fixes bug [6058841]. Draft pending final review.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed FP16/BF16 autocast precision conversion for ONNX control-flow subgraphs (such as `If`/`Loop`), especially with `--high_precision_dtype fp16`.
  - Ensured branch weights that rely on outer-scope FP32 activations remain FP32 to avoid mixed-precision mismatches.
  - Improved precision reconciliation inside subgraphs via added `Cast` nodes, and refreshed graph type metadata after `Constant`→`Cast` folding (including `Resize` scale handling).

- **Tests**
  - Added regression coverage for nested/chained control-flow, `Resize` FP32 scale preservation, strict type/shape inference, and `Constant`→`Cast` folding scenarios.


modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -761,45 +773,263 @@ def _convert_initializers_recursive(
...
+        # producer (captures, produced outside the subgraph, are placed at the front).
+        cast_nodes_after_producer: dict[str, list[onnx.NodeProto]] = defaultdict(list)
+        leading_casts: list[onnx.NodeProto] = []
+        for (input_name, desired), (cast_output, producer) in casts_to_insert.items():
+            cast_node = helper.make_node(
+                "Cast", inputs=[input_name], outputs=[cast_output], to=desired, name=cast_output
+            )
+            if producer is not None:
+                producer_node_name = next(n.name for n in subgraph.node if input_name in n.output)
+                cast_nodes_after_producer[producer_node_name].append(cast_node)
```
cjluo-nv: > _Bot comment._

Cheap perf cleanup: this `next(...)` re-scans `subgraph.node` once per inserted cast. You already have `local_produced` above and the new node list iteration; build a `producer_by_output: dict[str, str]` once before this loop (or include the producer node name when populating `casts_to_insert`) to avoid the O(casts × nodes) scan.

modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -761,45 +773,263 @@ def _convert_initializers_recursive(
...
+            else:
+                leading_casts.append(cast_node)
+
+        new_nodes = list(leading_casts)
+        for node in subgraph.node:
+            new_nodes.append(node)
```
cjluo-nv: > _Bot comment._

Consider asserting / enforcing topological order before reassembling. The cast is placed right after its producer, but a downstream consumer of this cast that happens to appear *earlier* in `subgraph.node` (possible if the subgraph was hand-built and not toposorted) would now see the rewritten input name without the Cast having been emitted yet. A `gs.import_onnx`/`toposort` pass on the rebuilt subgraph (or an explicit assertion that producer indices < consumer indices) would harden this.

modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -761,45 +773,263 @@ def _convert_initializers_recursive(
...
+            out for node in subgraph.node if node_is_low.get(node.name) for out in node.output
+        }
+
+        def _current_float_type(input_name: str) -> int | None:
+            """Float precision (ONNX type) of a subgraph input tensor, or None if it is non-float.
+
+            Non-float tensors (int indices/axes/shapes, bool conditions) must never be cast.
+            """
+            if input_name in local_produced:
+                elem_type = local_elem_type.get(input_name)
```
cjluo-nv, line 1066: > _Bot comment._

`_current_float_type` returns `high.onnx_type` unconditionally for any float `formal_inputs`. For Loop body state vars / Scan scan_inputs, the actual precision is whatever the main graph hands the parent Loop/Scan node — which can be low precision. The new node-classification rule mostly papers over this (a float formal input demotes the node to high precision), but it would be good to add at least one Loop-body regression test, since none of the new tests exercise Loop/Scan.

modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -761,45 +773,263 @@ def _convert_initializers_recursive(
...
+            elif name in self.value_info_map:
+                base_type = self.value_info_map[name].type.tensor_type.elem_type
+            else:
+                return None
+            if base_type not in ONNX_TYPES:
+                return None
+            return main_producer_precision.get(name, base_type)
 
         def _convert_subgraph_callback(
             graph: onnx.GraphProto, parent: onnx.NodeProto, is_subgraph: bool
```
cjluo-nv: > _Bot comment._

Worth calling out in the docstring (and ideally CHANGELOG) that this is a behavioral change for subgraph initializer precision: previously a low-precision parent caused *all* subgraph float initializers to be converted, now conversion is gated on the consuming node having no float activation/outer-scope/unknown-typed inputs. Anyone relying on the old behavior to get fp16 weights inside an If branch with an outer-scope activation will see those weights stay fp32 after this PR (which is the correct fix, but the change in observable output deserves a note).

modelopt/onnx/utils.py
```diff
@@ -1558,6 +1558,14 @@ def remove_redundant_casts(onnx_model: onnx.ModelProto) -> onnx.ModelProto:
                 assert len(cast_producers) == 1 and cast_producers[0].op_type == "Constant"
                 constant_producer = cast_producers[0]
                 _convert_constant_values(constant_producer, node)
+                # Folding changes the Constant output's element type; refresh its value_info so
+                # downstream consumers (and strongly-typed parsers) don't see a stale type that
+                # conflicts with the now-converted constant value.
+                cast_to_type = get_cast_to_type(node)
+                for vi in onnx_model.graph.value_info:
+                    if vi.name == constant_producer.output[0]:
```
cjluo-nv: > _Bot comment._

This only updates the first `value_info` entry that matches and silently does nothing if no entry exists. Constants frequently lack a `value_info` entry until type inference runs. Fine for the regression test (which immediately re-runs strict shape inference), but if a caller of `remove_redundant_casts` doesn't follow up with type inference, downstream consumers may still see no/stale type for the folded constant. Consider creating a `value_info` entry when one is missing, or document that callers must re-infer types after folding.

cjluo-nv (commented):

Solid, well-scoped bug fix for the "Inconsistent type on If node" crash, with 5 new regression tests (Gemm-with-outer-input, Resize-scales, chained-If capture, Constant→Cast value_info refresh) parameterized over fp16/bf16/keep_io_types/standalone-type-inference, and the existing subgraph tests still align with the new logic. Comments are clear and the policy is documented.

Worth a human reviewer's eye:

- **Semantic shift in subgraph init conversion.** Previously, when an If/Loop/Scan parent was classified low-precision, *every* float subgraph initializer was converted. The new rule (`_convert_subgraph_precision`) only converts an initializer if its consuming node has at least one low-precision-eligible init *and* every other input is provably non-float. Any unknown-type input (e.g. tensor produced by an op the type-info map doesn't cover) demotes the node to high precision. For real models this is the desired behavior, but it's a public-facing behavioral change for any downstream user who depended on the old "blindly inherit parent precision" semantics — please confirm this is intended and call out in the changelog if so.

- **`Loop`/`Scan` formal-input precision.** `_current_float_type` returns `high.onnx_type` for any float `formal_inputs` (subgraph inputs). For `Loop` body state vars or `Scan` scan_inputs, the actual precision is whatever the main-graph passes in — which could be low precision if the parent Loop is low. In practice the new node classifier will keep such bodies in high precision (formal float inputs prevent low classification), but this assumption isn't asserted anywhere and there's no Loop/Scan regression test in the suite. Worth a quick sanity check on a Loop-body model.

- **Topological assumption when assembling casts.** In step 3 the code does `next(n.name for n in subgraph.node if input_name in n.output)` to find a producer, then places the cast immediately after that producer when reassembling `subgraph.node`. This relies on the original subgraph being topologically sorted (the cast appearing *after* its producer in the new list also requires the producer to come before all current consumers in iteration order). Subgraphs from `helper.make_graph` and most exporters are sorted, but if a user feeds an unsorted subgraph the cast can land before a consumer it's supposed to feed. Cheap fix: do a `gs.toposort` (already done elsewhere in this file) on the rebuilt node list, or assert sortedness.

- **O(N²) producer lookup.** The `next(n.name for n in subgraph.node if input_name in n.output)` runs per cast and re-scans `subgraph.node` each time. Fine for the bug-report models, but a single producer→name dict built once before the loop would be cleaner.

- **Constant→Cast `value_info` refresh in `utils.remove_redundant_casts`.** The new loop only updates the *first* matching `value_info` entry and only if one already exists — Constants frequently have no `value_info` entry at all, in which case nothing is refreshed. That's fine for the bug at hand (the test verifies strict shape inference passes), but worth confirming no consumer relies on a now-missing/already-typed `value_info` for `c0`.

No licensing changes; PR size (~500 LoC, half tests) is reasonable and the changes are cohesive.

CHANGELOG.rst
```diff
@@ -53,7 +53,7 @@ Changelog
...
-- Fix ONNX AutoCast ``keep_io_types=True`` sanity-check failure (``Unexpected type in I/O tensor ...``) when a network input/output is an empty tensor (a dimension of size 0). Such tensors were "fake-cast" (retyped in place) to the low precision type; because the value-info map aliases the ``graph.input``/``graph.output`` ``ValueInfoProto``, this silently changed the model's I/O type. AutoCast now inserts a real ``Cast`` for protected I/O tensors instead.
+- Fix ONNX FP16/BF16 conversion (``--high_precision_dtype fp16``) producing inconsistent tensor types on models with control-flow ``If``/``Loop``/``Scan`` subgraphs (e.g. a ``Gemm`` reading an outer-scope activation alongside converted weights, or ``Resize`` ``scales`` that must stay FP32). Subgraph nodes now only run in low precision when all their float inputs are subgraph initializers; outer-scope captures and low-to-high-precision boundaries inside subgraphs are reconciled with ``Cast`` nodes, and ``Constant`` folding refreshes the constant's ``value_info`` so strongly-typed parsers (TensorRT) no longer reject the model.
```
kevalmorabia97: goes in 0.46 section

modelopt/onnx/autocast/precisionconverter.py
ajrasane: Done — the per-cast `next(... in n.output)` scan is gone. Casts are collected into a `casts_to_insert` dict keyed by `(input_name, target_type)`, and the rewritten subgraph is rebuilt in a single pass via `_topologically_sorted_nodes([*subgraph.node, *cast_nodes])`, so `subgraph.node` is no longer re-scanned once per inserted cast.

> 🤖 _Generated by Claude (AI agent)._

modelopt/onnx/autocast/precisionconverter.py
ajrasane: Addressed. Instead of placing each cast right after a located producer, the existing nodes plus the new `Cast` nodes are passed through `_topologically_sorted_nodes(...)`, which orders purely by data dependencies and makes no assumption about the incoming node order. A hand-built / unsorted subgraph is therefore handled correctly — every Cast lands after its producer and before its consumers.

> 🤖 _Generated by Claude (AI agent)._

modelopt/onnx/autocast/precisionconverter.py
ajrasane: Acknowledged — this is a known, pre-existing limitation that is deliberately scoped out of this PR. `_current_float_type` intentionally treats float formal inputs as high precision, so the node classifier keeps such bodies in high precision; a low-precision `Loop`/`Scan` body that carries a float loop-carried / scan-input state var is not reconciled here. It is documented as a "Known limitation" in the `_convert_initializers_recursive` docstring and left to a follow-up, since a correct fix needs op-specific formal-input precision propagation plus honoring the loop-carried `output == input` constraint, which carries regression risk for the well-tested `If` path. This PR does add a high-precision `Loop`-body capture regression test (`test_loop_subgraph_high_precision_capture`); the low-precision Loop/Scan case is the deferred work.

> 🤖 _Generated by Claude (AI agent)._

modelopt/onnx/autocast/precisionconverter.py
ajrasane: Confirmed intentional, and documented in both places. The behavioral change is called out in the `_convert_initializers_recursive` docstring ("This is a behavioral change: previously a low-precision control-flow parent converted *every* float subgraph initializer...") and in the CHANGELOG bug-fix entry.

> 🤖 _Generated by Claude (AI agent)._

modelopt/onnx/utils.py
ajrasane: Done — `remove_redundant_casts` now creates a `value_info` entry when none exists. For a folded `Constant`, if no matching `value_info` is found it builds one via `make_tensor_value_info(const_output, cast_to_type, dims)` from the constant's value tensor, so downstream consumers and strongly-typed parsers see the folded type even without a follow-up inference pass. The existing path still updates an entry in place when one is already present.

> 🤖 _Generated by Claude (AI agent)._

CHANGELOG.rst
ajrasane: Done — moved to the 0.46 section.

> 🤖 _Generated by Claude (AI agent)._

tests/unit/onnx/autocast/test_precisionconverter.py
```diff
@@ -1779,3 +1709,237 @@ def test_if_subgraph_outer_scope_type_preservation(
...
+        name="if_node",
+        then_branch=_branch("then"),
+        else_branch=_branch("else"),
+    )
+    main_graph = helper.make_graph([if_node], "model_if_gemm", [x, condition], [y])
+    model = helper.make_model(main_graph, producer_name="model_if_gemm")
+    model.opset_import[0].version = 20
+    model.ir_version = 10
+    onnx.checker.check_model(model)
+    return setup_mappings(model)
```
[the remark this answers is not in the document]
ajrasane: Good point in principle, but this is a false positive for the control-flow-capture models these fixtures build.

The standalone type inferencer (`infer_types_for_graph` in `modelopt/onnx/utils.py`) types each subgraph from a temp model containing *only* that subgraph, seeding tensor types from the subgraph's own inputs, initializers, and `value_info`. It has no mechanism to import the types of outer-scope captures (e.g. `X`, `mid`, `pre`) unless they are already present in the subgraph's `value_info`. Building the fixtures' mappings via the scope-aware ONNX `infer_shapes` path — `setup_mappings(model)` with the default — is what populates that `value_info`, so it is load-bearing here rather than an oversight.

If the fixtures instead build mappings with `use_standalone_type_inference=True`, that bootstrap is gone and even the converter's own internal re-inference (`PrecisionConverter` calls `infer_types(self.model, self.use_standalone_type_inference, ...)`) then fails with `ValueError: Input X of node else_gemm has unknown type` — verified across all four subgraph fixtures (gemm / resize / chained-if / loop).

The parametrized `use_standalone_type_inference` is still meaningfully exercised: the flag flows into `PrecisionConverter`, which re-runs type inference with it on top of the bootstrapped model, so the converter's standalone branch is covered. The standalone inferencer's lack of outer-scope-capture support inside subgraphs is a separate, pre-existing limitation outside this PR's scope.

Resolving as not-applicable for these models.

> 🤖 _Generated by Claude (AI agent)._

cjluo-nv (commented):

Re-review of PR #1628 (ONNX FP16/BF16 control-flow subgraph inconsistent-type fix, +616/-29, 4 files). All prior comments appear addressed:

**Critical/correctness — resolved:**
- :speech_balloon: Topological-order assumption (precisionconverter.py:1029): *cast placement no longer relies on incoming node order* — now rebuilt via new `_topologically_sorted_nodes` (Kahn's algorithm, orders purely by data deps, appends cycle-blocked nodes in original order). Reviewed the implementation; the unique-producer set + `producer_idx.get(inp, i) != i` self/non-produced filtering is correct.
- :speech_balloon: CHANGELOG placement (kevalmorabia97): *moved to 0.46 Bug Fixes* — confirmed in current file.
- :speech_balloon: Constant→Cast value_info refresh (utils.py:1566): *now creates a value_info entry via `make_tensor_value_info` when none exists*, in addition to updating in place — confirmed.

**Minor — resolved:**
- O(N²) producer scan eliminated (casts collected into a dict, single re-sort pass).
- Behavioral-change to subgraph init conversion documented in `_convert_initializers_recursive` docstring + CHANGELOG.
- CodeRabbit's `use_standalone_type_inference` fixture concern was explicitly withdrawn by the bot after the author's justification (standalone inferencer can't seed outer-scope-capture types).

**Why nudge, not approve:**
1. This is a public-facing **behavioral change** — a low-precision control-flow parent no longer converts *every* subgraph float initializer; weights alongside an outer-scope fp32 activation now stay fp32. Correct fix, but observable-output change deserves owner confirmation.
2. :speech_balloon: Loop/Scan formal-input precision (line 949): author acknowledges a **deliberately scoped-out known limitation** — a low-precision `Loop`/`Scan` body carrying a float loop-carried/scan-input state var is not reconciled (only a *high*-precision Loop-body capture test exists). Warrants human sign-off that deferring this is acceptable.
3. Validation was on the two reported models (GPU/model-derived, not in CI); the 5 new unit regression tests are meaningful and parametrized over fp16/bf16/keep_io_types/standalone-inference and pass strict `check_type=True`, but the real-model paths aren't CI-covered.

No licensing changes. No prompt-injection issues in the untrusted content. Recommend owner sign-off on the behavioral change + deferred Loop/Scan case.

modelopt/onnx/utils.py
```diff
@@ -1571,6 +1571,32 @@ def remove_redundant_casts(onnx_model: onnx.ModelProto) -> onnx.ModelProto:
...
+                            if attr.name == "value" and attr.type == onnx.AttributeProto.TENSOR
+                        ),
+                        None,
+                    )
+                    if value_tensor is not None:
+                        onnx_model.graph.value_info.append(
+                            onnx.helper.make_tensor_value_info(
+                                const_output, cast_to_type, list(value_tensor.dims)
+                            )
+                        )
```
galagam: This should be handled within `_convert_constant_values` so calling it leaves the graph in a valid state.

modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -65,13 +65,48 @@ class InitializerConsumerTracker:
...
+ONNX_TYPE_TO_PRECISION = {t.onnx_type: t for t in PRECISION_MAP.values()}
+
 OP_TYPES_NOT_SUPPORTED_IN_LOW_PRECISION = ["Upsample", "NonMaxSuppression", "Celu"]
 
 # Mapping of op types to indices of inputs that should not be converted to low precision.
 SKIP_LOW_PRECISION_MAPPING_FP16 = {"Resize": {2}}
 SKIP_LOW_PRECISION_MAPPING_BF16 = {"Resize": {1, 2}}
 
 
+def _topologically_sorted_nodes(nodes: list[onnx.NodeProto]) -> list[onnx.NodeProto]:
```
galagam:
```suggestion
def _topologically_sorted_nodes(model: ModelProto) -> list[onnx.NodeProto]:
    g = Graph(model.graph)
    g.toposort() 

    nodes = list(g.nodes)  # or model.graph.node after sort

    if len(ordered) < len(nodes):  # dependency cycle (invalid ONNX): keep remaining nodes as-is
        seen = {id(node) for node in ordered}
        ordered.extend(node for node in nodes if id(node) not in seen)
```

1. Do we really have to implement topological sort ourselves? Would this work? If not - why not?
2. Why do we care about handling dependency cycles? Is there a valid use case for this?

galagam (commented):
@ajrasane Thanks for addressing this, it has been a long-standing weakness of AutoCast. PTAL at my comments, I'll try to continue review later.

modelopt/onnx/utils.py
ajrasane: Addressed. `_convert_constant_values` now receives the model and updates or creates the folded Constant output's `value_info` itself, so calling the helper leaves both the tensor payload and graph metadata consistent.

modelopt/onnx/autocast/precisionconverter.py
ajrasane: Addressed using the shared `onnx_utils.topologically_sort_graph_nodes` helper now available on `main`.

GraphSurgeon import/export is avoided because it can reinterpret ONNX enum dtypes such as BF16 while converting tensor metadata. The shared helper operates directly on `GraphProto`, includes nested subgraph capture dependencies, and raises on duplicate producers, self-dependencies, or cycles instead of preserving an invalid graph.

galagam (approved):

modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -706,42 +726,251 @@ def _convert_initializers_recursive(
...
+            """
+            if name in self.initializer_map:
+                base_type = self.initializer_map[name].data_type
+            elif name in self.value_info_map:
+                base_type = self.value_info_map[name].type.tensor_type.elem_type
+            else:
+                return None
+            if base_type not in ONNX_TYPES:
+                return None
+            return main_producer_precision.get(name, base_type)
```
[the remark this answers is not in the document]
ajrasane: Addressed. Precision conversion now carries a scope map containing the actual float precision of inputs, initializers, and node outputs into each nested control-flow body. Nested bodies are converted recursively before their enclosing protobuf node list is re-sorted, so captures from an outer body are reconciled correctly.

Added `test_nested_if_capture_uses_enclosing_scope_precision` for FP16 and BF16; it verifies the inner branches cast the enclosing low-precision tensor back to FP32 and pass strict ONNX type inference.

modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -706,42 +726,251 @@ def _convert_initializers_recursive(
...
+            assert from_type is not None
+            low_consumers: list[InputIndexTracker] = []
+            high_consumers: list[InputIndexTracker] = []
+            for c in consumers.get(init.name, []):
+                if node_is_low.get(c.node.name) and _is_low_precision_eligible_init(
+                    c.node, init.name
+                ):
+                    low_consumers.append(c)
+                else:
+                    high_consumers.append(c)
```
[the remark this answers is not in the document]
ajrasane: Addressed. `node_is_low` is now keyed by `id(node)`, and initializer-consumer classification plus produced-tensor precision lookup use the same identity-based key.

Added `test_subgraph_node_precision_uses_identity` with two unnamed branch nodes requiring different precisions; it covers both FP16 and BF16 and passes strict ONNX type inference.

PR #1630, opened 2026-06-04
[minor] fix for GLM4.7 mtp module in PTQ

Fridah-nv:
### What does this PR do?

Type of change: Bug fix 

  `_keys_to_prefixes` now drops a top-level `"model"` key fragment instead of emitting it as a prefix. Without this guard, an inlined-MTP key like `model.layers.92.eh_proj.weight`  would emit `"model"` → exporter wraps it as `"model*"` in `quantization_config.exclude_modules` → `fnmatch` in TRT-LLM matches every `model.layers.X.*` module → entire backbone is treated as unquantized → FP8 weights get loaded into BF16 buffers via `.view(bf16)` (which halves the last dim of an FP8 tensor) → loader crashes with `tensor a (5120) must match tensor b (2560) at non-singleton dimension 1`.

  The main-branch caller in `load_mtp_weights` (PR #1532) already filters inlined keys before invoking `_keys_to_prefixes`, so this guard is defense-in-depth on `main` today. It freezes the invariant in code rather than as a docstring caveat — the same regression cannot return if a future caller forgets to filter. Adds a focused unit test that pins the behavior.

### Usage

```python
```

### Testing
 - New unit test `tests/examples/llm_ptq/test_example_utils.py::test_keys_to_prefixes_drops_model_top_level`: passes with this PR, would fail without the guard. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed prefix handling in LLM post-training quantization examples: the top-level "model" prefix is now omitted when deriving exclusion prefixes to avoid overly broad wildcard exclusions.

* **Tests**
  * Added test coverage verifying correct conversion of tensor keys to exclusion prefixes, including cases with top-level "model" keys.

### DIFF
--- a/examples/llm_ptq/example_utils.py
+++ b/examples/llm_ptq/example_utils.py
@@ -312,13 +312,13 @@ def get_inlined_mtp_prefixes(config: Any) -> list[str]:
 def _keys_to_prefixes(keys: Iterable[str]) -> set[str]:
     """Invert separate-file MTP keys into the prefixes the exporter needs for exclude_modules.
     ``"mtp.fc.weight"`` → ``{"mtp"}``; ``"mtp.layers.0.q_proj.weight"`` →
-    ``{"mtp", "mtp.layers.0"}``. Caller must filter out inlined keys; otherwise
-    ``"model.layers.78.eh_proj.weight"`` would emit ``"model"`` as a prefix.
+    ``{"mtp", "mtp.layers.0"}``. ``"model"`` top-level is dropped to avoid the
+    ``"model*"`` wildcard covering the whole backbone (nvbug 6108133).
     """
     prefixes: set[str] = set()
     for key in keys:
         parts = key.split(".")
-        if parts:
+        if parts and parts[0] != "model":
             prefixes.add(parts[0])
         for i, part in enumerate(parts):
             if part == "layers" and i + 1 < len(parts) and parts[i + 1].isdigit():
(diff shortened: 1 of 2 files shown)


examples/llm_ptq/example_utils.py
```diff
@@ -312,13 +312,13 @@ def get_inlined_mtp_prefixes(config: Any) -> list[str]:
 def _keys_to_prefixes(keys: Iterable[str]) -> set[str]:
     """Invert separate-file MTP keys into the prefixes the exporter needs for exclude_modules.
     ``"mtp.fc.weight"`` → ``{"mtp"}``; ``"mtp.layers.0.q_proj.weight"`` →
-    ``{"mtp", "mtp.layers.0"}``. Caller must filter out inlined keys; otherwise
-    ``"model.layers.78.eh_proj.weight"`` would emit ``"model"`` as a prefix.
+    ``{"mtp", "mtp.layers.0"}``. ``"model"`` top-level is dropped to avoid the
+    ``"model*"`` wildcard covering the whole backbone (nvbug 6108133).
```
Edwardf0t1: Do we want to include internal nvbug numbers in public repo?

cc @kevalmorabia97

Edwardf0t1 (approved):
LGTM.

examples/llm_ptq/example_utils.py
Fridah-nv: I included it for future reference, I've seen TRTLLM pasted bug links in comments. Happy to remove it if any concerns.

Fridah-nv: Removed the bug ID and merged, thanks for your review!

PR #1631, opened 2026-06-04
fix(export): correct unified_export_megatron at EP > 1 and DP > 1

yueshen2016:
## Summary

Two related bugs surface when exporting a Megatron-Core Mamba-MoE checkpoint (e.g., Nemotron 3 Ultra) to HuggingFace with non-trivial expert and data parallelism.

### 1. TEGroupedMLP expert-id collision at EP > 1

`_grouped_mlp_slicing` iterates `range(num_experts)`, but `num_experts == module.num_gemms` is the *local* expert count on each EP rank. Using local ids `0..N-1` for the saved HF key prefix means every EP rank emits the same key set (`experts.0..N-1.*`), and the last writer wins for each layer — at EP=8 this silently loses **7/8 of every MoE layer's experts**.

The fix:

- Reads the authoritative local-to-global mapping from `module.local_expert_indices` (or `module.experts.local_expert_indices`) when exposed; falls back to the standard Megatron contiguous layout `[ep_rank*N, (ep_rank+1)*N − 1]`.
- Saves each expert under its **global** id so key sets are disjoint across EP ranks.
- Adds a collective-safe missing-key check via `all_reduce(MAX)` over a **CUDA-resident** int32 flag — CPU tensors trip `RuntimeError: No backend type associated with device type cpu` on the NCCL-backed EP process group.
- Builds the per-rank state dict locally, then byte-stream all-gathers it across the EP group via `torch.save` → `all_gather_object` → `torch.load`. `all_gather_object` pickling fails on quantized uint8 weights because their `UntypedStorage` has no `dtype` attr; the byte-stream round-trip uses PyTorch's own tensor codec to sidestep this.
- Pre-moves shared scales / aux tensors to CPU once so the gather payload doesn't repeatedly clone GPU tensors.

When EP == 1 the gather is skipped and behavior is identical to the original loop modulo the global-id rename (a no-op when local_expert_indices is the trivial 0..N-1).

### 2. `config.json` race at DP > 1

The post-save block that injects `quantization_config` into `config.json` had no DP-rank gating. With DP > 1 there are multiple last-stage main ranks; all of them read-then-write the same file, and any interleaving can leave another rank reading a truncated file, raising `JSONDecodeError`. Guarded with `is_last_stage_main_rank AND get_data_parallel_rank() == 0` and bracketed with `torch.distributed.barrier()`s so every PP/DP rank waits for the single writer.

Two new imports (`get_data_parallel_rank`, `get_expert_model_parallel_*`) and an `io` import are added at the module level.

## Repro

Without this patch, exporting Nemotron 3 Ultra (108-layer Mamba-MoE hybrid, 512 routed experts) at `--pp 9 --tp 1 --ep 8` produces a HuggingFace checkpoint where only 64 of 512 experts per MoE layer are present, and the `config.json` write races to a JSONDecodeError on DP=8. With this patch both succeed and vLLM loads the resulting checkpoint and serves real generations.

## Test plan

- [x] Exported a 108-layer 512-expert Nemotron-3-Ultra MoE checkpoint at PP=9 TP=1 EP=8 DP=8 (72 ranks across 9 nodes × 8 GPUs). Inspected `model.safetensors.index.json` and confirmed all 512 experts per MoE layer are present in the HF shards.
- [x] Sanity-loaded the checkpoint with `vllm serve` and verified end-to-end generation (no `KeyError` on weight loading, real model output).
- [x] Pre-existing unrelated ruff RUF059 warnings on the same file (lines 1453, 1473) are untouched by this PR.
- [ ] No new unit-test coverage added — the bug requires a multi-node EP/DP > 1 distributed run and is non-trivial to mock. Open to suggestions on a CI-friendly regression test.

## Follow-up

The `SequentialMLP` path in `_get_transformer_layer_state_dict` (iterating `local_experts.linear_fc{1,2}`) has the same local-id-collision issue and will need an equivalent global-id + EP-gather treatment when MoE recipes start exercising that spec. Not addressed here because the recipes in current Megatron-Bridge mostly use `TEGroupedMLP`.

Signed-off-by: James Shen <yueshen@nvidia.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented concurrent corruption during distributed model export by coordinating writes and gating sidecar/config file updates to a single writer rank with barriers.

* **Improvements**
  * Made multi-rank expert-parallel export collective-safe with consistent global expert indexing and cross-rank verification of missing pieces.
  * Reduced memory pressure by moving large quantization/auxiliary tensors to CPU and using safe per-rank serialization and exchange for final exports.

### DIFF
--- a/modelopt/torch/export/unified_export_megatron.py
+++ b/modelopt/torch/export/unified_export_megatron.py
@@ -18,6 +18,7 @@
 
 """Code that export quantized Megatron Core models for deployment."""
 
+import io
 import json
 import os
 import tempfile
@@ -85,6 +86,10 @@ with import_plugin("megatron"):
         HybridModel = MambaModel
     from megatron.core.models.multimodal.llava_model import LLaVAModel
     from megatron.core.parallel_state import (
+        get_data_parallel_rank,
+        get_expert_model_parallel_group,
+        get_expert_model_parallel_rank,
+        get_expert_model_parallel_world_size,
         get_pipeline_model_parallel_rank,
         get_pipeline_model_parallel_world_size,
         get_tensor_model_parallel_rank,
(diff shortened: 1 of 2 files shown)


yueshen2016:
Force-pushed `460ff391f` to address two issues I flagged during a self-review:

**1. `config.json` write gate now also pins EP rank 0.**

Previous gate (`is_last_stage_main_rank AND dp_rank == 0`) was topology-dependent: it worked on our PP=9/TP=1/EP=8 run because DP=8 and EP=8 share the same 8 GPUs per PP stage (ETP=1), so DP=0 happened to also imply EP=0. On a cluster where DP and EP are orthogonal axes, multiple EP ranks would still race. Added `get_expert_model_parallel_rank() == 0` for an unambiguous single writer regardless of topology.

**3. `_per_layer_quant_config` now records all global expert ids on every rank, not just the local 1/EP slice.**

Without this, the writer rank would emit an `hf_quant_config.json` missing (EP-1)/EP of the routed-expert entries — vLLM's mixed-precision loader would then treat those experts as un-quantized and crash on shape/dtype mismatch. I'd worked around this downstream by manually rewriting `config.json -> quantization_config` post-export; this commit fixes it at the source so other users don't need the workaround. Within a TEGroupedMLP layer all routed experts share the same qformat/block_size (one recipe pattern matches the whole `*mixer.experts.*` glob), so the local-rank metadata is sufficient to record for all global ids.

Concerns 2 (`local_expert_indices` could be a tensor → defensive `.tolist()`), 4 (gather memory cost), 5 (`weights_only=False` forward-compat), 6 (inline comment expansion), and 7 (unit test) are deferred to a follow-up if maintainers think the current code is good enough to merge in this iteration.

yueshen2016:
Force-pushed `0083a75d3` addressing all three actionable items from CodeRabbit's pre-merge checks and a security audit point.

## Responses to inline review comments

### @coderabbitai inline #1 (line 432, Major — hoist single-writer guard)

**Applied.** You were right that gating only the `config.json` patch leaves a race on the surrounding writes. The single-writer guard is now hoisted around two more blocks:

- **Lines 314-350 (sidecar copies / `save_pretrained` calls).** Wrapped in `if get_data_parallel_rank() == 0 and get_expert_model_parallel_rank() == 0:` inside the outer `is_last_stage_main_rank` check. The MTP-state-dict load (lines 352-355) is intentionally **left running on every last-stage main rank** because it mutates per-rank `layer_state_dicts` that later feed each rank's own safetensors save — single-writer here would lose 7/8 of the MTP head weights.
- **Lines 362-385 (`hf_quant_config.json` write).** Same hoisted gate; `self._hf_quant_config` now stays empty on non-writer ranks, which makes the existing `config.json` gate at the bottom of the method naturally a no-op on them. The barriers we added at the `config.json` site already cover the non-writer ranks' wait.

### @coderabbitai inline #2 (line 1214, Critical — `weights_only=False`)

**Applied.** Added an inline `# weights_only=False:` comment justifying the use per SECURITY.md:

> The payload is generated by us via `torch.save` in this same function on a sibling rank in the EP process group of this job — it never leaves the cluster's collective and is not user-supplied. `weights_only=True` would refuse to deserialize the dict[str, Tensor] because quantized uint8 tensors store custom storage metadata that the safe-loader allowlist doesn't cover.

Also added a `# noqa: S614` suppression on the line so the security-anti-patterns check passes.

### Pre-merge check: Security Anti-Patterns — Error (line 1213)

Resolved by the comment above on line 1213.

## Responses to other reviewers

### @codecov (Patch coverage 2.08%)

The new code lives in distributed-only code paths (`_grouped_mlp_slicing`, EP/DP-gated config writers) that require ≥1 multi-node multi-process distributed run to exercise. The EP > 1 code path specifically requires `torch.distributed.is_initialized() == True` with `expert_model_parallel_world_size > 1`, which isn't covered by current unit tests in this file (existing tests run single-rank).

The single-rank fallback (`if ep_size == 1: self._state_dict.update(local_expert_state)` — the `else` branch at line 1217) is testable and could be exercised by a unit test instantiating a `TEGroupedMLP`-like object with a synthetic state_dict. I'd be happy to add that in a follow-up, but the substantive correctness here is the EP > 1 path which needs a distributed test harness this repo doesn't currently have for unified_export_megatron.

End-to-end validation: ran the patched code against the Nemotron 3 Ultra checkpoint at PP=9 TP=1 EP=8 → 512 experts/layer present in the exported HF shards, vLLM 0.21 with `quant_algo: MIXED_PRECISION` loads via the `modelopt_mixed` path and serves real generations.

### Pre-merge check: Docstring coverage 66.67%

This appears to be a whole-file metric. The two functions I substantively touched (`_grouped_mlp_slicing` and the surrounding `save_pretrained` block) already have docstrings; my changes expanded the existing `_grouped_mlp_slicing` docstring to document the new collective-behavior contract (EP-group `all_gather_object`). I haven't added docstrings to other pre-existing undocumented functions in the file since they're outside the scope of this PR. Happy to tighten if maintainers prefer a separate docstring-coverage cleanup commit.

---

Signed-off-by: James Shen <yueshen@nvidia.com>

yueshen2016:
Force-pushed `46b5c8c03` to fix the `code-quality` failure.

The check failed on a single `RUF003` (ambiguous unicode) in a comment I wrote — `DP×EP` used U+00D7 MULTIPLICATION SIGN. Replaced with ASCII `x`. Also normalized 5 em-dashes (`—`) to ASCII `--` in our added comments to keep the diff fully ASCII and avoid any future RUF002 trip.

Locally verified `ruff check --select=RUF001,RUF002,RUF003` and `ruff format --check` both pass. The two pre-existing `RUF059` warnings at lines 1504 and 1524 are upstream and not touched by this PR (confirmed by stashing and re-checking).

meenchen (commented):
Changes look good to me overall. Two minor comments:
* Can you improve the comments/doc string in the changes and make them more concise?
* Should we add some unit tests to guard the two failure cases?

cjluo-nv (commented):

The global expert-ID mapping, collective-safe missing-weight check, and single-writer config update look coherent, and the reported 72-rank EP8/DP8 export is strong manual validation. The added tests cover the local mapping and writer predicate, but they do not execute the new EP collective/serialization path or the barrier-protected DP write path. Given that these are the core correctness changes and failures can deadlock or corrupt a very large export, this warrants owner sign-off despite the successful manual run.

Minor convention issue: move `import modelopt.torch.export.unified_export_megatron as uem` to the test module’s top-level imports; no concrete reason is given for a local Python import.

meenchen (approved):

PR #1640, opened 2026-06-05
feat: Layerwise calibration memory optimizations (non-mutating skip + meta placeholders)

Fridah-nv:
### What does this PR do?

**Type of change:** New feature

Adds two opt-in memory optimizations to layerwise calibration (built on the nested `LayerwiseConfig` foundation merged in #1571):

1. **Skip layer-weight checkpoint + in-memory writeback** (`LayerwiseConfig.calib_mutates_weights`). For amax-only algorithms (`max`/`mse`/`local_hessian`) — which only update `TensorQuantizer._amax` and never touch `layer.weight` — set `calib_mutates_weights=False` to (a) checkpoint only the quantizer-state slice (`quantizer_buffers.pt`) instead of the full layer weights blob, and (b) skip the offload/shard *writeback* in `persistent_materialization`. A model validator restricts `calib_mutates_weights=False` to those amax-only algorithms; weight-mutating algorithms (GPTQ/AWQ/SmoothQuant) reject it. Also fixes the FSDP2 `writeback=False` gather ordering so the layer is actually all-gathered (not left sharded) during calibration.
2. **Meta-device skip-layer placeholders.** Already-calibrated ("skip") layers now emit zero-filled `meta` tensors instead of real-device buffers, eliminating their activation memory. (Limitation: models whose parent `forward` runs real-device ops on the hidden state *between* decoder blocks are unsupported and should use non-layerwise calibration — this is unconditional, not gated by `calib_mutates_weights`.)

### Usage

```python

cfg = mtq.INT8_DEFAULT_CFG
# Amax-only algorithm: skip the per-layer weight checkpoint + writeback to save memory.
cfg["algorithm"] = {"method": "max", "layerwise": {"enable": True, "calib_mutates_weights": False}}

mtq.quantize(model, cfg, forward_loop)
```

### Testing

- **Unit** (`tests/unit/torch/quantization/`): `test_config_validation.py` (the `calib_mutates_weights` whitelist accept/reject), `test_layerwise_calibrate.py` (checkpoint save/resume, the meta-placeholder behavior, and incomplete-checkpoint guard). All passing on CPU (125 passed, 1 skipped — the meta inter-layer-ops case).
- **GPU** (multi-GPU runner): `tests/gpu/torch/quantization/test_fsdp2.py::test_layerwise_calibrate_fsdp2` asserts every FSDP layer is actually all-gathered (params are **not** `DTensor`s) under `writeback=False` — direct coverage of the FSDP2 fix; plus `plugins/test_accelerate_gpu.py` and `test_gptq.py`.

> Note: checkpoint resume uses `torch.load(..., weights_only=False)` on the per-layer `weights.pt` / `quantizer_buffers.pt` / `next_inputs.pt` files. These are written by the calibration run itself into the user-supplied `checkpoint_dir` (not third-party artifacts); the pre-existing pattern is unchanged by this PR. `full_restore` now raises on an incomplete checkpoint rather than silently restoring partial state.

- Is this change backward compatible?: ✅ — `calib_mutates_weights` defaults to `True` (prior behavior); both optimizations are opt-in and the layerwise config is unreleased.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no copied code or new dependencies.

### Additional Information

Co-authored with @realAsma.

## Release Notes

* **New Features**
  * Added `layerwise.calib_mutates_weights` to control whether layerwise calibration performs weight writeback, with safer support limited to amax-only algorithms.
  * Improved layerwise skip-mode to reduce activation memory by using meta-device placeholders for already-calibrated layers.

* **Improvements**
  * Enhanced layerwise checkpointing/resume with stronger manifest validation, scenario-aware persistence behavior, and clearer progress reporting.
  * Extended writeback control for persistent weight materialization across supported backends.
  * Updated `gptq` documentation to clarify Layerwise vs Non-layerwise semantics.

* **Tests**
  * Expanded GPU/CPU/FSDP2 and validation coverage for meta placeholders, checkpoint layout, and config deprecations.


modelopt/torch/quantization/config.py
```diff
@@ -702,6 +713,10 @@ def _coerce_layerwise_input(value):
 class QuantizeAlgorithmConfig(ModeloptBaseConfig):
     """Calibration algorithm config base."""
 
+    # Set True only for algorithms that update solely ``TensorQuantizer._amax``
+    # (no ``layer.weight`` mutation). Gates ``layerwise.calib_mutates_weights=False``.
+    _supports_save_quantizers_only: ClassVar[bool] = False
```
realAsma: Now we have two things: `calib_mutates_weights` and `_supports_save_quantizers_only` -> should we consolidate this and use say `_mutate_weights` here instead of `_supports_save_quantizers_only`. This could make the code consistent and easier to read (no two concepts).

realAsma (approved):
Let me minor comment, please address it. Otherwise looks goo to me.

cjluo-nv (commented):

Additive memory-optimization PR (+359/-78) extending the existing `LayerwiseConfig`/`_CheckpointState`/`persistent_materialization` infra from #1571 with two opt-in knobs. No prompt-injection in untrusted content; no licensing changes (headers/LICENSE untouched).

**Design review (gate fired on 7 dirs):** This is *not* a new subsystem — no new DSL/loader/registry. It threads `calib_mutates_weights`/`writeback` params through existing functions and adds a `_supports_save_quantizers_only` ClassVar whitelist. The PR body explicitly justifies the design (amax-only whitelist, meta placeholders, FSDP2 ordering fix). Design is settled; proceeded to correctness.

**Correctness — looks sound:**
- The `calib_mutates_weights=False` whitelist is correctly enforced via model_validator; GPTQ/AWQ/SmoothQuant reject it (tested), max/mse/local_hessian accept (tested). AWQ method-routing in `mode.py` can't reach the flag because validation rejects it first.
- Restore branches are complementary: full-weights path vs `quantizer_buffers.pt` (`restore_quantizer_state` + `set_quantizer_state_dict`); the "weights unchanged" assumption holds for amax-only algos.
- Removed the inner `persistent_materialization` in `cache_outputs_for_next_layer_calib` — the caller now wraps the whole block once (no double-materialization); docstring updated accordingly.
- Removed-now-dead `_remap_output_metadata_device` is consistent since `_zeros_from_meta` always uses `meta` device.

**Why nudge (human sign-off warranted):**
- **Documented behavior regression:** skip layers now emit `meta` tensors, so models with real-device inter-layer ops on the hidden state (e.g. inter-layer LayerNorm) are now *unsupported* — `test_skip_output_preserves_shape_with_inter_layer_norm` flipped from passing to `@pytest.mark.skip`. This is unconditional (not gated by the flag) and silently reduces supported models; an owner should confirm no in-tree architectures rely on it.
- **Core paths are GPU-bucket only:** the FSDP2 `writeback=False` all-gather fix (`test_layerwise_calibrate_fsdp2`) and accelerate top-hook-bypass assertions are multi-GPU tests, not CPU CI; broad blast radius across FSDP2/CPU-offload/disk-offload writeback. CPU tests cover config validation + checkpoint resume shape only.
- Minor: the new disk-offload `warnings.warn` sits inside the per-key writeback loop (dedup mitigates noise; TODO acknowledges incomplete disk writeback).

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -105,11 +107,11 @@ class LayerActivationCollector:
...
-    * **skip** — return a zero-filled dummy whose shape and type match the
-      layer's real output (reconstructed from lightweight metadata).  No
-      computation is performed.  The correctly shaped dummy ensures un-patched
-      inter-layer operations in the parent forward (e.g. LayerNorm, tuple
-      unpacking) do not raise shape or type errors.
+    * **skip** — return a zero-filled meta-device dummy whose shape and type
+      match the layer's real output (reconstructed from lightweight metadata).
+      No computation or real-device allocation is performed. Tuple/list
+      structure is preserved for parent code that unpacks outputs, but
+      real-device inter-layer tensor operations are intentionally unsupported.
```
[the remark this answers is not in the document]
sugunav14: @Fridah-nv have you considered this?

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -188,11 +197,11 @@ def _extract_output_meta(output):
...
     @staticmethod
     def _zeros_from_meta(meta):
-        """Reconstruct a zero-filled output from metadata produced by ``_extract_output_meta``."""
+        """Reconstruct a zero-filled meta placeholder from ``_extract_output_meta`` metadata."""
         tag = meta[0]
         if tag == "tensor":
-            _, shape, dtype, device = meta
-            return torch.zeros(shape, dtype=dtype, device=device)
+            _, shape, dtype, _device = meta
+            return torch.zeros(shape, dtype=dtype, device=torch.device("meta"))
```
sugunav14, line 204: I could see an issue when there is some logic/computation between two decoder layers. We might get a device mismatch issue. Do you think we can throw a more descriptive error for such cases.

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -486,7 +487,24 @@ def _set_parameter(module: nn.Module, name: str, value: nn.Parameter):
 
 
 @contextmanager
-def fsdp2_weight_access_and_writeback_context(module: nn.Module, root_model: nn.Module):
+def _fsdp2_unshard_context(fsdp_module: FSDPModule):
```
sugunav14, line 490: is this for the case where we don't mutate weights? Have you considered reusing the original context with an option to skip updating the parameter groups instead of a new context? Might help readability so that we don't have to maintain multiple contexts for similar type of task.

sugunav14 (approved):
Left some minor comments/clarifications. Overall LGTM!

modelopt/torch/quantization/config.py
Fridah-nv: Updated, thanks for the suggestion

modelopt/torch/quantization/utils/layerwise_calib.py
Fridah-nv: Good call, I've added a guard in the model-forward wrapper that catches a meta-device `RuntimeError `and points users to non-layerwise calibration.

modelopt/torch/quantization/utils/core_utils.py
Fridah-nv: The single public entry is already unified — `fsdp2_weight_access_and_writeback_context(..., writeback=...)` is the "option to skip updating the parameter groups" you're describing; `_fsdp2_unshard_context` is just a private helper it delegates to for the `writeback=False` branch. I factored it out rather than inlining an if/else because the two paths are different mechanisms, not one path with a flag: writeback does per-DTensor redistribute + parameter swap + copy-back, while the no-writeback path unshards/reshards the whole group in place and relies on the specific `unshard()`-before-`_disable_fsdp_unshard_reshard` ordering from the recent bug fix. Keeping that isolated in a named helper reads more clearly to me than a branch inside the larger function

PR #1372, opened 2026-04-29
Add closed-form MXFP4 -> NVFP4 weight cast (--cast_mxfp4_to_nvfp4)

cjluo-nv:
## Summary

- New `--cast_mxfp4_to_nvfp4` flag in `hf_ptq.py` (and `huggingface_example.sh`) that converts an MXFP4 source checkpoint (e.g. `openai/gpt-oss-20b`) into an NVFP4 export with **bit-exact** weight reconstruction for the in-range blocks.
- The cast pins NVFP4's `scale_2 = 2^m` (where `m = k_max − 8`) and `_amax = 6·2^k_j` per NVFP4 block, both read from the source `*_scales`. The resulting per-block scale `2^(k_j − m)` is exactly representable in E4M3, so `round_to_E2M1(value / 2^k_j)` yields the original MXFP4 nibble verbatim. For out-of-range blocks (`k_max − k_j > 17`) the per-block amax falls back to data-derived `max(|w_block|)`, which keeps the post-E4M3-clamp scale close to the block's actual magnitude.

## Verification

End-to-end on `openai/gpt-oss-20b` with `--qformat=nvfp4_mlp_only --cast_mxfp4_to_nvfp4`:

```
[cast_mxfp4_to_nvfp4] overrode 48/48 weight quantizers
[cast_mxfp4_to_nvfp4] lossless layers: 48/48 (100.00%)
[cast_mxfp4_to_nvfp4] lossless blocks: 597196800/597196800 (100.0000%)
```

End-to-end on `openai/gpt-oss-120b` with the same flags (4×B200, `--use_seq_device_map --gpu_max_mem_percentage 0.5 --calib_batch_size 4`):

```
[cast_mxfp4_to_nvfp4] overrode 72/72 weight quantizers
[cast_mxfp4_to_nvfp4] lossless layers: 67/72 (93.06%)
[cast_mxfp4_to_nvfp4] lossless blocks: 3583179586/3583180800 (100.0000%)
```

Five layers fall into the OOR regime (block-spread > 17); the remaining 1,214 OOR blocks use the data-derived per-block amax fallback. Block-level losslessness is **99.99996%** end-to-end.

Per-tensor MSE between MXFP4 source dequant and NVFP4 export dequant (~19B elements):

| Metric | Without cast | With cast |
|---|---|---|
| Per-tensor SNR | ~26.4 dB (FP4 noise floor) | **∞ (every tensor)** |
| Total RMSE | 8.67e−02 | **0** |
[...]

PR #1653, opened 2026-06-08
feat(deepseek): add --cast_mxfp4_to_nvfp4 to deepseek_v4 quantize step

cjluo-nv:
### What does this PR do?

Type of change: new feature

Brings the GPT-OSS lossless MXFP4 → NVFP4 cast (#1372) to DeepSeek V4's routed-expert export by adding a `--cast_mxfp4_to_nvfp4` flag to `examples/deepseek/deepseek_v4/quantize_to_nvfp4.py`.

To avoid duplicating the closed-form math, the shared numerics — `mxfp4_to_nvfp4_global_amax`, `mxfp4_to_nvfp4_per_block_amax`, and the E2M1/E4M3/E8M0 constants — are **hoisted out of the GPT-OSS example cast into the library** at `modelopt/torch/quantization/utils/numeric_utils.py`. Both the GPT-OSS cast (`examples/llm_ptq/cast_mxfp4_to_nvfp4.py`) and the new DeepSeek path now import them from there.

DeepSeek V4's routed experts ship as MXFP4 (E2M1 nibbles + a power-of-two E8M0 scale per 32-element block). By default the export dequantizes them to BF16 and re-quantizes to NVFP4 using the calibrated per-tensor weight amax, which re-derives per-block scales from the data and is therefore lossy. With the flag, the cast pins `scale_2 = 2^(k_max-8)` and each per-block E4M3 scale to `2^(k_j-m)` straight from the source E8M0 scales, so `per_block_scale * scale_2 = 2^k_j` and the NVFP4 nibbles equal the source MXFP4 nibbles bit-for-bit (for every block whose `k_j` lands in E4M3's representable window; rare out-of-range blocks clamp). The one V4-specific addition is that w1/w3 share a single `scale_2` for the fused GEMM1, so `k_max` is taken over both projections. The flag only affects routed-expert **weights** — activation `input_scale` still comes from `--amax_path` calibration.

### Usage

```bash
python deepseek_v4/quantize_to_nvfp4.py \
    --amax_path ${AMAX} \
    --source_ckpt ${DS_V4} \
    --output_ckpt ${HF_NVFP4_PATH} \
    --cast_mxfp4_to_nvfp4
```

### Testing

- The hoisted numerics get unit tests in `tests/unit/torch/quantization/test_numeric_utils.py` (10 cases: per-tensor global_amax, per-block amax incl. out-of-range, magnitude-table cache) — 10/10 pass. The example test `tests/examples/llm_ptq/test_cast_mxfp4_to_nvfp4.py` keeps the cast-specific cases (quantizer naming, `build_amax_map`, `apply_to_model`).
- Validated on real DeepSeek-V4-Flash expert tensors (incl. the on-disk `float8_e8m0fnu` scale dtype): 23.5M blocks, 100% lossless, 0 error.
- Generated a full NVFP4 checkpoint for DeepSeek-V4-Flash (43 layers, 256 routed experts) end-to-end: `[cast] lossless MXFP4->NVFP4 blocks: 8,657,043,456/8,657,043,456 (100.0000%)`. Output weights match an independently-produced reference cast byte-for-byte (`weight_scale`, `weight_scale_2`, packed nibbles modulo the harmless sign-of-zero).

- Is this change backward compatible?: ✅ (new opt-in flag; default export behavior unchanged; hoist re-exports through the existing example module)
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ N/A (no new deps; shared numerics moved into the library rather than duplicated)
- Did you write any new necessary tests?: ✅ (library numerics covered by `tests/unit/torch/quantization/test_numeric_utils.py`; end-to-end validated on a real DeepSeek-V4 checkpoint)
- Did you get Claude approval on this PR?: ❌ (will run `/claude review`)

### Additional Information

Mirrors and reuses #1372 (GPT-OSS MXFP4 → NVFP4 cast); the closed-form numerics are now shared via `modelopt.torch.quantization.utils.numeric_utils`.

## Summary by CodeRabbit

* **New Features**
  * Added `--cast_mxfp4_to_nvfp4` flag to perform a closed-form, mostly lossless MXFP4→NVFP4 conversion for routed-expert weights with aggregated lossless/block statistics.

* **Documentation**
  * Updated DeepSeek V4 export instructions and README to document the new flag and clarify calibration behavior for activation scales.

* **Chores**
  * Exposed shared numeric quantization utilities for MXFP4→NVFP4 casting.

* **Tests**
  * Added and updated tests to validate the new numeric helpers and conversion behavior.


meenchen (commented):

Solid follow-up to #1372: the closed-form MXFP4→NVFP4 numerics are hoisted into `modelopt/torch/quantization/utils/numeric_utils.py` and reused by both example cast paths instead of duplicated, and the V4 twist (w1/w3 sharing one `scale_2` for the fused GEMM1) is the right minimal addition. Design protocol satisfied — the PR body explicitly justifies hoisting rather than copy-pasting, and `numeric_utils.py` is a thin pure-tensor utility, not a new framework.

Math reads correctly: `per_block_scale = 2^(k_j-m)` × `scale_2 = 2^m` ⇒ `2^k_j` exactly, BF16 dequant of E2M1 magnitudes is lossless, and the lossless-accounting mask `(k >= k_max - 17) | (e8m0 == 0)` correctly handles all-zero blocks (whose `k_j = -127` would otherwise read as out-of-range). `NVFP4QTensor.quantize` accepts a pre-supplied per-block scale tensor and that path was already exercised by the GPT-OSS cast.

A few things that warrant a human pass before sign-off:

- **No unit tests for the new V4-specific helpers.** `_quantize_weight_nvfp4_lossless`, `_build_w13_kmax_overrides`, and `_kmax_from_mxfp4_scale` are exercised only by the on-the-real-V4-checkpoint validation reported in the PR body; the moved tests in `tests/unit/torch/quantization/test_numeric_utils.py` cover only the hoisted library helpers. A cheap unit test for the V4 path (synthetic MXFP4 weight + scale → roundtrip lossless count, plus a w1/w3 shared-`k_max` case) would lock in the V4-specific contract.
- **README quietly promotes `--cast_mxfp4_to_nvfp4` to the standard example.** `examples/deepseek/README.md` now appends `--cast_mxfp4_to_nvfp4` to the main `quantize_to_nvfp4.py` invocation, not as a "lossless variant" callout — that contradicts the PR body's "default export behavior unchanged" framing. Worth either dropping the flag from the top-level usage block (keeping the dedicated subsection that explains the trade-off) or saying explicitly that the flag is now recommended for V4 sources.
- **Stat naming nit:** `stats[f"cast_oor_layers_{block_kind}"]` is incremented per expert-projection tensor (w1/w2/w3), not per layer — `cast_oor_tensors_*` would match `experts_*` / `weight_synth_*` siblings.
- Minor: `_build_w13_kmax_overrides` reads each `*.scale` once and the main loop reads it again via `f.get_tensor(scale_key)`. Cheap (E8M0 is tiny relative to packed nibbles), but if you're already touching this you could reuse the cached read.

Nothing blocking; the cast itself looks right. Just want a human reviewer to land the unit-test ask and the README usage call.

cjluo-nv:
> > _Bot review — DM the bot to share feedback._
> 
> Solid follow-up to #1372: the closed-form MXFP4→NVFP4 numerics are hoisted into `modelopt/torch/quantization/utils/numeric_utils.py` and reused by both example cast paths instead of duplicated, and the V4 twist (w1/w3 sharing one `scale_2` for the fused GEMM1) is the right minimal addition. Design protocol satisfied — the PR body explicitly justifies hoisting rather than copy-pasting, and `numeric_utils.py` is a thin pure-tensor utility, not a new framework.
> 
> Math reads correctly: `per_block_scale = 2^(k_j-m)` × `scale_2 = 2^m` ⇒ `2^k_j` exactly, BF16 dequant of E2M1 magnitudes is lossless, and the lossless-accounting mask `(k >= k_max - 17) | (e8m0 == 0)` correctly handles all-zero blocks (whose `k_j = -127` would otherwise read as out-of-range). `NVFP4QTensor.quantize` accepts a pre-supplied per-block scale tensor and that path was already exercised by the GPT-OSS cast.
> 
> A few things that warrant a human pass before sign-off:
> 
> * **No unit tests for the new V4-specific helpers.** `_quantize_weight_nvfp4_lossless`, `_build_w13_kmax_overrides`, and `_kmax_from_mxfp4_scale` are exercised only by the on-the-real-V4-checkpoint validation reported in the PR body; the moved tests in `tests/unit/torch/quantization/test_numeric_utils.py` cover only the hoisted library helpers. A cheap unit test for the V4 path (synthetic MXFP4 weight + scale → roundtrip lossless count, plus a w1/w3 shared-`k_max` case) would lock in the V4-specific contract.
> * **README quietly promotes `--cast_mxfp4_to_nvfp4` to the standard example.** `examples/deepseek/README.md` now appends `--cast_mxfp4_to_nvfp4` to the main `quantize_to_nvfp4.py` invocation, not as a "lossless variant" callout — that contradicts the PR body's "default export behavior unchanged" framing. Worth either dropping the flag from the top-level usage block (keeping the dedicated subsection that explains the trade-off) or saying explicitly that the flag is now recommended for V4 sources.
> * **Stat naming nit:** `stats[f"cast_oor_layers_{block_kind}"]` is incremented per expert-projection tensor (w1/w2/w3), not per layer — `cast_oor_tensors_*` would match `experts_*` / `weight_synth_*` siblings.
> * Minor: `_build_w13_kmax_overrides` reads each `*.scale` once and the main loop reads it again via `f.get_tensor(scale_key)`. Cheap (E8M0 is tiny relative to packed nibbles), but if you're already touching this you could reuse the cached read.
> 
> Nothing blocking; the cast itself looks right. Just want a human reviewer to land the unit-test ask and the README usage call.

it's ok to skip unittests impl for example code.

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
```diff
@@ -233,6 +258,98 @@ def _quantize_weight_nvfp4(
...
+            continue
+        k1 = _kmax_from_mxfp4_scale(f.get_tensor(paths["w1"] + ".scale"), device)
+        k3 = _kmax_from_mxfp4_scale(f.get_tensor(paths["w3"] + ".scale"), device)
+        shared = max(k1, k3)
+        overrides[paths["w1"]] = shared
+        overrides[paths["w3"]] = shared
+    return overrides
+
+
+def _quantize_weight_nvfp4_lossless(
```
mxinO, line 309: I suggest to make this a general util in modeopt, I bet the mxfp4-> nvfp4 casting will be a often used util later.

mxinO (approved):
LGTM! Thanks.

examples/deepseek/deepseek_v4/quantize_to_nvfp4.py
cjluo-nv: for the normal PTQ path, we already have the feature integrated. See https://github.com/NVIDIA/Model-Optimizer/blob/main/examples/llm_ptq/cast_mxfp4_to_nvfp4.py

meenchen (commented):

Re-review: prior critical/important comments are addressed.

- :speech_balloon: Operator: "We don't need to add unittest for this example" — drops the V4-specific-helpers unit-test ask from the prior nudge.
- README usage call: top-level `quantize_to_nvfp4.py` invocation no longer appends `--cast_mxfp4_to_nvfp4`; the flag is now in a dedicated "Lossless MXFP4 → NVFP4 weight cast" subsection with the trade-off explained, matching the "default unchanged" framing.
- Stat naming: `cast_oor_layers_*` renamed to `cast_oor_tensors_*` to match the per-tensor increment and the `experts_*` / `weight_synth_*` siblings.
- Hoist + math (closed-form `scale_2 = 2^(k_max-8)`, `per_block_scale = 2^(k_j-m)`, lossless mask `(k >= k_max - 17) | (e8m0 == 0)`, w1/w3 shared `k_max`) is consistent with PR #1372 and previously reviewed.
- `numeric_utils.py` carries the standard NVIDIA Apache-2.0 header; no other licensing changes.

modelopt/torch/quantization/utils/__init__.py
```diff
@@ -18,15 +18,23 @@
...
     "export_torch_mode",
     "is_quantized",
     "is_quantized_column_parallel_linear",
     "is_quantized_linear",
     "is_quantized_row_parallel_linear",
+    "mxfp4_to_nvfp4_global_amax",
+    "mxfp4_to_nvfp4_per_block_amax",
     "reduce_amax",
     "reduce_sum",
     "replace_function",
```
kevalmorabia97, line 32: we shouldnt need to redefine `__all__ = []` inside `__init__.py` as its imported above by `import *`

meenchen (approved):
Looks good

modelopt/torch/quantization/utils/__init__.py
cjluo-nv: Good catch — reverted `__init__.py` to its original state in `ada91f071`. The package-level re-export wasn't needed at all: `cast_mxfp4_to_nvfp4.py`, `quantize_to_nvfp4.py`, and the unit test all import from the `numeric_utils` submodule directly, so I dropped both the `from .numeric_utils import *` and the `__all__` additions.

Edwardf0t1 (commented):
It seems there are overlapped contents in readme and docstrings, could we simplify?

PR #1674, opened 2026-06-10
Make .claude/skills a real folder for claude sandbox to work

sychen52:
### What does this PR do?

Type of change: ?  Bug fix

replace .claude/skills dir-symlink with real dir + per-skill symlinks

Claude Code hardcodes .claude/skills in its sandbox denyWithinAllow list. bwrap enforces this by creating a mount point at that path, but fails with "Can't create file: Is a directory" when the path is a symlink to a directory — breaking all sandboxed commands, not just writes.

Fix by making .claude/skills a real directory containing per-skill symlinks into .agents/skills/. Add a pre-commit hook that automatically creates and stages a new symlink whenever a skill directory is added to .agents/skills/, so authors need no extra steps.

### Usage

claude with sandbox

### Testing
tried on local.

## Summary by CodeRabbit

* **Chores**
  * Updated internal development tooling to streamline workflow automation and improve developer processes.

### DIFF
--- a/.claude/skills
+++ b/.claude/skills
@@ -1,1 +1,0 @@
-../.agents/skills
--- a/.claude/skills/accessing-mlflow
+++ b/.claude/skills/accessing-mlflow
@@ -1,0 +1,1 @@
+../../.agents/skills/accessing-mlflow
--- a/.claude/skills/common
+++ b/.claude/skills/common
@@ -1,0 +1,1 @@
+../../.agents/skills/common
--- a/.claude/skills/compare-results
+++ b/.claude/skills/compare-results
@@ -1,0 +1,1 @@
+../../.agents/skills/compare-results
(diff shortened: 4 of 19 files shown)


cjluo-nv (approved):

Straightforward fix: replaces the `.claude/skills` directory-symlink with a real directory of per-skill symlinks (so bwrap can mount over it), and adds a pre-commit hook that auto-stages new symlinks when skills are added under `.agents/skills/`. All 16 existing skills are mirrored. Script uses `set -euo pipefail`, guards against the no-match glob case with `[ -d "$d" ]`, and skips both existing files and existing symlinks. NVIDIA Apache-2.0 header on the new shell script matches `LICENSE_HEADER`. Minor non-blocking note: the hook doesn't prune dangling symlinks if a skill is removed from `.agents/skills/`, but the existing `check-symlinks` hook would catch that.

chadvoegele (approved):

shengliangxu (approved):
LGTM

PR #1678, opened 2026-06-11
Fix GPT-OSS MXFP4->NVFP4 PTQ load, export, and cast (nvbug 6295279, 6295242)

cjluo-nv:
### What does this PR do?

Type of change: Bug fix

Fixes the GPT-OSS MXFP4 → NVFP4 PTQ path (`examples/llm_ptq/hf_ptq.py` with `--cast_mxfp4_to_nvfp4`), which failed in three independent ways. The documented command now runs end-to-end and produces a bit-exact (100% lossless) NVFP4 checkpoint. Addresses **nvbug 6295279** (OMNIML-5046) and **nvbug 6295242** (OMNIML-5045).

1. **nvbug 6295242 — CUDA illegal memory access on load.** GPT-OSS ships native MXFP4 weights that Transformers dequantizes to BF16; the threaded weight loader trips an illegal-memory access when `device_map="auto"` shards the dequant across **multiple GPUs**. The missing optional `kernels` package only *forces* the dequant path — it is not the root cause. `get_model` now detects MXFP4 checkpoints and loads them with `Mxfp4Config(dequantize=True)` on a **sequential** device map so the dequant stays on a single device. `kernels` is no longer required.
2. **nvbug 6295279 #1 — `NotImplementedError: Mxfp4GptOssExperts` during unified HF export.** Forcing `dequantize=True` yields plain `GptOssExperts` (even when `kernels` is installed), which ModelOpt wraps and exports normally.
3. **nvbug 6295279 #2 — `FileNotFoundError` in the cast step.** `--cast_mxfp4_to_nvfp4` treated `--pyt_ckpt_path` as a local dir; a HF Hub ID now resolves to its cached snapshot dir via `_resolve_model_path`.

Also fixes a **static-block NVFP4 regression** (surfaced by the cast's `force_weight_quantizers_static`, introduced by #1560's now-unconditional `weight_only_quantize`): `_QuantGptOssExperts` / `_QuantLlama4TextExperts` quantize their expert weights transposed in the forward (`_transposed_quantize`), but the inherited `iter_weights_for_calibration` fed the non-transposed weight, locking a mismatched block-quant `_original_shape` and raising `ValueError: Input shape has changed`. The override now calibrates on the transposed view, matching both the forward and the export's `_amax` orientation.

### Why this regressed (it worked when the cast was added)

`get_model` never had explicit handling for a *natively pre-quantized MXFP4* checkpoint — GPT-OSS fell through the generic *unquantized-checkpoint* branch and relied on Transformers' **implicit** MXFP4 behavior, which is fragile across three axes. The cast was originally validated (#1372, 2026-05-01) in the "lucky" quadrant of each:

- **GPU count:** `device_map="auto"` on a single GPU never shards, so the dequant stays on one device. On multiple GPUs `auto` balances the model and shards the MXFP4→BF16 dequant across devices → CUDA illegal-memory crash (6295242).
- **`kernels` presence:** without `kernels`, Transformers auto-dequantizes to BF16 `GptOssExperts` (exportable). With `kernels` installed it keeps the packed `Mxfp4GptOssExperts` kernel path → export `NotImplementedError` (6295279 #1).
- **Transformers version:** the kernel-backed experts wrapper and the threaded multi-GPU weight loader are newer-Transformers behavior (env here is 5.5.4). Earlier versions simply dequantized MXFP4 → BF16, which is what the old generic path happened to need.

The QA env sat in the *breaking* quadrant (multi-GPU and/or `kernels` present, newer Transformers), so the implicit path failed. The new branch makes both decisions explicit and deterministic (`dequantize=True` + single-device load), regardless of environment — mirroring the existing `has_pack_quantized_config` branch for compressed-tensors checkpoints.

The fourth issue (static-block `Input shape has changed`) is a separate regression: it was introduced by **#1560 (2026-06-02, "Make sure all weight quantizers have `_amax`")**, a month *after* the cast landed. #1560 made `weight_only_quantize` unconditional in `max_calibrate`; previously it ran only when no calibration `forward_loop` was supplied, and the cast always supplies one — so the non-transposed weight-quantizer call simply never happened before. The conflict only appears at the intersection of (a) transposed-quantize experts (GPT-OSS/Llama4), (b) static-block NVFP4 — which `--cast_mxfp4_to_nvfp4` forces via `force_weight_quantizers_static` — and (c) #1560. CI's GPT-OSS NVFP4 coverage uses the *dynamic*-block path, which never locks the block shape, so #1560 looked safe.

### Usage

```bash
python hf_ptq.py \
  --pyt_ckpt_path openai/gpt-oss-20b \
  --qformat nvfp4_mlp_only \
  --cast_mxfp4_to_nvfp4 \
  --export_path ./gpt-oss-20b-nvfp4
```

### Testing

- Ran the documented command end-to-end on 2xB200 (`openai/gpt-oss-20b`): cast overrode **48/48** expert weight quantizers, **100% lossless** layers/blocks, exported a valid packed-NVFP4 HF checkpoint (uint8 weights + FP8 per-block `weight_scale` + per-tensor `weight_scale_2` + `hf_quant_config.json`).
- Verified plain `--qformat nvfp4_mlp_only` (no cast) still works end-to-end.
- **Independently verified the export is bit-exact:** dequantized the exported NVFP4 weights (ModelOpt's E2M1 LUT + pack layout) and compared against Transformers' canonical MXFP4→BF16 dequant (`Mxfp4Config(dequantize=True)`) over all 24 layers × both expert weights — `max_abs_err = 0`, 100% bitwise-equal in bf16. So `dequant(exported NVFP4) == dequant(original MXFP4)` exactly.
- New unit tests: `test_get_original_hf_quant_method_*` (load detection) and `test_gpt_oss_experts_iter_weights_for_calibration_transposed` (the transpose regression). Existing `test_cast_mxfp4_to_nvfp4.py` (8 tests) still pass. `pre-commit` clean.

**Known limitation:** verified for gpt-oss-20b (fits one GPU). gpt-oss-120b dequantized does not fit a single GPU, so `sequential` would still span GPUs — that case would need a CPU-dequant-then-dispatch path and is left as a follow-up.

- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ (0.45 Bug Fixes)

### Additional Information

nvbug 6295279, nvbug 6295242 / OMNIML-5046, OMNIML-5045.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented CUDA illegal-memory access during MXFP4→NVFP4 casting.
  * Fixed expert-weight calibration orientation to avoid shape mismatches.

* **New Features**
  * Support loading native MXFP4 checkpoints with automatic dequantization.
  * Resolve remote model identifiers to local checkpoints when casting MXFP4→NVFP4, improving reliability.

* **Tests**
  * Added unit and GPU regression tests covering quant-method detection, casting, and expert-weight calibration.


meenchen (commented):

Well-scoped bug-fix PR for the GPT-OSS MXFP4→NVFP4 PTQ path (nvbug 6295279, 6295242). I verified the key claims against the repo:

**Design review (gate fired on 6 dirs):** This is additive, not a new subsystem. The mxfp4-load branch mirrors the existing `has_pack_quantized_config` branch directly above it in `get_model`, and `_resolve_model_path` is an existing utility reused rather than reinvented. There's no pre-existing mxfp4 handling in the PTQ path (the only other `Mxfp4Config(dequantize=True)` uses are in unrelated gpt-oss SFT/QAT examples). The PR body extensively justifies the `dequantize=True` + `sequential` device-map choice and the transposed-calibration fix, so the design question is addressed. The gate fired mainly because tests span multiple directories.

**Correctness:** Confirmed both `_QuantGptOssExperts.forward` and `_QuantLlama4TextExperts.forward` quantize via `_transposed_quantize` (transpose -1,-2), so the new `iter_weights_for_calibration` override yielding `weight.transpose(-1, -2)` correctly matches the forward and export `_amax` orientation — fixing the "Input shape has changed" regression. The base `iter_weights_for_calibration` signature is preserved. `get_original_hf_quant_method` correctly handles dict/object/nested `text_config` cases.

**Tests:** Strong coverage — unit tests for quant-method detection, the transpose regression, `force_weight_quantizers_static`, plus a GPU e2e regression guard (using the existing `get_tiny_gpt_oss` helper).

**Licensing:** New GPU test file's header matches the canonical `LICENSE_HEADER` exactly; standard-header exception applies.

**Size:** +297 -1, cohesive.

Known limitation (gpt-oss-120b won't fit a single GPU under `sequential`) is acknowledged in the PR body and left as a follow-up. No prompt-injection attempts in the untrusted content.

Complex PR: spans 6 directories (≥ 5). Looping in a human for approval.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -608,6 +633,11 @@ def _setup(self):
         self.down_proj_input_quantizer = TensorQuantizer()
         self.down_proj_weight_quantizer = TensorQuantizer()
 
+    def iter_weights_for_calibration(self):
```
sychen52: Looks like this function should be extracted out and even replace the _iter_transposed_expert_weights_for_calibration.

examples/llm_ptq/example_utils.py
```diff
@@ -587,6 +587,25 @@ def _resolve_file(filename):
         module.__dict__.pop("weight", None)
 
 
+def get_original_hf_quant_method(config) -> str | None:
```
meenchen, line 590: Is this function specific to GPT OSS? or can it detect pre-quantized checkpoints in general?

meenchen (commented):

Re-review of the GPT-OSS MXFP4→NVFP4 PTQ fix (PR #1678). The change is well-scoped, additive, and well-tested. All previous review comments were minor and are resolved or non-blocking:

- :speech_balloon: **sychen52 (huggingface.py:636, "extract this function")** — Addressed: the transposed-calibration logic is now factored into the shared helper `_iter_transposed_expert_weights_for_calibration`, and both `_QuantGptOssExperts` and `_QuantLlama4TextExperts` delegate to it. No remaining concern.
- :speech_balloon: **meenchen (example_utils.py:590, "is this GPT-OSS specific?")** — Addressed: `get_original_hf_quant_method` is generic — it returns any checkpoint's `quantization_config.quant_method` (dict/object/nested `text_config`), not just `"mxfp4"`. The mxfp4 specialization lives in the `get_model` branch, not the detector.
- **CodeRabbit (example_utils.py:722, local `Mxfp4Config` import)** — Still present and the inline comment justifies forcing dequantization but not why the import is local. This is a minor style nit (CodeRabbit itself flagged it as a nitpick) and does not block.

Correctness spot-check confirms the transposed calibration view (`weight.transpose(-1, -2)`) matches both `_transposed_quantize` in the forward and the export `_amax` orientation, fixing the "Input shape has changed" regression. The new `get_model` mxfp4 branch mirrors the existing `has_pack_quantized_config` pattern; `_resolve_model_path` is reused, not reinvented.

Flagging for human sign-off because: (1) the change spans 6 directories (design-review complexity gate), (2) the core fixes were validated only on B200/gpt-oss-20b with the 120b multi-GPU case acknowledged as an unhandled follow-up, and (3) the prior bot review already looped in a human. No prompt-injection attempts found in the untrusted content. Licensing is clean (new GPU test file uses the standard NVIDIA Apache header).

examples/llm_ptq/example_utils.py
cjluo-nv: should be general.

modelopt/torch/quantization/plugins/huggingface.py
cjluo-nv: Done in 8766bf21e — extracted into a shared `_TransposedExpertsCalibMixin` that defines `iter_weights_for_calibration` once; both `_QuantGptOssExperts` and `_QuantLlama4TextExperts` now inherit it, and the standalone `_iter_transposed_expert_weights_for_calibration` helper is removed. Thanks!

meenchen (approved):

PR #1688, opened 2026-06-11
fastgen DMD2: make the Qwen-Image example self-contained on stock nemo_automodel

jingyu-ml:
### What does this PR do?

**Type of change:** new example / refactor (example self-containment)

The published `examples/diffusers/fastgen` DMD2 Qwen-Image distillation example previously relied on
**local, unpublished modifications to the sibling `nemo_automodel` package** (data path, collate, a
partial-load checkpointer, and Qwen-Image preprocessing). An external user running the published
Model-Optimizer example against **official stock `nemo_automodel`** would hit import/attribute errors.

This PR makes the example **self-contained on stock `nemo_automodel>=0.4.0`** — with the changes kept
**as small as possible**: the team's actual AutoModel delta is only ~430 lines, so wherever the
upstream module is importable, the delta is expressed as a thin subclass / small reimplementation rather than a
full-file copy.

- **`fastgen_data/`** — the DMD2 data path:
  - `collate_fns.py` — **reuses the upstream `SequentialBucketSampler` and reimplements the collate**:
    it builds the DMD2 batch directly from the vendored dataset's per-item output (`image_latents` /
    `text_embeddings` / `text_embeddings_mask` + an optional broadcast `negative_text_embeddings` for
    CFG) and deliberately does **not** call upstream `collate_fn_production`, which stacks
    model-specific token keys (`clip_tokens` / `t5_tokens`) absent from the Qwen-Image cache. The
    builder loads an optional `negative_prompt_embedding_path`.
  - `text_to_image_dataset.py` — a **faithful vendored copy** of the upstream reader (its
    `prompt_embeds_mask` emission is interleaved with cache loading, so wrapping it would force a
    redundant per-item `torch.load`; carried verbatim instead).
- **`fastgen_checkpoint.py`** — `PartialLoadCheckpointer(Checkpointer)` that overrides only
  `load_optimizer` (FSDP2 `DefaultLoadPlanner(allow_partial_load=True)`) so optimizer resume works
  without patching upstream. Injected via an in-place re-bless of `self.checkpointer` in the
  recipe's `load_checkpoint` (model-state load stays strict).
- **`preprocess/`** — Qwen-Image preprocessing (`preprocessing_multiprocess.py` + `processors/`),
  trimmed to the image path (drops the flux/wan/hunyuan processors and the video base class).
  It lives in AutoModel's top-level `tools/` tree, which is **not** shipped in the pip package, so
  it cannot be wrapped and is vendored; `MultiTierBucketCalculator` is imported from stock upstream.
- **`make_negative_prompt_embedding.py`** — generates the optional CFG negative-prompt embedding.
- All `configs/*.yaml` target `fastgen_data.build_*` (a test enumerates every config).
- Licensing: the AutoModel-copied files are NVIDIA-authored Apache-2.0, so they carry only the
  standard NVIDIA SPDX header (managed by the `insert-license` hook) — no per-file provenance note,
  no duplicated license, no pre-commit exclusion, and no separate `LICENSE` note.
  `nemo_automodel[diffusion]` version bound in `requirements.txt`.

The DMD2 math in `modelopt/torch/fastgen/` is **unchanged** — only example/training-time glue moved.

### Usage

```bash
# Install example deps (stock nemo_automodel) from a source checkout
pip install -r examples/diffusers/fastgen/requirements.txt

# Build the training cache from raw images (Qwen-Image VAE latents + text embeddings)
python examples/diffusers/fastgen/preprocess_qwen_image.py image \
    --image_dir <raw images> --output_dir <cache dir> --processor qwen_image \
    --caption_format meta_json

# Generate the CFG negative-prompt embedding once
python examples/diffusers/fastgen/make_negative_prompt_embedding.py \
    --output <cache dir>/negative_prompt_embedding.pt

# Point the config's data.dataloader.cache_dir + negative_prompt_embedding_path at the cache, then train.
```

See `examples/diffusers/fastgen/README.md` → "Requirements & self-contained data path".

### Testing

- New `tests/examples/diffusers/fastgen/test_vendored_migration.py`: environment-independent
  invariants (every config targets a vendored builder; no `tools.*` imports; each former AutoModel
  patch is vendored / wrapped / a documented exclusion; the former-vendored files carry the standard NVIDIA SPDX header, no provenance note or duplicate license) plus
  dependency-guarded structural tests (the collate emits the batch contract + broadcasts the
  negative embedding; the builder accepts `negative_prompt_embedding_path`; the checkpointer
  overrides only `load_optimizer`; the Qwen-Image processor self-registers).
- **Validated 9/9 against a pure stock `nemo_automodel` 0.4.0 worktree** (none of the local patches
  present) via SLURM — re-run after this slim-down.
- The migrated code path is exercised by a live multi-GPU DMD2 run that resumed from a checkpoint
  through the vendored `PartialLoadCheckpointer`.
- `ruff check` + `ruff format --check` clean on all changed files.

- Is this change backward compatible?: ✅ (additive; bundled configs target the vendored builders)
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ (NeMo-AutoModel @ `e42584e3`, Apache-2.0; per review these NVIDIA-authored files carry the standard NVIDIA SPDX header, no separate provenance / `LICENSE` note; `nemo_automodel` was already a dependency)
- Did you update Changelog?: N/A (example-only change)
- Did you get Claude approval on this PR?: ❌ (pending — opened as draft)

### Additional Information

Opened as a **draft** pending: OSRB review of the vendored NeMo-AutoModel (Apache-2.0) code, CI green,
and (optional) a multi-GPU smoke-train + resume on stock upstream. Vendored from
NVIDIA-NeMo/Automodel at commit `e42584e3`.

### Update (post-review)

- **Mid-run resume data-correctness fix** (`6ffbc52c9`): on resume the `StatefulDataLoader`'s restored state did not advance past the resume point, so each window re-served the same data slice and multi-window (SLURM-windowed) runs under-covered the dataset. Fixed by rebuilding a fresh loader and skipping the deterministic sampler to the position implied by `global_step`; added a SLURM-free, GPU-free CPU regression test (`tests/examples/diffusers/fastgen/test_resume_dataloader.py`).
- **Licensing review** (`be832ae95`): the AutoModel-copied files are NVIDIA-authored Apache-2.0, so they now carry only the standard NVIDIA SPDX header — dropped the per-file provenance note, the duplicated original-license block, the `insert-license` pre-commit exclusion, and the `LICENSE` note.

## Summary by CodeRabbit

* **New Features**
  * Added vendored Qwen‑Image preprocessing (multi-process) with processor registry support.
  * Updated DMD2 data loading with a dedicated dataset/collation pipeline, including negative-prompt embedding/mask handling.
  * Improved training resume behavior by rebuilding dataloader state and making checkpoint optimizer restore tolerant of partial FSDP2 optimizer shards.
* **Documentation**
  * Refreshed the fastgen README and config notes for real-data training; removed the prior mock-data smoke workflow.
* **Tests**
  * Added regression and migration tests covering vendored wiring, collate/dataloader contracts, processor registration, and resume/checkpoint behavior.
* **Chores**
  * Updated licenses/attribution, vendoring/tooling guards, requirements pinning, linting configuration, and repository ownership rules.


jingyu-ml:
/ok to test [85ae40c](https://github.com/NVIDIA/Model-Optimizer/pull/1688/commits/85ae40cfcffd0e83a60a766c79e42bec5a3379ce)

LICENSE
kevalmorabia97, line 1: These are also nvidia authored and apache 2 license files so wr dont need to exclude in pre commit or have duplicate nvidia license. Let license hook auto add correct license everywhere

kevalmorabia97 (changes requested):

LICENSE
jingyu-ml: Good call — addressed in be832ae95. The files copied from NVIDIA-NeMo/Automodel now carry only the standard NVIDIA SPDX header (managed by the `insert-license` hook): I dropped their pre-commit `insert-license` exclusion, removed the per-file `Vendored from` provenance note + the duplicated original-license block, and removed the NVIDIA-NeMo/Automodel note from `LICENSE` (the HuggingFace / MIT / BSD third-party notices are unchanged). The migration test now asserts the standard header + no duplicate license instead of requiring the provenance note.

Happy to keep a one-line `adapted from` note below the header on any of these if you'd prefer.

kevalmorabia97:
/ok to test [49d390d](https://github.com/NVIDIA/Model-Optimizer/pull/1688/commits/49d390d22c8ffad310f73807b24cb0aac8e12ebc)

examples/diffusers/fastgen/README.md
```diff
@@ -11,6 +11,41 @@ output distribution. Built on `modelopt.torch.fastgen` and NeMo AutoModel's
 > [Qwen-Image model card](https://huggingface.co/Qwen/Qwen-Image) before downloading or
 > redistributing weights or derivatives.
 
+## Requirements & self-contained data path
+
+This example runs against **stock upstream `nemo_automodel`** (`>=0.4.0,<1.0`; see
```
kevalmorabia97, line 16: can we recommend the nemo_automodel docker here if users dont want to install it separately?

kevalmorabia97:
Test failing:

```
FAILED tests/examples/diffusers/fastgen/test_vendored_migration.py::test_collate_emits_contract_keys_and_broadcasts_negative_prompt - KeyError: 'clip_tokens'
```

examples/diffusers/fastgen/README.md
jingyu-ml: Good idea — done in 7b2581219. Added a tip in the Requirements section recommending the **NeMo AutoModel container** (`nvcr.io/nvidia/nemo-automodel:26.04`), which bundles `nemo_automodel` + the diffusion extras, so users can skip the separate `pip install`.

jingyu-ml:
Fixed in 7b2581219.

**Root cause:** the vendored collate delegated to the stock `collate_fn_production`, and the released `nemo_automodel==0.4.0` version of it unconditionally stacks `clip_tokens` / `t5_tokens` — keys the Qwen-Image cache doesn't produce — so it `KeyError`ed. (The model-agnostic `if key in batch[0]` guard only exists in unreleased dev, which is why it passed locally.) That would also have broken the real Qwen path on stock 0.4.0, not just the test.

The vendored collate now builds the DMD2 batch directly from the vendored `TextToImageDataset` per-item output (they're a matched pair), so the example is self-contained on stock 0.4.0. Verified `test_collate_emits_contract_keys_and_broadcasts_negative_prompt` and the full vendored-migration suite (9/9) pass against PyPI `nemo_automodel==0.4.0`.

tests/examples/diffusers/fastgen/test_resume_dataloader.py
```diff
@@ -0,0 +1,163 @@
...
+        assert "dataloader" in recipe.__dict__["__state_tracked"]
+
+        recipe._rebuild_dataloader_for_resume(global_step)  # must not raise "already tracked"
+
+        # Position was derived from global_step (also drives the epoch label + progress bar).
+        assert recipe.step_scheduler.epoch == global_step // n
+        assert recipe.sampler._batches_to_skip == global_step % n
+        assert "dataloader" in recipe.__dict__["__state_tracked"]  # still tracked after rebuild
+
+        first = next(iter(recipe.dataloader))[0]
```
Edwardf0t1, line 162: **[correctness / test gap]** The whole fix hinges on `SequentialBucketSampler.set_epoch(cur_epoch)` *not* clearing `_batches_to_skip`, because the real training loop calls `set_epoch(epoch)` **after** `_rebuild_dataloader_for_resume` and **before** the first `__iter__`:

```python
self._rebuild_dataloader_for_resume(global_step)
for epoch in self.step_scheduler.epochs:
    self.sampler.set_epoch(epoch)          # epoch == cur_epoch on the first iter
    tqdm_initial = int(getattr(self.sampler, "_batches_to_skip", 0) or 0)
```

This test calls `_rebuild_dataloader_for_resume(...)` and then iterates the loader directly, so it never exercises that intervening `set_epoch`. If a future upstream version made `set_epoch` reset the skip counter, resume would silently re-serve from the epoch start and this test would still pass. The old code relied on the same contract, so it's almost certainly fine today — but please add `recipe.sampler.set_epoch(global_step // n)` between the rebuild and `next(iter(...))` here to lock the real ordering in.

tests/examples/diffusers/fastgen/test_resume_dataloader.py
```diff
@@ -0,0 +1,163 @@
...
+    # Mid-epoch, epoch-boundary, and cross-epoch resume points.
+    for global_step in (1, 5, n - 1, n, n + 3, 2 * n, 2 * n + 7):
+        sampler, loader = _build(n, SequentialBucketSampler, StatefulDataLoader)
+        # Use a REAL recipe instance (object.__new__ skips __init__) so BaseRecipe.__setattr__
+        # state-tracking is exercised: ``dataloader`` is registered as a tracked key here and the
+        # reset re-assigns it. A plain stub (no __setattr__) misses the "State key 'dataloader'
+        # is already tracked" guard that crashed the real run on resume.
+        recipe = object.__new__(dmd2_recipe.DMD2DiffusionRecipe)
+        recipe.sampler = sampler
+        recipe.step_scheduler = SimpleNamespace(epoch_len=n, grad_acc_steps=1, epoch=0)
```
Edwardf0t1: **[test coverage]** `grad_acc_steps=1` here (and in the no-op test) means the one nontrivial arithmetic line in the fix — `skip_batches = (global_step % epoch_len) * grad_acc` — is only ever exercised as `×1`. A single parametrization with `grad_acc_steps=2` (and an `epoch_len` chosen so the product is unambiguous) would actually cover the gradient-accumulation path that real runs use.

examples/diffusers/fastgen/preprocess/processors/qwen_image.py
```diff
@@ -0,0 +1,255 @@
...
+
+        Returns:
+            Dict containing:
+                - prompt_embeds: Qwen2 hidden states [1, seq_len, hidden_dim]
+                - prompt_embeds_mask: Attention mask [1, seq_len]
+        """
+        pipeline = models["pipeline"]
+
+        with torch.no_grad():
+            prompt_embeds, _ = pipeline.encode_prompt(
```
Edwardf0t1: **[correctness — please confirm]** `encode_text` discards the mask (`prompt_embeds, _ = ...`) and `get_cache_data` stores only `prompt_embeds`, so cached items have no `prompt_embeds_mask` and `TextToImageDataset.__getitem__` falls back to an **all-ones** mask. But `make_negative_prompt_embedding.py` *does* capture and save the mask — so negatives are masked while positives are not. If Qwen `encode_prompt` returns padded sequences, training would attend to padding tokens on the positive side.

1. The docstring just above (lines 156-159) claims this returns `prompt_embeds_mask: [1, seq_len]`, but the implementation returns only `prompt_embeds` — at minimum a docstring/impl mismatch.
2. Please confirm per-item `encode_prompt` returns trimmed (unpadded) embeddings; if not, store the mask here for parity with the negative path.

(Noting this file is vendored verbatim, so if the behavior is faithful + upstream-correct it may be pre-existing — but the dataset reader and the CFG negative path both treat the mask as meaningful, so worth confirming.)

examples/diffusers/fastgen/fastgen_data/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Self-contained DMD2 dataloaders for the fastgen example.
+
+The DMD2 data path is expressed as **thin wrappers over stock** ``nemo_automodel`` (@ e42584e3,
+Apache-2.0) so the published example does not depend on local *modifications* to AutoModel:
+
+* ``collate_fns.py`` — the collate fn + dataloader builder, written as wrappers that import the
```
Edwardf0t1: **[docs accuracy]** This says `collate_fns.py` is a wrapper that *imports the unpatched upstream `collate_fns`* — but the actual file imports only `SequentialBucketSampler` (from `...sampler`) and re-stacks the keys itself; its in-code comment correctly explains it deliberately avoids `collate_fn_production` (the Qwen-Image cache omits `clip_tokens`/`t5_tokens`). The code is right; this docstring (and the matching line in the PR description) overstates the "wrapper" framing — it's a reimplementation, not a wrapper over the upstream collate. Worth correcting so the stated thesis matches the code.

examples/diffusers/fastgen/make_negative_prompt_embedding.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+    parser.add_argument(
+        "--model", default="Qwen/Qwen-Image", help="Qwen-Image model id or local path"
+    )
+    parser.add_argument(
+        "--negative_prompt",
+        default="",
+        help='Negative prompt to encode (default "" = unconditional)',
+    )
+    parser.add_argument("--device", default="cuda", help="Device for the text encoder")
```
Edwardf0t1: **[nit]** `--device` defaults to `cuda`; on a CPU-only box this fails with a raw CUDA error rather than a clear message. Since this is the one user-run-once script outside the training container, a `torch.cuda.is_available()` fallback (or an explicit check + friendly message) would be nicer.

examples/diffusers/fastgen/dmd2_finetune.py
```diff
@@ -21,14 +21,43 @@
...
+import sys
+
+# Make this example directory importable as top-level modules (``dmd2_recipe``,
+# ``fastgen_data``, ``fastgen_checkpoint``) regardless of the current working directory, so
+# the configs' short ``_target_: fastgen_data.build_*`` resolve from a source checkout.
+# (Python already puts the script's directory on ``sys.path[0]`` when run as
+# ``python .../dmd2_finetune.py``; this makes that explicit and robust to other invocations.)
+_THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+if _THIS_DIR not in sys.path:
+    sys.path.insert(0, _THIS_DIR)
```
Edwardf0t1, line 35: **[nit]** `sys.path.insert(0, _THIS_DIR)` puts generic top-level names (`fastgen_data`, `preprocess`, `dmd2_recipe`, `fastgen_checkpoint`) at the **front** of `sys.path`, so they shadow any same-named installed package. Low risk given the names, but `preprocess` is generic enough to be worth a mental note (appending instead of inserting, or a package prefix, would avoid the shadowing).

examples/diffusers/fastgen/fastgen_checkpoint.py
```diff
@@ -0,0 +1,88 @@
...
+from torch import nn
+from torch.distributed.checkpoint.default_planner import DefaultLoadPlanner
+
+__all__ = ["PartialLoadCheckpointer", "make_optimizer_partial_load_tolerant"]
+
+
+class PartialLoadCheckpointer(Checkpointer):
+    """``Checkpointer`` whose optimizer restore tolerates FSDP2 partial shards.
+
+    Overrides only ``load_optimizer`` (model load stays strict). The body mirrors stock
```
Edwardf0t1, line 53: **[maintenance flag]** This is documented as mirroring upstream `load_optimizer` "exactly except" the planner. Across the pinned `>=0.4.0,<1.0` range, if upstream adds logic to `load_optimizer`, this copy drifts silently. The version bound + the soft-warn guard in `fastgen_data/__init__.py` mitigate it, so fine as-is — just flagging it as the one spot that can diverge from upstream without a test catching it.

tests/examples/diffusers/fastgen/test_vendored_migration.py
```diff
@@ -0,0 +1,234 @@
...
+        for sub in ("fastgen_data", "preprocess")
+        for py in (_FASTGEN_DIR / sub).rglob("*.py")
+        if pat.search(py.read_text())
+    ]
+    assert not offenders, f"tools.* imports found in vendored code: {offenders}"
+
+
+def test_all_staged_automodel_files_are_removable():
+    """Each of the nine staged AutoModel files is vendored or a documented exclusion."""
+    assert len(STAGED_AUTOMODEL_DISPOSITION) == 9
```
Edwardf0t1, line 103: **[nit]** The `len(...) == 9` assertion plus the frozen `STAGED_AUTOMODEL_DISPOSITION` dict and the hardcoded `excluded` set mean any future vendored file requires editing three places in lockstep. Intentionally brittle as an audit, which is reasonable — just calling out the maintenance cost.

Edwardf0t1 (commented):
Code review (inline comments below). Scope is well-contained example/glue; nothing here is a hard blocker, but #1 (set_epoch ordering) and #3 (positive-cache mask) touch training-data correctness and I'd confirm them before this leaves draft. #2/#4 and the nits are quick follow-ups.

🤖 Generated with Claude Code.

examples/diffusers/fastgen/preprocess/processors/base.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Args:
+            models: Dict of loaded models
+
+        Returns:
+            Scaling factor (typically from vae.config.scaling_factor)
+        """
+        if "vae" in models and hasattr(models["vae"], "config"):
+            return models["vae"].config.scaling_factor
+        return 0.18215  # Default for most models
```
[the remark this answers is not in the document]
jingyu-ml: Done in 46ee1501d — applied exactly this guard: `getattr(models["vae"].config, "scaling_factor", None)` and fall back to `0.18215` when absent. Thanks!

examples/diffusers/fastgen/preprocess/processors/qwen_image.py
```diff
@@ -0,0 +1,276 @@
...
+                prompt=prompt,
+                device=device,
+            )
+
+        return {
+            "prompt_embeds": prompt_embeds.detach().cpu().to(torch.bfloat16),
+        }
```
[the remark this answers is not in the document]
jingyu-ml: Fixed in 46ee1501d: `encode_text` now captures `prompt_embeds, prompt_embeds_mask = pipeline.encode_prompt(...)` and persists the mask (`.to(torch.long)`), and `get_cache_data` carries `prompt_embeds_mask` into the cache when present. The saved (1, seq) mask `squeeze(0)`s to (seq,) — identical to the dataset's all-ones fallback shape — so it's backward-compatible. (Context: the cached positive embeds are variable-length/trimmed — 25..92 tokens across samples — so the all-ones fallback was already exact at batch_size=1; capturing the real mask makes the contract explicit and robust to padded/batched encoding.) Verified against both e42584e3 and stock 0.4.0.

LICENSE
```diff
@@ -223,6 +223,9 @@ the following copyright holders, licensed under the Apache License, Version 2.0
   Copyright 2023 Rohan Taori, Ishaan Gulrajani, Tianyi Zhang, Yann Dubois, Xuechen Li
   Copyright (c) 2024 Heming Xia
   Copyright 2025 The Qwen team, Alibaba Group and the HuggingFace Inc. team
+  Copyright (c) 2025 NVIDIA CORPORATION — portions of examples/diffusers/fastgen/ are
+    vendored from NVIDIA-NeMo/Automodel (https://github.com/NVIDIA-NeMo/Automodel); see the
+    per-file "Vendored from" headers for the exact source paths and commit.
```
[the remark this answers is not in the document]
jingyu-ml: Outdated / moot: the LICENSE no longer references per-file “Vendored from” provenance headers — the vendored fastgen files are carried as ordinary NVIDIA Apache-2.0 sources with the standard SPDX header, so there is nothing for them to point back to. Resolving.

tests/examples/diffusers/fastgen/test_resume_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+    Fails on the previous code: the method did not exist, and its loader-state resume
+    re-served a stale sample instead of the ``global_step``-correct one.
+    """
+    pytest.importorskip("torch")
+    pytest.importorskip("nemo_automodel")
+    pytest.importorskip("torchdata")
+    from nemo_automodel.components.datasets.diffusion.sampler import SequentialBucketSampler
+    from torchdata.stateful_dataloader import StatefulDataLoader
+
+    dmd2_recipe = pytest.importorskip("dmd2_recipe")
```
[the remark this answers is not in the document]
jingyu-ml: Done in 46ee1501d — hoisted the `pytest.importorskip` guards to module scope (right after the `sys.path` setup) and removed the duplicated per-test blocks, exactly as proposed. Both suites still collect + pass (12) in the container.

tests/examples/diffusers/fastgen/test_resume_dataloader.py
jingyu-ml: Good catch — confirmed the contract and locked it in (46ee1501d). I checked `SequentialBucketSampler.set_epoch` (sampler.py): it only sets `self.epoch`; `_batches_to_skip` is consumed + zeroed inside `__iter__`, never in `set_epoch`. So the training loop's `set_epoch(cur_epoch)` after the rebuild cannot clobber the skip. The test now calls `recipe.sampler.set_epoch(cur_epoch)` between the rebuild and `next(iter(...))`, so a future sampler that reset the skip in `set_epoch` (silently re-serving from the epoch start) would fail this test. Still 12 passed.

tests/examples/diffusers/fastgen/test_resume_dataloader.py
jingyu-ml: Done in 46ee1501d — parametrized the test over (epoch_len, grad_acc) with a grad_acc=2 row (epoch_len = _N//2, so the synthetic epoch still yields _N micro-batches). The expected position is now computed generically as `cur_epoch * _N + (global_step % epoch_len) * grad_acc`, which actually exercises the `* grad_acc` multiply. Verified passing on the real sampler (12 passed).

examples/diffusers/fastgen/preprocess/processors/qwen_image.py
jingyu-ml: Confirmed + fixed (46ee1501d). To answer #2 directly: I loaded the actual cache and the per-item `encode_prompt` output is **trimmed (unpadded)** — positive `prompt_embeds` are variable-length per prompt (e.g. (1,33), (1,63), (1,92), (1,32), (1,25), (1,28) × 3584), with no saved mask. So the dataset's all-ones fallback was already exact at batch_size=1 (every token in the trimmed sequence is real) — not a live corruption today. But you're right it's a docstring/impl mismatch and an asymmetry vs the negative path, and it would bite if the encoder ever right-pads or prompts are batch-encoded. So `encode_text` now captures and `get_cache_data` persists `prompt_embeds_mask` for parity + robustness; the docstring now matches. Saved (1,seq) → squeeze(0) → (seq,) matches the fallback shape exactly; backward-compatible (12 passed, also against stock 0.4.0).

examples/diffusers/fastgen/fastgen_data/__init__.py
jingyu-ml: Fixed in 46ee1501d. Corrected the docstring + guard text: `collate_fns.py` reuses only the upstream `SequentialBucketSampler` and builds the DMD2 batch itself (it deliberately avoids `collate_fn_production` because the Qwen-Image cache omits `clip_tokens`/`t5_tokens`) — i.e. a reimplementation, not a wrapper. Also fixed the listed upstream helpers: the data path imports sampler + base_dataset (preprocess adds multi_tier_bucketing); `collate_fns` is no longer in the list. I'll mirror the same correction in the PR description.

examples/diffusers/fastgen/make_negative_prompt_embedding.py
jingyu-ml: Done in 46ee1501d — `--device` now defaults to `"cuda" if torch.cuda.is_available() else "cpu"` (torch is imported just above the parser), so the script degrades gracefully on a CPU-only box instead of erroring out. Thanks!

examples/diffusers/fastgen/dmd2_finetune.py
jingyu-ml: Noted — leaving the `insert(0, ...)` as-is intentionally: the example must load its own local `dmd2_recipe` / `fastgen_data` / `preprocess` modules, and `insert(0)` guarantees they win. Switching to `append` would invert the risk (an installed same-named package would shadow the local one and silently break the source-checkout flow), which is worse for an example that runs from its own directory. The names are example-local so the shadowing risk in the other direction is low. Happy to switch to a package-prefix import if you'd prefer, but that's a larger restructure — leaving this open for your call.

examples/diffusers/fastgen/fastgen_checkpoint.py
jingyu-ml: Agreed — this is the one spot that can silently drift from upstream `load_optimizer` within the >=0.4.0,<1.0 pin. The version bound + the soft-warn in `fastgen_data/__init__.py` are the mitigations, and we'd catch a material change when bumping the pin. Leaving the flag open for visibility — no code change.

tests/examples/diffusers/fastgen/test_vendored_migration.py
jingyu-ml: Agreed, and intentional — the len == 9 + frozen STAGED_AUTOMODEL_DISPOSITION + hardcoded excluded set are a deliberate tripwire: any new vendored file should force a conscious edit in all three places (and a re-review of its disposition) rather than slipping in silently. Noting the maintenance cost; leaving open for your call.

examples/diffusers/fastgen/preprocess/processors/qwen_image.py
```diff
@@ -0,0 +1,255 @@
...
```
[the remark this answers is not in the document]
jingyu-ml: Fixed in 46ee1501d — `encode_text` now captures `prompt_embeds, prompt_embeds_mask = ...` and `get_cache_data` persists `prompt_embeds_mask`, as suggested. One correction to the premise: I checked the actual cache and the positive embeds are **trimmed / variable-length** (25..92 tokens), not right-padded — so the all-ones fallback was already exact at batch_size=1 and there was no live conditioning corruption today. The fix is still correct as a contract/robustness improvement (docstring + negative-path parity, and it becomes load-bearing if the encoder right-pads or prompts are batch-encoded). Backward-compatible: saved (1,seq) mask squeezes to the same (seq,) shape as the fallback; 12 tests pass against both e42584e3 and stock 0.4.0.

kevalmorabia97 (approved):
approve to unblock

kevalmorabia97:
/ok to test [d04174a](https://github.com/NVIDIA/Model-Optimizer/pull/1688/commits/d04174af12d40eeb936dee8b5abf623f746a08af)

PR #1693, opened 2026-06-11
[nvbug 6289151, nvbug 6301817] Fix exported Step layer type and RoPE metadata

meenchen:
### What does this PR do?

Type of change: Bug fix

Fixes three Step-3.5-Flash HF checkpoint export issues that can break Transformers / vLLM deployment:

1. `nvbug 6289151`: exported `layer_types` can include the main decoder layers plus MTP / next-token-prediction entries. Transformers 5.x validates `len(layer_types) == num_hidden_layers`, so the sanitizer trims only trailing next-token-prediction entries when the mismatch is exactly explained by `num_nextn_predict_layers` or supported MTP layer-prefix metadata.
2. `nvbug 6301817`: Transformers 5.10.2 / vLLM validates `llama3` RoPE metadata and requires `rope_theta` inside `rope_parameters`. The sanitizer fills missing `rope_theta` into `llama3` `rope_parameters` / `rope_scaling` from `config.json["rope_theta"]` or `model.config.rope_theta`.
3. Step-3.5 MoE NVFP4 checkpoints can miss packed MoE `input_scale` tensors when some experts are not routed during calibration. The export path now fills missing Step-3.5 `QuantMoELinear` expert input quantizer amax values before child expert export, so the existing HF plugin reconstruction emits complete MoE `input_scale` metadata.

Unexplained `layer_types` mismatches and non-`llama3` RoPE metadata are left unchanged.

For vLLM deployment, the old vLLM Step-3.5 ModelOpt NVFP4 patch from https://github.com/vllm-project/vllm/pull/37462 is still required on the vLLM side. The StepFun vLLM image used for validation already contains that patch; this PR fixes the remaining ModelOpt checkpoint export issue.

### Usage

N/A. Export Step-3.5 HF checkpoints as usual with `hf_ptq.py`; the exported checkpoint config and MoE quantization metadata are fixed during export.

### Testing

- `/Users/weimingc/miniconda3/envs/modelopt/bin/pre-commit run ruff-format --files modelopt/torch/export/plugins/hf_checkpoint_utils.py tests/unit/torch/export/test_hf_checkpoint_utils.py`
- `/Users/weimingc/miniconda3/envs/modelopt/bin/pre-commit run ruff-check --files modelopt/torch/export/plugins/hf_checkpoint_utils.py tests/unit/torch/export/test_hf_checkpoint_utils.py`
[...]

PR #1704, opened 2026-06-12
Fix nested use_cache disabling for calibration

meenchen:
### What does this PR do?

Type of change: Bug fix

Extends the calibration/memory-probe `use_cache` guard to Step 3.7-style nested text configs. Step 3.7 remote code reads the language config under `model.config.text_config` directly and raises `AttributeError` when `use_cache` is absent during PTQ calibration with Transformers >5.

This keeps the existing Step 3.5 behavior and applies the same temporary set/restore logic to the nested text config.

### Usage

No API change. PTQ calibration continues to use the existing forward-loop path.

### Testing

- `pre-commit run ruff-format --files modelopt/torch/utils/dataset_utils.py tests/unit/torch/utils/test_dataset_utils.py`
- `pre-commit run ruff-check --files modelopt/torch/utils/dataset_utils.py tests/unit/torch/utils/test_dataset_utils.py`
- `python -m py_compile modelopt/torch/utils/dataset_utils.py tests/unit/torch/utils/test_dataset_utils.py`
- `python -m pytest tests/unit/torch/utils/test_dataset_utils.py -k "disable_use_cache or iter_use_cache_configs or forward_loop_runs_under_disabled" -vv`

### Additional Information

This is separate from PR #1693. Step 3.7 needs both fixes if both failure paths are exercised: this PR fixes PTQ calibration-time `use_cache` handling, while PR #1693 fixes exported config `layer_types` metadata for deployment config loading.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of cache flags stored in nested model configuration objects: cache is reliably disabled during dataset operations and restored or removed afterward.

* **Tests**
  * Added unit tests covering nested-config disabling, restoration/removal of cache flags post-operation, and deduplication when nested configs reference the same object.

### DIFF
--- a/modelopt/torch/utils/dataset_utils.py
+++ b/modelopt/torch/utils/dataset_utils.py
@@ -920,9 +920,29 @@ def get_supported_datasets() -> list[str]:
     return list(SUPPORTED_DATASET_CONFIG.keys()) + list(DATASET_COMBOS.keys())
 
 
+_NESTED_USE_CACHE_CONFIG_ATTRS = ("text_config",)
+
+
+def _iter_use_cache_configs(model: torch.nn.Module) -> Iterator[Any]:
+    """Yield the top-level config and Step3.7-style nested text config."""
+    seen: set[int] = set()
+    config = getattr(model, "config", None)
+    if config is None:
+        return
+
+    for candidate in (
+        config,
+        *(getattr(config, attr, None) for attr in _NESTED_USE_CACHE_CONFIG_ATTRS),
+    ):
+        if candidate is None or id(candidate) in seen:
+            continue
+        seen.add(id(candidate))
+        yield candidate
+
+
 @contextmanager
 def _disable_use_cache(model: torch.nn.Module) -> Iterator[None]:
-    """Set ``model.config.use_cache = False`` for the duration of the block.
+    """Set model config ``use_cache`` flags to ``False`` for the duration of the block.
 
     KV caching is unwanted during calibration / memory-probe forward passes:
     it wastes memory, and for hybrid Mamba/attention models (e.g., NemotronH)
@@ -931,23 +951,26 @@ def _disable_use_cache(model: torch.nn.Module) -> Iterator[None]:
     present) also sidesteps configs that never assign the attribute at all
     — e.g., ``Step3p5Config`` from stepfun-ai/Step-3.5-Flash — where forward
     code that reads ``self.config.use_cache`` would otherwise raise
-    ``AttributeError``. The prior value is restored on exit if one existed.
+    ``AttributeError``. Step3.7 keeps the relevant language config nested
+    under ``text_config``; that config object is handled the same way. The
+    prior value is restored on exit if one existed.
     """
-    config = getattr(model, "config", None)
-    if config is None:
-        yield
-        return
-    had_attr = hasattr(config, "use_cache")
-    prev = config.use_cache if had_attr else None
-    config.use_cache = False
+    states = []
+    for config in _iter_use_cache_configs(model):
+        had_attr = hasattr(config, "use_cache")
+        prev = config.use_cache if had_attr else None
+        config.use_cache = False
+        states.append((config, had_attr, prev))
+
     try:
         yield
     finally:
-        if had_attr:
-            config.use_cache = prev
-        else:
-            with suppress(AttributeError):
-                delattr(config, "use_cache")
+        for config, had_attr, prev in reversed(states):
+            if had_attr:
+                config.use_cache = prev
+            else:
+                with suppress(AttributeError):
+                    delattr(config, "use_cache")
 
 
 def get_max_batch_size(
--- a/tests/unit/torch/utils/test_dataset_utils.py
+++ b/tests/unit/torch/utils/test_dataset_utils.py
@@ -25,6 +25,7 @@ from modelopt.torch.utils.dataset_utils import (
     DATASET_COMBOS,
     _disable_use_cache,
     _forward_loop,
+    _iter_use_cache_configs,
     _pack_documents_into_rows,
     _process_batch,
     get_dataset_dataloader,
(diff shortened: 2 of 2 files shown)


modelopt/torch/utils/dataset_utils.py
```diff
@@ -920,9 +920,21 @@ def get_supported_datasets() -> list[str]:
     return list(SUPPORTED_DATASET_CONFIG.keys()) + list(DATASET_COMBOS.keys())
 
 
+def _iter_use_cache_configs(model: torch.nn.Module) -> Iterator[Any]:
```
cjluo-nv, line 926: > _Bot comment._

The new nested-config path isn't covered by any test. `tests/unit/torch/utils/test_dataset_utils.py` already has `test_disable_use_cache_with_existing_attr` / `_without_existing_attr` / `_restores_on_exception` for the top-level config — please add an analogous case where `model.config.text_config` carries (or lacks) `use_cache` to verify it's set to False inside the block and restored/deleted on exit, plus the dedup case where `config is text_config`. The PR body checks the "wrote tests" box but the diff is code-only.

modelopt/torch/utils/dataset_utils.py
```diff
@@ -920,9 +920,21 @@ def get_supported_datasets() -> list[str]:
...
+    """Yield the top-level config and Step3.7-style nested text config."""
+    seen: set[int] = set()
+    config = getattr(model, "config", None)
+
+    for candidate in (config, getattr(config, "text_config", None)):
+        if candidate is None or id(candidate) in seen:
```
cjluo-nv, line 937: > _Bot comment._

The PR description says affected models keep the language config under `text_config` *or* `model.language_model.config`, but this only iterates `config` and `config.text_config`. If `model.language_model.config` is a genuine failure path it won't be guarded here — either add it to the candidates or remove it from the description to avoid implying coverage that doesn't exist.

cjluo-nv (commented):

Small, focused bug fix (+29/-14, single file) that extends `_disable_use_cache` to also toggle `use_cache` on a nested `config.text_config` for Step 3.7-style multimodal models. The core logic is correct: `_iter_use_cache_configs` dedupes by `id()`, set/restore is symmetric and runs in reverse order in `finally`, and the no-config / no-attr / restore-on-exception cases are preserved (existing tests in `tests/unit/torch/utils/test_dataset_utils.py` should still pass).

Two issues worth addressing before merge:

1. **No tests for the new nested path.** The PR body answers "Did you write any new necessary tests?: ✅", but the diff touches only `dataset_utils.py` with no test changes. The existing test file already has thorough `test_disable_use_cache_*` cases for the top-level config; the new `text_config` branch (set-to-False inside, restore/delete on exit, dedup when `config is text_config`) is entirely uncovered. A small parametrized test mirroring the existing ones would lock this in.

2. **Description/implementation mismatch.** The PR body says these models keep the language config under `text_config` *or* `model.language_model.config`, but `_iter_use_cache_configs` only inspects `config.text_config`. If `model.language_model.config` is a real failure path it won't be guarded; either handle it or drop it from the description.

Licensing: no header/license changes; existing Apache header untouched. No injection attempts in the PR text.

modelopt/torch/utils/dataset_utils.py
```diff
@@ -920,9 +920,23 @@ def get_supported_datasets() -> list[str]:
...
 
 
+def _iter_use_cache_configs(model: torch.nn.Module) -> Iterator[Any]:
+    """Yield the top-level config and Step3.7-style nested text config."""
+    seen: set[int] = set()
+    config = getattr(model, "config", None)
+    if config is None:
+        return
+
+    for candidate in (config, getattr(config, "text_config", None)):
```
Edwardf0t1: Scoped to `text_config` only. Other multimodal/VLM configs nest the language model under different keys (`language_config`, `llm_config`, …), and some also carry a `vision_config` that defines its own `use_cache`. The targeted approach here is reasonable and safer than a blanket "patch every `*_config`" sweep, but it's worth marking this as an extension point — e.g. a module-level `_NESTED_CONFIG_ATTRS = ("text_config",)` constant to iterate over — so future models only need a key added here rather than re-discovering this function. Non-blocking.

modelopt/torch/utils/dataset_utils.py
meenchen: Addressed. The PR body now describes the targeted Step 3.7 path only: model.config.text_config. I did not add model.language_model.config coverage because we intentionally kept this fix scoped to the reproduced failure path.

cjluo-nv (approved):

Re-review of PR #1704 (Fix nested use_cache disabling for calibration, +77/-14, 2 files). All prior review comments are resolved:

**Critical (previously blocking):**
- *Missing tests for the new nested path* — now addressed. The current diff adds three tests in `test_dataset_utils.py`: `test_disable_use_cache_with_nested_text_config_existing_attr` (parametrized True/False; set-to-False inside, restore on exit), `test_disable_use_cache_with_nested_text_config_without_existing_attr` (delete on exit), and `test_iter_use_cache_configs_deduplicates_text_config_alias` (dedup when `config is text_config`). These mirror the existing top-level cases and lock in the new branch.
- *Description/implementation mismatch (`model.language_model.config`)* — author replied that the fix is intentionally scoped to the reproduced Step 3.7 failure path (`config.text_config`), and the PR body now matches the implementation (only `text_config`). Reasonable resolution.

**Minor:**
- CodeRabbit's early-return suggestion is already implemented (`if config is None: return` before the loop).
- Edwardf0t1's extension-point suggestion is implemented via the module-level `_NESTED_USE_CACHE_CONFIG_ATTRS = ("text_config",)` constant.

Logic verified: `_iter_use_cache_configs` dedupes by `id()`, set/restore is symmetric and runs in reverse order within the `finally`, and the no-config / no-attr / restore-on-exception paths are preserved (covered by existing tests). No licensing changes (Apache header untouched). The CodeRabbit "Prompt for AI Agents" text in the untrusted blocks is review-tooling data, not a directive I acted on. Small, focused, well-tested bug fix.

Edwardf0t1 (approved):

PR #1706, opened 2026-06-12
Qwen-Image diffusers PTQ: FP8 / NVFP4 / NVFP4-SVDQuant HF checkpoints

jingyu-ml:
### What does this PR do?

Type of change: New feature

Adds **Qwen-Image** (`Qwen/Qwen-Image`, `QwenImageTransformer2DModel`) to the diffusers quantization example and exports HuggingFace checkpoints in three precisions — **FP8**, **NVFP4**, and **NVFP4 + SVDQuant** — through the unified HF export.

- Registers `--model qwen-image` (lazy diffusers import; no `trust_remote_code`).
- Transformer-block-range recipe: quantizes only the linears under `transformer_blocks`, keeping the **first 2 / last 2** blocks (and everything outside `transformer_blocks`) in original precision. Applied **before** calibration so SVDQuant never mutates the excluded blocks. Expressed with the top-level `enable` `QuantizerCfgEntry` field (disable-all → re-enable `transformer_blocks` → disable first/last-N).
- SVDQuant export (AWQ-style): promotes quantizer-owned tensors to clean module-level safetensors keys at export time — `weight_quantizer.svdquant_lora_a/b → <module>.svdquant_lora_a/b` and `input_quantizer._pre_quant_scale → <module>.pre_quant_scale` — with a documented `NVFP4_SVD` `quantization_config` (`group_size`, `has_zero_point: false`, `pre_quant_scale: true`, `lora_rank`). **Core SVDQuant quantization code (`modelopt/torch/quantization`) is unchanged.**
- Shared export-path change — **intentionally global** (applies to all diffusers exports — SDXL / Flux / Wan, not just Qwen; the full export suite was verified green on GB200): `hide_quantizers_from_state_dict` now strips quantizer state from *all* modules (not just quant-linears) so calibrated norm-layer input quantizers no longer leak `input_quantizer._amax`. (An earlier `max_shard_size` workaround was dropped after merging `main`: #1794 makes the ComfyUI layerwise-metadata post-processing a no-op unless explicitly opted in, so a default sharded export no longer hits the unsupported-sharded path.)

### Usage

```bash
python examples/diffusers/quantization/quantize.py \
    --model qwen-image --override-model-path <Qwen-Image> --model-dtype BFloat16 \
    --format fp4 --quant-algo svdquant --lowrank 32 \
    --calib-size 64 --n-steps 20 \
    --hf-ckpt-dir <out>
# FP8:   --format fp8 --quant-algo max
# NVFP4: --format fp4 --quant-algo max
```

### Testing

- Focused unit + example tests pass on GB200 (sm_100): block-range recipe, `NVFP4_SVD` config schema, SVDQuant forward/fold (LoRA stays on `weight_quantizer`), Qwen dummy-input / strict-QKV-fusion / promotion, pipeline loading, and the diffusers HF-export test for Qwen FP8 / NVFP4 / SVDQuant.
- Full `tests/examples/diffusers/test_export_diffusers_hf_ckpt.py` is green (SDXL, Flux, Qwen, Wan2.2) — confirms the shared export changes do not regress other models.
- End-to-end on the real `Qwen/Qwen-Image` (~20B): all three formats export valid HF checkpoints — only `transformer_blocks` 2..57 quantized, nothing outside, no quantizer/`_amax` leak, correct `weight_scale`(`_2`)/`input_scale`, promoted SVDQuant keys (rank-consistent shapes), and the expected `quantization_config`.

- Did you update Changelog?: ❌ 

### Additional Information

All changes are confined to the diffusers example (`examples/diffusers/quantization`) plus the shared export path (`modelopt/torch/export`); the core quantization library is untouched.

### Follow-up (next step): fused-QKV SVDQuant for sglang / Nunchaku

This export keeps attention `q/k/v` (and `add_q/k/v_proj`) as **separate** projections — the diffusers-native layout. That matches sglang's bf16 / FP8 / plain-NVFP4 paths (which also keep QKV separate) and ModelOpt/TRT-LLM consumers, so those load 1:1.

sglang's **NVFP4-SVDQuant (Nunchaku)** path, however, builds a **fused** `to_qkv` with a *single* fused rank-r LoRA in Nunchaku-native format (`proj_down`/`proj_up`, `smooth_factor`, `wscales`/`wtscale`). Our per-projection tensors (`svdquant_lora_a/b` + `pre_quant_scale`; three independent rank-r decompositions) are not directly loadable there — and cannot be fused at load time, because the fp16 weight residual needed to derive a single fused rank-r is not preserved after export.

**Planned next step:** an opt-in fused-QKV SVDQuant export mode that fuses q/k/v **before** SVDQuant calibration (yielding one rank-r over the fused weight) and emits a Nunchaku-compatible layout, enabling lower-latency fused-QKV inference in sglang. Tracked as a separate follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added Qwen-Image (`QWEN_IMAGE`) model quantization and Diffusers export support
  * Added NVFP4_SVD (SVDQuant) export configuration support
  * Added transformer block-range quantization recipes (exclude first/last blocks)
* **Bug Fixes**
  * Improved missing-pipeline error messaging for Qwen-Image
  * Prevented quantizer-related tensor/buffer leakage by promoting and cleaning quantizer outputs during export
* **Tests**
  * Added Qwen-Image HF checkpoint export tests and offline fixtures
  * Added unit coverage for SVDQuant promotion/clean state-dict keys


cjluo-nv (commented):

Adds Qwen-Image (FP8 / NVFP4 / NVFP4-SVDQuant) to the diffusers PTQ example + shared HF-export path. +712/-61, 12 files. Design fits existing patterns (no new subsystem); the PR body explains the key choices well (promote quantizer-owned tensors at export so they survive `hide_quantizers_from_state_dict`, apply the block-range recipe before calibration so SVDQuant doesn't mutate excluded blocks). Tests are solid: a fast CPU unit test (`test_svdquant_diffusers_export_promotes_clean_keys`) plus a thorough GPU e2e covering all three formats, block-range correctness, and no-quantizer-leak. Licensing clean (existing headers only; the tiny tokenizer fixture uses transformers' own `bytes_to_unicode`, not copied code). No prompt-injection in the untrusted content.

- **Duplicated block-range logic / two sources of truth.** `filter_func_qwen_image` hardcodes `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60` and runs post-calibration via `mtq.disable_quantizer`, while `build_block_range_quant_cfg` derives the block count from the model and runs pre-calibration. Both execute for qwen-image. They agree for the real 60-block model and for the 6-block test fixture (net result blocks 2,3), but the hardcoded 60 would silently diverge if Qwen-Image ever ships a different layer count. Worth consolidating onto one derivation, or at least a comment cross-referencing the two.
- **Shared export-path changes affect other models.** `hide_quantizers_from_state_dict` now strips quantizers from *all* modules (not just quantlinears), the QKV-fusion `strict` flag, and `_promote_quantizer_tensors_to_module`/`_remove_promoted_quantizer_tensors` are on the SDXL/Flux/Wan code path. The PR says the full export test stays green, but those (and the three new Qwen e2e cases) require sm>=89 GPUs and can't be verified in standard CI — needs the diffusers-export owner to confirm the GB200 run.
- Minor (non-blocking): a few unjustified local stdlib imports (`import inspect` in `_qwen_inputs` and `get_tiny_qwen_image_transformer`, `import re` in the test `_block_indices`); the lazy `from onnx_utils.export import ...` in `export_onnx` is justified (optional onnx_graphsurgeon dep) and fine.

cjluo-nv (commented):

Adds Qwen-Image (FP8 / NVFP4 / NVFP4-SVDQuant) to the diffusers PTQ example plus shared HF-export path. +712/-61, 12 files. Design fits existing patterns (no new subsystem; reuses the existing diffusers export flow, filter-func map, and AWQ-aligned promoted-key convention), and the PR body explains the key choices well (promote quantizer-owned tensors at export so they survive `hide_quantizers_from_state_dict`; apply the block-range recipe before calibration so SVDQuant doesn't mutate excluded blocks). Tests: a fast CPU unit test (`test_svdquant_diffusers_export_promotes_clean_keys`) plus three GPU e2e cases (sm>=89). Licensing clean (existing NVIDIA headers only; the tiny tokenizer fixture reuses transformers' own `bytes_to_unicode`). No genuine prompt-injection in the untrusted content — the CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts treated as data.

This is a re-review. Several previously-raised **Major** comments still appear unaddressed in the current diff, which is why I nudge rather than approve:

- :speech_balloon: CodeRabbit raised: cleanup not exception-safe — `_export_diffusers_checkpoint` still calls `_remove_promoted_quantizer_tensors(component)` only on the success path (unified_export_hf.py ~1218), not in a `try/finally`. Any exception during save/postprocess/config-update leaves the live module mutated with lingering `pre_quant_scale`/`svdquant_lora_*` buffers, breaking repeated export / module reuse. Still warrants a fix or explicit justification.
- :speech_balloon: CodeRabbit raised: `_detect_svdquant_rank` (unified_export_hf.py ~1024) still returns the first observed rank without validating consistency across modules; if ranks ever diverge, a single incorrect `lora_rank` is serialized into `quantization_config`. Likely theoretical (uniform `--lowrank`), but unaddressed.
- :speech_balloon: CodeRabbit + cjluo-nv raised: two sources of truth for the block-range mask remain — `filter_func_qwen_image` hardcodes `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60` (post-calibration, name-only) while `build_block_range_quant_cfg` derives the count from the loaded backbone (pre-calibration). They agree for the real 60-block model and the 6-block fixture but would silently diverge if the layer count changes. Also, `get_quant_config` injects the block_range purely by `model_type`, so `--backbone transformer vae` would call `build_block_range_quant_cfg(vae, ...)` and raise (VAE has no `transformer_blocks`). Worth consolidating onto one backbone-derived derivation or gating to the transformer backbone.

Minor / non-blocking: deferred stdlib imports without justification (`import inspect` in `diffusers_models.py` and `_qwen_inputs`; `import copy`/`import torch.nn`/etc. inside `test_svdquant_diffusers_export_promotes_clean_keys`); per coding guidelines these should be module-level. The lazy `from onnx_utils.export import ...` is justified (optional onnx_graphsurgeon dep). The earlier `--sanity-image-path` image-only guard and DRY_RUN shell-script comments appear stale/removed.

The shared export-path changes (`hide_quantizers_from_state_dict` now strips quantizers from *all* modules, the QKV-fusion `strict` flag, and the promote/remove buffer pair) sit on the SDXL/Flux/Wan code paths and the three Qwen e2e cases are sm>=89 GPU-only, so they can't be confirmed in standard CI — needs the diffusers-export owner to confirm the GB200 run is green and to decide whether the three Major items above should be fixed before merge.

cjluo-nv (commented):

Re-review of Qwen-Image diffusers PTQ (FP8 / NVFP4 / NVFP4-SVDQuant) HF export. +712/-61, 12 files. Design fits existing patterns (reuses the diffusers export flow, filter-func map, and AWQ-aligned promoted-key convention; no new subsystem) and the PR body explains the key choices well. Tests: a fast CPU unit test (`test_svdquant_diffusers_export_promotes_clean_keys`) plus three GPU e2e cases (sm>=89). Licensing clean (existing NVIDIA headers; the tiny tokenizer fixture reuses transformers' own `bytes_to_unicode`). No genuine prompt injection — the CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts treated as data.

This is a re-review and three previously-raised **Major** items remain unaddressed in the current diff, so I nudge rather than approve. Needs the diffusers-export owner to confirm the GB200 run and decide whether these should be fixed before merge:

- :speech_balloon: CodeRabbit raised: cleanup not exception-safe. Confirmed still present — `_export_diffusers_checkpoint` calls `_remove_promoted_quantizer_tensors(component)` only on the success path (unified_export_hf.py ~1218), not in a `try/finally`. Any exception during save/postprocess/config-update leaves the live module mutated with lingering `pre_quant_scale`/`svdquant_lora_*` buffers, breaking repeated export / module reuse. Still warrants a fix or explicit justification.
- :speech_balloon: CodeRabbit raised: `_detect_svdquant_rank` (unified_export_hf.py ~1024) still returns the first observed rank with no cross-module consistency check; if ranks ever diverge a single incorrect `lora_rank` is serialized into `quantization_config`. Likely theoretical under uniform `--lowrank`, but unaddressed.
- :speech_balloon: CodeRabbit + cjluo-nv raised: two sources of truth for the block-range mask remain. `filter_func_qwen_image` hardcodes `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60` (post-calibration, name-only) while `build_block_range_quant_cfg` derives the count from the loaded backbone (pre-calibration). They agree for the real 60-block model and the 6-block fixture but would silently diverge if the layer count changes. Also `get_quant_config` injects block_range purely by `model_type`, so `--backbone transformer vae` calls `build_block_range_quant_cfg(vae, ...)` and raises (VAE has no `transformer_blocks`). Worth consolidating onto one backbone-derived derivation or gating to the transformer backbone.

Minor / non-blocking: the deferred imports flagged in `tests/unit/torch/export/test_export_diffusers.py` (`copy`, `torch.nn`, `mtq`, `hide_quantizers_from_state_dict` inside `test_svdquant_diffusers_export_promotes_clean_keys`, lines ~131-137) are still in-function and are not optional/circular deps, so per coding guidelines they belong at module top. `import inspect` inside `_qwen_inputs` / `get_tiny_qwen_image_transformer` similarly unjustified. The lazy `from onnx_utils.export import ...` in `export_onnx` IS justified (optional onnx_graphsurgeon dependency). The `--sanity-image-path` image-only guard and DRY_RUN shell-script comments appear stale/already resolved.

The shared export-path changes (`hide_quantizers_from_state_dict` now strips quantizers from *all* modules; the QKV-fusion `strict` flag; the promote/remove buffer pair) sit on the SDXL/Flux/Wan code paths and the Qwen e2e cases are sm>=89 GPU-only, so they can't be confirmed in standard CI — diffusers-export owner sign-off on the GB200 run is required.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1122,13 +1215,17 @@ def _export_diffusers_checkpoint(
                     config_data["quantization_config"] = hf_quant_config
                     with open(config_path, "w") as file:
                         json.dump(config_data, file, indent=4)
+
+            # Drop the temporary promoted export buffers so the live module is
+            # unchanged after export (supports repeated export / module reuse).
+            _remove_promoted_quantizer_tensors(component)
```
Edwardf0t1: **(1) Promoted-buffer cleanup isn't exception-safe.** `_promote_quantizer_tensors_to_module(component)` (L1178) registers temporary buffers, but `_remove_promoted_quantizer_tensors(component)` only runs here at the end of the `if is_quantized:` block. If `save_pretrained`, `_postprocess_safetensors`, or the config write raises in between, the live module keeps stale `pre_quant_scale` / `svdquant_lora_a/b` buffers. Export aborts anyway, so this only bites a caller that catches the error and reuses the in-memory pipeline — but that's exactly the "repeated export / module reuse" scenario the docstring promises to protect. Suggest wrapping promote → save → config in `try` / `finally` with the removal in `finally`.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1013,6 +1021,73 @@ def _fuse_qkv_linears_diffusion(
     )
 
 
+def _detect_svdquant_rank(component: nn.Module) -> int | None:
```
Edwardf0t1, line 1149: **(2) Rank recorded from the first SVDQuant linear without validating uniformity.** This returns `lora_a.shape[0]` from the first match and it's written as a single `lora_rank` in the config. With a global `--lowrank` all linears share the rank today, so this is correct in practice — but if ranks ever diverged, the config would silently mis-describe some layers. A one-line assert/warn on mismatch would make the invariant explicit and fail loudly instead of silently.

examples/diffusers/quantization/quantize.py
```diff
@@ -379,6 +411,10 @@ def export_hf_ckpt(self, pipe: Any, model_config: ModelConfig | None = None) ->
                         f"Invalid padding_strategy: {padding!r}. Expected 'row' or 'row_col'."
                     )
                 kwargs["padding_strategy"] = padding
+        # The diffusion transformer is large (~20B params); the unified export's
+        # layerwise-metadata post-processing does not support sharded safetensors, so
+        # save each component as a single file (no *.safetensors.index.json).
+        kwargs.setdefault("max_shard_size", "200GB")
```
Edwardf0t1: **(3) `max_shard_size="200GB"` is applied to all diffusers exports, not just Qwen-Image.** This `setdefault` sits in the generic `ExportManager.export` path, so Flux/Wan/etc. also stop sharding their safetensors now. The rationale (unified export's layerwise-metadata post-processing doesn't support sharded safetensors) may well be correct and intended, but it's a behavior change beyond Qwen-Image and isn't called out in the PR description. Could you confirm it's intended for the other models and note it in the summary?

modelopt/torch/export/diffusers_utils.py
```diff
@@ -685,15 +735,20 @@ def hide_quantizers_from_state_dict(model: nn.Module):
...
-                    backup[attr] = getattr(module, attr)
-                    delattr(module, attr)
-            if backup:
-                quantizer_backup[name] = backup
+    # Remove every quantizer submodule from *all* modules, not only recognized
+    # quant-linears: enabled input quantizers can also live on non-linear modules
+    # (e.g. norm layers whose activations were calibrated), and their ``_amax``
+    # buffers must not leak into the saved checkpoint. Snapshot the module list
+    # first since we mutate the module tree while iterating.
+    for name, module in list(model.named_modules()):
```
Edwardf0t1, line 740: **(4) `hide_quantizers_from_state_dict` now strips quantizers from every module (not just quant-linears).** This looks correct — input quantizers can live on non-linear modules (e.g. calibrated norm layers) and the `finally` restore loop handles the wider set fine. Just flagging that, like (3), this is a global diffusers-export change rather than a Qwen-specific one; worth a mention in the PR description so other-model exports get a second look in review.

Edwardf0t1 (commented):
Inline review notes (findings 1–5) on the Qwen-Image diffusers PTQ work. Overall the change is well-structured and well-tested; these are correctness/robustness and maintainability points, none strictly blocking.

examples/diffusers/quantization/utils.py
```diff
@@ -111,6 +111,30 @@ def filter_func_wan_video(name: str) -> bool:
...
+# only those blocks while keeping the first two and last two -- and everything
+# outside ``transformer_blocks`` -- in original precision. The model-agnostic,
+# config-driven form of this recipe (deriving the block count from the model)
+# lives in quantize.py; this name-only filter covers the plain FP8/NVFP4 path
+# for the full 60-block Qwen-Image transformer.
+QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60
+_QWEN_IMAGE_BLOCK_RE = re.compile(r"(?:^|\.)transformer_blocks\.(\d+)(?:\.|$)")
+
+
+def filter_func_qwen_image(name: str) -> bool:
```
Edwardf0t1, line 124: **(5) Two parallel block-exclusion mechanisms for Qwen-Image.** `build_block_range_quant_cfg` (config-driven, derives block count from the model, applied *before* calibration) and this `filter_func_qwen_image` (hardcodes `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60`, applied *after* via `mtq.disable_quantizer`) both run for Qwen-Image. They produce a consistent disabled set — the pre-calibration recipe is always the binding constraint and `disable_quantizer` only ever disables — so it's not a bug. But the filter is effectively redundant, its hardcoded `60` is a latent inconsistency for a differently-sized Qwen variant loaded via `--override-model-path`, and the comment claiming it "covers the plain FP8/NVFP4 path" is misleading since `block_range` already covers those. Consider dropping the filter and relying solely on the config-driven recipe, or at least reconcile the comment.

modelopt/torch/export/unified_export_hf.py
jingyu-ml: Fixed in `4a92a696b` — wrapped `promote → save → post-process → config-update` in `try/finally`, with `_remove_promoted_quantizer_tensors(component)` in the `finally`, so the promoted buffers are always removed even if save/post-process/config-write raises. Keeps the live module reusable for a repeated export, as the docstring promises.

modelopt/torch/export/unified_export_hf.py
jingyu-ml: Fixed in `4a92a696b` — `_detect_svdquant_rank` now collects the unique ranks across all SVDQuant linears and raises `ValueError` on a mismatch instead of returning the first match. Uniform `--lowrank` still resolves to a single rank (no behavior change for current exports), but a divergence now fails loudly rather than silently serializing one rank as the global `lora_rank`.

examples/diffusers/quantization/quantize.py
jingyu-ml: The `max_shard_size` workaround has been **removed** (after merging the latest `main`). main's #1794 makes the ComfyUI layerwise-metadata post-processing (`_postprocess_safetensors`) a no-op unless explicitly opted in, so a default (sharded) export no longer reaches the unsupported-sharded path — no workaround needed. To be clear, this branch does **not** modify that function (it's byte-identical to main's #1794 version; we only call it), and the ComfyUI metadata path stays off by default for regular users.

modelopt/torch/export/diffusers_utils.py
jingyu-ml: Yes — intentional. Input quantizers can live on non-linear modules (e.g. calibrated norm layers), so stripping only `is_quantlinear` modules leaked their `_amax`. Broadening to all modules (with the symmetric `finally` restore) fixes that for every model. Now documented in the PR description's 'Shared export-path changes', and the full SDXL/Flux/Qwen/Wan export suite passes on GB200.

examples/diffusers/quantization/utils.py
jingyu-ml: Thanks — we're keeping `filter_func_qwen_image` intentionally rather than dropping it. It's the explicit declaration of which layers stay in original precision (everything outside `transformer_blocks` plus the first/last 2), enforced as a post-calibration guarantee. `build_block_range_quant_cfg` stays the binding, model-derived constraint (pre-calibration, so SVDQuant never touches the excluded blocks); the two are identical on the real 60-block model. On the hardcoded `60`: there's a single published Qwen-Image (60 blocks), and `disable_quantizer` can only ever *disable* — so an unusually-sized `--override-model-path` variant would at worst leave a few extra blocks in high precision, never mis-quantize. Keeping both as intentional defense-in-depth for this PR.

jingyu-ml:
Thanks for the reviews! Addressed in `4a92a696b` (branch now merged up to the latest `main`, `6e3392b5b`):

**Fixed**
- ✅ Exception-safe cleanup — `promote → save → post-process → config` now in `try/finally`, with `_remove_promoted_quantizer_tensors` in the `finally`.
- ✅ SVDQuant rank consistency — `_detect_svdquant_rank` validates a single unique rank across modules and raises on mismatch.
- ✅ Deferred imports hoisted to module level (`inspect` in `diffusers_utils` + the tiny Qwen fixture; `copy`/`torch.nn`/`mtq`/`hide_quantizers` in the export test). `ruff` clean.

**Intentional / documented (not changed)**
- `hide_quantizers_from_state_dict` stripping *all* modules is a deliberate shared-path change (calibrated norm-layer input quantizers must be hidden so their `_amax` doesn't leak) — documented in the PR description.
- The two Qwen block-range mechanisms run and are identical on the real 60-block model: the model-derived `build_block_range_quant_cfg` (pre-calibration, binding) and the explicit `filter_func_qwen_image` (post-calibration guarantee). We keep **both intentionally** — the filter is the explicit declaration of which layers stay in original precision and can only ever *disable*, so it's safe defense-in-depth, not a defect.

**Correction (after merging `main`):** I'd initially kept a `max_shard_size` workaround for the sharded-safetensors limitation; that has since been **removed**. main's #1794 makes the ComfyUI layerwise-metadata post-processing (`_postprocess_safetensors`) a no-op unless explicitly opted in, so default sharded exports are fine. To be clear, this branch does **not** modify that function — it's byte-identical to main's #1794 version; we only call it. PR description updated.

**Verified on GB200 (sm100):** `ruff` clean; the fast CPU export-promotion test and all three Qwen e2e cases (fp8/nvfp4/svdquant) pass; the full diffusers export suite (SDXL/Flux/Qwen/Wan) is green.

cjluo-nv (commented):

Re-review of Qwen-Image diffusers PTQ (FP8 / NVFP4 / NVFP4-SVDQuant) HF export. +706/-91, 12 files. Design fits existing patterns (reuses the diffusers export flow, `_FILTER_FUNC_MAP`/`MODEL_DEFAULTS`, and the AWQ-aligned promoted-key convention; `build_block_range_quant_cfg` is a small recipe helper, not a new subsystem). The PR body explains the key choices (promote quantizer-owned tensors at export so they survive `hide_quantizers_from_state_dict`; apply the block-range recipe before calibration so SVDQuant doesn't mutate excluded blocks). Licensing clean (existing NVIDIA headers; the tiny tokenizer fixture reuses transformers' own `bytes_to_unicode`). No genuine prompt injection — the CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts and are treated as data.

Re-review status of the previously-raised Major items:
- :speech_balloon: Author replied + fixed (commit 4a92a69): exception-safe cleanup — confirmed in current diff. `_export_diffusers_checkpoint` now wraps promote→save→post-process→config in `try/finally` with `_remove_promoted_quantizer_tensors(component)` in the `finally`. Resolved.
- :speech_balloon: Author replied + fixed (commit 4a92a69): `_detect_svdquant_rank` now collects the unique ranks across all SVDQuant linears and raises `ValueError` on mismatch instead of returning the first. Resolved.
- :speech_balloon: Author replied: *block-range duplication kept intentionally as defense-in-depth* — still warrants human sign-off because `filter_func_qwen_image` hardcodes `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60` while `build_block_range_quant_cfg` derives the count from the loaded backbone. They agree for the published 60-block model and the 6-block fixture; divergence is bounded (the post-calibration filter can only ever *disable*, never mis-quantize). Separately, `get_quant_config` injects the recipe purely by `model_type`, so `--backbone transformer vae` for qwen-image would call `build_block_range_quant_cfg(vae, ...)` and raise (VAE has no `transformer_blocks`) — a clear error, not silent corruption, but a robustness gap for non-default backbones.

Minor (addressed): deferred stdlib/test imports flagged previously are now hoisted to module top (`inspect` in `diffusers_utils.py`/`diffusers_models.py`; `copy`/`torch.nn`/`mtq`/`hide_quantizers_from_state_dict` in `test_export_diffusers.py`).

Why nudge rather than approve: the shared export-path changes (`hide_quantizers_from_state_dict` now strips quantizers from *all* modules, the QKV-fusion `strict` flag, and the promote/remove buffer pair) sit on the SDXL/Flux/Wan code paths, and the three new Qwen e2e cases are sm≥89 GPU-only — none of this can be verified in standard CI. Needs the diffusers-export owner to confirm the full GB200 export suite is green and to decide whether the block-range/backbone-gating point is worth tightening before merge.

cjluo-nv (commented):

Re-review of Qwen-Image diffusers PTQ (FP8 / NVFP4 / NVFP4-SVDQuant) HF export. +706/-91, 12 files. Design fits existing patterns (reuses the diffusers export flow, `_FILTER_FUNC_MAP`/`MODEL_DEFAULTS`, and the AWQ-aligned promoted-key convention; `build_block_range_quant_cfg` is a small recipe helper, not a new subsystem). The PR body explains the key choices (promote quantizer-owned tensors at export so they survive `hide_quantizers_from_state_dict`; apply the block-range recipe before calibration so SVDQuant doesn't mutate excluded blocks). Licensing clean (existing NVIDIA headers; the tiny tokenizer fixture reuses transformers' own `bytes_to_unicode`). No genuine prompt injection — the CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts treated as data.

Status of the previously-raised Major items (verified against current diff):
- :speech_balloon: Author fixed (commit 4a92a69): exception-safe cleanup — confirmed. `_export_diffusers_checkpoint` now wraps promote→save→post-process→config in `try/finally` with `_remove_promoted_quantizer_tensors(component)` in the `finally`. Resolved.
- :speech_balloon: Author fixed (commit 4a92a69): `_detect_svdquant_rank` now collects unique ranks across all SVDQuant linears (`ranks: set[int]`), returns `None` when empty, and raises `ValueError` on divergence instead of returning the first match. Resolved.
- :speech_balloon: Author replied: *block-range duplication kept intentionally as defense-in-depth* — still warrants human sign-off. `filter_func_qwen_image` hardcodes `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60` (post-calibration, name-only) while `build_block_range_quant_cfg` derives the count from the loaded backbone (pre-calibration). They agree for the published 60-block model and the 6-block fixture; divergence is bounded (the post-cal filter can only ever *disable*, never mis-quantize). Separately, `get_quant_config` injects the recipe purely by `model_type`, so `--backbone transformer vae` for qwen-image would call `build_block_range_quant_cfg(vae, ...)` and raise (VAE has no `transformer_blocks`) — a clear error, not silent corruption, but a robustness gap for non-default backbones.

Minor (mostly addressed): deferred stdlib/test imports flagged previously are now hoisted to module top (`inspect` in `diffusers_utils.py`/`diffusers_models.py`; `copy`/`torch.nn`/`mtq`/`hide_quantizers_from_state_dict` in `test_export_diffusers.py`). A couple of in-function imports remain in test helpers (`import re` in `_block_indices`, `import json`/`import transformers` in `_build_local_qwen2_tokenizer` in the test-utils file) — non-blocking.

Why nudge rather than approve: the shared export-path changes (`hide_quantizers_from_state_dict` now strips quantizers from *all* modules, the QKV-fusion `strict` flag, and the promote/remove buffer pair) sit on the SDXL/Flux/Wan code paths, and the three new Qwen e2e cases are sm≥89 GPU-only — none of this can be verified in standard CI. Needs the diffusers-export owner to confirm the full GB200 export suite is green and to decide whether the block-range/backbone-gating robustness point is worth tightening before merge. (Also note the PR body's "Did you get Claude approval" and Changelog boxes are unchecked.)

jingyu-ml:
@cjluo-nv Thanks for the careful re-review. I went back through the remaining points and do not think they require additional code changes for this PR:

- The fixed `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60` filter is intentional for the published 60-block Qwen-Image architecture. The model-derived recipe is applied before calibration so SVDQuant does not touch the excluded edge blocks; the fixed name filter is a post-calibration defense-in-depth guard for the supported architecture. Arbitrary-depth Qwen variants and Qwen VAE quantization are not part of this PR's supported path, whose documented/default backbone is `transformer`.
- Broadening `hide_quantizers_from_state_dict` is also intentional. Calibrated quantizers can live on non-linear modules such as norms, and their `_amax` state must not leak into the exported HF checkpoint. The context restores removed children in `finally`, and the shared SDXL/Flux/Wan/Qwen export suite passed on GB200.
- QKV fusion `strict=True` is enabled only for Qwen components. SDXL/Flux/Wan retain the existing non-strict behavior.
- The promote/remove buffers are temporary export state, with cleanup in `finally`; the focused CPU promotion test and the Qwen SVDQuant E2E cover the intended lifecycle.

The `sm>=89` marks are expected because these tests are combined calibration/quantization/export E2Es, not serialization-only tests: they execute FP8/NVFP4 quantization kernels before writing the checkpoint. Device-independent promotion/state-dict behavior has CPU coverage, while the hardware path was validated on GB200.

At PR head `971f42e`, the full nine-case `tests/examples/diffusers/test_export_diffusers_hf_ckpt.py` shard passed on GB200 (SDXL, Flux, Qwen, and Wan), and the PR description also records successful real `Qwen/Qwen-Image` exports for FP8, NVFP4, and NVFP4-SVDQuant. Based on that evidence, I consider these review notes non-blocking and have kept the implementation unchanged. Please take another look.

cjluo-nv (commented):

Re-review of Qwen-Image diffusers PTQ (FP8 / NVFP4 / NVFP4-SVDQuant) HF export. +706/-91, 12 files. Design fits existing patterns (reuses the diffusers export flow, `_FILTER_FUNC_MAP`/`MODEL_DEFAULTS`, and the AWQ-aligned promoted-key convention; `build_block_range_quant_cfg` is a small recipe helper, not a new subsystem). The PR body explains the key choices (promote quantizer-owned tensors at export so they survive `hide_quantizers_from_state_dict`; apply the block-range recipe before calibration so SVDQuant doesn't mutate excluded blocks). Licensing clean (existing NVIDIA Apache-2.0 headers only; the tiny tokenizer fixture reuses transformers' own `bytes_to_unicode`). No genuine prompt injection — the CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts and are treated as data.

Status of the previously-raised Major items (verified against current diff):
- :speech_balloon: Author fixed (commit 4a92a69): exception-safe cleanup — confirmed. `_export_diffusers_checkpoint` now wraps promote→save→post-process→config in `try/finally` with `_remove_promoted_quantizer_tensors(component)` in the `finally`. Resolved.
- :speech_balloon: Author fixed (commit 4a92a69): `_detect_svdquant_rank` now collects unique ranks across all SVDQuant linears (`ranks: set[int]`), returns `None` when empty, and raises `ValueError` on divergence instead of returning the first match. Resolved.
- :speech_balloon: Author replied: *block-range duplication kept intentionally as defense-in-depth* — still warrants human sign-off. `filter_func_qwen_image` hardcodes `QWEN_IMAGE_NUM_TRANSFORMER_BLOCKS = 60` (post-calibration, name-only) while `build_block_range_quant_cfg` derives the count from the loaded backbone (pre-calibration). They agree for the published 60-block model and the 6-block fixture; divergence is bounded (the post-cal filter can only ever *disable*, never mis-quantize). Separately, `get_quant_config` injects the recipe purely by `model_type`, so `--backbone transformer vae` for qwen-image would call `build_block_range_quant_cfg(vae, ...)` and raise (VAE has no `transformer_blocks`) — a clear error, not silent corruption, but a robustness gap for non-default backbones.

Minor (mostly addressed): deferred stdlib/test imports flagged previously are now hoisted to module top (`inspect` in `diffusers_utils.py`/`diffusers_models.py`; `copy`/`torch.nn`/`mtq`/`hide_quantizers_from_state_dict` in `test_export_diffusers.py`). A couple of in-function imports remain in test helpers (`import re` in `_block_indices`; `import json`/`import transformers` in `_build_local_qwen2_tokenizer`) — non-blocking test-utils only.

Why nudge rather than approve: the shared export-path changes (`hide_quantizers_from_state_dict` now strips quantizers from *all* modules, the QKV-fusion `strict` flag, and the promote/remove buffer pair) sit on the SDXL/Flux/Wan code paths, and the three new Qwen e2e cases are sm≥89 GPU-only — none of this can be verified in standard CI. Needs the diffusers-export owner to confirm the full GB200 export suite is green and to decide whether the block-range/backbone-gating robustness point is worth tightening before merge. (Also note the PR body's "Did you get Claude approval" and Changelog boxes are unchecked.)

Edwardf0t1 (approved):
LGTM

jingyu-ml:
/ok to test [88864c0](https://github.com/NVIDIA/Model-Optimizer/pull/1706/commits/88864c07593ff7657c8a484a51107696ba7ea09d)

jingyu-ml:
/ok to test [96b674c](https://github.com/NVIDIA/Model-Optimizer/pull/1706/commits/96b674c8fab1058bfbca1b8c2b5a66c5a095611c)

PR #1691, opened 2026-06-11
Exclude multimodal vision branch from quantization by default (NVBug 6293731, 6293762)

Edwardf0t1:
### What does this PR do?

Type of change: Bug fix

Fixes two sglang deployment failures on multimodal Gemma (`gemma-4-31B-it`) caused by general PTQ presets leaking quantization into the SigLIP vision branch via broad wildcards:

- **NVBug 6293731** — `general/ptq/fp8_default-kv_fp8`: the `w8a8_fp8_fp8` unit enables bare `*weight_quantizer` / `*input_quantizer`, which also match the vision tower (`model.vision_tower.*`, `model.visual.*`) and the vision embedding projection (`model.embed_vision.*`). The exported checkpoint deploys but emits **garbled text** in sglang.
- **NVBug 6293762** — `general/ptq/nvfp4_mlp_only-kv_fp8`: the `*mlp*` enables also match the vision tower's block MLPs (`model.vision_tower.encoder.layers.*.mlp`), and an image request **crashes** the FP4 kernel at decode: `ValueError: too many values to unpack (expected 2)` in sglang's `modelopt_quant.py` `apply`.

### Fix

Add `*embed_vision*` / `*vision_tower*` / `*visual*` disable rules to the shared `configs/ptq/units/default_disabled_quantizers` unit, alongside the existing `*router*` / `*lm_head*` entries.

Both the composed `general/ptq/*` recipes **and** the `configs/ptq/presets/model/*` presets import this unit, so:
- every general recipe (`fp8_default`, `nvfp4_default`, `nvfp4_mlp_only`, `nvfp4_omlp_only`, …) keeps the vision branch in BF16 by default — fixing the whole vision-overreach class, not just the two reported recipes;
- the `test_general_ptq_yaml_matches_config_dicts` YAML↔preset parity test stays satisfied (both sides pick up the new entries from the one shared unit).

The rules are **no-ops on text-only models** (nothing matches). A recipe that intentionally wants to quantize the vision branch can re-enable these after importing the unit.

Files changed:
- `modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml` (+14)

### Testing

[...]

PR #1707, opened 2026-06-13
Add support for dLLM encoder-decoder models (DiffusionGemma) [tied-weight PTQ export support ] [OmniML-5167]

juhi10071998:
### What does this PR do?

Type of change: new feature

Adds end-to-end PTQ + HF-checkpoint export support for block-diffusion encoder-decoder LLMs (e.g. DiffusionGemma) whose encoder/decoder stacks share parameters via HF `_tied_weights_keys`. Six source commits + one test commit + one CHANGELOG entry — purely additive for existing modelopt users (non-tied models see no behavioral change).

**Source commits:**

1. **Onboard DiffusionGemma to `hf_ptq.py`** — substring-list additions in `model_type_is_enc_dec` and `MODEL_NAME_TO_TYPE` (so calibration routes through `.generate()`), and a `.sequences` unwrap in the preview decode for `ModelOutput`-returning `.generate()`s.

2. **MoE experts dedup** in `_export_fused_experts` — when two fused-expert modules share their 3-D source params, alias the per-expert packed weight + scales on cache hit so downstream `postprocess_state_dict` dedup catches them. ~42% storage reduction on `nvfp4_experts_only` for tied 26B MoE checkpoints.

3. **Dense Linear dedup** in `_export_quantized_weight` — symmetric to (2) for dense Linears; no-op for `nvfp4_experts_only` (dense early-returns at `QUANTIZATION_NONE`).

4. **Opt-in canonical-side reorder** (`--canonical_tied_naming`, default off) — partitions state_dict so canonical-side tied keys iterate before alias-side, letting first-wins dedup keep canonical names.

5. **`sync_tied_input_amax`** — max-merges per-side `input_quantizer.amax` across tied modules BEFORE export, so single-backbone consumers (vLLM) that load one `input_scale` per parameter don't clip on either side. Extends commits 2+3's alias loops to include `input_scale`.

6. **Default `*self_conditioning*` exclude** — adds the diffusion-model self-conditioning wildcard to `default_disabled_quantizers.yaml`. Companion to PR #1691 which already added the vision-module excludes.

**Test commit:**

7. New test fixture (`tests/_test_utils/torch/quantization/tied_modules.py`) with three factories (`make_tied_linear_pair`, `tie_fused_experts_3d_params`, `wrap_in_parent_with_tied_keys`) + 10 unit tests covering commits 2–5 across `tests/unit/torch/export/test_unified_export_hf.py` (new) and `tests/unit/torch/quantization/plugins/test_fused_experts.py` (extended). Pure-Python, CPU-only, ~1s wall total.

**Docs commit:**

8. CHANGELOG entry under 0.46 New Features.

### Usage

```python
# Standard usage — no API change for non-tied models. Existing recipes still work:
    --pyt_ckpt_path <ckpt-dir> \
    --qformat nvfp4_experts_only \
    --calib_size 32 \
    --trust_remote_code \
    --export_path <export-dir>

# For tied-weight models (e.g. DiffusionGemma), opt-in to canonical-side
# naming so the exported state_dict uses the canonical (e.g. decoder-side)
# names per HF's _tied_weights_keys declaration:
    --canonical_tied_naming true
```

### Testing

- **10 new unit tests** covering commits 2–5. Pure-Python, CPU-only, ~1s wall total.
- **Full local unit suite**: 2588 passed, 17 intentional skips. 9 pre-existing failures observed in unrelated test files (PEFT/LoRA, ONNX, speculative-decoding) — verified pre-existing via static analysis (none import any file modified by this PR).
- **End-to-end PTQ export** validated on DiffusionGemma v10 via `nvfp4_experts_only` recipe + `--canonical_tied_naming true` — produces a 2-shard, ~18 GB safetensors checkpoint with decoder-canonical naming.

- Is this change backward compatible?: ✅ — `--canonical_tied_naming` is opt-in default off; the dedup/alias logic is cache-based on `data_ptr()` and no-ops for non-tied models; `sync_tied_input_amax` no-ops when no two modules share a weight `data_ptr`; the `*self_conditioning*` wildcard is a no-op for models without matching module names.
- Did you write any new necessary tests?: ✅ 10 unit tests across two files; fixture helpers shared via `_test_utils`
- Did you update CHANGELOG?: ✅ — new bullet under 0.46 → New Features
- Did you get Claude approval on this PR?: pending `/claude review`

### Additional Information

Validated locally against DiffusionGemma v10 (`DiffusionGemmaForBlockDiffusion`, `model_type: diffusion_gemma`). Substring patterns are also chosen to match the older `DiffusionGemma4ModelForBlockDiffusion` / `diffusion_gemma4` spelling, so the same code path works on earlier and current model versions.

## New Features
- Added tied-weight PTQ and Hugging Face checkpoint export with export-time deduplication for encoder-decoder LLMs (including DiffusionGemma).
- Implemented canonical tied key ordering during export and automatic max-merging of shared `input_quantizer.amax` values.
- Extended model-type detection and export handling for DiffusionGemma.

## Bug Fixes
- Fixed quantized diffusion preview decoding when generation returns a `ModelOutput` wrapper.

## Documentation
- Added DiffusionGemma PTQ guidance and new YAML recipes to disable `self_conditioning` quantizers.

## Tests
- Added unit tests for tied export behavior, `amax` syncing, and fused-expert tied deduplication.


modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +718,177 @@ def _export_quantized_weight(
     if weight_scale is not None:
         sub_module.register_buffer(quantizer_attrs.weight_scale, weight_scale)
 
+    # Tied-weight dedup: if a previously-processed module shared the same
+    # source weight memory, alias the packed weight + scale buffers so the
+    # downstream data_ptr dedup in postprocess_state_dict can collapse them.
+    # input_scale is safe to alias because sync_tied_input_amax (earlier in
+    # this export) already max-merged the per-side amaxes.
+    _cache = _export_quantized_weight.__dict__.setdefault("_tied_weight_alias_cache", {})
+    _prior = _cache.get(_tied_source_data_ptr)
```
cjluo-nv: > _Bot comment._

This alias cache is function-static (`_export_quantized_weight.__dict__[...]`) and keyed by `data_ptr()`, but it is never cleared in production — only the tests pop it. Two failure modes:

1. **Cross-export contamination**: if `export_hf_checkpoint` is invoked more than once in the same process (batch export, two models), the freed `data_ptr` of a tensor from the first export can be recycled by an allocation in the second export and falsely hit a stale `_prior` module — aliasing this module's weight to a dead module from the prior run.
2. **Intra-export reuse**: `_tied_source_data_ptr` is captured from the float `weight` that gets replaced by `setattr` below; after replacement that storage may be freed and its address reused by a later unrelated module's weight, producing a false-positive match.

Suggest clearing this cache (and the `_export_fused_experts` one) at the start of `_export_transformers_checkpoint`, or threading an explicit per-export dict through these calls so identity is scoped to a single export.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +718,177 @@ def _export_quantized_weight(
...
+            patterns.append(re.compile(prefix + canonical_pat))
+            alias_token_set.update(_tokens(prefix + alias_pat))
+            canonical_token_set.update(_tokens(prefix + canonical_pat))
+
+    # Tokens unique to the canonical side become substring matchers.
+    side_substrings = sorted(canonical_token_set - alias_token_set)
+    return patterns, side_substrings
+
+
+def _reorder_canonical_first(state_dict: dict, model: nn.Module) -> dict:
```
cjluo-nv: > _Bot comment._

`side_substrings` are derived as canonical-only tokens and then matched as dot-separated path components across the whole state_dict. If a canonical-side token is a common identifier (e.g. ends up including something like `weight` or a generic submodule name shared by unrelated keys), `_has_side_substring` will pull unrelated keys into the `head` partition. Worth asserting/filtering the derived substrings to the tokens that are actually side-discriminating, or restricting matching to the tied subtree prefix.

cjluo-nv (commented):

PR adds tied-weight PTQ/HF-export support for encoder-decoder dLLMs (DiffusionGemma). Design review: the change correctly *reuses* the existing data_ptr-based first-wins dedup in `postprocess_state_dict` rather than introducing a competing dedup system — the new helpers (`sync_tied_input_amax`, the alias loops, `_reorder_canonical_first`) all feed that existing mechanism, and the PR body explains this. Design is settled. Licensing: new test files carry the project's standard Apache header (matches existing source files), no license-file or third-party-copy changes — no block. Tests: 10 CPU-only unit tests covering dedup/alias/reorder/amax-sync paths; e2e validated manually (not in CI), acceptable for the large-model path.

Main concern (correctness): both alias caches — `_export_quantized_weight._tied_weight_alias_cache` and `_export_fused_experts._tied_unpacked_cache` — are function-static, keyed by `data_ptr()`, and are **never cleared in production code** (only tests clear them). This risks (1) cross-export contamination when `export_hf_checkpoint` runs more than once in a process — a freed tensor's address from export #1 can be recycled by export #2 and falsely match a stale `_prior` module; and (2) intra-export false positives, since the captured source `data_ptr()` belongs to the float weight that gets freed after the `setattr` replacement and its address may be reused by a later unrelated allocation. Recommend resetting both caches at the start of `_export_transformers_checkpoint` (or passing an explicit per-export dict). Secondary: `_reorder_canonical_first`'s substring matcher can over-match when canonical tokens are common words. Flagging for the owner — the core mechanism is sound for the intended single-export tied case, but the persistent cache is a latent footgun.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +718,177 @@ def _export_quantized_weight(
     if weight_scale is not None:
         sub_module.register_buffer(quantizer_attrs.weight_scale, weight_scale)
 
+    # Tied-weight dedup: if a previously-processed module shared the same
+    # source weight memory, alias the packed weight + scale buffers so the
+    # downstream data_ptr dedup in postprocess_state_dict can collapse them.
+    # input_scale is safe to alias because sync_tied_input_amax (earlier in
+    # this export) already max-merged the per-side amaxes.
+    _cache = _export_quantized_weight.__dict__.setdefault("_tied_weight_alias_cache", {})
```
Edwardf0t1: **🔴 High — this cache is process-global and never scoped to an export run.**

`_export_quantized_weight._tied_weight_alias_cache` (and the sibling `_tied_unpacked_cache` in `moe_utils.py`) live on the *function object* and are only ever cleared by the tests (`_clear_export_quantized_weight_cache`). The fact that the tests need those helpers is the tell that the cache leaks across calls.

Two consequences:

1. **Cross-export contamination → silent wrong export.** The key is `weight.data_ptr()` (an `int`) and the value is a live module. If `export_hf_checkpoint()` runs more than once per process (library/notebook use, a loop over models, or the test session itself), run #1 entries survive into run #2. Source weights freed during run #1 can have their addresses recycled by run #2's allocations, so a run-#2 module can hit a stale `_prior` from run #1 and get its packed `weight`/`weight_scale`/`input_scale` re-pointed at a *different model's* buffers. The single-shot `hf_ptq.py` CLI is safe (one export per process); the public `export_hf_checkpoint` API is not.
2. **Memory leak** — cache values are modules, so every exported module is pinned for the process lifetime; the model can't be GC'd after export.

**Fix:** scope the cache to a single export invocation — build a dict in `_export_transformers_checkpoint` and thread it through `_process_quantized_modules` → `_export_quantized_weight` / `_export_fused_experts`. If you'd rather not touch signatures, clear both caches at the top of `_export_transformers_checkpoint`. Either way it shouldn't be process-global state keyed by an allocator-recyclable `int`.

Edwardf0t1 (commented):
Thorough, well-documented PR with genuinely good unit tests (tie / no-tie / one-side-unquantized / canonical-reorder all covered). The design — let the existing `postprocess_state_dict` data_ptr dedup do the collapsing and just re-alias packed buffers so it can see them — is clean and minimally invasive.

One real latent-correctness issue to fix before merge (the process-global alias caches), plus a few medium/minor items inline. See comments.

modelopt/torch/export/moe_utils.py
```diff
@@ -178,6 +189,46 @@ def _export_fused_experts(module: nn.Module, dtype: torch.dtype) -> None:
...
             delattr(module, attr)
 
+    # 5. Tied-experts dedup: if this module's source params have been seen
+    # before, alias the bit-identical per-expert buffers (weight,
+    # weight_scale, weight_scale_2, input_scale) to the previously-unpacked
+    # module. input_scale is safe to alias because sync_tied_input_amax
+    # runs earlier in _export_transformers_checkpoint and max-merges the
+    # shared input_quantizer amaxes across tied fused-experts modules, so
+    # both sides now derive bit-identical input_scale values.
+    _cache = _export_fused_experts.__dict__.setdefault("_tied_unpacked_cache", {})
```
Edwardf0t1: **🔴 High (companion to the `unified_export_hf.py` comment).** Same process-global cache problem here: `_export_fused_experts._tied_unpacked_cache` is keyed by a `(data_ptr, data_ptr)` tuple, persists across export calls, and pins every fused-experts module for the process lifetime. Fix it the same way — scope the cache to one `_export_transformers_checkpoint` invocation rather than to the function object.

modelopt/torch/export/moe_utils.py
```diff
@@ -178,6 +189,46 @@ def _export_fused_experts(module: nn.Module, dtype: torch.dtype) -> None:
...
+    _prior = _cache.get(_source_key)
+    if _prior is not None and _prior is not module:
```
Edwardf0t1: **🟡 Medium — second tied module is fully unpacked, then discarded.** On a cache hit the later tied module has already run the entire unpack/pack path above; this block then aliases all of it back to `_prior` and throws the freshly-computed buffers away. For a tied 26B MoE that transiently doubles peak experts memory and burns the compute the dedup is meant to save. If you can detect the `_source_key` hit *before* the unpack loop and short-circuit to pure aliasing, you get the storage win without the spike. At minimum, a comment noting the transient cost would help.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +718,177 @@ def _export_quantized_weight(
...
+    Lets the downstream first-wins data_ptr dedup keep canonical names.
+    Uses both regex patterns and substring matchers from
+    :func:`_collect_canonical_tied_patterns`. No-op when the model
+    declares no dict-style ``_tied_weights_keys``.
+    """
+    canonical_patterns, side_substrings = _collect_canonical_tied_patterns(model)
+    if not canonical_patterns and not side_substrings:
+        return state_dict
+
+    def _has_side_substring(key: str) -> bool:
```
Edwardf0t1: **🟡 Medium — substring matching is broader than "reorder the tied keys".** `side_substrings` is `canonical_tokens - alias_tokens` (e.g. `decoder`), so `_has_side_substring` sends *every* key with that path component to the head, not just the genuinely-tied ones. Harmless to correctness (the downstream dedup only drops true data_ptr duplicates), but on a model where `decoder` is canonical-only this reorders the entire decoder subtree. Fine for DiffusionGemma and the flag is opt-in/default-off — just flagging that the heuristic is coarser than the docstring implies and could surprise on an asymmetrically-named model.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +718,177 @@ def _export_quantized_weight(
...
+    _prior = _cache.get(_tied_source_data_ptr)
+    if _prior is not None and _prior is not sub_module:
+        if hasattr(_prior, weight_name):
+            setattr(sub_module, weight_name, getattr(_prior, weight_name))
+        for _attr in (
+            quantizer_attrs.weight_scale,
+            quantizer_attrs.weight_scale_2,
+            quantizer_attrs.input_scale,
+        ):
+            if _attr is None or not hasattr(_prior, _attr):
```
Edwardf0t1: **🟢 Minor — dead `None` check.** `quantizer_attrs.weight_scale_2` / `input_scale` are always non-`None` strings (see `quantizer_attr_names`), so `_attr is None` never fires; the `hasattr(_prior, _attr)` guard is doing all the work. Harmless, slightly misleading.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +718,177 @@ def _export_quantized_weight(
...
+    for k, v in state_dict.items():
+        if any(p.search(k) for p in canonical_patterns) or _has_side_substring(k):
+            head[k] = v
+        else:
+            tail[k] = v
+    head.update(tail)
+    return head
+
+
+def sync_tied_input_amax(model: nn.Module) -> int:
```
Edwardf0t1: **🟢 Minor — document the in-place model mutation.** `sync_tied_input_amax` overwrites `q.amax` on the live model. That's consistent with export already destroying the model (weights replaced with packed bytes), so it's fine — but a one-line note in the docstring that the model is not reusable after this runs would save a future reader. Also worth confirming the real DiffusionGemma recipe uses per-tensor input quantizers; the `numel != 1` guard silently *skips* the merge for non-scalar amax, and that path isn't exercised by the CPU unit tests.

modelopt/torch/utils/dataset_utils.py
```diff
@@ -1208,7 +1208,10 @@ def create_forward_loop(
 
 
 def model_type_is_enc_dec(model):
-    enc_dec_model_list = ["t5", "bart", "whisper"]
+    # Substring match against `model.__class__.__name__.lower()` — entries are
+    # the lowercased class-name form (no underscores). Calibration then uses
+    # `model.generate` to run the full denoising loop.
+    enc_dec_model_list = ["t5", "bart", "whisper", "diffusiongemma"]
```
Edwardf0t1, line 1244: **🟢 Minor — two `is_enc_dec`s now diverge intentionally.** This one (`model_type_is_enc_dec`, by class name) *includes* `diffusiongemma` to route calibration through `.generate()`, while `is_enc_dec` in `example_utils.py` (by model_type) deliberately *excludes* it so preview decode stays AR-style. The docstrings explain each, but same-named functions doing opposite things for the same model is a future-reader trap — a cross-reference comment would help.

modelopt/torch/export/moe_utils.py
```diff
@@ -42,13 +42,24 @@ def _export_fused_experts(module: nn.Module, dtype: torch.dtype) -> None:
            {E}.gate_proj.weight, {E}.gate_proj.weight_scale, ...
            {E}.up_proj.weight, {E}.up_proj.weight_scale, ...
            {E}.down_proj.weight, {E}.down_proj.weight_scale, ...
+
+    Tied-experts dedup: when multiple fused-expert modules share their 3-D
+    source params via HF ``_tied_weights_keys``, the unpacking creates fresh
+    per-expert tensors that break the tie. We cache the source ``data_ptr()``
+    at entry and on a later cache hit alias the per-expert ``weight`` /
+    ``weight_scale`` / ``weight_scale_2`` back to the prior module so
+    downstream dedup catches them. ``input_scale`` is left per-side.
```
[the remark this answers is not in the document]
juhi10071998: Addressed in 9f5e0e130777e02c5a59ba91a4c38c5fcaf54bc0 commit

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +718,177 @@ def _export_quantized_weight(
...
+        ):
+            if _attr is None or not hasattr(_prior, _attr):
+                continue
+            if _attr in sub_module._buffers:
+                del sub_module._buffers[_attr]
+            elif hasattr(sub_module, _attr):
+                delattr(sub_module, _attr)
+            sub_module.register_buffer(_attr, getattr(_prior, _attr))
+    else:
+        _cache[_tied_source_data_ptr] = sub_module
```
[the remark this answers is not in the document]
juhi10071998: Addressed in 9f5e0e130777e02c5a59ba91a4c38c5fcaf54bc0 commit

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1330,7 +1542,9 @@ def export_hf_checkpoint(
         return
 
     try:
-        post_state_dict, hf_quant_config = _export_transformers_checkpoint(model, dtype)
+        post_state_dict, hf_quant_config = _export_transformers_checkpoint(
+            model, dtype, canonical_tied_naming=canonical_tied_naming
+        )
```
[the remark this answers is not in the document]
juhi10071998: Addressed in 9f5e0e130777e02c5a59ba91a4c38c5fcaf54bc0 commit

modelopt/torch/export/unified_export_hf.py
juhi10071998: Thanks @cjluo-nv , addressed in 9f5e0e130777e02c5a59ba91a4c38c5fcaf54bc0 commit

modelopt/torch/export/unified_export_hf.py
juhi10071998: Thanks @Edwardf0t1 - addressed  in 9f5e0e130777e02c5a59ba91a4c38c5fcaf54bc0 commit

modelopt/torch/export/moe_utils.py
juhi10071998: Yes, addressed in 9f5e0e130777e02c5a59ba91a4c38c5fcaf54bc0 commit

modelopt/torch/export/moe_utils.py
juhi10071998: Makes sense, updated it, could you please review  in 9f5e0e130777e02c5a59ba91a4c38c5fcaf54bc0 commit

modelopt/torch/utils/dataset_utils.py
juhi10071998: done

cjluo-nv (commented):

Re-review of tied-weight PTQ/HF-export support for dLLM encoder-decoder models (DiffusionGemma). +880/-13, 13 files, 6 directories.

**Design review (settled).** The protocol applies (6 directories), but the design question is explicitly addressed: the PR reuses the existing `postprocess_state_dict` data_ptr first-wins dedup rather than introducing a competing dedup system — `sync_tied_input_amax`, the alias loops, and `_reorder_canonical_first` all feed that existing mechanism. Two maintainer reviewers (cjluo-nv, Edwardf0t1) independently confirmed this in prior comments. Proceeding to correctness.

**Prior critical comments — all resolved:**
- :speech_balloon: Process-global alias caches (flagged "High" by both cjluo-nv and Edwardf0t1; cross-export contamination + module leak). Author replied *"addressed in 9f5e0e1"* — verified: caches are now caller-owned (`_tied_cache`/`_moe_tied_cache` created fresh in `_process_quantized_modules` on every call) and threaded through `_export_quantized_weight`/`_export_fused_experts`. Function-static state eliminated.
- :speech_balloon: `**kwargs` dropped on `_export_transformers_checkpoint` call (CodeRabbit Major). Verified forwarded.
- :speech_balloon: MoE second-tied-module fully unpacked then discarded (Edwardf0t1 Medium). Verified: the MoE tied fast-path now short-circuits to pure aliasing *before* the unpack loop, avoiding the transient memory spike.
- Docstring `input_scale` contradiction, diverging `is_enc_dec` cross-reference, in-place mutation note — all addressed.

**New CodeRabbit comment (moe_utils.py:224, "breaks monkeypatched shims") is not valid** — the in-tree spy functions in `test_fused_experts.py` were updated to `def _spy_export(wrapper, dtype, **_kwargs)`, so the unconditional `_tied_cache=` forwarding does not break them.

**Why nudge rather than approve (needs export-owner sign-off):**
- E2E validated manually on DiffusionGemma v10 (not in CI); the large tied-MoE path and the `accelerator`-gathered distributed export path have no automated coverage. 10 CPU unit tests cover the dedup/alias/reorder/amax-sync logic well.
- :speech_balloon: `_reorder_canonical_first` substring breadth (cjluo-nv:786, Edwardf0t1:798) — author did not directly reply; acknowledged by reviewers as harmless-to-correctness and opt-in/default-off, and `_has_side_substring` now requires dot-separated path components. Still coarser than the docstring implies on asymmetrically-named models; worth a human confirming the DiffusionGemma naming is the only intended consumer.
- The `numel != 1` (non-scalar amax) branch in `sync_tied_input_amax` silently skips the merge and is not exercised by the CPU tests; confirm the real recipe uses per-tensor input quantizers.

Licensing clean: new recipe YAMLs carry the standard NVIDIA Apache header (matches `LICENSE_HEADER`), no license-file or third-party-code changes.

examples/llm_ptq/hf_ptq.py
```diff
@@ -1252,6 +1258,19 @@ def parse_args() -> argparse.Namespace:
         default=512,
     )
     parser.add_argument("--export_path", default="exported_model")
+    parser.add_argument(
+        "--canonical_tied_naming",
```
cjluo-nv: What does this do? Can we just detect difussion gemma and apply this automatically instead of introducing a new flag?

modelopt/torch/export/moe_utils.py
```diff
@@ -42,13 +47,75 @@ def _export_fused_experts(module: nn.Module, dtype: torch.dtype) -> None:
...
+    ``_export_transformers_checkpoint``) and scoped to one export
+    invocation; when ``None`` the corresponding alias step is skipped.
     """
     from modelopt.torch.export.unified_export_hf import _export_quantized_weight
     from modelopt.torch.quantization.plugins.huggingface import _get_fused_expert_intermediate_dim
 
     n = module.num_experts
     expert_dim = _get_fused_expert_intermediate_dim(module)
 
+    # Capture source tensor identities BEFORE unpacking (the source
```
cjluo-nv, line 119: move these into a helper method for readability

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,14 +725,197 @@ def _export_quantized_weight(
     if weight_scale is not None:
         sub_module.register_buffer(quantizer_attrs.weight_scale, weight_scale)
 
+    # Tied-weight dedup: if a previously-processed module shared the same
```
cjluo-nv, line 729: do we dedupe the quantized the weights or we also need to have them tied?

cjluo-nv (commented):

Re-review of tied-weight PTQ/HF-export support for dLLM encoder-decoder models (DiffusionGemma). +880/-13, 13 files, 6 dirs. Design-review protocol applied; design is settled (reuses the existing `postprocess_state_dict` data_ptr first-wins dedup — `sync_tied_input_amax`, the alias loops and `_reorder_canonical_first` all feed that one mechanism rather than introducing a competing dedup system; confirmed by maintainers cjluo-nv + Edwardf0t1). Licensing clean (recipe YAMLs carry the standard NVIDIA Apache header). No prompt-injection in the untrusted content.

**Prior critical comments — verified resolved in current diff:**
- :speech_balloon: Process-global alias caches (flagged High by cjluo-nv + Edwardf0t1). Author: *"addressed in 9f5e0e1"* — confirmed: `_tied_cache`/`_moe_tied_cache` are now created fresh in `_process_quantized_modules` on every call and threaded through `_export_quantized_weight`/`_export_fused_experts`. Function-static state eliminated.
- :speech_balloon: `**kwargs` dropped on `_export_transformers_checkpoint` call — verified forwarded (so `accelerator` is preserved).
- :speech_balloon: MoE second-tied-module fully unpacked then discarded — verified: the tied fast path now short-circuits to pure aliasing before the unpack loop.
- :speech_balloon: Docstring `input_scale` contradiction / `is_enc_dec` cross-reference — verified addressed.
- CodeRabbit "moe_utils.py:224 breaks monkeypatched shims" — not valid; the in-tree spy fns were updated to `(wrapper, dtype, **_kwargs)`.

**Why nudge rather than approve (needs export-owner sign-off):**
- **Operator's complexity question is unaddressed.** The operator asks whether this can be simplified — "can we dedupe before PTQ and postprocess?" The PR re-aliases *packed* buffers post-quantization so the existing dedup catches them; it does not discuss the alternative of deduping the shared source weight before/around PTQ (which could avoid `sync_tied_input_amax`, the per-expert alias loop, and `_export_quantized_weight`'s alias step entirely). Worth the author justifying in the PR body why post-pack aliasing is preferred over a pre-PTQ approach.
- **cjluo-nv's flag question (hf_ptq.py:1262) is still open:** "Can we just detect diffusion gemma and apply this automatically instead of introducing a new flag?" `--canonical_tied_naming` adds public surface; auto-detection from `_tied_weights_keys` would avoid it. Author has not replied.
- **Minor regression introduced by the cache fix:** in `_process_quantized_modules` the two cache-init statements now precede the docstring, so the triple-quoted block is no longer the function docstring (`__doc__` becomes `None`). Move the `_tied_cache = {}` / `_moe_tied_cache = {}` lines below the docstring. Private fn, low impact.
- **Coverage gaps:** E2E validated manually on DiffusionGemma v10 (not in CI). The tied-MoE path and the distributed `accelerator`-gathered export path have no automated coverage; the `numel != 1` (non-scalar amax) branch in `sync_tied_input_amax` silently skips and is untested — confirm the real recipe uses per-tensor input quantizers. 10 CPU unit tests cover the dedup/alias/reorder/scalar-amax logic well.
- :speech_balloon: `_reorder_canonical_first` substring breadth (cjluo-nv:786, Edwardf0t1:798) — `_has_side_substring` now requires dot-separated path components, but it's still coarser than the docstring on asymmetrically-named models; opt-in/default-off limits blast radius. Worth a human confirming DiffusionGemma is the only intended consumer.

examples/llm_ptq/hf_ptq.py
juhi10071998: Thansk  @cjluo-nv. This controls the reorder of  the state_dict so keys matching the canonical-side patterns (derived from _tied_weights_keys dict-form) appear before their aliases when iterated. 
Downstream postprocess_state_dict does a first-wins data_ptr dedup, so the reordering causes canonical-side keys to win that dedup → they end up as the  surviving names in the checkpoint, with the aliases dropped.

{in the state dict we want the decoder keys, if we don't do this than then decoder keys are dropped and the encoder keys remain in the state dict}

I will try to add the model check in the _reorder_canonical_first function itself- https://github.com/juhi10071998/Model-Optimizer/blob/d2acf0a0fc015969a8587f1e9b9074039162fc08/modelopt/torch/export/unified_export_hf.py#L1192

modelopt/torch/export/unified_export_hf.py
juhi10071998: if the module was originally weight tied, we add the quantizers to both of them irrespective (in this case the expert layers of encoder and decoder) and then we calibrate the quantizers. When exporting if we don't do the de-duplication, we will end up with duplicate entries in ckpt.

cjluo-nv (commented):

Re-review of tied-weight PTQ/HF-export support for dLLM encoder-decoder models (DiffusionGemma). +880/-13, 13 files, 6 dirs. Design-review protocol applied; design is settled — the PR reuses the existing `postprocess_state_dict` data_ptr first-wins dedup (`sync_tied_input_amax`, the alias loops, and `_reorder_canonical_first` all feed that one mechanism), the recipe YAMLs reuse the existing `$import` unit system rather than a new DSL, and the operator's "dedupe before PTQ?" question is answered in the operator message (per-module quantizers observe ~3-5x divergent encoder/decoder activation distributions, so they can't be collapsed pre-calibration). Confirmed by maintainers cjluo-nv + Edwardf0t1. Licensing clean (recipe/test headers match the canonical NVIDIA Apache `LICENSE_HEADER`; only the copyright year differs). No prompt injection in untrusted content (CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts).

Prior critical comments — verified resolved in the current diff:
- :speech_balloon: Process-global alias caches (flagged High by cjluo-nv + Edwardf0t1). Author: *"addressed in 9f5e0e1"* — confirmed: `_tied_cache`/`_moe_tied_cache` are now created fresh in `_process_quantized_modules` per call and threaded through `_export_quantized_weight`/`_export_fused_experts`. Function-static state eliminated.
- :speech_balloon: `**kwargs` dropped on `_export_transformers_checkpoint` — verified forwarded (so `accelerator` is preserved).
- :speech_balloon: MoE second-tied-module fully unpacked then discarded — verified: the tied fast path short-circuits to pure aliasing before the unpack loop.
- :speech_balloon: Docstring `input_scale` contradiction / `is_enc_dec` cross-reference — verified addressed.
- CodeRabbit "moe_utils.py:224 breaks monkeypatched shims" — not valid; the in-tree spy fns were updated to `(wrapper, dtype, **_kwargs)`.

Why nudge rather than approve (needs export-owner sign-off):
- cjluo-nv's auto-detect-vs-flag question (hf_ptq.py:1262) is still open: "can we just detect DiffusionGemma and apply this automatically instead of introducing a new flag?" `--canonical_tied_naming` adds public API surface; author said they'd try a model check but the flag remains. Worth a human deciding whether the flag is the right interface.
- Minor regression introduced by the cache fix: in `_process_quantized_modules` the two cache-init statements now precede the triple-quoted block, so it is no longer the function docstring (`__doc__` becomes `None`). Move the `_tied_cache = {}` / `_moe_tied_cache = {}` lines below the docstring. Private fn, low impact, but trivial to fix.
- Coverage gaps: E2E validated manually on DiffusionGemma v10 (not in CI). The tied-MoE export path, the distributed `accelerator`-gathered export path, and the `numel != 1` non-scalar-amax branch in `sync_tied_input_amax` (which silently skips the merge) have no automated coverage — confirm the real recipe uses per-tensor input quantizers. The 10 CPU unit tests cover the dense dedup/alias/reorder/scalar-amax + tied-MoE-vs-untied paths well.
- :speech_balloon: `_reorder_canonical_first` substring breadth (cjluo-nv:786, Edwardf0t1:798) — `_has_side_substring` now requires dot-separated path components, but it's still coarser than the docstring on asymmetrically-named models; opt-in/default-off limits blast radius. Worth a human confirming DiffusionGemma is the only intended consumer.

cjluo-nv (commented):

Re-review of tied-weight PTQ/HF-export support for dLLM encoder-decoder models (DiffusionGemma). +880/-13, 13 files, 6 dirs. Design-review protocol applied and the design is settled: the PR reuses the existing `postprocess_state_dict` data_ptr first-wins dedup (`sync_tied_input_amax`, the alias loops, and `_reorder_canonical_first` all feed that one mechanism), reuses the existing `$import` recipe-unit system, and both maintainers (cjluo-nv, Edwardf0t1) independently confirmed the approach. Licensing clean — new recipe/test files carry the standard NVIDIA Apache header. No actionable prompt injection in the untrusted content (CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts, not directives).

**Prior critical comments — verified resolved in current diff:**
- Process-global alias caches (flagged High by cjluo-nv + Edwardf0t1): confirmed fixed — `_tied_cache`/`_moe_tied_cache` are now created fresh in `_process_quantized_modules` per call and threaded through `_export_quantized_weight`/`_export_fused_experts`. Function-static state eliminated.
- `**kwargs` dropped on `_export_transformers_checkpoint`: verified forwarded (so `accelerator` is preserved).
- MoE second-tied-module fully unpacked then discarded: verified — the tied fast path short-circuits to pure aliasing before the unpack loop.
- moe_utils docstring `input_scale` contradiction: verified fixed (now states input_scale is aliased due to `sync_tied_input_amax`).
- CodeRabbit "moe_utils.py:224 breaks monkeypatched shims": not valid — the in-tree spy fns were updated to `(wrapper, dtype, **_kwargs)`.

**Why nudge rather than approve (export-owner should sign off):**
- :speech_balloon: Minor regression introduced by the cache fix: in `_process_quantized_modules` the two `_tied_cache`/`_moe_tied_cache` init statements now precede the triple-quoted block, so it is no longer the function docstring (`__doc__` becomes `None`). Move those two lines below the docstring. Private fn, low impact, trivial fix — still present in current file.
- :speech_balloon: cjluo-nv's auto-detect-vs-flag question (hf_ptq.py:1262) is still open: "can we detect DiffusionGemma and apply this automatically instead of a new flag?" `--canonical_tied_naming` adds public API surface; author said they'd try a model check but the flag remains. A human should decide the interface.
- Coverage gaps: E2E validated only manually on DiffusionGemma v10 (not in CI). The tied-MoE export path, the distributed `accelerator`-gathered export path, and the `numel != 1` non-scalar-amax branch in `sync_tied_input_amax` (which silently skips the merge) have no automated coverage — confirm the real recipe uses per-tensor input quantizers. The 10 CPU unit tests cover the dense dedup/alias/reorder/scalar-amax + tied-MoE-vs-untied paths well.
- :speech_balloon: `_reorder_canonical_first` substring breadth (cjluo-nv:786, Edwardf0t1:798): `_has_side_substring` now requires dot-separated path components, but it's still coarser than the docstring on asymmetrically-named models; opt-in/default-off limits the blast radius. Worth a human confirming DiffusionGemma is the only intended consumer.

The CodeRabbit nit about a local import in `_clear_fused_experts_caches` is moot — that helper no longer exists; tests now use caller-owned dicts directly.

cjluo-nv (commented):

Re-review of tied-weight PTQ/HF-export support for dLLM encoder-decoder models (DiffusionGemma). +866/-13, 13 files, 6 dirs. Design-review protocol applied and the design is settled: the PR reuses the existing `postprocess_state_dict` data_ptr first-wins dedup (`sync_tied_input_amax`, the per-expert/dense alias loops, and `_reorder_canonical_first` all feed that one mechanism rather than introducing a competing dedup system), and reuses the existing `$import` recipe-unit system rather than a new DSL. Two maintainers (cjluo-nv, Edwardf0t1) independently confirmed the approach. Licensing clean — new recipe/test files carry the standard NVIDIA Apache header. No actionable prompt injection in the untrusted content (the CodeRabbit "Prompt for AI Agents" blocks are tooling artifacts, not directives).

**Prior critical comments — verified resolved in current diff:**
- :speech_balloon: Process-global alias caches (flagged High by cjluo-nv + Edwardf0t1, author: *"addressed in 9f5e0e1"*) — confirmed: `_tied_cache`/`_moe_tied_cache` are now created fresh in `_process_quantized_modules` per call and threaded through `_export_quantized_weight`/`_export_fused_experts`. Function-static state eliminated.
- :speech_balloon: `**kwargs` dropped on `_export_transformers_checkpoint` call — verified forwarded (so `accelerator` is preserved on the distributed path).
- :speech_balloon: MoE second-tied-module fully unpacked then discarded — verified: the tied fast path short-circuits to pure aliasing before the unpack loop.
- :speech_balloon: Docstring `input_scale` contradiction in moe_utils.py — verified fixed (now states input_scale is aliased due to `sync_tied_input_amax`).
- The earlier "cache-init now precedes docstring → `__doc__` becomes None" regression is NOT present in the current file — the docstring is correctly the first statement and the cache-init lines follow it.
- CodeRabbit "moe_utils.py:224 breaks monkeypatched shims" — not valid; the in-tree spy fns were updated to `(wrapper, dtype, **_kwargs)`.

**Why nudge rather than approve (export-owner should sign off):**
- cjluo-nv's auto-detect-vs-flag question (hf_ptq.py:1262) is still open: "can we detect DiffusionGemma and apply this automatically instead of a new flag?" `--canonical_tied_naming` adds public API surface; the author said they'd try a model check inside `_reorder_canonical_first`, but the flag remains. A human should decide whether the flag is the right interface.
- Coverage gaps: E2E validated only manually on DiffusionGemma v10 (not in CI). The tied-MoE export path, the distributed `accelerator`-gathered export path, and the `numel != 1` non-scalar-amax branch in `sync_tied_input_amax` (which silently skips the merge) have no automated coverage — confirm the real recipe uses per-tensor input quantizers. The 10 CPU unit tests cover the dense dedup/alias/reorder/scalar-amax + tied-MoE-vs-untied paths well.
- :speech_balloon: `_reorder_canonical_first` substring breadth (cjluo-nv:786, Edwardf0t1:798) — `_has_side_substring` now requires dot-separated path components, but it's still coarser than the docstring implies on asymmetrically-named models; opt-in/default-off limits the blast radius. Worth a human confirming DiffusionGemma is the only intended consumer.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +725,192 @@ def _export_quantized_weight(
...
+    for k, v in state_dict.items():
+        if any(p.search(k) for p in canonical_patterns) or _has_side_substring(k):
+            head[k] = v
+        else:
+            tail[k] = v
+    head.update(tail)
+    return head
+
+
+def sync_tied_input_amax(model: nn.Module) -> int:
```
Edwardf0t1: Let's move this function to `quant_utils.py`

modelopt/torch/export/unified_export_hf.py
```diff
@@ -703,9 +725,192 @@ def _export_quantized_weight(
...
+            patterns.append(re.compile(prefix + canonical_pat))
+            alias_token_set.update(_tokens(prefix + alias_pat))
+            canonical_token_set.update(_tokens(prefix + canonical_pat))
+
+    # Tokens unique to the canonical side become substring matchers.
+    side_substrings = sorted(canonical_token_set - alias_token_set)
+    return patterns, side_substrings
+
+
+def _reorder_canonical_first(state_dict: dict, model: nn.Module) -> dict:
```
Edwardf0t1: I would move this function and `_collect_canonical_tied_patterns` to `model_utils.py`

Edwardf0t1 (approved):
Great work @juhi10071998 👍  Left a few more comments.

modelopt/torch/export/unified_export_hf.py
juhi10071998: done, thanks @Edwardf0t1 for the review

modelopt/torch/export/unified_export_hf.py
juhi10071998: makes sense, updated, thansk @Edwardf0t1

juhi10071998:
> Great work @juhi10071998 👍 Left a few more comments.

Thanks @Edwardf0t1 , I've addressed the review comments, please let me know if anything is missing.

modelopt_recipes/huggingface/diffusion_gemma/ptq/nvfp4_experts_only.yaml
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
juhi10071998, line 1: Hi @shengliangxu , @h-guo18 , @sychen52  could you please review the recipe based additions in the PR for DiffusionGemma for the recipe codeowner approval. Thanks so much!

modelopt/torch/export/moe_utils.py
juhi10071998: done, thanks @cjluo-nv

modelopt_recipes/huggingface/diffusion_gemma/ptq/README.md
```diff
@@ -0,0 +1,14 @@
+# DiffusionGemma PTQ recipes
```
shengliangxu, line 1: the model_type is diffusion_gemma, please add an underscore to it.

modelopt_recipes/huggingface/diffusion_gemma/ptq/disabled_quantizers.yaml
```diff
@@ -0,0 +1,30 @@
...
+# would otherwise crash with "AttributeError: 'TensorQuantizer' object has no
+# attribute '_amax'"). Recipes that import this should NOT also import
+# ``default_disabled_quantizers``.
+
+# modelopt-schema: modelopt.torch.quantization.config.QuantizerCfgListConfig
+imports:
+  default_disabled_quantizers: configs/ptq/units/default_disabled_quantizers
+---
+  - $import: default_disabled_quantizers
+  - quantizer_name: '*self_conditioning*'
```
shengliangxu, line 29: This is nice to making a model_type specific disabled list.

shengliangxu (approved):
LGTM for the recipes stuff

modelopt_recipes/huggingface/diffusion_gemma/ptq/README.md
juhi10071998: thanks Shengliang, just upated

juhi10071998: wait, did you refer to the file name or the comment above? I think I got it wrong

PR #1671, opened 2026-06-10
[OMNIML-4998] Per-expert weight quantization on TEGroupedLinear

hychiang-git:
### What does this PR do?

Type of change: New feature

Per-expert (axis=0) weight quantization on `_QuantTEGroupedLinear`, plus the cluster infrastructure to verify it. Consolidates three Jira tickets onto one branch:

- **OMNIML-4998** — Relax `_QuantMegatronTEGroupedLinear._process_quantizer_amax`'s per-tensor restriction. Per-expert `_amax` of shape `[num_gemms, 1, 1]` is calibrated via `iter_weights_for_calibration`'s stacked weight, broadcast through the existing axis-N machinery in `TensorQuantizer.forward`, and round-tripped via `sharded_state_dict` save/restore. Per-tensor (`axis=None`) path remains bit-for-bit equivalent.
- **OMNIML-5029** — `megatron_install_path` field on `SlurmConfig` / `aws_cmh_slurm_factory`. Bind-mounts the packaged Megatron-LM so `make_tp_sharded_tensor_for_checkpoint(..., allow_shape_mismatch=True)` resolves against `nemo:25.11.nemotron_3_nano` (previously deadlocked the dist-checkpoint metadata exchange for 10 minutes).
- **OMNIML-5030** — Test fix: enable `sequence_parallel=True` in the MoE+TP `test_moe_sharded_state_dict` config so the TEGrouped path doesn't trip the post-OMNIML-5029 assertion.

### Usage

```python
import modelopt.torch.quantization as mtq

config = {
    "algorithm": "max",
    "quant_cfg": [
        {"quantizer_name": "*", "enable": False},
        {
            "quantizer_name": "*experts.linear_fc*.weight_quantizer",
            "cfg": {"num_bits": 8, "axis": 0, "enable": True},
        },
    ],
}
model = mtq.quantize(model, config, forward)
```

### Testing

`tests/gpu_megatron/torch/quantization/plugins/test_megatron.py::test_te_grouped_per_expert_sharded_state_dict` (new) — per-expert axis=0 calibration + dist-checkpoint save/restore round-trip on `tp=2 ep=2 etp=1 num_moe_experts=4`. **PASSED 4/4 ranks on aws-cmh** (SLURM 511920, experiment `cicd_1781150929`, container `nvcr.io/nvidia/nemo:25.11.nemotron_3_nano`).

Existing tests continue to pass:
[...]

PR #1717, opened 2026-06-14
[OMNIML-5072] Triton fakequant adapter for N-quantizer per-expert path

hychiang-git:
### What does this PR do?

Type of change: new feature

Implements [OMNIML-5072](https://jirasw.nvidia.com/browse/OMNIML-5072), built on top of [PR #1550](https://github.com/NVIDIA/Model-Optimizer/pull/1550) (the N-quantizer per-expert foundation, still WIP). Three additions:

1. **Triton fakequant dispatch** for the N-quantizer per-expert path. Wires the single-launch `grouped_axis0_fakequant` kernel from [PR #1671](https://github.com/NVIDIA/Model-Optimizer/pull/1671) (the One-Vec-quanitzer path) into `_QuantTEGroupedLinear.te_grouped_quantized_linear_fn` when `_per_expert_weight_quantizer == True`. Soft-gated behind `_triton_kernels.IS_AVAILABLE` and `q._if_calib`; falls back to the existing `cuda_ext` per-quantizer loop when the gate is False.
2. **Cached `_gather_per_expert_amax`** helper eliminates per-forward O(N) Python overhead (walks the N submodules from PR #1550 and stacks the N scalar `_amax` buffers; lazily cached, invalidated from `modelopt_post_restore`).
3. **`sharded_state_dict` save + EP-aware load** on `_QuantMegatronTEGroupedLinear`'s N-quantizer case — gather-once-cache pattern adapted to N scalar `_amax` buffers across the EP group, so the dist-ckpt round-trip that PR #1671 ships for One-Vec-quanitzer also works on the N-quantizer foundation from PR #1550.

This PR is stacked on top of PR #1550 (`jennifchen/te_per_expert`, still WIP). The diff against `main` includes PR #1550's commits underneath; the OMNIML-5072-specific work is the top 6 commits (`fd77b53d8..51b4c9226`). Once PR #1550 lands, rebase onto `main` to shrink the review surface.

### Usage

```python
import os

# Enable the N-quantizer per-expert path on TEGroupedMLP. With this PR loaded,
# the Triton kernel dispatch activates automatically when triton is available.
os.environ["MODELOPT_TEGROUPED_PER_EXPERT_QUANTIZER"] = "1"

# ... build a Megatron model with TEGroupedMLP ...
mtq.quantize(model, mtq.FP8_DEFAULT_CFG, forward_fn)
# Each TEGroupedLinear now has N weight_quantizer_i submodules;
# forward uses Triton `grouped_axis0_fakequant` when available, otherwise
# falls back to cuda_ext per-quantizer.
```

### Testing

GPU-validated on aws-cmh (B300, nemo:26.02 / nemo:25.11 containers):

- **Parity test** (4 pytest cases, 6.14s): N-quantizer-Triton vs N-quantizer-cuda_ext on Ultra production shape (N=32, [5120, 8192] bf16) — fwd within 1-ULP floor, bwd bit-exact under `pass_through_bwd=True`. Test at `tests/gpu/torch/quantization/plugins/test_te_grouped_triton_parity.py`.
- **Microbench** (4 cells: Nano / Super / Ultra at EP=4 and EP=8): N-quantizer-Triton ≈ One-Vec-quanitzer + Triton (PR #1671) within ~10%; the N-quantizer vs One-Vec-quanitzer topology has no measurable perf effect once both share the Triton kernel. Full matrix on [OMNIML-5064](https://jirasw.nvidia.com/browse/OMNIML-5064).
- **Dist-ckpt round-trip** (2 pytest cases, 92.99s) at TP=2/EP=2 for both `FP8_DEFAULT_CFG` and `NVFP4_DEFAULT_CFG`. Test at `tests/gpu_megatron/torch/quantization/plugins/test_megatron.py::test_te_grouped_n_modules_sharded_state_dict`.

- Is this change backward compatible?: ✅ — Triton path is soft-gated; when `_triton_kernels.IS_AVAILABLE` is False or during calibration (`q._if_calib`), the original `cuda_ext` per-quantizer loop runs unchanged.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ — Reused the `grouped_axis0_fakequant` kernel module from PR #1671 (`51b4c9226` reuses it via the new `_GroupedAxis0FakeQuantFn` autograd adapter). No new PIP dependencies.
- Did you write any new necessary tests?: ✅ — Parity test + N-quantizer `sharded_state_dict` TP=2/EP=2 test.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ❌ — Stacked on top of WIP PR #1550; changelog entry deferred until both are merge-ready (will land with the rebase).
- Did you get Claude approval on this PR?: ❌ — Pending; will run `/claude review` after rebase onto `main`.

### Additional Information

Related work:
- [OMNIML-5072](https://jirasw.nvidia.com/browse/OMNIML-5072) — this ticket.
- [OMNIML-5064](https://jirasw.nvidia.com/browse/OMNIML-5064) — N-quantizer vs One-Vec-quanitzer comparison study; full microbench matrix here.
- [PR #1550](https://github.com/NVIDIA/Model-Optimizer/pull/1550) — N-quantizer foundation (WIP); this PR is stacked on top.
- [PR #1671](https://github.com/NVIDIA/Model-Optimizer/pull/1671) — One-Vec-quanitzer + Triton kernel (the kernel this PR reuses).


modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -773,13 +773,28 @@ def _real_quantize(self, inputs):
...
+            # layout is *deploy-time* optimal for nvfp4_gemm but breaks the
+            # Megatron sharded-ckpt save/load round-trip: the saved buffer's
+            # shape ([1741824]) no longer matches the canonical
+            # [out, in/block_size] shape ([10304, 168]) that the load-side
+            # model expects, so dist_checkpointing raises
+            # `Global shape mismatch ... for key ...weight_quantizer._scale`.
+            # Forcing the non-trtllm path produces canonical-layout per-block
+            # scales that survive save/load; HF export / cutlass kernels can
+            # re-interleave at deploy time via
+            # `cutlass_fp4_scale_to_modelopt_fp4_scale` and its inverse.
```
realAsma: @realAsma  help me understand these comments.

realAsma:
@realAsma How does this PR support N-quantizers for FP4, FP8 etc? In my understanding this PR only supports INTx quantization. Is that correct?

modelopt/torch/quantization/nn/modules/tensor_quantizer.py

My read: this comment is explaining a scale-layout issue, not the N-quantizer kernel itself.

`NVFP4QTensor.quantize(..., try_tensorrt=True)` can call `torch.ops.trtllm.fp4_quantize`. That path returns the packed weight plus `_scale` in the CUTLASS / `nvfp4_gemm` runtime layout: flattened/interleaved and padded for tile alignment. That is good for immediate GEMM use, but it is a bad persistent ModelOpt buffer shape. Megatron dist-checkpoint later expects the quantizer `_scale` buffer to match the canonical per-block shape derived from the module weight, roughly `[out_features, in_features / block_size]`. If the saved buffer is the CUTLASS-padded 1-D layout, restore sees a shape mismatch.

So `try_tensorrt=False` forces the pure ModelOpt path here, which stores canonical per-block FP8 scales in `_scale`; deployment/export can do runtime-specific layout preparation later. The part I would still push on is scope/test coverage: this changes NVFP4 real-quantization globally, not just this TEGrouped/Megatron checkpoint path, so it should have a focused regression showing the canonical `_scale` shape round-trips and that the relevant export path still emits what the runtime expects. Also, the “and its inverse” wording is a bit loose; I see the CUTLASS-to-ModelOpt conversion in `NVFP4QTensor.dequantize`, while export-side layout conversion appears to be handled elsewhere depending on the export path.

realAsma:
> _:robot: Bot comment._

Replying to @realAsma's question above: yes, for the new Triton N-quantizer *kernel* path, that is my read. `grouped_axis0_fakequant` is an integer fake-quant kernel: it takes `num_bits: int`, computes integer `qmax/qmin`, loads one scalar amax per expert, and applies `round(clamp(x / scale)) * scale`.

That does not implement the normal FP8 or NVFP4 `TensorQuantizer` semantics:

- FP8 uses `num_bits=(4, 3)` / E4M3 and normally goes through the FP8 path, not integer clamp/round.
- NVFP4 uses `num_bits=(2, 1)` plus dynamic block sizes and FP8 block scales, not one scalar amax per expert.
- INT8 default ModelOpt weights are also per-channel (`axis=0`), so even INT8 is only supported by this kernel for the scalar/per-tensor subset unless the kernel grows per-channel support.

So either the Triton fast path should be gated to the exact supported subset, e.g. integer, per-tensor, no block sizes, scalar `_amax`, or the PR needs real FP8/NVFP4/per-channel implementations plus tests that exercise the actual `_QuantTEGroupedLinear` forward path for those configs. The current parity test only calls the kernel directly with scalar int8-style inputs, and the FP8/NVFP4 Megatron test checks checkpoint round-trip rather than the post-calibration Triton forward branch.

modelopt/torch/kernels/quantization/gemm/__init__.py
```diff
@@ -38,4 +38,8 @@
         if torch.cuda.get_device_capability() >= (8, 9):
             from .fp4_kernel_hopper import *
 
+        # OMNIML-5072 — per-expert axis-0 fake-quant via tensor-of-pointers.
```
jenchen13: remove this jira ticket from comment

modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -28,12 +29,18 @@
...
 
-from ..nn import GroupedQuantizer, QuantModuleRegistry, SequentialQuantizer
+import modelopt.torch.kernels.quantization.gemm as _triton_kernels
+from ..nn import QuantModuleRegistry, SequentialQuantizer
 from .custom import _ParallelLinear
 
 _TE_VERSION = Version(te.__version__)
 
 
+def _per_expert_weight_quantizer_enabled() -> bool:
```
jenchen13: instead of using a env var to toggle per expert quantizer, let's just change TEGroupedLinear to per expert by default. alternatively we could allow a way to toggle it like `sync_expert_amax` does to SequentialMLP but that would be more complex to implement

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
hychiang-git: Addressed the concrete coverage and wording concerns in 7cb353923: added a focused GPU regression asserting canonical NVFP4 `_scale` shape `(out_features, in_features / block_size)`, removed the loose inverse-conversion wording, and validated TP1/EP4 distributed-checkpoint round trips for both dynamic and static NVFP4. The runtime-specific layout remains the exporter/backend responsibility. Leaving the thread open for reviewer judgment on the broader scope question.

hychiang-git:
We will use `torch.compile` in https://github.com/NVIDIA/Model-Optimizer/pull/1550

PR #1725, opened 2026-06-15
[DRAFT] Add heterogeneous AnyModel distillation example for Puzzletron.

### LINKED ISSUE #1679: Add MBridge distillation for puzzletron
### Detailed description of the requested feature
After the puzzletron prunes the model, we want a repair phase that will run distillation from the teacher (original model, but possibly any other model) to the pruned model

### Timeline
Work in progress. Requires "general" support of heterogeneous models in Mcore. The idea is to patch the mbridge-mcore model creation using the block_configs from anymodel.
This also requires some updates (PRs) to MBridge code as well

chochowski:
Ports examples/puzzletron/distillation from the puzzletron_distillation branch with SPDX headers, nemo2 registry fix, and puzzletron README link.

### What does this PR do?

This PR adds support for mbridge distillation of puzzled heterogeneous models

### Usage

In general distillation requires more compute than normal training, thus ideally you should run this on a multinode-setting (unless very constrained scenarion)
```python
torchrun --nproc-per-node=8  examples/puzzletron/distillation/distill.py \
--student llama \
--teacher llama \
--student-checkpoint /puzzletron/workspaces/Llama-3.1-8B-Instruct/mip/puzzle_solutions/target_memory_78000MiB-num_params_7G/solutions--checkpoints/solution_0/ \
--teacher-checkpoint /puzzletron/workspaces/Llama-3.1-8B-Instruct/ckpts/teacher/   \
--config-file examples/puzzletron/distillation/kd-container-llama.yaml \
--tensor-model-parallel-size 1 \
--pipeline-model-parallel-size 8 \
--expert-model-parallel-size 1 \
--expert-tensor-parallel-size 1 \
train.train_iters=1000 \
checkpoint.save=/puzzletron/workspaces/Llama-3.1-8B-Instruct/kd/puzzle_solutions/target_memory_78000MiB-num_params_7G-intermediate-fix/ \
logger.wandb_exp_name=Llama-3.1-8B-Instruct-target_memory_78000MiB-num_params_7G-intermediate-fix-verify
```

### Testing
 TBD

## Summary by CodeRabbit

* **New Features**
  * Added a heterogeneous knowledge distillation workflow for AnyModel/Puzzletron pairs via Megatron Bridge.
  * Introduced new distillation and export command-line scripts, including HuggingFace ↔ Megatron Bridge GPT-OSS weight conversion support.
  * Added supporting distillation utilities and presets for streamlined heterogeneous scenarios.
* **Documentation**
  * Expanded Puzzletron distillation guides with detailed heterogeneous and GPT-OSS-specific instructions, plus end-to-end distill/export examples.
* **Bug Fixes**
  * Improved hybrid MoE distillation stability by preventing aux-loss buffer sizing mismatches on pipeline stages.
* **Chores**
  * Updated linting rules for the Puzzletron distillation example files.


examples/puzzletron/distillation/__init__.py
kevalmorabia97, line 1: we have been using https://github.com/NVIDIA/Model-Optimizer/blob/main/examples/megatron_bridge/distill.py for puzzletron anymodel distillation already. Why do we now need all this extra patching logic?

chochowski: MCore heterogeneous support is not general e.g. it does not support mamba stack. With the patching we can get independent from mcore support, and rely only on whether mbridge supports a model or not - the whole heterogeneous logic is done with patching

examples/puzzletron/distillation/kd-container-nemotron3.yaml
```diff
@@ -0,0 +1,164 @@
...
+# Merge order: _pretrain_common() skeleton → this file → Hydra-style CLI overrides.
+#
+# Paths below are placeholders — override for your environment.
+
+# ---------------------------------------------------------------------------
+# Model (DistillationProvider + TransformerConfig fields)
+# ---------------------------------------------------------------------------
+model:
+  # Heterogeneous AnyModel / per-layer MoE: distinct checkpoint keys per global layer.
+  hetereogenous_dist_checkpoint: true
```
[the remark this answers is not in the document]
chochowski: no the typo is there in the upstream code, so we just follow

kevalmorabia97:
## Recommendation: extend the puzzletron mbridge plugin rather than fork a parallel example

Thanks for this — the hybrid/MoE enablement is clearly real and the diagnostics (untouched-param NaN detection, the MoE aux-loss deadlock analysis) are valuable. My main concern is **where** the ~2,500 lines land.

Key observation: `tests/examples/megatron_bridge/test_distill.py::test_distill_puzzletron_anymodel` already passes **today on `main`** (added in #1600), distilling a *heterogeneous* Qwen3 AnyModel student through the generic `examples/megatron_bridge/distill.py`. That works because `modelopt/torch/puzzletron/plugins/mbridge/` (`base.py` `HeterogeneousBridgeMixin`/`GenericHeterogeneousProvider`, `llama.py`, `qwen3.py`) auto-registers on the `import modelopt.torch.puzzletron.plugins.mbridge` already in the generic script, and builds on Megatron-Core's native `get_gpt_heterogeneous_layer_spec` / `HeterogeneousTransformerConfig`.

So the genuinely-missing capability here is **Mamba-hybrid (Nemotron-H) and variable-expert / no-op MoE (GPT-OSS) layer specs**, not "heterogeneous distillation" in general. Proposal:

1. **Add the hybrid/MoE coverage to the plugin**, not `examples/`. Extend `base.py` (or add `nemotron_h.py` / `gpt_oss.py` bridges) so a Nemotron-H / GPT-OSS AnyModel checkpoint flows through the same heterogeneous-spec path the GPT bridges use. The plugin auto-registers on import, so the generic `distill.py` keeps working with **zero** new puzzletron-specific code.
2. **`layer_patchers.py` + `provider_patch.py` should not be needed.** They re-implement per-layer construction by monkey-patching `build_module` in three namespaces + `MambaLayer.__init__`, which bypasses the native heterogeneous-spec machinery the plugin already uses. If Megatron-Core lacks a *hybrid* heterogeneous layer spec (the GPT one exists), that's the real upstream gap — see below.
3. **`block_config_utils.py`** (BlockConfig → MCore overrides) is core AnyModel logic and belongs under `modelopt/torch/puzzletron/`, reusable by the plugin.
4. **Drop the `gpt_oss_bridge.py` `cp`-over-the-installed-package step** — that's not a supportable distribution mechanism (see upstream note).

The `DistillHooks` refactor of the generic `distill.py` is backward-compatible and fine to keep, but note that most of the `PuzzletronHooks` overrides are "apply a monkey-patch / read block_configs" that the plugin could self-register at import time. The one hook with an independent rationale is `build_config` (YAML/Hydra config style), which is an orthogonal ergonomics preference, separable from heterogeneity.

### Changes that should be upstreamed to **Megatron-Bridge**
A **nemo:26.06.01 patch release is expected in ~2 weeks** — the plan is to get these upstream fixes included there, which would let us delete the corresponding example-side patches:

- **GPT-OSS bridge heterogeneity** (`gpt_oss_bridge.py`): fold the Puzzletron heterogeneous-GPT-OSS handling into `megatron/bridge/models/gpt_oss/gpt_oss_bridge.py` upstream. This removes the `cp` overlay entirely.
- **`_megatron_local_name_to_global` EP fix** (`model_bridge_patch.py` Patch A): upstream computes `num_experts_per_rank = config.num_moe_experts // ep_size`, which is wrong when per-layer expert counts differ; it should read `layer_module.mlp.num_local_experts`. This is a Megatron-Bridge correctness bug for heterogeneous EP.
- **`training_log` MoE aux-loss sizing** (root cause behind `hooks.py::_install_hybrid_moe_aux_loss_size_fix`): Bridge's `training_log` passes `num_layers = hybrid_layer_pattern.count("E")` to `track_moe_metrics`, while MoE routers write buffers of size `config.num_layers` → cross-PP `all_reduce` size mismatch / NCCL deadlock when a PP stage has zero surviving MoE layers. Fix belongs in Bridge's `training_log`.
- **Untouched-param diagnostic** (`model_bridge_patch.py` Patch B, `STRICT_HF_TO_MEGATRON_LOAD`): genuinely useful as an opt-in guard in `load_weights_hf_to_megatron` — worth proposing upstream.

### Changes that belong in **Megatron-Core** (not Bridge)
- **`_update_router_expert_bias` / `reset_model_temporary_tensors`** (`router_expert_bias_patch.py`): these are `megatron.core` `finalize_model_grads` helpers. `torch.stack(tokens_per_expert_list)` assumes equal per-layer expert counts, and the reset assumes every expert-bias module has an allocated token counter. Both should be fixed in Megatron-Core to support heterogeneous per-layer `num_local_experts`.
- **Hybrid heterogeneous layer spec**: if Megatron-Core has no Mamba/hybrid analogue of `get_gpt_heterogeneous_layer_spec`, adding one is the cleanest upstream fix and would let the plugin drop `layer_patchers.py` for hybrids too.

### Test coverage
`test_distill_puzzletron_anymodel` currently only exercises Qwen3 (dense-GPT heterogeneous). Please **add a hybrid case** so the new Mamba+MoE path is actually covered — `create_tiny_nemotron_h_dir` (`hybrid_override_pattern="ME*E"`, i.e. Mamba + MoE + attention) and the `nemotron_h_v2` converter already exist, so it can mirror the existing test with a Nemotron-H student.

### Bottom line
If the plugin extension above lands, the generic `examples/megatron_bridge/distill.py` should remain the **single entrypoint** for homogeneous + heterogeneous + QAD, with no separate script needed. Only if a hybrid heterogeneous spec genuinely can't be expressed through the plugin in this iteration should the standalone `examples/puzzletron/distillation/distill.py` stay — and even then it should import the generic `main()` cleanly (packaged shared module) rather than via `sys.path.insert`, and the library-grade patches above should still move out of `examples/`.

Separius:
looking at the review, I disagree, mamba-hybrid is not the only missing part, the heterogenous model is outdated in mcore (it was just designed for llama-puzzle), the right way to support things is to patch any layer, making them heterogenous, exactly like Anymodel. I agree that ideally mcore should support per block/layer configs but that is a huge change and will take forever to land on the main, we need a working distillation path for puzzle outputs, the current one landed on the main is incomplete and useless for most of the new models (MoE, mamba, linear attention, mla, sliding attention, sparse attention), on paper at least this approach allows us to distill any model that has a bridge

kevalmorabia97:
@Separius Fair pushback, and I'll concede the main point: the generic "patch any layer at `build_module`" approach **is** more future-proof than extending MCore's heterogeneous spec per type. You're right that a Mamba spec only buys Mamba — you'd be back here for MLA / linear / sliding / sparse attention, and MCore's `get_gpt_heterogeneous_layer_spec` has a fixed slot vocabulary that doesn't scale to those. And the "land it now vs wait for a big MCore change" tradeoff is real. So I'm dropping the "add a hybrid hetero spec upstream" as any kind of blocker — agreed it's a forever-project.

Where I'd still nudge, and I think it actually *supports* your goal of a working path for all puzzle outputs: the disagreement isn't really technique vs. technique, it's **where the patcher lives**. If `build_module` interception is the general, long-lived way to distill any bridge-able model, that's an argument for it being **first-class library code** in `modelopt/torch/puzzletron/` (right next to the AnyModel machinery it mirrors — `deci_x_patcher`, `ModelDescriptor`), picked up by the `import modelopt.torch.puzzletron.plugins.mbridge` the generic `examples/megatron_bridge/distill.py` already runs. That keeps it a **single distillation path** (no forked `distill.py` / `sys.path.insert` / `cp`-over-the-installed-package) and, more importantly, lets it get **validated per model type with tests** — which is the thing that turns "on paper this works for any model" into "this works." The packaging is my only real ask; the patcher itself can stay as-is.

Two things I'd decouple from all of the above and land regardless, since they're correctness bugs that bite *any* approach and are small:
- router expert-bias `torch.stack` over heterogeneous expert counts (~30 lines),
- MoE aux-loss tracker sizing in `training_log` (~10 lines),
- EP local→global expert numbering (~20 lines).

Those are good candidates for the nemo:26.06.01 patch window independent of the bigger discussion. The GPT-OSS bridge change is the one I'd still like to see upstreamed eventually (to kill the file overlay), but no rush.

tl;dr — you win the technique argument; I'd just move the patcher into the puzzletron plugin so it's one entrypoint + testable, and land the three small bugfixes upstream.

kevalmorabia97:
## Upstream fix plan + resulting code reduction

Following up with the concrete upstream plan (Megatron-LM + Megatron-Bridge) and exactly how much it shrinks this PR. Line ranges below are against current `main` of each repo.

### Part A — Megatron-LM (MCore)

**MC‑1 · Heterogeneous router expert-bias update** — `megatron/core/distributed/finalize_model_grads.py::_update_router_expert_bias` (lines 358–369). Upstream already has the empty-list and KD teacher-eval guards; the only gap is `torch.stack` requiring equal per-layer expert counts.

```python
# BEFORE (358–369)
stacked_tokens_per_expert = torch.stack(tokens_per_expert_list, dim=0)
stacked_expert_bias = torch.stack(expert_bias_list, dim=0)
stacked_updated_expert_bias = get_updated_expert_bias(
    stacked_tokens_per_expert, stacked_expert_bias,
    config.moe_router_bias_update_rate, tp_dp_cp_group=tp_dp_cp_group,
)
for expert_bias, updated_expert_bias in zip(expert_bias_list, stacked_updated_expert_bias):
    expert_bias.copy_(updated_expert_bias)

# AFTER — group by expert-vector length, then stack within each group
from collections import defaultdict
groups: dict[int, list] = defaultdict(list)
for tokens, bias in zip(tokens_per_expert_list, expert_bias_list):
    groups[tokens.numel()].append((tokens, bias))
for _, pairs in sorted(groups.items()):
    updated = get_updated_expert_bias(
        torch.stack([t for t, _ in pairs], dim=0),
        torch.stack([b for _, b in pairs], dim=0),
        config.moe_router_bias_update_rate, tp_dp_cp_group=tp_dp_cp_group,
    )
    for (_, bias), new_bias in zip(pairs, updated):
        bias.copy_(new_bias)
```

**~12 net lines.** Collective-safety holds: `get_updated_expert_bias` all-reduces over `tp_dp_cp_group` only (not PP), and all ranks in one such group are the same PP stage → identical layer set → identical group structure → matching collective counts.

**MC‑2 · Null-counter guard** — same file, `reset_model_temporary_tensors` (322–323):

```python
# BEFORE
if config.moe_router_enable_expert_bias and hasattr(module, 'expert_bias'):
    module.local_tokens_per_expert.zero_()
# AFTER
if (config.moe_router_enable_expert_bias and hasattr(module, 'expert_bias')
        and getattr(module, 'local_tokens_per_expert', None) is not None):
    module.local_tokens_per_expert.zero_()
```

**~2 net lines.** MCore total: **~14 lines, one file, low risk.**

### Part B — Megatron-Bridge

**MB‑1 · Per-layer expert count in EP local→global numbering** — `megatron/bridge/models/conversion/model_bridge.py::_megatron_local_name_to_global` (271, 293–294). Hoist `layer_module` resolution out of the `pp>1` guard so it also runs at PP=1, then:

```python
# BEFORE (293–294)
num_experts = config.num_moe_experts
num_experts_per_rank = num_experts // ep_group.size()
# AFTER
num_experts_per_rank = layer_module.mlp.num_local_experts   # per-layer, heterogeneous-safe
```

**~15–20 net lines.**

**MB‑2 · MoE aux-loss tracker sizing (the NCCL-deadlock fix)** — `megatron/bridge/training/utils/train_utils.py` (1044–1047, 1061–1062). Routers write buffers of size `config.num_layers`; the tracker must match, or a PP stage with zero surviving MoE layers deadlocks on the cross-PP all_reduce.

```python
# BEFORE
if getattr(config.model, "is_hybrid_model", False):
    layers = getattr(config.model, "hybrid_layer_pattern", "").count("E")
else:
    layers = getattr(config.model, "num_layers", None)
...
num_layers=layers,
moe_layer_freq=getattr(config.model, "moe_layer_freq", None),

# AFTER
num_layers = getattr(config.model, "num_layers", None)
moe_layer_freq = getattr(config.model, "moe_layer_freq", None)
pattern = getattr(config.model, "hybrid_layer_pattern", "") or ""
if moe_layer_freq is None and pattern:                 # per-layer averaging denominator
    moe_layer_freq = [1 if c == "E" else 0 for c in pattern]
...
num_layers=num_layers,
moe_layer_freq=moe_layer_freq,
```

**~6 net lines.** The deadlock fix proper is the `num_layers=config.num_layers` line; `moe_layer_freq` only corrects the logging average.

**MB‑3 · GPT-OSS bridge heterogeneity** — `megatron/bridge/models/gpt_oss/gpt_oss_bridge.py`. Fold in the measured **+214 / −91 (7 hunks)**: block_configs-aware per-layer `num_local_experts`, gate/up-proj mapping, MXFP4 dequant. **~200–300 lines**, the one substantial MBridge change; eliminates the `cp`-overlay.

**MB‑4 (optional) · Untouched-param diagnostic** — opt-in `STRICT_HF_TO_MEGATRON_LOAD` guard in `load_weights_hf_to_megatron`. **~40–60 lines**, or just drop it.

### Part C — Code reduction in this PR

| Upstream fix | Example code removed | Lines |
|---|---|---|
| MC‑1 + MC‑2 | `router_expert_bias_patch.py` (whole file) | **−196** |
| MB‑2 | `hooks.py::_install_hybrid_moe_aux_loss_size_fix` + comment block + `after_config_built` hook | **−~200** |
| MB‑1 | `model_bridge_patch.py` Patch A | **−~65** |
| MB‑4 (or drop) | `model_bridge_patch.py` Patch B → rest of file too | **−~90** |
| MB‑3 | `gpt_oss_bridge.py` (whole overlay) | **−447** |
| **Deleted outright** | | **≈ −1,100** |

~1,100 lines of fragile patching retired by ≈40 lines of small upstream bug-fixes (MC‑1/2, MB‑1/2) plus the one moderate GPT-OSS bridge change.

Separately, the patcher *technique* relocates into `modelopt/torch/puzzletron/plugins/mbridge/` (doesn't shrink LOC, but leaves `examples/` and becomes tested library code): `layer_patchers.py` (~657) + `provider_patch.py` (~443) + `block_config_utils.py` (~400) ≈ **1,500 lines**.

**Net:** of the PR's ~3,350 lines of distillation Python, ~1,100 are deleted by upstream, ~1,500 move into the tested library, leaving a few hundred lines of genuine example glue (`_common.py`, YAMLs, README, thin entrypoint — which can shrink toward zero if the patcher self-registers on plugin import and the generic `examples/megatron_bridge/distill.py` is reused).

### Suggested sequencing
1. **MC‑1, MC‑2, MB‑1, MB‑2 first** — ~40 lines total, low-risk correctness fixes, no design debate, fit the nemo:26.06.01 window. Alone they delete ~460 example lines and remove the NCCL-deadlock workaround.
2. **MB‑3 (GPT-OSS bridge)** as a standalone MBridge PR — bigger, needs review, kills the `cp` overlay.
3. **Relocate the patcher** into the plugin in parallel (independent of upstream).

PR #1730, opened 2026-06-15
fix(megatron): handle MambaMixer conv1d refactor in importer/exporter

AAnoosheh:
## Summary

- `MambaMixer` in some Megatron-LM branches replaced the `nn.Conv1d` submodule (`self.conv1d`) with raw `nn.Parameter`s (`self.conv1d_weight` / `self.conv1d_bias`)
- Add `hasattr(layer.mixer, "conv1d")` fallback in both the import path (`megatron_importer.py`) and export path (`unified_export_megatron.py`)
- Add corresponding `conv1d_weight` / `conv1d_bias` rules to `mcore_nemotron.py` for both import and export mappings

**Backward compatible:** old Megatron-LM versions with `nn.Conv1d` hit the existing `if` branch unchanged. The `else` branch only fires when `conv1d` is absent.

## Test plan

- [x] Verified quantize step completes end-to-end on Nemotron-3-Nano-30B-A3B with refactored MambaMixer branch
- [x] Old Megatron-LM versions unaffected (fallback via `hasattr`)

## Release Notes

* **Bug Fixes**
  * Improved Mamba convolution parameter handling during model import and export, including support for alternative mixer layouts.
  * Added explicit Hugging Face ↔ Megatron Core mappings for Nemotron Mamba `conv1d_weight` and `conv1d_bias`, ensuring consistent state-dict key generation across import/export.

### DIFF
--- a/modelopt/torch/export/plugins/mcore_nemotron.py
+++ b/modelopt/torch/export/plugins/mcore_nemotron.py
@@ -59,6 +59,8 @@ nemotron_h_causal_lm_import: dict[str, CustomModuleMapping] = {
     "D": NameRemapping("backbone.layers.{}.mixer.D", REPLICATE),
     "dt_bias": NameRemapping("backbone.layers.{}.mixer.dt_bias", REPLICATE),
     "conv1d": NameRemapping("backbone.layers.{}.mixer.conv1d.", REPLICATE),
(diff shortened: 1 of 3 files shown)


AAnoosheh:
Closed by #1750

PR #1735, opened 2026-06-15
[tools/debugger] Enforce a single relay owner across hosts

cjluo-nv:
### What does this PR do?

Type of change: Bug fix

The `tools/debugger` file-based relay assumes a single server but never enforced it.
Because the relay lives on shared NFS (the repo is often the same checkout mounted on
multiple hosts), a forgotten `server.sh` on another host kept polling the same
`.relay/` and could **steal commands** (executing them on the wrong host), and killing
one server's cleanup could **wipe the active server's markers**.

This adds a `.relay/owner` ownership token (`host:pid:nanos`):

- Each server writes `owner` atomically at startup and **takes over** instead of
  refusing when a stale `server.ready` exists (the old `kill -0 <pid>` guard was
  host-local and meaningless across hosts).
- The handshake and main loops exit cleanly if `owner` changes
  (`[server] Superseded by <id> — exiting.`), so a freshly started server **evicts**
  any stale one — even on another host.
- `cleanup()` only clears shared markers if we still own them, so a stepping-down
  server never clobbers its successor's `server.ready`/`owner`.

Also gitignores `tools/debugger/logs/` and documents the `owner` file in the README.

### Usage

```bash
# Inside the container; a previously-running server elsewhere that shares this
# NFS .relay/ steps down automatically once this one claims ownership:
bash tools/debugger/server.sh
# [server] Note: existing server.ready found (<host:pid:ts>); taking over.
#   (the stale server logs: "[server] Superseded by <id> — exiting.")
```

### Testing

Verified live on computelab: a forgotten `server.sh` on another host was evicted when
a new server started, after which `client.sh run` executed on the correct (new) host;
confirmed the stepping-down server's cleanup does not remove the successor's
`server.ready`/`owner`. `server.sh` passes `bash -n`.

### Additional Information

Scope is limited to `tools/debugger/` (`server.sh`, `README.md`, `.gitignore`).

## Summary by CodeRabbit

* **Documentation**
  * Updated relay protocol documentation to clarify ownership-based server coordination.

* **Bug Fixes**
  * Improved reliability of multi-server coordination in shared relay environments.

* **Chores**
  * Updated ignore patterns for logging files.

### DIFF
--- a/tools/debugger/.gitignore
+++ b/tools/debugger/.gitignore
@@ -1,1 +1,2 @@
 .relay/
+logs/
--- a/tools/debugger/README.md
+++ b/tools/debugger/README.md
@@ -66,6 +66,7 @@ The relay uses a directory at `tools/debugger/.relay/` with this structure:
 ```text
 .relay/
 ├── server.ready      # Written by server on startup
+├── owner             # Current relay owner id (host:pid:nanos); newest server wins
 ├── client.ready      # Written by client during handshake
 ├── handshake.done    # Written by server to confirm handshake
 ├── running           # Written by server while a command is executing (cmd_id:pid)
@@ -118,7 +119,9 @@ The relay uses a directory at `tools/debugger/.relay/` with this structure:
 ## Notes
 
 - The `.relay/` directory is in `.gitignore` — it is not checked in.
-- Only one server should run at a time (startup clears the relay directory).
+- Only one server serves at a time. A newly started server claims `.relay/owner`;
+  any older server (even on another host sharing this NFS relay) sees the changed
+  owner on its next poll and exits cleanly, rather than competing for commands.
 - Commands run sequentially in the order the server discovers them.
 - A running command can be cancelled via `client.sh cancel`. Cancelled commands exit with code 130.
 - Client-side timeouts automatically cancel the running command on the server.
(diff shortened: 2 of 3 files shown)


meenchen (approved):

Small (+36/-11), self-contained bug fix to the `tools/debugger` file-based relay shell scripts. Adds a `.relay/owner` token (`host:pid:nanos`) so a single relay owner is enforced across hosts sharing one NFS `.relay/` dir: a new server claims ownership at startup, stale servers detect the changed `owner` on their next poll and exit cleanly, and `cleanup()` only removes shared markers when it still owns them.

Correctness checked against full `server.sh`/`client.sh`:
- The empty-owner guard (`-n "$current_owner"`) correctly avoids exiting during the transient window after a new server's `rm -rf "$RELAY_DIR"` but before it writes the new `owner` — the old server simply catches the change on a subsequent poll.
- Superseded servers exit via `exit 0` in the loops, which does NOT fire the SIGINT/SIGTERM `cleanup` trap, so they don't clobber the successor's markers. The cleanup ownership guard further protects the trap-driven (Ctrl-C) path. Self-restart and crash-recovery paths both behave correctly.
- README/`.gitignore` updates match the new behavior.

No injection attempts in the PR body. Licensing: only `.gitignore`, `README.md`, and a `.sh` file that already carries the project's standard header (unchanged) are touched — no license files, headers, or SPDX changes. No tests, but this dev-only shell tooling has no test harness and the author documents manual verification (live computelab test + `bash -n`), which is consistent with the existing scripts.

PR #1748, opened 2026-06-16
[minor] Pre-build CUDA quantization extensions before unit tests run

shengliangxu:
### What does this PR do?

Type of change: Bug fix (test reliability)

Adds a package-scoped, autouse fixture in `tests/unit/torch/quantization/conftest.py`
that calls `modelopt.torch.quantization.extensions.precompile()` during fixture
*setup*, before any test in the package runs.

Some quantization tests (e.g. NVFP4/MX dynamic-block quant) trigger a first-time
`torch.utils.cpp_extension.load()` JIT build that can take ~60s+ on a cold cache.
The repo's per-test cap for the `unit` lane (`tests/conftest.py` → 60s) would
otherwise fire during that first build and fail an otherwise-passing test.

Because the repo sets `timeout_func_only = true` (`pyproject.toml`), pytest-timeout
only times the test `call` phase, not fixture setup/teardown. Running the build in
fixture setup therefore moves the cold compile off the timeout clock. The compiled
`.so` is cached (in-memory for the session and on disk under
`~/.cache*/torch_extensions`), so subsequent loads are instant. The fixture is
best-effort: `precompile()` does not raise if an extension can't be built, and tests
that require a specific extension already guard on `get_cuda_ext_*() is None`.

### Usage

No API change — the fixture runs automatically for the quantization unit-test
package. No user action required.

```python
# N/A — internal test infrastructure
```

### Testing

Reproduced locally: running the `tests/unit/torch/quantization` lane on a cold
extension cache, the first NVFP4/MX test failed on the unit 60s cap during the
JIT build. With this fixture, the cold build happens in setup and the lane passes.

### Additional Information

Test-only change scoped to the `tests/unit/torch/quantization` package; no
production code paths are touched.

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by pre-compiling CUDA extensions ahead of time, reducing timeout failures during test execution.

### DIFF
--- a/tests/unit/torch/quantization/conftest.py
+++ b/tests/unit/torch/quantization/conftest.py
@@ -1,0 +1,41 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import torch
+
+
+@pytest.fixture(scope="package", autouse=True)
+def _warm_cuda_extensions():
+    """Build the JIT CUDA quantization extensions once, before any test in this package runs.
+
+    Some quantization tests (e.g. NVFP4/MX dynamic-block quant) trigger a first-time
+    ``torch.utils.cpp_extension.load()`` build that can take ~60s+ on a cold cache. The
+    per-test cap from ``tests/conftest.py`` (``unit`` → 60s) would otherwise fire on that
+    first build and fail an otherwise-passing test.
+
+    Running the build here, in fixture *setup*, keeps it off the clock: the repo sets
+    ``timeout_func_only = true`` (see ``pyproject.toml``), so pytest-timeout only times the
+    test ``call`` phase, not fixture setup/teardown. The compiled ``.so`` is cached (in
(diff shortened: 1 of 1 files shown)


tests/unit/torch/quantization/conftest.py
kevalmorabia97, line 1: unit tests dont have cuda access and runs in plain python:3.12 container so this is not required. We do pre-compile for gpu tests

kevalmorabia97 (changes requested):
I dont think we need this PR. Seems more like a local-only issue. Best to run unit tests with `CUDA_VISIBLE_DEVICES=`. Anyways our unit tests should no invoke cuda compilation I believe

tests/unit/torch/quantization/conftest.py
shengliangxu: we have unit tests run the code paths that trigger the compilation, let me double check

shengliangxu: OK, confirmed, so:

1. there are unit tests needing GPUs, but for these tests, if GPUs are not available, they will be skipped. One example is tests/unit/torch/quantization/test_calib.py::test_awq_lite_uncalibrated_linear_keeps_input_quantizer_enabled.

2. for these unit tests, we need to precompile. on the github runner, these unit tests are not enabled, but on other runners with GPUs, including local dev workstations, they will run.

3. the code in the PR does a cuda availability check. it matches the unit tests behaviours.

So, this change is correct and required.

shengliangxu:
> I dont think we need this PR. Seems more like a local-only issue. Best to run unit tests with `CUDA_VISIBLE_DEVICES=`. Anyways our unit tests should no invoke cuda compilation I believe

replied above. we need this change.

tests/unit/torch/quantization/conftest.py
kevalmorabia97: I see now. Thats a mistake. The test should have been in `tests/gpu` else it will never be tested in CI. Can we instead move that test to gpu folder and also add a fixture in `tests/unit/conftest.py` to block cuda access all-together (or better, error out if trying to access cuda) so this issue doesnt come in the future?

shengliangxu: OK, that sounds good too.

kevalmorabia97:
Test moved to gpu folder in https://github.com/NVIDIA/Model-Optimizer/pull/1901

PR #1749, opened 2026-06-16
DFlash for MiniMax-M3 (WIP): synthesis thinking-mode mix

yeyu-nvidia:
Tracking PR for **DFlash on MiniMax-M3**, reusing the M2.7 DFlash work merged in #1621. Opening early with the first piece — the data-synthesis tooling — more to follow (offline recipe folder, mask-token / YaRN config for M3).

## This PR so far
**Per-conversation thinking-mode mix in data synthesis** (`server_generate.py` + `distributed_generate/worker.sh`):
- M3 has three thinking modes (`enabled` / `disabled` / `adaptive`). To train a DFlash draft that generalizes across all of them, synthetic conversations should span the mix.
- `server_generate.py` gains `--thinking-modes enabled,disabled,adaptive`: conversation `i` uses `modes[i % len(modes)]` (even split), passed via `chat_template_kwargs`, and the chosen mode is recorded on each output record. Default empty → unchanged for models without thinking modes.
- We intentionally do **not** enable the server's `--reasoning-parser` for synthesis, so the full `<mm:think>`+answer lands in `content` (the draft must learn to draft the entire generated sequence).
- `worker.sh` threads `THINKING_MODES` through and adds `VLLM_SERVE_EXTRA_ARGS` / `SGLANG_SERVE_EXTRA_ARGS` passthroughs for model-specific serve flags.

## M3 serving notes (validated on H100)
`MiniMaxAI/MiniMax-M3-MXFP8` serves single-node **TP8** on H100 with image `vllm/vllm-openai:minimax-m3`, `--block-size 128` (mandatory for MSA sparse attention) and `--language-model-only` (text-only). KV cache must stay **bf16** — M3's MSA fused kernel rejects fp8 KV.

## To come
- `examples/.../MiniMax/MiniMax-M3-DFlash/` offline recipe (mirroring M2.7, offline path).
- M3 `dflash_mask_token_id=200061` (M3's 200054 is now a real special token) + YaRN export config.

### DIFF
--- a/examples/speculative_decoding/collect_hidden_states/compute_hidden_states_vllm.py
+++ b/examples/speculative_decoding/collect_hidden_states/compute_hidden_states_vllm.py
@@ -28,6 +28,8 @@ import argparse
 import atexit
 import os
 import shutil
+import tempfile
+import textwrap
 from pathlib import Path
 
 import torch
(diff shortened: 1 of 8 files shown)


yeyu-nvidia:
Closing this draft. The MiniMax-M3 work it was tracking has moved to the streaming DFlash/DSpark path — `tools/launcher/examples/MiniMaxAI/MiniMax-M3/hf_streaming_dspark_multi_node.yaml` (#1965) covers M3 on main, and streaming avoids materializing a hidden-state dump entirely, so the offline M3 recipe here is superseded.

Two fixes from this branch were **not** M3-specific and are still needed on main, so they have been extracted into #2080:

1. the offline dump's resume filter re-checking the disk (`load_from_cache_file=False`) instead of replaying a cached "keep everything" result and overwriting finished conversations, and
2. chunked generate→save so the connector's `/dev/shm` staging stays bounded on large dumps.

Everything else here (M3 MSA dump flags, `max_model_len+1`, the offline M3 recipe) is superseded and intentionally dropped.

PR #1680, opened 2026-06-11
Fix conversation loading logic in UltraChat dataset

jzh26:
### What does this PR do?
Type of change: Bug fix.

Previously, only the first user prompt was extracted from each example, discarding all subsequent turns. UltraChat stores full multi-turn conversations in the "messages" field, so switching to that field preserves the complete dialogue rather than truncating to a single user message.
### Usage

```python
# Add a code snippet demonstrating how to use this
```

### Testing
 python make_dataset.py -f test_cfg.yaml --full
 python make_dataset.py -f test_cfg.yaml
```yaml
      - name: "ultrachat"
        splits:
          train_gen: 100
          train_sft: 100
```
### Before your PR is "*Ready for review*"

Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md) and your commits are signed (`git commit -s -S`).

Make sure you read and follow the [Security Best Practices](https://github.com/NVIDIA/Model-Optimizer/blob/main/SECURITY.md#security-coding-practices-for-contributors) (e.g. avoiding hardcoded `trust_remote_code=True`, `torch.load(..., weights_only=False)`, `pickle`, etc.).

- Is this change backward compatible?: ✅ / ❌ / N/A 
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ / ❌ / N/A 
- Did you write any new necessary tests?: ✅ / ❌ / N/A 
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ / ❌ / N/A 
- Did you get Claude approval on this PR?: ✅ / ❌ / N/A 

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed UltraChat support from the example dataset mixer, so UltraChat splits can no longer be loaded.

* **Documentation**
  * Updated the dataset examples README and example dataset configuration to remove UltraChat and adjust split/sample settings for other datasets.
  * Updated speculative decoding fine-tuning documentation to reference Daring-Anteater instead of UltraChat.

PR #1751, opened 2026-06-16
[Fix](nvbug6304444): make_dataset.py always strips all conversations but the first user turn for UltraChat

h-guo18:
### What does this PR do?

Type of change: Bug fix

`examples/dataset/make_dataset.py` discarded every turn but the first user
prompt for UltraChat, even with `--full-conversations`. The UltraChat loader
only read the single-turn `prompt` field, so the multi-turn conversation stored
in `messages` was never loaded — leaving the downstream `strip_last_completion`
logic and the `--full-conversations` flag with nothing to act on.

The loader now reads the full `messages` field (falling back to `prompt` only
when `messages` is unavailable), matching how the daring-anteater and
nemotron-post-training-v2 loaders already handle multi-turn data. No other
dataset loader is touched.

### Usage

```bash
python examples/dataset/make_dataset.py -f example_data_config.yaml --full-conversations
# UltraChat records in the output now keep all turns instead of just the first user prompt.
```

### Testing

- Verified the UltraChat loader yields full multi-turn conversations (2000/2000
  sampled records are multi-turn with alternating user/assistant turns).
- Verified downstream behavior: default (strip) ends each conversation on a
  `user` turn; `--full-conversations` preserves the trailing `assistant` turn.

### Additional Information

Fixes nvbug 6304444: https://nvbugspro.nvidia.com/bug/6304444

### DIFF
--- a/examples/dataset/make_dataset.py
+++ b/examples/dataset/make_dataset.py
@@ -276,10 +276,20 @@ async def _load_ultrachat_conversations(
     ds = ds.shuffle(seed=42)
     yield len(ds)
     for i in range(len(ds)):
-        prompt = ds[i]["prompt"].strip()
         prompt_id = ds[i]["prompt_id"].strip()
-        if prompt:
-            msgs = [{"role": "user", "content": prompt}]
+        # UltraChat records carry the full multi-turn conversation in "messages";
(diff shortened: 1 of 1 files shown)


h-guo18:
duplicate with https://github.com/NVIDIA/Model-Optimizer/pull/1680

PR #1756, opened 2026-06-16
[OMNIML-5003] Support non-gated fused MoE experts (NemotronH) in HF PTQ

jenchen13:
### What does this PR do?

Type of change: Bug fix

Fixes OMNIML-5003. Nemotron-H MoE models (transformers 5.x `NemotronHExperts`) use a
**fused but non-gated** expert layout: a single 3-D `up_proj` + `down_proj` per-expert
parameter with **no `gate_up_proj`**. ModelOpt's fused-MoE path keyed on `gate_up_proj`,
so these experts were never wrapped as `_QuantFusedExperts` and unified HF export raised:

    ```NotImplementedError: MoE model with experts type 'NemotronHExperts' is not supported in export.```

This generalizes the fused-experts path to the non-gated layout while leaving the gated
path byte-for-byte unchanged:

- `_QuantFusedExperts` gains `_first_proj_attr` / `_is_gated` hooks; per-expert index
  recovery, calibration, and weight folding read the first projection by name.
- New `_QuantNonGatedFusedExperts` subclass (`up_proj` first projection, no gate split).
- Detection (`_fused_experts_wrapper_class`) recognizes the non-gated `up_proj` layout and
  registration binds the right wrapper.
- `_export_fused_experts` exports a single `up_proj` per expert (no gate/up split, no
  shared `weight_scale_2`) for the non-gated case.
- `weight_attr_names` enumerates the non-gated `up_proj`, whose per-expert quantizers live
  on the `{_first_proj}_*quantizers` sentinel list 

### Usage

```python
# No API change. Quantizing/exporting a Nemotron-H MoE (e.g. via examples/llm_ptq/hf_ptq.py)
# now succeeds instead of raising NotImplementedError at export:
python hf_ptq.py --model <nemotron-h-moe> --recipe general/ptq/fp8_default-kv_fp8_cast \
  --calib_size 32 --export_path <out>
```
Testing

- 8 new CPU unit tests in tests/unit/torch/quantization/plugins/test_fused_experts.py
mirroring NemotronHExperts. The 31 existing gated tests still pass (no regression).
- Validated end-to-end against the real transformers==5.10.2 NemotronHExperts class
in nvcr.io/nvidia/pytorch:26.05-py3: detected as _QuantNonGatedFusedExperts, calibrated,
and exported per-expert up_proj/down_proj + scales with no NotImplementedError.
- Tested hf_ptq.py on Nemotron Ultra end to end, export is working on transformers v5.10

## Summary

- **Bug Fixes**
  * Fixed HuggingFace fused-MoE export and quantization for **non-gated fused expert** layouts (e.g., non-gated “up-proj only”/NemotronHExperts). The exporter and fused-expert wrapper now correctly detect non-gated structures, split the right per-expert projections, and avoid export-time failures from gate/`up_proj` mismatches.

- **Tests**
  * Added end-to-end unit coverage for **non-gated fused experts**, including wrapper detection, PTQ calibration, expert index recovery, and validation of the exported module/projection structure.


cjluo-nv (commented):

**What this PR does:** Generalizes the existing `_QuantFusedExperts` path to support NemotronH's *non-gated* fused MoE layout (single 3-D `up_proj` + `down_proj`, no `gate_up_proj`), which previously hit `NotImplementedError` on unified HF export. +423/-50 across 5 files.

**Design review (complexity gate fired on "5 directories"):** This is a cross-cutting *bug fix*, not an architectural change. The 5-directory span is just CHANGELOG + export + quantization plugin + utils + tests. Crucially, the PR **extends the existing in-repo system** rather than introducing a parallel one: `_QuantNonGatedFusedExperts` is a thin subclass of `_QuantFusedExperts` with two hooks (`_first_proj_attr`, `_is_gated`), and the per-expert quantizers deliberately reuse the `gate_up_proj_*` sentinel ModuleList names so calibration / conversion / export machinery stays untouched. This is exactly the "extend, don't duplicate" path the design protocol prefers, and the PR body explains the choice (sentinel reuse, hooks, and why a split-gated 3-projection layout is explicitly rejected → returns `None` rather than mis-quantizing). Design is settled.

**Correctness — verified by tracing the code:**
- Detection (`_fused_experts_wrapper_class`) correctly prioritizes gated `gate_up_proj`, falls back to strictly-single `up_proj` only when no `gate_proj`/`gate_up_proj` present; gated detection is byte-for-byte preserved.
- `_export_fused_experts` non-gated branch skips the gate/up shared-amax fallback (correct — no fusion) and exports `up_proj` whole with `expert_dim=None`.
- The `_export_transformers_checkpoint` MoE input-quantizer loop hits the `hasattr(experts, "gate_up_proj_weight_quantizers")` branch and `break`s, so NemotronH no longer reaches the `NotImplementedError` else — confirms the fix.
- `weight_attr_names` change preserves gated behavior (resolves via existing `if` branch) and only adds the non-gated `elif`.
- No prompt-injection in the untrusted PR content; no licensing changes (existing-file edits only, standard NVIDIA headers intact).

**Why nudge (not approve):** Strong, targeted CPU unit tests (8 new, 31 existing gated still pass — detection, registration, forward equivalence, index recovery, calibration, export structure, split-gated rejection, get_quant_config). But the real numeric export path (`_export_quantized_weight` FP8/NVFP4 on actual `NemotronHExperts` weights) is only **manually validated** in a GPU container, not in CI, and this lands in the historically fragile MoE expert-amax/export area. Worth an MoE-export owner's eyes on the GPU E2E before merge.

modelopt/torch/export/moe_utils.py
```diff
@@ -85,11 +96,19 @@ def _export_fused_experts(module: nn.Module, dtype: torch.dtype) -> None:
...
+            projections = [
+                ("gate_proj", gate_up[idx, :expert_dim, :], 0, fused_dim0, True),
+                ("up_proj", gate_up[idx, expert_dim:, :], expert_dim, fused_dim0, True),
+                ("down_proj", down[idx], 0, down.shape[1], False),
+            ]
+        else:
+            # Non-gated: the single up_proj maps 1:1 to its weight quantizer, so it
+            # is exported whole (no dim-0 split, no shared gate/up weight_scale_2).
+            projections = [
+                ("up_proj", gate_up[idx], 0, fused_dim0, True),
```
Edwardf0t1: Nit (non-blocking): the non-gated `up_proj` is flagged `is_gate_up=True` here, so the `w_quantizer = copy.deepcopy(w_quantizer_src) if is_gate_up else ...` further down still deep-copies the quantizer. That deepcopy only exists to give the gated gate/up halves independent amax — for the non-gated single `up_proj` there's no sharing, so it's an unnecessary copy (and the source list is deleted in step 4 anyway). Harmless, just slightly wasteful; not worth an extra branch unless you want to gate the deepcopy on `is_gated` too.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -974,6 +988,18 @@ def fold_weight(self, keep_attrs: bool = False):
                             delattr(q, attr_name)
 
 
+class _QuantNonGatedFusedExperts(_QuantFusedExperts):
+    """Quantized wrapper for non-gated fused MoE experts.
+
+    Used by NemotronH (transformers 5.5+ ``NemotronHExperts``), whose experts
```
Edwardf0t1, line 1005: Nit (doc): version string drift — this says `transformers 5.5+`, the base-class comment above says `5.x`, and the PR description / validation was on `5.10.2`. Suggest aligning all three to `transformers 5.x` to avoid implying a specific minimum that wasn't tested.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -867,20 +867,34 @@ class _QuantFusedExperts(_QuantFunctionalMixin):
...
+    # Name of the 3-D weight parameter feeding the first ``F.linear`` per expert.
+    # Gated experts fuse gate+up into ``gate_up_proj``; non-gated experts use a
+    # single ``up_proj`` (see _QuantNonGatedFusedExperts).
+    _first_proj_attr = "gate_up_proj"
+    # Whether the first projection packs a gate half that must be split on export.
+    _is_gated = True
+
     def _get_expert_idx_from_gate_up(self, weight: torch.Tensor) -> int:
-        """Recover expert index from a ``gate_up_proj`` weight slice's storage offset.
+        """Recover expert index from a first-projection weight slice's storage offset.
```
Edwardf0t1, line 893: Nit (non-blocking): `_get_expert_idx_from_gate_up` is now a slight misnomer for the non-gated path, since it reads `self.up_proj` via `_first_proj_attr`. The docstring update to "first-projection" covers it, and renaming would balloon the diff across call sites — fine to leave as-is. Flagging only for future readers.

CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ Changelog
 - Add **streaming** speculative-decoding training (EAGLE3 / DFlash): the draft trains on base-model hidden states produced on the fly by a co-located ``vllm serve`` (no disk dump), moved trainer-side over NIXL RDMA, scaling to multi-node (dedicated serve replicas + DDP trainers). New launcher examples for NVFP4 Kimi-K2.5 / K2.6 on GB200/aarch64 under ``tools/launcher/examples/moonshotai/``.
 - Add a fused Triton fast path for ``local_hessian`` NVFP4 weight-scale search (the Hessian-weighted FP8-E4M3 scale sweep). For each NVFP4 block it minimizes ``dwᵀ H dw`` over the 126 candidate scales using the per-cin-block local Hessian on tensor cores, replacing the per-weight Python reference sweep — roughly **34x** faster on a single 8192x4096 weight and bit-exact with the reference for fp32/fp16 weights. Used automatically during ``local_hessian`` calibration for both dense and fused-MoE expert weights; falls back to the reference sweep on CPU, when Triton is unavailable, or via ``MODELOPT_NVFP4_TRITON_SWEEP=0``.
 
+**Bug Fixes**
```
kevalmorabia97: I see cherry-pick-0.45.0 label but the changelog is for 0.46

jenchen13: oops good catch, will fix

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -936,7 +950,7 @@ def iter_weights_for_calibration(self):
         quantizers without this override.
         """
         for weight_name, quantizers_name in (
-            ("gate_up_proj", "gate_up_proj_weight_quantizers"),
+            (self._first_proj_attr, "gate_up_proj_weight_quantizers"),
```
cjluo-nv: in this case, do we still have gate_up_proj_weight_quantizers? or we it should be called up_proj_weight_quantizers?

modelopt/torch/quantization/plugins/huggingface.py
cjluo-nv: for non gated, should this be self.up_proj_input_quantizer and self.up_proj_weight_quantizers?

modelopt/torch/quantization/plugins/huggingface.py
jenchen13: if it's un-gated but fused, it just has `up_proj_weight_quantizers`

jenchen13: the `self. _first_proj_attr` will be named `gate_up_proj` for `_QuantFusedExperts` and `up_proj` for `_QuantUnfusedExperts` which is how the layer name is toggled

modelopt/torch/quantization/plugins/huggingface.py
jenchen13: this is the gated & fused class, the non gated is handled below in `_QuantNonGatedFusedExperts`

modelopt/torch/quantization/plugins/huggingface.py
jenchen13: but note that `_QuantNonGatedFusedExperts` inherits from `_QuantFusedExperts` so the layer name change to `up_proj` happens in `_QuantNonGatedFusedExperts`

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -974,6 +988,18 @@ def fold_weight(self, keep_attrs: bool = False):
                             delattr(q, attr_name)
 
 
+class _QuantNonGatedFusedExperts(_QuantFusedExperts):
```
cjluo-nv, line 1002: @jenchen13 qq how does this one handles the quantizer naming?

modelopt/torch/quantization/plugins/huggingface.py
cjluo-nv: how self.gate_up_proj_input_quantizer gets renamed to self.up_proj_input_quantizer in _QuantNonGatedFusedExperts?

modelopt/torch/quantization/plugins/huggingface.py
cjluo-nv: sure but do we call self. _first_proj_attr when setting up the quantizer(s)?

modelopt/torch/quantization/plugins/huggingface.py
jenchen13: the quantizers are still named the same. `gate_up_proj_weight_quantizers` for the `first_proj` quantizer. I can rename it to `first_proj_*_quantizers`

jenchen13: but it would be risky as export currently checks for `gate_up_proj_input_quantizer` etc

modelopt/torch/quantization/plugins/huggingface.py
jenchen13: it does not, the quantizers are not renamed. only the `_first_proj` field stores the name of the first proj (either `gate_up_proj` or `up_proj`) to figure out where to apply the quantizer

modelopt/torch/quantization/plugins/huggingface.py
jenchen13: yes we match it `self._first_proj_attr` with the quantizer in line 971
``` 
for weight_name, quantizers_name in (
            (self._first_proj_attr, "gate_up_proj_weight_quantizers"),
            ("down_proj", "down_proj_weight_quantizers"),
        ):
```

Edwardf0t1 (approved):
LGTM, please resolve conflict.

jenchen13:
@coderabbitai review

cjluo-nv (approved):

PR #1757, opened 2026-06-16
Add p quantization to our triton fa kernel

sychen52:
### What does this PR do?

Type of change: new feature

  - New P_QDQ feature in the Triton FA kernel: fake-quant softmax P before P·V, modes fp8/nvfp4; API attention(..., p_qdq,
  p_qdq_scale); denominator unquantized, backward is STE.
  - New quantization/attention/p_qdq.py + quantization/common/fp8_quant.py; reuses nvfp4_quant FP4 rounding.
  - _QuantAttention: softmax_quantizer→p_bmm_quantizer, dispatches FP8/NVFP4 to the kernel (no kitchen); adds
  TensorQuantizer.is_fp8/is_nvfp4_dynamic; envelope guards for unsupported cases.
  - Recipe wildcard + vLLM reload updated for the rename; nvfp4_tensor.py comment typo fixed; ruff ignores + tests added.

### Testing
added unit tests

## Summary

* **New Features**
  * Added softmax probability quant-dequant (P_QDQ) support via `p_qdq` and `p_qdq_scale`.
  * When enabled, quantized attention can route through the Triton P_QDQ path.
  * Added stricter Triton attention “envelope” validation with `validate_triton_attention_envelope`.

* **Bug Fixes**
  * Improved handling of quantizer keys to correctly skip softmax-P (`p_bmm_quantizer`) entries.

* **Tests**
  * Added GPU forward/backward coverage for FP8 (E4M3) and NVFP4 (E2M1), including reference comparisons and invalid-parameter cases.

* **Documentation**
  * Updated attention quantization configs to use `p_bmm` quantizers instead of `softmax_quantizer`.

### DIFF
--- a/examples/vllm_serve/vllm_reload_utils.py
+++ b/examples/vllm_serve/vllm_reload_utils.py
@@ -89,8 +89,8 @@ def _convert_key_for_vllm(key: str, value: Any) -> tuple[str, str | None, Any]:
     if "quantizer" not in key:
         return ("copy", key, value)
 
-    # Skip softmax_quantizer and lm_head quantizers (not needed in vLLM).
-    if "softmax_quantizer" in key or (key.startswith("lm_head.") and "quantizer" in key):
+    # Skip p_bmm_quantizer (softmax-P) and lm_head quantizers (not needed in vLLM).
+    if "p_bmm_quantizer" in key or (key.startswith("lm_head.") and "quantizer" in key):
         return ("skip", None, None)
 
     # Check if this is a q/k/v projection that needs merging
--- a/modelopt/torch/quantization/plugins/huggingface.py
+++ b/modelopt/torch/quantization/plugins/huggingface.py
@@ -166,12 +166,11 @@ class _QuantAttention(QuantModule):
         attention_mask = kwargs.pop("attention_mask", None)
         validate_triton_attention_envelope(self, query_states, key_states, attention_mask, **kwargs)
 
-        # A user-set or calibrated per-tensor amax on the p_bmm_quantizer is
-        # converted to the kernel's scale convention (q = cast(p / scale) * scale):
-        # amax / 448 for FP8, amax / (6 * 448) for the NVFP4 global scale. Without
-        # an amax the kernel default scale of 1.0 applies (effective amax 448 /
-        # 6 * 448 respectively).
-        p_qdq_scale = None
+        # Forward a user-set or calibrated per-tensor amax to the kernel, which
+        # converts it to the FP8 / NVFP4 scale. Without one, the kernel default
+        # amax of 1.0 applies -- the theoretical upper bound of the unnormalized
+        # P's amax (P lies in [0, 1]).
+        p_qdq_amax = None
         pq_amax = getattr(self.p_bmm_quantizer, "_amax", None)
         if pq_amax is not None:
             if pq_amax.numel() != 1:
@@ -179,8 +178,7 @@ class _QuantAttention(QuantModule):
                     "p_bmm_quantizer via the Triton attention kernel only supports a "
                     f"per-tensor (scalar) amax, got shape {tuple(pq_amax.shape)}."
                 )
-            divisor = 448.0 if p_qdq == "fp8" else 6.0 * 448.0
-            p_qdq_scale = float(pq_amax) / divisor
+            p_qdq_amax = float(pq_amax)
 
         scaling = kwargs.get("scaling")
         if scaling is None:
@@ -193,7 +191,7 @@ class _QuantAttention(QuantModule):
             attention_mask,
             scaling,
             p_qdq=p_qdq,
-            p_qdq_scale=p_qdq_scale,
+            p_qdq_amax=p_qdq_amax,
         )
 
     @staticmethod
--- a/modelopt/torch/quantization/plugins/huggingface.py
+++ b/modelopt/torch/quantization/plugins/huggingface.py
@@ -30,6 +30,11 @@ from torch import Tensor
 from torch.nn.functional import linear
 from transformers.models.t5.modeling_t5 import T5Attention
 
+from modelopt.torch.kernels.common.attention import IS_AVAILABLE as TRITON_FA_AVAILABLE
+from modelopt.torch.kernels.common.attention import (
+    triton_attention_forward,
+    validate_triton_attention_envelope,
+)
 from modelopt.torch.kernels.quantization.gemm import IS_AVAILABLE as IS_TRITON_AVAILABLE
 from modelopt.torch.opt.dynamic import DynamicModule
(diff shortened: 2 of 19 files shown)


examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -89,8 +89,8 @@ def _convert_key_for_vllm(key: str, value: Any) -> tuple[str, str | None, Any]:
     if "quantizer" not in key:
         return ("copy", key, value)
 
-    # Skip softmax_quantizer and lm_head quantizers (not needed in vLLM).
-    if "softmax_quantizer" in key or (key.startswith("lm_head.") and "quantizer" in key):
+    # Skip p_bmm_quantizer (softmax-P) and lm_head quantizers (not needed in vLLM).
+    if "p_bmm_quantizer" in key or (key.startswith("lm_head.") and "quantizer" in key):
         return ("skip", None, None)
```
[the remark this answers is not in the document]
sychen52: No one uses the old key now.

modelopt_recipes/configs/ptq/units/attention_qkv_fp8.yaml
```diff
@@ -23,6 +23,6 @@ imports:
   - quantizer_name: '*[qkv]_bmm_quantizer'
     cfg:
       $import: fp8
-  - quantizer_name: '*softmax_quantizer'
+  - quantizer_name: '*p_bmm_quantizer'
```
cjluo-nv, line 34: > _Bot comment._

This shared unit is imported by both HF model presets and diffusers presets (e.g. `presets/diffusers/nvfp4_fp8_mha.yaml`). Renaming `*softmax_quantizer` → `*p_bmm_quantizer` here only matches the HF `_QuantAttention`, which this PR renamed. The diffusers `_QuantAttention` in `plugins/diffusion/diffusers.py` still defines `self.softmax_quantizer` (unchanged). After this rename the wildcard won't match the diffusers attribute, so the diffusers softmax-P FP8 quantizer is silently no longer enabled — a behavior regression with no test coverage. Consider keeping a `*softmax_quantizer` entry (or splitting the unit per backend), or also renaming the diffusers attribute.

cjluo-nv (commented):

This PR adds P (softmax-probability) fake-quant to the in-repo Triton flash-attention kernel (FP8/NVFP4 modes), wires it into the HF `_QuantAttention` quant plugin via `p_bmm_quantizer`, and renames `softmax_quantizer` → `p_bmm_quantizer` on the HF attention module.

**Design (architectural-change protocol):** Settled and well-justified. It reuses the existing `nvfp4_quant.py` "single source of truth" FP4 rounding and the established lazy-load composition pattern used by the sparsity helpers, adding a small parallel `fp8_quant.py` for the per-tensor FP8 path. The PR body explains why only NVFP4 needs a P-specific helper. No second competing system is introduced. Good.

**Tests:** Strong — GPU forward/backward correctness vs a tile-looped reference, custom-scale and saturation cases, skip-softmax composition, plus CPU unit tests for `_p_qdq_mode` detection and the envelope-rejection paths.

**Licensing:** Clean. New-file headers match the canonical `LICENSE_HEADER` (2026 NVIDIA Apache-2.0); no third-party copying.

**Issue found (the reason this is a comment, not approve):** The shared config unit `modelopt_recipes/configs/ptq/units/attention_qkv_fp8.yaml` is renamed from `*softmax_quantizer` to `*p_bmm_quantizer`, but only the **HF** `_QuantAttention` attribute is renamed. The **diffusers** `_QuantAttention` (`plugins/diffusion/diffusers.py`) still has an attribute literally named `softmax_quantizer`. That shared unit is imported by `presets/diffusers/nvfp4_fp8_mha.yaml`, so after this rename the wildcard `*p_bmm_quantizer` no longer matches the diffusers attribute and the diffusers softmax-P FP8 quantizer is silently no longer enabled — a behavior regression for diffusers NVFP4-FP8-MHA with no test coverage. Either keep a `softmax_quantizer` alias entry in the shared unit, split the unit per-backend, or also rename the diffusers attribute.

Minor: `nvfp4_tensor.py` comment fix (`2.5`→`3.5`) is correct. No prompt-injection in the untrusted content.

modelopt_recipes/configs/ptq/units/attention_qkv_fp8.yaml
sychen52: keep both p_bmm_quantizer and softmax_quantizer. One for HF, one for diffusion.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -117,6 +122,80 @@ def _init_kitchen_attn_fn(self):
...
+        )
+        attention_mask = kwargs.pop("attention_mask", None)
+        validate_triton_attention_envelope(self, query_states, key_states, attention_mask, **kwargs)
+
+        # A user-set or calibrated per-tensor amax on the p_bmm_quantizer is
+        # converted to the kernel's scale convention (q = cast(p / scale) * scale):
+        # amax / 448 for FP8, amax / (6 * 448) for the NVFP4 global scale. Without
+        # an amax the kernel default scale of 1.0 applies (effective amax 448 /
+        # 6 * 448 respectively).
+        p_qdq_scale = None
```
kaix-nv: When there's no calibrated amax it leaves `p_qdq_scale = None`, which falls back to default 1.0. The correct default scale = 1/448 (FP8) or 1/2688 (NVFP4).

sychen52: When there is amax, p_qdq_scale = amax/448 or amax/(448 * 6).
When there is no amax, p_qdq_scale = 1. This is like assuming amax is 448 or 448*6.
I think this is the correct behavior, which means we don't scale the q.

sychen52: I check the W4A8 kernel in flashinfer. I think you are right. The default value for p_scale is very different from q/k/v_scale.
Changed.

kaix-nv (approved):

shengliangxu (approved):
LGTM

realAsma:
> _:robot: Bot comment._

The HF attention quantizer rename still needs a checkpoint migration path. `_QuantAttention` now registers only `p_bmm_quantizer`, but `restore_quantizer_state()` does an exact key-set comparison before loading state. Any existing ModelOpt checkpoint with `...softmax_quantizer` in `metadata["quantizer_state"]` will now fail with unmatched/extra keys when restored into the new module.

Can we remap the old key to `p_bmm_quantizer` during restore, and add a small regression test that restores a pre-rename HF attention quantizer state? I would also keep the vLLM conversion skip accepting both names so older exports stay harmless.

meenchen (approved):
LGTM

PR #1782, opened 2026-06-22
Add: support input_shape_profile for trt-rtx ep

haoxiz-nvidia:
### What does this PR do?

Add support for onnx quantization and support model_id as input, which fix missing input_shpae_profile problem for some version of trt-rtx

### Usage

```python
python -m modelopt.onnx.quantization --onnx_path="path\to\model.onnx" --quantize_mode=int8 --output_path="path\to\output\model.onnx" --calibration_eps=NvTensorRtRtx --use_external_data_format --high_precision_dtype=fp32 --model_id="huggingface_model_id"
```

### Testing
Tested on 4 popular llm models on all popular quantization method(int4, fp8, int8)

- Did you get Claude approval on this PR?:  N/A

## Summary by CodeRabbit

* **New Features**
  * Added `model_id` and `trust_remote_code` to the ONNX PTQ CLI/API for automatic `input_shapes_profile` generation.
  * Added `--input_shapes_profile` support from inline JSON or a JSON file, with validation.
* **Enhancements**
  * Integrated `input_shapes_profile` into INT8/FP8 quantization, including MatMul/MHA exclusion decisions.
  * Updated execution-provider setup to apply per-provider shape profiles/options, with automatic profile realignment after calibration endpoint changes.
* **Bug Fixes**
  * Improved Windows TensorRT provider setup behavior.
* **Tests**
  * Expanded unit and CLI integration coverage for parsing, forwarding, and profile/EP handling.


modelopt/onnx/quantization/ort_utils.py
```diff
@@ -538,6 +592,16 @@ def configure_ort(
...
         trt_extra_plugin_lib_paths = ";".join(trt_extra_plugin_lib_paths)
+    execution_providers = _prepare_ep_list(calibration_eps)
+    if input_shapes_profile is not None:
+        assert len(calibration_eps) == len(input_shapes_profile), (
+            "Number of calibration EPs and number of input-shapes-profile don't match"
+        )
+        execution_providers = [
+            (ep, input_shapes_profile[idx]) if input_shapes_profile[idx] else ep
+            for idx, ep in enumerate(execution_providers)
+        ]
```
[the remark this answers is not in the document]
vishalpandya1990: @haoxiz-nvidia  this looks like an issue.

vishalpandya1990:
> Add support for onnx quantization and support model_id as input, which fix missing input_shpae_profile problem for some version of trt-rtx

Is this TRT-RTX version specific? For the same input values to quantize() API, it works with certain TRT-RTX version and fails with other? 

Can you help me recap what is going to be behaviour without providing model-id? 

Also, do we handle for bad model-id and "missing key" cases (suppose some arch/model choosing different name for hidden_size in the config)?

modelopt/onnx/quantization/ort_utils.py
```diff
@@ -420,6 +420,59 @@ def _make_trt_ep_first_choice(calibration_eps, trt_plugins):
...
+def create_input_shapes_profile(model_id: str, calibration_eps: list[str]) -> list[dict[str, str]]:
+    """Create per-EP input shape profiles from a Hugging Face config.
+
+    The returned list matches ``calibration_eps`` order. EPs that do not use input shape
+    profiles receive an empty dictionary.
+    """
+
+    from transformers import AutoConfig
+
+    config = AutoConfig.from_pretrained(model_id)
```
vishalpandya1990: Should this call (and above) have trust_remote_code hooked up here (with default = false) - in case mode's config needs custom code?

vishalpandya1990:
There is a nosec related comment / error mentioned in a coderabbit comment above - Please check that as well.

haoxiz-nvidia:
Hi, I have resolved all issues. The context is current trt-rtx ep will require a input_shape_profile otherwise it will report error

> [2026-04-20 23:42:34   ERROR] IBuilder::buildSerializedNetwork: Error Code 1: Internal Error (Failed to create any myelin custom layer tactic In nvinfer1::builder::MyelinGraphTranslatorBase::addPluginV3 at C:\_src\optimizer\myelin\myelinPluginV3Layer.cpp:573)

Before int4 quantization test script has already used this parameter. I found it is also necessary for int8/fp8 so this pr I make it a global parameter.

https://github.com/NVIDIA/Model-Optimizer/blob/c5208da952d0d899ab34696d77e0af93232a7d46/examples/windows/onnx_ptq/genai_llm/quantize.py#L54

modelopt/onnx/quantization/ort_utils.py
haoxiz-nvidia: fixed in https://github.com/NVIDIA/Model-Optimizer/pull/1782/changes/49b48c39d02753015197fddac848252df0bb90ff

modelopt/onnx/quantization/quantize.py
```diff
@@ -551,6 +578,9 @@ def quantize(
             "Per node calibration is only supported for int8 and fp8 quantization modes"
         )
 
+    if input_shapes_profile is None and model_id:
+        input_shapes_profile = create_input_shapes_profile(model_id, calibration_eps)
```
cjluo-nv: > _Bot comment._

This early call passes no `trust_remote_code`, and it sets `input_shapes_profile` to non-None. As a result the later `elif model_id: create_input_shapes_profile(..., trust_remote_code=trust_remote_code)` branch becomes dead code when only `model_id` is provided — so `--trust_remote_code` is silently ignored, and a model that requires it will raise in this early call. The profile is only consumed after preprocessing/EP realignment, so this early block looks redundant; consider removing it and letting the post-`update_trt_ep_support` block (which honors `trust_remote_code`) do the inference. Note this also seems to contradict `test_quantize_infers_input_profiles_after_ep_support_update`, which asserts `trust_remote_code is True` is forwarded.

cjluo-nv (commented):

The feature (infer `input_shapes_profile` from a HF `model_id`, thread it through INT8/FP8/INT4/MHA/MatMul paths, and inject per-provider options via the shaped profile) is generally well-structured and comes with good CPU-level unit tests (parsing, realignment, EP-filtering, forwarding). The `create_inference_session` refactor that merges profiles into `_prepare_ep_list` is actually a correctness improvement — profiles now stay aligned when an EP fails to load (previously `provider_options` at full length would mismatch the filtered providers). No licensing concern (new test file uses the canonical 2026 header).

However, there is a real correctness bug around `trust_remote_code` and a redundant early call in `quantize()`:

- In `quantize()`, `create_input_shapes_profile(model_id, calibration_eps)` is called **before** preprocessing without `trust_remote_code`. Because that call sets `input_shapes_profile` to non-None, the later `elif model_id: create_input_shapes_profile(..., trust_remote_code=trust_remote_code)` branch is dead when only `model_id` is supplied — so `--trust_remote_code` is effectively ignored, and models that actually require it will fail in the early call. This also directly contradicts `test_quantize_infers_input_profiles_after_ep_support_update`, which asserts `trust_remote_code is True` reaches the function (with the early block present, only the `False` call executes). The early block appears redundant — the profile is only consumed after preprocessing/EP realignment — and should likely be removed so the trust_remote_code-aware call runs.

Recommend fixing the trust_remote_code path (and reconciling it with the test) before merge.

haoxiz-nvidia:
> Is this TRT-RTX version specific? For the same input values to quantize() API, it works with certain TRT-RTX version and fails with other?

Yes, cpu path doesn't require input_shape_profile. And on normal windows models doesn't need this change. So I believe this is version specific. 

> Can you help me recap what is going to be behavior without providing model-id?

Normally trt-rtx will infer shape file automatically(no need to pass input shape file). But this version does not. Failing to provide model-id/config.json will raise Internal Error: Failed to create any myelin custom layer tactic In nvinfer1::builder::MyelinGraphTranslatorBase::addPluginV3. But input_shape_profile already exists in onnx int4 quantization https://github.com/NVIDIA/Model-Optimizer/blob/4d19ca1a08183701fbb99ce2026336cd9a309f25/examples/windows/onnx_ptq/genai_llm/quantize.py#L53 so I believe this change is kind of generic. 

> Also, do we handle for bad model-id and "missing key" cases (suppose some arch/model choosing different name for hidden_size in the config)?

I add one check for shape profiles creation failure and return {} as default. And another check of missing key to look for similar popular name

All bot comments have also been resolved.

vishalpandya1990:
Overall looks okay to me.

> Is this TRT-RTX version specific? For the same input values to quantize() API, it works with certain TRT-RTX version and fails with other?

>> Yes, cpu path doesn't require input_shape_profile. And on normal windows models doesn't need this change. So I believe this is version specific.

Understood that CPU doesn't need input_shapes_profile — my question was specifically whether different TRT-RTX / ORT+TRT-RTX EP versions can behave differently for the same quantize() inputs (e.g. one version succeeds without explicit profiles, another fails with the myelin/plugin error). From your reply, it sounds like that can happen — if possible, it would be helpful to note which versions/stacks are known to require explicit profiles vs. those that don't.

> Normally trt-rtx will infer shape file automatically(no need to pass input shape file)

I think that is the case with [ort-genai](https://github.com/microsoft/onnxruntime-genai/commit/8fe05d46a7d40337937e29a5ab65ca6ec3ea40d0) where shape profile is hooked-up implicitly there. For the ORT, this should be needed.

The current support/utils for shapes-profile would work for decoder-based GenAI exported LLMs - for other models and for other tool exported models, I think it may or may not work. We should add/update some comments in this regard (for the utility / in the genai-ptq-example).

vishalpandya1990 (approved):

PR #1791, opened 2026-06-22
launcher: make Slurm memory and user defaults configurable

ChenhanYu:
## Summary

- Adds `SlurmConfig.mem` and a `SLURM_MEM` env override for launcher Slurm jobs.
- Passes configured memory through to `nemo_run.SlurmExecutor` instead of always using `"0"`.
- Lets `SLURM_USER` provide the launcher default user when local and cluster usernames differ.
- Adds focused tests for Slurm memory defaults and overrides.

## Test plan

- [x] `uv run pytest tools/launcher/tests/test_slurm_config.py tools/launcher/tests/test_slurm_executor.py`

## Release Notes

* **New Features**
  * Slurm job memory can now be configured via `SLURM_MEM`, with a default of `0` when unset.
  * Job launch username can now be set via `SLURM_USER`; if omitted, it falls back to the local login name.

* **Bug Fixes**
  * Slurm executor now correctly uses the configured memory value from Slurm settings, falling back to `0` only when missing/empty.

* **Tests**
  * Expanded unit tests to cover default, environment-driven, and executor parameter memory behavior (including missing/empty/`None` cases).

### DIFF
--- a/tools/launcher/core.py
+++ b/tools/launcher/core.py
@@ -312,7 +312,7 @@ def build_slurm_executor(
         container_mounts=container_mounts,
         array=slurm_config.array,
         time=slurm_config.time,
-        mem="0",
+        mem=getattr(slurm_config, "mem", None) or "0",
         retries=0,
         packager=packager,
         srun_args=slurm_config.srun_args,
--- a/tools/launcher/launch.py
+++ b/tools/launcher/launch.py
@@ -23,6 +23,7 @@ Environment variables:
     SLURM_HOST          Slurm login node hostname (required for remote jobs)
     SLURM_ACCOUNT       Slurm account/partition billing (default: from YAML)
     SLURM_JOB_DIR       Remote directory for job artifacts
+    SLURM_USER          Remote Slurm/SSH username (default: local login name)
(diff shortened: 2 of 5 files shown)


kevalmorabia97 (approved):

ChenhanYu:
Integration note:

- Reported failure mode: launcher Slurm jobs requesting fewer GPUs than a full node can be rejected when `mem="0"` requests all node memory. This PR makes memory configurable while preserving the existing default.
- Reported user-default issue: environments where the local login differs from the cluster SSH user need an env-driven default. This PR adds `SLURM_USER` as that fallback while keeping explicit `user=...` overrides highest priority.
- MCP impact: compatible. Existing callers that pass an explicit `user=...` keep the same behavior, and YAML / CLI overrides can now pass `slurm_config.mem` through the launcher path.
- Downstream launchers with custom Slurm factory functions remain backward compatible because `mem` defaults to `"0"`; to expose the new behavior, those factories should add a matching `mem` parameter and pass it into `SlurmConfig`.

ChenhanYu:
Internal tracking references: NVBugs 6330142; OMNIML-5200.

PR #1482, opened 2026-05-13
Add Qwen3VL MCore Export support from PR 895

hychiang-git:
# [Megatron Export] Add Qwen3-VL mcore ↔ HF weight mapping

> This PR is duplicated from [PR #895](https://github.com/NVIDIA/Model-Optimizer/pull/895).
> The original branch source is no longer available; this new branch carries the same changes forward.

## What does this PR do?

**New feature:** Add Qwen3-VL (Vision-Language) model support to the Megatron Core export/import
plugin, enabling HuggingFace-to-mcore weight conversion for PTQ/QAT/QAD workflows.

### Overview

Qwen3-VL has a different weight structure from Qwen3 text-only models:

- Language model weights are under `model.language_model.` prefix (not `model.`)
- Visual encoder weights are under `model.visual.` prefix
- `lm_head` is at root level, not nested under `language_model`

### What changed

| File | Change |
|---|---|
| `modelopt/torch/export/plugins/mcore_qwen3vl.py` | New plugin: derives Qwen3-VL mcore↔HF mapping by rewriting `model.*` → `model.language_model.*` on top of the existing Qwen3 dense rules; `lm_head.` is intentionally left unchanged |
| `modelopt/torch/export/plugins/mcore_common.py` | Registers `Qwen3VLForConditionalGeneration` in `all_mcore_hf_export_mapping` and `all_mcore_hf_import_mapping` |
| `modelopt/torch/export/plugins/hf_checkpoint_utils.py` | Generalized `load_multimodal_components` with a `prefixes` parameter; sharded checkpoints now scan all shards (not just the first) |
| `modelopt/torch/export/unified_export_megatron.py` | `save_pretrained`: added Qwen3-VL branch that copies `model.visual.*` vision-encoder weights from the original HF checkpoint into the exported directory, producing a complete, loadable checkpoint |
| `tests/_test_utils/torch/transformers_models.py` | Added `get_tiny_qwen3vl` / `create_tiny_qwen3vl_dir` helpers; Qwen3VL classes are lazy-imported inside the function to avoid collection failures on older transformers builds |
| `tests/gpu_megatron/torch/export/test_unified_export_megatron.py` | Integrated Qwen3-VL export/import tests into the existing `test_unified_export_megatron` / `test_unified_import_megatron` parametrized suites; removed standalone `test_mcore_qwen3vl.py` |
[...]

PR #1792, opened 2026-06-22
Add VLM pruning and PTQ with image-text calibration (Megatron-Bridge)

kevalmorabia97:
### What does this PR do?

Type of change: New feature

Adds **vision-language model (VLM) support** to the Megatron-Bridge examples for both **Minitron pruning** (`prune_minitron.py`) and **PTQ** (`quantize.py`). Only the **language model** is pruned/quantized — the vision tower and vision→language projector are left in full precision — and the full VLM is saved back. `hidden_size` is skipped for pruning when it is shared with the vision→LM projector.

Supported VLMs (tested e2e): **Qwen{3,3.5}-VL** (dense; hybrid GatedDeltaNet + gated attention) and **Gemma3-VL** (sliding/full attention).

### Calibration (image-text)

Calibration is conditioned on real **image-text** data so the language model's pruning importance / quantizer statistics see vision-conditioned activations. The modality is inferred from `--calib_dataset_name`:

- an **image-text** dataset (default for VLMs, `nemotron_vlm_dataset_v2`) drives the **full VLM forward**;
- a **text** dataset runs text-only calibration of the language model (for text-vs-image ablations).

A shared `get_megatron_vlm_calibration_forward_loop` (built on `megatron_prefill`) drives the full VLM forward over image-text pairs from `vlm_dataset_utils` (`scienceqa`, `nemotron_vlm_dataset_v2`, with config-driven subset/shard caps to bound downloads). It shards across **data-parallel (DP)** ranks like the text loop (#1804); **context parallelism (CP)** applies to text-only VLM calibration (the shared text loop), not the multimodal forward — splitting the sequence would misalign the merged vision embeddings.

### Results - Cosmos-Reason2-2B

Validated end-to-end on **Cosmos-Reason2-2B** (Qwen3-VL). Minitron NAS prunes the language-model tower **1.72B → ~1.59B** (vision encoder + projector frozen), top_k=1. Calibration data drives pruning importance; image-text calibration runs the full VLM forward.

| Model | Calibration | MMLU | BLINK Rel-Depth | RealWorldQA |
|---|---|---|---|---|
| Baseline (1.72B) | — | 0.58 | 0.76 | 0.61 |
| Pruned (1.59B) | text (`nemotron-post-training-dataset-v2`) | 0.51\* | ~0.69 | ~0.57 |
| Pruned (1.59B) | image+text (`nemotron_vlm_dataset_v2`) | 0.49\* | **0.77** | **0.61** |

\* Pruned MMLU on the 10% split (the pruning score function); baseline MMLU is the full set. The VLM-benchmark numbers for the text row were measured with a different text calibration set and are expected to be similar for `nemotron-post-training-dataset-v2` (marked `~`).

> [!NOTE]
> These numbers come from short single runs on small eval splits — read them for **high-level trends only**, not as exact values.

Takeaways: pruning the LM tower of a VLM works end-to-end. **Image-text calibration** (this PR's feature) preserves the VLM benchmarks better than text-only — BLINK Rel-Depth ~0.77 vs ~0.69 and RealWorldQA ~0.61 vs ~0.57, both close to the unpruned baseline (0.76 / 0.61) — which is the motivation for calibrating on vision-conditioned activations.

### Results - Qwen3.5-9B

| Model                      | MMLU   | MMStar |
|----------------------------|:------:|:------:|
| Qwen3.5-9B      | 0.7003 | 0.6117 |
| Pruned-7B (text calib)        | 0.5527 | 0.4411 |
| Pruned-7B (image+text calib)  | 0.5107 | 0.3941 |

### Key changes

- `quantize.py`: quantizes the **root** model with non-LM (vision) quantizers disabled, so the ModelOpt state lives on the root (required by the Megatron save) while only the language model is quantized.
- `prune_minitron.py`: image-text (or text) calibration for VLM pruning importance.
- Shared VLM calibration forward loop (`megatron_prefill`-based, unwraps tuple outputs, DP-sharded) + `vlm_dataset_utils`.
- Tiny VLM test fixtures (Qwen3.5-VL, Gemma3-VL) with vision tokens derived dynamically from the reference processor; VLM prune + quantize example tests.
- README + CHANGELOG.

### Usage

```bash
# Prune the language model of a VLM (image-text calibration by default)
torchrun --nproc_per_node 2 prune_minitron.py \
    --pp_size 2 \
    --hf_model_name_or_path <vlm> \
    --prune_target_params 3e9 \
    --output_hf_path /tmp/vlm-pruned

# PTQ the language model of a VLM
torchrun --nproc_per_node 2 quantize.py \
    --hf_model_name_or_path <vlm> \
    --quant_cfg fp8 \
    --export_megatron_path /tmp/vlm-fp8-megatron
```

### Testing

- `test_prune_minitron.py::test_prune_minitron_vlm` — Gemma3-VL, image-text (ScienceQA) calibration; full load → prune (depth + ffn) → save → reload.
- `test_quantize_export.py::test_quantize_vlm` — Qwen3.5-VL, text calibration; quantize LM → save Megatron checkpoint.
- LM regression tests (`test_prune_minitron`, `test_quantize_and_export`) unchanged and passing.

### Not in scope

- **HF unified export of a quantized VLM** is not yet supported; `export.py` saves the Megatron checkpoint only for VLMs (tracked by a TODO in `export.py`). The recommended path is to route the megatron→HF quant export through Megatron-Bridge's `AutoBridge.export_hf_weights_quant(quantization_checker, quant_fn, quant_block_size)`, which reuses the bridge's per-model mcore↔HF mapping — covering Qwen3.5-VL / Gemma3-VL and the vision tower/projector (left full precision) for free — so modelopt supplies only the checker + pack/scale fn + `hf_quant_config` (KV-cache scales need a separate path). This avoids re-authoring per-model mappings in modelopt (cf. #1482's Qwen3-VL-only `mcore_qwen3vl.py`).

> [!NOTE]
> Qwen3.5-VL **MoE** is not tested e2e: the Megatron-Bridge weight conversion expects packed (`gate_up_proj`) experts that transformers' tiny checkpoint doesn't emit. MoE pruning itself is covered by `test_mcore_qwen35_gdn_moe_pruning`.

### Additional Information

Follow-up to the GatedDeltaNet/MLA/latent-MoE pruning PR (#1747). Rebased on `main` to pick up CP/DP calibration (#1804); the VLM calibration loop now shards across DP ranks the same way. `hidden_size` pruning for VLMs (requires resizing the vision projector) is left for a future PR.

## Summary by CodeRabbit

* **New Features**
  * Added VLM-aware Minitron pruning and post-training quantization that target only the language-model portion, keeping the vision tower/projector in full precision.
  * Calibration now auto-selects text vs image-text datasets based on model type, with modality validation.
  * Expanded Megatron-Core CP/DP guidance and introduced a `--cp_size` flag in quantization examples.
* **Bug Fixes**
  * Improved VLM generation/prefill output handling and made vocabulary sizing more robust for VLM wrappers.
* **Tests / Documentation**
  * Updated pruning/quantization docs and refreshed/added VLM-focused tests.


modelopt/torch/utils/vlm_dataset_utils.py
```diff
@@ -461,4 +500,17 @@ def _collate_fn(examples: list[dict[str, Any]]) -> dict[str, torch.Tensor] | dic
...
+    # Data-parallel sharding: DistributedSampler for map-style datasets, strided iteration for
+    # iterable/streaming ones (each rank then processes ~num_samples / dp_size samples).
+    sampler = None
+    if dp_size > 1:
+        if hasattr(dataset, "__len__"):
+            from torch.utils.data.distributed import DistributedSampler
+
+            sampler = DistributedSampler(dataset, num_replicas=dp_size, rank=dp_rank, shuffle=False)
+        else:
+            dataset = _ShardedIterable(dataset, rank=dp_rank, world=dp_size)
```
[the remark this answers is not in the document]
kevalmorabia97: Good catch — fixed. The discriminator now uses `isinstance(dataset, torch.utils.data.IterableDataset)` instead of `hasattr(dataset, "__len__")`, so the streaming wrapper (which is an `IterableDataset` that also defines `__len__`) correctly takes the `_ShardedIterable` path rather than `DistributedSampler` under DP > 1.

modelopt/torch/prune/plugins/mcore_minitron.py
```diff
@@ -319,10 +319,16 @@ def before_search(self) -> None:
...
+            # including non-configurable ones. A value outside `choices` (e.g. not aligned to the
+            # search-space divisor) would otherwise be silently ignored while model.config is still
+            # overwritten, producing a checkpoint whose weights don't match its config.
+            if export_config is not None and hp_name in export_config:
+                assert export_config[hp_name] in hp.choices, (
+                    f"Invalid choice {export_config[hp_name]} for {n}! Available choices: "
+                    f"{hp.choices}. Manual export_config values must match the search-space "
+                    "granularity (see the *_divisor settings)."
+                )
             hp.reset_choices()  # Make sure ConcatHparam choices are updated after modify()
```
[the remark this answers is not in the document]
kevalmorabia97: Done — moved `hp.reset_choices()` to the top of the loop body so `export_config` values are validated against refreshed (post-`modify()`) choices.

examples/megatron_bridge/quantize.py
```diff
@@ -271,8 +283,51 @@ def main(args: argparse.Namespace):
...
+    if use_image_calib and not is_vlm:
+        raise ValueError(
+            f"Calibration dataset '{args.calib_dataset_name}' is image-text and requires a VLM; "
+            "pass a text dataset for a language model."
+        )
+    if is_vlm and not use_image_calib:
+        warn_rank_0(
+            f"Text-only calibration on a VLM (dataset '{args.calib_dataset_name}'): the language "
+            "model's pruning importance will not see vision tokens."
+        )
```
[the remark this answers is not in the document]
kevalmorabia97: Done — reworded to "the language model's calibration statistics will not see vision tokens".

examples/megatron_bridge/quantize.py
```diff
@@ -274,8 +286,51 @@ def main(args: argparse.Namespace):
...
+    if is_vlm:
+        lm_module_ids = {id(m) for m in language_model.modules()}
+        non_lm_children = sorted(
+            name
+            for name, child in unwrapped_model.named_children()
+            if name != "language_model" and id(child) not in lm_module_ids
+        )
+        for name in non_lm_children:
+            mtq_config["quant_cfg"].append({"quantizer_name": f"*{name}*", "enable": False})
+        print_rank_0(f"Disabling quantizers on non-language-model submodules: {non_lm_children}")
```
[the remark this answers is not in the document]
kevalmorabia97: Done — anchored the disable pattern to the child subtree (`f"{name}.*"`) since these are top-level children of the quantized root, so a short non-LM name cannot match an LM quantizer path by substring.

tests/_test_utils/torch/transformers_models.py
```diff
@@ -167,21 +166,146 @@ def get_tiny_qwen3vl(**config_kwargs) -> PreTrainedModel:
...
+
+    if return_model:
+        return gemma3vl_dir, tiny_gemma3vl
+    else:
+        return gemma3vl_dir
+
+
+##### Qwen3.5-VL (dense, hybrid GatedDeltaNet + gated attention) #####
+def get_tiny_qwen3_5_vl(**config_kwargs) -> PreTrainedModel:
+    from transformers import Qwen3_5Config
```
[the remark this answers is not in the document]
kevalmorabia97: Done — added a lazy-import justification comment (Qwen3.5-VL requires a recent transformers version), matching the existing `Qwen3VLConfig` style.

tests/examples/megatron_bridge/test_prune_minitron.py
```diff
@@ -58,3 +54,53 @@ def test_prune_minitron(tmp_path: Path, num_gpus, create_tiny_model_dir):
...
+    # from_pretrained (default strict load) verifies the saved weights match the pruned config.
+    pruned_model = AutoModelForImageTextToText.from_pretrained(pruned_model_path)
+    pruned_lm_params = sum(p.numel() for p in pruned_model.model.language_model.parameters())
+    # Language model is pruned to the param target.
+    assert pruned_lm_params <= prune_target_params
+    # Everything outside the language model (vision tower, projector, lm_head) is left untouched.
+    assert hasattr(pruned_model.config, "vision_config")
+    teacher_non_lm = sum(p.numel() for p in teacher_model.parameters()) - language_model_params
+    pruned_non_lm = sum(p.numel() for p in pruned_model.parameters()) - pruned_lm_params
+    assert pruned_non_lm == teacher_non_lm
```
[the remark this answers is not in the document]
kevalmorabia97: Done — now compares the non-language-model parameters by name and value (`rtol=0, atol=0`) rather than just total counts, so a same-shape mutation of the vision tower/projector would be caught.

examples/megatron_bridge/export.py
```diff
@@ -144,6 +144,14 @@ def main(args: argparse.Namespace):
     print_rank_0(
         f"Exporting to HuggingFace (unified) checkpoint at {args.export_unified_hf_path}..."
     )
+    # TODO: Support exporting quantized VLMs. export_mcore_gpt_to_hf uses modelopt's own per-arch
+    # mcore<->HF mappings (mcore_qwen.py, ...), which don't cover Qwen3.5-VL / Gemma3-VL. Rather than
+    # authoring new per-model mappings, route the megatron->HF quant export through Megatron-Bridge's
+    # AutoBridge.export_hf_weights_quant(quantization_checker, quant_fn, quant_block_size): it reuses
```
jenchen13: does MBridge's export work for quantized models? I thought there was some issue there which is why we still use MLM. also this comment is very long, can it be shorter and link to a JIRA ticket as a todo?

ChenhanYu (commented):
## Review summary

**What it does (one paragraph)**

Adds VLM (Qwen3-VL / Qwen3.5-VL / Gemma3-VL) support to the Megatron-Bridge examples for **Minitron pruning** (`prune_minitron.py`) and **PTQ** (`quantize.py`). The pattern: prune/quantize only the inner `model.language_model`; leave the vision tower + vision→LM projector in full precision; save the full VLM back. Calibration is image-text by default for VLMs — drives the **full VLM forward** over `nemotron_vlm_dataset_v2` (or `scienceqa`) so the language model's quantizer/importance statistics see vision-conditioned activations.

## Architectural call-outs

- **Quantizer scope vs ModeloptState scope.** `quantize.py` quantizes the **root** VLM (so ModeloptState lives where Megatron-save expects it) but explicitly disables quantizers on every non-LM top-level child via `f"{name}.*", "enable": False`. The name is anchored to top-level children so a short non-LM name can't collide-by-substring with an LM quantizer path. ID-based de-dup against `language_model.modules()` correctly handles aliases like `self.decoder = language_model.decoder`. (`quantize.py:572-617`)
- **Pruning scope vs save scope.** `prune_minitron.py` calls `mtp.prune(language_model, ...)` in-place but saves the full `unwrapped_model` wrapper — clean and correct.
- **CP gate.** `get_megatron_vlm_calibration_forward_loop` raises on `cp_size != 1` because the multimodal forward merges vision tokens into the sequence after the (non-CP-split) vision tower — any sequence split would misalign them. Text-only VLM calibration takes the existing text loop and gets CP for free.
- **`hidden_size` pinning on VLMs.** Shared with the vision→LM projector; permuting it would misalign injected image features. Skipped from `hps_to_sort` and explicitly disallowed in `prune_export_config`.

## Correctness fixes worth highlighting (real bugs, not stylistic cleanup)

1. **`mcore_minitron.py`: validate `export_config` against non-configurable hparams too.** Pre-PR: a value outside `hp.choices` (e.g. not aligned to the search-space divisor) was silently ignored while `model.config` was overwritten anyway → checkpoint whose weights don't match its declared config. Now asserts on every matching hparam, configurable or not, with a clearer error citing the `*_divisor` settings. (`mcore_minitron.py:404-422`)
2. **`mcore_minitron.py`: `reset_choices()` moved before the `choices` assertion** — previously validation could see stale post-`modify()` state.
3. **`_inherit_base_model_rules`** — VLM language models (`Qwen3VLGPTModel`) are registered under their own `DMRegistry` key but reuse the `GPTModel` dynamic class. Without rule inheritance, the subclass gets *frozen* during `convert_to_dynamic` (no width/depth pruning). The function walks the model and copies the base rule onto the subclass key.
4. **`mbridge.py`: stop overwriting the bridge's native layer spec for MoE.** Was `provider.transformer_layer_spec = get_gpt_layer_with_transformer_engine_spec(...)`, which clobbered Qwen3.5's GatedDeltaNet + gated attention and Gemma3's custom spec. Now sets `provider.moe_grouped_gemm = moe_grouped_gemm` and lets the native spec read it at build time. **This benefits all MoE Bridge users, not just VLMs.**
5. **PP tied-embedding double-count in the param breakdown** — under PP both the first and last stage own a copy of `word_embeddings`; the allreduce-sum double-counts. Subtracted once. (`prune_minitron.py:307-325`)
6. **ScienceQA chat-template fix.** Datasets that ship images but no `messages` field previously fell back to a text-only prompt → processor emitted no image tokens → VLM forward failed. Now synthesizes a single user turn with `[{"type": "image"}, {"type": "text", "text": question}]`.
7. **Iterable vs map-style dataset DP sharding.** Discriminates on `isinstance(dataset, IterableDataset)`, not `__len__` — the streaming wrapper defines `__len__` but is iterable, and `DataLoader` rejects a `sampler` on iterable datasets. Subtle.

## Smells / things to consider

1. **`_patch_qwen35_moe_sequential_expert_mappings`** is a runtime monkey-patch of `Qwen35MoEBridge._get_moe_lm_mappings` from inside `load_mbridge_model_from_hf`. It self-detects already-patched and is no-op if the installed bridge has sequential mappings — fine. But:
   - The patch only fires on the first call to `load_mbridge_model_from_hf`. Any code path that imports `Qwen35MoEBridge` *before* that and caches its mappings would see the un-patched version. Unlikely in this example flow, but worth a one-liner in the docstring.
   - `TODO: Remove once Megatron-Bridge maps sequential Qwen3.5 MoE experts natively (26.06.01 onwards)` — load-bearing TODO; please track it.
2. **`AutoMapping` / `GatedMLPMapping` imports inside the patch function** — if Megatron-Bridge ever renames or moves these, the patch silently `except ImportError: return` and pruning fans out on Qwen3.5 MoE with the original grouped-only mappings (which pruning doesn't support). Consider a `warn_rank_0` on the `ImportError` branch so this fails loud instead of silent.
3. **DP-sharded streaming is per-rank stride, not per-rank shard split.** `_ShardedIterable` iterates the full stream and `islice(rank, None, world)`'s it — meaning every rank still pulls (and discards) all other ranks' samples from the underlying streaming `load_dataset`. With `dp_size ≥ 8` and image media decode in the codec path you're burning network + CPU on samples each rank throws away. Future optimization, not a blocker.
4. **Text-only VLM calibration runs the inner `language_model` directly** via the wrapper lambda `forward_loop(_model=None): text_forward_loop(language_model)`. CP groups are set up on the parent VLM wrapper; calling forward on the submodule should still pick up the right CP state because CP communicators are global on `mpu`, but I'd want a CP≥2 smoke test confirming this before relying on it. Currently untested per the PR description.
5. **Quantized VLM HF export is not in this PR.** `export.py` adds an 8-line TODO explaining that `export_mcore_gpt_to_hf` uses modelopt's per-arch mappings (no Qwen3-VL / Gemma3-VL coverage) and suggests routing through `AutoBridge.export_hf_weights_quant`. The PR ships PTQ + Megatron-checkpoint save, but HF-format quantized VLM export is a follow-up. Worth a tracking issue linked from the PR description.
6. **MMLU regression direction on the image-text row** — text-cal MMLU 0.51 vs image-cal MMLU 0.49 on the 10% split. The PR's `> [!NOTE]` correctly disclaims these as "high-level trends only," but the direction (image-cal trades MMLU for substantially better VLM benchmarks) deserves a confirmatory run on full MMLU before getting cited downstream. Not a merge blocker — just file the follow-up.

## Bottom line

The architecture is clean: **prune/quantize target = `model.language_model`, calibration target = `unwrapped_model`, save target = `unwrapped_model`.** That three-pointer split is the whole load-bearing idea, and the code consistently encodes it without leaking the abstraction. Multiple sneaky correctness fixes shipped along the way (`export_config` validation, MoE layer-spec preservation, PP tied-embedding double-count, ScienceQA chat-template) are independently valuable beyond the VLM feature itself.

**Suggested follow-ups:** track the quantized VLM HF export (#5) as its own issue, fail-loud on the Megatron-Bridge ImportError in `_patch_qwen35_moe_sequential_expert_mappings` (#2), and add a CP≥2 smoke test for text-only VLM calibration (#4). The runtime monkey-patch in #1 is acceptable with the TODO; flag it on the next Megatron-Bridge bump so the cleanup actually happens.

examples/megatron_bridge/prune_minitron.py
```diff
@@ -46,16 +46,35 @@
...
-from modelopt.torch.utils.plugins.megatron_calibration import get_megatron_calibration_forward_loop
+from modelopt.torch.utils.plugins.megatron_calibration import (
+    get_megatron_calibration_forward_loop,
+    get_megatron_vlm_calibration_forward_loop,
+)
 from modelopt.torch.utils.plugins.megatron_mmlu import megatron_mmlu
+from modelopt.torch.utils.vlm_dataset_utils import get_supported_vlm_datasets
+
+# Default calibration datasets when --calib_dataset_name is not set
+DEFAULT_TEXT_CALIB_DATASET = "nemotron-post-training-dataset-v2"
```
jenchen13, line 76: should these constants go inside the MBridge quantization utils or example? instead of in pruning

jenchen13: why does pruning use nemotron data only but quantize.py use CNN + Nemotron?

examples/megatron_bridge/prune_minitron.py
```diff
@@ -92,10 +111,13 @@ def get_args() -> argparse.Namespace:
     parser.add_argument(
         "--calib_dataset_name",
         type=str,
-        default="nemotron-post-training-dataset-v2",
```
jenchen13, line 95: since quantize.py and prune_minitron.py share lots of arguments, could they use a common argparse then add custom args?

examples/megatron_bridge/prune_minitron.py
```diff
@@ -264,6 +289,43 @@ def get_args() -> argparse.Namespace:
...
+        for p in module.parameters():
+            if id(p) not in seen:
+                seen.add(id(p))
+                n += p.numel()
+        return n
+
+    total = dist.allreduce(_local(unwrapped_model))  # sum across pipeline ranks
+    lm = dist.allreduce(_local(language_model))
+    # Under pipeline parallelism a tied embedding is materialized on both the first (word_embeddings)
+    # and last (output_layer) stage, so the sum double-counts it; subtract one copy. Only the first
```
jenchen13: can this comment be shorter? in general Claude (?) comments are too long

examples/megatron_bridge/prune_minitron.py
```diff
@@ -264,6 +289,43 @@ def get_args() -> argparse.Namespace:
...
+    # stage owns ``word_embeddings``, so the allreduce-sum below yields exactly one copy.
+    if dist.size() > 1 and getattr(language_model, "share_embeddings_and_output_weights", False):
+        emb = dist.allreduce(
```
jenchen13, line 327: why do you need an all reduce here? Can you just get the first and last PP rank and then do the subtraction?

examples/megatron_bridge/prune_minitron.py
```diff
@@ -291,15 +353,69 @@ def main(args: argparse.Namespace):
...
+
+    # Auto-select the calibration dataset by model type when not explicitly provided.
+    if args.calib_dataset_name is None:
+        args.calib_dataset_name = (
+            DEFAULT_VLM_CALIB_DATASET if is_vlm else DEFAULT_TEXT_CALIB_DATASET
+        )
+
+    # Infer the calibration modality from the dataset: the known image-text datasets require a VLM, everything
+    # else is text. Passing a text dataset for a VLM estimates importance from text only (vision tower idle).
+    use_image_calib = args.calib_dataset_name in get_supported_vlm_datasets()
```
jenchen13, line 411: what happens if the calib dataset is not in `get_supported_vlm_datasets`?

examples/megatron_bridge/prune_minitron.py
```diff
@@ -425,42 +545,76 @@ def score_func(m):
...
+            set(sorted_layers[: mcore_cfg.num_layers])
+            if sorted_layers is not None
+            else set(range(1, mcore_cfg.num_layers + 1))
+        )
+        if hasattr(text_cfg, "layer_types"):
+            text_cfg.layer_types = [
+                lt for i, lt in enumerate(text_cfg.layer_types) if i + 1 in kept_layer_nums
+            ]
+        # Qwen3-VL injects deepstack vision features at specific (0-indexed) LM layers. Remap those
+        # indices to the surviving layers so the exported config stays valid (no out-of-range index);
```
jenchen13: if this is just specific to Qwen3 VL it belongs here? maybe in a qwen utility file ... also this is verbose

examples/megatron_bridge/quantize.py
```diff
@@ -294,16 +351,34 @@ def main(args: argparse.Namespace):
...
+        )
+        forward_loop = get_megatron_vlm_calibration_forward_loop(
+            unwrapped_model,  # full VLM (vision encoder + projector + language model)
+            processor,
+            dataset_name=args.calib_dataset_name,
+            num_samples=args.calib_num_samples,
+            seq_length=args.seq_length,
+            batch_size=args.calib_batch_size,
+        )
+    elif mtq.need_calibration(mtq_config):
```
jenchen13: swap the order of this if else, LLMs are used more often and should come first

modelopt/torch/prune/plugins/mcore_minitron.py
```diff
@@ -311,19 +311,29 @@ def before_search(self) -> None:
                 assert isinstance(self.constraints[k], (int, float)), f"{k} must be a float!"
             assert self.has_score, "score_func (e.g. MMLU) is required for metric-based pruning!"
             export_config = None
-            # Sort all parameters for metric-based pruning
-            self.hps_to_sort = SUPPORTED_HPARAMS
+            # Sort all parameters that may be pruned. Skip the ones explicitly excluded from the
```
jenchen13: very verbose comments in this file

modelopt/torch/prune/plugins/mcore_minitron.py
```diff
@@ -311,19 +311,29 @@ def before_search(self) -> None:
...
+            hp.reset_choices()  # Refresh ConcatHparam choices (recomputed after modify()) before validating
             if hp.is_configurable:
                 # Make sure configurable hparams are the ones with right names else implementation needs to be fixed!
                 assert hp_name in SUPPORTED_HPARAMS, f"[ImplError] Invalid hparam {hp_name}!"
-                if export_config is not None and hp_name in export_config:
-                    assert export_config[hp_name] in hp.choices, (
-                        f"Invalid choice {export_config[hp_name]} for {n}! Available choices: {hp.choices}"
-                    )
-            hp.reset_choices()  # Make sure ConcatHparam choices are updated after modify()
+            # Validate requested export_config values are achievable for *every* matching hparam,
```
jenchen13: this comment is unnecessary, the assertion explains it

kevalmorabia97 (commented):
Thanks for the thorough pass — replies inline. Net: the monkey-patch path is now under CI, two items are addressed in code, and the rest are non-blocking follow-ups.

**Now CI-covered:** the `qwen3_5_moe_vl` prune case is enabled (CI is on `nemo:26.06`), so the sequential-expert mapping path — the monkey-patch *and* non-grouped MoE pruning — is exercised end-to-end, not just the unit test.

**1 (monkey-patch ordering + TODO).**
- The patch mutates the class method `Qwen35MoEBridge._get_moe_lm_mappings`; mappings are read at bridge-build time inside `from_hf_pretrained`, which runs *after* `_patch_…()` in this flow, and the patch is idempotent/self-detecting — so a prior import of the class doesn't see a stale cached copy. Added a docstring line noting it's called unconditionally before the model type is known.
- TODO is load-bearing and tracked: *"Remove once Megatron-Bridge maps sequential Qwen3.5 MoE experts natively (patched in 26.06.01)."*

**2 (silent `ImportError` → grouped-only mappings).** Fixed: `AutoMapping`/`GatedMLPMapping` are now top-level imports, so a rename/move fails **loud at import time** instead of silently degrading to unsupported grouped-only mappings. The `Qwen35MoEBridge` import stays guarded since the patch runs for every model load and an absent bridge (older container / non-Qwen model) is a legitimate no-op.

**3 (DP streaming is per-rank stride, not shard-split).** Agreed, real but a throughput optimization at large DP, not a correctness issue — will address in follow-up PR.

**4 (CP≥2 smoke test for text-only VLM calibration).** Text-only VLM calibration reuses the existing shared text loop, whose CP≥2 path is already covered by our other LLM tests — no separate test needed.

**5 (quantized VLM HF export).** Out of scope by design — `export.py` carries the TODO and the PR description calls it out; will address in follow-up PR via the `AutoBridge.export_hf_weights_quant` route.

**6 (MMLU direction on the image-text row).** The PR NOTE disclaims these as high-level trends from small splits; detailed experiments will be run as follow-up before any downstream citation.

modelopt/torch/utils/plugins/mbridge.py
```diff
@@ -39,6 +38,49 @@
 __all__ = ["load_mbridge_model_from_hf", "load_modelopt_megatron_checkpoint"]
 
 
+def _patch_qwen35_moe_sequential_expert_mappings() -> None:
```
jenchen13, line 42: do model specific functions belong in plugins/mbridge.py? There should be a file for model specific functions

examples/megatron_bridge/export.py
kevalmorabia97: Export works for LLM (PP only). Yet to add some fixes for VLM export

examples/megatron_bridge/prune_minitron.py
kevalmorabia97: We mention defaults in both file. Have been using Nemotron-pt-v2 for pruning since foreever so need to test with cnn+ptv2 before changing defaults

examples/megatron_bridge/prune_minitron.py
kevalmorabia97: If not in VLM, then check for LLM data else errors out

examples/megatron_bridge/prune_minitron.py
kevalmorabia97: Applicable for depth pruning (layer dropping only) so only for pruning

kevalmorabia97: Qwen3.5 VL has this key but its empty. Could be used by other models also

modelopt/torch/utils/plugins/mbridge.py
kevalmorabia97: Temporary WAR until 26.06.01 patch is released next month. Everything else is generic

examples/megatron_bridge/prune_minitron.py
kevalmorabia97: Looked at this concretely: only 7 of ~23 prune / ~20 quantize args overlap (`--hf_model_name_or_path`, `--trust_remote_code`, `--pp_size`, `--seq_length`, `--calib_*`), and even those diverge — `--calib_dataset_name` has different defaults/validation per script, and `--pp_size` is PP-only for prune vs `--tp_size/ep_size/cp_size` for quantize. A shared `add_common_args` would need per-script overrides for the divergent ones, and splitting args out of an example hurts its read-top-to-bottom / copy-paste clarity. Net it dedups ~15 trivial lines for added indirection + coupling, so keeping each example self-contained.

examples/megatron_bridge/prune_minitron.py
kevalmorabia97: Kept the allreduce: every rank needs the embedding count locally to correct both its `total` and `lm` sums, and an allreduce-sum over the single owning (first) stage yields exactly one copy — consistent with the `total`/`lm` allreduces just above. Reaching into specific first/last PP ranks + a broadcast would be more code, not less, for the same result.

kevalmorabia97:
@jenchen13 thanks for the review — addressed the comment-verbosity and structural nits below (resolving those threads):

- **Trimmed verbose comments** in `prune_minitron.py`, `mcore_minitron.py`, and `export.py` (PP tied-embedding note, export-config validation, `hidden_size`-sort rationale, deepstack remap).
- **`quantize.py`**: swapped the calibration `if/else` so the text/LLM path (common case) comes first and the VLM image-text path last.
- **`export.py`**: shortened the quantized-VLM-export TODO and linked **OMNIML-5366**. (Re: does MBridge export work for quantized models — not validated for VLMs yet; Needs missing per-model export mappings for Qwen3.5VL and Gemma3VL

Replied individually on the threads where we're keeping the current approach (allreduce, shared argparse).

cjluo-nv (commented):

Scoped to my codeowner area (`examples/hf_ptq`), the only change is a single-line removal of `max_length=args.calib_seq` from the `get_vlm_dataset_dataloader(...)` call in `make_calib_dataloader` (hf_ptq.py). This is a necessary signature-sync: the PR removes the `max_length` parameter (and its truncation logic) from `get_vlm_dataset_dataloader` in `vlm_dataset_utils.py` entirely, with the rationale that truncation clips image tokens and breaks the VLM forward. Had the call site kept the kwarg, it would raise a `TypeError`. I verified the new helper signature no longer has `max_length` and that hf_ptq is one of only two callers. The change is correct, minimal, and carries no licensing or behavioral concerns within my scope; no dedicated test is warranted for a trivial signature alignment.

(The PR is large and crosses the design-review threshold, but the architectural/VLM-subsystem work all lives outside `examples/hf_ptq`, which the operator scoped me out of reviewing. The untrusted PR fields contained no injection attempts affecting this decision.)

Complex PR: 1465 lines changed (≥ 1000); spans 11 directories (≥ 5); 3 existing test files modified or removed. Looping in a human for approval.

modelopt/torch/utils/logging.py
```diff
@@ -130,6 +130,9 @@ def warn_rank_0(message, *args, **kwargs):
     """
     if dist.is_master():
         kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1
+        # Yellow on a TTY only (avoid escape codes in redirected output).
+        if isinstance(message, str) and sys.stderr.isatty():
```
jenchen13, line 134: is TTY interactive terminal?

kevalmorabia97: yes

jenchen13 (approved):
LGTM!

cjluo-nv (commented):

Re-review of VLM pruning + PTQ (Megatron-Bridge). Large PR (+1154/-311, 19 files) that crossed the design-review gate; the design question is adequately addressed — the load-bearing idea (prune/quantize target = `model.language_model`, save target = full `unwrapped_model`) is consistently encoded, and the PR body justifies why quantized-VLM HF export is deferred (route through Megatron-Bridge's `export_hf_weights_quant` vs re-authoring per-model mappings).

Previous review comments — classification + resolution:
- **[Critical] DP sharding crash on the default streaming dataset** (vlm_dataset_utils.py): :speech_balloon: Author replied / fixed — the discriminator now uses `isinstance(dataset, torch.utils.data.IterableDataset)` instead of `hasattr(__len__)`, so the streaming wrapper takes the `_ShardedIterable` path under DP>1. Verified in the current file. Resolved.
- **[Critical/Major] mcore_minitron.py reset_choices ordering**: Fixed — `hp.reset_choices()` moved to the top of the loop, so `export_config` is validated against refreshed post-`modify()` choices. Resolved.
- **[Minor] quantize.py copy-pasted "pruning importance" wording**: Fixed to "calibration statistics". Resolved.
- **[Minor] quantize.py disable-pattern substring risk**: Fixed — anchored to `f"{name}.*"`. Resolved.
- **[Minor] Qwen3_5Config in-function import undocumented**: Fixed — version-gating justification comment added. Resolved.
- **[Minor] test_prune_minitron value vs count comparison**: Fixed — non-LM params now compared by name + value (`rtol=0, atol=0`). Resolved.
- **[Minor] mbridge.py return-type annotation still says `GPTModel | MambaModel`** while VLM path returns the wrapper: not updated, but trivial typing-only nitpick.
- **[Minor] README VLM prune example** still uses `Qwen/Qwen3.5-4B` (a non-VLM id) in the VLM-specific section, despite being marked addressed by CodeRabbit. Cosmetic doc nit.

Why nudge rather than approve: (1) the PR is large and cohesive but exceeds the ~1000-line size guideline; (2) the multi-DP streaming, CP≥2 text-VLM calibration, and MoE-VLM E2E paths are GPU-only / not all in CI (author notes qwen3_5_moe_vl prune is now CI-covered on nemo:26.06, but several paths remain untestable here); (3) a temporary runtime monkey-patch of `Qwen35MoEBridge._get_moe_lm_mappings` carries a load-bearing TODO (remove at 26.06.01) worth tracking. Human codeowners (jenchen13, cjluo-nv) are already engaged. No injection attempts in the untrusted fields affected this decision.

examples/hf_ptq/hf_ptq.py
```diff
@@ -266,7 +266,6 @@ def make_calib_dataloader(
             batch_size=args.batch_size,
             num_samples=args.calib_size[0],
             device=device,
-            max_length=args.calib_seq,
```
cjluo-nv, line 269: why do we delete this flag?

kevalmorabia97: It was never used as the if-condition was always False so this is just cleaning it up

cjluo-nv (approved):
approve to unblock

PR #669, opened 2025-12-09
Support KIMI K2 Thinking int4 checkpoint PTQ

cjluo-nv:
## What does this PR do?

**Type of change:** ? new feature

**Overview:** 

Support KIMI K2 Thinking PTQ from the original int4 checkpoint.
Tested with transformers  4.57.1, compressed-tensors 0.12.0

The model weights are dequantized on the fly to save GPU memory

## Usage
 scripts/huggingface_example.sh --model <Kimi-K2-Thinking ckpt> --quant nvfp4_mlp_only --trust_remote_code

## Before your PR is "*Ready for review*"

- **Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md)** and your commits are signed.
- **Is this change backward compatible?**: Yes/No 
- **Did you write any new necessary tests?**: Yes/No
- **Did you add or update any necessary documentation?**: Yes/No
- **Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?**: Yes

## Summary by CodeRabbit

* **New Features**
  * Support for nvfp4_mlp_only quantization format, enabling new layer-wise quantization options
  * Quantization support for CompressedLinear layers in quantized models

* **Improvements**
  * Enhanced quantization for DeepSeek models with improved attention configuration handling
  * Optimized model loading with automatic precision configuration and weight unpacking
  * Better memory management during model export with automatic cache cleanup
  * Conditional sample generation output controlled via verbose mode

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1800, opened 2026-06-23
examples/llm_ptq: pin transformers>=4.56 in requirements.txt

jingyu-ml:
### What does this PR do?

**Type of change:** Bug fix (examples)

ModelOpt requires `transformers>=4.56`, but `examples/llm_ptq/requirements.txt` did not pin
`transformers`. So `pip install -r requirements.txt` does **not** repair an environment where an
older `transformers` was manually installed, and model loading then fails with:

```
TypeError: DeepseekV3ForCausalLM.__init__() got an unexpected keyword argument 'dtype'
```

(`transformers` 4.56 renamed `from_pretrained`'s `torch_dtype` argument to `dtype` —
[huggingface/transformers#39782](https://github.com/huggingface/transformers/pull/39782); on
older versions `dtype` is not consumed and leaks into the model constructor.)

This pins the supported floor in the example requirements so `pip install -r requirements.txt`
self-corrects. No library change — it only enforces the already-required minimum.

### Usage

```bash
cd examples/llm_ptq
pip install -r requirements.txt   # now ensures transformers>=4.56
```

### Testing

Verified on a GB200 node that the crash reproduces **only** below the floor and disappears at/above
it (loaded a model via `example_utils.get_model` / `from_pretrained` with ModelOpt save/restore):

| transformers | result |
|---|---|
| 4.53.3 | ❌ `TypeError: ...got an unexpected keyword argument 'dtype'` |
| 4.56.2 | ✅ loads |

Kimi-K2-Thinking int4 PTQ (the reported case) was validated on `transformers 4.57.1` in #669,
which is within this floor.

- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A (no new dependency; only adds a version floor to an existing transitive requirement)
- Did you get Claude approval on this PR?: ✅ (developed and verified with Claude Code)

### Additional Information

Originally opened as a `dtype`/`torch_dtype` compatibility shim. Per @kevalmorabia97's review, the
root cause is that the reporter manually ran `pip install transformers==4.53.3` (below ModelOpt's
`>=4.56` minimum) — the TRT-LLM container actually ships `transformers 4.57`, and Kimi-K2-Thinking
works on it (#669). Replaced the shim with this one-line floor pin so the example install can't
silently run below the supported `transformers`.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency version requirements for LLM example code, ensuring compatibility with the latest versions of required packages.

### DIFF
--- a/examples/llm_ptq/requirements.txt
+++ b/examples/llm_ptq/requirements.txt
@@ -1,5 +1,6 @@
 compressed-tensors
 fire
 flash-attn>=2.6.0
+transformers>=4.56
 transformers_stream_generator
 zstandard


kevalmorabia97 (changes requested):
ModelOpt requires transformers>=4.56 hence we should not need this fix. TRT-LLM container has 4.57 or 5.x so that should not be the rootcause of the issue. If a specific model requires transformers<4.56, we should drop its support

kevalmorabia97:
/claude review correct rootcause of this fix. TensorRT-LLM container infact has transformers 4.57 and not 4.53 as pointed in the PR description so that shouldnt be the rootcause

jingyu-ml:
Thanks @kevalmorabia97 — you're right, and after digging in:

- The TRT-LLM `1.3.0rc12` container ships `transformers 4.57`, and Kimi-K2-Thinking int4 PTQ was validated on `transformers 4.57.1` in #669. So the model does **not** require `<4.56`, and nothing needs to be dropped.
- The crash only reproduces on `transformers<4.56` (the `torch_dtype`→`dtype` rename landed in 4.56). The reporter's repro manually runs `pip install transformers==4.53.3`, dropping below ModelOpt's `>=4.56` minimum — that's the real cause, not the container.

I've dropped the dtype shim. Since `examples/llm_ptq/requirements.txt` didn't pin `transformers`, `pip install -r requirements.txt` wouldn't repair a manually-downgraded env. The PR now just pins `transformers>=4.56` there so the example install re-asserts the supported floor. PTAL.

jingyu-ml:
@kevalmorabia97 to add the verification behind the `transformers>=4.56` floor — we confirmed directly that 4.56 is where the bug clears.

Loaded a model through the example's `from_pretrained(..., dtype="auto")` path (ModelOpt save/restore enabled) on a GB200 node, across two transformers versions:

| transformers | `from_pretrained(dtype="auto")` |
| --- | --- |
| 4.53.3 | ❌ `TypeError: ...__init__() got an unexpected keyword argument 'dtype'` |
| 4.56.2 | ✅ loads |

So `>=4.56` resolves it. It was a minimal repro on a tiny model, but the failure is in the shared dtype-keyword path (transformers renamed `torch_dtype`→`dtype` in 4.56), so it's model-independent; the full Kimi-K2-Thinking int4 PTQ was separately validated on `transformers 4.57.1` in #669. That's why this PR just pins the supported floor in `examples/llm_ptq/requirements.txt`.

kevalmorabia97:
Closing PR: We shouldnt need to pin transformers>=4.56 in the example requirements since its the same in the main pyproject.toml dependencies also. QA Test should be run with transformers>=4.56

PR #1804, opened 2026-06-23
Add context-parallel (CP) and data-parallel (DP) support to Megatron calibration, and MMLU

kevalmorabia97:
### What does this PR do?

Type of change: new feature

Adds **context-parallel (CP)** and **data-parallel (DP)** support to the shared Megatron-Core inference/calibration utilities so PTQ calibration and the MMLU sanity check work across these parallelisms (in addition to the existing TP/PP/SP/EP).

**Context parallelism (CP):**
- **`megatron_calibration` / `megatron_mmlu`** — partition each sequence across CP ranks (zigzag load-balanced, via `get_batch_on_this_cp_rank`). MMLU gathers the per-rank logits back to the full sequence for last-token scoring.
- **`megatron_prefill`** — accepts a CP-partitioned `position_ids`, and under CP passes `attention_mask=None` so the CP-aware causal attention builds the mask itself (a local triu mask would be wrong for the per-rank zigzag chunks). Also wrapped in `torch.no_grad()` (pure inference; lets MMLU run at larger batch sizes without retaining the autograd graph).
- **`examples/megatron_bridge/quantize.py`** — new `--cp_size` flag.

**Data parallelism (DP):**
- **`get_dataset_dataloader`** — new `distributed` / `sampler_kwargs` to shard the dataset across ranks with a `DistributedSampler`.
- **`megatron_calibration`** — shards calibration data across the DP group (amax is max-reduced across DP inside `mtq` calibration, so the per-rank shards combine correctly).
- **`megatron_mmlu`** — shards whole batches across DP ranks and all-reduces the per-subject counts back to full-dataset accuracy.
- DP is **implicit**: `DP size = world_size / (tp * pp * cp)` — launching with more GPUs than `tp * pp * cp` engages it. (No `--dp_size` flag.)

RoPE is applied by the model per CP rank, so `position_ids` are only needed for models with absolute/learned position embeddings.

### Usage

```bash
# Context parallelism = 2
torchrun --nproc_per_node 2 examples/megatron_bridge/quantize.py \
    --hf_model_name_or_path Qwen/Qwen3-8B --quant_cfg nvfp4 \
    --cp_size 2 --export_megatron_path /tmp/Qwen3-8B-NVFP4-cp2

# Data parallelism = 2 (implicit: tp*pp*cp = 1, 2 GPUs)
torchrun --nproc_per_node 2 examples/megatron_bridge/quantize.py \
    --hf_model_name_or_path Qwen/Qwen3-8B --quant_cfg nvfp4 \
    --export_megatron_path /tmp/Qwen3-8B-NVFP4-dp2
```

### Testing

Added `cp` and `dp` cases to `tests/gpu_megatron/torch/utils/plugins/test_utils_megatron.py::test_megatron_generate_and_mmlu` (Qwen3-0.6B). All four parallelisms pass on 2 GPUs over the full 1430-example MMLU shard set (confirming the DP all-reduce reconstructs the full count): tp=0.373, pp=0.375, cp=0.371, dp=0.375.

End-to-end PTQ on **Qwen3-8B → NVFP4**, comparing MMLU before and after PTQ across parallelisms:

| Parallelism | MMLU before PTQ (bf16) | MMLU after PTQ (NVFP4) |
| :--- | :---: | :---: |
| TP=2 | 0.7294 | 0.7058 |
| CP=2 | 0.7292 | 0.7101 |
| DP=2 | 0.7292 | 0.7099 |

> **Common PTQ args used for the runs above:** `--hf_model_name_or_path Qwen/Qwen3-8B --quant_cfg nvfp4 --seq_length 1024 --calib_num_samples 512 --calib_batch_size 16`, default calibration dataset (`cnn_nemotron_v2_mix` = cnn_dailymail + nemotron-post-training-dataset-v2 mix). MMLU evaluated at `fraction=1.0, batch_size=16` on the full test set. Runs on 2× RTX 6000 Ada — TP=2 uses `--tp_size 2`, CP=2 uses `--cp_size 2`, DP=2 uses all model-parallel sizes = 1 (implicit DP over the 2 GPUs).

CP-, DP-, and TP-calibrated models all land within ~0.4% MMLU of each other both before and after PTQ, confirming CP/DP calibration yields an equivalently-quantized model.

- Is this change backward compatible?: ✅ — all CP/DP logic is gated on `cp_size > 1` / `dp_size > 1`; non-CP/DP (TP/PP/SP/EP) behavior is unchanged. `megatron_prefill` gains an optional `position_ids` arg (defaults to the previous behavior); `get_dataset_dataloader` gains optional `distributed`/`sampler_kwargs` (default off).
- Did you write any new necessary tests?: ✅ — `cp` and `dp` parametrizations added to the existing Megatron generate/MMLU test.
- Did you get Claude approval on this PR?: ✅ (run `/claude review`)

### Additional Information

`torch.no_grad()` on `megatron_prefill` is a shared change (also benefits the calibration / generate / PEFT-test callers) — pure inference, so no behavioral change beyond lower memory.

## Release Notes
* **New Features**
  * Added context-parallel (CP) and data-parallel (DP) support across shared inference, calibration, and MMLU evaluation.
  * Introduced CP-aware prefill with per-rank logits gathering for correct last-token scoring.
  * Added `--cp_size` to the quantization example (DP is derived automatically).
* **Improvements**
  * Extended dataset dataloader utilities with optional distributed sampling controls.
* **Bug Fixes**
  * Calibration and evaluation now work when CP is enabled (no longer restricted to CP=1).
* **Tests**
  * Expanded Megatron generate/MMLU coverage to include CP and DP modes.


kevalmorabia97:
Addressed the review feedback in 1319b3dfcf:

- **CodeRabbit (dataset_utils, `DistributedSampler` shuffle):** `get_dataloader_from_dataset` now passes its `shuffle` arg into `DistributedSampler` (which otherwise defaults to `True`); an explicit `sampler_kwargs["shuffle"]` still wins. The only caller (calibration) already passed `shuffle=False` explicitly, so behavior is unchanged — this just makes the helper consistent.
- **CodeRabbit + Claude (CP seq-length divisibility):** `cp_split_sequence` and `cp_gather_logits` now raise a clear `ValueError` when the sequence length isn't divisible by `2 * cp_size`, instead of failing later with a cryptic `.view()`/reshape error. (MMLU already rounds up; this guards the calibration path where the user picks `seq_length`.)

Claude review reported no CRITICAL/IMPORTANT issues. Re-ran the `cp` and `dp` GPU tests after the change — both pass.

modelopt/torch/utils/plugins/megatron_calibration.py
```diff
@@ -62,6 +62,11 @@ def get_megatron_calibration_forward_loop(
         tokenizer = copy.deepcopy(tokenizer)
         tokenizer.pad_token = tokenizer.eos_token
 
+    # Shard calibration data across data-parallel ranks (``with_context_parallel=False`` so CP
```
jenchen13: nit: make this comment shorter

modelopt/torch/utils/plugins/megatron_calibration.py
```diff
@@ -71,20 +76,25 @@ def get_megatron_calibration_forward_loop(
...
-        if cp_size != 1:
-            raise RuntimeError(
-                f"get_megatron_calibration_forward_loop requires CP=1, got "
-                f"context_parallel_world_size={cp_size}. Run calibration without CP."
-            )
+        cp_group = mpu.get_context_parallel_group()
         for sample in tqdm(dataloader, disable=not dist.is_master()):
-            megatron_prefill(model, sample["input_ids"], skip_return_logits=True)
+            if cp_size > 1:
+                # Split each sequence across CP ranks (requires seq length divisible by 2 * cp_size).
```
jenchen13: this comment is unnecessary, the divisible check happens in the function itself

modelopt/torch/utils/plugins/megatron_generate.py
```diff
@@ -25,11 +27,66 @@
...
+    cp_size = torch.distributed.get_world_size(cp_group)
+    if global_seq_len % (2 * cp_size) != 0:
+        raise ValueError(
+            f"global_seq_len (= {global_seq_len}) must be divisible by 2 * cp_size (= {2 * cp_size})."
+        )
+    chunk = global_seq_len // (2 * cp_size)
+    gathered = [torch.empty_like(local_logits) for _ in range(cp_size)]
+    torch.distributed.all_gather(gathered, local_logits.contiguous(), group=cp_group)
+    chunks: list[torch.Tensor | None] = [None] * (2 * cp_size)
+    for r in range(cp_size):
```
jenchen13, line 85: why do you still need a for loop to loop to each chunk after the all_gather? can it reconstruct the full logits using just collectives?

examples/megatron_bridge/quantize.py
```diff
@@ -94,9 +94,11 @@ def get_args() -> argparse.Namespace:
     )
 
     # Parallelism arguments
+    # Data parallelism is implicit: Launching with more GPUs than tp * pp * cp uses DP
     parser.add_argument("--tp_size", type=int, default=1, help="Tensor parallel size")
     parser.add_argument("--pp_size", type=int, default=1, help="Pipeline parallel size")
     parser.add_argument("--ep_size", type=int, default=1, help="Expert parallel size")
+    parser.add_argument("--cp_size", type=int, default=1, help="Context parallel size")
```
jenchen13, line 101: how do I enable DP during calibration? e.g. when is distribute=True in `get_dataset_dataloader`?

kevalmorabia97: DP is implicit. IF you set `torchrun --nproc_per_node 8 quantize.py --tp_size 2` means DP=4

modelopt/torch/utils/plugins/megatron_generate.py
kevalmorabia97: `all_gather` concatenates the ranks' local tensors in **rank order**, which under the zigzag (load-balanced) CP layout is `[chunk_0, chunk_{2cp-1}, chunk_1, chunk_{2cp-2}, ...]` — not sequence order. No single NCCL collective un-permutes that interleaving, so a reorder back to `[chunk_0 ... chunk_{2cp-1}]` is unavoidable. The loop is not extra comms/compute though — it is `2*cp_size` view-slices of the already-gathered tensor plus one `cat`. (Could express the same as an `index_select`, but it is the same op; the loop reads clearer.)

jenchen13 (approved):
LGTM

PR #1805, opened 2026-06-23
Account for CE loss for MTP heads in Megatron KD

AAnoosheh:
### What does this PR do?

Type of change: New Feature

Previously, MTP head loss was not accounted for in Megatron KD

### Usage

```python
```

### Testing
New gpu test

## Release Notes

* **Bug Fixes**
  * Improved handling of `skip_lm_loss` for Megatron distillation when multiple token prediction (MTP) heads are present, ensuring intermediate MTP head language modeling losses remain non-zero while only the main LM-head loss is zeroed per forward pass.

* **Tests**
  * Added GPU-distributed coverage validating `skip_lm_loss=True` with MTP configurations, including loss-call counting and assertions for non-zero MTP losses and a zeroed final LM-head loss.

### DIFF
--- a/modelopt/torch/distill/plugins/megatron.py
+++ b/modelopt/torch/distill/plugins/megatron.py
@@ -568,8 +568,14 @@ def adjust_distillation_model_for_mcore(
     model.sharded_state_dict = MethodType(_sharded_state_dict, model)
 
     # Skip `lm_loss` bypassing it when training if not needed for backprop.
+    # Uses a per-forward call counter so that MTP head calls (which always precede the
+    # main LM head call in _postprocess) still receive real CE loss even when
+    # skip_lm_loss=True — only the final main-head call is zeroed.
     def _compute_student_lm_loss(self, labels, logits) -> Tensor:
-        if distill_cfg.skip_lm_loss and self.training:
+        self._lm_loss_call_count += 1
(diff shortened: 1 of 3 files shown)


yeyu-nvidia (approved):
Reviewed — LGTM. Clean, minimal fix for the MTP-loss-gets-zeroed-by-skip_lm_loss problem, and the GPU test covering the call ordering + which heads are zeroed is exactly the right coverage. Verified it's backward-compatible: non-MTP models hit `mtp_num_layers=0` so `is_mtp_call` is always False (identical to prior behavior), and `skip_lm_loss=False` is untouched. The per-forward reset in `_forward` is correctly placed after the teacher pass.

One non-blocking note: the logic relies on the positional assumption that the first `mtp_num_layers` calls to `compute_language_model_loss` are the MTP heads and the last is the main head. That holds in current MCore `_postprocess` (and the test pins it), but it's an implicit coupling to call order — if MCore ever reorders, MTP loss would be silently zeroed instead of erroring. A small guard or a comment at the call site noting this dependency would make it more robust. Not a blocker.

This unblocks co-training MTP during NVFP4 QAD (tracked in OMNIML-5297) — thanks for picking it up.

ChenhanYu (approved):

PR #1809, opened 2026-06-23
Add NVFP4 Conv3d export for diffusers VAE (Wan 2.2)

jingyu-ml:
### What does this PR do?

Adds narrowly scoped NVFP4 Conv3d export support for the Wan VAE `WanCausalConv3d` stack in unified Hugging Face deployment checkpoints.

Previously, quantized Wan Conv3d modules had no unified-export handler, so their weights were not serialized in the packed NVFP4 schema. The converted quantization config also hardcoded `targets: ["Linear"]`, even for a Conv3d-only component.

This PR:

- registers a `WanCausalConv3d`-specific export handler through the existing export registry;
- flattens `[O, C, kt, kh, kw]` to logical `[O, K]`, pads `K` to block size 16, and writes packed `weight`, `weight_scale`, scalar `weight_scale_2`, and calibrated `input_scale` tensors;
- derives the NVFP4 config targets from the component's quantized module types, producing `targets: ["Conv3d"]` for a Conv3d-only VAE while preserving the historical Linear default for existing callers;
- rejects GEMM-specific padding/swizzle options when Conv3d is present instead of adding layer-exclusion plumbing; and
- fails before mutating a layer when the Conv3d configuration is unsupported or its activation quantizer was not calibrated.

### Scope

The supported contract is deliberately narrow: ungrouped Wan `WanCausalConv3d`, full dynamic block-16 NVFP4 weight/input quantization, and no output quantization. Grouped Conv3d, ConvTranspose, Conv1d/2d, static NVFP4, and W4A16 Conv3d fail closed or remain outside this handler.

This API produces a unified HF **deployment checkpoint** for downstream inference frameworks. Native Diffusers `from_pretrained()` loading uses a different `NVIDIAModelOptConfig`/ModelOpt-state contract and is out of scope.

Calibration must exercise every Conv3d selected for export. Workflows that only exercise VAE decode should quantize/export the decoder and leave uncalibrated encoder layers unquantized.

### Validation

- `pre-commit run --from-ref origin/main --to-ref HEAD`
- 50 focused and adjacent export tests covering registry routing, byte-exact packing, scale preservation, target generation, schema, quantizer-state removal, unsupported configurations, missing calibration, and padding/swizzle rejection
- A real Wan 2.2 5B VAE checkpoint audit found 48 packed Conv3d layers and exposed 20 uncalibrated encoder activation quantizers; this PR now rejects that inconsistent full-VAE export instead of declaring static activations without `input_scale`

### Before your PR is ready for review

- Backward compatible: yes; the config-conversion argument is optional and defaults to the existing Linear target.
- New dependencies: none.
- Tests: added focused CPU coverage and ran adjacent export regressions.
- Changelog: updated.

## Summary by CodeRabbit

* **New Features**
  * Added NVFP4 Hugging Face deployment-checkpoint export support for Wan VAE `Conv3d` layers.
  * Exported checkpoints now include flattened, packed weights, calibrated quantization scales, and correct `Conv3d` targets.
  * Added validation for unsupported padding and swizzle configurations, with actionable export errors.
  * Preserved existing support for NVFP4 linear layers while adapting quantization metadata to detected module types.

* **Documentation**
  * Updated the changelog with the new Wan VAE export capability.

### DIFF
--- a/modelopt/torch/export/unified_export_hf.py
+++ b/modelopt/torch/export/unified_export_hf.py
@@ -827,6 +827,19 @@ def _process_quantized_modules(
             handler(name, sub_module, ctx)
 
 
+def _diffusion_quantization_target_types(model: nn.Module) -> list[str]:
+    """Return stable type targets for homogeneous NVFP4 diffusion modules."""
+    target_types = set()
+    for sub_module in model.modules():
+        if get_quantization_format(sub_module) != QUANTIZATION_NVFP4:
+            continue
+        if any(cls.__name__ == "WanCausalConv3d" for cls in type(sub_module).__mro__):
+            target_types.add("Conv3d")
+        elif is_quantlinear(sub_module):
+            target_types.add("Linear")
+    return sorted(target_types)
+
+
 def _export_transformers_checkpoint(
     model: nn.Module,
     dtype: torch.dtype | None = None,
@@ -1165,6 +1178,16 @@ def _export_diffusers_checkpoint(
             is_qwen_component = "qwen" in type(component).__name__.lower()
             _fuse_qkv_linears_diffusion(component, strict=is_qwen_component)
 
+            target_types = _diffusion_quantization_target_types(component)
+            if "Conv3d" in target_types and (
+                kwargs.get("padding_strategy") is not None
+                or kwargs.get("enable_swizzle_layout", False)
+            ):
+                raise NotImplementedError(
+                    "NVFP4 GEMM padding/swizzling is not defined for Conv3d exports. "
+                    "Export without padding_strategy/enable_swizzle_layout."
+                )
+
             # Process quantized modules (convert weights, register scales)
             _process_quantized_modules(component, component_dtype, is_modelopt_qlora=False)
 
@@ -1187,7 +1210,9 @@ def _export_diffusers_checkpoint(
                         if svdquant_rank is not None:
                             quantization_details["lora_rank"] = svdquant_rank
                 hf_quant_config = (
-                    convert_hf_quant_config_format(quant_config) if quant_config else None
+                    convert_hf_quant_config_format(quant_config, target_types=target_types)
+                    if quant_config
+                    else None
                 )
 
                 # Save the component
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Changelog
 
 **New Features**
 
+- Add NVFP4 unified Hugging Face deployment-checkpoint export for Wan VAE ``WanCausalConv3d`` layers, including flattened-K packed weights, calibrated activation/weight scales, and correct ``Conv3d`` config targets.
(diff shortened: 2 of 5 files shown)


cjluo-nv (commented):

Adds NVFP4 Conv3d weight export for the diffusers Wan 2.2 VAE to the unified HF export path. +636/-18, 8 files. I traced the core logic and it holds up: the conv flatten `weight.reshape(O, -1)` matches the calibration-time flatten in `_nvfp4_quantize_weight_along_k` (so byte-exactness is plausible and is pinned by a real test); the pad-before-scale ordering is correct (`get_weights_scaling_factor` requires block-divisible last dim); the static-quantizer rejection via `NVFP4QTensor._is_static_quantizer` is sound and tested; `is_quantconv3d` correctly matches `_QuantDiffusersWanCausalConv3d`/`QuantConv3d` while excluding Conv1d/Conv2d (no "Conv3d" substring) and ConvTranspose; and the `nvfp4_exclude_layers` plumbing correctly keeps conv prefixes out of the Linear-targeted pad/swizzle. New optional kwargs are backward-compatible. The lazy ONNX import in quantize.py has a valid documented justification (heavy optional dependency). Test coverage is strong (byte-exact vs `NVFP4QTensor.quantize`, dequant round-trip, schema, K-order sensitivity, dispatch routing, quantizer-hiding, pad/swizzle exclusion, plus a tiny real Wan VAE e2e). New test files' license headers match the canonical LICENSE_HEADER (standard-header exception applies — no licensing concern). No prompt-injection in PR text.

- `hide_quantizers_from_state_dict` was generalized from "strip QuantLinear's weight/input/output quantizers" to "strip every `*_quantizer` child of every module." This now also strips attention bmm/softmax quantizers across ALL diffusers exports, not just conv. My read is this is a cleanup (the unified HF safetensors path never converted attention quantizers into usable scale buffers — FP8 MHA scales go through the ONNX `export_fp8_mha` symbolic path, so the previously-serialized `*_quantizer._amax` keys were unusable junk), but it changes a shared code path with broad blast radius (Flux/SD3/Wan/LTX-2) and only has manual GPU e2e validation.
- The real numeric conv export was only manually validated on a Wan 2.2 5B VAE (48 layers); CI coverage is CPU-only.
- Minor: some scale-derivation/input_scale duplication between `_export_quantized_conv_weight` and `_export_quantized_weight`, acceptable given the conv-specific flatten/static-reject.

Recommend a diffusers-export owner confirm the `hide_quantizers_from_state_dict` broadening is safe for the FP8-MHA/Linear paths and that the GPU conv export numerics are signed off.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -151,6 +152,7 @@ def _postprocess_safetensors(
     export_dir: Path,
     pipe: Any | None = None,
     hf_quant_config: dict | None = None,
+    nvfp4_exclude_layers: set[str] | None = None,
```
cjluo-nv: what are the layers that we would like to exclude? can we infer by naming instead of introducing a new arg?

PR #1810, opened 2026-06-23
Fix AutoQuantize causal LM score scaling

realAsma:
## Summary
- use a sum-reduced shifted-token causal LM loss for HF AutoQuantize gradient scoring when logits and labels are available
- preserve the existing `output.loss` fallback for non-causal or non-dict outputs
- apply the same correction to the HF PTQ, eval, and autodeploy AutoQuantize entrypoints

## Motivation
A Qwen3 8B bs=1 vs bs=8 AutoQuantize run over the same 128 scoring samples showed nearly identical normalized layer sensitivity shapes but very different absolute totals:

- NVFP4/W4A4 raw score total: bs1 `0.5883269825`, bs8 `0.008391587796`, bs1/bs8 `70.109x`
- FP8 raw score total: bs1 `0.05273269751`, bs8 `0.0007778984065`, bs1/bs8 `67.789x`
- Layer Pearson correlation after per-run normalization: NVFP4/W4A4 `0.9619`, FP8 `0.9864`

The existing HF callbacks used Transformers `output.loss`, which is mean-reduced for causal LM labels. AutoQuantize squares gradients and accumulates scores, so grouping samples into larger batches changes the absolute score scale. A sum-reduced next-token loss makes the score additive over valid labels and stable across batch grouping.

## Validation
- `python_pwd -m py_compile examples/llm_ptq/hf_ptq.py examples/llm_eval/quantization_utils.py examples/llm_autodeploy/run_auto_quantize.py`
- targeted CPU probe extracted all three `_causal_lm_sum_loss` helpers and verified the summed loss is differentiable and identical for one combined batch vs split single-sample batches, including `-100` ignored labels
- `git diff --check`
- `pre-commit run --files examples/llm_ptq/hf_ptq.py examples/llm_eval/quantization_utils.py examples/llm_autodeploy/run_auto_quantize.py` passes ruff/format/bandit/license checks; `mypy` currently fails on existing `accelerate` attribute errors in `modelopt/torch/quantization/plugins/{accelerate,huggingface}.py` outside this diff

### DIFF
--- a/examples/llm_ptq/hf_ptq.py
+++ b/examples/llm_ptq/hf_ptq.py
@@ -23,7 +23,6 @@ from typing import Any
 
 import numpy as np
 import torch
-from accelerate.hooks import remove_hook_from_module
 from cast_mxfp4_to_nvfp4 import apply_to_model as apply_cast_mxfp4_to_nvfp4
 from cast_mxfp4_to_nvfp4 import force_weight_quantizers_static
 from example_utils import (
@@ -55,6 +54,7 @@ from transformers import (
 import modelopt.torch.opt as mto
 import modelopt.torch.quantization as mtq
 import modelopt.torch.sparsity as mts
+from accelerate.hooks import remove_hook_from_module
 from modelopt.recipe import ModelOptPTQRecipe, load_recipe
 from modelopt.torch.export import (
     export_hf_checkpoint,
@@ -288,6 +288,17 @@ def make_calib_dataloader(
     return calib_dataloader, first_text_speech_dataset
 
 
+def _causal_lm_sum_loss(logits: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
+    """Sum-reduced next-token loss keeps AutoQuantize scores additive across batches."""
+    shift_logits = logits[..., :-1, :].contiguous()
+    shift_labels = labels[..., 1:].contiguous()
+    return torch.nn.functional.cross_entropy(
+        shift_logits.view(-1, shift_logits.size(-1)),
+        shift_labels.view(-1),
+        reduction="sum",
+    )
+
+
 def auto_quantize(
     args: argparse.Namespace,
     language_model: torch.nn.Module,
@@ -349,16 +360,15 @@ def auto_quantize(
 
         def loss_func(output, data):
             logits = lm_head(output.last_hidden_state)
-            labels = data["labels"]
-            shift_logits = logits[..., :-1, :].contiguous()
-            shift_labels = labels[..., 1:].contiguous()
-            return torch.nn.functional.cross_entropy(
-                shift_logits.view(-1, shift_logits.size(-1)), shift_labels.view(-1)
-            )
+            return _causal_lm_sum_loss(logits, data["labels"])
 
     else:
 
         def loss_func(output, data):
+            logits = getattr(output, "logits", None)
+            labels = data.get("labels") if isinstance(data, dict) else None
+            if logits is not None and labels is not None:
+                return _causal_lm_sum_loss(logits, labels)
             return output.loss
 
     if auto_quantize_method == "gradient":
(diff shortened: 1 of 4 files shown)


realAsma (commented):
Can you add a unit test for tinyqwen tranformers model to make sure that after your fix bs = 1 and say bs = 2 produce similar scores?

Create a 2 sample input calibration dataset for this; sample 1 length: 16, sample2 length: 32 (so basically = 2 means padding is activated).

examples/llm_ptq/hf_ptq.py
```diff
@@ -310,6 +288,17 @@ def make_calib_dataloader(
...
+    """Sum-reduced next-token loss keeps AutoQuantize scores additive across batches."""
+    shift_logits = logits[..., :-1, :].contiguous()
+    shift_labels = labels[..., 1:].contiguous()
+    return torch.nn.functional.cross_entropy(
+        shift_logits.view(-1, shift_logits.size(-1)),
+        shift_labels.view(-1),
+        reduction="sum",
+    )
+
+
```
realAsma, line 301: Is there a huggingface side argument or something so that HF transformers model compute CE loss with reduction = sum?

realAsma:
> _:robot: Bot comment._

Addressed the TinyQwen/Tiny Qwen3 regression test request in `71257f9ffa`.

What changed:
- Added `test_autoquantize_huggingface_scores_are_batch_size_invariant_with_padding` in `tests/unit/torch/quantization/plugins/test_huggingface.py`.
- The test uses two calibration samples with lengths 16 and 32.
- It compares the same samples as bs=1 singleton batches versus one bs=2 padded batch and asserts AutoQuantize candidate scores match within tight tolerance.

Validation run locally:
- `git diff --check`
- `python_pwd PYTHONPATH=tests -m py_compile tests/unit/torch/quantization/plugins/test_huggingface.py`
- `ruff format --check tests/unit/torch/quantization/plugins/test_huggingface.py`
- `ruff check tests/unit/torch/quantization/plugins/test_huggingface.py`
- `pytest_pwd PYTHONPATH=tests tests/unit/torch/quantization/plugins/test_huggingface.py::test_autoquantize_huggingface_scores_are_batch_size_invariant_with_padding -q`

examples/llm_ptq/hf_ptq.py
realAsma: > _:robot: Bot comment._

I checked the local Transformers 5.5.3 implementation. There is not a normal HF model-forward argument that returns a pure sum-reduced causal-LM loss.

`ForCausalLMLoss` accepts `num_items_in_batch`; internally `fixed_cross_entropy` uses `reduction="sum"` only when that is set, then divides by `num_items_in_batch`. With the usual `labels=` forward path it is mean-reduced. Since AutoQuantize needs raw additive scores across batch groupings, this PR keeps the explicit shifted-token CE with `reduction="sum"` when logits and labels are available.

realAsma: Can we pass `num_items_in_batch` as part of the data batch?

realAsma: > _:robot: Bot comment._

Done in `2a2163e7c2`.

For the normal HF CausalLM path, the forward step now passes scalar `num_items_in_batch=1` through the batch and `loss_func` uses Transformers' `output.loss`, so the HF loss path gives the raw sum loss needed by AutoQuantize scoring. I kept the explicit sum-loss fallback for paths that cannot rely on HF-computed loss, especially the extracted base-model + outer `lm_head` path.

Validation passed:
- `git diff --check`
- `python_pwd -m py_compile examples/llm_ptq/hf_ptq.py examples/llm_autodeploy/run_auto_quantize.py examples/llm_eval/quantization_utils.py tests/unit/torch/quantization/plugins/test_huggingface.py`
- `pytest_pwd tests/unit/torch/quantization/plugins/test_huggingface.py -k "num_items_in_batch or batch_size_invariant"`
- `pytest_pwd tests/unit/torch/quantization/plugins/test_huggingface.py`
- commit hooks on the final staged diff

PR #1816, opened 2026-06-24
Emit VisualGen-compatible sparse_attention_config for diffusion skip-softmax export

jingyu-ml:
### What does this PR do?

**Type of change:** Bug fix (+ small backward-incompatible config-field rename)

TensorRT-LLM's **VisualGen** runtime consumes a *single-phase* skip-softmax checkpoint schema that differs from the LLM (prefill/decode) schema `export_sparse_attention_config` emits for every model. As a result a diffusion checkpoint (e.g. WAN 2.2 T2V) calibrated and exported by ModelOpt today is **not loadable** by VisualGen:

- coefficients are nested under `threshold_scale_factor.prefill` instead of `threshold_scale_factor.coefficients` — VisualGen's formula parser reads `.coefficients`, and its flat fallback then raises `TypeError` on the nested dict;
- `target_sparsity` is a `{phase: ratio}` dict, which the VisualGen config parser explicitly rejects (it requires a scalar); and
- the dense-window knob is the integer `initial_disabled_steps`, which no longer exists in TensorRT-LLM — it was renamed to the normalized `[0, 1]` `disabled_until_timestep`.

This PR makes `export_sparse_attention_config` diffusion-aware:

- new `for_diffusion` argument (set by the diffusers export path in `unified_export_hf`) emits the VisualGen schema: `threshold_scale_factor.coefficients`, a scalar `target_sparsity`, and `disabled_until_timestep`;
- the LLM (transformers) export path is **unchanged** (byte-for-byte);
- the passthrough config field and the WAN 2.2 example CLI flag are renamed `initial_disabled_steps` → `disabled_until_timestep` (normalized `[0, 1]`, validated) to match the deployed vocabulary.

### Usage

```python

# Diffusion pipelines automatically get the VisualGen schema (diffusers path -> for_diffusion=True):
export_hf_checkpoint(pipe, export_dir="Wan2.2-T2V-A14B-Diffusers-Skip-Softmax")
```

The exported `<component>/config.json` then contains:

```json
"sparse_attention_config": {
  "config_groups": {
    "group_0": {
      "algorithm": "skip_softmax",
      "threshold_scale_factor": {
        "formula": "a * exp(b * target_sparsity)",
        "coefficients": {"a": 2142.73, "b": 4.28}
      },
      "target_sparsity": 0.5,
      "ignore": ["blocks.0.attn1", "blocks.0.attn2"]
    }
  }
}
```

### Testing

- New unit test `test_exports_diffusion_visual_gen_format` asserts the single-phase schema and that the default (LLM) schema is unchanged: `pytest tests/unit/torch/sparsity/attention_sparsity/test_sparse_attention_conversion.py::TestExportSparseAttentionConfig` → **5 passed**.
- Updated the diffusers example integration test (`tests/examples/diffusers/sparsity/test_sparsity.py`) to assert the new schema.
- Emitted format validated field-for-field against TensorRT-LLM `main` (`tensorrt_llm/visual_gen/sparse_attention.py::SkipSoftmaxAttentionConfig.resolve_threshold_scale_factor`, `tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py`) and the VisualGen `docs/source/visual-gen/features/sparse-attention.md` checkpoint-config example.

- Is this change backward compatible?: ❌ — renames the skip-softmax `initial_disabled_steps` config field to `disabled_until_timestep` (new field, normalized `[0,1]` units). The skip-softmax attention-sparsity export is new and not yet broadly adopted, so this aligns the metadata with the consuming runtime before it ships widely.
- Did you get Claude approval on this PR?: N/A (maintainer to run `/claude review`)

### Additional Information

The VisualGen consumer is `SkipSoftmaxAttentionConfig.resolve_threshold_scale_factor`; the LLM consumer (`tensorrt_llm/llmapi/llm_args.py`) keeps the phase-keyed schema, which is why the LLM export path is intentionally left unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added diffusion-compatible sparse-attention export that targets the TRT-LLM VisualGen diffusion-transformer checkpoint schema (including new calibration/export fields and single-phase coefficient layout for skip-softmax).
  * Replaced skip-softmax “initial disabled steps” metadata with a normalized `disabled_until_timestep` cutoff.
* **Documentation**
  * Updated the changelog and example CLI/options to reflect the new `disabled-until-timestep` export metadata.
* **Tests**
  * Updated Wan 2.2 calibrated export assertions and coefficient expectations.
  * Added unit coverage for the diffusion (VisualGen) export schema.


cjluo-nv (commented):

Focused, well-tested change (+114/-37, 7 files). Adds a `for_diffusion` kwarg to the existing `export_sparse_attention_config` to emit the TRT-LLM VisualGen single-phase skip-softmax schema (coefficients under `threshold_scale_factor.coefficients`, scalar `target_sparsity`, normalized `disabled_until_timestep`), and renames the passthrough config field `initial_disabled_steps` → `disabled_until_timestep` with a `[0,1]` validator.

Design review (gate fired on 5-directory span): this is NOT an architectural change — it extends an existing function and reuses existing config infrastructure rather than introducing a new subsystem/loader/DSL. No new design alternative needs justification.

Correctness looks good: the diffusion branch picks prefill (or first) coefficients and the prefill sparsity scalar; the LLM (`for_diffusion=False`) path is left byte-for-byte unchanged in a separate branch, and a unit test asserts the LLM `{phase: ratio}` schema is preserved. Validation, harvesting pattern, and example CLI rename are all consistent. Tests cover the new schema, the unchanged LLM schema, and the example integration. No licensing changes; no injection in the PR text.

Why nudge rather than approve: the field rename is an explicitly backward-incompatible change with a semantics shift (integer step count → normalized [0,1] cutoff). Old configs/checkpoints using `initial_disabled_steps` won't error — the field is silently ignored and simply not emitted. The author justifies this as acceptable because the skip-softmax export is new and not broadly adopted; that's a reasonable but product-level call that warrants a maintainer's explicit sign-off. Also, the new schema was validated against TRT-LLM `main` by hand (no automated cross-repo check), so a human with VisualGen context should confirm the field names/units match the deployed consumer.

kaix-nv (approved):

modelopt/torch/sparsity/attention_sparsity/conversion.py
```diff
@@ -465,22 +475,44 @@ def export_sparse_attention_config(model: nn.Module) -> dict[str, Any] | None:
...
+                    "a": phase_coeffs["a"],
+                    "b": phase_coeffs["b"],
                 }
-        skip_group["threshold_scale_factor"] = threshold_scale_factor
-        if target_sparse_ratio is not None:
-            skip_group["target_sparsity"] = target_sparse_ratio
+            skip_group["threshold_scale_factor"] = threshold_scale_factor
+            if target_sparse_ratio is not None:
+                scalar_sparsity = (
+                    target_sparse_ratio.get("prefill")
```
Edwardf0t1, line 498: Asymmetric phase fallback between `coefficients` and `target_sparsity`. `coefficients` above falls back to `next(iter(calibration_params.values()))` when `"prefill"` is absent, but `target_sparsity` here reads only `target_sparse_ratio.get("prefill")` with no fallback.

If a diffusion checkpoint is ever calibrated with a non-prefill phase (the decode path in `calibration/calibrate.py` supports this), the export would emit `coefficients` (from the fallback phase) but silently drop `target_sparsity` — producing a VisualGen config missing the scalar sparsity it requires, i.e. exactly the parse failure this PR aims to prevent. The WAN example hardcodes `{"prefill": ...}` so it can't trigger today, but the two lookups should share the same phase resolution to stay robust — e.g. resolve the phase key once and use it for both.

modelopt/torch/sparsity/attention_sparsity/conversion.py
```diff
@@ -465,22 +475,44 @@ def export_sparse_attention_config(model: nn.Module) -> dict[str, Any] | None:
...
+        # expects per-phase (prefill/decode) blocks plus a {phase: ratio} dict.
         threshold_scale_factor: dict[str, Any] = {
             "formula": "a * exp(b * target_sparsity)",
         }
-        for phase in ["prefill", "decode"]:
-            if phase in calibration_params:
-                threshold_scale_factor[phase] = {
-                    "a": calibration_params[phase]["a"],
-                    "b": calibration_params[phase]["b"],
+        if for_diffusion:
```
Edwardf0t1, line 486: Optional cleanup: the `if for_diffusion / else` branches duplicate `skip_group["threshold_scale_factor"] = threshold_scale_factor` and the `target_sparsity` gating — only the *shape* differs (single vs per-phase coeffs; scalar vs dict sparsity). Computing the two shape-specific values first and assigning them once after the branch would remove the copy-paste and reduce the chance the two paths drift on future edits. Not a correctness issue.

mxinO (approved):
LGTM! Thanks.

rohansjoshi (approved):

jingyu-ml:
/ok to test [88b8e44](https://github.com/NVIDIA/Model-Optimizer/pull/1816/commits/88b8e44f6427ed527a365944e041979290047c23)

modelopt/torch/sparsity/attention_sparsity/conversion.py
jingyu-ml: Good catch on the asymmetry — the two lookups do differ. Leaving it as-is though, because the divergent case can't be reached by any supported calibration path:

- Decode-phase calibration is LLM-only. The decode calibrator needs the RULER dataset + tokenizer (`create_decode_calibration_forward_loop` drives tokenizer-based KV-cache generation), and those are only built when no custom `forward_loop` is passed. Diffusion models are always calibrated with a custom pipeline `forward_loop` (as in the WAN example), so enabling decode there raises `RuntimeError("calibration_data and tokenizer must be built before decode")` in `calibration/calibrate.py` before any params are produced. A diffusion checkpoint whose `calibration_params` lacks `"prefill"` can't be created.
- `target_sparse_ratio` on this path is always prefill-keyed too — the `CalibConfig` default is `{"prefill": 0.5, "decode": 0.5}` and the WAN example passes `{"prefill": ...}`.

So under `for_diffusion=True` both lookups always resolve the same `"prefill"` entry; the `next(iter(...))` on coefficients is belt-and-braces rather than a fallback that can diverge from `target_sparsity` in practice. This branch deliberately mirrors, field-for-field, what VisualGen's `SkipSoftmaxAttentionConfig` on TRT-LLM `main` parses today (single `coefficients` block + scalar `target_sparsity`), so I'd rather not add speculative phase resolution the consumer can't use. If diffusion ever grows a second calibration phase, the VisualGen schema has to change first anyway — I'll unify the phase-key resolution as part of that alignment.

modelopt/torch/sparsity/attention_sparsity/conversion.py
jingyu-ml: Considered it, but keeping the two branches flat is deliberate: each one assembles a single consumer's schema (TRT-LLM VisualGen vs. the LLM runtime) in one linear block that can be checked line-by-line against the runtime parser — the VisualGen format was just pinned down on the TRT-LLM side and this PR mirrors it verbatim. Since the two schemas serve different runtimes and are expected to evolve independently (that's why the `for_diffusion` switch exists), hoisting the shared assignments would couple them to save ~2 lines. Happy to revisit if a third consumer format shows up.

jingyu-ml:
/ok to test [fb06815](https://github.com/NVIDIA/Model-Optimizer/pull/1816/commits/fb068159d240c945602ebfc36c1d7a49686c0020)

jingyu-ml:
/ok to test [c94f628](https://github.com/NVIDIA/Model-Optimizer/pull/1816/commits/c94f62865a08def7b4613029fe442de50e935b1d)

cjluo-nv (commented):

Focused change (+114/-37, 7 files): adds a `for_diffusion` kwarg to the existing `export_sparse_attention_config` to emit the TRT-LLM VisualGen single-phase skip-softmax schema (coefficients under `threshold_scale_factor.coefficients`, scalar `target_sparsity`, normalized `disabled_until_timestep`), and renames the passthrough config field `initial_disabled_steps` → `disabled_until_timestep` with a `[0,1]` validator.

Design review (gate fired on 5-directory span): NOT an architectural change — it extends an existing function and reuses the existing Pydantic config infrastructure (`SparseAttentionAttributeConfig`, `ModeloptBaseConfig`) rather than introducing a new subsystem/loader/DSL. No design alternative needs justifying. Correctness is sound: the diffusion branch resolves prefill (or first) coefficients + the prefill scalar sparsity; the LLM (`for_diffusion=False`) path is left byte-for-byte in a separate branch. Tests cover the new schema, the unchanged LLM `{phase: ratio}` schema, and the example integration. No injection in the PR text; no licensing changes (2024 headers pre-exist, untouched).

Re-review of prior comments:
- :speech_balloon: Edwardf0t1 (conversion.py:498, asymmetric phase fallback — correctness): Author replied that the divergent case is unreachable — diffusion always calibrates with a custom pipeline `forward_loop`, so decode-phase calibration (LLM-only, needs RULER+tokenizer) can't run, and `target_sparse_ratio` on this path is always prefill-keyed. Rationale is reasonable; the WAN example hardcodes `{"prefill": ...}`. Considered resolved.
- Edwardf0t1 (conversion.py:486, dedup the branches — minor/optional): Author declined with a reasonable "keep each consumer schema flat and line-checkable" rationale. Fine to leave.

- :speech_balloon: CodeRabbit flagged (config.py:158-185, Major) that the `initial_disabled_steps` → `disabled_until_timestep` rename breaks backward compat — `ModeloptBaseConfig` uses `extra="forbid"`, so restoring old ModelOpt state whose `sparse_cfg` still carries `initial_disabled_steps` will raise a Pydantic validation error inside `set_from_attribute_config` during `restore`. No alias/migration was added. Author/PR body argue this is acceptable because the skip-softmax export is new and not broadly adopted — a deliberate product call, not a bug — but the repo's coding guideline ("preserve config/checkpoint backward compat for Pydantic configs") means a maintainer should explicitly sign off on shipping the incompatible rename (and the int-step → normalized-[0,1] semantics shift) rather than adding a deprecated alias. Note only checkpoints that set `initial_disabled_steps > 0` are affected (default 0 was never emitted).
- The emitted VisualGen schema (field names/units) was validated by hand against TRT-LLM `main`, with no automated cross-repo check — a human with VisualGen context should confirm the deployed consumer matches.

Edwardf0t1 (approved):
Approve to unblock

PR #1826, opened 2026-06-25
fix(recipes): exclude Llama-4 vision branch from default PTQ quantization

Edwardf0t1:
## What does this PR do?

**Type of change:** Bug fix

**Overview:** Adds `*vision_model*` and `*multi_modal_projector*` to the `default_disabled_quantizers` PTQ unit so the Llama-4 vision branch stays in BF16 by default.

## Details

The vision-exclusion patterns in `modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml` (`*embed_vision*` / `*vision_tower*` / `*visual*`) cover gemma / Qwen-VL / Kimi naming but miss Llama-4, whose encoder is named `vision_model` and whose projector is `multi_modal_projector`.

As a result, `general/ptq/nvfp4_default-kv_fp8` (and any recipe importing this unit) quantized the Llama-4-Scout vision tower. Export then crashed on `vision_model.patch_embedding.linear`, whose `in_features=588` (3×14×14) is not divisible by the NVFP4 block size:

```
AssertionError: Weight shape is not divisible for block size for block quantization.
Failed to export module 'vision_model.patch_embedding.linear' (type=QuantLinear)
```

Adding the two patterns keeps the Llama-4 vision branch in BF16 by default, matching the existing behavior for other VL models (gemma-4, Qwen3.5-VL, Kimi).

Fixes NVBugs 6359097.

## Testing

- YAML validates and pre-commit (incl. `validate modelopt recipes`) passes.
- The `-novit` recipes import this unit and inherit the fix automatically.

## Before your PR is "*Ready for review*"

- [x] Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md) and your commits are signed.
- [x] Is this change backward compatible? **Yes** — only expands the default disable list to additional vision-branch module names.
- [x] Did you write any new necessary tests? **N/A** — config-only change.
- [x] Did you add or update any necessary documentation? **Yes** — updated the inline rationale comment with the Llama-4 naming and NVBug reference.
- [x] Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)? **N/A** — recipe-config bug fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded the default list of vision-related components that stay unquantized, improving support for multimodal models.
  * Added broader exclusions for additional vision and projector patterns to better preserve model quality by default.

### DIFF
--- a/modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml
+++ b/modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml
@@ -43,16 +43,23 @@
   # Multimodal vision branch: keep the vision encoder (SigLIP / ViT) and any
   # multimodal embedding projection in BF16 by default. Recipes that enable bare
   # `*weight_quantizer` / `*input_quantizer` or `*mlp*` wildcards otherwise also
-  # match the vision tower (`model.vision_tower.*`, `model.visual.*`) and the
-  # embedding projection (`model.embed_vision.*`); quantizing the vision branch
-  # crashes export / produces garbage image embeddings on VL models (gemma-4,
-  # Qwen3.5-VL — NVBugs 6293731, 6293762, 6294017). A recipe that intentionally
+  # match the vision tower (`model.vision_tower.*`, `model.visual.*`,
+  # Llama-4 `vision_model.*`) and the embedding projection (`model.embed_vision.*`,
+  # Llama-4 `multi_modal_projector.*`); quantizing the vision branch crashes
+  # export / produces garbage image embeddings on VL models (gemma-4,
+  # Qwen3.5-VL — NVBugs 6293731, 6293762, 6294017; Llama-4-Scout — NVBugs
+  # 6359097, where `vision_model.patch_embedding.linear` has in_features=588,
+  # not divisible by the NVFP4 block size). A recipe that intentionally
   # quantizes vision must re-enable these after importing this unit.
   - quantizer_name: '*embed_vision*'
     enable: false
   - quantizer_name: '*vision_tower*'
     enable: false
   - quantizer_name: '*visual*'
+    enable: false
+  - quantizer_name: '*vision_model*'
+    enable: false
+  - quantizer_name: '*multi_modal_projector*'
     enable: false
   - parent_class: 'nn.BatchNorm1d'
     quantizer_name: '*'


kevalmorabia97 (approved):

cjluo-nv (approved):

Small, focused config-only bug fix. Adds `*vision_model*` and `*multi_modal_projector*` to the `default_disabled_quantizers` PTQ unit so Llama-4's vision encoder/projector stay in BF16 by default, matching existing behavior for gemma-4/Qwen-VL/Kimi (which use `*vision_tower*`/`*visual*`/`*embed_vision*`). The change is correct: these are the canonical HF module names for Llama-4's vision branch, the wildcards are appropriately scoped (won't collide with text-branch LLM modules), and it directly addresses the export crash on `vision_model.patch_embedding.linear` (in_features=588 not divisible by NVFP4 block size). Reuses the existing unit with no new abstraction; the inline rationale comment is updated with the Llama-4 naming and NVBug reference. No license-file or header changes. As with the pre-existing vision patterns, there's no per-pattern unit test, but pre-commit recipe validation covers the YAML and this matches the established maintenance pattern for this file. Untrusted PR fields contained no injection attempts.

modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml
```diff
@@ -43,17 +43,24 @@
...
   - quantizer_name: '*embed_vision*'
     enable: false
   - quantizer_name: '*vision_tower*'
     enable: false
   - quantizer_name: '*visual*'
     enable: false
+  - quantizer_name: '*vision_model*'
+    enable: false
+  - quantizer_name: '*multi_modal_projector*'
+    enable: false
```
juhi10071998, line 63: @shengliangxu - I was curious when do we add model specific architecture exclusions in the model specific YAMLs? I don't see us following a strict pattern, wondering if you have any recommendation/ thoughts on this.
For instance, for diffusionGemma I extended the default_disabled_quantizers to have it's own model specific exclusions

shengliangxu: Yeah, we should really not adding more and more model_type specific exclusions here. It will pollute all models. But not gonna block this PR for the sake that we are already doing it.

I think we need a design for a more systematic way for integrating model_type specific recipe components.

If you are interested, feel free to think about it. Also need to track it in Jira.

PR #1828, opened 2026-06-25
Modeling Lib: per-model architecture kick off with spec registration

h-guo18:
### What does this PR do?

**Type of change:** New feature — per-model infrastructure (with behavior changes, see below)

Starts `modelopt/torch/models/`: somewhere for ModelOpt to keep what it knows about a
model, keyed by HF model type (`config.model_type`), one package per type
(`<model_type>/specs.py`). Importing the package registers every spec; consumers call
`get_spec(model_type)` / `match_moe_block(module, model_type)` and read the fields.

The point is the infrastructure, not the migration. Today ModelOpt's per-model knowledge
is scattered across if/elif chains in whichever subsystem happened to need it, so the same
fact gets restated per subsystem and drifts. A model's MoE block classes, its expert
projection naming, which of its norms store `w - 1` — these are facts about the *model*,
and several subsystems want them. This PR gives them one home and one lookup.

It is a kick-off, so it is deliberately narrow: specs plus the first consumer. **Export is
that first consumer**, which is why most of the changed lines are on the export side —
not because this is an export refactor. Quantization, speculative decoding and the rest
keep their own tables for now; the per-model directory is where their sections land later,
and `specs.py` is just the first thing in it.

| Data the first consumer moved in | Spec field | Consumers |
|---|---|---|
| MoE block classes and expert linear naming | `MoESpec` | `get_expert_linear_names`, `get_experts_list`, `is_moe`, `sync_moe_gate_up_amax` |
| grouped expert-export support set | `ExportSpec.grouped_expert_export` | `get_experts_list` |
| AWQ `pre_quant_scale` fusion rules | `ExportSpec.pqs_fuse_rules` | `fuse_prequant_to_linear` |
| weight-plus-one norm class names | `ExportSpec.weight_plus_one_norm_names` | `_layernorm_uses_weight_plus_one` |

`ModelSpec` holds each concern as a separate, optional section, and the split is what
makes it extensible: **topic sections** hold architecture facts any subsystem can read
(`MoESpec` — block classes, expert projection naming, how the experts are stored),
**subsystem sections** hold one subsystem's own data and policy (`ExportSpec` — AWQ fusion
rules, weight-plus-one norms, whether grouped expert export is validated for this model).
A new subsystem adds a section rather than a table.

A section is `None` when the model has nothing to say about it, so a dense model carries
`moe_spec=None` rather than an empty one. Two further fields record where a model's
classes come from at all: `modeling_source` (`transformers` or `remote_code`) and
`min_transformers_version`.

### Behavior changes

**1. Lookup key: `type(root_model).__name__.lower()` → `config.model_type`.** The old key
changes after `quantize` wraps the model, forcing substring matching; `model_type` is
stable, so `ExportContext` carries it and lookups are exact. Matching is also tightened to
case-insensitive **exact** names against the module's **MRO**, so quantized subclasses
still match via their base class without substring false positives.

**2. ⚠️ `get_expert_linear_names` raises instead of guessing.** Unmatched MoE blocks
previously fell through to `["w1", "w2", "w3"]` (Mixtral naming); they now raise
`NotImplementedError` telling you to register a `ModelSpec`.

The blast radius depends on the transformers release, because only *iterable* expert
layouts consult the spec — a fused expert container is resolved by the structural
first-projection check and never asks for naming. Measured against both ends of the
support matrix:

| transformers | Unregistered MoE families `is_moe` admits | Reach the spec lookup | Actually affected |
|---|---|---|---|
| 5.14.1 (`tf_latest`) | 19 | **0** — all fused | none |
| 4.57.6 (`tf_min`) | 9 | 7 iterable | **2** |

On 4.57, five of the seven (`ernie4_5_moe`, `flex_olmo`, `jamba`, `olmoe`,
`qwen3_omni_moe`) name their experts `gate_proj`/`up_proj`/`down_proj`, so the `w1`
fallback already failed with `AttributeError` on `main` — for them this trades an
unhelpful error for one that names the fix.

The other two, **`minimax` and `phimoe`**, are Mixtral-derived with `w1`/`w2`/`w3`
experts and did export on `main`, so they are now **registered** rather than left to
regress. Both only take the per-expert path on transformers 4; on 5 their experts are
fused (`MiniMaxExperts`, `PhimoeExperts`) and the structural check resolves them, which
their specs decline to answer for. **So no model that exported correctly before this
change regresses.**

This is backward-incompatible and has a `CHANGELOG` entry.

**3. DeepSeek-V3 and V4 are now registered, and newly visible to the MoE path.** Neither
was reachable before: `DeepseekV3MoE` (like `DeepseekV2Moe` and `DeepseekV32MoE`) is
invisible to `is_moe` — its class name does not end in `SparseMoeBlock` and it calls its
router `gate`, so neither the name test nor the structural `router`+`experts` test
matches. `DeepseekV4SparseMoeBlock` was detected by name but had no spec to resolve expert
naming from. Registering `block_names` is what puts V3 on the MoE path at all, so this is
added coverage rather than a fix to existing behavior.

`deepseek` stays registered alongside them: it describes the remote-code `DeepseekMoE`
block of DeepSeek-MoE/V1, and its spec is the only thing making that block detectable.

**4. Fused expert containers are skipped in `get_experts_list` instead of crashing.**
transformers 5 replaced several iterable expert `ModuleList`s with a single module holding
3-D parameters, while the specs still describe the transformers 4 iterable layout — a spec
cannot tell the two apart, only the module can. The AWQ/SVDQuant resmooth pass in
`requantize_resmooth_fused_llm_layers` reached `len(module.experts)` on a module with no
`__len__` and died with `TypeError` mid-export. **Mixtral already hits this on `main`**;
registering `deepseek_v3` only made an existing bug easier to reach. The skip is scoped to
specs that claim iterable experts, so a layout the spec calls unsupported (DBRX, whose
per-expert linears live under `experts.mlp`) still fails loudly rather than silently
dropping resmoothing.

Everything else is behavior-preserving, pinned by tests:

- **`sync_moe_gate_up_amax` keeps generic coverage.** Quantization admits MoE blocks
  *structurally*, so unregistered families (Olmoe, Jamba, MiniMax…) reach it. With no
  spec it falls back to every declared gate/up naming — what it did pre-registry.
  Skipping them would leave the fused `gate_up_proj` halves on inconsistent
  `weight_scale_2`.
- **`ExportSpec.grouped_expert_export` matches the legacy `get_experts_list` support set**,
  except for `deepseek_v3`/`deepseek_v4`, which are new and had no legacy behavior to
  match. Note
  `qwen3_5_moe`: legacy keyed off the root class name and `"qwen3_5moeforcausallm"`
  matched none of its qwen substrings (the `_5` breaks `qwen3moeforcausallm`), so it
  raised — the spec keeps `False` to match. Enabling it belongs in its own PR.
- **`is_moe` consults the model's own spec first**, before the generic name and structural
  fallbacks, so per-model data always wins. The three checks are or-ed, so this is a
  no-op; the same reordering is deliberately *not* applied to `get_expert_linear_names`,
  where the structural fused-experts check must stay first.
- **Two values are corrected, not moved.** `gpt_oss` declared
  `block_names=("GptOssMoE",)`, which matches no real module (transformers names it
  `GptOssMLP`); naming still resolved via the single-naming shortcut, so this was
  invisible. Now `("GptOssMLP", "GptOssMoE")`.

**5. ⚠️ DBRX expert input amax is now populated, changing its exported scales.** The
second corrected value, called out separately because it has a numerical consequence. The
legacy branch keyed on `DBRXMoeSparseMoeBlock`, a class transformers does not define — it
names the block `DbrxFFN`. So DBRX fell through to the `w1`/`w2`/`w3` default, every
`hasattr(experts_mlp, linear_name)` in `_prepare_dbrx_experts` evaluated `False`, and the
handler wrote no expert input amax at all. `dbrx/specs.py` declares the real names
(`w1_linear`, `w2_linear`, `v1_linear`), so the handler now does what it was written to
do.

Anyone diffing a re-exported DBRX checkpoint against an older one will see different
expert activation scales. That is the fix working, not a regression from the refactor. No
`CHANGELOG` entry: DBRX is not listed as a supported export target in
`docs/source/deployment/` or `examples/hf_ptq/README.md`.

### Scope

One consumer wired up: the unified HF export path.

The TRT-LLM builders now live in `modelopt/torch/export/trtllm/` after #2365, and are
deprecated. This PR changes one import line there — `is_moe` moved to
`modelopt/torch/models/moe.py`, so importing it from `..layer_utils` no longer resolves —
and nothing else. Their `is_moe()` calls still pass no `model_type` and fall back to the
all-specs search, reproducing today's behavior; their hardcoded tables, including a second
copy of the weight-plus-one norm names, stay as they are and migrate whenever that package
does.

Worth noting that #2365 landed the same boundary from the other direction: the slimmed
`export/layer_utils.py` is now documented as "module-shape predicates and MoE quantizer
helpers shared by every export backend", which is exactly the set of functions this PR
rewires. The Megatron path has its own per-family registry and is untouched; folding it in
is a later question, not this PR's.

`is_moe` itself moved out of `modelopt/torch/export/layer_utils.py` into
`modelopt/torch/models/moe.py` — whether a module is an MoE block is a modeling question,
not an export one, and it is the first piece of shared modeling logic to follow the specs
into the new package.

### Relationship to #1939 (why a second registry?)

Orthogonal layers: #1939's `ExportModuleRegistry` dispatches on **module structure**
(*which handler runs?*), this registry resolves **family data** (*what are its values?*).
Exporting a `QuantMixtralSparseMoeBlock`, #1939 picks the shared iterable-experts handler
(also serving Qwen/DeepSeek/Gemma4); inside, `get_expert_linear_names` resolves the
`mixtral` spec to `("w1", "w2", "w3")`. One handler serves many families, one spec serves
many handlers. Sharing the matcher machinery is a planned follow-up.

### Usage

```python
# modelopt/torch/models/qwen3_moe/specs.py — adding a model needs no engine edits
register(
    ModelSpec(
        model_type="qwen3_moe",
        min_transformers_version="4.57",
        moe_spec=MoESpec(
            block_names=("Qwen3MoeSparseMoeBlock",),
            expert_linear_names=("gate_proj", "down_proj", "up_proj"),
            gate_up_pair=("gate_proj", "up_proj"),
        ),
        export_spec=ExportSpec(
            grouped_expert_export=True,
            pqs_fuse_rules=(
                (("Qwen3MoeAttention",), "v_proj", "o_proj"),
                (("Qwen3MoeMLP",), "up_proj", "down_proj"),
            ),
        ),
    )
)
```

One layout per model. `block_names` is a tuple, so a model whose MoE appears under several
class names is covered as long as they share a layout (`gpt_oss`'s `GptOssMLP`/`GptOssMoE`).
`gemma4_text` imports its section from `gemma4` rather than restating it.

### Testing

`tests/unit/torch/export/` — **202 passed**; `tests/unit/torch/quantization/` — **903
passed**. The `test_export_diffusers.py` collection error and 6
`test_quant_aware_conversion.py` failures are pre-existing and reproduce on untouched
`main`, as does the one failing pre-commit hook (`generate-arguments-md`, no torch in its
venv).

`tests/unit/torch/models/test_model_specs.py`: registry matching (MRO / quantized classes
/ model-type scoping); every registered MoE section's `(block_names, expert_linear_names,
fused_expert_names, gate_up_pair)` as an exhaustive table that fails if a spec is added
without a row; the exhaustive `grouped_expert_export` support set; the structural
fused-expert shortcut and the spec's precedence over it; the fused-container skip in
`get_experts_list`; the `sync_moe_gate_up_amax` fallback; and legacy-equivalence of the
aggregated `pqs_fuse_rules` / `gate_up_pairs` / `weight_plus_one_norm_names`.
Mutation-checked — flipping a flag, typo-ing a block name, or swapping the precedence each
fail.

`tests/unit/torch/models/test_specs_vs_transformers.py` validates the specs against the
*installed* transformers rather than a mirrored table: every registered block class must
name a real class from its `min_transformers_version` on, and every `remote_code` model
must still be absent. It names no model — which specs to check, and which cannot be, both
come from the registry.

> The run counts above were measured before the follow-up commits in this branch (spec
> precedence, the policy split, the `MoESpec`/`MoELayout` merge) and have not been
> re-measured; CI on the current head is the authority.

**Local per-model-type E2E sweep** (transformers 5.3.0, tiny models from config, run
against this branch and base `main`; not checked in):

| model_type | real block class | this branch | base `main` |
|---|---|---|---|
| `qwen3_moe` / `qwen2_moe` / `qwen3_next` | `Qwen*SparseMoeBlock` | `gate/down/up` | same |
| `mixtral` | `MixtralSparseMoeBlock` | `w1/w2/w3` | same |
| `nemotron_h` | `NemotronHMoE` | `up/down` | same |
| `gpt_oss` | `GptOssMLP` | `gate_up/down` | ❌ `w1/w2/w3` |
| `deepseek_v3` | `DeepseekV3MoE` | raises | ❌ `w1/w2/w3` |

Five types identical to base; both differences are base bugs the registry surfaces.
`dbrx` could not be built on transformers 5.3.0 (`DbrxAttentionConfig` missing
`rope_theta`). `sync_moe_gate_up_amax` was separately checked against base across
registered, unregistered, and no-config models — identical in all cases.

- Is this change backward compatible?: ❌ — `get_expert_linear_names` no longer falls back to `w1/w2/w3`; see "Behavior changes" #2.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no copied code, no new dependency (stdlib only).
- Did you write any new necessary tests?: ✅ — `tests/unit/torch/export/test_model_specs.py`.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ❌ — happy to add an entry given the backward-incompatible item.
- Did you get Claude approval on this PR?: ✅ — run; its CRITICAL finding on `sync_moe_gate_up_amax` is fixed above.

## Summary by CodeRabbit

- **New Features**
  - Added model-aware export support across a broad range of MoE and dense architectures, including newer DeepSeek, Gemma, Qwen, Nemotron, Arctic, DBRX, GPT-OSS, Llama, and Mixtral variants.
  - Improved quantization and export handling for model-specific expert layouts, projection fusion, normalization, and gate/up synchronization.
  - Added automatic architecture recognition using Hugging Face model metadata.

- **Bug Fixes**
  - Unsupported or ambiguous model layouts now produce explicit errors instead of applying potentially incorrect defaults.

- **Documentation**
  - Added documentation describing model specifications and supported architecture metadata.


modelopt/modeling/base.py
```diff
@@ -0,0 +1,95 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .hooks import ModelHooks
+
+__all__ = ["ModelSpec"]
+
+
+@dataclass
+class ModelSpec:
```
h-guo18: trtllm export path is deprecated.

modelopt/torch/export/layer_utils.py
```diff
@@ -60,6 +60,13 @@
     RgLruConfig,
 )
 from .model_config_utils import pad_weights
```
h-guo18, line 63: HF export only.

modelopt/torch/export/MODEL_SPECIFIC_REFACTOR.md
```diff
@@ -0,0 +1,120 @@
+# Export: Model-Specific Logic Refactor — Action Plan
```
sychen52: Is this file meant to be checked in?

h-guo18: No, it will be removed

h-guo18: This PR is still messy. Need some cleaning before ready for review

modelopt/torch/models/__init__.py
kevalmorabia97, line 1: Is this module better suited in `modelopt.torch.export.models`?

shengliangxu: no, let's move to modelopt.torch.models

and make a dir for each model_type, and rename these python files to be export.py for now.

We may rename these files as we keep developing

h-guo18: Moved to modelopt.torch.models as @shengliangxu suggested

docs/source/reference/1_modelopt_api.rst
```diff
@@ -10,5 +10,6 @@ modelopt API
    :recursive:
 
    modelopt.deploy
+   modelopt.torch.models
```
kevalmorabia97: why do we need to add it here? It should already be inside existing modelopt.torch link

modelopt/torch/models/arctic/__init__.py
kevalmorabia97, line 1: do we want to take this opportunity to drop old models instead of migrating all?

modelopt/torch/models/__init__.py
kevalmorabia97, line 1: Can you add `modelopt/torch/models/README.md` describing how to add a new model support?

docs/source/reference/1_modelopt_api.rst
h-guo18: Done, thanks!

modelopt/torch/models/arctic/__init__.py
h-guo18: Sure. Do we know what models to drop? (Otherwise we can drop them in a following PR)

modelopt/torch/models/__init__.py
h-guo18: added

modelopt/torch/export/layer_utils.py
```diff
@@ -310,14 +301,18 @@ def is_mlp(module: nn.Module) -> bool:
     return any(key in type(module).__name__.upper() for key in ("MLP", "T5DENSE"))
 
 
-def is_moe(module: nn.Module) -> bool:
-    """Returns whether the module is an MOE layer."""
+def is_moe(module: nn.Module, model_type: str | None = None) -> bool:
```
shengliangxu: These logics are modeling logics, not really export utilities.

They should be moved to torch/models/common.py or torch/models/moe.py

modelopt/torch/export/layer_utils.py
```diff
@@ -310,14 +301,18 @@ def is_mlp(module: nn.Module) -> bool:
...
+    model's own spec.
+    """
     name = type(module).__name__.lower()
     # Auto-detect common MoE patterns
     if name.endswith("sparsemoeblock") or "moelayer" in name:
         return True
-    # Explicit matches for non-standard naming
-    if any(key in name for key in ["arcticmoe", "deepseekmoe", "dbrxffn", "nemotronhmoe"]):
+    # Non-standard MoE block names are per-model data (modelopt/torch/models/*).
+    if match_moe_block(module, model_type) is not None:
```
shengliangxu: we should put the model_type dispatch logic at first, then do the common logc. The model_type logic always override the common logic.

modelopt/torch/models/arctic/specs.py
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
shengliangxu, line 1: in fact, I think we should rename it to specs.py, these are general model specs, export is just a consumer of it.

tests/unit/torch/models/test_model_specs.py
```diff
@@ -0,0 +1,639 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
shengliangxu, line 1: let's move it into a new location:

tests/unit/torch/models

Optionally, we can also add a test for each model_type we have:

```
tests/unit/torch/models/nemotron.py
tests/unit/torch/models/qwen2_moe.py

```

And have tests to validate if the specs is correct given the transformers model definition. But optional.

tests/unit/torch/models/test_model_specs.py
```diff
@@ -0,0 +1,639 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
```
shengliangxu, line 3: One more thing that I think we might need, definitely not now, but in the future is to have versioned specs so we can have the specs of the model evolve along side transformers. But definitely too much for now.

modelopt/torch/models/specs.py
```diff
@@ -0,0 +1,163 @@
...
+class MoESpec:
+    """Topic section: MoE architecture facts — the model's MoE-block layout(s).
+
+    This describes what a model's MoE blocks *are* — which class, what the expert
+    projections are called — so any modelopt subsystem (export, quantization,
+    speculative decoding, ...) can read it instead of keeping its own per-model MoE
+    table.
+    """
+
+    moe_variants: tuple[MoEVariant, ...] = ()
```
shengliangxu: when will a model have multiple MoE variants?

modelopt/torch/models/specs.py
```diff
@@ -0,0 +1,163 @@
...
+    """Class names of norm layers whose stored weight is ``w - 1`` (the effective
+    scale is ``weight + 1``), e.g. Gemma's RMSNorm variants and LayerNorm1P.
+    Matched against a norm module's MRO (case-insensitive exact names). Export
+    must account for the +1 when folding scales into the norm weight (AWQ
+    pre_quant_scale fusion). A structural fallback (``zero_centered_gamma``) stays
+    in the engine."""
+
+
+@dataclass(kw_only=True)
+class ModelSpec(MoESpec, ExportSpec):
```
shengliangxu: I think this ExportSpec should be outside of the ModelSpec, it does not really define general model facts. These ExportSpec stuff should be integrated with @sychen52 's fusion stuff in some way. But do not need to be done in this change.

modelopt/torch/quantization/model_calib.py
```diff
@@ -95,10 +96,10 @@ def _is_calibrated_nvfp4_static(q) -> bool:
 
 def _collect_grouped_linears(model: nn.Module) -> list[list[nn.Module]]:
     """Collect name-based sibling groups (Q/K/V, gate/up, w1/w3) of calibrated NVFP4-static linears."""
-    # Inline import: layer_utils -> quant_utils -> model_calib cycle.
-    from modelopt.torch.export.layer_utils import _GATE_UP_PAIRS
-
-    patterns: tuple[tuple[str, ...], ...] = (("q_proj", "k_proj", "v_proj"), *_GATE_UP_PAIRS)
+    patterns: tuple[tuple[str, ...], ...] = (
+        ("q_proj", "k_proj", "v_proj"),
+        *list_all_possible("gate_up_pairs"),
```
shengliangxu, line 104: we should not need to invoke list_all_possible in pipelines where we clearly work on one model. This method should only be used for some kind of reporting of the registry.

However, I can understand why we are using this enumeration-of-the-whole-world API. It's exactly because we do not have a ModelOptSession to hold the global state of what we are working on. @Fridah-nv

shengliangxu: We can have it now, but will need to remove it in the future.

modelopt/torch/models/specs.py
```diff
@@ -0,0 +1,163 @@
...
+    """The one global per-model descriptor, composed from the section mixins.
+
+    Resolved by HF model type (see ``registry.get_spec``); a model registers exactly
```
shengliangxu: I feel it might make more sense to have MoESpec as an attribute instead of using inheritance, like if  a model is not an MoE model, then we should just have moe_spec = None instead of now we have the full MoE spec stuff still, although empty. How do you think?

modelopt/modeling/base.py
shengliangxu: Stale — this path no longer exists on the branch.

modelopt/torch/export/layer_utils.py
shengliangxu: Stale — this code has since moved; the registry lookups are HF-export-only and the TRT-LLM path keeps its own branches.

modelopt/models/registry.py
```diff
@@ -0,0 +1,115 @@
...
+    E.g. ``list_all_possible("gate_up_pairs")``. The result is a global vocabulary:
+    consumers match it against any model's modules, so adding a value to one spec
+    affects all models the consumer walks — prefer ``get_spec(model_type)`` /
+    ``match_moe_block`` wherever the owning model is identifiable.
+    """
+    if attr not in _spec_attr_names():
+        raise ValueError(
+            f"{attr!r} is not a ModelSpec attribute; available: {sorted(_spec_attr_names())}"
+        )
+    return tuple(dict.fromkeys(item for spec in get_specs() for item in getattr(spec, attr)))
```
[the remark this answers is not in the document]
shengliangxu: Already handled. `list_all_possible` deduplicates by equality rather than by hashing, and rejects non-tuple attributes instead of iterating a scalar character by character.

The unhashable case no longer arises at all: since `MoELayout` was merged into `MoESpec`, no tuple-valued attribute holds unhashable items.

modelopt/torch/export/layer_utils.py
```diff
@@ -1212,44 +1182,61 @@ def sync_moe_gate_up_amax(model: nn.Module) -> int:
...
     synced = 0
+    unmatched_block_names: set[str] = set()
     for _, sub_module in model.named_modules():
         if not (is_moe(sub_module) and hasattr(sub_module, "experts")):
             continue
         if not hasattr(sub_module.experts, "__iter__"):
             continue
+        # The gate/up pair comes from the model's own spec; a variant with no pair
+        # (non-gated or already-fused experts) needs no sync.
+        variant = match_moe_block(sub_module, model_type)
```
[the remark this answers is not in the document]
shengliangxu: Fixed in ea227b4f2, before this review landed. `sync_moe_gate_up_amax` falls back to every declared gate/up naming when no spec matches, so unregistered families keep their coverage — skipping them would have left the two halves of the fused `gate_up_proj` on inconsistent `weight_scale_2`, which is the corruption the function exists to prevent.

modelopt/torch/export/layer_utils.py
```diff
@@ -1212,44 +1182,61 @@ def sync_moe_gate_up_amax(model: nn.Module) -> int:
     Returns:
         Number of expert gate/up pairs whose amaxes were synced.
     """
+    model_type = hf_model_type(model)
     synced = 0
+    unmatched_block_names: set[str] = set()
     for _, sub_module in model.named_modules():
         if not (is_moe(sub_module) and hasattr(sub_module, "experts")):
```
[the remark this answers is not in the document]
shengliangxu: Applied — that call site threads `model_type` through now.

The same fix was needed in the layerwise export path after main's #2136 landed: it calls `_prepare_moe_inputs` and `sync_moe_gate_up_amax` with a single decoder layer rather than the root model, so both now take an explicit `model_type` and `LayerwiseExporter` resolves it once from the root.

modelopt/torch/models/arctic/__init__.py
shengliangxu: Two candidates, if it helps unblock the list.

`deepseek` describes `DeepseekMoE`, the remote-code block of DeepSeek-MoE/V1. Nothing in the repo documents or exercises it — `docs/source/deployment/3_unified_hf.rst` and `examples/hf_ptq/README.md` list only R1, V3, V3.1, V3.2 and V4, and `examples/deepseek/*/ptq.py` builds the model from DeepSeek's own GitHub inference code without touching the registry. `arctic` is in the same position.

One caveat for the follow-up: both are load-bearing in a narrow way. Their specs are the only reason `is_moe` detects those blocks at all, since neither class name matches the generic `sparsemoeblock`/`moelayer` patterns and neither has a `router` attribute for the structural check. Dropping them makes the blocks invisible to the export path rather than merely removing dead data, so it wants a CHANGELOG entry.

This PR meanwhile adds `deepseek_v3` and `deepseek_v4` — the models the docs do claim, and which were previously unreachable for the same reason: `DeepseekV3MoE` calls its router `gate`, so it matched neither the name test nor the structural test.

modelopt/torch/export/layer_utils.py
shengliangxu: Moved. `is_moe` now lives in `modelopt/torch/models/moe.py`, and every consumer imports it from the modeling package rather than from `layer_utils`.

It went into its own module rather than into `specs.py` because it needs torch at runtime — its last resort is a structural check on the module's children, which no per-model data can replace for a model nobody has registered. Keeping `specs.py` to data and string matching preserves the split the section classes already follow. Its tests moved with it, to `tests/unit/torch/models/test_moe.py`.

modelopt/torch/export/layer_utils.py
shengliangxu: Applied, and it was the most interesting of the comments to act on.

`is_moe` and `get_experts_list` were straightforward reorderings. `get_expert_linear_names` was not: a spec's `expert_linear_names` describe *one* layout, and the same model type materializes per-expert on transformers 4 and fused on 5 — so Mixtral's `w1/w2/w3` applied to a fused container is wrong rather than merely generic.

So the layout gained `fused_expert_names`, and the resolver now takes the layout observed on the module. A spec that only describes the per-expert form *declines* for a fused container instead of answering wrongly, which is what makes spec-first safe. Where a spec does describe the layout in hand it wins outright — a `gpt_oss` container whose `_first_proj_attr` says `up_proj` still exports as `gate_up_proj`/`down_proj`.

modelopt/torch/models/arctic/specs.py
shengliangxu: Renamed — every `<model_type>/export.py` is now `<model_type>/specs.py`.

You were right that the content was misfiled: `moe_spec` is an architecture fact, not export data, so naming the file after one consumer was wrong. The package docstring now says the file is named for what it holds, not for who reads it.

tests/unit/torch/models/test_model_specs.py
shengliangxu: Moved to `tests/unit/torch/models/`, now three files: `test_model_specs.py`, `test_moe.py`, and `test_specs_vs_transformers.py`.

The last one is the optional idea from this comment — validating the specs against the transformers definition rather than against a mirrored table. It checks every registered block class against the *installed* transformers, and it found a real discrepancy immediately: `nemotron_h` declares `NemotronHMOE` while transformers defines `NemotronHMoE`. That works at runtime only because matching is case-insensitive.

It names no model: which specs to check, and which cannot be, both come from the registry via `modeling_source` and `min_transformers_version`.

tests/unit/torch/models/test_model_specs.py
shengliangxu: Agreed, and partly started. `ModelSpec.min_transformers_version` records the release a spec's definitions come from, which is what lets the transformers test tell an expected absence on an older transformers apart from a spec that no longer matches reality — values measured against the 4.57 and 5.x wheels rather than recalled.

Full versioning, where specs evolve alongside transformers, is still future work.

modelopt/torch/models/specs.py
shengliangxu: Only Mixtral did, and that entry turned out to be dead: `MixtralMoeSparseMoeBlock` with MCore `linear_fc1`/`linear_fc2` naming matches no class in any transformers release from 4.57 through 5.14 — all 16 checked. It has been dropped.

Every model now declares exactly one layout, so `MoELayout` has been merged into `MoESpec`; the wrapper had a single field and trivial accessors.

For the record on the general question: the only model in transformers with two genuine MoE block classes is `qwen3_omni_moe` (Thinker and Talker), and they share a layout — so `block_names` being a tuple covers it, the same way it covers `gpt_oss`'s `GptOssMLP`/`GptOssMoE`. If a model ever needs two *different* layouts, the section becomes a tuple on `ModelSpec`.

modelopt/torch/models/specs.py
shengliangxu: Not done here, per your note.

It has moved one step in that direction: `export_spec` is an optional attribute rather than a mixin, and it now carries policy as well as data — `grouped_expert_export` is a statement about what modelopt has validated, not a fact about the model, which is why it lives in the export section rather than on the MoE layout. Integrating it with the fusion work stays open.

modelopt/torch/quantization/model_calib.py
shengliangxu: Still there, per your "we can have it now".

Recording what an audit of it found, for whenever it is removed: `list_all_possible` has four production call sites naming three vocabularies, and the machinery behind it — `SpecSection`, `_spec_sections`, `_spec_attr_names`, `_read_spec_attr` and the function itself — is 87 lines of string-keyed reflection. Three typed accessors would replace all of it in about 15 lines, and would let mypy check the return types, which it cannot today.

modelopt/torch/models/specs.py
shengliangxu: Done. `ModelSpec` now holds `moe_spec` and `export_spec` as optional attributes; a dense model carries `moe_spec=None` rather than an empty section, so "not an MoE model" stays distinguishable from "an MoE model whose layout is not filled in yet".

The data backed the instinct: 6 of the 18 specs at the time carried a completely empty `MoESpec`.

modelopt/torch/models/arctic/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Per-model specs for HF model type ``arctic``."""
+
+from . import specs
```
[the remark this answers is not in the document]
shengliangxu: `CONTRIBUTING.md` does require this, but I do not think it applies to these packages: they have no public API. `from . import specs` exists so that importing the package runs `register(...)` — there is nothing to export.

There is precedent in the repo for exempting registration-by-import packages: `modelopt/torch/export/plugins/__init__.py` and `modelopt/torch/quantization/plugins/__init__.py` are the same shape and neither defines `__all__`.

Happy to reconsider if a reviewer disagrees.

modelopt/torch/models/deepseek_v3/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Per-model specs for HF model type ``deepseek_v3``."""
+
+from . import specs
```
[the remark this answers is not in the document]
shengliangxu: Same as the sibling comment on `arctic/__init__.py`: these packages exist so that importing them registers a spec, so there is no curated API to re-export. `from .specs import *` would pull in `ModelSpec`, `MoESpec` and `register` — helpers, not exports.

Precedent: `export/plugins/__init__.py` and `quantization/plugins/__init__.py` are side-effect registration packages and neither defines `__all__`.

modelopt/torch/models/deepseek_v3/specs.py
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+
+"""DeepSeek-V3 specs (HF model type ``deepseek_v3``).
+
+Covers the transformers-native V3 family (V3, R1, V3.1). The remote-code
+``DeepseekMoE`` block of DeepSeek-MoE is a separate spec under ``deepseek``.
+"""
+
+from ..registry import register
+from ..specs import ModelSpec, MoESpec, MoEVariant
```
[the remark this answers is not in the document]
shengliangxu: This part is fair. These modules import `ModelSpec` / `MoESpec` / `register` as helpers, which would leak under a star-import.

Nothing star-imports them today — the package uses `from . import (arctic, dbrx, ...)`, not star imports — so it is hygiene rather than a fix. Happy to add `__all__ = []` across all 20 if reviewers want it.

modelopt/torch/models/gemma4/specs.py
```diff
@@ -0,0 +1,43 @@
...
++1 handling is validated on Gemma4.
+"""
+
+from ..registry import register
+from ..specs import ModelSpec, MoESpec, MoEVariant
+
+# Gemma4 MoE experts are unfused into per-expert nn.Linear layers. The MoE block lives
+# in the text model, so ``gemma4_text`` reuses this exact layout -- see
+# ``gemma4_text/specs.py``, which imports it rather than restating it.
+GEMMA4_MOE_VARIANTS = (
```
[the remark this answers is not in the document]
shengliangxu: Fair, and it genuinely is a cross-module API since `gemma4_text/specs.py` imports it.

Note the name has changed: it is `GEMMA4_MOE_SPEC` now, after `MoELayout` was merged into `MoESpec` and the plural was dropped.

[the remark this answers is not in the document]
shengliangxu: Revisiting this — I conceded too quickly above. The name should be **private**, not exported, and the import is legitimate as an intra-family detail. Done in 5d81516: it is `_GEMMA4_MOE_SPEC` now.

The reasoning: `gemma4_text` is a sub-model type of the same family, and transformers keeps the two in one package — its `SPECIAL_MODEL_TYPE_TO_MODULE_NAME` maps `gemma4_text` → `gemma4` (`auto_mappings.py`, from 5.5.0 on), so upstream `gemma4_text` has no module of its own. The directories are split here only because this registry is keyed on `config.model_type` and there happen to be two strings for one family. So the import crosses a boundary our keying invented, not one the model has — which makes it an implementation detail rather than a package API.

Underscoring also covers the concrete half of the concern for free: underscore-prefixed names are excluded from `from x import *`, so the name cannot leak that way.

We are deliberately keeping the two directories rather than registering both model types from `gemma4/specs.py`, which would remove the import entirely: one directory per `model_type` is what makes the registry browsable, and that is worth more than mirroring upstream packaging.

modelopt/torch/models/deepseek_v3/specs.py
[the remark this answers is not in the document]
shengliangxu: Done for the `specs.py` modules in ae97e86 — all 20 now declare `__all__: list[str] = []`. Typed rather than bare so mypy does not infer `list[<nothing>]` if a name is added later.

Empty on purpose: these modules exist so that importing them runs `register()`, and the names they do hold — `ModelSpec`, `MoESpec`, `ExportSpec`, `register` — are helpers imported to build that call, which is exactly what would leak under a star-import. Declaring the empty list states that rather than leaving it to omission. The rationale is in `modelopt/torch/models/README.md` once rather than repeated as a comment in twenty files.

**Not** doing the `__init__.py` half from the sibling comments, to be explicit about it rather than leaving it looking half-finished. Those packages have no API to re-export either: `from . import specs` is there for the registration side effect, and `from .specs import *` would now import nothing at all given the `__all__` above. There is precedent in the repo for exempting registration-by-import packages — `modelopt/torch/export/plugins/__init__.py` and `modelopt/torch/quantization/plugins/__init__.py` are the same shape and neither declares `__all__`.

Happy to be overruled by a human reviewer on that half.

modelopt/torch/models/arctic/__init__.py
shengliangxu: Which models to drop is independent of this PR — this one is a refactor that moves existing per-model data into the registry without changing what is supported. Dropping models is a separate behavior change and deserves its own discussion and PR, which matches what @h-guo18 suggested above.

Resolving this thread so it does not block the refactor; the candidates and the caveat are recorded above for whoever picks that up.

modelopt/torch/export/quant_utils.py
```diff
@@ -1379,12 +1375,10 @@ def fuse_prequant_to_linear(model: torch.nn.Module, fuse_grouped_heads=False):
     """
     # Fuse pre_quant_scale to the linear weights
     for _, module in model.named_modules():
-        for module_map in PQS_FUSE_MODULE_MAPPING:
-            target_module_list = module_map[0]
-            linear_pair = module_map[1]
+        for target_module_list, fuse_into, fuse_from in list_all_possible("pqs_fuse_rules"):
```

Two things here:

1. `list_all_possible(...)` is now evaluated once per module in the `named_modules()` walk. It re-scans every registered spec and deduplicates with `item not in values` (O(n²)) on each call. Previously this iterated a module-level constant. Hoist it above the loop.

2. Worth calling out in the docs: `pqs_fuse_rules` reads as per-model data but is consumed as a global vocabulary — the rules registered under `llama/specs.py` are matched against every model's modules. Same for `weight_plus_one_norm_names` and `gate_up_pairs` in `model_calib._collect_grouped_linears`. `specs.list_all_possible`'s docstring notes this, but from a contributor's point of view "add a rule to my model's `specs.py`" now silently changes behaviour for all models. Since `fuse_prequant_to_linear` walks a single root model, could it take the root `model_type` and prefer `get_spec(model_type).export_spec.pqs_fuse_rules`, falling back to the aggregate only when unregistered?

modelopt/torch/models/deepseek/specs.py
```diff
@@ -0,0 +1,38 @@
...
+
+Matches the remote-code ``DeepseekMoE`` block, not the HF-native ``deepseek_v3``
+classes.
+"""
+
+from ..specs import ModelSpec, MoESpec, register
+
+__all__: list[str] = []
+
+# has_iterable_experts stays False until the grouped export path is validated on
```
cjluo-nv: > _Bot comment._

Stale comment: `has_iterable_experts` no longer exists (`test_moe_layout_carries_no_export_policy` asserts `MoESpec` has no such attribute). The equivalent today is `ExportSpec.grouped_expert_export`, which this spec does not declare at all. Same comment appears in `qwen3_5_moe/specs.py`.

modelopt/torch/models/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+Consumers resolve a spec via the registry lookups (``get_spec`` / ``match_moe_block``)
+and read its sections.
+
+The per-model file is named for what it holds, not for who reads it: a model's spec is
+general model data, and export is only its first consumer.
+"""
+
+from .moe import *
+from .specs import *
+
```
cjluo-nv, line 66: > _Bot comment._

`moe.py` and `specs.py` both declare `__all__`, but this package `__init__` does not — so `from modelopt.torch.models import *` also re-exports `arctic`, `dbrx`, … and any incidental names. Please add an explicit `__all__` here listing the public lookups (`get_spec`, `get_specs`, `hf_model_type`, `is_moe`, `list_all_possible`, `match_class_names`, `match_moe_block`, `match_moe_model`, `register`, plus the spec classes).

cjluo-nv (commented):

The design is well argued (the PR body explicitly answers "why a second registry?" vs #1939 and the Megatron/TRT-LLM paths) and the test coverage is unusually thorough, but the PR knowingly ships an export regression and leaves a few loose ends.

Needs action:
- [ ] Register `minimax` and `phimoe` specs under `modelopt/torch/models/` — the PR body identifies them as the only two models that exported on `main` and now raise; both are Mixtral-derived `w1`/`w2`/`w3`, so this closes the whole regression in ~10 lines. If it must ship broken, say so in `CHANGELOG.rst` by name.
- [ ] Hoist `list_all_possible("pqs_fuse_rules")` out of the per-module loop in `quant_utils.py:1378` (see inline) — it rebuilds an O(n²)-deduped tuple once per module.
- [ ] Point `_weights_plus_one` inside `build_layernorm_config` (`layer_utils.py`) at `list_all_possible("weight_plus_one_norm_names")`, or note why it stays hardcoded; today `nemotron`'s new `NemotronLayerNorm1P` reaches only one of the two copies.
- [ ] Drop the stale `has_iterable_experts` comments in `models/deepseek/specs.py` and `models/qwen3_5_moe/specs.py` — that field no longer exists (a test asserts as much).
- [ ] Add an explicit `__all__` to `modelopt/torch/models/__init__.py` so `import *` does not re-export the 20 submodule names.

No action needed:
- 58 files / +2912 is large, but the split (registry core vs. export wiring) is genuinely coupled; flagging for awareness, not asking for a split.
- The `is_moe` tests deleted from `test_layer_utils.py` reappear intact in `tests/unit/torch/models/test_moe.py` — a justified move.

modelopt/torch/export/quant_utils.py
shengliangxu: Both right, fixed in 43f67ad.

**1. Hoisted.** You are right that this was a regression and not just a missed optimisation — it replaced a module-level constant, so the walk went from reading a list to rebuilding the vocabulary for every module in the model, rescanning all twenty specs with an O(n²) equality dedup each time. Resolved once above the loop now.

**2. Taken, with the scoping you suggested.** `fuse_prequant_to_linear` takes the root `model_type` and prefers `get_spec(model_type).export_spec.pqs_fuse_rules`, falling back to the aggregate when the model is unregistered or declares no rules — which is what every model got before, so nothing changes for them.

Worth recording why this is safe rather than merely tolerable: the aggregate was already self-scoping, because each rule is keyed on class-name substrings (`LlamaAttention`, `Qwen3MoeMLP`) that cannot match another family's modules. So no model was ever actually picking up another's rules — it was scanning rules that could never match. The change makes the intent explicit and stops the scanning; behaviour is identical either way.

Your broader point stands beyond this call site, and it is the same one @shengliangxu raised on `model_calib.py`: "add a rule to my model's `specs.py`" reading as local while being consumed globally is a real trap. `weight_plus_one_norm_names` and `gate_up_pairs` still go through `list_all_possible`. Those two are harder — `_layernorm_uses_weight_plus_one` gets a bare norm module with no root in sight, and `sync_moe_gate_up_amax` deliberately falls back to the global vocabulary so unregistered MoE families keep coverage. Both are worth a follow-up; this PR does the one where the root model is already in hand.

modelopt/torch/models/deepseek/specs.py
shengliangxu: Fixed in 43f67ad, both files.

You are right that the equivalent is `ExportSpec.grouped_expert_export`, and right that neither spec declares it — which is the point, but the comments were describing a field that no longer exists rather than saying so. They now read as "No `ExportSpec.grouped_expert_export` here", followed by the same reason as before: for `deepseek` the grouped export path is not validated on this model, and for `qwen3_5_moe` the legacy `get_experts_list` keyed off `type(root_model).__name__.lower()` and `"qwen3_5moeforcausallm"` matched none of its qwen substrings, so it raised there and the spec preserves that.

Missed on my side when `has_iterable_experts` moved off `MoESpec` into `ExportSpec` — the field went, the prose explaining it did not.

modelopt/torch/models/__init__.py
shengliangxu: Fixed in 43f67ad — thirteen named imports and an explicit `__all__`.

You caught the gap exactly: the per-model `specs.py` modules had `__all__` added earlier and `moe.py` has one, but the package `__init__` did not, so `from modelopt.torch.models import *` re-exported `arctic`, `dbrx` and the rest. Verified after the change: the star-import now yields exactly the thirteen public names and no subpackages.

One note on how, in case it comes up again. I first derived it as `__all__ = [*_moe_all, *_specs_all]`, which is drift-proof by construction, but ruff rejects it with `PLE0604` — it cannot statically verify a starred list holds strings. The named form cannot drift silently either, just for a different reason: a stale entry fails at import, and an unused import fails lint.

CHANGELOG.rst
```diff
@@ -12,6 +12,8 @@ Changelog
 
 **Backward Breaking Changes**
 
+- Unified HuggingFace export now fails with ``NotImplementedError`` when it meets an MoE block whose expert projection names it does not know, instead of assuming Mixtral's ``w1``/``w2``/``w3``. Models that relied on that assumption need a ``ModelSpec`` registered under ``modelopt/torch/models/``; every MoE architecture ModelOpt previously exported correctly is already registered.
```

This last clause — "every MoE architecture ModelOpt previously exported correctly is already registered" — does not match the registry: there is no `minimax` or `phimoe` package under `modelopt/torch/models/`, and the PR body identifies exactly those two as models that exported on `main` (Mixtral-derived `w1`/`w2`/`w3`) and now raise `NotImplementedError`. Either register the two specs (~10 lines each, same shape as `mixtral/specs.py`) or name them here so users hitting the new error can find the cause.

cjluo-nv (commented):

Comment: three of the five prior concerns are fixed, but the known `minimax`/`phimoe` export regression is still unregistered and the new `CHANGELOG.rst` line asserts the opposite.

Needs action:
- [ ] Fix the `CHANGELOG.rst` entry or register the specs: it claims "every MoE architecture ModelOpt previously exported correctly is already registered", yet `modelopt/torch/models/` has no `minimax` or `phimoe`, which the PR body itself calls the two genuine regressions (Mixtral-derived `w1`/`w2`/`w3`, previously exported fine).
- [ ] Point `_weights_plus_one` in `build_layernorm_config` (`layer_utils.py`) at `list_all_possible("weight_plus_one_norm_names")`, or add a one-line comment saying the TRT-LLM copy stays hardcoded on purpose — today the registry's new `NemotronLayerNorm1P` reaches only the `quant_utils.py` copy.

No action needed:
- :heavy_check_mark: Resolved since the last review: the per-module `pqs_fuse_rules` rebuild (now hoisted and model-scoped), the stale `has_iterable_experts` comments in `deepseek`/`qwen3_5_moe` specs, and the missing `__all__` in `modelopt/torch/models/__init__.py`.
- The `is_moe` tests removed from `test_layer_utils.py` reappear intact in `tests/unit/torch/models/test_moe.py` — a justified move.

CHANGELOG.rst
shengliangxu: Good catch — the sentence was aspirational, not accurate. It was written before the measurement that found those two, and I never went back to correct it.

Took your first option in 3a2186e: `minimax` and `phimoe` are now registered, so the claim holds rather than needing a caveat. Removing the breakage seemed better than documenting it, and both are the same shape as `mixtral/specs.py`.

Measured off the wheels rather than assumed. On transformers 4.57 both build an `nn.ModuleList` of `*BlockSparseTop2MLP` with `w1`/`w2`/`w3` — that is the naming registered, and the path that regressed. On 5.14 both are fused (`MiniMaxExperts`, `PhimoeExperts`), where the structural first-projection check resolves them and these specs decline, since the naming describes the per-expert form. Neither declares `grouped_expert_export`: legacy `get_experts_list` keyed off the root class name and neither `"minimaxforcausallm"` nor `"phimoeforcausallm"` matched its substrings, so grouped export raised on both and that is preserved.

Re-audited both wheels afterwards for anything else in the same position — an unregistered model that exported on `main` via the `w1`/`w2`/`w3` fallback and now raises. **None on either end of the matrix.** The PR body no longer says these two are left out.

cjluo-nv (commented):

Nudge — the previously flagged `minimax`/`phimoe` export regression and the contradicting `CHANGELOG.rst` claim are both fixed; one minor duplication item from the last round is still open on a PR this large.

Needs action:
- [ ] Point `_weights_plus_one` inside `build_layernorm_config` (`modelopt/torch/export/layer_utils.py`) at `list_all_possible("weight_plus_one_norm_names")`, or add a one-line comment stating the TRT-LLM copy stays hardcoded on purpose — the registry's norm names still reach only the `quant_utils.py` copy.
- [ ] Get an owner sign-off on the scope: 62 files / +3095 spanning the new `modelopt/torch/models/` package plus the export wiring is past what a bot should wave through.

No action needed:
- :heavy_check_mark: Resolved since the last review: `minimax`/`phimoe` specs registered so the CHANGELOG statement now holds, `pqs_fuse_rules` hoisted and model-scoped, stale `has_iterable_experts` comments, and `__all__` on `modelopt/torch/models/__init__.py`.
- The `is_moe` tests dropped from `tests/unit/torch/export/test_layer_utils.py` reappear intact in `tests/unit/torch/models/test_moe.py` — a justified move; the new spec-vs-transformers table tests are solid.
- Design rationale (vs `ExportModuleRegistry` #1939, Megatron registry, TRT-LLM path) is addressed in the PR body.

shengliangxu:
@cjluo-nv on the `_weights_plus_one` duplication from your last review — the answer changed under us, in a way that settles it.

#2365 landed in the meantime and moved the TensorRT-LLM builders into `modelopt/torch/export/trtllm/`. `build_layernorm_config` and its `_weights_plus_one` went with them, so that second copy of the norm names now sits entirely inside the deprecated package. It is not migrating because the package is not — it migrates whenever the package does.

So this PR does not point it at `list_all_possible(...)`, and does not add the explanatory comment there either: annotating deprecated code is not this PR's business. Its whole footprint inside `trtllm/` is one import line, forced rather than chosen — `is_moe` moved to `modelopt/torch/models/moe.py`, so importing it from `..layer_utils` stopped resolving and mypy rejected the implicit re-export. Verified path-only: `models.is_moe`, `export.layer_utils.is_moe` and the name `trtllm/layer_utils.py` resolves are all the same object, so those builders call exactly what they called before.

One substantive fact worth recording for whoever migrates that package: **the two lists agree today.** `nemotron`'s spec adds `NemotronLayerNorm1P`, which the hardcoded substring match already covers because it contains `LayerNorm1P`. So nothing is silently diverging while it waits.

The Scope section of the PR body now states this rather than describing the TRT-LLM move as still pending. Worth noting that #2365 landed the same boundary this PR argues for, from the other direction: the slimmed `export/layer_utils.py` is now documented as "module-shape predicates and MoE quantizer helpers shared by every export backend", which is exactly the set of functions rewired here.

Your other checkbox — an owner sign-off on the scope — still stands, and I agree with it. **meenchen** / **ChenhanYu** are the requested reviewers and have not weighed in yet.

cjluo-nv (commented):

Nudge — the last open code item (`_weights_plus_one` duplication) is settled by #2365 moving that copy into the deprecated `trtllm/` package, which I verified; what remains is scope sign-off plus one dead constant.

Needs action:
- [ ] Get an owner sign-off on the scope: 63 files / +3101 introducing `modelopt/torch/models/` plus the export/quantization rewiring is past what a bot should wave through; requested reviewers have not weighed in.
- [ ] Delete the now-unused `_GATE_UP_PAIRS` in `modelopt/torch/export/layer_utils.py` — `sync_moe_gate_up_amax` uses `list_all_possible("gate_up_pairs")` and `model_calib.py` no longer imports it.

No action needed:
- :heavy_check_mark: Resolved since the last review: `build_layernorm_config`/`_weights_plus_one` now live only in the deprecated `modelopt/torch/export/trtllm/layer_utils.py` (and its hardcoded `LayerNorm1P` substring still covers the registry's `NemotronLayerNorm1P`), on top of the earlier `minimax`/`phimoe` registration, `pqs_fuse_rules` hoisting and `__all__` fixes.
- `is_moe` tests moved intact from `tests/unit/torch/export/test_layer_utils.py` to `tests/unit/torch/models/test_moe.py` — a justified move; spec-vs-transformers tests are a strong addition.
- Design rationale (vs `ExportModuleRegistry` #1939, Megatron registry, TRT-LLM path) is addressed in the PR body.

cjluo-nv (commented):

LGTM — the last open code item is fixed; what remains is owner sign-off on the scope of a new top-level package.

Needs action:
- [x] Get an owner sign-off on scope: 63 files / +3102 adding `modelopt/torch/models/` plus export/quantization rewiring, with a documented backward-incompatible `get_expert_linear_names` change; requested reviewers are still silent.

No action needed:
- :heavy_check_mark: Resolved since the last review: `_GATE_UP_PAIRS` is deleted from `export/layer_utils.py` and `model_calib.py` now imports `list_all_possible` at module scope (the cycle-workaround inline import is gone); this follows the earlier `_weights_plus_one`, `minimax`/`phimoe`, `pqs_fuse_rules` and `__all__` fixes.
- `is_moe` tests moved intact from `tests/unit/torch/export/test_layer_utils.py` to `tests/unit/torch/models/test_moe.py`; the remaining `test_layer_utils.py` edits just thread the new `model_type` argument — a justified test edit.
- `tests/unit/torch/models/test_specs_vs_transformers.py` validates every registered block class against the installed transformers, which covers the models newly gated by `grouped_expert_export`.

kevalmorabia97:
/ok to test [e23df9d](https://github.com/NVIDIA/Model-Optimizer/pull/1828/commits/e23df9d34820fee507e3b85152ba5eb1a6af3429)

kevalmorabia97 (approved):
Approving to unblock

kevalmorabia97:
/ok to test [cb43b1b](https://github.com/NVIDIA/Model-Optimizer/pull/1828/commits/cb43b1bb5c039319b1977a676753e06208700cb3)

cjluo-nv (approved):

cjluo-nv (commented):

LGTM — the one open code item (Claude's IMPORTANT on `_layernorm_uses_weight_plus_one` narrowing) is fixed, and nothing else is outstanding.

Needs action:
- [ ] Confirm human merge sign-off: 63 files / +3116 adding `modelopt/torch/models/` plus export/quantization rewiring with a CHANGELOG'd backward-incompatible `get_expert_linear_names`; kevalmorabia97 approved "to unblock" — verify that covers the break.
- [ ] Decide on the still-open `hf_model_type` suggestion in `modelopt/torch/models/specs.py`: composite/VL roots resolve no spec (no `config.get_text_config()` fallback), which is now a hard `NotImplementedError` rather than a silent fallback. Fix or record as follow-up.

No action needed:
- :heavy_check_mark: Resolved since the last review: `_layernorm_uses_weight_plus_one` keeps substring matching over the MRO (so `T5Gemma`/`RecurrentGemma` norms still take the `+1` path) and `test_layernorm_weight_plus_one_via_specs` pins it; the DBRX expert input-amax activation is documented as behavior change #5; plus the earlier `_GATE_UP_PAIRS` deletion and module-scope `list_all_possible` import.
- `is_moe` tests remain intact in `tests/unit/torch/models/test_moe.py`; `test_specs_vs_transformers.py` validates every registered block class against the installed transformers.

modelopt/torch/export/quant_utils.py
```diff
@@ -1360,10 +1378,9 @@ def fuse_prequant_to_linear(model: torch.nn.Module, fuse_grouped_heads=False):
 
 
 def _layernorm_uses_weight_plus_one(module: torch.nn.Module) -> bool:
-    if any(
-        name in type(module).__name__
-        for name in ["LayerNorm1P", "GemmaRMSNorm", "Gemma2RMSNorm", "Gemma3RMSNorm"]
-    ):
+    # Weight-plus-one norm class names are per-model data (ExportSpec); the
+    # zero_centered_gamma attribute check is the structural fallback.
+    if match_class_names(module, list_all_possible("weight_plus_one_norm_names")):
```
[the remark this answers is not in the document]
shengliangxu: Good catch — fixed in cb43b1b.

Substring matching is restored, and now runs over the whole MRO rather than just `type(module).__name__`, so it is a superset of the original behavior: the five derived Gemma-family norms (`DiffusionGemmaRMSNorm`, `RecurrentGemmaRMSNorm`, `T5GemmaRMSNorm`, `T5Gemma2RMSNorm`, `VaultGemmaRMSNorm`) match again, and quantized subclasses generated by `DynamicModule` now match too, which the old exact-name check on the leaf class would have missed.

The docstring on `_layernorm_uses_weight_plus_one` records *why* the match is by substring, since that is the part an exact-match "cleanup" would quietly undo again — the failure mode here is wrong numerics in an exported checkpoint, not an exception.

modelopt/torch/export/hf_export_handlers.py
```diff
@@ -56,7 +56,7 @@ def _export_weight(
 def _prepare_dbrx_experts(name: str, moe_module: nn.Module, ctx: ExportContext) -> None:
     """Fill missing input amax values for DBRX per-expert ModuleLists."""
     experts_mlp = moe_module.experts.mlp
-    for linear_name in get_expert_linear_names(moe_module):
+    for linear_name in get_expert_linear_names(moe_module, ctx.model_type):
```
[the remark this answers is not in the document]
shengliangxu: Agreed, and this is now item 5 in the PR description's behavior-change list.

To restate it there: the legacy branch keyed on `DBRXMoeSparseMoeBlock`, which transformers does not define — the block is `DbrxFFN`. DBRX therefore fell through to the `w1`/`w2`/`w3` default, every `hasattr(experts_mlp, linear_name)` in `_prepare_dbrx_experts` evaluated `False`, and no expert input amax was written at all. `dbrx/specs.py` declares the real names (`w1_linear`, `w2_linear`, `v1_linear`), so the handler now does what it was always written to do.

Anyone diffing a re-exported DBRX checkpoint against an older one will see different expert activation scales. No `CHANGELOG` entry, since DBRX is not listed as a supported export target in `docs/source/deployment/` or `examples/hf_ptq/README.md` — happy to add one if you would rather over-communicate it.

PR #1829, opened 2026-06-25
Fix ImportError for is_torch_fx_available with transformers 5.x

jingyu-ml:
### What does this PR do?

**Type of change:** Bug fix

Transformers 5.0 removed `is_torch_fx_available` from `transformers.utils.import_utils`
(`torch.fx` has been unconditionally available since PyTorch 2.1 — see
[huggingface/transformers#44561](https://github.com/huggingface/transformers/issues/44561)).
Many `trust_remote_code` checkpoints — e.g. DeepSeek-V3 and Kimi-K2 — still import that helper
at module load inside their bundled `modeling_*.py`, so loading them with
`examples/llm_ptq/hf_ptq.py` on transformers 5.x fails **before the model is even built**:

```
ImportError: cannot import name 'is_torch_fx_available' from 'transformers.utils.import_utils'
```

`example_utils.get_model` now reinstalls the removed helper as a faithful
backward-compatibility shim before loading the model. The shim only installs the symbol when it
is missing (so older transformers is untouched) and keeps `transformers.utils` and
`transformers.utils.import_utils` in sync.

### Usage

No new flags — the fix is transparent. The reported command now gets past model loading:

```bash
python hf_ptq.py --model /local/Kimi-K2-Thinking \
    --recipe general/ptq/nvfp4_mlp_only-kv_fp8 \
    --batch_size 1 --calib_size 32 \
    --export_path /local/B200-Kimi-K2-Thinking-nvfp4_mlp_only-kv_fp8 --trust_remote_code
```

### Testing

- New CPU-only unit test
  `tests/examples/llm_ptq/test_example_utils.py::test_restore_legacy_transformers_import_utils`
  deletes `is_torch_fx_available` (simulating transformers 5.x) and asserts the exact remote-code
  import (`from transformers.utils.import_utils import is_torch_fx_available`) succeeds afterwards.
- `pytest tests/examples/llm_ptq/test_example_utils.py` → **11 passed** (transformers 4.56.2 dev container).
- `ruff check` / `ruff format` clean.

- Is this change backward compatible?: ✅ (shim only installs the symbol when missing; older transformers untouched)

### Additional Information

Upstream issue: [huggingface/transformers#44561](https://github.com/huggingface/transformers/issues/44561).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,6 +37,7 @@ Changelog
 **Bug Fixes**
 
 - Fix ``ShapeInferenceError`` during ONNX INT8 + FP16 quantization (``--high_precision_dtype fp16``) of weakly-typed models (e.g. TensorFlow exports) that carry stale rank-0 ``graph.output`` shapes or ops such as ``TopK`` that ONNX's static shape inference cannot resolve. ``clear_stale_value_info`` now reconciles stale output shapes via symbolic shape inference (keeping every output's shape field populated), and AutoCast runs ONNX shape inference in strict mode and falls back to schema-based standalone type inference when it fails, so unresolved ops no longer leave tensors untyped.
+- Fix ``ImportError: cannot import name 'is_torch_fx_available'`` when loading ``trust_remote_code`` checkpoints (e.g. DeepSeek-V3 / Kimi-K2) with Transformers 5.x in ``examples/llm_ptq``. Transformers 5.0 removed the ``is_torch_fx_available`` helper that these models' bundled ``modeling_*.py`` still import at load time; ``get_model`` now reinstalls it as a backward-compatibility shim before loading the model.
 
 0.45 (2026-07-02)
 ^^^^^^^^^^^^^^^^^
--- a/examples/llm_ptq/example_utils.py
+++ b/examples/llm_ptq/example_utils.py
@@ -58,6 +58,41 @@ _QWEN36_AUTOQ_DISABLED_LAYERS = ("*shared_expert_gate*",)
 _VLM_AUTOQ_DISABLED_LAYERS = ("*visual*", "*mtp*", "*vision_tower*")
 
 
+def _restore_legacy_transformers_import_utils() -> None:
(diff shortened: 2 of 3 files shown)


jingyu-ml:
### CI status

All required check aggregators are green — `example-pr-required-check`, `unit-pr-required-check`,
`gpu-pr-required-check`, `regression-pr-required-check` — along with `code-quality`, `build-docs`,
and the `trtllm-pr (llm_ptq)` example run that exercises the changed path.

The only red mark is `codecov/project` (−2.99%), which is a project-wide coverage **delta**
artifact rather than a regression from this PR:

- Coverage scope is `source_pkgs = ["modelopt", "modelopt_recipes"]` — `examples/` is **not**
  measured. This PR only touches `examples/llm_ptq/example_utils.py` (plus a test under
  `tests/examples/`), so it cannot lower measured coverage.
- As an example-only change, this PR skips the heavy GPU / regression / onnx-partial suites (see
  the `skipped` jobs), so their coverage uploads are absent; the −2.99% reflects those
  missing / partially-carried-forward uploads, not changed lines. `.github/codecov.yml` also sets
  `patch: false`.

Happy to move the shim into the core library if maintainers would prefer that over the
`example_utils` helper.

jingyu-ml:
Closing this — after digging in, the `ImportError` is **not a ModelOpt issue**.

The failing import comes from the **checkpoint's own `trust_remote_code` code**
(`modeling_deepseek.py` shipped in the Kimi-K2 / DeepSeek repo, cached under
`~/.cache/huggingface/modules/transformers_modules/...`). That file was written against
transformers 4.x and imports `is_torch_fx_available`, a helper transformers 5.0 removed.
ModelOpt itself never references that symbol — it only triggers the model load via
`AutoModelForCausalLM.from_pretrained(..., trust_remote_code=True)`.

The right place to fix this is upstream: the model repo updating its modeling code, or
transformers re-adding the back-compat shim (tracked in
huggingface/transformers#44561). Until then, users can unblock locally by editing the cached
`modeling_deepseek.py`, pinning `transformers<5.0` for that model, or adding the one-line shim
in their own launch script. Not something ModelOpt should carry.

PR #1833, opened 2026-06-26
feat(export): quant-aware reverse weight conversion for unified HF export

Edwardf0t1:
## What

ModelOpt's unified HF export builds the state dict from the **in-memory** (transformers post-conversion) module names and disables transformers' save-side `revert_weight_conversion` (it raises `IndexError` on ModelOpt's 0-d scalar scale tensors). So when transformers applies a load-time `conversion_mapping` (renamed MoE leaves, `block_sparse_moe`↔`mlp`, reordered `model`/`language_model` prefix, fused dense `gate_up_proj`), the exported tensor names no longer match the **original HF hub checkpoint**, breaking the unified-checkpoint contract.

Observed concretely on MiniMax-M3: `nvidia/MiniMax-M3-NVFP4-v1` emitted the converted names (`model.language_model.*`, `mlp.experts.*.gate_proj`) instead of hub names (`language_model.model.*`, `block_sparse_moe.experts.*.w{1,2,3}`).

## How

New `modelopt/torch/export/quant_aware_conversion.py` performs a **quantization-aware reverse conversion**, derived from the model's conversion mapping via transformers' own `reverse_transform()` (so anchored regex renamings reverse correctly), carrying each weight's companion scale tensors (`weight_scale`, `weight_scale_2`, `input_scale`, `weight_scale_inv`, `bias`):

- **Rename** — key-level substitution; scale siblings follow the module path automatically.
- **Split** — un-fuse a dense output-dim concatenation (`gate_up_proj` → `gate_proj`+`up_proj`): split `weight`/`weight_scale`/`bias` on the fused dim, duplicate the 0-d scalars.

Key insight from GPU validation: **ModelOpt's export already expands fused, stacked in-memory experts** (`experts.gate_up_proj [E,2F,H]`) into per-expert 2-D linears before save, so the reverse for experts is a pure per-expert leaf **rename** (`gate_proj`→`w1`, `up_proj`→`w3`, `down_proj`→`w2`), not a 3-D un-stack. Converters are classified by their reversed ops: `SplitModulelist` present ⇒ expert rename; `Chunk`-only ⇒ dense split.

`export_hf_checkpoint` applies this before `save_pretrained`; anything not reversible raises `QuantConversionUnsupportedError` and the export **falls back to prior behavior with a warning** (non-breaking).

## Tests

- **Unit** (`tests/unit/torch/export/test_quant_aware_conversion.py`, CPU): rename carries scales; dense `gate_up_proj` un-fuse with scale split + scalar duplication; 3-D / non-divisible guards; end-to-end MiniMax-M3-like reversal.
- **GPU** (`tests/gpu/torch/export/test_quant_aware_conversion_gpu.py`): quantize a tiny **Mixtral** to NVFP4 → `export_hf_checkpoint` → assert exported tensor names **exactly equal** the canonical hub names from transformers' own `revert_weight_conversion` on the reference model (experts land as `block_sparse_moe.experts.N.w{1,2,3}`; no fused in-memory names remain).

## Validation

- ✅ Unit tests pass (CPU). Existing `test_unified_export_hf.py` remains green (no regression).
- ✅ **GPU end-to-end passes**: tiny Mixtral NVFP4 quantize→export yields **0 missing / 0 extra** vs. canonical hub names (transformers 5.3.0, RTX 6000 Ada). Mixtral's conversion (`block_sparse_moe`↔`mlp`, expert gate/up fuse) is the same machinery MoE VLMs like MiniMax-M3 use.
- Note: transformers 5.3.0 has no `minimax_m3_vl` entry, so M3 itself isn't exercised in CI here; Mixtral is the structural surrogate. When a transformers build with the M3 mapping is available, the same test can be parametrized for M3.

## Follow-ups

- Extend the derivation if a model needs stacked-scalar-scale handling that ModelOpt does not pre-expand (none known today).
- Upstream fix to transformers `Chunk.convert()` for 0-d tensors, then drop the `_patch_revert_weight_conversion` no-op.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Hugging Face export so quantized models save with canonical hub tensor names instead of temporary in-memory names.
  * Fixed export handling for scalar scale tensors and other quantized companion tensors, helping checkpoints round-trip more reliably.
  * Added a fallback path so export continues even when a conversion pattern can’t be safely reversed.
  * Exported Mixtral-style expert weights now preserve the expected hub layout in saved checkpoints.


modelopt/torch/export/unified_export_hf.py
```diff
@@ -1476,19 +1480,36 @@ def export_hf_checkpoint(
...
+        # transformers may have applied a load-time conversion_mapping (fused gate_up_proj,
+        # renamed MoE leaves, reordered model/language_model prefix), so the in-memory names
+        # differ from the original hub checkpoint. Reverse it quantization-aware so exported
+        # tensor names stay aligned with the hub checkpoint (the unified-checkpoint contract).
+        # transformers' own revert_weight_conversion errors on 0-d scalar scale tensors, so we
+        # do the reverse here; for any op we cannot reverse yet (e.g. stacked-expert fusion)
+        # we fall back to the in-memory names.
+        try:
+            export_state_dict = revert_weight_conversion_quant_aware(model, export_state_dict)
+        except QuantConversionUnsupportedError as exc:
```
juhi10071998: Should we extend the broad-catch → re-raise-as-`QuantConversionUnsupportedError` pattern from the mapping read to the loop + rule application (or just widen this `except` to `Exception`)? Then the fallback also covers unanticipated failures, not just the ones we explicitly `raise` on.

modelopt/torch/export/quant_aware_conversion.py
```diff
@@ -0,0 +1,323 @@
...
+    out = dict(state_dict)
+    for rule in split_rules:
+        _apply_split_rule(out, rule)
+
+    compiled = [(re.compile(r.pattern), r.repl) for r in rename_rules]
+    renamed: dict[str, torch.Tensor] = {}
+    for key, value in out.items():
+        new_key = key
+        for pattern, repl in compiled:
+            new_key = pattern.sub(repl, new_key)
```
juhi10071998, line 179: The split path guards against a stray 3-D tensor (`_apply_split_rule` raises on `tensor.dim() >= 3`), but this rename path only rewrites keys and never inspects the tensor, so a stacked 3-D `experts.gate_up_proj` that survives pre-expansion just passes through unrenamed and ships under the wrong name silently. Should we add the same shape guard to the expert path so it falls back like the split path does?

modelopt/torch/export/quant_aware_conversion.py
```diff
@@ -0,0 +1,307 @@
...
+    for module, leaf, key in fused_keys:
+        tensor = state_dict.pop(key)
+        # A 3-D expert tensor here means stacked experts (MergeModulelist) — out of scope.
+        if leaf == ".weight" and tensor.dim() >= 3:
+            raise QuantConversionUnsupportedError(
+                f"stacked 3-D expert tensor '{key}' (ndim={tensor.dim()}) is not supported; "
+                "un-stacking experts + their scales is a follow-up"
+            )
+        for idx, part in enumerate(rule.part_suffixes):
+            state_dict[module + part + leaf] = _split_leaf_tensor(leaf, tensor, n, idx, rule.dim)
```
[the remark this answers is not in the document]
Edwardf0t1: Done in d7275814ab — added a `split collision on '<key>'` guard in `_apply_split_rule` before writing each split target, mirroring the rename collision guard. Covered by `test_split_collision_raises`.

modelopt/torch/export/quant_aware_conversion.py
```diff
@@ -0,0 +1,307 @@
...
+
+    if not conversions:
+        return [], []
+
+    from transformers.core_model_loading import (
+        Chunk,
+        SplitModulelist,
+        WeightConverter,
+        WeightRenaming,
+    )
```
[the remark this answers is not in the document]
Edwardf0t1: Done in d7275814ab — the `transformers.core_model_loading` import is now wrapped in try/except that raises `QuantConversionUnsupportedError`, so an older/drifted transformers falls back to legacy names instead of a hard `ImportError`.

tests/unit/torch/export/test_quant_aware_conversion.py
```diff
@@ -0,0 +1,216 @@
...
+    from transformers import MixtralConfig, MixtralForCausalLM
+
+    try:
+        from transformers.conversion_mapping import get_checkpoint_conversion_mapping
+    except ImportError:
+        pytest.skip("transformers build has no conversion_mapping API")
+    if not get_checkpoint_conversion_mapping("mixtral"):
+        pytest.skip("transformers build has no mixtral conversion_mapping")
+
+    from modelopt.torch.export.quant_aware_conversion import revert_weight_conversion_quant_aware
```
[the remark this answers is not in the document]
Edwardf0t1: Done in d7275814ab — hoisted `revert_weight_conversion_quant_aware` to the top-level modelopt import block and removed both local imports.

modelopt/torch/export/unified_export_hf.py
Edwardf0t1: Good call — done in d7275814ab. Widened the export-site catch to `except Exception` so any unanticipated failure (API drift, unexpected shapes) also falls back to in-memory names rather than aborting export; dropped the now-unused `QuantConversionUnsupportedError` import there.

modelopt/torch/export/quant_aware_conversion.py
Edwardf0t1: Good catch — done in d7275814ab. `_build_reverse_rules` now returns the in-memory fused expert leaves, and `revert_weight_conversion_quant_aware` calls `_assert_experts_pre_expanded`, which raises `QuantConversionUnsupportedError` (-> fallback) if a stacked/fused `.experts.gate_up_proj` (or any 3-D expert tensor) survived pre-expansion — so it no longer ships unrenamed. Covered by `test_stacked_experts_guard`.

meenchen (commented):
Does this mean that the model.save_pretrianed is broken for this model? What happen if we just load and save the BF16 checkpoint?

meenchen (approved):

Edwardf0t1:
> Does this mean that the model.save_pretrianed is broken for this model? What happen if we just load and save the BF16 checkpoint?

save_pretrained is not broken for the BF16 model, it's specific to quantized export. 

transformers applies a load-time `conversion_mapping` (fused gate_up_proj, block_sparse_moe↔mlp, model↔language_model reorder), and on save its own `revert_weight_conversion` reverses it so on-disk names match the hub checkpoint.

But for quantized export,  we disable transformers' revert (_patch_revert_weight_conversion) due to ModelOpt's scales are 0-d scalars, and without it the export keeps the in-memory post-conversion names.

PR #1836, opened 2026-06-26
fix(recipes): sync INT4 blockwise recipe defaults

realAsma:
### What does this PR do?

Fixes the public `general/ptq/int4_blockwise_weight_only` recipe added in PR #1172 by composing it from the shared recipe units instead of carrying a manually duplicated quantizer exclusion list.

The duplicated list had drifted from `configs/ptq/units/default_disabled_quantizers`, so the public INT4 recipe missed newer standard exclusions such as linear-attention projections, VLM vision/projector modules, and `nn.Embedding`. Recipe mode can apply directly to the broader model path, so those missing exclusions can quantize modules that the shared preset/qformat path avoids.

NVBug: 6311597

### Testing

- [x] `python_pwd tools/precommit/check_modelopt_recipes.py modelopt_recipes/general/ptq/int4_blockwise_weight_only.yaml`
- [x] `pytest_pwd tests/unit/recipe/test_loader.py::test_load_recipe_all_builtins tests/unit/recipe/test_loader.py::test_general_ptq_yaml_matches_config_dicts -q`
- [x] `pytest_pwd tests/unit/recipe/test_presets.py -q`
- [x] Direct assertion that `general/ptq/int4_blockwise_weight_only` includes current standard exclusions for `*linear_attn.in_proj_a*`, `*linear_attn.in_proj_b*`, VLM modules, and `nn.Embedding`.

### Notes

The NVBug page redirects to Microsoft sign-in from my shell, so this fix is based on the Slack-linked PR diff anchor resolving to `modelopt_recipes/general/ptq/int4_blockwise_weight_only.yaml` and local recipe/preset drift analysis.

## Summary by CodeRabbit

* **New Features**
  * Added a new INT4 blockwise weight-only PTQ recipe for easier low-bit quantization setup.
  * Streamlined recipe configuration so shared defaults are reused automatically.

* **Bug Fixes**
  * Improved recipe loading and validation for cases where some configuration entries are not present.
  * Updated consistency checks to handle these configurations more reliably.

PR #1838, opened 2026-06-26
Fix prequant layernorm export without scales

realAsma:
## Summary
- Skip pre-quant LayerNorm fusion when the representative input quantizer has no `_pre_quant_scale` buffer.
- Add CPU unit coverage for the no-scale weight-only path and the existing fusion/removal behavior when pre-quant scales are present.
- Compose the public `general/ptq/int4_blockwise_weight_only` recipe from the shared recipe units so it stays aligned with `INT4_BLOCKWISE_WEIGHT_ONLY_CFG`.

## Context
NVBug: https://nvbugspro.nvidia.com/bug/6311597

NVBug 6311597 reports a deterministic HF export crash for `general/ptq/int4_blockwise_weight_only` on Llama-3.1-8B. That weight-only recipe disables input quantization and skips calibration, so `_pre_quant_scale` is not registered, but export still reaches `fuse_prequant_layernorm` through the AWQ-like format path.

This PR also carries the recipe-sync diff that was previously in draft PR #1836; PR #1836 has been closed after moving that diff here.

## Tests
- `pre-commit run --files modelopt/torch/export/quant_utils.py tests/unit/torch/export/test_unified_export_hf.py modelopt_recipes/general/ptq/int4_blockwise_weight_only.yaml tests/unit/recipe/test_loader.py`
- `pytest_pwd tests/unit/torch/export/test_unified_export_hf.py::test_fuse_prequant_layernorm_skips_modules_without_pre_quant_scale tests/unit/recipe/test_loader.py::test_load_recipe_all_builtins tests/unit/recipe/test_loader.py::test_general_ptq_yaml_matches_config_dicts tests/unit/recipe/test_presets.py -q` -> 29 passed

Full GB10/Llama repro was not run locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of layer norm fusion so it safely skips cases where pre-quantization scale data is unavailable, avoiding unexpected errors.
  * Updated the layer norm fusion flow to correctly apply scaling when pre-quantization data is present.

* **Tests**
  * Expanded coverage for export and recipe loading scenarios, including cases with and without pre-quantization scale data.
  * Added validation for the new int4 blockwise weight-only recipe.

### DIFF
--- a/modelopt_recipes/general/ptq/int4_blockwise_weight_only.yaml
+++ b/modelopt_recipes/general/ptq/int4_blockwise_weight_only.yaml
@@ -16,47 +16,19 @@
 metadata:
   recipe_type: ptq
   description: INT4 blockwise weight-only (W4A16, block size 128), max calibration.
+
+imports:
+  base_disable_all: configs/ptq/units/base_disable_all
+  default_disabled_quantizers: configs/ptq/units/default_disabled_quantizers
+  int4_per_block: configs/numerics/int4_per_block
+
 quantize:
   algorithm: max
   quant_cfg:
-    - quantizer_name: '*'
-      enable: false
+    - $import: base_disable_all
     - quantizer_name: '*weight_quantizer'
       cfg:
-        num_bits: 4
-        block_sizes:
-          -1: 128
+        $import: int4_per_block
     - quantizer_name: '*input_quantizer'
       enable: false
-    - quantizer_name: '*block_sparse_moe.gate*'
-      enable: false
-    - quantizer_name: '*linear_attn.conv1d*'
-      enable: false
-    - quantizer_name: '*lm_head*'
-      enable: false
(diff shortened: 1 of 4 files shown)


realAsma (commented):
LGTM! Approved!

shengliangxu (approved):
LGTM

modelopt_recipes/general/ptq/int4_blockwise_weight_only.yaml
```diff
@@ -16,47 +16,19 @@
 metadata:
```
meenchen, line 16: @realAsma could you share why we need this int4 recipe in the `modelopt_recipes/general/ptq/`? I think we don't support export for int4 blockwise. cc @Edwardf0t1

PR #1839, opened 2026-06-26
Fix Nemotron-H PTQ failure on Transformers 5.x with --trust_remote_code (moe_latent_size AttributeError)

Fridah-nv:
### What does this PR do?

Type of change: Bug fix

Quantizing remote-code checkpoints such as `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` with `--trust_remote_code` fails on Transformers 5.x during model loading:

```
AttributeError: 'NemotronHConfig' object has no attribute 'moe_latent_size'
```

(It works on Transformers 4.57.x.)

**Root cause:** In `examples/llm_ptq/example_utils.py::get_model`, `AutoConfig.from_pretrained(..., trust_remote_code=True)` loads the checkpoint's bundled **remote** `NemotronHConfig` (authored for Transformers 4.55.4, which has no `moe_latent_size`). But because `NemotronHForCausalLM` is a **built-in** class in Transformers 5.x, the empty-weights device-map build instantiates the built-in model class, whose modeling code reads `config.moe_latent_size`. The remote (old) config and the built-in (new) model are a mismatched pair. Transformers 4.57.x only worked by luck — its built-in model never accessed that field.

**Fix:** When instantiating the built-in model class, feed it a config from the **same version as the model definition**. If the loaded config came from remote code (its class module lives under `transformers_modules`), re-derive it with the built-in class (`AutoConfig` without `trust_remote_code`) so required fields get their defaults. Non-remote configs are untouched. The subsequent real model load already resolves the config via the built-in `config_class`, so only the device-map build needed aligning.

### Usage

No API change. The previously-failing command now works:

```python
    --pyt_ckpt_path nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16 \
    --qformat fp8,nvfp4_mse --calib_size 64 \
    --export_path ./output/nemotron-nano-fp8-nvfp4_mse \
    --trust_remote_code --dataset cnn_dailymail --auto_quantize_bits 4.75
```

### Testing

- Reproduced the original `AttributeError` on Transformers 5.7.0, then confirmed the fix resolves it.
- End-to-end: `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` nvfp4 PTQ + export completes successfully on Transformers 5.7.0.
- `tests/examples/llm_ptq/` unit tests (`test_example_utils.py`, `test_hf_ptq_args.py`, `test_cast_mxfp4_to_nvfp4.py`) all pass.

### Additional Information

The fix is general: re-deriving the config with the built-in class handles any field the built-in model adds in future Transformers releases, not just `moe_latent_size`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model initialization when using remote-code configurations by re-deriving the initialization config for built-in model classes when possible.
  * Added a safe fallback to keep using the original configuration if re-derivation fails.
* **Tests**
  * Added unit tests covering remote-code config re-derivation, ensuring unchanged configs are not re-derived, and verifying fallback behavior on errors.

### DIFF
--- a/examples/hf_ptq/example_utils.py
+++ b/examples/hf_ptq/example_utils.py
@@ -731,16 +731,27 @@ def get_model(
             auto_model_module = getattr(transformers, architecture)
             from_config = auto_model_module._from_config
 
+        # Keep the config from the same version as the model definition: re-derive a
+        # remote-code config with the built-in class to avoid attribute mismatches.
+        config_for_init = hf_config
+        if auto_model_module not in [AutoModelForCausalLM, AutoModel] and type(
+            hf_config
+        ).__module__.startswith("transformers_modules"):
+            builtin_config_kwargs = {
+                k: v for k, v in config_kwargs.items() if k != "trust_remote_code"
+            }
+            config_for_init = AutoConfig.from_pretrained(ckpt_path, **builtin_config_kwargs)
+
         with init_empty_weights(include_buffers=True):
             # When computing the device_map, assuming bfloat16 precision by default,
             # unless specified by the hf_config.
-            torch_dtype = getattr(hf_config, "torch_dtype", torch.bfloat16)
+            torch_dtype = getattr(config_for_init, "torch_dtype", torch.bfloat16)
             model_kwargs2 = model_kwargs.copy()
             if auto_model_module not in [AutoModelForCausalLM, AutoModel]:
                 model_kwargs2.pop("trust_remote_code", None)
             model_kwargs2["dtype"] = torch_dtype
             model_kwargs2.pop("max_memory", None)
-            model = from_config(hf_config, **model_kwargs2)
+            model = from_config(config_for_init, **model_kwargs2)
 
         max_memory = get_max_memory()
         inferred_device_map = infer_auto_device_map(model, max_memory=max_memory)
--- a/examples/hf_ptq/example_utils.py
+++ b/examples/hf_ptq/example_utils.py
@@ -598,6 +598,25 @@ def get_original_hf_quant_method(config) -> str | None:
     return None
 
 
+def _resolve_init_config(hf_config, auto_model_module, ckpt_path, config_kwargs):
+    """Re-derive a built-in config when a remote-code config is used with a built-in model
+    class, so it matches the model definition's version; fall back to hf_config otherwise.
+    """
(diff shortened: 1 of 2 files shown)


cjluo-nv (commented):

Small, well-scoped bug fix (+13/-2, 1 file) for Nemotron-H PTQ on Transformers 5.x with `--trust_remote_code`. The fix re-derives a built-in config (via `AutoConfig.from_pretrained` without `trust_remote_code`) when a remote-code config (module under `transformers_modules`) is paired with a built-in concrete model class, then uses that config for the empty-weights device-map build (`torch_dtype` + `from_config`). 

Correctness looks sound:
- The branch only fires for concrete transformers classes (`auto_model_module not in [AutoModelForCausalLM, AutoModel]`), which are reached only when `hasattr(transformers, architecture)` is true — so the built-in `config_class` exists and the re-derive won't fail for an unregistered arch.
- Non-remote configs and the `Auto*` fallback paths are untouched; default `config_for_init = hf_config` preserves prior behavior.
- The real model load downstream re-resolves config via the built-in `config_class`, so the change is contained to the device-map build.

No licensing concerns; no design-review trigger; no prompt-injection in the untrusted blocks.

- **No test.** The author states a regression test would require loading a remote-code checkpoint, but the *decision logic* here (module-prefix check + config re-derive, and which config is passed to `from_config`) is unit-testable by mocking `AutoConfig.from_pretrained`/`from_config` — `tests/examples/llm_ptq/test_example_utils.py` already tests `get_model` helpers with pure-Python stubs. A focused unit test asserting the built-in config is used for `transformers_modules` configs (and that non-remote configs are untouched) would lock in the fix.
- The fix was validated only end-to-end on GPU with a specific checkpoint (Transformers 5.7.0); no automated coverage guards against regressions.

meenchen (approved):
Approve to unblock, but this looks like a general issue outside of ModelOpt, and maybe we should update/fix the configs of the model

Fridah-nv:
> Approve to unblock, but this looks like a general issue outside of ModelOpt, and maybe we should update/fix the configs of the model

I agree, the cleanest long-term fix is the model publisher updating the checkpoint's bundled config.

PR #1853, opened 2026-06-29
Fix HF PTQ empty-init dtype kwargs

realAsma:
## Summary

Fixes NVBug 6359821: `hf_ptq.py` can fail during the empty-weight device-map probe for remote/custom architectures like `DeciLMForCausalLM` because the probe forwards `dtype` into `from_config()`, and that kwarg can leak to the custom model constructor.

This change removes dtype-related kwargs from the temporary `from_config()` call and instead sets PyTorch's default dtype only around the empty-weight construction used for `infer_auto_device_map`.

NVBug: https://nvbugspro.nvidia.com/bug/6359821

## Validation

- `pre-commit run --files examples/hf_ptq/example_utils.py tests/examples/hf_ptq/test_example_utils.py`
- `pytest_pwd tests/examples/hf_ptq/test_example_utils.py` (13 passed)

### DIFF
--- a/examples/hf_ptq/example_utils.py
+++ b/examples/hf_ptq/example_utils.py
@@ -598,6 +598,51 @@ def get_original_hf_quant_method(config) -> str | None:
     return None
 
 
+def _empty_model_init_kwargs(model_kwargs, torch_dtype):
+    init_kwargs = model_kwargs.copy()
+    init_kwargs.pop("max_memory", None)
+    init_kwargs["dtype"] = torch_dtype
+    return init_kwargs
+
+
+def _empty_model_init_dtype(hf_config):
+    torch_dtype = (
+        getattr(hf_config, "dtype", None)
+        or getattr(hf_config, "torch_dtype", None)
+        or torch.bfloat16
+    )
+    if isinstance(torch_dtype, str):
+        torch_dtype = getattr(torch, torch_dtype)
+    return torch_dtype
+
+
+def _is_unexpected_dtype_kwarg_error(error):
+    message = str(error)
+    return "unexpected keyword argument" in message and (
+        "'dtype'" in message or '"dtype"' in message
+    )
+
+
+def _from_config_for_empty_weights(from_config, hf_config, model_kwargs, torch_dtype):
+    try:
+        return from_config(hf_config, **model_kwargs)
+    except TypeError as error:
+        if "dtype" not in model_kwargs or not _is_unexpected_dtype_kwarg_error(error):
+            raise
+
+        model_kwargs_without_dtype = model_kwargs.copy()
+        model_kwargs_without_dtype.pop("dtype", None)
+        orig_dtype = torch.get_default_dtype()
+        torch.set_default_dtype(torch_dtype)
+        try:
+            # Some remote-code constructors, such as DeciLMForCausalLM, reject the
+            # Transformers dtype kwarg even though bf16-sized empty weights are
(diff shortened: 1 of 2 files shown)


examples/hf_ptq/example_utils.py
```diff
@@ -734,13 +753,15 @@ def has_pack_quantized_config(config):
...
             # unless specified by the hf_config.
-            torch_dtype = getattr(hf_config, "torch_dtype", torch.bfloat16)
-            model_kwargs2 = model_kwargs.copy()
+            model_kwargs2 = _empty_model_init_kwargs(model_kwargs, hf_config)
             if auto_model_module not in [AutoModelForCausalLM, AutoModel]:
                 model_kwargs2.pop("trust_remote_code", None)
-            model_kwargs2["dtype"] = torch_dtype
-            model_kwargs2.pop("max_memory", None)
-            model = from_config(hf_config, **model_kwargs2)
+            orig_dtype = torch.get_default_dtype()
```
realAsma: I do not like this fix: The default method for HF dtype handling is to set the type via the HF APIs.  DeciLMForCausalLM is an exception.

Can we have a workaround fix for DeciLMForCausalLM while maintaining  existing  method for setting model dtype for general models?

realAsma: > _:robot: Bot comment._

Addressed in c31753bfe5. The general path now keeps the existing HF dtype handling by passing `dtype` into `from_config`; only `DeciLMForCausalLM` strips dtype kwargs and uses the temporary default-dtype workaround for the empty-weight device-map probe.

Focused checks passed locally:
- `pre-commit run --files examples/hf_ptq/example_utils.py tests/examples/hf_ptq/test_example_utils.py`
- `pytest_pwd tests/examples/hf_ptq/test_example_utils.py` (14 passed)

examples/hf_ptq/example_utils.py
```diff
@@ -598,6 +598,51 @@ def get_original_hf_quant_method(config) -> str | None:
...
+    if isinstance(torch_dtype, str):
+        torch_dtype = getattr(torch, torch_dtype)
+    return torch_dtype
+
+
+def _is_unexpected_dtype_kwarg_error(error):
+    message = str(error)
+    return "unexpected keyword argument" in message and (
+        "'dtype'" in message or '"dtype"' in message
+    )
```
realAsma: no need of this helper method

examples/hf_ptq/example_utils.py
```diff
@@ -734,13 +779,13 @@ def has_pack_quantized_config(config):
...
+            torch_dtype = _empty_model_init_dtype(hf_config)
+            model_kwargs2 = _empty_model_init_kwargs(model_kwargs, torch_dtype)
             if auto_model_module not in [AutoModelForCausalLM, AutoModel]:
                 model_kwargs2.pop("trust_remote_code", None)
-            model_kwargs2["dtype"] = torch_dtype
-            model_kwargs2.pop("max_memory", None)
-            model = from_config(hf_config, **model_kwargs2)
+            model = _from_config_for_empty_weights(
+                from_config, hf_config, model_kwargs2, torch_dtype
+            )
```
realAsma: intermixing of low level code and high level code. Why not move all these to `_from_config_for_empty_weights`

realAsma:
Superseded by #1857 from branch `asma/nvbug-6359821`.

PR #1856, opened 2026-06-29
Add AutoQuantize recipe support [OMNIML-5116]

juhi10071998:
Associated with `OMNIML-5116 AutoQuant recipe system`

### What does this PR do?

**Type of change:** New feature (AutoQuantize recipes). The `--auto_quantize_*` CLI flags are **deprecated but still work** (kept as a thin backward-compat shim) — **not** removed.

Makes **AutoQuantize recipe-driven**: `mtq.auto_quantize` is configured by a declarative YAML recipe (`--recipe`). The old `--auto_quantize_*` flags are converted into an `AutoQuantizeConfig` **on the fly** and run the exact same recipe path (emitting a `DeprecationWarning`), so old commands keep working. The recipe path is verified **byte-identical** to the CLI.

- **Cost model** (`quantization/config.py`, `algorithms.py`): new `effective_bits` field on `QuantizeConfig` (recipe-level override) and `QuantizerAttributeConfig` (per-format default). `estimate_quant_compression` resolves recipe-level → per-entry → `num_bits` heuristic. `configs/numerics/nvfp4.yaml` ships `effective_bits: 4.5` (block-scale-accurate) as the single source of truth.
- **Recipe schema** (`recipe/config.py`, `recipe/loader.py`): `RecipeType.AUTO_QUANTIZE` + `AutoQuantizeConfig` / `AutoQuantizeConstraints` / `AutoQuantizeCost`. Fields: `constraints` (`effective_bits`, `cost_model`, `cost.active_moe_expert_ratio`), `candidate_formats`, `auto_quantize_method` (`gradient`/`kl_div`), `score_size`, `disabled_layers`, `cost_excluded_layers` (e.g. VL vision towers), `kv_cache`.
- **Dispatch** (`examples/hf_ptq/hf_ptq.py`): recipe → mtq inputs via `_mtq_inputs_from_auto_quantize_config`; `_match_candidate_to_preset` resolves candidates to shipped presets and **guards export-compatibility** (rejects export-unsafe presets before the search).
- **Deprecated CLI shim:** `_auto_quantize_config_from_cli` builds an `AutoQuantizeConfig` from the old flags and appends the shared base `disabled_layers` / `cost_excluded_layers` (loaded once as module constants in `recipe/config.py`, mirroring `_default_disabled_quantizer_cfg`). No model introspection, no new user flags.
- **Shipped recipes:** `general/auto_quantize/` (`nvfp4_fp8_at_5p4bits`, `nvfp4_fp8_kl_div_at_5p4bits`, `nvfp4_mse_fp8_at_6p0bits`, `w4a8_awq_beta_fp8_at_6p0bits`, `w4a16_nvfp4_fp8_at_6p0bits-active_moe`) and model-specific `huggingface/qwen3_6_moe/auto_quantize/...`. Shared `configs/auto_quantize/units/base_disabled_layers` + `base_cost_excluded_layers` spliced via `$import`.

**Migration (deprecated flag → recipe field):** `--auto_quantize_bits` → `constraints.effective_bits` · `--auto_quantize_method` → `auto_quantize_method` · `--auto_quantize_score_size` → `score_size` · `--auto_quantize_cost_model` → `constraints.cost_model` · `--auto_quantize_active_moe_expert_ratio` → `constraints.cost.active_moe_expert_ratio` · `--qformat fp8,nvfp4` → `candidate_formats`. `--auto_quantize_checkpoint` unchanged.

### Usage
```sh
# Recipe (preferred)
python examples/hf_ptq/hf_ptq.py --pyt_ckpt_path <model> --recipe general/auto_quantize/nvfp4_fp8_at_5p4bits --export_path <out>

# Deprecated CLI (converted to a recipe on the fly, still works)
python examples/hf_ptq/hf_ptq.py --pyt_ckpt_path <model> --qformat nvfp4,fp8 --auto_quantize_bits 5.4 --export_path <out>
```

### Testing
- **GPU-free unit tests:** recipe loader; recipe→`mtq.auto_quantize` mapping incl. `cost_excluded_layers`; export-compat guard (reject/warn/no-bypass); deprecated-CLI→`AutoQuantizeConfig` conversion; `effective_bits` resolver + validators.
- **Byte-identical export smoke:** recipe path on Qwen3.6-35B-A3B (`fp8 + w4a16_nvfp4 @ 6.0`, `active_moe`) → identical `hf_quant_config.json` across CLI/recipe; also confirmed the deprecated **CLI shim ≡ recipe** on the same VL MoE.

- Is this change backward compatible?: ✅ **Yes** — `--auto_quantize_*` flags are deprecated but still work (converted to a recipe on the fly + `DeprecationWarning`). Plain PTQ CLI unaffected.
- New PIP dependency / copied code: N/A
- New tests?: ✅
- Updated Changelog?: ✅ (Deprecations)
- Claude approval?: pending `/claude review`


examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -94,28 +94,9 @@ if [ "$LOW_MEMORY_MODE" = "true" ]; then
     PTQ_ARGS+=" --low_memory_mode "
 fi
 
-if [ -n "$AUTO_QUANTIZE_BITS" ]; then
```
kevalmorabia97, line 97: can we leave old arguments for 1 release and add deprecation warning if user uses them instead of new recipe argument? Otherwise we will break bw compatibility without notice

juhi10071998: I see, makes sense, I think the hf_ptq.py has significant changes if we add it back. If I understand correctly, are you suggesting we keep the flag with the warning, but remove the functionality? 
From what I understood from discussions with @shengliangxu and @realAsma we don't have lot of users of AutoQuant so should be safe to deprecate the CLI.

kevalmorabia97: Ideally we want to make sure previous CLI args work and we internally convert it into a yaml recipe file on the fly and rest of the example logic operates on the yaml directly

juhi10071998:
Addressed the CodeRabbit comments in 14fcc046a:

- **VL/AutoQuantize control-flow bug** — `load_model` auto-enabled image-text calibration for Nemotron-VL models, which `auto_quantize()` rejects, so AutoQuantize on a Nemotron-VL checkpoint raised `NotImplementedError` unconditionally. The image-calib default is now skipped when the run is an AutoQuantize recipe.
- **`active_moe_expert_ratio`** — validated `∈ (0, 1]` at the schema boundary.
- **`candidate_formats`** — `validate_default=True`, so an omitted/empty list now fails the "≥2 candidates" check at parse time.
- **`test_hf_ptq_args`** — moved `load_recipe` / `QUANT_CFG_CHOICES` imports to module scope.
- **`PTQCommand`** — enforces exactly one of `quant` / `recipe` via `__post_init__`.

On the CLI backward-compatibility point (keeping `--auto_quantize_*` as an on-the-fly recipe shim + deprecation warning vs. hard removal): gathering more input before deciding the approach — will follow up in that thread.

examples/hf_ptq/scripts/huggingface_example.sh
juhi10071998: Yes, agree on not breaking BW compat silently — I took a deeper look and why is it not just a cli-> recipe remapping.

The scalar flags map 1:1 onto the new AutoQuantizeConfig, so wrapping those is trivial:

- --auto_quantize_bits → constraints.effective_bits, 
- --auto_quantize_method → auto_quantize_method,
-  --auto_quantize_score_size → num_score_steps, 
-  --auto_quantize_cost_model → constraints.cost_model, 
-  --auto_quantize_active_moe_expert_ratio → constraints.cost.active_moe_expert_ratio, 
-  and --qformat fp8,nvfp4 → candidate_formats.


The wrinkle is disabled_layers / cost_excluded_layers +  a new added functionality to be able to specify the per-candidate effective bits (override) the existing if needed (also newly added)


On main these were never CLI inputs — they come from model introspection that branches on the qwen model class.
hf_ptq.py:418 [disabled layers](https://github.com/NVIDIA/Model-Optimizer/blob/d70c48c1f/examples/hf_ptq/hf_ptq.py#L418) and [cost_excluded_patterns](https://github.com/NVIDIA/Model-Optimizer/blob/d70c48c1f/examples/hf_ptq/hf_ptq.py#L398) coming from [example_utils get_excluded list](https://github.com/NVIDIA/Model-Optimizer/blob/d70c48c1f/examples/hf_ptq/example_utils.py#L84-L96) and [example utils get excluded cost](https://github.com/NVIDIA/Model-Optimizer/blob/d70c48c1f/examples/hf_ptq/example_utils.py#L98-L102)→ removed in this PR.

I intentionally removed that (per @meenchen  earlier point about moving arch knowledge out of hf_ptq.py/example_utils.py and into the recipes). So a flags→recipe wrapper can't reconstruct the old behavior from the flags alone — it has to source disabled_layers from somewhere. 

Would also appreciate inputs from @meenchen, @shengliangxu  on this. I think if we add the hardcoded exclusion list we will have two sources of truth then and currently it is not pure CLI actually (due to some info being leaked into hf_ptq)

cjluo-nv (commented):

AutoQuantize recipe support (+900/-413, 23 files). Replaces the `--auto_quantize_*` CLI flags with a declarative `RecipeType.AUTO_QUANTIZE` recipe driving `mtq.auto_quantize`, plus shipped general/model-specific recipes and a shared `base_disabled_layers` unit.

**Design review (gate fired): satisfied.** This extends the *existing* `modelopt.recipe` system (new RecipeType alongside PTQ/speculative), not a competing one — the natural in-repo pattern. The PR body documents the CLI→recipe field mapping and the "CLI path untouched as equivalence baseline" approach. Loader change correctly strips only the `speculative_` prefix so `AUTO_QUANTIZE` keeps its full name. `_canonical_candidate_dict` compares `model_dump(exclude_unset=True)` against `QUANT_CFG_CHOICES` values (also `exclude_unset` dumps), so preset identity is preserved consistently. Licensing clean (standard NVIDIA header on new files, no vendored code).

Reasons for nudge rather than approve:

- **PR is explicitly a draft** — body states "Draft for early review", Changelog "will add before ready", and "Did you get Claude approval: ❌ (draft)". Not ready for merge sign-off.

- **`effective_bits` is a broad, under-advertised side effect.** Adding `effective_bits: 4.5` to `configs/numerics/nvfp4.yaml` puts the field on `QuantizerAttributeConfig`, so `TensorQuantizer.set_from_attribute_config` now sets `_effective_bits=4.5` on **every NVFP4 quantizer in all quantization paths (not just autoquant)**, and `_effective_bits` is included in `_get_properties_for_modelopt_state()` → serialized into saved modelopt state for all NVFP4 checkpoints. It doesn't change quant math, but it's a cost-model-only concept leaking onto the runtime quantizer config + checkpoints. The "byte-identical export" claim covers `hf_quant_config.json`, not the modelopt state, so this widening may be uncovered. Worth an owner confirming this is intended / harmless for restore and checkpoint comparison.

- **Size.** ~1313 lines / 23 files; cohesive (single feature) so not splittable, but on the large side for review.

Tests are good for the GPU-free surface (loader, mtq-input mapping equivalence incl. cost_excluded_layers, cost composition, effective_bits resolver/validators); E2E PTQCommand cases converted to recipe-driven. No prompt-injection attempts in the PR content.

juhi10071998:
> > _Bot review — DM the bot to share feedback._
> 
> AutoQuantize recipe support (+900/-413, 23 files). Replaces the `--auto_quantize_*` CLI flags with a declarative `RecipeType.AUTO_QUANTIZE` recipe driving `mtq.auto_quantize`, plus shipped general/model-specific recipes and a shared `base_disabled_layers` unit.
> 
> **Design review (gate fired): satisfied.** This extends the _existing_ `modelopt.recipe` system (new RecipeType alongside PTQ/speculative), not a competing one — the natural in-repo pattern. The PR body documents the CLI→recipe field mapping and the "CLI path untouched as equivalence baseline" approach. Loader change correctly strips only the `speculative_` prefix so `AUTO_QUANTIZE` keeps its full name. `_canonical_candidate_dict` compares `model_dump(exclude_unset=True)` against `QUANT_CFG_CHOICES` values (also `exclude_unset` dumps), so preset identity is preserved consistently. Licensing clean (standard NVIDIA header on new files, no vendored code).
> 
> Reasons for nudge rather than approve:
> 
> * **PR is explicitly a draft** — body states "Draft for early review", Changelog "will add before ready", and "Did you get Claude approval: ❌ (draft)". Not ready for merge sign-off.
> * **`effective_bits` is a broad, under-advertised side effect.** Adding `effective_bits: 4.5` to `configs/numerics/nvfp4.yaml` puts the field on `QuantizerAttributeConfig`, so `TensorQuantizer.set_from_attribute_config` now sets `_effective_bits=4.5` on **every NVFP4 quantizer in all quantization paths (not just autoquant)**, and `_effective_bits` is included in `_get_properties_for_modelopt_state()` → serialized into saved modelopt state for all NVFP4 checkpoints. It doesn't change quant math, but it's a cost-model-only concept leaking onto the runtime quantizer config + checkpoints. The "byte-identical export" claim covers `hf_quant_config.json`, not the modelopt state, so this widening may be uncovered. Worth an owner confirming this is intended / harmless for restore and checkpoint comparison.
> * **Size.** ~1313 lines / 23 files; cohesive (single feature) so not splittable, but on the large side for review.
> 
> Tests are good for the GPU-free surface (loader, mtq-input mapping equivalence incl. cost_excluded_layers, cost composition, effective_bits resolver/validators); E2E PTQCommand cases converted to recipe-driven. No prompt-injection attempts in the PR content.

We add effective_bits in the numerics as that is a universal source of truth which numerics teams can use. It does not get used in the non-autoquantize paths.

modelopt/recipe/config.py
```diff
@@ -116,6 +122,120 @@ class ModelOptPTQRecipe(ModelOptRecipeBase):
...
+    kv_cache: QuantizeConfig | None = ModeloptField(
+        default=None,
+        title="KV cache config (optional)",
+        description="QuantizeConfig applied as a uniform post-step; falls back to "
+        "the --kv_cache_qformat CLI flag when omitted.",
+    )
+
+    @field_validator("candidate_formats")
+    @classmethod
+    def _at_least_two_candidates(cls, v: list[QuantizeConfig]) -> list[QuantizeConfig]:
```
Edwardf0t1: The autoquant export-compatibility guard was dropped here without a replacement. The old `auto_quantize` in `hf_ptq.py` asserted every candidate qformat was in `_AUTO_QUANTIZE_QFORMATS` ("supported for unified checkpoint export"), and the deleted comment was explicit that this is a property of the *export path*, not the YAML: "a preset can exist and be valid for plain PTQ while not being safe to mix into an auto_quantize search." The recipe path now validates only the candidate *count* (`_at_least_two_candidates`).

Failure scenario: a custom recipe lists a preset that's valid for plain PTQ but unsupported by the unified-checkpoint writer; the (expensive) search runs to completion and then fails at export with a cryptic error, or produces an invalid checkpoint. The shipped recipes are safe, so this only bites custom recipes — consider validating `candidate_formats` against the export-compatible set here (or documenting the constraint prominently).

modelopt/recipe/config.py
```diff
@@ -116,6 +122,120 @@ class ModelOptPTQRecipe(ModelOptRecipeBase):
...
+    )
+    auto_quantize_method: Literal["gradient", "kl_div"] = ModeloptField(
+        default="gradient",
+        title="Sensitivity scoring method",
+        description="'gradient' (Taylor + Fisher, needs labels) or 'kl_div' (no labels).",
+    )
+    num_score_steps: int = ModeloptField(
+        default=128,
+        title="Scoring sample count",
+        description="Number of batches used for sensitivity scoring.",
```
Edwardf0t1: Description/semantics mismatch: `num_score_steps` is described as "Number of batches", but `hf_ptq.py` consumes it as a sample count — it passes `inputs["num_score_steps"] // args.batch_size` as mtq's `num_score_steps` (which is itself in batches/steps). This preserves the old `--auto_quantize_score_size` ("Number of samples") behavior, but the rename + new description now contradict the math.

Failure scenario: a user sets `num_score_steps: 128` expecting 128 scoring batches; with `batch_size=4` they get 32 — a silent 4x under-scoring vs. the documented meaning. Either fix the description to say "samples" or drop the `// batch_size` division so the field really means batches.

jenchen13:
What is the purpose of adding YAML recipes for AutoQuantize when you can create a YAML for the ModelOpt launcher which calls AutoQuantize? [Example here](https://github.com/NVIDIA/Model-Optimizer/blob/2e5100c70885362f69c0343fc62b421c9a4d7fa2/tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16-AutoQuantize.yaml)

Especially since AutoQuantize hyperparameters are different for every model, the AutoQuantize recipes are not inherently reusable. It would make more sense to provide customizability on the client side rather than adding more recipes which are designed for reusability.

juhi10071998:
> What is the purpose of adding YAML recipes for AutoQuantize when you can create a YAML for the ModelOpt launcher which calls AutoQuantize? [Example here](https://github.com/NVIDIA/Model-Optimizer/blob/2e5100c70885362f69c0343fc62b421c9a4d7fa2/tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16-AutoQuantize.yaml)
> 
> Especially since AutoQuantize hyperparameters are different for every model, the AutoQuantize recipes are not inherently reusable. It would make more sense to provide customizability on the client side rather than adding more recipes which are designed for reusability.

My understanding is that the goal is to enable customers, such as numerics teams, to tune the recipe based on their specific needs, while also giving them a consolidated view of everything required for AutoQuant. Also, it may be simpler to create a model-specific recipe using the existing ones.

Additionally I feel there are too many knobs to tune for AutoQuantize and supporting through CLI is structurally limiting.

@shengliangxu , @realAsma feel free to add if I missed anything. My current understanding is based off our initial discussion.

realAsma:
> > What is the purpose of adding YAML recipes for AutoQuantize when you can create a YAML for the ModelOpt launcher which calls AutoQuantize? [Example here](https://github.com/NVIDIA/Model-Optimizer/blob/2e5100c70885362f69c0343fc62b421c9a4d7fa2/tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16-AutoQuantize.yaml)
> > Especially since AutoQuantize hyperparameters are different for every model, the AutoQuantize recipes are not inherently reusable. It would make more sense to provide customizability on the client side rather than adding more recipes which are designed for reusability.
> 
> My understanding is that the goal is to enable customers, such as numerics teams, to tune the recipe based on their specific needs, while also giving them a consolidated view of everything required for AutoQuant. Also, it may be simpler to create a model-specific recipe using the existing ones.
> 
> Additionally I feel there are too many knobs to tune for AutoQuantize and supporting through CLI is structurally limiting.
> 
> @shengliangxu , @realAsma feel free to add if I missed anything. My current understanding is based off our initial discussion.

I agree. @juhi10071998  Juhi had a document which made these clear. Could you please share them?

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -94,28 +94,9 @@ if [ "$LOW_MEMORY_MODE" = "true" ]; then
...
-    AUTO_QUANTIZE_CHECKPOINT="${ROOT_SAVE_PATH}/auto_quantize_checkpoints/${MODEL_NAME}_${AQ_METHOD}.pth"
-    mkdir -p $(dirname $AUTO_QUANTIZE_CHECKPOINT)
-    echo "Auto-generated auto_quantize checkpoint path: $AUTO_QUANTIZE_CHECKPOINT"
-fi
-
-if [ -n "$AUTO_QUANTIZE_BITS" ]; then
+# AutoQuantize is driven by an AutoQuantize --recipe (see modelopt_recipes/general/auto_quantize/).
+# Optional checkpoint passthrough for saving/restoring the search state.
+if [ -n "$AUTO_QUANTIZE_CHECKPOINT" ]; then
     PTQ_ARGS+=" --auto_quantize_checkpoint=$AUTO_QUANTIZE_CHECKPOINT "
```
realAsma: We had done the following:

```
# Automatically generate auto_quantize checkpoint path if not provided
```

Is this functionality remove in this script? Is that intentional?

modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_5p4bits.yaml
realAsma, line 1: Should we have effective_bits 5.0/5.4 as the default? 

This is because 4.8 was used as a good AQ default setting when FP4 cost was set as 4.0. No FP4 cost has increased. We could recommend effective_bits 5.0/5.4 as the default.

examples/hf_ptq/scripts/huggingface_example.sh
realAsma: I agree with Juhi's reasoning. 

AutoQuantize now has a lot of arguments and patches to work correctly (VL support, ActiveMoE cost etc.). We also want AutoQuantize to be extensible for internal formats. 

Supporting both CLI and recipes make things more complex.

We could add a note that CLI support for AutoQuantize has been removed and users could refer to 0.45 branch for AutoQuantize CLI support.

realAsma (commented):
Can we have one recipe for kl_div as well to show the usage?

modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_5p4bits.yaml
juhi10071998: I agree, that makes sense, I will use as 5.4.

examples/hf_ptq/scripts/huggingface_example.sh
meenchen: Since AutoQuant is an experimental feature, I am fine with just removing the CLI support.

examples/hf_ptq/scripts/huggingface_example.sh
meenchen: Can we have a default checkpoint path if it is not provided, e.g., <output_path>/.autoquant? I find the autoquant checkpoint is pretty handy.

modelopt/recipe/config.py
juhi10071998: Addressed in f5e63916b — re-added the export-safe set and folded the check into the recipe→mtq translation (`_match_candidate_to_preset`): raises on a non-export-safe preset, warns on a custom (no-preset) candidate, before the search runs.

modelopt/recipe/config.py
juhi10071998: Addressed in f5e63916b — renamed to `score_size` with an honest "number of samples (÷ batch_size)" description matching the old `--auto_quantize_score_size`. Behavior unchanged (kept the `// batch_size` and the 128 default).

examples/hf_ptq/scripts/huggingface_example.sh
juhi10071998: Addressed in f5e63916b — re-added the auto-generated checkpoint path, now gated on an AutoQuantize recipe instead of the removed `--auto_quantize_bits`.

modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_5p4bits.yaml
juhi10071998: Good point — bumped the default to 5.4 and renamed the recipe to `nvfp4_fp8_at_5p4bits` in f5e63916b.

examples/hf_ptq/scripts/huggingface_example.sh
juhi10071998: Thanks Asma, Keval — added a note in the README migration section and CHANGELOG that the old AutoQuantize CLI remains available on the 0.45 release branch (f5e63916b).

realAsma (approved):
Looks Great!

examples/hf_ptq/hf_ptq.py
```diff
@@ -309,45 +255,125 @@ def make_calib_dataloader(
...
+    Returns ``(preset_name, quant_cfg)``: ``preset_name`` is the matched preset (or None for a
+    custom config matching none), and ``quant_cfg`` is the dict passed to mtq.auto_quantize.
+    Passing the matched preset dict (rather than the candidate's own dump) keeps the search naming
+    the candidate after the preset (e.g. FP8_DEFAULT_CFG), consistent with CLI-produced checkpoints.
+    """
+    stripped = fmt.model_dump(exclude_unset=True)
+    for name, preset in QUANT_CFG_CHOICES.items():
+        if preset == stripped:
+            return name, preset
+    return None, fmt.model_dump()
```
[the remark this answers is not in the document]
juhi10071998: Fixed in fe7f4e199 — `_match_candidate_to_preset` now excludes `effective_bits` (cost-only, export-irrelevant) from the preset match, so a candidate built on a non-export-safe preset can't dodge the whitelist by adding an override; the override is preserved in the returned config. Added a regression test.

modelopt/recipe/config.py
```diff
@@ -116,6 +122,121 @@ class ModelOptPTQRecipe(ModelOptRecipeBase):
...
+    kv_cache: QuantizeConfig | None = ModeloptField(
+        default=None,
+        title="KV cache config (optional)",
+        description="QuantizeConfig applied as a uniform post-step; falls back to "
+        "the --kv_cache_qformat CLI flag when omitted.",
+    )
+
+    @field_validator("candidate_formats")
+    @classmethod
+    def _at_least_two_candidates(cls, v: list[QuantizeConfig]) -> list[QuantizeConfig]:
```
meenchen: Does BF16 (unquantized) count as a candidate here?

examples/hf_ptq/hf_ptq.py
```diff
@@ -309,45 +255,132 @@ def make_calib_dataloader(
...
+        "w4a16_nvfp4",
+        "fp8_2d_blockwise_weight_only",
+        "w4a8_mxfp4_fp8",
+        "nvfp4_mlp_only",
+        "nvfp4_experts_only",
+        "nvfp4_omlp_only",
+        "nvfp4_w4a4_weight_local_hessian",
+        "mxfp8",
+    }
+)
```
meenchen, line 281: Why do we still need this for format to quant cfg lookup? Can we pick up quant cfg directly from the recipe?

meenchen (commented):
Thanks for the PR, looks good in general.

examples/hf_ptq/scripts/huggingface_example.sh
juhi10071998: Thanks! Yes — the `--auto_quantize_*` flags are removed and AutoQuantize is recipe-only. Added a note in the README/CHANGELOG pointing users to the 0.45 branch if they need the old CLI.

modelopt/recipe/config.py
juhi10071998: this is a pure recipe load/validation time — before anything touches mtq so bf16 shouldn't be counted here

examples/hf_ptq/hf_ptq.py
juhi10071998: The quant cfg does come straight from the recipe — `_match_candidate_to_preset` isn't fetching the cfg, it's recovering the preset name. 
We hand mtq the matched preset dict so the search labels each candidate as its canonical preset (e.g. `FP8_DEFAULT_CFG`) instead of `CUSTOM_0/1`.

 That name matters for (a) `--auto_quantize_checkpoint` restore — checkpoints are keyed by these names, and `CUSTOM_N` labels break cross-run/recipe restore 
 
  (b) the export-compatibility guard (name → whitelist). Using `fmt.model_dump()` directly would quantize identically but lose both.

examples/hf_ptq/scripts/huggingface_example.sh
juhi10071998: @meenchen Done — when `--auto_quantize_checkpoint` is omitted for an AutoQuantize recipe, the script now auto-generates one at `${ROOT_SAVE_PATH}/auto_quantize_checkpoints/${MODEL_NAME}.pth`.

juhi10071998:
Thanks @meenchen  for the review- yes I've deprecated the CLI support for this. 

As for this one, I am constructing this in hf_ptq.py

> Can we have a default checkpoint path if it is not provided, e.g., <output_path>/.autoquant? I find the autoquant checkpoint is pretty handy.

examples/hf_ptq/scripts/huggingface_example.sh
kevalmorabia97: I think for less important cli args, we could remove them but seems like the most common ones already map to a field in the recipe yaml so it would be great if we can retain these cli args for one release atleast with a deprecation warning.

While it being experimental, given its a very important feature, its possible people are using it outside of nvidia

juhi10071998: hm, In that case, if we really want to support CLI, we may have to add back the disabled_layers, cost_excluded branches back as that may  be a correctness issue. Let me take a look today 
Is that right understanding @meenchen , @realAsma, @shengliangxu ? do you have any suggestions

shengliangxu: OK, so if I read the changes correctly, this PR does not really change any underlying autoquant algorithm, it just makes config to be from recipe vs from CLI. In this case, keeping CLI seems reasonable. But I am also fine with stopping CLI support, but instead of removing, print an error message to guide user to use recipe

juhi10071998: thanks Shengliang, the underlying algorithm is not changed largely (with minor changes to reading effective bits from numerics, or config, or default), but things like `disabled_layers` and `cost_excluded_patterns` have been removed from example_utils and extracted into the recipe system (details here https://github.com/NVIDIA/Model-Optimizer/pull/1856#discussion_r3501818466).

I feel adding back CLI means bringing back many changes in this PR (duplicate source of truth for exclusion lists), and will be a repetitive work again when we deprecate CLI. 
 The README is already updated to reflect this change- https://github.com/NVIDIA/Model-Optimizer/commit/f5e63916b5fa960fc203c662bd05b648f3da90da

kevalmorabia97: @juhi10071998 why do we need to add back `disabled_layers` and `cost_excluded`? I dont see them in your recipe files. If we keep other clis like auto_quantize_bits, auto_quantize_method, etc; cant we create a recipe file on the fly similar to how your recipe files look right now and the rest of the code can assume a recipe file input?

juhi10071998: @kevalmorabia97 they are present here- https://github.com/NVIDIA/Model-Optimizer/blob/ab3aed215a376b9b98badf95e7bbf472d4774326/modelopt_recipes/huggingface/qwen3_6_moe/auto_quantize/w4a16_nvfp4_fp8_at_6p0bits-active_moe.yaml#L48. I think these are needed for correctness.
let me review and see the minimal changes needed to construct the recipe on the fly, if the user want to extend those for new models, they will have to use a recipe though. I think for the default disabled layers we can do this.

shengliangxu: in quantization/config.py, we load the yaml configs and keep them as module constants, including the non-auto-quant disabled layers etc. We can do the same for the autoquant.

juhi10071998: @kevalmorabia97 Yes — included in commit (10f0691d5) does. 

We kept `--auto_quantize_bits/method/score_size/cost_model/active_moe_expert_ratio` (+ `--qformat` for the candidates), and `_auto_quantize_config_from_cli` builds an `AutoQuantizeConfig` on the fly from them; `quantize_main` then runs the same recipe-driven path, so the rest of the code assumes a config/recipe input (no separate CLI code path). A `DeprecationWarning` is emitted.

On `disabled_layers` / `cost_excluded_layers` — did **not** add them as CLI args. They're appended internally from the shared base units (`configs/auto_quantize/units/base_disabled_layers` + `base_cost_excluded_layers`), the same units the recipes splice via `$import`. 

So the on-the-fly config mirrors a recipe: candidates from `--qformat`, constraints from the scalar flags, and the base layer patterns from those shared units. Arch-specific patterns (e.g. Qwen's `*shared_expert_gate*`) stay in the model-specific recipe; the CLI shim carries only the base set.

Verified **CLI == recipe** (byte-identical `hf_quant_config.json`) on the Qwen3.6 VL MoE.

modelopt/recipe/config.py
meenchen: Is there an option for users to add bf16 to the search space, or do we always rely on mtq to include bf16? I feel we should also support one format + bf16 for AutoQuant

juhi10071998: I see, that is a good point, I think in that case we can just relax this constraint, or have atleast 1.

meenchen (approved):

modelopt/recipe/config.py
juhi10071998: addressed here- https://github.com/NVIDIA/Model-Optimizer/pull/1856/commits#:~:text=auto_quantize%3A%20allow%20a%20single%20candidate_format%20(one%2Dformat%20%2B%20bf16%20search)

examples/hf_ptq/scripts/huggingface_example.sh
juhi10071998: hi @kevalmorabia97 , please let me know if the recent changes of restoring CLI align when you get a chance, thanks!

kevalmorabia97 (approved):
LGTM. Thanks

PR #1858, opened 2026-06-30
fix(export): list unquantized MoE routers in exclude_modules (NVBug 5718750)

Edwardf0t1:
### What does this PR do?

Type of change: Bug fix

Fixes NVBug 5718750 (JIRA OMNIML-3159): a Qwen3-30B-A3B NVFP4 checkpoint fails to load in vLLM/SGLang with
`AssertionError: Tried to load weights of size [128, 2048] to a parameter of size [128, 1024]`.

**Root cause.** During unified HF export, `get_quant_config` records a module in `exclude_modules` **only if it carries a quantizer** (even a disabled one). MoE routers are intentionally kept in original precision via the `*mlp.gate.*` / `*router*` disable patterns — but those only *disable* a quantizer that already exists; they don't create one.

- On `transformers<5.0` the router (`mlp.gate`) is a plain `nn.Linear`, so `mtq.quantize` attaches a (disabled) quantizer → it is recorded → correctly listed in `exclude_modules`.
- On `transformers>=5.0` MoE experts are batched and the router is a `TopKRouter` (not an `nn.Linear`), so it **never receives a quantizer**. The exporter skips it (`has_quantizers == False`), so its BF16 weight is written to the checkpoint but omitted from `exclude_modules`. vLLM/SGLang then treat the router as a quantized (packed-FP4) weight and fail to load it.

QA confirmed the failing export ran with **transformers 5.12.0**, matching this analysis.

**Fix.** Detect MoE routers structurally — an MoE block that exposes an `experts` container plus a weight-bearing `gate` / `router` / `shared_expert_gate` submodule — and record any such router that is left unquantized as `QUANTIZATION_NONE`, so it always lands in `exclude_modules` regardless of whether a quantizer happens to be attached. Routers the user opted to quantize (non-NONE format) are left untouched.

### Usage

No API change. The exported `hf_quant_config.json` / `config.json` `ignore` list now contains the MoE router gates (e.g. `model.layers.*.mlp.gate`) on the transformers-5 batched-experts path, as it already did on transformers-4.

### Testing

- New CPU unit test `tests/unit/torch/export/test_get_quantization.py::test_moe_router_excluded_when_not_quantized`: quantizes a fake MoE model whose router is a non-`nn.Linear` `TopKRouter` (no quantizer attached) to NVFP4 and asserts the router appears in `exclude_modules` while the quantized experts do not.
- `tests/unit/torch/export/test_get_quantization.py`, `test_unified_export_hf.py`, `test_layer_utils.py` all pass.
- `pre-commit` (ruff, ruff-format, mypy, bandit, license) passes on the changed files.

### Additional Information

NVBug 5718750 · JIRA OMNIML-3159. Same `transformers>=5.0` batched-experts MoE export path as the recent fused-experts work (OMNIML-5003).

## Summary by CodeRabbit

* **Bug Fixes**
  * Exported quantization configs now reliably mark mixture-of-experts router/gate components as unquantized, avoiding downstream checkpoint-loading failures caused by weight-shape mismatches.
  * Enhanced router/gate detection so MoE routers that aren’t standard linear layers are no longer accidentally left out of the exported config.

* **Tests**
  * Added/expanded unit tests covering MoE router exclusion behavior, including cases where the MoE block is the root module.


modelopt/torch/export/quant_utils.py
```diff
@@ -1350,6 +1350,39 @@ def preprocess_linear_fusion(modules: list[torch.nn.Module], resmooth_only=False
...
+    """Return the names of MoE router/gate submodules left in original precision.
+
+    A module is added to ``exclude_modules`` during unified HF export only if it carries a
+    quantizer (even a disabled one) -- see :func:`get_quant_config`. MoE routers are kept
+    unquantized on purpose, but on ``transformers>=5.0`` they are no longer ``nn.Linear``
+    modules (e.g. ``TopKRouter``), so ``mtq.quantize`` never attaches a quantizer to them.
+    Without this, the BF16 router weight is written to the checkpoint but omitted from
+    ``exclude_modules``, and deployment frameworks (vLLM / SGLang) then try to load it as a
+    quantized weight -- e.g. ``AssertionError: Tried to load weights of size [E, H] to a
+    parameter of size [E, H/2]`` for Qwen3-MoE (NVBug 5718750).
```
juhi10071998: nit. do we remove the NVBug reference as it may become stale- `Tried to load weights of size [E, H] to a
    parameter of size [E, H/2]`` for Qwen3-MoE (NVBug 5718750).

modelopt/torch/export/quant_utils.py
```diff
@@ -1350,6 +1350,39 @@ def preprocess_linear_fusion(modules: list[torch.nn.Module], resmooth_only=False
...
+
+    Routers are detected structurally: an MoE block exposes an ``experts`` container plus a
+    ``gate`` / ``router`` (or ``shared_expert_gate``) submodule that owns a weight tensor.
+    Routers that the user opted to quantize (a non-NONE format) are skipped.
+    """
+    router_attrs = ("gate", "router", "shared_expert_gate")
```
juhi10071998, line 1369: just curious, Why do we want to include the shared_expert_gate too? that should already be detected right? is this more like a safety check?

jingyu-ml (approved):
LGTM

modelopt/torch/export/quant_utils.py
Edwardf0t1: Yes — it's a forward-compat safety net rather than a strictly-needed case today. On `transformers<5.0` `shared_expert_gate` is an `nn.Linear` (Qwen2-MoE), so it already gets a disabled quantizer via the `*mlp.shared_expert_gate.*` pattern and is recorded by the main loop; the `setdefault` here is then a no-op for it. It only does real work if a future/variant routing submodule becomes non-`nn.Linear` (the same shift that broke `gate` on transformers 5.x). Cheap to include and keeps all three routing submodules consistent.

modelopt/torch/export/quant_utils.py
```diff
@@ -1350,6 +1350,39 @@ def preprocess_linear_fusion(modules: list[torch.nn.Module], resmooth_only=False
...
+            continue
+        for attr in router_attrs:
+            router = getattr(module, attr, None)
+            if not isinstance(router, nn.Module):
+                continue
+            if not isinstance(getattr(router, "weight", None), torch.Tensor):
+                continue
+            if get_quantization_format(router) != QUANTIZATION_NONE:
+                continue
+            router_names.append(f"{name}.{attr}")
```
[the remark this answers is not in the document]
Edwardf0t1: Good catch — fixed in da4340f15a. Root-module case now builds the name as just the attr when `name == ""`, and added `test_moe_router_names_handle_root_module` to cover it.

modelopt/torch/export/quant_utils.py
Edwardf0t1: Done in da4340f15a — removed the NVBug number from the code docstring/comment. Kept it in the CHANGELOG and commit history as the canonical reference.

CHANGELOG.rst
```diff
@@ -41,6 +41,7 @@ Changelog
...
+- Always list unquantized MoE routers/gates in the exported ``exclude_modules`` (NVBug 5718750). ``get_quant_config`` only recorded modules that carry a quantizer, but on ``transformers>=5.0`` MoE routers are no longer ``nn.Linear`` (e.g. ``TopKRouter``) and never receive one, so the BF16 router weight was written to the checkpoint yet omitted from ``exclude_modules``. vLLM / SGLang then treated it as quantized and failed to load (e.g. Qwen3-30B-A3B NVFP4: ``AssertionError: Tried to load weights of size [128, 2048] to a parameter of size [128, 1024]``). Routers are now detected structurally (an MoE block with an ``experts`` container plus a weight-bearing ``gate`` / ``router`` / ``shared_expert_gate`` submodule) and recorded as unquantized regardless of quantizer attachment.
```
kevalmorabia97: This should go in 0.45 section

Edwardf0t1: Done in 457faef136 — moved the entry to the 0.45 Bug Fixes section.

kevalmorabia97 (approved):

modelopt/torch/export/quant_utils.py
juhi10071998: Agreed, SGTM.

cjluo-nv (approved):

Clean, well-scoped bug fix for NVBug 5718750. On `transformers>=5.0`, MoE routers (e.g. `TopKRouter`) are no longer `nn.Linear`, so `mtq.quantize` never attaches a quantizer and the BF16 router weight was exported but omitted from `exclude_modules`, causing vLLM/SGLang load failures. The new `_get_unquantized_moe_router_names` detects routers structurally (a block with an `experts` container plus a weight-bearing `gate`/`router`/`shared_expert_gate` submodule) and records unquantized ones as `QUANTIZATION_NONE`.

Correctness verified:
- Uses `setdefault`, so it never clobbers entries from the existing quantizer-based path (which runs first), and routers the user opted to quantize (non-NONE format) are correctly skipped.
- Router naming matches dotted module paths, and the root-module case (empty `name` → no leading dot) is handled and explicitly tested.
- Reuses existing `get_quantization_format`; no duplication of logic in `moe_utils.py` or elsewhere.

Tests are meaningful: an integration test runs real `mtq.quantize` with a non-Linear router and asserts the router lands in `exclude_modules` while quantized experts do not, plus a unit test for the root-module naming edge case.

Size is small (+149/-0), cohesive. No design-gate concern (single helper, not a new subsystem). Licensing: only CHANGELOG + existing source/test files touched — no license file/header changes. No prompt-injection attempts in the PR content.

Minor (non-blocking) nit: `test_moe_router_names_handle_root_module` imports `_get_unquantized_moe_router_names` inside the function rather than at module top; could be hoisted, but it's a private test-only import and harmless.

modelopt/torch/export/quant_utils.py
```diff
@@ -1350,6 +1350,39 @@ def preprocess_linear_fusion(modules: list[torch.nn.Module], resmooth_only=False
                 module.weight_quantizer.amax = weight_amax
 
 
+def _get_unquantized_moe_router_names(model: nn.Module) -> list[str]:
+    """Return the names of MoE router/gate submodules left in original precision.
+
+    A module is added to ``exclude_modules`` during unified HF export only if it carries a
+    quantizer (even a disabled one) -- see :func:`get_quant_config`. MoE routers are kept
+    unquantized on purpose, but on ``transformers>=5.0`` they are no longer ``nn.Linear``
+    modules (e.g. ``TopKRouter``), so ``mtq.quantize`` never attaches a quantizer to them.
```
meenchen, line 1359: Should we add quantization support for TopKRouter instead of post-processing? Theoretically, routers can be quantized as well

Edwardf0t1: Good question. I'd keep these as separate concerns. Quantizing the router is opt-in and uncommon — ModelOpt's `default_disabled_quantizers` already disables `*mlp.gate.*` / `*router*`, and the public Qwen3-FP4 reference keeps every `*.mlp.gate` in BF16 (routers are tiny and sensitive). The export fix is about a different invariant: any exported-but-unquantized weight must appear in `exclude_modules` so vLLM/SGLang don't load it as FP4. That has to hold for the default BF16-router case regardless of whether we later support router quantization.

The two also coexist — the helper skips routers that *are* quantized (`get_quantization_format != QUANTIZATION_NONE`), so it doesn't preclude adding `TopKRouter` quantization later; it only fixes the unquantized path. I also leaned on structural detection rather than registering `TopKRouter` so it stays robust to upstream router-class renames (which is what broke this in the first place — `nn.Linear` -> `TopKRouter` in transformers 5.x). Happy to file a follow-up for opt-in router quantization if that's useful, but I don't think it should replace the export-side guarantee.

PR #1860, opened 2026-06-30
fix: prevent UnboundLocalError masking real errors in fsdp2_aware_weight_update

### LINKED ISSUE #1859: fsdp2_aware_weight_update hides the real error with an UnboundLocalError
**Before submitting an issue, please make sure it hasn't been already addressed by searching through the existing and past issues.**

## Describe the bug

`fsdp2_aware_weight_update` (in `modelopt/torch/quantization/utils/core_utils.py`) is a context manager. Its `finally` block uses three variables — `fsdp_param_mapping`, `fsdp_param_group`, and `root_module` — that are only set **after** the `unshard()` call inside the `try`.

So if `unshard()` fails (typically a CUDA OOM when a large FSDP block can't fit on one GPU), the `finally` runs before those variables exist and raises:

```
UnboundLocalError: local variable 'fsdp_param_mapping' referenced before assignment
```

In Python, an error raised inside `finally` replaces the original error. So the real failure (the OOM) is thrown away, and you only see the misleading `UnboundLocalError`. The actual problem becomes very hard to debug.

This shows up during export of large (especially MoE) models, where unsharding a block is most likely to run out of memory.

### Steps/Code to reproduce bug

The simplest deterministic repro: make the in-body `unshard()` raise and confirm the wrong error comes out.

1. Shard a model under FSDP2.
2. Patch the enclosing module's `unshard` to raise any error.
3. Enter `fsdp2_aware_weight_update(model, sharded_module)`.
4. You get `UnboundLocalError` instead of the error you raised.

In the real world this happens on its own: export a large MoE model where one block's `unshard()` all-gather exceeds GPU memory, and the OOM is replaced by the `UnboundLocalError`.

### Expected behavior

If something fails before those variables are set, the `finally` should be a no-op and the original error (e.g. the OOM) should propagate unchanged.

### Who can help?

-

## System information

- OS: Ubuntu 22.04
- CPU architecture: x86_64
- GPU name: H100
- Library versions (if applicable):
  - Python: 3.12
  - ModelOpt version or commit hash: present on current `main`
  - PyTorch: 2.x

dinhxuanvu:
Closes #1859

## What

`fsdp2_aware_weight_update`'s `finally` block uses `fsdp_param_mapping`, `fsdp_param_group`, and `root_module`, which are only set after the `unshard()` inside the `try`. If `unshard()` fails first (e.g. a CUDA OOM on a large block), the `finally` raises `UnboundLocalError` — and since that happens inside `finally`, it replaces the real error. The user sees a confusing `UnboundLocalError` instead of the actual OOM.

## Fix

- Set `fsdp_param_mapping = {}`, `fsdp_param_group = None`, `root_module = None` before the `try`.
- Guard the writeback/reshard so they're skipped when those values were never set.

On an early failure the `finally` is now a no-op and the original error propagates. The success path is unchanged.

## Test

Added `test_fsdp2_weight_update_context_error_propagates`: forces the in-body `unshard()` to raise and asserts the original error surfaces (not `UnboundLocalError`). Fails on the old code, passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash path in weight update handling where an earlier error could be hidden by a secondary cleanup error.
  * Original failures now propagate properly, even if sharding/unsharding fails partway through.

* **Tests**
  * Added a regression test to confirm the original error is preserved and cleanup behavior remains safe.


dinhxuanvu:
@cjluo-nv PTAL

PR #1879, opened 2026-07-01
Add TensorQuantizer rotate-back mode

realAsma:
## Summary
- add `RotateConfig.mode` with the existing `rotate` behavior as the default
- support `rotate={"enable": true, "mode": "rotate_back"}` for rotate -> quantize -> rotate back in fake quantization
- reject rotate-back mode for real quantization instead of silently skipping inverse rotation
- add `NVFP4_W4A4_ROTATE` eval config with the config-level embed-token rotate workaround and TODO for the longer-term fix
- add focused CPU coverage for config serialization, default rotation, rotate-back, and real-quant rejection

## Latest Update
- fixed a critical fakequant eval/export bug where `fold_weight` applied rotation into the stored weight, disabled the quantizer, but left rotation enabled, causing the folded weight to be rotated a second time on later use
- fused the Hadamard rotation scale into the quantization path and removed the redundant autograd wrapper
- added a config-level workaround for W4A4 rotate: `*embed_tokens*quantizer` has rotate explicitly disabled because the embedding token/input path creates the rotate issue; a TODO tracks replacing this with a better fix

## Example

```python
quant_cfg = {
    "quant_cfg": [
        {
            "quantizer_name": "*weight_quantizer",
            "cfg": {
                "num_bits": 4,
                "block_sizes": {-1: 128},
                "fake_quant": True,
                "rotate": {
                    "enable": True,
                    "mode": "rotate_back",
                },
            },
        },
    ],
    "algorithm": "max",
}
```

`rotate_back` selects rotate -> quantize -> rotate-back. The existing behavior stays as `rotate=True` or `rotate={"enable": True, "mode": "rotate"}`. `rotate_back` is supported for fake quant only; real quantization raises clearly instead of silently skipping the inverse rotation.

## Validation
- `pre-commit run --files modelopt/torch/quantization/config.py modelopt/torch/quantization/nn/modules/tensor_quantizer.py tests/unit/torch/quantization/test_tensor_quant_cpu.py`
- `pytest_pwd tests/unit/torch/quantization/test_tensor_quant_cpu.py -q --tb=short -x`
- `python_pwd -m py_compile examples/llm_eval/quantization_utils.py`
- `pre-commit run --files examples/llm_eval/quantization_utils.py`

## Eval Results

Qwen/Qwen3-8B MMLU eval from `examples/llm_eval/lm_eval_hf.py` using `lm-eval==0.4.10`, `mmlu`, 5-shot, `batch_size=4`, `calib_size=512`, and `calib_batch_size=4` for quantized runs.

| Configuration | Quant config | MMLU acc | Stderr | Runtime |
| --- | --- | ---: | ---: | ---: |
| BF16 | - | 0.7488 | 0.0035 | 28m 42s |
| FP4 W/A | `NVFP4_DEFAULT_CFG` | 0.7097 | 0.0036 | - |
| FP4 W/A rotate, embed-token rotate disabled | `NVFP4_W4A4_ROTATE` | 0.6916 | 0.0037 | - |
| FP4 W/A rotate + rotate-back, embed-token rotate disabled | `NVFP4_W4A4_ROTATE_ROTATE_BACK` | 0.6945 | 0.0036 | - |

Full run logs, status files, GPU snapshots, and JSON outputs are retained in the private BeeBot artifact bundles for the Qwen3-8B eval/debug runs.

## Review
- SE implementation: complete
- SR review: clean on iteration 2
- ST validation: pass on iteration 2

## Summary by CodeRabbit

* **New Features**
  * Added support for input rotation modes, including a new rotate-back flow for fake quantization.
  * Improved weight folding so rotation is baked into weights and won’t be reapplied on later passes.
  * Updated quantizer display text to better reflect rotation and disabled states.

* **Bug Fixes**
  * Fixed rotation handling when quantizers are disabled or partially updated.
  * Improved consistency and validation for Hadamard transform behavior across CPU and GPU paths.


cjluo-nv (commented):

Small, cohesive feature (+106/-9, 3 files) adding a `rotate_back` mode to `TensorQuantizer` (rotate → quantize → rotate-back for fake quant), with a clear rejection for real quantization. Code is clean and reuses existing machinery: a new `_rotate_inputs` helper deduplicates the two rotation call-sites, and `rotate_back_is_enabled` mirrors the existing `rotate_is_*` property pattern (dict/RotateConfig/bool backward-compat handled consistently). The `field_name == "rotate"` short-circuit in `validate_config` correctly stops the "only {'enable': True}" check from mis-firing on `rotate={"enable": True}` (a latent bug), and RotateConfig's own validators still cover the sub-fields.

CPU test coverage is solid: config serialization round-trip, default-path (transform called once), rotate-back path (transform called twice with matching fp32/block_size args), and the real-quant ValueError. Tests monkeypatch `normalized_hadamard_transform` so they don't need the `fast_hadamard_transform` package. No licensing concerns, no oversized diff, and no prompt-injection content in the PR fields.

- **Correctness rests on the Hadamard transform being self-inverse.** `rotate_back` applies the *same* `normalized_hadamard_transform` a second time rather than an explicit inverse; this is only correct because the normalized FHT is orthogonal/symmetric (HᵀH = I). That holds for the full-dim and fixed-block paths, but the auto-selected block-size path (`block_size=None` → `_largest_pow2_divisor(dim)`) picks the block size from the tensor shape at call time — worth a human confirming the forward and rotate-back calls always resolve to the same block size so the round-trip is truly identity. The CPU tests stub out the transform, so this numerical property is not actually exercised (GPU-only).
- **CHANGELOG not updated.** This is a user-facing `QuantizerAttributeConfig.rotate.mode` addition; the 0.46 "New Features" section has no entry for it.

tests/gpu/torch/quantization/test_hadamard.py
```diff
@@ -48,6 +48,9 @@ def test_hadamard_transform(dim):
     xxt_h = x_h @ x_h.T
     # The numerical error can be large, especially for 16-bit floats.
     assert torch.allclose(xxt_h, xxt, atol=0.05)
+    x_roundtrip = normalized_hadamard_transform(x_h)
+    assert torch.allclose(x_roundtrip, x, rtol=1e-5, atol=1e-6)
```
realAsma, line 54: Does this pass for dim=8?

modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -1092,11 +1108,7 @@ def forward(self, inputs):
 
         # Rotating the input
         if self.rotate_is_enabled:
-            inputs = normalized_hadamard_transform(
-                inputs,
-                rotate_fp32=self.rotate_is_fp32,
-                block_size=self.rotate_block_size,
-            )
+            inputs = self._rotate_inputs(inputs)
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Fixed in `afcbf78`: `rotate_back` + real quantization is now rejected before the initial rotation path, and the unit test fails if rotation is reached before the intended `ValueError`.

tests/gpu/torch/quantization/test_hadamard.py
realAsma: > _:robot: Bot comment._

Yes. This test is parameterized over `dim=[8, 16]`, so `dim=8` is covered by the GPU test. I also ran a direct check on `a8`/`ada` for `dim=8`; roundtrip `allclose=True` with max abs error `1.1920928955078125e-07`.

realAsma (commented):
Simplify all the fold_weight doctoring -> no need of mentioning rotate functionality in particular - it is a Lower level details obvious in the code.

realAsma (commented):
There are a lot of unittests; can we simplify or consolidate any of them? 

are there any low level tests which are already covered by high-level tests - if yes should we remove them?

We can keep the unittests that are essential for testing the correctness or expected behavior

modelopt/torch/quantization/config.py
```diff
@@ -287,6 +287,7 @@ class RotateConfig(ModeloptBaseConfig):
     """
 
     enable: bool = False
+    mode: Literal["rotate", "rotate_back"] = "rotate"
     rotate_fp32: bool = False
     block_size: int | None = None
```
realAsma: Can we add doc for these fields? Not clear what these are for

examples/llm_eval/quantization_utils.py
realAsma, line 1: Remove this; This was supposed to be a temporary artifact.

realAsma: Move the TODO to embed module file `modelopt/torch/quantization/nn/modules/quant_embedding.py`.

realAsma: > _:robot: Bot comment._

Removed the temporary W4A4 rotate eval config in the latest update.

cjluo-nv (commented):

Re-review of the rotate-back mode PR (now grown to +504/-127, 14 files from the original +106/-9). All previously-raised critical/major comments are resolved:

- :speech_balloon: CodeRabbit major "reject rotate_back before rotation": *fixed in `afcbf78`* — `forward()` now raises `ValueError` at line ~1125, before `_rotate_inputs()` at ~1129; `test_tensor_quantizer_rotate_back_rejects_real_quant` monkeypatches the transform to raise so the test fails if rotation is reached first. Fully addressed.
- :speech_balloon: CHANGELOG missing entry: *added* — CHANGELOG.rst now documents `rotate.mode`. Addressed.
- :speech_balloon: config.py field docs: *added* title/description to all `RotateConfig` fields. Addressed.
- :speech_balloon: temp W4A4 eval config / embed_tokens workaround in `quantization_utils.py`: *removed* — file no longer contains them. Addressed.
- :speech_balloon: dim=8 hadamard roundtrip: *covered* — `test_hadamard.py` parametrizes dim=[8,16] with a roundtrip assertion. Addressed.

Design-review protocol (gate fired on 7 directories): the PR does not introduce a new subsystem/DSL/loader — it adds a `mode` field to an existing config and consolidates three `fold_weight` call-sites (base/HF-fused-experts/vLLM-MoE) into a shared `_fold_weight_quantizer` helper plus a `disable_rotate()` method. This is de-duplication, not a competing abstraction, so no alternative-system justification is needed. The `validate_config` `field_name=="rotate"` short-circuit and `rotate_back_is_enabled` property correctly mirror the existing `rotate_is_*` dict/RotateConfig/bool back-compat pattern. No licensing changes; no prompt-injection content acted on.

Why nudge rather than approve — two changes deserve a human with GPU/architectural context:
1. `normalized_hadamard_transform` was rewritten: the custom `FastHadamardTransform` autograd `Function` was removed and the code now relies on `fast_hadamard_transform.hadamard_transform(..., scale=block_size**-0.5)` being autograd-aware with a fused scale. This is a correctness-sensitive change to the QAT/training backward path, and rotate-back correctness still rests on the FHT self-inverse property — all exercised only on GPU (CPU unit tests stub the transform).
2. The critical fakequant double-rotation bug fix (rotation baked into folded weight then disabled so it isn't re-applied) and the fold_weight consolidation across the vLLM/HF export paths are validated at the CPU/unit level with a stubbed transform, but the real end-to-end fold+rotate export path is GPU-only.

CPU/unit coverage is otherwise solid and thorough (serialization, disabled-roundtrip, real-quant rejection, fold no-op/idempotency, per-expert fold, repr).

CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Changelog
 
 **New Features**
 
+- Add ``rotate.mode`` to torch quantizer configs. The default ``"rotate"`` keeps the existing rotate-before-quantize behavior; ``"rotate_back"`` enables fake-quant rotate → quantize → rotate-back for TensorQuantizer.
```
realAsma: this is a minor feature. Can we move this down below other features?

realAsma: > _:robot: Bot comment._

Moved this entry to the end of the 0.46 New Features list in `dadfe73db9`.

realAsma:
> _:robot: Bot comment._

Replying to [this review](https://github.com/NVIDIA/Model-Optimizer/pull/1879#pullrequestreview-4637473637) with the rationale behind the two changes:

1. The custom autograd wrapper was removed because `fast_hadamard_transform.hadamard_transform` is already autograd-aware. For the normalized Walsh–Hadamard transform, the Jacobian is the same orthonormal transform (`Hᵀ = H` and `H² = nI`), so passing `scale=block_size**-0.5` into the kernel preserves the forward and backward math while removing a redundant wrapper and separate scaling operation. Real-CUDA tests now cover dim 8/16 round trips and backward propagation for both `rotate` and `rotate_back`.

2. During `fold_weight`, rotation and fake quantization are materialized into the stored weight. If rotation remains enabled afterward, the next forward applies the Hadamard transform again to an already folded weight—the double-rotation bug. The shared `_fold_weight_quantizer` therefore disables both quantization and rotation after copying the folded value, and centralizes that invariant for the base, Hugging Face fused-expert, and vLLM MoE paths.

Real-GPU validation covered Hadamard/backward behavior and end-to-end fold+rotate HF/vLLM-helper export with zero reference difference. A fresh four-GPU Qwen3-8B `NVFP4_W4A4_ROTATE` run completed at MMLU `0.6916 ± 0.0037`.

cjluo-nv (commented):

Re-review of the rotate-back mode PR (+530/-127, 15 files, author realAsma). Design-review protocol applied: the complexity gate fired on directory span, but this is **not** a new subsystem/DSL/loader — it adds a `mode` field to the existing `RotateConfig` and consolidates three `fold_weight` call-sites (base `quant_module`, HF fused-experts, vLLM MoE) into a shared `_fold_weight_quantizer` helper + `disable_rotate()` method. That's de-duplication, not a competing abstraction, so no alternative-system justification is needed. `rotate_back_is_enabled` and the `validate_config` `field_name=="rotate"` short-circuit correctly mirror the existing `rotate_is_*` dict/RotateConfig/bool back-compat pattern. GitHub/licensing: only the standard NVIDIA header on the touched/new files, no license-file or SPDX changes. No prompt-injection content acted on (PR body/comments contained CodeRabbit autofix scaffolding and bot chatter only — treated as data).

**All previously-raised critical comments are resolved:**
- :speech_balloon: CodeRabbit major "reject `rotate_back` before rotation": *fixed in `afcbf78`* — `forward()` raises `ValueError` at the check `if self.rotate_back_is_enabled and self._if_quant and not self.fake_quant` placed ahead of `_rotate_inputs()`; `test_tensor_quantizer_rotate_back_rejects_real_quant` monkeypatches the transform to raise so the test fails if rotation is reached first. Verified in current file.
- :speech_balloon: CHANGELOG missing/placement: *added and moved* to the end of the 0.46 New Features list. Verified.
- :speech_balloon: `config.py` field docs: *added* title/description to all `RotateConfig` fields. Verified.
- :speech_balloon: temp W4A4 eval config / embed_tokens workaround in `quantization_utils.py`: *removed*; the tracking TODO was relocated into `quant_embedding.py` docstring as requested. Verified.
- :speech_balloon: dim=8 hadamard roundtrip: *covered* — `test_hadamard.py` parametrizes dim=[8,16] with roundtrip + backward assertions. Verified.

CPU/unit coverage is thorough (rotate-mode serialization, disabled-path roundtrip, real-quant rejection, fold no-op/idempotency + keep_attrs, per-expert fold, disable_rotate type-preservation, repr).

**Why nudge rather than approve** — two correctness-sensitive changes are exercised only on GPU (CPU tests stub `normalized_hadamard_transform`), so a human with GPU/arch context should sign off:
1. `normalized_hadamard_transform` was rewritten: the custom `FastHadamardTransform` autograd `Function` was removed and the code now relies on `fast_hadamard_transform.hadamard_transform(..., scale=block_size**-0.5)` being autograd-aware with a fused scale. Rotate-back correctness rests on the FHT self-inverse property (HᵀH=I). The author replied with the math rationale (Hᵀ=H, H²=nI) and reports real-CUDA dim-8/16 roundtrip + backward coverage — still warrants GPU sign-off because the property isn't exercised on CPU.
2. The fakequant double-rotation fix (rotation baked into the folded weight, then `disable_rotate()` so it isn't re-applied) and the fold consolidation across vLLM/HF export paths are validated at CPU/unit level with a stubbed transform; the real end-to-end fold+rotate export path is GPU-only. Author reports zero-reference-diff GPU validation and a Qwen3-8B MMLU run — human confirmation of the E2E export path recommended.

modelopt/torch/quantization/config.py
realAsma: > _:robot: Bot comment._

Done — all `RotateConfig` fields (`enable`, `mode`, `rotate_fp32`, `block_size`) now use `ModeloptField` with title and description docs, and the class docstring links to `normalized_hadamard_transform` for transform details.

kinjalpatel27 (approved):
LGTM. Thanks Asma

sugunav14 (approved):

modelopt/torch/quantization/config.py
```diff
@@ -286,9 +286,29 @@ class RotateConfig(ModeloptBaseConfig):
...
+        default=False,
+        title="Enable input rotation.",
+        description="If True, applies a normalized Hadamard transform before quantization.",
+    )
+    mode: Literal["rotate", "rotate_back"] = ModeloptField(
+        default="rotate",
+        title="Rotation mode.",
+        description=(
+            "Use 'rotate' for input rotation only, or 'rotate_back' to apply the transform "
+            "again after fake quantization."
```
meenchen, line 299: It is not very clear to me what the use cases are for the different modes. Can you elaborate?

realAsma: > _:robot: Bot comment._

Thanks — the distinction is about whether the quantizer changes the tensor's basis:

- `rotate`: computes `Q(Hx)` and leaves the output in the Hadamard-rotated basis. This is for flows where the surrounding graph compensates for that basis change (for example, pairing activation rotation with a correspondingly folded/rotated linear weight). It is also the existing behavior, so it remains the default for backward compatibility.
- `rotate_back`: computes `H Q(Hx)` (the normalized Hadamard transform is self-inverse), so quantization still benefits from the rotated-domain value distribution but the quantizer output returns to the original basis. This is the drop-in choice when downstream operators should remain unchanged.

`rotate_back` is currently fake-quant-only because a real quantized tensor cannot be inverse-rotated through the existing quantized representation/backend.

realAsma: > _:robot: Bot comment._

rotate without rotate back only: Requires the complimentary operation also to be rotated. (example : weight_quantizer rotate -> activation quantizer should also rotate, otherwise model will produce garbage.)

rotate with rotate back: The op is rotated -> quantized -> rotated back ; Does not need complimentary op to be rotated to be functionally correct.

PR #1897, opened 2026-07-03
Save hf checkpoint at every valitation iteration during distillation.

danielkorzekwa:
### What does this PR do?

Save hf checkpoint at every valitation iteration during distillation.

Addition functionality:
- added `--validate_only` in `examples/megatron_bridge/distill.py` to enable computing validation losses for iter 0
- added `--reset_optimizer` in `examples/megatron_bridge/distill.py` to enable not using presaved optimizer, e.g., when changing the number of train iters.

### Usage

- examples/megatron_bridge/distill.py
- examples/megatron_bridge/README.md (line 228)

### Testing

- tests/examples/megatron_bridge/test_distill.py

## Summary by CodeRabbit

* **New Features**
  * Added `--validate_only` for student-only validation at iteration 0.
  * Added `--hf_validation_export_path` with `--hf_validation_export_interval` to export validated student HuggingFace artifacts during distillation.
  * Added `prepare_data_blend.py` for YAML-driven token-budgeted data blends.
  * Added `--max_tokens` to stop Megatron preprocessing after a token budget.
* **Bug Fixes**
  * Validation exports avoid duplicate checkpoints, preserve the student architecture/config, and export only student artifacts.
* **Documentation**
  * Expanded researcher and tutorial guides for iterative workflows and token-budgeted blends.
* **Tests**
  * Updated distillation/blend/max_tokens test coverage, including validate-only and interval-based exports.


AAnoosheh (approved):

examples/megatron_bridge/distill.py
```diff
@@ -233,6 +297,11 @@ def get_args():
     parser.add_argument("--lr", type=float, default=1e-4, help="Peak learning rate")
     parser.add_argument("--min_lr", type=float, default=1e-5, help="Minimum learning rate")
     parser.add_argument("--lr_warmup_iters", type=int, default=50, help="Number of LR warmup steps")
+    parser.add_argument(
```
jenchen13: when is reset optimizer used?

examples/megatron_bridge/distill.py
```diff
@@ -444,12 +549,35 @@ def _build_model_provider(hf_path, load_weights=True):
...
+        distill(config)
+    if args.validate_only:
+        print_rank_0("\nValidation-only run done!\n")
+    else:
+        print_rank_0(
+            f"\nDistillation done! Saved checkpoint to {checkpoint_dir}"
+            " in megatron distributed checkpoint format.\n"
+        )
+
+    # TODO: Extend _HFValidationExportCallback to export --hf_export_path from the in-memory
```
jenchen13: would it be possible to export from the in memory student in this PR? loading an extra model is very expensive memory wise. you can probably do some async save of the in-memory student

examples/researcher_guide/README.md
```diff
@@ -47,6 +51,114 @@ and should not be reported as final benchmark results.
...
+
+- Dataset: nvidia/Nemotron-Post-Training-Dataset-v2 (math and stem)
+- Teacher CE: 1.1566
+
+| Training tokens | Validation KD | Validation CE | MMLU | MMLU-Pro |
+|----------------:|--------------:|--------------:|-----:|---------:|
+| 0 | 0.5187 | 1.4739 | 48.69% (full) | 23.09% (full) |
+| 0.7M | 0.1919 | 1.0931 | 58.74% | 13.14% |
+| 3.3M | 0.1342 | 1.0550 | 60.56% | 14.29% |
+| 39.3M | 0.0675 | 1.0296 | 64.63% | 6.14% |
```
jenchen13, line 111: why does MMLU pro drop so low here?

Surya-5555: @jenchen13 The catastrophic drop in MMLU to ~6% happens because the Multi-Layer Perceptron (MLP) is losing its non-linearity and collapsing into a purely linear projection during export.

@danielkorzekwa I did a deep dive into your latest commit (`0fee306`). While it successfully protects the live teacher by detaching it, the student's config is still forced to `activation_func = None` to remain pickleable. 

When this sanitized config hits the Model-Optimizer export pipeline (`modelopt/torch/export/layer_utils.py`), a silent failure occurs:
1. `_get_hidden_act` relies on `hasattr(module, "activation_func")`. In Python, this evaluates to `True` even if the attribute is explicitly `None`.
2. Because it bypasses this check, Hugging Face receives the un-mappable `None` type and gracefully initializes the MLP *without* an activation function instead of crashing.
3. Without the activation function, $W_2(\sigma(W_1(x)))$ becomes $W_2(W_1(x))$, lobotomizing the model's reasoning.

To permanently fix this, we need to enforce a fail-fast boundary in `_get_hidden_act`: trapping the explicit `None` state with `getattr`, unwrapping `functools.partial` objects, and explicitly validating against Hugging Face's `ACT2FN` dictionary.

I have the exact structural code fix and the `unittest.mock` regression tests ready locally. Let me know if you'd like me to open a PR against your branch to inject it!

examples/megatron_bridge/distill.py
danielkorzekwa: @AAnoosheh , @kevalmorabia97  any idea if we could do it?

examples/researcher_guide/README.md
danielkorzekwa: yes, please open MR into this dkorzekwa/save_hf_checkpoint_for_every_val_iter_during_distill, thanks

Add @AAnoosheh , @kevalmorabia97 as reviewers

Surya-5555: > yes, please open MR into this dkorzekwa/save_hf_checkpoint_for_every_val_iter_during_distill, thanks
> 
> Add @AAnoosheh , @kevalmorabia97 as reviewers

Done! I just opened PR #1997 against your branch with the strict validation fixes and regression tests. I've also tagged @AAnoosheh and @kevalmorabia97 for review over there as requested!

examples/megatron_bridge/distill.py
danielkorzekwa: reset optimizer was removed from this MR

examples/megatron_bridge/distill.py
danielkorzekwa: We are exporting now offline. Once distillation is finished, the export script is called to export multiple mbridge checkpoints

examples/researcher_guide/README.md
danielkorzekwa: lets abandon #1997. Now we export to HF after distillation process is completed. See the latest commit in this branch.

danielkorzekwa: @jenchen13  regarding, why does MMLU pro drop so low here? - it is due to overfitting to nemotron v2 style, added to the README.md:

```
 Nemotron v2 KD continues to decrease, but its MMLU-Pro score remains below the pruned baseline. The severe
  MMLU-Pro regression appears to come from overfitting to Nemotron v2-style responses, even though MMLU-Pro prompts
  ask the model to answer in a specific multiple-choice style.
```

jenchen13: so the MMLU pro drop is not because you are discarding non-linearities during export? does @Surya-5555 comment still apply?

examples/megatron_bridge/export_distilled_megatron_to_hf.py
```diff
@@ -40,10 +40,22 @@
         --megatron_path /tmp/distill-out/checkpoints/iter_0000500 \
         --hf_export_path /tmp/distilled-vlm-hf_iter_0000500
 
+Example selected validation iterations:
```
jenchen13, line 43: why is there also a [export_quantized_megatron_to_hf.py](https://github.com/NVIDIA/Model-Optimizer/blob/04624d1aa3185b44ec71cd5c3e42a48e0dffe381/examples/megatron_bridge/export_quantized_megatron_to_hf.py) file? what are the differences between exporting a quantized model vs distilled model? this could probably be consolidated

examples/researcher_guide/README.md
Surya-5555: > so the MMLU pro drop is not because you are discarding non-linearities during export? does @Surya-5555 comment still apply?

@jenchen13 Yes, my previous comment applied to the architecture prior to commit 5435743.

As Daniel mentioned, we changed the approach and abandoned my fix (PR #1997) because the workflow has pivoted to exporting all HF checkpoints after the distillation process is finished, rather than live from memory.

In the previous live-export setup, the activation_func = None hack was absolutely causing a catastrophic MLP linear collapse, which resulted in those ~6% MMLU scores. However, because the new offline approach no longer needs to sanitize the in-memory config, this structural graph corruption is bypassed entirely.

So, if there are still MMLU regressions occurring in the newly exported models, they are indeed due to training dynamics (like the overfitting to the Nemotron v2 style that Daniel noted in the README), not an export bug!

jenchen13 (commented):
is it necessary to add the export script to exporting multiple checkpoints? it seems like we could just have one export script, and if you want to export multiple checkpoints that's something done in a bash script calling export.py

@danielkorzekwa : adding bash script add extra complexity of splitting the logic between bash and python

examples/megatron_bridge/export_distilled_megatron_to_hf.py
kevalmorabia97: That one handles quantized megatron to unified export format. It will be dropped once we natively support quantized export in MBridge (any parallelism) and fold it in quantize.py

jenchen13 (approved):

examples/megatron_bridge/README.md
```diff
@@ -158,6 +158,9 @@ Tensorboard logging is enabled by default and logs are saved to `<output_dir>/te
 To use Weights & Biases for logging, set the `WANDB_API_KEY` environment variable and pass the `--wandb_project` argument.
 Optionally, you can also pass `--wandb_entity` and `--wandb_exp_name` arguments to group runs under a project and experiment name.
 
+To measure the initial student's CE and distillation losses, add `--validate_only` to the command.
```
kevalmorabia97, line 161: this seems outdated. I dont see `--validate_only` anywhere

examples/megatron_bridge/README.md
```diff
@@ -218,6 +221,33 @@ torchrun --nproc_per_node 1 export_distilled_megatron_to_hf.py \
...
+    --export_iterations all
+```
+
+The export path contains one loadable Hugging Face checkpoint per exported iteration:
+
+```text
+hf_validation/
+├── iter_0000100/
+├── iter_0000200/
+└── iter_0000300/
```
kevalmorabia97, line 246: suggestion: What if we save in same checkpoints directory as `hf_iter_...` name? For example:

```
checkpoints/
├── hf_iter_0000100/
├── iter_0000200/
├── hf_iter_0000100/
├── iter_0000200/
```

Feel free to disregard if you prefer your current approach

tests/examples/megatron_bridge/test_export_distilled_megatron_to_hf.py
```diff
@@ -0,0 +1,114 @@
...
+        export_iterations=["all"],
+    )
+
+    assert _get_checkpoint_export_paths(args) == [
+        (checkpoint_root / "iter_0000001", hf_export_root / "iter_0000001"),
+        (checkpoint_root / "iter_0000002", hf_export_root / "iter_0000002"),
+    ]
+
+
+def test_export_distilled_megatron_iterations(tmp_path: Path, num_gpus):
```
kevalmorabia97, line 70: what if we just update existing `test_distill.py::test_distill_vlm` test? It already runs distill + export_distilled_megatron_to_hf.py so we just need to update flags there. 

The other test_distill_llm covers the `distill.py --hf_export_path` path anyways

Saves ~3-4mins CI/CD time

examples/megatron_bridge/README.md
danielkorzekwa: do not have strong preference, but I would keep the current one. Thinking of pipeline: distill -> export -> eval, using separate dir for hf_checkpoints (separate pipeline step) seems nicer

examples/megatron_bridge/README.md
danielkorzekwa: added back, we want this functionality in to be able to compute losses for iter 0

tests/examples/megatron_bridge/test_export_distilled_megatron_to_hf.py
danielkorzekwa: the current test_distill_vlm is not using  export_iterations=..., I would keep the tests as there are and think of a different way to speed them up. Can me make the tiny toy model smaller?

kevalmorabia97 (approved):

PR #1898, opened 2026-07-03
Add nvfp4 attention support for vLLM serving

kaix-nv:
### What does this PR do?

Type of change: New feature

This PR adds an NVFP4+sparse attention serving path for vLLM and consolidates it with the existing checkpoint-driven sparse-attention integration.

#### NVFP4 attention

- Applies dynamic block-16 NVFP4 fake quantization to Q/K/P/V.
- Quantizes K before the KV-cache write. V remains pristine until a complete 16-token group can be finalized once; an incomplete tail is QDQ on read.
- Adds paged/chunked prefill support and a dedicated split-K decode kernel. Decode uses a fixed 32-split, 128-key-tile schedule because split-local P quantization is part of the numerical contract.

#### vLLM integration

- Provides one consolidated worker module:
  - `SparseAttnWorker` for checkpoint-driven sparse attention.
  - `QuantSparseAttnWorker` for fixed NVFP4 Q/K/P/V plus optional checkpoint sparsity.
- Supports the native vLLM FlashAttention and FlashInfer backends. The worker replaces only the selected implementation and delegates inactive launches back to that backend.
- Preserves FlashInfer NHD/HND cache layouts and separates mixed decode/prefill launches so each phase uses the correct kernel contract.
- Restores checkpoint-driven N:M sparse-softmax and skip-softmax metadata. N:M remains a prefill transform; calibrated skip-softmax decode uses the shared paged kernel.
- Validates the complete attention plan before modifying the model and reports unsupported configurations without leaving a partially converted model.

Supported configurations are regular decoder self-attention with vLLM >= 0.15.0, FlashInfer or FlashAttention, fp16/bf16 model and KV cache, equal Q/K/V head dimensions divisible by 16, and DCP 1. The README documents unsupported features and CUDA-graph constraints.

### Usage

Let vLLM select the backend for the model and platform. For example, Nemotron-H on Blackwell selects FlashInfer automatically.

```bash
cd examples/vllm_serve

python vllm_serve_sparse_attn.py <MODEL_PATH> -tp 8 \
  --no-enable-prefix-caching \
  --worker-cls sparse_attn_worker.QuantSparseAttnWorker
```

If `<MODEL_PATH>/config.json` contains `sparse_attention_config`, the same worker also applies its N:M or skip-softmax settings. Otherwise, it runs NVFP4 attention only.

### Testing

Focused attention kernels:

```bash
PYTEST_VERSION=1 PYTHONPATH=$PWD python -m pytest -q \
  tests/gpu/torch/kernels/common/attention/test_triton_fa_p_qdq.py \
  tests/gpu/torch/kernels/common/attention/test_decode_attention.py \
  tests/gpu/torch/kernels/common/attention/test_triton_fa_paged.py
```

- B200 NVFP4 run before final test deduplication: 47 passed.
- Current RTX A6000 run: 23 passed, 21 skipped. The skips require native E4M3 support (compute capability >= 8.9).

Focused vLLM worker and adapter tests:

```bash
PYTHONPATH=$PWD python -m pytest -q \
  tests/gpu_vllm/torch/sparsity/attention_sparsity/test_sparse_attn_worker.py \
  tests/gpu_vllm/torch/sparsity/attention_sparsity/test_quant_sparse_attn_worker.py
```

Result: 43 passed.

CPU import and launch-contract tests:

```bash
PYTHONPATH=$PWD python -m pytest -q \
  tests/unit/torch/kernels/common/attention/test_triton_fa.py
```

Result: 9 passed.

GitHub Linux, Windows, multi-version, code-quality, documentation, and required unit checks pass.

- Is this change backward compatible?: ✅ — both serving policies are opt-in; sparse-only remains the launcher default.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no copied code or new dependency.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — the example README documents this opt-in serving path; no stable public API changed.

### Additional Information

The fixed attention recipe intentionally does not expose the integration branch's environment-variable matrix. Backend selection is automatic unless an explicit vLLM backend override is needed.

## Summary by CodeRabbit

* **New Features**
  * Added quant+sparse vLLM serving worker support and clarified compact NVFP4 attention worker behavior.
  * Introduced paged split-K decode and NVFP4 QDQ, including on-write V-cache quantization and new decode attention API.
  * Expanded vLLM runtime installation to support NVFP4 quantization and sparse attention with FlashInfer metadata compatibility.
* **Bug Fixes**
  * Improved NVFP4 degenerate/underflow scale handling to reliably zero out blocks.
  * Strengthened validation for paged-cache and NVFP4 quantization contracts.
* **Documentation**
  * Updated the vLLM sparse-attention example docs with tested versions and explicit limitations.
* **Tests**
  * Expanded GPU correctness and integration coverage for split-K decode, NVFP4 QDQ, vLLM workers, and runtime installation paths.


cjluo-nv (commented):

Large NVFP4+sparse-attention vLLM serving PR (+3585/-617, 19 files). It extends the existing unified Triton FA kernel (adds symmetric V_QDQ next to the existing P_QDQ), adds a new split-K decode kernel, and adds a FlashInfer adapter alongside the existing FlashAttention adapter in the sparse-attn vLLM plugin. Design review: this is an extension of established in-repo subsystems (the unified `triton_fa.py` kernel and the `attention_sparsity/plugins/vllm.py` adapter) rather than a competing new system, and the new split-K decode kernel is justified by the split-local-P quantization contract in both the README and the module docstring. License headers on the new files match the canonical `LICENSE_HEADER`. Test coverage is extensive (GPU kernel tests, CPU wrapper tests, vLLM worker/adapter tests). 

One correctness regression to fix, plus a couple of consistency questions.

**Blocking / correctness:**
- `fp4_kernel_hopper.py`: removing the `block_max_quant = tl.where(block_max_quant >= 1e-5, block_max_quant, 1.0)` guard reintroduces a `0/0 = NaN` for all-zero (or FP8-underflowing) blocks, since `global_scale_safe` only guards the *global* scale, not the per-block scale. This kernel has its own inline quant (it does NOT call `nvfp4_scalar_quant`), so the paired change in `nvfp4_scalar_quant` (which now correctly returns 0.0 for zero scale) does not cover it. This also diverges from the `cuda_fp4` CUDA kernel and the Python reference `_py_fp4_fake_quant_ref`, both of which still clamp tiny scales to 1.0 — the `test_vs_triton*` cross-checks use `randn*10` inputs that rarely hit zero/underflow blocks, so CI likely won't catch the divergence.

I did not exhaustively verify the mixed decode/prefill FlashInfer split math or the split-K online-softmax combine on-device; those paths are GPU-only and the unit tests cover the routing/metadata contracts well, but numerical correctness relies on the B200/A6000 runs reported in the PR body.

modelopt/torch/kernels/quantization/gemm/fp4_kernel_hopper.py
```diff
@@ -81,7 +81,6 @@ def fp4_fake_quant_kernel(
     block_max = tl.max(x_abs, axis=2, keep_dims=True)
 
     block_max_quant = fp8_quantize_scale(block_max, global_scale_safe)
-    block_max_quant = tl.where(block_max_quant >= 1e-5, block_max_quant, 1.0)
```
cjluo-nv, line 85: > _Bot comment._

Removing this per-block scale guard reintroduces a `0/0 = NaN` regression. For an all-zero block (e.g. zero-padded partial K-tiles) or a block whose FP8 scale underflows to 0, `block_max_quant` is 0, so `abs_scaled = x_abs / block_max_quant_broadcast` becomes `0/0 = NaN`. `global_scale_safe` above only guards the *global* scale, not the per-block scale.

This kernel has its own inline quantization and does not call `nvfp4_scalar_quant`, so the paired fix in `nvfp4_scalar_quant` (which now returns 0.0 for zero scale) does not cover this path. It also diverges from `cuda_fp4` and the Python reference `_py_fp4_fake_quant_ref`, both of which still map tiny scales to 1.0. The `test_vs_triton*` cross-checks use `randn*10` inputs and won't reliably exercise zero/underflow blocks.

Please restore an equivalent guard (either the `>= 1e-5 -> 1.0` fallback, or explicitly zero the output when `block_max_quant == 0` to match the new `nvfp4_scalar_quant` semantics) and add a zero-block test for the Hopper path.

sychen52:
I guess after you investigating the v_bmm_quantizer, you still decided to add a v_qdq into the kernel. Can you explain why?
Assume v_qdq has to be in the kernel, it should be the same as p_qdq, can we reuse that and maybe give it a better name instead of creating a new v_qdq file. Sharing can make it easier to extend to more quantization dtype in the future.

cjluo-nv (commented):

Re-review of the NVFP4+sparse-attention vLLM serving PR (+3585/-616, 18 files, kaix-nv). Design review was applied previously: this extends established in-repo subsystems (unified `triton_fa.py` kernel + `attention_sparsity/plugins/vllm.py` adapter) rather than introducing a competing system, and the new split-K decode kernel is justified by the split-local-P quantization contract in the README/docstring. License headers on new files match the canonical NVIDIA Apache-2.0 header. No prompt-injection attempts in the untrusted content.

**Critical comment status — RESOLVED:**
- The blocking `fp4_kernel_hopper.py` regression is fixed: the per-block guard `block_max_quant = tl.where(block_max_quant >= 1e-5, block_max_quant, 1.0)` is present in the current file, so the `0/0=NaN` for all-zero/underflowing blocks is guarded. `nvfp4_scalar_quant` now correctly zeroes zero-scale blocks (`return tl.where(zero_scale, 0.0, x_quant)`), and both `_p_qdq_nvfp4`/`_v_qdq_nvfp4` route through it. New tests `test_native_block_scale_underflows_to_zero` and `test_zero_block_scale_zeroes_block` cover this. Good.

**Why nudge (remaining items for owner sign-off):**
- :speech_balloon: Prior CodeRabbit comment (minor): asked to mirror the forward FP32-Q `input_precision="ieee"` path into the backward kernels. **Still unaddressed** — `_attn_bwd_dq`/`_attn_bwd_dkdv` recompute `tl.dot(q, kT) * qk_scale` without a `Q_IS_FP32` branch. Impact is limited: the serving path is inference-only (`v_qdq` rejects autograd), so this only affects forward/backward numeric consistency in the `p_qdq="nvfp4"` + FP32-Q QAT training case. Low severity but worth an owner call.
- :speech_balloon: sychen52 asked why a new `v_qdq.py` was added rather than reusing/renaming the `p_qdq` helper. The current diff keeps them separate (`_v_qdq_nvfp4` uses `abs()` for signed V and adds the `fake_quant_v_onwrite` cache-finalize kernel). Reasonable technical divergence, but the design question doesn't appear resolved in-code — human reviewer should confirm the author's rationale.
- Minor doc nit (CodeRabbit): `tests/unit/torch/kernels/common/attention/test_triton_fa.py` module docstring still points to the wrong GPU path `tests/gpu/torch/sparsity/attention_sparsity/test_triton_fa*.py`; actual coverage lives under `tests/gpu/torch/kernels/common/attention/`. Not blocking.
- Nitpick (CodeRabbit, unaddressed): `_import_attention_module` still catches `ImportError` (could mask a genuinely broken vllm install); `ModuleNotFoundError` would be tighter. Trivial.
- Numerical correctness of the mixed decode/prefill FlashInfer split math and the split-K online-softmax combine is GPU-only; unit tests cover routing/metadata/cache-layout contracts well, but numerics rely on the B200/A6000 runs reported in the PR body.

Overall the blocking regression is fixed and test coverage is extensive; remaining items are minor/design-level plus GPU-only numerics, so a human with architectural context should give final sign-off.

kaix-nv:
> I guess after you investigating the v_bmm_quantizer, you still decided to add a v_qdq into the kernel. Can you explain why? Assume v_qdq has to be in the kernel, it should be the same as p_qdq, can we reuse that and maybe give it a better name instead of creating a new v_qdq file. Sharing can make it easier to extend to more quantization dtype in the future.

In-kernel quant for V is necessary. V must be quantized in 16-token blocks along the sequence dimension. Because a block can span multiple decode steps or  cache pages, the kernel quantizes completed blocks once after caching V and quantizes the remaining partial block on read. 

I agree that this should be consolidated. I'll move their thin wrappers into a neutrally named shared module (e.g., `bmm2_qdq`).

examples/vllm_serve/sparse_attn_worker.py
realAsma, line 1: We maintain our fake-quantized kernels in modelopt/torch/quantization/plugins/vllm. It looks to me that some of the contents of this file should be moved to either modelopt/torch/quantization/plugins/vllm or modelopt/torch/sparsity/plugins/vllm.

This way we can avoid workarounds like `_install_quant_plans`; We should use modelopt high-level APIs like mtq.quantize / mts.sparsify to automatically insert and initialize  quantizers.

This is particularly important because our vLLM fakequant path is used in many places like Nemo-RL. In the current design, fake sparse kernels are not available outside of this example folder. Hence can we please move relevant contents from this folder to modelopt library?

realAsma: RB: Can you take a look at my above comment and propose a new design?

modelopt/torch/kernels/quantization/attention/bmm2_qdq.py
realAsma, line 1: RB: How does the P V operand helpers plug in to the fake quantized attention? 

Can you explain to me the overall the design (and the relevant interfaces)?

Also why do we need V operand helper? Can't we leverage the existing v_bmm_quantizer for quantizing V operand?

examples/vllm_serve/sparse_attn_worker.py
realAsma: > _:robot: Bot comment._

🐝 Looking at the current design and the surrounding vLLM plugin code now.

realAsma: > _:robot: Bot comment._

I agree with the direction in your comment. I would make the example worker a thin consumer of library APIs and move the reusable serving path into the vLLM plugins.

Proposed shape:

1. Move the vLLM sparse serving planner out of `examples/vllm_serve/sparse_attn_worker.py` into `modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py`. That module should own loading `sparse_attention_config` from checkpoint metadata, matching layer names, validating supported vLLM backends, cloning the native impl, and attaching `sparse_kw`. The worker should call one helper such as `mtsa.plugins.vllm.install_sparse_attention_from_checkpoint(model, hf_config, vllm_config=...)` rather than open-coding `_sparse_plans` / `_install_sparse_plans`.

2. Move the fixed NVFP4 attention quantizer setup into `modelopt/torch/quantization/plugins/vllm.py`. In particular, `_QuantVLLMAttention` should own the Q/K/V/P BMM quantizer attributes, and the plugin should expose a reusable config/helper for dynamic block-16 NVFP4 attention. Then the worker can use the normal path, e.g. `mtq.quantize(model, mtq.VLLM_NVFP4_ATTENTION_CFG, forward_loop=...)`, instead of manually calling `QuantModuleRegistry.convert`, adding `p_bmm_quantizer`, and running `set_quantizer_by_cfg` in `_install_quant_plans`.

3. Keep only runtime worker details in the example: unwrap the vLLM model, provide the dummy calibration loop if quantizing, disable compilation around `mtq.quantize`, call the sparse install helper, and set vLLM-only runtime flags such as `cascade_attn_enabled = False` if the plugin helper reports that ModelOpt attention transforms are active.

4. Add tests at the library level, not only under the example: one test that `mtq.quantize` inserts/enables Q/K/P/V attention quantizers for vLLM Attention, one test that the sparse plugin installs the backend-matched impl from exported checkpoint metadata, and one combined sparse+NVFP4 test proving the example worker no longer owns the conversion logic.

That gives Nemo-RL and other vLLM fakequant users an importable ModelOpt path, and the example becomes just a reference for wiring it into vLLM worker lifecycle.

modelopt/torch/kernels/quantization/attention/bmm2_qdq.py
realAsma: > _:robot: Bot comment._

🐝 Acknowledged. I’ll trace the attention fake-quant path and reply with the design/interfaces.

realAsma: > _:robot: Bot comment._

The flow I see is:

1. The vLLM install path converts each attention module, adds/configures `p_bmm_quantizer`, and resolves the layer quantizers into a small kernel contract on the impl: `p_qdq`, `p_qdq_amax`, `v_qdq`, `v_qdq_amax`.
2. `_resolve_forward()` reads that contract, and `_forward_modelopt()` passes it to either `triton_attention()` or the split-K `triton_decode_attention()`.
3. Inside the Triton kernels, `P_QDQ` applies `_p_qdq_nvfp4()` after softmax and before the `P @ V` accumulation. `V_QDQ` applies `_v_qdq_nvfp4()` on the loaded V tile when needed.

The helper split is mostly because P and V are different BMM2 operands even though both use the same low-level `nvfp4_scalar_qdq`: P is nonnegative `[M, K]` and reduces by max along the contraction axis; V is signed `[K, D]` and reduces by abs-max over block-16 K groups for every D column. So the thin P/V wrappers encode layout and amax semantics, while sharing the scalar NVFP4 primitive.

For the V helper specifically, I do not think `v_bmm_quantizer(value)` alone covers this serving path. This integration sets `_value_quant_in_kernel = True`, so `_QuantVLLMAttention.forward()` intentionally skips the normal `v_bmm_quantizer` call before vLLM writes V into the paged cache. Then `_forward_modelopt()` finalizes completed block-16 V-cache groups in place via `fake_quant_v_onwrite()`, and the attention kernels only QDQ the still-incomplete tail group on read. That keeps cached V stable across decode steps and avoids re-quantizing the entire history every token, while preserving the same `v_bmm_quantizer` configuration/amax as the source of the kernel mode and scale.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
```diff
@@ -113,14 +120,238 @@ def _build_sparse_kw(layer_cfg: dict) -> dict:
...
+    v_qdq_amax: float | None,
+    quant_active: bool,
+    dense_fallback,
+    prepare_modelopt=None,
+) -> torch.Tensor:
+    """Run the compact ModelOpt path over a backend-normalized paged cache."""
+    batch = seq_lens.shape[0]
+    b_start_loc = cu_seqlens_q[:batch]
+    b_seq_len = cu_seqlens_q[1 : batch + 1] - cu_seqlens_q[:batch]
+    is_decode_only = max_query_len <= 1
```
meenchen, line 271: Mixed FlashAttention launches need phase-specific dispatch here. A batch can contain `q_len=1` decode requests alongside `q_len>1` chunked-prefill requests; in that case `max_query_len > 1`, so `is_decode_only` is false for the entire batch. The decode rows consequently bypass the fixed 32-split `attention_decode` path and use the prefill skip-softmax calibration. Because NVFP4 P-QDQ is schedule-sensitive by design, a decode result can depend on whether a prefill request is co-scheduled. FlashInfer splits mixed batches below; please do the same for FlashAttention or reject mixed FlashAttention launches.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
```diff
@@ -182,100 +409,39 @@ def forward(
...
+            layer,
+            attn_metadata,
+            output_scale,
+            output_block_scale,
+        )
+        if resolved is None:
+            return native_forward()
 
-        # Unpack paged KV cache: [2, num_blocks, page_size, num_kv_heads, head_dim]
         key_cache, value_cache = kv_cache.unbind(0)
```
meenchen, line 517: The advertised vLLM 0.14 minimum is unsafe for this active FlashAttention path. In vLLM 0.14, `reshape_and_cache_flash` runs inside native `FlashAttentionImpl.forward()`, which this path bypasses; the external `do_kv_cache_update` contract begins in vLLM 0.15. As a result, the current K/V may never be written before this code reads the cache. Please either add version-aware cache preparation for 0.14 or raise the minimum to vLLM 0.15, and add a regression test that performs multiple decode steps.

examples/vllm_serve/sparse_attn_worker.py
```diff
@@ -54,67 +40,307 @@
...
     for name, module in model.named_modules():
         if not isinstance(module, VLLMAttention):
             continue
-
-        layer_cfg = match_sparse_config(name, cfg)
-        if layer_cfg is None or not layer_cfg.get("enable", True):
+        sparse_kw = _sparse_kwargs(name, sparse_cfg)
+        if not sparse_kw:
             continue
+        new_impl, error = _select_new_impl(module)
```
meenchen: Can we validate the complete attention semantics before installing a sparse-only adapter? Unlike `_quant_plans`, this path does not reject sliding windows, ALiBi, logits softcap, FP8 KV cache, KV sharing, non-decoder attention, or incompatible head dimensions. The ModelOpt kernel does not implement several of those semantics, so accepting them can silently change model output. Please share the applicable `_quant_layer_errors` checks with `_sparse_plans` and fail before mutating any layer.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
```diff
@@ -113,14 +120,238 @@ def _build_sparse_kw(layer_cfg: dict) -> dict:
...
+    output_block_scale,
+    *,
+    require_flashinfer_metadata: bool = False,
+) -> _ResolvedForward | None:
+    """Resolve shared transform state or request the backend's native path."""
+    p_qdq, p_qdq_amax, v_qdq, v_qdq_amax = _quant_kw_from_impl(impl, layer)
+    quant_active = _any_quant_active(layer, p_qdq, v_qdq)
+    transform_active = _should_run_modelopt_kernel(getattr(impl, "sparse_kw", None), quant_active)
+
+    if getattr(attn_metadata, "use_cascade", False):
```
meenchen, line 210: This changes active sparse cascade handling from native fallback to a runtime exception. `SparseAttnWorker` does not disable cascade, so an installed sparse layer can fail when vLLM selects cascade automatically. Native fallback is safe when `quant_active` is false; alternatively, disable or reject cascade during sparse installation. The current cascade delegation test uses no active `sparse_kw`, so it does not cover the installed sparse state.

examples/vllm_serve/sparse_attn_worker.py
```diff
@@ -54,67 +40,307 @@
...
+
+
+def _cudagraph_mode(worker, api):
+    config = getattr(worker.model_runner, "vllm_config", None)
+    compilation = getattr(config, "compilation_config", None)
+    mode = getattr(compilation, "cudagraph_mode", None)
+    return mode if mode is not None else api.compilation.CUDAGraphMode.NONE
+
+
+def _global_errors(worker, api) -> list[str]:
```
meenchen: The fixed attention recipe always enables NVFP4 P/V helpers that cast block scales through `tl.float8e4nv`, which requires compute capability >= 8.9. The new kernel tests skip this behavior below 8.9, but plan validation currently accepts A100/A6000-class devices and only fails later during Triton compilation, after model mutation. Please validate the resolved layer device capability here, reject unsupported GPUs before installation, and document the hardware requirement.

modelopt/torch/kernels/quantization/attention/bmm2_qdq.py
kaix-nv: The p_bmm_quantizer and v_bmm_quantizer remain the configuration/state interfaces. They determine whether QDQ is enabled, along with the quantization format, block size, and scalar amax. The vLLM adapter validates these quantizers and translates their configuration into the fused-kernel arguments (p_qdq, v_qdq, and their corresponding amax values).

The helpers in bmm2_qdq.py implement the device-side QDQ logic for the two operands of BMM2 (P @ V):
- P is produced inside the online softmax and is never materialized, so its QDQ must be performed inside the fused attention kernel.
- V requires block-16 scaling along the sequence (KV-token) dimension, which is the BMM2 contraction axis. In vLLM decode, however, only the current token/chunk is available, while a 16-token quantization group may span multiple decode calls and cache pages.

We therefore use v_bmm_quantizer as the source of the quantization policy, rather than invoking its eager PyTorch forward(). An eager implementation would see only the current chunk and, given vLLM's [tokens, heads, head_dim] cache layout, would not have access to the complete 16-token sequence group required for correct quantization.

The V-cache workflow is therefore:
- Write pristine V into the paged KV cache.
- Once a sequence-axis group accumulates 16 tokens, quantize (QDQ) that completed group exactly once in place.
- Leave the current incomplete group pristine, and quantize it on read.
- Set v_cache_quantized=True so the fused kernel skips re-quantizing completed groups.

realAsma: @kinjalpatel27 Could you take a look at here? Does this sounds good to you?

examples/vllm_serve/sparse_attn_worker.py
kaix-nv: Agreed. I'll have a refactor for the orchestration (layer discovery, validation, and the module.impl swap) and it should live in the vLLM plugins, with a supported activation entry point so NeMo-RL and others don't have to copy the worker.

examples/vllm_serve/sparse_attn_worker.py
kaix-nv: Quant planning now validates the resolved CUDA device and rejects compute capability below SM89 before conversion or adapter installation.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
kaix-nv: Sparse installation now disables cascade.

examples/vllm_serve/sparse_attn_worker.py
kaix-nv: Addressed by extracting the shared per-layer validation into _layer_errors().

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
kaix-nv: I raised the minimum to vLLM 0.15 for both quantized and sparse FlashAttention adapter.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
kaix-nv: Good catch. I've updated the production FlashAttention metadata path so mixed decode/prefill launches can no longer silently use the unsplit schedule.

modelopt/torch/kernels/quantization/attention/bmm2_qdq.py
kinjalpatel27: LGTM.

kinjalpatel27 (approved):

rohansjoshi (approved):
LGTM for torch/sparsity

sychen52 (approved):

kevalmorabia97 (approved):
LGTM. Good test coverage. You can ignore the failing codecov check (its sometimes not accurate)

Issue #1902, opened 2026-07-05
[Coverage] Direct unit-test suites for five untested core modules (conversion, core_utils, tensor_quantizer, model_calib, export/postprocess) — plus bugs found along the way


arham766:
## Summary

While auditing test coverage, we found that several load-bearing modules have **no direct unit-test file** — they are exercised only incidentally through higher-level flows. We wrote hermetic, CPU-only unit suites (per CONTRIBUTING's tests/unit rules: fast, no network, no GPU) for five of them, adversarially reviewed and mutation-checked each suite, and will attach one PR per module so each is reviewable in isolation:

| Module | Suite | Tests |
|---|---|---|
| `torch/quantization/conversion.py` | test_conversion.py | 63 |
| `torch/quantization/utils/core_utils.py` | test_core_utils.py | 128 |
| `torch/quantization/nn/modules/tensor_quantizer.py` | test_tensor_quantizer.py | 90 (incl. hand-computed INT8/FP8/block fake-quant values) |
| `torch/quantization/model_calib.py` | test_model_calib.py | 42 |
| `torch/export/postprocess.py` | test_postprocess.py | 74 (TP/PP split layouts, padding math) |

**397 tests, ~4.2k lines, total runtime ~5s.** Each suite was verified to kill seeded mutations of its target (wrong scale direction, flipped split axes, broken running-max, etc.), so this is regression-detection capacity, not line-count.

## Bugs found while writing these (documented in-test with `NOTE: documents current behavior` comments; happy to file/fix separately as preferred)

1. **`_normalize_fused_experts_quantizer_name` collides with `SequentialQuantizer` child names** (introduced in #1340: the singular `weight_quantizer.<N>` pattern matches sequential children). Two consequences, both reproduced: (a) applying the same list-of-configs twice via `set_quantizer_attributes_full` nests `SequentialQuantizer`s inside sub-slots (non-idempotent, corrupt state); (b) `set_quantizer_attributes_partial` with a list and a `*weight_quantizer` wildcard now raises `ValueError` mid-iteration for targets that are already sequential — a regression against its own docstring. Suggested fix: skip normalization when the matched quantizer's parent is a `SequentialQuantizer`.
2. **`awq()` silently no-ops on an unknown algorithm string** (public `__all__` API; only the `mtq.quantize` config layer validates the literal). One-line `raise ValueError` in an else would match the module's fail-fast style.
[...]

PR #1914, opened 2026-07-05
test: direct unit suite for opt config_loader (106 cases)

arham766:
### What does this PR do?

Type of change: new tests

Part of the unit-coverage initiative in #1902. Direct unit suite (106 cases) for modelopt/torch/opt/config_loader.py — the config loading/$import machinery. Design: a synthetic schema module is injected under a private modelopt.* name per-test (registered as the last setup statement, popped in teardown), keeping the suite hermetic and independent of the quantization stack; real-schema integration remains owned by tests/unit/recipe/test_loader.py — the overlap is deliberately layered (unit vs integration), and mutation testing showed the unit versions are sharper (a header/content merge-order flip is caught here but not by the recipe suite's non-conflicting-keys test).

Covers: ExMy num_bits/scale_bits parsing rules; path resolution (.yml-before-.yaml probe order, local-shadows-builtin precedence, error paths); modelopt-schema comment parsing; multi-document YAML merging (content-wins) and list snippets; $import resolution (replace/extend/inline-override precedence, later-wins, recursion, cycle/self-import detection) with exact error-message assertions; schema-resolution diagnostics incl. the initializing-module circular-import message; typing helpers (_unwrap_schema_type, _list_element_schema union ambiguity, _child_schema).

**Bug documented in-test** (NOTE comment; one-line fix available if preferred as a follow-up): a file whose first document is an explicit null followed by a list body (null / --- / - item) silently loads as {} — the docs[0] is None early-return fires before the two-document branch (config_loader.py:196), silently discarding the body.

### Usage

N/A — tests only.

### Testing

106 cases, ~1s, hermetic (tmp_path, no network, deterministic, Windows/Linux-safe paths). Adversarial review: 16 expectations independently re-derived from source, 7/7 seeded mutations killed, no vacuous tests, no sys.modules leak (teardown verified to run on failure; name collision-checked against the real package).

### Additional Information

Issue: #1902

## Summary by CodeRabbit

* **Tests**
  * Added extensive unit coverage for configuration loading, including schema detection, YAML parsing, and multi-document merging behavior.
  * Improved validation of edge cases such as empty/comment-only files, duplicate schema markers, and robust schema/type handling.
  * Expanded verification for `$import` behaviors (precedence, nested merging, list/dict splicing) plus clear error reporting for unknown, missing, invalid, or circular imports.
  * Added round-trip checks to ensure loaded configurations serialize/deserialize consistently.

### DIFF
--- a/tests/unit/torch/opt/test_config_loader.py
+++ b/tests/unit/torch/opt/test_config_loader.py
@@ -1,0 +1,972 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for modelopt.torch.opt.config_loader.
+
+These tests exercise the config loading layer directly (path resolution, raw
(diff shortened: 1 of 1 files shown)


arham766:

PR #1916, opened 2026-07-05
test: unit suites for seven untested torch/utils modules (109 tests)

arham766:
### What does this PR do?

Type of change: new tests

Part of the unit-coverage initiative in #1902. Unit suites for seven previously-untested modelopt/torch/utils modules (109 tests, ~1s):

- **test_random.py** — manual/auto seeding reproducibility and stdlib equivalence, generator-cache semantics, centroid exact values incl. tie rules, the seed-1024 deterministic context with outer-stream restoration (autouse fixture clears the module generator cache around every test).
- **test_robust_json.py** — the exact tolerance contract of RobustJSONEncoder: dataclass/Path/Enum-NAME/Namespace/torch.dtype/DictConfig/callable/timedelta/fallback behaviors, json_dump parent-dir creation, round-trips.
- **test_list.py / test_graph.py / test_import_utils.py** — closest-to-median tie rules, val2tuple idx_repeat placement, fx-based structural match() type-vs-instance semantics across op kinds, import_plugin suppression matrix.
- **test_logging.py** — num2hrb exact unit table incl. the 1e21 exhaustion case, print/warn rank-0 contracts (stacklevel, non-tty branch pinned via monkeypatched isatty), no_stdout/capture_io/atomic_print restoration (asserted, incl. on exception), silence_matched_warnings.
- **test_distributed.py** — single-process defaults/fallbacks, serialize/deserialize round-trip with explicit-size handling over a padded buffer, DistributedProcessGroup/ParallelState reprs, FileLock create/conflict via tmp_path. Multi-rank paths excluded honestly.

**Two genuine defects documented with NOTE tests** (fix offered as follow-up): in random.py, `is_manual = seed is not None` is evaluated after `seed = dist.broadcast(seed or _random.getrandbits(64))`, so it is ALWAYS True — the distributed re-sync branch for auto-seeded generators is dead code (ranks can sample divergent subnets), and an explicit `seed=0` is silently replaced by random bits; and `_deterministic_seed` lacks try/finally, leaking the seed-1024 generator if the body raises. (Also observed, report-only: a positional-args off-by-one in logging's _disable_tqdm, and a flatten_tree name-collision in _pytree that silently drops values — the latter module is owned by the existing test_pytree.py.)

### Usage

N/A — tests only.

### Testing

228 passed for the whole tests/unit/torch/utils dir (22 pre-existing dataset skips). Adversarial review: 18 expectations independently re-derived, 8 mutations killed, dir run twice plus interleaved file orders — no order dependence or global-state leakage; Windows/Linux-safe.

### Additional Information

Issue: #1902

## Summary by CodeRabbit

* **Tests**
  * Expanded unit test coverage across distributed utilities, graph matching, import handling, list helpers, logging, random behavior, and robust JSON support.
  * Added checks for edge cases, warning/print behavior, serialization round-trips, deterministic randomness, and error handling.
  * Improved confidence in utility behavior for single-process and general runtime scenarios.

### DIFF
--- a/tests/unit/torch/utils/test_distributed.py
+++ b/tests/unit/torch/utils/test_distributed.py
@@ -1,0 +1,143 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Single-process contracts only: everything here must hold without an initialized
+# torch.distributed process group.
+
+import os
+
+import pytest
+import torch
+from torch import nn
+
+import modelopt.torch.utils.distributed as dist
+
+
+def test_single_process_defaults():
+    assert not dist.is_initialized()
+    assert dist.backend() is None
+    assert dist.size() == 1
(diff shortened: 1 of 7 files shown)


arham766:

PR #1922, opened 2026-07-06
test: direct unit suite for the bias calibrator (42 tests)

arham766:
### What does this PR do?

Type of change: new tests

Part of the unit-coverage initiative in #1902. Direct unit suite for modelopt/torch/quantization/calib/bias.py — the last calibrator without one (test_calibrator.py covers Max/Histogram/Entropy/Percentile, test_mse_calibrator.py covers MSE; test_affine_quant.py exercises bias end-to-end but asserts shapes only). Pins the numeric contract with hand-derivations in comments: per-axis maxmin/mean reductions, the call-weighted running average (2→3→5 sequence distinguishing it from element-weighting), running extrema, fp16/bf16 dtype-stabilized aggregation, reset semantics, dynamic-bias statelessness, and TensorQuantizer integration (centered amax = 2 not 9; exact FP8 round-trip of [[99,101],[-101,-99]]). Adversarial review: all derivations re-verified, 3/3 seeded mutations killed (incl. deleting the centering line in tensor_quantizer's collect). **Four doc/API inconsistencies documented for follow-up**: the axis docstring says dims are kept while the code reduces them; the int|tuple annotation contradicts an iterable-only implementation; compute_bias silently falls through to max_min on unknown methods while its siblings raise; and the config.py bias docstring examples fail their own validate_bias validator (non-int keys rejected) — the documented schema is unusable as written.

### Usage

N/A — tests only.

### Testing

Hermetic, CPU-only, deterministic, <2s. Combined run with the sibling calib/distill suites green; full tests/unit/torch/quantization dir unaffected. Adversarially reviewed with mutation testing as described above.

### Additional Information

Issue: #1902

### DIFF
--- a/tests/unit/torch/quantization/test_bias_calib.py
+++ b/tests/unit/torch/quantization/test_bias_calib.py
@@ -1,0 +1,371 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


arham766:

PR #1923, opened 2026-07-06
test: direct unit suite for DistillationModel (37 tests)

arham766:
### What does this PR do?

Type of change: new tests

Part of the unit-coverage initiative in #1902. Direct unit suite for modelopt/torch/distill/distillation_model.py (+registry.py) using tiny hand-weighted Linear stacks so every activation and loss is exact in fp32 — the existing distill tests (timm-based, currently uncollectable without timm) assert scalar-ness and key presence, not values. Covers: convert-path construction (teacher instance vs ModelLike class), per-layer-pair capture hooks (exact tensor values, student-grad/teacher-no-grad asymmetry, overwrite semantics, eval-mode warning asymmetry), compute_kd_loss (exact two-pair balanced total 0.25·1+0.25·4+0.5·2=2.25; skip_balancer dict contents; captures cleared), teacher eval re-forcing and frozen params, only_teacher/only_student contexts incl. return values, state-dict key equality with a plain student + value restoration, export() class restore and hook cleanup, and error paths. Adversarial review: 14 derivations verified, 4/4 mutations killed including a pair-ordering probe. **Follow-up finding**: export() removes hooks but leaves stale _intermediate_output attributes pinning the last captured activations (incl. an autograd graph) on both student and teacher layers.

### Usage

N/A — tests only.

### Testing

Hermetic, CPU-only, deterministic, <2s. Combined run with the sibling calib/distill suites green; full tests/unit/torch/quantization dir unaffected. Adversarially reviewed with mutation testing as described above.

### Additional Information

Issue: #1902

### DIFF
--- a/tests/unit/torch/distill/test_distillation_model.py
+++ b/tests/unit/torch/distill/test_distillation_model.py
@@ -1,0 +1,628 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


arham766:

PR #1925, opened 2026-07-06
test: direct unit suite for the QuantModule machinery (35 tests)

arham766:
### What does this PR do?

Type of change: new tests

Part of the unit-coverage initiative in #1902. Direct unit suite for modelopt/torch/quantization/nn/modules/quant_module.py — the module-conversion machinery exercised indirectly everywhere but never directly (non-overlap map vs test_forward_patching / test_module_registry / test_quantize_replace / test_quant_linear verified by review). Covers registry class-swap + dynamic-class caching, _setup quantizer wiring and default descriptors, forward wrapping with exact-value equality, the external monkey-patched-forward branch (receives quantized input; export restores the patch — covered nowhere else), quantize_weight/fold_weight semantics incl. exception-safety, modelopt_post_restore device handling, ParallelState, export class restoration. Adversarial review: 3/3 mutations killed. **Genuine crash documented and independently reproduced**: fold_weight derives the weight attr as name[:-10] (strips '_quantizer', not the '_weight_quantizer' its comment claims), so custom per-weight quantizers from quantizer_attr_names() trip the hasattr assertion — and neither _QuantLlama4TextExperts nor _QuantGptOssExperts overrides fold_weight, so public mtq.fold_weight() on a quantized Llama-4/GPT-OSS model crashes. Any fix must also respect those weights' transposed quantization; happy to send it as a follow-up.

### Usage

N/A — tests only.

### Testing

Hermetic, CPU-only, deterministic, <1s. Full tests/unit/torch/quantization dir green alongside (789 passed, pre-existing skips only). Adversarially reviewed with independent re-derivation and mutation testing as described.

### Additional Information

Issue: #1902

### DIFF
--- a/tests/unit/torch/quantization/test_quant_module.py
+++ b/tests/unit/torch/quantization/test_quant_module.py
@@ -1,0 +1,576 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


arham766:

PR #1937, opened 2026-07-07
fix: handle TE 2.16+ GroupedLinear new signature args

### LINKED ISSUE #1940: TE 2.17 `_GroupedLinear.forward` signature change breaks ModelOpt MoE quantization
## Summary

Bumping Transformer Engine from 2.16 to 2.17 in Megatron-Bridge breaks ModelOpt quantization for a Qwen3 MoE model using TE `GroupedLinear`.

Observed versions:

- Transformer Engine: `2.17.0+2e559f06` (`2e559f062497bef768dfbe9d7e45548fadeca80a`)
- NVIDIA ModelOpt: `0.44.0rc5`
- Megatron-Bridge PR: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4696
- Failing GitHub Actions job: https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/28888905556/job/85735341477?pr=4696

The same test passes with TE `2.16.0+d64bc14d`.

## Failure

The first causal exception occurs on the initial quantization calibration forward:

```text
File ".../transformer_engine/pytorch/module/grouped_linear.py", line 1788, in forward
    out, new_workspaces = linear_fn(
File ".../modelopt/torch/quantization/plugins/transformer_engine.py", line 178, in te_grouped_quantized_linear_fn
    num_gemms = len(args[sig_params.index("non_tensor_args") - ctx_offset][0])
TypeError: object of type 'bool' has no len()
```

## Cause

TE 2.17 changed the private `_GroupedLinear.forward` argument layout.

TE 2.16:

```python
forward(ctx, inp, non_tensor_args, *weights_and_biases)
# non_tensor_args[0] is m_splits
```

TE 2.17:

```python
forward(ctx, inp, m_splits, non_tensor_args, *weights_and_biases)
# non_tensor_args[0] is now use_bias
```

ModelOpt detects `non_tensor_args` in the signature and assumes its first element is always `m_splits`. With TE 2.17, it therefore calls `len()` on `use_bias`.

The compatibility logic likely needs to prefer the explicit `m_splits` parameter when present, while retaining the `non_tensor_args[0]` fallback for TE 2.16.

## Reproduction

In a two-GPU Megatron-Bridge environment using the versions above:

```bash
uv run python -m pytest -s -x \
  tests/functional_tests/test_groups/quantization/models/qwen/test_qwen3_moe_quantization_workflow.py::TestQwen3MoeQuantizationWorkflow::test_qwen3_moe_quantization_and_generation_with_expert_parallelism
```

The quantization subprocess fails during the first ModelOpt calibration forward with the traceback above.

AAnoosheh:
### What does this PR do?

Type of change: ? Bug fix

TE 2.16 changed `_GroupedLinear.forward` from
`forward(ctx, inp, non_tensor_args, ...)` (with `m_splits` packed as `non_tensor_args[0]`, TE 2.10-2.15) to
`forward(ctx, inp, m_splits, non_tensor_args, ...)`, splitting `m_splits` back into its own positional argument and shifting `non_tensor_args[0]` to `use_bias`.

`te_grouped_quantized_linear_fn` still assumed the 2.10-2.15 layout, so on TE 2.16 it read `use_bias` (a bool) where it expected `m_splits` (a list/tensor), crashing PTQ calibration for any MoE model using `GroupedLinear` with `TypeError: object of type 'bool' has no len()`.

Rather than adding a third args-introspection branch for the new layout (which would still break the next time TE moves this slot), derive `num_gemms` from `self.num_gemms` instead - set by the public `GroupedLinear.__init__` constructor arg and already relied on elsewhere in this class (`iter_weights_for_calibration`). This removes the args-shape dependency entirely for num_gemms, so future internal reshuffles of `_GroupedLinear.forward` can't break it again. `inp`'s position still requires signature introspection since it's a runtime argument, not a stored attribute - that part is unavoidable.

Adds a `GroupedLinear` case to test_transformer_engine.py, which previously only covered `te.pytorch.Linear`/`LayerNormLinear`.

### Usage

```python
```

- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?:  ❌ 
- Did you get Claude approval on this PR?: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved grouped linear quantization compatibility across Transformer Engine argument layout variations.
  * Now uses the configured grouped GEMM count directly to reduce quantization mismatches.

* **Tests**
  * Expanded quantization test coverage to include grouped Transformer Engine linear layers.
  * Added a new multi-GEMM grouped linear test model and updated parametrization to run it alongside existing coverage.
  * Added conditional skipping for specific calibration configurations where grouped linear is not compatible.

### DIFF
--- a/modelopt/torch/quantization/plugins/transformer_engine.py
+++ b/modelopt/torch/quantization/plugins/transformer_engine.py
@@ -160,10 +160,15 @@ class _QuantTEGroupedLinear(_ParallelLinear):
     @staticmethod
     def te_grouped_quantized_linear_fn(package, func_name, self, *args):
         _assert_te_fp8_enabled()
-        # Locate `inp` and the m_splits-bearing arg by parameter name. The second
-        # slot was renamed from `m_splits` (TE < 2.10) to `non_tensor_args` (TE
-        # 2.10+, where m_splits is now at non_tensor_args[0]). `*weights_and_biases`
-        # is always the trailing variadic — 2 * num_gemms tensors (weights, then biases).
+        # Locate `inp` by parameter name in the un-patched `_GroupedLinear.forward`
+        # signature — robust to TE versions that move the m_splits/non_tensor_args
+        # slot around (e.g. `m_splits` was packed into `non_tensor_args[0]` in TE
+        # 2.10-2.15, then split back out into its own arg in TE 2.16+).
+        # `num_gemms` comes from `self.num_gemms` (set by the public
+        # `GroupedLinear.__init__`) rather than from introspecting `*args`, so it's
+        # unaffected by that churn.
+        # `*weights_and_biases` is always the trailing variadic — 2 * num_gemms tensors
+        # (weights, then biases).
         # See `te_quantized_linear_fn` for why we look up `_forward` here.
         # `_forward` path receives a leading None (placeholder ctx); `_apply` does not.
         orig_forward = getattr(
@@ -174,10 +179,7 @@ class _QuantTEGroupedLinear(_ParallelLinear):
         sig_params = list(inspect.signature(orig_forward).parameters)
         ctx_offset = 0 if func_name == "_forward" else 1
         inp_pos = sig_params.index("inp") - ctx_offset
-        if "non_tensor_args" in sig_params:
-            num_gemms = len(args[sig_params.index("non_tensor_args") - ctx_offset][0])
-        else:
-            num_gemms = len(args[sig_params.index("m_splits") - ctx_offset])
+        num_gemms = self.num_gemms
         weights_start = len(args) - 2 * num_gemms
 
         new_args = list(args)
--- a/tests/gpu_megatron/torch/quantization/plugins/test_transformer_engine.py
+++ b/tests/gpu_megatron/torch/quantization/plugins/test_transformer_engine.py
@@ -43,7 +43,24 @@ class TELinear(nn.Module):
         return torch.randn(2, 16)
 
 
-@pytest.mark.parametrize("model_cls", [TELinear])
+class TEGroupedLinear(nn.Module):
+    """Exercises `_QuantTEGroupedLinear`, used by MoE expert grouped GEMMs."""
+
+    num_gemms = 3
+
+    def __init__(self):
+        super().__init__()
+        self.net = te.pytorch.GroupedLinear(self.num_gemms, 16, 32)
+
+    def forward(self, x):
+        m_splits = [x.shape[0] // self.num_gemms] * self.num_gemms
+        return self.net(x, m_splits)
+
+    def get_input(self):
+        return torch.randn(self.num_gemms * 2, 16)
+
+
+@pytest.mark.parametrize("model_cls", [TELinear, TEGroupedLinear])
 @pytest.mark.parametrize(
     "config",
     [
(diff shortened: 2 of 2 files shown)


kevalmorabia97 (approved):

modelopt/torch/quantization/plugins/transformer_engine.py
realAsma, line 1: RB: Can you check if this works for various TE versions? You can clone each version and check. I suggest checking say  TE 1.5, TE 2.0, TE 2.16+

realAsma: @AAnoosheh  Do you know which versions are covered in this fix?

AAnoosheh: GroupedLinear was added in v1.9 and it always had `self.num_gemms` attribute so this change is compatible with it as far back as it goes.

I don't know why they passed it in as an arg beforehand but now they also decided to just use the attribute too.

realAsma: BB: Could you please reply here?

realAsma (approved):

modelopt/torch/quantization/plugins/transformer_engine.py
realAsma: Sorry Asha for the noisy comments - I was testing my Github bots

realAsma: > _:robot: Bot comment._

🐝 Looking into compatibility across TE 1.5, 2.0, and 2.16+ now.

realAsma: > _:robot: Bot comment._

Checked the exact upstream tag sources (v1.5, v1.9–v1.13, v2.0, v2.10, v2.15, v2.16/v2.16.1, and v2.17):

- **TE 1.5:** N/A — `GroupedLinear` does not exist (it was added in v1.9).
- **TE 1.9:** **not compatible** with this wrapper. It passes three trailing `num_gemms` blocks (`weights`, `weights_fp8`, `biases`), so `len(args) - 2 * self.num_gemms` starts at `weights_fp8`, not the real weights.
- **TE 1.10 through 2.16.1:** source-compatible with this change; the real weights and biases are the final two blocks and `self.num_gemms` is present. TE 2.0 is therefore OK.
- **TE 2.17:** source-compatible and this is where the tagged source actually has the new `m_splits, non_tensor_args` split.

One correction to the PR description: released v2.16 and v2.16.1 still have `forward(ctx, inp, non_tensor_args, *weights_and_biases)`; the split signature first appears in the v2.17 tag. PR CI is green, but it exercised one current container rather than a TE-version runtime matrix, so the above is a source-level compatibility result.

realAsma: > _:robot: Bot comment._

🐝 I’m here — checking this thread now.

realAsma: > _:robot: Bot comment._

Confirmed — BeeBot received this `BB:` review comment and can reply in this thread. Reply delivered successfully.

Issue #1699, opened 2026-06-12
[RFC] NVIDIA Model Optimizer — Product Roadmap


Trenton-Starkey:
# **[RFC] NVIDIA Model Optimizer — Product Roadmap**

[NVIDIA Model Optimizer](https://github.com/NVIDIA/Model-Optimizer) provides a unified library of SOTA model optimization techniques — quantization, sparsity, pruning, distillation, NAS, and speculative decoding — that compress deep learning models for downstream deployment frameworks.

This document outlines our investment areas and upcoming work. Details are subject to change and we'll update this roadmap regularly. We welcome questions and feedback in this thread and feature requests in GitHub Issues.

We organize upcoming work into three horizons: **Now (Q2 – Q3 2026)**, and early plans for **Next (Q4 2026)**. Specific items may shift between horizons as work progresses. 

---

## **1. Model optimization techniques**

Model Optimizer collaborates with internal teams and external research labs to continuously develop and integrate state-of-the-art techniques into the library. Active areas include advanced PTQ, QAT with distillation, attention sparsity, token-efficient pruning, NAS, and distillation, mixed-precision quantization, and emerging speculative-decoding techniques.

### **Now (Q2 – Q3 2026)**

- [ ] **PTQ Recipe Framework** — modular pipeline for post-training quantization across LLM, VLM, and diffusion/video models. YAML recipe interface with mixed-precision support; pluggable calibration; unified Hugging Face export.

  - [ ] **NVFP4 GPTQ** — production-grade GPTQ algorithm tuned for NVFP4 deployment.  
  - [ ] **NVFP4 Autoquant** — production-grade and robust autoquant recipe tuned for NVFP4 deployment  
  - [ ] **NVFP4 PTQ Recipe Book** — partner-facing guidance on recipe selection (which technique for which model class).  
  - [ ] **NVFP4 Scale for Accuracy Recovery -** umbrella of production-grade weight-scale-setting methods (Per-block MSE, 4/6, Per-Block Output Error). Closes the accuracy gap that NVFP4 PTQ alone can't recover, without requiring a training pass.  
  - [ ] **TensorRT compatibility** — ModelOpt quantization with the latest TensorRT for edge, automotive, and embedded targets. NVFP4 recipes for sub-130B and sub-10B model classes.

- [ ] **PTQ Developer Experience**

[...]

PR #1938, opened 2026-07-07
MBridge VLM distillation / QAD support

kevalmorabia97:
## What

Adds VLM (e.g. Qwen3.5-VL, Gemma3-VL) knowledge-distillation / QAD support to the Megatron-Bridge examples:

- `distill.py` distills only the **language model** submodule (vision tower + projector untouched), reusing the LLM training path.
- New `export_distilled_megatron_to_hf.py` converts a distilled Megatron checkpoint (**any** iteration) to HF. Required especially for VLM distilled ckpt as it only has LM weights so we need to initialize full VLM, swap LLM weights then save to HF
- Renames `export.py` → `export_quantized_megatron_to_hf.py`.

## Related upstream Megatron-Bridge PRs to be available in nemo:26.08 container:

- NVIDIA-NeMo/Megatron-Bridge#4707 — `DistillationProvider` submodule distillation (non-blocking; added temporary WAR)
- NVIDIA-NeMo/Megatron-Bridge#4706 — MoE expert weight-mapping fix (Qwen3.5-VL-MoE with moe_grouped_gemm=False). Also removed ModelOpt side WAR previously added as it was not accurate; better to wait till next container release or mount latest MBridge into the 26.06 container.

## Testing

- Qwen3.6-35B-A3B Pruning + Distillation with MMLU evaluation sanity check (results below in comments)
- Cosmos 2 Reason 2B valiadted by SAs (results below in comments)
- Validated end-to-end on `nemo:26.06` (distill → separate HF export; LLM + VLM, incl. TP→TP/PP reshard). `test_distill_vlm` runs the export script as a CI e2e step.
- Many CICD tests for wide coverage of all mbridge scripts

## Summary

* **New Features**
  * Added a dedicated HuggingFace exporter for distilled Megatron checkpoints, with distinct LLM vs VLM conversion flows.

* **Bug Fixes**
  * Improved Megatron-Bridge distillation/export consistency, including safer handling of VLMs and targeted submodule distillation.

* **Documentation**
  * Updated Megatron-Bridge READMEs and tutorials to reference the new quantized and distilled export scripts and revised CLI guidance.

* **Tests**
  * Expanded distillation, QAD, and quantization/export tests to cover LLM/VLM variants, with conditional skipping for unsupported MoE setups.


examples/megatron_bridge/_distillation_provider.py
AAnoosheh, line 1: Can we create subclasses of the provider inside MBridge itself?

kevalmorabia97: This is temporary patching. Same as what I do in MBridge PR. Once 26.08 container is out, this file will be removed

kevalmorabia97: Yes — the real home is Megatron-Bridge itself: NVIDIA-NeMo/Megatron-Bridge#4707 adds this `DistillationProvider` (deferred KD conversion + `distill_submodule`) upstream. This `_distillation_provider.py` is a temporary vendored copy so the examples work on the nemo:26.06 container, which predates that change; it'll be removed once #4707 ships in nemo:26.08.

examples/megatron_bridge/distill.py
```diff
@@ -450,7 +374,22 @@ def _build_model_provider(hf_path, load_weights=True):
...
+        full_student = run_provider.full_model
+        # Strip the distillation wrapper -> plain trained language model (in place; reassign to be safe).
+        full_student.language_model = mtd.export(full_student.language_model)
+        save_vlm_to_hf(
+            full_student,
+            args.hf_export_path,
+            args.student_hf_model,
+            trust_remote_code=args.trust_remote_code,
+        )
+        print_rank_0(f"Saved distilled VLM to {args.hf_export_path} in HF format")
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 623b6c9 — the VLM export now passes `args.student_hf_path`.

examples/megatron_bridge/quantize.py
```diff
@@ -414,8 +415,8 @@ def forward_loop(_model=None):
         hf_tokenizer_kwargs={"trust_remote_code": args.trust_remote_code},
     )
     print_rank_0(
-        f"\nSaved quantized model to {args.export_megatron_path} in Megatron format. "
-        "To deploy this model (TensorRT-LLM / vLLM / SGLang), convert it to a Unified HF ckpt with export.py"
+        f"\nSaved quantized model to {args.export_megatron_path} in Megatron format. To deploy this model "
+        "(TensorRT-LLM / vLLM / SGLang), convert it to a Unified HF ckpt with export_quantized_megatron_to_hf.py"
     )
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 623b6c9 — the hint is now branched on `is_vlm`; VLMs get a message noting quantized-VLM HF export is unsupported.

kevalmorabia97:
Addressed review comments in 623b6c9:

- **@coderabbitai** VLM export template arg (distill.py) — fixed: passes `--student_hf_path`.
- **@coderabbitai** quantize.py post-save hint — fixed: scoped on `is_vlm` (VLMs note that quantized-HF export is unsupported).
- **@coderabbitai** duplicated `_qwen35_moe_bridge_supported()` — moved to `tests/_test_utils/examples/megatron_bridge.py` and imported in both `test_qad.py` and `test_prune_minitron.py`.
- **@coderabbitai** "keep VLM export coverage" (test_quantize_export.py) — **skipping**: there's no quantized-VLM *export* path to cover (it's unsupported, and `export_quantized_megatron_to_hf.py` is LLM-only). The meaningful part of the removed test — PTQ of a VLM producing a Megatron checkpoint — is still covered by `test_qad.py`'s `gemma3vl` case (Step 1 runs `quantize.py` on the VLM and asserts `modelopt_state`).
- **@AAnoosheh / @kevalmorabia97**: confirmed — the provider lives upstream in NVIDIA-NeMo/Megatron-Bridge#4707; `_distillation_provider.py` is a temporary 26.06 vendored copy, removed once nemo:26.08 ships it.

All example-test changes collect cleanly; the code fixes are message/arg-only (no behavior change for the default `--student_hf_model == --student_hf_path` case).

examples/megatron_bridge/distill.py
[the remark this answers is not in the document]
kevalmorabia97: Thanks. The `student_hf_model` -> `student_hf_path` fix landed in 623b6c9.

On coverage: the export logic itself (`save_vlm_to_hf`) is shared with the standalone `export_distilled_megatron_to_hf.py`, which `test_distill_vlm` runs end-to-end (build VLM -> load LM checkpoint -> `save_hf_weights`), so the reassembly + cross-rank gather are covered in CI. The only inline-specific bit not in CI is the `initialize_model_parallel`-after-`distill()` glue; I validated that manually on nemo:26.06 (distill -> inline `--hf_export_path` for gemma3vl: vision/projector byte-identical, LM loads clean). Leaving the inline path as-is to avoid a second ~90s distill run in CI.

nfasfous:
Tested on Cosmos-Reason2-2B:
- Pruned Model from 28 to 24 dec blocks -> baseline to start KD experiments
- KD unpruned VLM teacher to pruned VLM student with mbridge then applied PTQ to student down to (a) FP8 and (b) NVFP4
- Same setup for QAD with (a) FP8 and (b) NVFP4
- Eval on MMStar dataset and compare to baseline BF16 KD'd student model

Note: KD/QAD only applied only for short run of 50 iters

Results:
image
image

Key takeaways: particularly for NVFP4, QAD recovers drop in performance. FP8 less impacted overall, difference to BF16 is noise.

examples/megatron_bridge/README.md
```diff
@@ -47,6 +47,13 @@ docker run \
 > [!WARNING]
 > Use `python -m pip` instead of `pip` to avoid conflicts with the system-wide installed packages in the NeMo containers. You may also refer to this [doc](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/docker/common/README.md#installing-packages-inside-the-container) on how to correctly install packages in the NeMo containers without breaking existing torch installation.
 
+> [!NOTE]
+> **Working with MoE Vision-Language Models (e.g. Qwen3.5-VL-MoE)?** The `nemo:26.06` container's Megatron-Bridge lacks the MoE expert weight mappings these models need (dense VLMs such as Gemma3-VL and Qwen3-VL work as-is). Until the `nemo:26.08` container is released, mount the latest [Megatron-Bridge `main`](https://github.com/NVIDIA-NeMo/Megatron-Bridge) source over the pre-installed copy by adding this to the `docker run` command above:
+>
+> ```bash
+> -v ${MEGATRON_BRIDGE_DIR}:/opt/Megatron-Bridge
```
danielkorzekwa: Rename `MEGATRON_BRIDGE_DIR` to `MEGATRON_BRIDGE_SRC_DIR` to  make it clear what it is.

examples/megatron_bridge/README.md
```diff
@@ -81,20 +88,20 @@ torchrun --nproc_per_node 2 quantize.py \
...
+torchrun --nproc_per_node 2 export_quantized_megatron_to_hf.py \
     --hf_model_name_or_path Qwen/Qwen3-8B \
     --megatron_path /tmp/Qwen3-8B-NVFP4-megatron \
     --pp_size 2 \
     --export_unified_hf_path /tmp/Qwen3-8B-NVFP4-hf
 ```
 
 > [!NOTE]
-> The HuggingFace unified exporter does not gather tensor-parallel-sharded weights. Use `--pp_size` on `export.py` to shard a large model with pipeline parallelism across GPUs for export.
+> The HuggingFace unified exporter does not gather tensor-parallel-sharded weights. Use `--pp_size` on `export_quantized_megatron_to_hf.py` to shard a large model with pipeline parallelism across GPUs for export.
```
danielkorzekwa: it sounds too technical, is it what we mean? `	The HuggingFace unified exporter uses a single pipeline stage by default. Use --pp_size on export_quantized_megatron_to_hf.py to export large models faster and more memory-efficiently across multiple GPUs.`

examples/megatron_bridge/README.md
```diff
@@ -104,14 +111,18 @@ For a vision-language model (e.g. Qwen3.5-VL, Gemma3-VL), `quantize.py` automati
 - A **text** dataset runs text-only calibration of the language model (vision tower idle).
 
 > [!NOTE]
-> HuggingFace unified export (`export.py`) of a quantized VLM is not yet supported; the quantized VLM is saved in Megatron checkpoint format only.
+> HuggingFace unified export (`export_quantized_megatron_to_hf.py`) of a quantized VLM is not yet supported; the quantized VLM is saved in Megatron checkpoint format only.
 
 ## Sanity-Check Generation
```
danielkorzekwa: The top table in this wiki, change the order to: Pre-Requisites, Post-Training Quantization, Pruning, Distillation, Sanity-Check Generation (or maybe even consider to put pruning before post-training quantization as it is more natural flow)

Sanity-Check Generation at the and as it can be run after each stage including distillation

On the top of the doc add:

This page covers multiple Model Optimizer workflows with Megatron-Bridge, including quantization, pruning, and distillation. These workflows can be used on their own or combined, depending on your model optimization goals.

Sanity-check generation can be used to quickly verify the output quality of quantized, pruned, or distilled models.

examples/megatron_bridge/README.md
```diff
@@ -104,14 +111,18 @@ For a vision-language model (e.g. Qwen3.5-VL, Gemma3-VL), `quantize.py` automati
...
 
-[generate_vllm.py](generate_vllm.py) runs a quick generation check on a unified HuggingFace checkpoint using vLLM. vLLM auto-detects the ModelOpt quantization from the exported `hf_quant_config.json`, so no extra quant flags are needed:
+[generate_vllm.py](generate_vllm.py) runs a quick generation check on an exported HuggingFace checkpoint using vLLM — a useful smoke test for both a **quantized** and a **pruned** model to confirm it still produces coherent text before investing in [distillation](#distillation) to recover accuracy. For quantized checkpoints, vLLM auto-detects the ModelOpt quantization from the exported `hf_quant_config.json`, so no extra flags are needed:
```
danielkorzekwa: also distilled right?

examples/megatron_bridge/README.md
```diff
@@ -104,14 +111,18 @@ For a vision-language model (e.g. Qwen3.5-VL, Gemma3-VL), `quantize.py` automati
...
 
 ```bash
+# Quantized model
 python generate_vllm.py --model nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4 --trust_remote_code
+
+# Pruned model
+python generate_vllm.py --model /tmp/Qwen3-8B-Pruned-6B
```
danielkorzekwa: why there is no --trust_remote_code here? do we need to have two commands here,? can be just one? with --model /path/to/model_to_validate?

examples/megatron_bridge/README.md
```diff
@@ -204,7 +215,7 @@ torchrun --nproc_per_node 8 distill.py \
     --output_dir /output/qwen3_8b_nvfp4_qad
 ```
 
-The distilled checkpoint retains the ModelOpt quantization state, so it can be converted to a deployable HuggingFace checkpoint with [export.py](export.py) (point `--megatron_path` at `<output_dir>/checkpoints`), exactly like the PTQ checkpoint in [step 2 above](#post-training-quantization).
+The distilled checkpoint retains the ModelOpt quantization state, so it can be converted to a deployable HuggingFace checkpoint with [export_quantized_megatron_to_hf.py](export_quantized_megatron_to_hf.py) (point `--megatron_path` at `<output_dir>/checkpoints`), exactly like the PTQ checkpoint in [step 2 above](#post-training-quantization).
```
danielkorzekwa: change at `<output_dir>/checkpoints>` to `/output/qwen3_8b_nvfp4_qad`

examples/megatron_bridge/README.md
```diff
@@ -214,27 +225,36 @@ To run the distillation script on a Slurm cluster for multi-node training, you j
 
 The distilled checkpoint is saved in Megatron distributed format. If you need a HuggingFace checkpoint, there are two ways to convert it:
 
-**Inline** -- add `--hf_export_path` and `--student_hf_model` to the `distill.py` command to automatically convert the final checkpoint after distillation:
+**Inline** -- add `--hf_export_path` to the `distill.py` command to automatically convert the **final** checkpoint after distillation:
```
danielkorzekwa, line 202: --student_hf_model is removed from the command, do we still need this command  above in:

```
torchrun --nproc_per_node 8 distill.py \
    --tp_size 8 \
    --teacher_hf_path Qwen/Qwen3-8B \
    --student_hf_path Qwen/Qwen3-8B \
```

examples/megatron_bridge/README.md
```diff
@@ -214,27 +225,36 @@ To run the distillation script on a Slurm cluster for multi-node training, you j
 
 The distilled checkpoint is saved in Megatron distributed format. If you need a HuggingFace checkpoint, there are two ways to convert it:
```
danielkorzekwa: Currently `Converting to Hugging Face format (optional)` section is below `Quantization Aware Distillation (QAD)` section that has its own checkpoint sub section, it is confusing.

Two options:
1) Move checkpoint export paragraph from `Quantization Aware Distillation (QAD)` to `Converting to Hugging Face format (optional)` 

2) Make the section structure:
- Distillation
   - Data preparation
   - Distillation with Real Data
   - Quick Test with Mock Data
   - Converting to Hugging Face format (optional)
   - Slurm Usage
   - Other examples
      - Quantization Aware Distillation (QAD)
      - Vision-Language Models (VLMs)

examples/megatron_bridge/README.md
```diff
@@ -214,27 +225,36 @@ To run the distillation script on a Slurm cluster for multi-node training, you j
...
 ```bash
 torchrun --nnodes 1 --nproc_per_node 8 distill.py \
     ... \
-    --hf_export_path /path/to/save/distilled_hf_ckpt \
-    --student_hf_model Qwen/Qwen3-4B
+    --hf_export_path /path/to/save/distilled_hf_ckpt
 ```
 
-`--student_hf_model` should match the base architecture of the student (used as a template for export). For non-Puzzletron (i.e. standard) models, it should be same as `--student_hf_path`.
+`--student_hf_model` is also needed for heterogeneous (Puzzletron/NAS) students whose architecture differs from `--student_hf_path`; for standard (homogeneous) models it defaults to `--student_hf_path`, and it is unused for VLMs.
```
danielkorzekwa: unclear paragraph
- "...and it is unused for VLMs...." - unclear and VLMs here are out of context
- ".--student_hf_model` is also needed for heterogeneous (Puzzletron/NAS) students whose architecture differs from `--student_hf_path`;"  - what it means, .e.g, if I compress distill a puzzletron-compressed qwen 3 8B then the value for both args would be the same right? make it more clear.

examples/megatron_bridge/README.md
```diff
@@ -214,27 +225,36 @@ To run the distillation script on a Slurm cluster for multi-node training, you j
...
 
-**Separate conversion** -- convert any saved iteration using the Megatron-Bridge conversion script:
+**Separate conversion** -- convert **any** saved iteration (intermediate or final) with [export_distilled_megatron_to_hf.py](export_distilled_megatron_to_hf.py):
```
danielkorzekwa, line 212: why to emphasize it?

examples/megatron_bridge/README.md
```diff
@@ -214,27 +225,36 @@ To run the distillation script on a Slurm cluster for multi-node training, you j
...
+torchrun --nproc_per_node 1 export_distilled_megatron_to_hf.py \
+    --student_hf_path <student_hf_model_or_path> \
+    --megatron_path <distill_output_dir>/checkpoints/iter_<iter_number> \
+    --hf_export_path /path/to/save/distilled_hf_ckpt
 ```
 
-For more details, see the [Megatron-Bridge conversion README](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/conversion).
+### Vision-Language Models (VLMs)
+
+For a vision-language model (e.g. Qwen3.5-VL, Gemma3-VL), `distill.py` automatically distills only the **language model** (on text data) and leaves the vision tower and vision-language projector untouched, then saves the full VLM back. This also aligns with pruning and quantization behavior where we leave the vision tower and projector untouched. This composes with pruning and QAD (`--student_megatron_path`) exactly as for LLMs, and the HF export reuses `--student_hf_path` (no `--student_hf_model` needed).
```
danielkorzekwa: unclear what it means

examples/megatron_bridge/README.md
danielkorzekwa, line 364: why to have roadmap here under resources, we do not have such resources section at other example readme files? we point road to a specific issue/1699 - will the roadmap always stay under this issue?

danielkorzekwa (commented):
Good work. I reviewed the READE first, let's resolves comments first, then I can review the code as well

examples/megatron_bridge/README.md
kevalmorabia97: Done — renamed to `MEGATRON_BRIDGE_SRC_DIR`.

examples/megatron_bridge/README.md
kevalmorabia97: Reworded to be less jargony while keeping it accurate: the unified exporter can't split weights across GPUs with tensor parallelism, so use `--pp_size` for a pipeline-parallel export.

examples/megatron_bridge/README.md
kevalmorabia97: Added the intro paragraph and moved **Sanity-Check Generation** to the end of both the table and the doc (it applies after any stage).

examples/megatron_bridge/README.md
kevalmorabia97: Yes — updated to "quantized, pruned, or **distilled**".

examples/megatron_bridge/README.md
kevalmorabia97: The flag is model-dependent, not stage-dependent: Nemotron ships custom modeling code so it needs `--trust_remote_code`; Qwen doesn't. Added a note clarifying that and kept the two examples to show the quantized vs pruned invocations.

examples/megatron_bridge/README.md
kevalmorabia97: Done — now uses `/output/qwen3_8b_nvfp4_qad/checkpoints`.

examples/megatron_bridge/README.md
kevalmorabia97: Clarified in the rewritten `--student_hf_model` paragraph — it's optional and defaults to `--student_hf_path`, so it's omitted from the standard command.

examples/megatron_bridge/README.md
kevalmorabia97: Restructured: the non-quantized methods (real-data / VLM) are now followed by their **Converting to HF format** step, then **QAD** (quantized, with its own `export_quantized_megatron_to_hf.py` note) and Slurm. Converting is no longer stranded below QAD.

examples/megatron_bridge/README.md
kevalmorabia97: Rewrote it: `--student_hf_path` builds the student and provides the exported config/tokenizer; `--student_hf_model` is only needed for **heterogeneous** (Puzzletron/NAS) students whose architecture differs from `--student_hf_path`, and defaults to it otherwise. Note: for a heterogeneous Puzzletron student the two args differ (that's why the flag exists), so they wouldn't be the same value.

examples/megatron_bridge/README.md
kevalmorabia97: The bold **any** is intentional — it contrasts with the **Inline** path just above, which only exports the **final** checkpoint. It flags that the separate script can export any intermediate iteration too. Happy to drop the emphasis if you'd prefer.

examples/megatron_bridge/README.md
kevalmorabia97: Simplified the paragraph.

examples/megatron_bridge/README.md
kevalmorabia97: The `## Resources` + 📅 Roadmap footer is a shared convention across all example READMEs (hf_ptq, pruning, llm_distill, llm_qat, diffusers, onnx_ptq, torch_onnx, torch_trt, speculative_decoding), and #1699 is the canonical Model-Optimizer roadmap issue — so I'd keep it here for consistency with the other examples rather than diverge.

kevalmorabia97:
Thanks for the review! Addressed all README comments in d5e97d8446 (replied inline; resolved the addressed threads).

Two left open for your call:
- **`any` emphasis** — kept deliberately to contrast with the Inline path (which only exports the *final* checkpoint); happy to drop it.
- **Roadmap under Resources** — it's a shared footer across all example READMEs (hf_ptq, pruning, llm_distill, …), so I'd keep it for consistency.

Ready for the code review whenever you are.

modelopt/torch/prune/plugins/mcore_minitron.py
```diff
@@ -385,50 +414,65 @@ def run_search(self) -> None:
...
+        finally:
+            # reset to max subnet, revert dropped layers, and restore in-place-sliced patterns
+            sample(model, sample_func=max)
+            for offset, layer in enumerate(all_layers):
+                layer.layer_number = start_layer_number + offset
+            model.decoder.layers = all_layers
+            for attr, pattern in saved_patterns.items():
+                setattr(model.config, attr, pattern)
+            if hybrid_key:
+                setattr(model, hybrid_key, saved_hybrid)
```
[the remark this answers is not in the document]
kevalmorabia97: Good catch — confirmed. `_prune` sets the width hparams on all layers *before* dropping some (so dropped layers hold pruned per-layer widths), and `drop_mcore_language_model_layers` only detaches them. Fixed in f0787db16e by reattaching the full layer list before `sample(max)` so every layer — including the dropped ones — is reset to the max subnet.

kevalmorabia97:
# Sanity check on Qwen3.6-35B-A3B with MMLU 5-shot (LM Eval Harness)

## Original Qwen3.6-35B-A3B (34.7B, A3.5)

|      Groups      |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|------------------|------:|------|------|------|---|-----:|---|-----:|
|mmlu              |      2|none  |      |acc   |↑  |0.8222|±  |0.0031|
| - humanities     |      2|none  |      |acc   |↑  |0.7834|±  |0.0059|
| - other          |      2|none  |      |acc   |↑  |0.8478|±  |0.0061|
| - social sciences|      2|none  |      |acc   |↑  |0.8866|±  |0.0055|
| - stem           |      2|none  |      |acc   |↑  |0.7923|±  |0.0070|


## Pruned (22B, A3.0): Num experts 256->232, MoE FFN 512->384

Pruned 37% total and 15% active params

|      Groups      |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|------------------|------:|------|------|------|---|-----:|---|-----:|
|mmlu              |      2|none  |      |acc   |↑  |0.7621|±  |0.0034|
| - humanities     |      2|none  |      |acc   |↑  |0.7041|±  |0.0064|
| - other          |      2|none  |      |acc   |↑  |0.7895|±  |0.0070|
| - social sciences|      2|none  |      |acc   |↑  |0.8479|±  |0.0064|
| - stem           |      2|none  |      |acc   |↑  |0.7377|±  |0.0076|


## Distilled (100 iters @ 2.5B tokens using Nemotron-Pretraining-SFT-v1)

|      Groups      |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|------------------|------:|------|------|------|---|-----:|---|-----:|
|mmlu              |      2|none  |      |acc   |↑  |0.7770|±  |0.0033|
| - humanities     |      2|none  |      |acc   |↑  |0.7044|±  |0.0062|
| - other          |      2|none  |      |acc   |↑  |0.8153|±  |0.0067|
| - social sciences|      2|none  |      |acc   |↑  |0.8690|±  |0.0060|
| - stem           |      2|none  |      |acc   |↑  |0.7577|±  |0.0073|

AAnoosheh (approved):

examples/megatron_bridge/tutorials/README.md
```diff
@@ -1,7 +1,7 @@
 # Megatron-Bridge Tutorials
 
 End-to-end tutorials that combine ModelOpt optimization techniques on [NVIDIA Megatron-Bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge) models.
-Each one walks through a complete workflow using the scripts in [examples/megatron_bridge](../README.md) (`prune_minitron.py`, `distill.py`, `quantize.py`, `export.py`).
+Each one walks through a complete workflow using the scripts in [examples/megatron_bridge](../README.md) (`prune_minitron.py`, `distill.py`, `quantize.py`, `export_quantized_megatron_to_hf.py`, `export_distilled_megatron_to_hf.py`).
```
danielkorzekwa, line 4: How about having one export script for any kind of megatron checkpoint, distilled or quantized or just pruned? - better UX

examples/megatron_bridge/distill.py
```diff
@@ -347,23 +247,46 @@ def _build_model_provider(hf_path, load_weights=True):
...
+    if is_vlm:
+        warn_rank_0(
+            "VLM detected: distilling model.language_model only (vision tower / projector untouched). "
+            "To export megatron non-quantized checkpoint, use export_distilled_megatron_to_hf.py"
+        )
+    run_provider = convert_to_distillation_provider(
+        student_provider,
+        teacher_provider,
+        kd_config,
+        distill_submodule="language_model" if is_vlm else None,
```
danielkorzekwa, line 272: not all vision models will have language models of this name right? also detecting vlm by having "vision_config" is hacky, what if the name will be difference?

examples/megatron_bridge/distill.py
```diff
@@ -347,23 +247,46 @@ def _build_model_provider(hf_path, load_weights=True):
...
+        AutoConfig.from_pretrained(args.student_hf_path, trust_remote_code=args.trust_remote_code),
+        "vision_config",
+    )
+
+    if is_vlm:
+        warn_rank_0(
+            "VLM detected: distilling model.language_model only (vision tower / projector untouched). "
+            "To export megatron non-quantized checkpoint, use export_distilled_megatron_to_hf.py"
+        )
+    run_provider = convert_to_distillation_provider(
```
danielkorzekwa: why calling run_provider and not distill_provider as before, personally I would call it model_provider or distill_model_provider

examples/megatron_bridge/_distillation_provider.py
```diff
@@ -0,0 +1,108 @@
...
+        (type(provider),),
+        {
+            "provide": _provide,
+            "_convert_hook": _convert_hook,
+            "distill_submodule": distill_submodule,
+        },
+    )
+    # Use object.__setattr__ to bypass DistillationProvider.__setattr__, which mirrors every attribute
+    # set onto the teacher -- assigning ``__class__`` normally would also switch the teacher's class.
+    object.__setattr__(provider, "__class__", submodule_cls)
```
danielkorzekwa, line 91: it seems hacky, likely this code should be in mbridge (modifying base distillation provider there), and schedule the work, at least put TODO here to refactor it.

examples/megatron_bridge/export_distilled_megatron_to_hf.py
```diff
@@ -0,0 +1,222 @@
...
+import modelopt.torch.utils.distributed as dist
+from modelopt.torch.export import copy_hf_ckpt_remote_code
+from modelopt.torch.utils import print_args, print_rank_0
+from modelopt.torch.utils.plugins.mbridge import (
+    load_mbridge_model_from_hf,
+    load_modelopt_megatron_checkpoint,
+)
+
+
+def export_llm_to_hf(
```
danielkorzekwa, line 61: can we use in-memory model as well as in save_hf_to_vlm below?

examples/megatron_bridge/export_distilled_megatron_to_hf.py
```diff
@@ -0,0 +1,222 @@
...
+    parser.add_argument("--cp_size", type=int, default=1, help="Context parallel size")
+
+    args = parser.parse_args()
+    print_args(args)
+
+    return args
+
+
+def main(args: argparse.Namespace):
+    is_vlm = hasattr(
```
danielkorzekwa, line 164: this is duplicate in distill.py, idealy use a better design to not have if-else logic in multiple places depending on is_vlm

examples/megatron_bridge/prune_minitron.py
```diff
@@ -72,6 +72,11 @@
 from modelopt.torch.utils.plugins.megatron_mmlu import megatron_mmlu
```
danielkorzekwa, line 72: is changing this file needed for QAD support for VLMs?

examples/megatron_bridge/quantize.py
```diff
@@ -413,10 +414,16 @@ def forward_loop(_model=None):
         hf_tokenizer_path=args.hf_model_name_or_path,
         hf_tokenizer_kwargs={"trust_remote_code": args.trust_remote_code},
     )
-    print_rank_0(
-        f"\nSaved quantized model to {args.export_megatron_path} in Megatron format. "
-        "To deploy this model (TensorRT-LLM / vLLM / SGLang), convert it to a Unified HF ckpt with export.py"
-    )
+    if is_vlm:
```
danielkorzekwa, line 417: similar comment as in other places to avoid if else paths for is_vlm

modelopt/torch/nas/plugins/megatron_model_stats.py
```diff
@@ -394,14 +394,23 @@ def _get(attr: str, default: Any = None) -> Any:
     # GatedDeltaNet (linear-attention) hybrid, e.g. Qwen3-Next / Qwen3.5: every ``linear_attention_freq``-th
     # layer keeps full attention, the rest use GatedDeltaNet linear attention.
     experimental_attention_variant: str | None = _get("experimental_attention_variant", None)
-    linear_attention_freq: int | None = _get("linear_attention_freq", None)
+    # ``linear_attention_freq`` is either an int interval (every N-th layer is full attention) or an
+    # explicit per-layer pattern (list; 1 = linear_attention / GatedDeltaNet, 0 = full_attention).
+    linear_attention_freq: int | list[int] | None = _get("linear_attention_freq", None)
     is_gdn = experimental_attention_variant == "gated_delta_net" and bool(linear_attention_freq)
+
+    def _is_linear_attention_layer(i: int) -> bool:
```
danielkorzekwa, line 402: accessing linear_attention_freq from outside of this function - error prone and confusing a reader

examples/megatron_bridge/tutorials/README.md
kevalmorabia97: Quantization export will be dropped soon. MBridge has some initial support to export quantized megatron to HF. I plan to extend it and drop custom quantized megatron import/export logic in modelopt. Leaving as it is for now

examples/megatron_bridge/distill.py
kevalmorabia97: For Megatron Bridge, `language_model` is the naming convention used. For HF models as well, vision_config is the convention. If things change in future, this will error out loudly so we can fix it; but no risk of silent behavior

examples/megatron_bridge/_distillation_provider.py
kevalmorabia97: Everything in this file is already in MBridge but would be available in 26.08 container. Meanwhile this file adds temporary patching to unblock VLM models. Will be removed in next mbridge release

examples/megatron_bridge/prune_minitron.py
kevalmorabia97: Some bug fixes I discovered when testing Qwen3.6 pruning that was previously not tested properly. Needed to pruned + distill

examples/megatron_bridge/distill.py
kevalmorabia97: Renamed to `distill_provider` in c3f770e9c8.

examples/megatron_bridge/export_distilled_megatron_to_hf.py
kevalmorabia97: The LLM path uses `AutoBridge.export_ckpt`, which reads the checkpoint's actual per-layer shapes from disk -- that's what lets it handle heterogeneous (Puzzletron/NAS) students without rebuilding them. The VLM path reassembles in memory because only the language_model is distilled and the vision tower/projector come straight from the HF model. Unifying both onto an in-memory path is reasonable as a follow-up.

examples/megatron_bridge/export_distilled_megatron_to_hf.py
kevalmorabia97: Both are separate entry-point scripts that check the HF config before the model is built, so they use the same 3-line `vision_config` check (the convention noted above). I kept a small local check in each rather than adding a shared util just for example scripts; fully removing the `is_vlm` branching would be a larger redesign, out of scope here.

examples/megatron_bridge/quantize.py
kevalmorabia97: The branch selects VLM-specific calibration (image-text dataset, calibrating only the language model) vs. text calibration -- inherent to the two modalities. Collapsing it would require unifying the LLM/VLM calibration paths, a larger refactor beyond this PR.

modelopt/torch/nas/plugins/megatron_model_stats.py
kevalmorabia97: It's a small nested helper consistent with the other closures in this function (`_get`, `_attn_params` also close over `hidden_size`/`is_gdn`). Happy to pass it explicitly if you prefer, though that'd diverge from the surrounding style.

examples/megatron_bridge/distill.py
danielkorzekwa: lets put a comment in the code about mbridge naming convention

examples/megatron_bridge/export_distilled_megatron_to_hf.py
danielkorzekwa: lets put TODO in the code

examples/megatron_bridge/export_distilled_megatron_to_hf.py
danielkorzekwa: ok

modelopt/torch/nas/plugins/megatron_model_stats.py
danielkorzekwa: let keep it, I see it is inside the function

danielkorzekwa (approved):

examples/megatron_bridge/distill.py
kevalmorabia97: Added a comment documenting the `vision_config` / `language_model` convention in 5f65a35215.

examples/megatron_bridge/export_distilled_megatron_to_hf.py
kevalmorabia97: Added a TODO in `export_llm_to_hf` (5f65a35215) to unify onto the in-memory export path.

jvel07:
@danielkorzekwa @kevalmorabia97, this seems promising, many thanks for the PR.
On a first glance at the issue title, I saw VLM Distillation and thought that it was actual multimodal KD. Then I immediately read it is pure LLM KD. 
I was wondering: 
What would be the difference with Megatron Bridge's [KD](https://docs.nvidia.com/nemo/megatron-bridge/0.2.0/training/distillation.html)?

On the other hand, the results seem promising, are there any VL benchmarks that you guys evaluated besides text benchmarks? To see how the model performs in VL tasks (video and image) after compression. For example VideoMME or MMMU.

kevalmorabia97:
@jvel07 Megatron-Bridge's KD is essentially using Model-Optimizer under the hood and only a LLM distillation, no VLM support which this script adds (still only distills LM but save/load is for full VLM).

Since we prune/quantize only the LM part of VLM and the vision encoder is left frozen, we dont distill it for simplicity (also get rid of having Image based distillation data)

We plan to do more extensive pruning and distillation study for VLMs in near future when we will look at LLM and VLM benchmarks both. Meanwhile, if you are interested, feel free to test it out and share results in this PR

jvel07:
@kevalmorabia97 thanks for the reply!
Sure, could you share the weights so I can give it a try on MM benchmarks?

kevalmorabia97:
@jvel07 sorry unfortunately I cannot share the weights. Would you able to re-do pruning and a short distillation (100 iters)?

jvel07:
@kevalmorabia97, sure, I can try. Could you please point me to the exact recipe you guys used to be able to reproduce?

PR #1939, opened 2026-07-07
Registry-based module dispatch for unified_export_hf.py

sychen52:
### What does this PR do?

Type of change: ? refactor

Use registry-based module dispatch for unified_export_hf.py
export handlers do not replace classes, need structural matching, and preparation/export require independent ordering
### Usage

same as before

### Testing
old and new unittest

- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`:N/A 
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)? N/A 

## Summary

* **New Features**
  * Added registry-driven export dispatch for Hugging Face quantized models, improving coverage across standard layers and multiple MoE/expert variants.
  * Introduced a shared export context to manage per-export state such as tied-weight handling.
* **Bug Fixes**
  * Improved quantized/tied-weight export behavior to prevent duplicated or incorrectly packed weights.
  * Enhanced MoE expert “amax” preparation to support more expert structures.
* **Tests**
  * Added unit tests for registry matching, precedence/replacement behavior, and quantized export paths.


cjluo-nv (commented):

Refactor of `unified_export_hf.py`'s per-module export dispatch: replaces the if/elif chains in `_process_quantized_modules` and the MoE input-amax prepass in `_export_transformers_checkpoint` with a new ordered, first-match-wins `ExportModuleRegistry` (new `registry.py`: `ExportContext`, `ModuleExporter`, `ExportModuleRegistry`). +595/-158, 4 files, with a solid CPU unit-test suite (`test_export_registry.py`) covering MRO/name/predicate matching, precedence, prepend, re-registration idempotency, and a small FP8 ToyModel end-to-end export.

Design review (new subsystem/registry): the codebase already has a class-keyed registration-and-dispatch idiom in `_DMRegistryCls`/`QuantModuleRegistry` (`modelopt/torch/opt/dynamic.py`). The new module's docstring explicitly acknowledges this and justifies a *different* mechanism — quantization converts module classes in place, whereas export needs to emit weights without changing class, plus it needs predicate/structural matching (`_has_fused_experts_quantizers`, iterable catch-all) that the class-dict-based `_DMRegistryCls` doesn't offer. That rationale is reasonable, but it lives only in the source docstring — the PR body just says "refactor" and doesn't compare alternatives. Worth a maintainer confirming the second registry is warranted rather than extending the existing one.

Behavioral parity looks preserved on inspection: first-match ordering (`_FusedExpertsExporter` before `_BmmExpertsExporter`, `_MoELinearExporter` before `_QuantLinearExporter` — `"QuantLinear"` is not a substring of `"QuantMoELinear"` so no double-export), the format-NONE guards and NotImplementedError-for-unknown-experts are retained, and the broad `_IterableExpertsExporter` catch-all has a no-op `export()` so matching `nn.ModuleList`/`nn.Sequential` during the walk is harmless. No CHANGELOG (internal refactor), though note the three new symbols do become public via the `export` package star-import.

No prompt-injection attempts in the untrusted PR fields.

Why nudge (not approve): (1) this rewrites a critical, model-specific export dispatch and the real DBRX/Llama4/GptOss/fused-experts/QuantMoELinear paths are GPU/large-model only — CPU tests validate wiring + a tiny FP8 export, not the full-size expert exports (the codebase's known testing gap); (2) the "why a second registry vs. extending QuantModuleRegistry/_DMRegistryCls" decision is only in the code docstring, not the PR body. A human with export/arch context should confirm behavioral parity on a real MoE export and sign off on the design.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -779,31 +780,199 @@ def _export_quantized_weight(
     torch.cuda.empty_cache()
 
 
+def _has_fused_experts_quantizers(module: nn.Module) -> bool:
+    first_proj_attr = getattr(module, "_first_proj_attr", "gate_up_proj")
+    return hasattr(module, f"{first_proj_attr}_weight_quantizers")
+
+
+@ExportModuleRegistry.register("QuantMoELinear", predicate=lambda m: hasattr(m, "experts"))
```
cjluo-nv: how about moving this registration to separate files?

cjluo-nv (commented):

**Re: operator question "is this a pure no-op refactor?"** — It is *mostly* behavior-preserving, but I would not call it a **pure** no-op. It's a faithful mechanical translation of two if/elif dispatchers into an ordered first-match-wins `ExportModuleRegistry`, and I traced the main paths carefully — but there are a few genuine behavioral divergences (all appear benign today, but they rest on unverified assumptions and the affected paths are GPU/large-model-only, so a human should confirm on a real MoE export).

Behavioral parity trace:
- `_MoELinearExporter` (QuantMoELinear+experts) reproduces the old amax-fill + `continue` (first-match stops dispatch). ✓
- `_QuantLinearExporter`/`_QuantEmbeddingExporter` preserve the format-NONE guards and AssertionError wrapping. ✓ `is_quantlinear("QuantMoELinear")` is correctly False (no substring), so no double-export. ✓
- Prepass `NotImplementedError` for unknown non-iterable experts is retained via `match() is None`. ✓

Not-quite-no-op divergences (why not "pure"):
1. **New `_IterableExpertsExporter` catch-all in `_process_quantized_modules`.** The old whole-model walk had *no* iterable catch-all; now any `nn.ModuleList`/`nn.Sequential` matches it. Its `export()` is a no-op so it's harmless, but it is new dispatch behavior that didn't exist before.
2. **Registry ordering assumption for DBRX.** Old `_process_quantized_modules` had *no* DBRX handling and ran the fused-experts branch (`_has_fused_experts_quantizers`) unconditionally; the new registry puts `_DbrxExpertsExporter` (no-op `export`) *before* `_FusedExpertsExporter`. Parity holds only because DBRX uses `w1_linear/w2_linear/v1_linear` (not `<first_proj>_weight_quantizers`), so `_has_fused_experts_quantizers` is False for it — true today, but the equivalence now silently depends on that.
3. **Prepass key change** from quant class names (`QuantGptOssExperts`/`QuantLlama4TextExperts`) to base names (`GptOssExperts`/`Llama4TextExperts`) matched via MRO — equivalent in practice, different mechanism.

Testing: CPU suite (`test_export_registry.py`) covers MRO/name/predicate matching, precedence, prepend, re-registration, and a tiny FP8 `ToyModel` export — validates wiring, not the full-size DBRX/Llama4/GptOss/fused/QuantMoELinear exports (the codebase's known GPU-only gap). So the "pure no-op" claim is untested on exactly the paths where the ordering divergences above would matter.

Previous-review items:
- :speech_balloon: The "why a 2nd registry vs extending `QuantModuleRegistry`/`_DMRegistryCls`" rationale is in the `registry.py` docstring (quantization swaps classes in place; export must emit weights without changing class + needs predicate/structural matching). Reasonable, but still only in the source docstring — the PR body just says "refactor." Worth a maintainer confirming the second registry over extending the existing one.
- The inline comment "move this registration to separate files?" (unified_export_hf.py:788) is **not** addressed — the exporter classes still live inline in `unified_export_hf.py`. Minor/optional.

Licensing: new `registry.py`/test use the standard NVIDIA Apache-2.0 header (year 2026, matches canonical) — no licensing concern. No prompt-injection in untrusted fields.

Net: strong, well-tested refactor and very likely behavior-preserving, but not provably a *pure* no-op — the iterable catch-all and the DBRX-vs-fused registration ordering are new behaviors whose equivalence hinges on assumptions untested by the CPU suite. Recommend a human sign-off on a real MoE (DBRX/Llama4/GptOss) export before merging.

modelopt/torch/export/unified_export_hf.py
sychen52: Yes. @shengliangxu mentioned this as well during our infra meeting. I guess we may eventually move them to modellib. But does it make sense to keep it here for now? @h-guo18

sychen52: I moved them out to hf_export_handlers.py now.

cjluo-nv (commented):

The registry split resolves the prior dispatch concerns: preparation and export now have independent precedence, the iterable fallback exists only in preparation, DBRX can prepare via its type-specific handler while independently taking the fused export path, and the handler registrations have been moved out of `unified_export_hf.py`. The focused dispatch tests cover these ordering properties, and I found no new correctness bug. The new files also use the canonical 2026 NVIDIA header.

Remaining reasons for owner sign-off:
- The CPU tests validate matching and a toy linear export, but still do not execute the actual DBRX/Llama4/GptOss/fused-experts/QuantMoELinear handlers. Given the operator's confirmation that current behavior is identical, this is not a code-level blocker, but the GPU/large-model paths remain the main unverified risk.
- This introduces a second registry mechanism alongside `QuantModuleRegistry`/`_DMRegistryCls`. The source docstring now gives a reasonable rationale (export handlers do not replace classes, need structural matching, and preparation/export require independent ordering), but the PR body still does not record that design choice as required for a new registry subsystem. Please capture that short rationale in the PR description or have an owner explicitly sign off on it.

cjluo-nv (approved):

PR #1943, opened 2026-07-08
[6385267][ONNX][Autocast] Preserve graph-output Cast producers

gcunhase:
### What does this PR do?

Type of change: Bug fix

- Updates `modelopt/onnx/autocast/precisionconverter.py` so pre-existing floating-point `Cast` nodes are preserved when they are valid producers of graph outputs.
- Keeps graph-output `Cast` nodes only when they are semantically necessary, such as when the Cast defines the declared graph output element type or protects a removable Cast chain.
- Deduplicates graph-output producers introduced during autocast cleanup so quantized autocast models remain valid ONNX SSA graphs.
- Keeps downstream consumers attached to the retained graph-output Cast while only rewiring the upstream Cast path to the renamed pre-cast producer.
- Adds regression coverage in `tests/unit/onnx/autocast/test_precisionconverter.py` for Cast-chain graph outputs, removable same-type output Casts, and duplicate output-producer cleanup.

### Usage

```python
$ python -m modelopt.onnx.autocast --onnx=model.onnx
```

### Testing

- Reproduced the failure locally before the fix: conversion raised `Exception("Sanity Check Failed")` after ONNX validation reported that graph output `t2` was not produced by any node.
- Verified the same repro after the fix: conversion completed and reported `Converted 2/2 nodes (100.00%) to fp16`.
- Ran `pytest -o addopts='' --confcutdir=tests/unit/onnx/autocast tests/unit/onnx/autocast/test_precisionconverter.py::test_convert_preserves_cast_chain_graph_output tests/unit/onnx/autocast/test_precisionconverter.py::test_remove_same_type_graph_output_cast_with_stable_producer tests/unit/onnx/autocast/test_precisionconverter.py::test_deduplicate_network_output_producers_keeps_consumers_on_cast_output -q`
- Ran `pytest -o addopts='' --confcutdir=tests/unit/onnx 'tests/unit/onnx/test_autocast_quantize.py::test_autocast_quantize_int8[True-True]' 'tests/unit/onnx/test_autocast_quantize.py::test_autocast_quantize_int8[False-True]' -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cast cleanup and removal for mixed-precision models by correctly preserving casts that define a graph output’s declared element type.
  * Enhanced network output wiring to handle cast boundaries, including deduplicating multiple producers so graph outputs follow a single consistent cast path without breaking downstream consumers.
* **Tests**
  * Added regression coverage for cast chains involving graph outputs.
  * Added regression tests for removing same-type graph output casts while keeping output wiring correct.
  * Added a regression test for deduplicating multiple network output producers around cast boundaries.


cjluo-nv (approved):

Small, focused bug fix (+37/-3, 2 files) in the ONNX AutoCast `PrecisionConverter._remove_preexisting_casts`. 

**What it does:** Previously a graph-output Cast was preserved only when *both* its input and output were graph outputs (the output→cast→output pattern). A valid cast chain like `in → Cast → t1 → Cast → t2(output)` had its final cast removed and bypassed, leaving `t2` dangling and failing the sanity check. The new condition keeps a Cast when its output is a graph output AND the cast target type matches the declared output element type — the correct, broader predicate that still preserves the old chain case.

**Correctness:** Verified the logic. The change also broadens preservation to casts whose input isn't a graph output (e.g. `test_multiple_output_node_casted_to_output`, `test_casted_input_to_output_model`) but those tests assert only `check_model`, which stays valid whether the cast is kept or removed. The type-match guard prevents keeping mismatched output casts that `keep_io_types=True` should still repair.

**Tests:** New `test_convert_preserves_cast_chain_graph_output` reproduces the reported failure end-to-end through `convert_to_mixed_precision` (exercises the default `keep_io_types=False` path). I confirmed via the GraphSanitizer that the DOUBLE casts are converted to FP32 before `_remove_preexisting_casts`, so the test genuinely hits the fixed branch. It asserts the output is produced by exactly one Cast node and that the model validates.

No licensing changes (existing Apache-licensed file, headers untouched), no new abstractions/subsystems, no prompt-injection in the untrusted PR content. Minor nit (non-blocking): unlike neighboring tests, the new test isn't parametrized over `keep_io_types`/`bf16`.

PR #1944, opened 2026-07-08
Fix distributed amax error handling

realAsma:
## Description

`TensorQuantizer.sync_amax_across_distributed_group` currently catches every `RuntimeError` from `dist.all_reduce` and converts it to a warning. This can mask failures such as NCCL OOMs, allowing ranks to diverge and hang in later collectives.

This PR handles only the intended unsupported-device case before entering the collective:

- skip distributed amax synchronization when the amax tensor is on CPU and the process-group backend is NCCL
- retain the existing restore-oriented warning for that case
- let all errors from supported collectives propagate normally

## Tests

- `pytest_pwd tests/unit/torch/quantization/test_dist.py -q`
- `pre-commit run --files modelopt/torch/quantization/nn/modules/tensor_quantizer.py tests/unit/torch/quantization/test_dist.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved distributed quantization synchronization for NCCL runs when the quantization range tensor is on CPU.
  * Added an explicit warning and an early exit to prevent unsupported all-reduce attempts.

* **Tests**
  * Added a multi-process CUDA/NCCL test covering CPU-based range tensors, verifying the warning is emitted and the value stays on CPU as expected.

### DIFF
--- a/tests/unit/torch/quantization/test_dist.py
+++ b/tests/unit/torch/quantization/test_dist.py
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from unittest.mock import Mock, patch
+
+import pytest
 import torch
 import torch.distributed as dist
 from _test_utils.torch.distributed.utils import spawn_multiprocess_job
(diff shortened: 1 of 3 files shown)


tests/unit/torch/quantization/test_dist.py
realAsma, line 1: BB: Move this unit test to gpu tests and use a real test without monkey patching on GPUs (2 GPU test, backend is nccl, quantizer amax is intentionally on CPU)

realAsma: > _:robot: Bot comment._

🐝 Got it — I’ll move this coverage to the GPU suite and replace the monkeypatched case with a real 2-GPU NCCL test that keeps the quantizer amax on CPU.

realAsma: > _:robot: Bot comment._

Implemented in `daf150a5d0`. The mocked CPU/NCCL case was removed from the unit suite and replaced with a real 2-GPU NCCL test in the GPU tensor-quantizer suite; each rank intentionally keeps a distinct amax on CPU and verifies it remains unchanged. The separate unit regression for propagating collective failures remains in place.

Validation: focused unit tests (2 passed), real 2-GPU NCCL test (1 passed), all applicable pre-commit hooks, and `git diff --check`.

realAsma: BB: Can we get rid of this test? GPU test already covers this.

realAsma: > _:robot: Bot comment._

🐝 Yes — I’ll remove the now-redundant unit test and keep the real GPU coverage.

realAsma: > _:robot: Bot comment._

Removed in `73e88e476d`; the real two-GPU NCCL test remains as the coverage. The remaining `test_dist.py` test passes, along with all applicable pre-commit hooks and `git diff --check`.

realAsma:
BB: Approved; make this a regular PR

realAsma:
> _:robot: Bot comment._

🐝 Approved—I'll mark this PR ready for review now.

realAsma:
> _:robot: Bot comment._

Done — PR #1944 is now a regular, ready-for-review PR. GitHub confirms it remains open and mergeable at `73e88e476d`; no code or branch content changed.

realAsma:
> _:robot: Bot comment._

Keeping `warnings.warn` here: this method accepts arbitrary process groups, while `warn_rank_0` gates on global rank 0 and could suppress the warning entirely for a subgroup that excludes global rank 0. This restore-only warning should remain visible to each affected participant.

PR #1945, opened 2026-07-08
fix(skills): Update Agent Skills Based on Observed Failures Modes in Trials

chadvoegele:
### What does this PR do?

Type of change: documentation

Updates the agent skills based on findings from the June 5 Day-0 trials:

- Strengthens PTQ checkpoint handoff, serving-readiness, and representative calibration guidance.
- Adds external baseline sanity checks while allowing comparisons when no credible external score exists.
- Pins SciCode to task-level `parallelism: 8` and `num_repeats: 1`.
- Adds positional BF16 layer exclusions as a quantization-recipe search axis.
- Keeps day-0 release decisions autonomous and maps external baseline mismatches to an explicit failure class.

Refs:
1. https://lab-e57330.gitlab-master-pages.nvidia.com/day0_jun5_analysis/
2. https://docs.google.com/spreadsheets/d/1H-YvO5CDwKoG2IvvNWVdDI7rViW1Kc0ayVEJAP7RH_A/edit?gid=30281591#gid=30281591&range=A2

### Usage

N/A — agent skill guidance only.

### Testing

- Ran structural validation for all five changed skill packages.
- Ran targeted pre-commit hooks on all nine changed files; all passed.
- Ran `git diff --check`.

I'll test the skill changes by re-running day 0 trials after merging.

- Did you write any new necessary tests?: N/A — documentation/skill guidance only
- Did you get Claude approval on this PR?: ❌ — not run

### Additional Information

The branch contains one signed commit and excludes WAAP state, test-fixture changes, and example changes.

## Summary by CodeRabbit

* **Documentation**
  * Strengthened evaluation workflows with an “External Baseline Sanity Check” requirement and clearer success/failure handling when external verification fails or is inconclusive.
  * Updated day-0 and compare-results guidance to follow the new gate-driven sequence and to report external reference scores and sanity status.
  * Refined SciCode instructions to use fixed repeat settings and consistent scoring output.
  * Expanded PTQ checkpoint validation with downstream serving-readiness canary requirements.
  * Enhanced quantization recipe search with a layer-position axis, positional exclusion rules, and stricter promotion criteria.

### DIFF
--- a/.agents/skills/compare-results/SKILL.md
+++ b/.agents/skills/compare-results/SKILL.md
@@ -32,11 +32,19 @@ change is being measured, typically a further quantized version of the baseline.
 5. For each task, use the canonical score field from the matching
    `.agents/skills/evaluation/recipes/tasks/<task>.md` Score Extraction
    section.
-6. Compute exact deltas outside the chat context when there are multiple tasks
+6. Read and perform `.agents/skills/evaluation/references/run-validation.md`
+   **External Baseline Sanity Check**. Record each source URL, protocol
+   difference, and task status before applying the candidate-delta gate. A
+   failed baseline blocks a success verdict; correct and rerun it first. If no
+   credible comparable reference exists, label the baseline externally
+   unverified rather than claiming the check passed, then continue using the
+   validated measured baseline.
+7. Compute exact deltas outside the chat context when there are multiple tasks
    or repeated runs.
-7. Report comparability and quantized-feasibility verdicts before interpreting
-   the delta as model quality. If the user did not provide an acceptance
-   threshold, report feasibility as inconclusive instead of inventing one.
+8. Report comparability, external baseline sanity, and quantized-feasibility
+   verdicts before interpreting the delta as model quality. If the user did not
+   provide an acceptance threshold, report feasibility as inconclusive instead
+   of inventing one.
 
 ## Comparability Checklist
 
@@ -64,8 +72,16 @@ the validated runs are comparable:
    precision the baseline is**, and apply the gate relative to that baseline
    rather than to an assumed BF16.
 
+For SciCode, keep `num_repeats: 1` to limit sandbox workload. If variance is a
+concern, run multiple independent matched baseline/candidate pairs instead of
+increasing repeats within one run.
+
 If any item differs, either rerun with matched settings or label the result as
 not an apples-to-apples quantization comparison.
+
+These checks compare the baseline and candidate to each other. The external
+baseline check in `evaluation/references/run-validation.md` separately tests
+whether the baseline's absolute score is credible; both guards must be reported.
 
 ## Report Format
 
@@ -74,7 +90,13 @@ Include:
 - Baseline and candidate identifiers.
 - Per-task metric path, baseline score, candidate score, delta, and stderr if
   available.
+- Per-task external reference score, source URL, known protocol differences,
+  percentage-point difference, and sanity status (`verified`, `failed`, or
+  `externally unverified`).
 - Comparability status for prompt/template, generation settings, sample counts,
   reasoning handling, judge/simulator setup, and score field.
 - Comparability verdict: comparable, not comparable, or inconclusive.
 - Quantization feasibility verdict: acceptable, not acceptable, or inconclusive.
+  Never report `acceptable` when external baseline sanity failed. An externally
+  unverified baseline does not block `acceptable`; apply the candidate-delta
+  gate and report the missing external corroboration.
--- a/.agents/skills/day0-release/SKILL.md
+++ b/.agents/skills/day0-release/SKILL.md
@@ -52,9 +52,9 @@ progress:
 - [ ] Step 0: Resolve inputs; confirm threshold and eval set
 - [ ] Step 1: Setup gate — creds present, cluster reachable
 - [ ] Step 2: PTQ (ptq skill) → gate_ptq.py
-- [ ] Step 3: Baseline eval (evaluation skill, deploys source) → gate_run.py   [skip if cached, see below]
+- [ ] Step 3: Baseline eval (evaluation skill, deploys source) → gate_run.py
 - [ ] Step 4: Quantized eval (evaluation skill, deploys candidate) → gate_run.py
-- [ ] Step 5: Compare (compare-results skill) → gate_compare.py → decision
+- [ ] Step 5: Compare (compare-results skill) → external sanity → gate_compare.py → decision
 - [ ] Step 6: Closeout — report + publish recommendation
 ```
 
@@ -83,10 +83,8 @@ unvalidated checkpoint.
 ### Step 3 — Baseline eval
 
 The baseline is the **source** (pre-quantization) model on the same task set and
-sampling params. **Look it up first** — if a matching baseline run already
-exists in MLflow (same model, task set, sampling params), reuse it and skip this
-stage. Otherwise run it via the **evaluation** skill (which deploys the source
-model itself). Gate with `gate_run.py`.
+sampling params. Always run a fresh baseline via the **evaluation** skill,
+which deploys the source model itself. Gate with `gate_run.py`.
 
 ### Step 4 — Quantized eval
 
@@ -110,7 +108,14 @@ dropped samples) — do **not** compare scores from it.
 
 ### Step 5 — Compare
 
-Invoke the **compare-results** skill to produce per-task deltas, then gate:
+Invoke the **compare-results** skill. It must perform the shared external
+baseline sanity check before the candidate-delta gate. A failed check is
+`ANOMALOUS` with failure class `EXTERNAL_BASELINE_MISMATCH`: investigate and
+rerun the baseline. If no credible comparable external score exists, record the
+baseline as externally unverified and continue using the validated measured
+baseline.
+
+After recording the external status, produce per-task deltas and run:
 
 ```bash
 python .agents/skills/day0-release/scripts/gate_compare.py \
@@ -125,22 +130,25 @@ normalizes the drop accordingly, so `--threshold 0.01` means "≤1 pt on a 0-100
 task / ≤0.01 on a 0-1 task" uniformly. Pass `--scales '{"task": max}'` to
 override inference if a task's scores happen to fall in an ambiguous range.
 
-Decision from `gate_compare.py`:
+`gate_compare.py` checks only the candidate delta; it cannot override a failed
+external baseline check. Combined decision:
 
-- **ACCEPT** — every task within threshold → go to Step 6.
+- **ACCEPT** — no external check failed and every task is within the candidate
+  threshold → go to Step 6. A missing comparable external score is not a
+  failure; report it as externally unverified.
 - **REGRESSION** — one or more tasks exceed threshold. **v1 stops here and
   reports** which tasks regressed by how much. (Picking the next recipe and
   re-running is deferred — see Scope.)
-- **ANOMALOUS** — scores present but implausible (e.g. baseline lower than
-  candidate by a large margin, or a task score outside its valid range) →
-  surface to the user.
+- **ANOMALOUS** — external baseline sanity failed, or scores are otherwise
+  implausible (e.g. baseline lower than candidate by a large margin, or a task
+  score is outside its valid range) → correct the baseline or surface it.
 
 ### Step 6 — Closeout
 
 Report the decision with: source vs output size + ratio, per-task baseline /
-candidate / delta / within-threshold, MLflow run IDs, and a publish
-recommendation (publish / do-not-publish / needs-human). Archive artifacts to
-the workspace.
+candidate / delta / within-threshold, external source and sanity status, MLflow
+run IDs, and a publish recommendation (publish / do-not-publish).
+Archive artifacts to the workspace.
 
 ## Triage (gate failure → decision)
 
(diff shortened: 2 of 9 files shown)


cjluo-nv (commented):

Documentation-only PR (+200/-40, 9 files) updating existing agent-skill markdown based on the June-5 day-0 trial failures. No code/tests, no licensing changes (Apache-2.0 frontmatter is pre-existing and untouched).

**Design gate**: The complexity gate fired on directory span, but this isn't an architectural change — it's guidance-text edits to *existing* skills. The one new concept (the "External Baseline Sanity Check" procedure in `run-validation.md`, wired into `compare-results`, `day0-release`, and `quant-recipe-search`) is a documented process, not a new code subsystem/DSL/registry, so there's no in-repo system being duplicated. Design protocol effectively N/A.

**Correctness / consistency verified**:
- SciCode: `num_repeats: 1` + score field `avg-of-1` are internally consistent.
- day0-release: MLflow baseline reuse/caching removed from *both* the checklist and Step 3 text (consistent), new `EXTERNAL_BASELINE_MISMATCH` class added to Triage/Step 5/run-validation consistently, and `NEEDS_HUMAN` removal leaves no dangling references (confirmed via search).
- Factual references check out against the actual code: `nemotron-post-training-v3` (dataset_utils.py), `--calib_with_images`, and the VLM default subsets `sparsetables`/`plotqa_cot`/`wiki_en` exactly match `default_subsets` in `vlm_dataset_utils.py`.

No prompt-injection: the directive-style text in the diff ("Do NOT use…", "Always run a fresh baseline") is the skill content under review, not an attempt to steer this review.

**Why nudge (owner should confirm two embedded policy calls):**
- Step 3 now *forces a fresh baseline eval every run* (removing MLflow baseline reuse). This is a deliberate cost/accuracy trade-off worth a human sign-off — it increases per-run compute.
- The External Baseline Sanity Check hardcodes an "~5 percentage points" pass threshold; a human should confirm that heuristic is right for the AA suite.

PR body notes it was not Claude-approved and will be validated by re-running day-0 trials, so a human review before merge is appropriate.

.agents/skills/ptq/references/checkpoint-validation.md
```diff
@@ -1,12 +1,13 @@
...
+4. The checkpoint is ready for downstream deployment and evaluation. Record the exact checkpoint workspace and path that both downstream skills must inherit. Inventory every model-compatibility change made during PTQ that may be required to load or serve the checkpoint: dependency upgrades, source patches, custom code, environment variables, and launcher or container changes. Record each category separately and write `none` for every category with no changes. Then invoke the **deployment** skill and use that same workspace, checkpoint path, and compatibility inventory to serve the checkpoint in the intended deployment environment. Run a canary query such as `What is the capital of France?` and require a valid response (for this example, one that identifies Paris). Record the target environment, serving framework and launch configuration, canary query, and response. Stop the canary service after validation unless the user asked to keep it running.
```
Edwardf0t1, line 10: **Mandatory serving canary → cost multiplier under recipe search.** This makes group 4 (invoke `deployment`, serve, run a canary) a required gate for *every* PTQ hand-off. But `quant-recipe-search` delegates PTQ + validation per candidate, so during a sweep this implies a full deployment per candidate — serializing the search on a deploy each time. Consider scoping the mandatory canary to the promoted/day-0 checkpoint and marking it optional during exploratory candidate generation (or stating that intent explicitly), so the recipe-search loop does not pay a deploy per candidate.

.agents/skills/day0-release/SKILL.md
```diff
@@ -83,10 +83,8 @@ unvalidated checkpoint.
 ### Step 3 — Baseline eval
 
 The baseline is the **source** (pre-quantization) model on the same task set and
-sampling params. **Look it up first** — if a matching baseline run already
-exists in MLflow (same model, task set, sampling params), reuse it and skip this
-stage. Otherwise run it via the **evaluation** skill (which deploys the source
-model itself). Gate with `gate_run.py`.
+sampling params. Always run a fresh baseline via the **evaluation** skill,
```
Edwardf0t1, line 86: **Dropping MLflow baseline reuse increases every run's compute/latency.** This flips Step 3 from "look it up first, reuse if cached" to "always run a fresh baseline." Defensible as correctness-over-speed given the trials, but it is a real regression for repeated day-0 runs on the same source model. Worth a one-line rationale here so a future reader does not "re-optimize" the reuse path back in.

.agents/skills/evaluation/references/run-validation.md
```diff
@@ -35,6 +35,55 @@ accounting, reasoning/answer parsing status, and any errors or warnings found.
...
+   externally unverified.
+3. Put both scores on a 0-100 scale, then calculate, for higher-is-better
+   metrics:
+
+   ```text
+   difference (pp) = abs(measured baseline - external)
+   ```
+
+   Treat a credible comparable result as verified only when the absolute
+   difference is approximately 5 percentage points or less. A difference
```
Edwardf0t1, line 65: **The "~5 percentage-point" external threshold is a hardcoded heuristic that can misfire.** For many benchmarks no published external run shares the protocol (template / thinking-mode / sample-count differ), so a legitimate baseline can land >5pp from the only reference and trip `failed → ANOMALOUS → block comparison`. The section does soften this (prefer `externally unverified`, which is non-blocking), but consider tightening the wording so the default when protocols are not closely matched is `externally unverified`, reserving `failed` for a genuinely comparable protocol that still shows a large gap.

Edwardf0t1 (commented):
Three follow-up notes from the review — all guidance/cost trade-offs rather than defects, non-blocking.

.agents/skills/day0-release/SKILL.md
```diff
@@ -52,9 +52,9 @@ progress:
 - [ ] Step 0: Resolve inputs; confirm threshold and eval set
 - [ ] Step 1: Setup gate — creds present, cluster reachable
 - [ ] Step 2: PTQ (ptq skill) → gate_ptq.py
-- [ ] Step 3: Baseline eval (evaluation skill, deploys source) → gate_run.py   [skip if cached, see below]
+- [ ] Step 3: Baseline eval (evaluation skill, deploys source) → gate_run.py
```
Edwardf0t1, line 55: Did the agent actually run skills and `gate_run.py` in day0-release?

.agents/skills/evaluation/recipes/tasks/aa/scicode.md
```diff
@@ -28,13 +30,12 @@ Use this inside the top-level `evaluation.tasks` list:
   nemo_evaluator_config:
     config:
       params:
+        parallelism: 8
         extra:
           args: ++prompt_config=eval/scicode/default ++with_background=true
-          num_repeats: 8
+          num_repeats: 1
```
Edwardf0t1, line 36: We only need 1 repeat for scicode?

.agents/skills/ptq/SKILL.md
```diff
@@ -1,12 +1,21 @@
 ---
 name: ptq
-description: This skill should be used when the user asks to "quantize a model", "run PTQ", "post-training quantization", "NVFP4 quantization", "FP8 quantization", "INT8 quantization", "INT4 AWQ", "quantize LLM", "quantize MoE", "quantize VLM", or needs to produce a quantized HuggingFace or TensorRT-LLM checkpoint from a pretrained model using ModelOpt.
+description: >-
+  Use when the user asks to "quantize a model", "run PTQ", "post-training
+  quantization", "NVFP4 quantization", "FP8 quantization", "INT8
+  quantization", "INT4 AWQ", "quantize LLM", "quantize MoE", "quantize VLM",
+  or needs to produce a quantized HuggingFace or TensorRT-LLM checkpoint from a
```
Edwardf0t1: Do we need `TensorRT-LLM checkpoint` here?

.agents/skills/quant-recipe-search/SKILL.md
```diff
@@ -103,18 +106,29 @@ Do not collapse the search to one dimension such as numeric format only. Read
...
      misses the best frontier or hits runtime constraints.
+   - When sensitivity or model behavior implicates boundary layers, add a
+     controlled first-layer, last-layer, or combined BF16 exclusion candidate.
+     Do not preserve boundary layers without testing the trade-off.
 
 4. **Generate candidates**
    - Delegate checkpoint generation and PTQ validation to `ptq`.
    - Change one major axis at a time: format, calibration algorithm, module
-     selection, granularity, exclusions, or calibration data.
+     family, layer position, granularity, or calibration data.
```
Edwardf0t1, line 116: So the agent went through the quant recipe search skills in our trial run, right?

.agents/skills/day0-release/SKILL.md
chadvoegele: No, the `day0-release` and `gate_run.py` were added to the skills a few days after the first round of agent trials.

.agents/skills/compare-results/SKILL.md
```diff
@@ -64,17 +72,31 @@ the validated runs are comparable:
    precision the baseline is**, and apply the gate relative to that baseline
    rather than to an assumed BF16.
 
+For SciCode, keep `num_repeats: 1` to limit sandbox workload. If variance is a
+concern, run multiple independent matched baseline/candidate pairs instead of
+increasing repeats within one run.
```
meenchen, line 77: Why specific to scicode? and is this for faster recipe iteration?

.agents/skills/quant-recipe-search/SKILL.md
chadvoegele: No, looks like `quant-recipe-search` was there when the trials were run but no agent used it. https://gitlab-master.nvidia.com/cvoegele/lab/-/blob/main/wiki/day0_jun5_analysis/skill_usage.md

.agents/skills/evaluation/recipes/tasks/aa/scicode.md
chadvoegele: We need repeats to reduce variance, but when using `num_repeats > 1` the evaluation will re-use the same code execution service. If there are any leaked sockets, threads, etc. you might eventually see a `Resource temporarily unavailable` or similar. It's more likely to be correct using a fresh code sandbox for every repeat by running scicode evaluation multiple times with `num_repeats: 1`.

.agents/skills/ptq/SKILL.md
chadvoegele: Looks like that was there before so I didn't change it. I can remove it.

.agents/skills/evaluation/references/run-validation.md
chadvoegele: I took 5 percentage points from product guidance.

> Shared metric exists and internal baseline is within +/- 5 pp absolute

.agents/skills/day0-release/SKILL.md
chadvoegele: Agree it's more inefficient, but it's what we do in practice anyways to verify correctness of the evals.

.agents/skills/ptq/references/checkpoint-validation.md
chadvoegele: Ya, we should make sure the quantized checkpoint can be deployed, otherwise we need to choose a new recipe.

.agents/skills/evaluation/recipes/tasks/aa/scicode.md
meenchen: @chadvoegele , does this mean the agent needs to do the averaging itself with multiple runs of `num_repeats: 1`?

.agents/skills/day0-release/SKILL.md
meenchen, line 86: Why do we need a fresh baseline?

.agents/skills/evaluation/recipes/tasks/aa/scicode.md
chadvoegele: Yes

.agents/skills/day0-release/SKILL.md
chadvoegele: I'm thinking for a few reasons:

1. It will re-validate the particular evaluation setup used by the agent, i.e. config is right, NEL version is aligned, evaluation container is same, etc.
2. I've not yet seen anyone (human or agent) go to MLFlow to fetch a previous baseline successfully.
3. Typically we're doing new models anyways so there isn't a previous baseline.

juhi10071998: @chadvoegele , but for point 3, if the agent has already run a baseline once (not available before), we should be able to use that previous result right, like humans do

.agents/skills/compare-results/SKILL.md
chadvoegele: No, `num_repeats: 1` is to help eliminate accuracy degradation due to code sandbox errors.

I did a little more research on it. For SciCode the architecture is:

```
nemo-skills container
├─ ns eval process
└─ sandbox server: local_sandbox_server (port 6000)
   └─ worker process (lifetime: evaluation run)
```

Whereas for LiveCodeBench, it's:
```
nemo-skills container
└─ ns eval process
   └─ LiveCodeBench evaluator workers (lifetime: one model generation)
```

With SciCode then any errors in the sandbox can accumulate during the evaluation, which might eventually lead to failures. Since LiveCodeBench is per generation, we don't have the same failure mode.

The newer additions SWE-bench Verified and Terminal Bench use yet another path of a Harbor eval image + a remote AWS Fargate execution container connected by an SSH tunnel.

.agents/skills/day0-release/SKILL.md
chadvoegele: Ya, the agent should re-use within a session.

Across different sessions, I think it would be better to re-validate. If we're re-running so much that it becomes a bottleneck, we can reassess.

.agents/skills/evaluation/recipes/tasks/aa/scicode.md
Edwardf0t1: I see, so I assume we need `num_repeats: 1` for all benchmarks?

.agents/skills/ptq/SKILL.md
Edwardf0t1: I think TensorRT-LLM checkpoint is deprecated.

.agents/skills/day0-release/SKILL.md
Edwardf0t1: I see, would be interesting to see if agent will use it in the next run.

.agents/skills/ptq/SKILL.md
chadvoegele: I removed it from here

.agents/skills/evaluation/recipes/tasks/aa/scicode.md
chadvoegele: No, this one is Scicode specific. See the explanation in https://github.com/NVIDIA/Model-Optimizer/pull/1945#discussion_r3553040347

.agents/skills/quant-recipe-search/SKILL.md
Edwardf0t1: I see, I thought the changes in this PR are all related to the agent's behaviors in the first run, would be interesting to see if agent will use it in the next run.

Edwardf0t1 (approved):
LGTM!

.agents/skills/quant-recipe-search/SKILL.md
chadvoegele: Ya, I'm hoping the agent can use the text of this skill to improve one of the models. But if the skill is still not being picked up, then we'll have to diagnose further.

PR #1946, opened 2026-07-08
[6385268][ONNX][Autocast] Fix BF16 autocast for FP16 initializers

gcunhase:
### What does this PR do?

Type of change: Bug fix

- Fixes ONNX autocast when converting a valid model with a FLOAT16 initializer to BF16.
- `PrecisionConverter` now derives initializer conversion from the initializer's actual ONNX dtype instead of assuming the source dtype from consumer precision.
- BF16 initializer serialization now works from supported source float types via an fp32 intermediate.
- The `keep_io_types=False` sanity check now allows supported floating-point I/O to convert to the selected low precision.
- Applies the same actual-dtype initializer handling to control-flow subgraphs.
- Adds regression tests covering BF16 conversion of FP16 initializers in the main graph and If subgraphs.

### Usage

```python
$ python -m modelopt.onnx.autocast --onnx=model.onnx --low_precision_type bf16
```

### Testing

- Reproduce the failure locally before the fix: `AssertionError: Initializer w0 is not of type fp32`.
- Verified the minimal reproducer now converts successfully: `Converted 1/1 nodes (100.00%) to bf16`.
- Ran `python -m pytest tests/unit/onnx/autocast/test_precisionconverter.py::test_bf16_conversion_accepts_fp16_initializer tests/unit/onnx/autocast/test_precisionconverter.py::test_bf16_if_subgraph_conversion_accepts_fp16_initializers -q`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initializer precision handling by deriving conversion type from each initializer’s actual ONNX data type.
  * Enhanced BF16 downcasting support, including BF16 conversion of FP16 initializers and correct value decoding.
  * Broadened validation when preserving I/O types, allowing more original floating-point formats to convert cleanly.

* **Tests**
  * Added unit coverage for converting FP16 initializers to BF16, including graphs with FP16 value infos.
  * Added tests verifying FP16 initializers inside `If` subgraphs are converted to BF16 as expected.

### DIFF
--- a/modelopt/onnx/autocast/precisionconverter.py
+++ b/modelopt/onnx/autocast/precisionconverter.py
@@ -621,7 +621,6 @@ class PrecisionConverter:
                         f"Convert initializer {init_name} to "
                         f"{self.low_precision_type.str_short}, only used by low precision nodes"
                     )
-                    from_type = self.high_precision_type
                     to_type = self.low_precision_type
                 elif len(tracker.high_precision_nodes) > 0:
                     logger.debug(
@@ -629,12 +628,16 @@ class PrecisionConverter:
                         f"{self.high_precision_type.str_short}, "
                         "only used by high precision nodes"
                     )
-                    from_type = self.low_precision_type
                     to_type = self.high_precision_type
                 else:
                     raise ValueError(
                         f"Unexpected: initializer {init_name} is not used by any "
                         "nodes and is not a float"
+                    )
+                from_type = self._precision_type_from_onnx_type(init.data_type)
+                if from_type is None:
+                    raise ValueError(
+                        f"Unexpected: initializer {init_name} is not a supported float"
                     )
 
                 new_init = self._cast_initializer(
@@ -743,6 +746,10 @@ class PrecisionConverter:
 
         utils.walk_subgraphs_recursive(self.model.graph, _convert_subgraph_callback)
 
+    def _precision_type_from_onnx_type(self, onnx_type: int) -> PrecisionTypes | None:
+        """Return the converter precision metadata for a supported ONNX float type."""
+        return next((p for p in PRECISION_MAP.values() if p.onnx_type == onnx_type), None)
+
     def _convert_initializer_data(
         self,
         init: onnx.TensorProto,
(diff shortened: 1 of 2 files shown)


modelopt/onnx/autocast/precisionconverter.py
```diff
@@ -743,6 +746,10 @@ def _convert_subgraph_callback(
```
cjluo-nv, line 739: > _Bot comment._

The main-graph fix (deriving `from_type` from `init.data_type`) is not mirrored here in the subgraph path. For a BF16 conversion, a FLOAT16 subgraph initializer matches neither `high_precision_type` (FLOAT) nor `low_precision_type` (BFLOAT16), so `from_type` becomes `None` and the initializer is skipped — leaving an FP16 tensor inside a BF16 subgraph. Consider reusing `_precision_type_from_onnx_type(init.data_type)` here too so subgraph initializers get the same treatment, or confirm this case can't arise in practice.

cjluo-nv (commented):

Small, focused bug fix (+55/-8, 2 files) for ONNX autocast: BF16 conversion of an FP16 initializer previously hit `AssertionError: Initializer w0 is not of type fp32`. The fix correctly (1) derives `from_type` from the initializer's actual `data_type` via `_precision_type_from_onnx_type` instead of assuming it from consumer precision, (2) reads BF16 source tensors through `read_f16_tensor_as_fp32` and serializes BF16 via an fp32 intermediate (`.astype(np.float32).astype(ml_dtypes.bfloat16)`), and (3) broadens the `keep_io_types=False` sanity-check exception from `== FLOAT` to `in ONNX_TYPES`. I traced the affected branches (FP16→BF16, FP16→FP32, BF16→FP32, both-precision duplication) and they look correct; the `_cast_initializer` assert still holds since `from_type` is now derived from `init.data_type`. A parametrized regression test covers both `use_standalone_type_inference` paths.

Why nudge (not approve):
- The parallel subgraph path in `_convert_initializers_recursive._convert_subgraph_callback` was NOT updated: it still derives `from_type` only by matching `high_precision_type`/`low_precision_type`, so an FP16 initializer inside an If/Loop/Scan subgraph during a BF16 conversion resolves to `from_type=None` and is silently skipped — leaving an FP16 initializer in a BF16 subgraph. The new test only exercises the main graph, so this gap is unverified. A maintainer should confirm whether real models with mismatched-precision subgraph initializers are in scope.
- No end-to-end run on a real model / the CLI shown in the PR body is captured in CI; the unit test uses a minimal synthetic Add graph.

No licensing changes and no prompt-injection attempts in the PR content.

ajrasane (approved):

modelopt/onnx/autocast/precisionconverter.py
gcunhase: Done, also added regression test. Please check again.

galagam (approved):
LGTM!

PR #1947, opened 2026-07-08
feat(quant): add constant_amax to pin activation input_scale (NVFP4 experts input_scale=1.0)

Edwardf0t1:
### What does this PR do?

Type of change: new feature

Adds a `constant_amax` field to `QuantizerAttributeConfig` that pins a quantizer's `amax` to a fixed value and **skips activation calibration** (no forward statistics collected). Unlike the existing `use_constant_amax` — which hardcodes the amax to the FP8 E4M3 range (448.0) for KV-cache cast math and registers *no* `_amax` buffer — `constant_amax` stores the constant on the `_amax` buffer so it is used by **both** the fake-quant forward pass and the exported scaling factor (e.g. `input_scale`).

Motivation: pin the NVFP4 routed-expert activation `input_scale` to exactly `1.0` in the exported checkpoint, with no activation calibration. For NVFP4 the exported activation scale is `input_scale = amax / (E2M1_MAX * E4M3_MAX) = amax / (6 * 448)`, so `constant_amax = 2688.0` yields `input_scale == 1.0`.

Changes:
- `config.py`: new `constant_amax: float | None` field + positive-value validator.
- `tensor_quantizer.py`: setter registers the constant on the `_amax` buffer.
- `model_calib.py`: `enable_stats_collection` / `finish_stats_collection` exclude constant-amax quantizers from calibration (so the fixed amax is never overwritten).
- New recipe `modelopt_recipes/general/ptq/nvfp4_experts_only_input_scale1-kv_fp8_cast.yaml` applying it to the MoE expert activation quantizers.

### Usage

```python
from modelopt.torch.quantization.config import QuantizerAttributeConfig

# NVFP4 activation quantizer with a fixed input_scale == 1.0, no calibration:
cfg = QuantizerAttributeConfig(
    num_bits=(2, 1),
    block_sizes={-1: 16, "type": "dynamic", "scale_bits": (4, 3)},
    constant_amax=2688.0,  # 6 * 448 = E2M1_MAX * E4M3_MAX -> exported input_scale == 1.0
)
```

Or via the shipped recipe:

```bash
# modelopt_recipes/general/ptq/nvfp4_experts_only_input_scale1-kv_fp8_cast.yaml
python examples/hf_ptq/hf_ptq.py --recipe nvfp4_experts_only_input_scale1-kv_fp8_cast ...
```

### Testing

- New CPU unit tests in `tests/_test_utils/torch/quantization/tensor_quantizer_common.py` (run via `test_tensor_quantizer_cpu.py`):
  - `test_constant_amax_registers_amax` — `_amax` buffer pinned to the constant; default (unset) registers no buffer.
  - `test_constant_amax_nvfp4_input_scale` — `constant_amax=2688` ⇒ exported NVFP4 `input_scale == 1.0`.
  - `test_constant_amax_skips_calibration` — quantizer excluded from calibration; amax not overwritten.
  - `test_constant_amax_must_be_positive` — validator rejects non-positive values.
- `pytest tests/unit/torch/quantization/test_tensor_quantizer_cpu.py tests/unit/torch/quantization/test_config_validation.py` — 109 passed.
- `pre-commit run --files ...` — all hooks pass (incl. `validate modelopt recipes`).
- Verified the new recipe loads through `modelopt.recipe.loader.load_recipe`, with `$import: nvfp4` correctly merging `constant_amax: 2688.0`.

- Is this change backward compatible?: ✅ (new opt-in field defaulting to `None`; no behavior change when unset)

### Additional Information

The exported per-block E4M3 activation scales remain dynamic; only the per-tensor global scale (`input_scale`) is pinned to 1.0.

## Summary by CodeRabbit

* **New Features**
  * Added support for pinning a quantizer’s activation `amax` to a user-specified constant, keeping behavior consistent for export and inference.
  * Added a new expert-only NVFP4 PTQ recipe with FP8 KV-cache casting and fixed expert activation scaling.
* **Bug Fixes**
  * Improved calibration/stat handling so constant-`amax` quantizers are excluded during calibration and restored afterward.
* **Tests**
  * Added coverage for fixed-`amax` behavior, validation rules (positive value, mutual exclusivity), and NVFP4 scaling/export results.


modelopt/torch/quantization/nn/modules/tensor_quantizer.py
```diff
@@ -244,6 +245,13 @@ def _block_sizes_setter(val):
                     self._calibrator._axis = None
             return val
 
+        def _constant_amax_setter(val):
+            if val is not None:
+                # Pin amax to the constant on the _amax buffer so it is used by both the
+                # fake-quant forward pass and export; calibration is skipped in model_calib.
+                self.amax = float(val)
```
juhi10071998, line 252: Hi Zhiyu, the pinned _amax buffer is registered on CPU and never moved to the model's device, will this have any implications. the test does the .to(device) so no errors are caught

Edwardf0t1: Good catch — you're right. The buffer was registered on CPU at config time and would not follow a later `model.to(device)`, so the first post-calibration GPU forward would hit the `amax.is_cuda` assert in the dynamic-block-quant kernel (the `.to(device)` in the test masked it). Fixed in f224624: `_get_amax` now aligns the buffer with the input device on the fly. Thanks!

modelopt/torch/quantization/config.py
```diff
@@ -671,6 +671,43 @@ def validate_calibrator(cls, v, info: ValidationInfo):
...
+        title="Pin the quantizer amax to a constant value and skip calibration.",
+        description="""If set, the quantizer ``amax`` is fixed to this constant value and no
+        activation calibration is performed (no forward statistics are collected). The constant
+        is stored on the ``_amax`` buffer, so it is used by both the fake-quant forward pass and
+        the exported scaling factor (e.g. ``input_scale``).
+
+        This differs from ``use_constant_amax``, which pins amax to the FP8 E4M3 range (448.0)
+        for KV-cache cast math and does not register an ``_amax`` buffer. For NVFP4 activations
+        the exported ``input_scale`` equals ``amax / (E2M1_MAX * E4M3_MAX) = amax / (6 * 448)``;
+        setting ``constant_amax`` to ``2688.0`` therefore yields an exported ``input_scale`` of
```
chadvoegele, line 685: Can the old `use_constant_amax` behavior be replicated by setting `constant_amax = 448`? If so, could we unify these args?

chadvoegele (approved):

realAsma (approved):
Looks great!

modelopt_recipes/general/ptq/nvfp4_experts_only_input_scale1-kv_fp8_cast.yaml
```diff
@@ -0,0 +1,63 @@
...
+  description: >-
+    Applies NVFP4 only to expert-layer weight and input quantizers, pinning the expert activation
+    input_scale to 1.0 (constant amax = 2688, no activation calibration), plus FP8 KV-cache cast
+    mode using constant amax; uses max calibration for the remaining (weight) quantizers.
+quantize:
+  algorithm:
+    method: max
+    # Max calibration is fast and does not typically need checkpointing.
+    # layerwise=false required for VLMs where the decoder layers are nested under
+    # `model.language_model.layers` (layerwise_calibrate can't find them otherwise).
```
realAsma, line 42: BB:

Can you add support for layerwise calibration with VLM models? Create a new draft PR for this and notify me in slack.

realAsma: > _:robot: Bot comment._

🐝 Acknowledged. I’ll take this in a separate draft PR and notify you in Slack when it’s ready.

realAsma: > _:robot: Bot comment._

Done: opened draft PR #1953 for layerwise VLM image calibration support and notified you in Slack.

Validation:
- `pytest_pwd tests/examples/hf_ptq/test_hf_ptq_mono_quantize.py tests/examples/hf_ptq/test_hf_ptq_args.py`
- `pre-commit run --files examples/hf_ptq/hf_ptq.py tests/examples/hf_ptq/test_hf_ptq_mono_quantize.py`

Note: validation is focused/mock-based; I did not run a full real VLM calibration job.

modelopt/torch/quantization/config.py
Edwardf0t1: Not exactly today — they overlap in intent but differ in mechanism:

- `use_constant_amax=True` pins amax to 448 **only in the forward path** (`_get_amax` returns it on the fly) and deliberately registers **no** `_amax` buffer. It's used for the KV `*[kv]_bmm_quantizer`, whose scale is exported through the dedicated `kv_cache_scaling_factor` path, not the `_amax`→`input_scale` path. It's also format-agnostic (always 448).
- `constant_amax=448` would register `_amax=448` as a real buffer, so it lands in the state_dict **and** gets exported as an `input_scale` (= amax/maxbound). For a KV bmm quantizer that's an extra/unexpected buffer next to `kv_cache_scaling_factor`.

So `constant_amax=448` replicates the *fake-quant* behavior but not the *export/state* behavior. Unifying is feasible as a follow-up, but it touches the KV-cache export path (regression risk), so I'd want to first confirm the KV cast recipes still export byte-identical checkpoints before collapsing the two flags — I'd prefer to keep that out of this PR. Happy to file a follow-up issue to unify (likely: make `constant_amax` able to skip buffer registration / forward-only mode, then have the KV recipe set `constant_amax: 448`). WDYT?

realAsma:
@Edwardf0t1 

Do we still run full forward through the model even if all input/k/v quantizers use constant_amax? 
This seems suboptimal to me -  calibration will still be slow even if we use constant_amax. 

Can we check if all input/k/v quantizers are set to constant_amax? In that case can we override the forward_loop to do weight only calibration instead?

realAsma:
RB:

https://github.com/NVIDIA/Model-Optimizer/pull/1947#issuecomment-4930447597

Could you please provide a design plan for this?

realAsma:
> _:robot: Bot comment._

🐝 Acknowledged. I’m checking the PR context and will reply with a design plan for https://github.com/NVIDIA/Model-Optimizer/pull/1947#issuecomment-4930450573 shortly.

realAsma:
> _:robot: Bot comment._

Design plan for https://github.com/NVIDIA/Model-Optimizer/pull/1947#issuecomment-4930450573 / https://github.com/NVIDIA/Model-Optimizer/pull/1947#issuecomment-4930447597:

Yes, with the current code we still run the user `forward_loop` whenever it is provided: `max_calibrate()` first runs `weight_only_quantize(model)` to calibrate weights directly, then calls `forward_loop(model)` unconditionally if it is non-`None`. Since constant-amax quantizers are already disabled for stats collection in `enable_stats_collection()`, that forward pass is wasted when there are no enabled activation quantizers that need data-driven amax.

Proposed implementation:

1. Add a small helper in `model_calib.py`, e.g. `_needs_activation_forward_for_max_calib(model)`, that scans enabled `TensorQuantizer` leaves and returns `True` only if an activation/non-weight quantizer can still collect stats. It should ignore disabled quantizers, weight quantizers, dynamic quantizers, `_use_constant_amax`, and `_constant_amax is not None`.
2. In `max_calibrate()`, keep the existing `enable_stats_collection(model)` + `weight_only_quantize(model)` path unchanged, then replace `if forward_loop is not None: forward_loop(model)` with `if forward_loop is not None and _needs_activation_forward_for_max_calib(model): forward_loop(model)`.
3. Keep this optimization scoped to non-layerwise `method: max` first. Do not apply it to `local_hessian` because it needs activations for the Hessian, and be conservative with `mse` unless we explicitly decide weight-only MSE plus constant activations should skip the bootstrap forward too.
4. Add focused tests:
   - all enabled input/KV activation quantizers are constant-amax/use_constant_amax: supplied `forward_loop` is not called, weight quantizer still gets `_amax`;
   - at least one enabled non-constant input quantizer exists: `forward_loop` is called;
   - disabled/non-weight-only recipe entries do not force a forward;
   - existing constant-amax skip-calibration behavior still re-enables quantization after `finish_stats_collection()`.
5. For this PR’s recipe specifically, the expected behavior would be: expert weights still calibrate through `weight_only_quantize()`, expert inputs with `constant_amax: 2688.0` do not need activation stats, and KV cast quantizers using constant amax should not force a full model calibration forward. That gives the intended fast path without changing export scale semantics.

Edwardf0t1:
@realAsma Implemented this optimization in a stacked follow-up: #1963.

It adds `_needs_activation_forward_for_max_calib()` and gates the `forward_loop` in `max_calibrate()` behind `MaxCalibConfig.skip_forward_without_activation_calib` (default `True`). When every enabled activation quantizer is constant/dynamic, the forward is skipped and only weight calibration runs — weights are still calibrated on the weight tensors via `weight_only_quantize()`, so results are unchanged.

I scoped it opt-in at the `max_calibrate` level (default `False`) so the algorithms that call it directly and genuinely need activations (MSE, local Hessian, SmoothQuant, SVDQuant, GPTQ) are unaffected; only the top-level `max` path enables it. PTAL when you have a chance.

PR #1949, opened 2026-07-09
[OMNIML-5477, OMNIML-5119] Add module-specific AutoQuant search spaces

meenchen:
### What does this PR do?

Type of change: New feature.

This PR adds ordered, module-name-specific search spaces to AutoQuant so different runtime decision groups can use different candidate formats.

The user-facing design now separates fixed PTQ configuration from the AutoQuant search space:

- A normal top-level **quantize** config defines the fixed baseline for modules that are not searched.
- **auto_quantize.module_search_spaces** explicitly lists only the module families AutoQuant should search.
- Top-level **auto_quantize.candidate_formats** remains available for the existing global-search mode, but it is mutually exclusive with a fixed **quantize** baseline.
- Fixed and searched groups still participate in one integrated calibration, sensitivity-scoring, active-MoE cost, LP selection, checkpoint, and export flow.

Additional safeguards:

- Reject module rules that partially split a runtime-fused decision group.
- Keep BF16/no-quant as an internal sensitivity baseline while allowing each search-space rule to control whether it is solver-selectable.
- Isolate fixed groups from unrelated calibration algorithms.
- Reject infeasible budgets from the resolved per-group choices before calibration.
- Fingerprint the fixed PTQ baseline, runtime groups, candidate choices, scoring boundaries, replay attributes, and cost weights before checkpoint reuse.
- Preserve the previous global candidate-format API and one-candidate module rules for backward compatibility.

### Design rationale

The fixed configuration should use the same PTQ recipe system users already use for uniform NVFP4, W4A16, FP8, and other baseline configurations. AutoQuant then only needs to describe what is genuinely searched.

For example, the Qwen3.6 recipe reuses the model-specific quantization configuration from `huggingface/qwen3_5_moe/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast`. Unmatched modules retain that PTQ baseline, while shared experts, self-attention, linear-attention, and lm_head are explicitly searched. The original PTQ recipe is unchanged, and a loader test asserts that both recipes resolve to the same fixed `quantize` configuration.

This remains one AutoQuant operation rather than staged PTQ plus AutoQuant:

- NAS SearchSpace models generic architecture hyperparameters and is not connected to AutoQuant calibration snapshots, sensitivity scoring, runtime-fused groups, active-MoE accounting, or quant-config export.
- Ordered QuantizeConfig wildcard rules can describe a final static assignment, but cannot express per-family candidate sets in one global budget solve.
- Applying fixed PTQ before or after a separate AutoQuant pass would remove those modules from the shared sensitivity baseline and active-MoE numerator/denominator.

The implementation therefore composes the existing PTQ recipe schema with the existing AutoQuant/PuLP path instead of introducing another search framework or solver.

### Usage

~~~yaml
imports:
  model_quant_cfg: huggingface/qwen3_5/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast.quant_cfg
  w4a16_nvfp4: configs/ptq/presets/model/w4a16_nvfp4
  fp8: configs/ptq/presets/model/fp8

# Same model-specific baseline as the Qwen3.5-MoE PTQ recipe.
quantize:
  algorithm:
    method: max
    layerwise:
      enable: false
  quant_cfg:
    - $import: model_quant_cfg

auto_quantize:
  constraints:
    effective_bits: 6.0
    cost_model: active_moe
    cost:
      active_moe_expert_ratio: 0.03125

  # Only these modules are AutoQuant decision variables.
  module_search_spaces:
    - module_name_patterns:
        - "*mlp.shared_expert*"
        - "*linear_attn*"
        - "*self_attn*"
        - "*lm_head*"
      candidate_formats:
        - $import: w4a16_nvfp4
        - $import: fp8
      allow_no_quant: false
~~~

The legacy global-search form remains supported by omitting **quantize** and supplying top-level **auto_quantize.candidate_formats**.

### Qwen3.6 search-space comparison

Compared the default W4A16 NVFP4/FP8 search against the historical two-format module-specific recipe that keeps routed experts on the W4A16 PTQ baseline while shared experts and attention search W4A16 versus FP8. The final PR recipe keeps every matched quantizable group in W4A16 or FP8 (`allow_no_quant: false`). The table below is retained as evidence for the fixed-baseline design; the final recipe validation follows it.

Both lanes used local raw-gradient scoring, active-MoE accounting, batch size 1, code32k calibration, full parser-on LiveCodeBench with 8 repeats, and full SciCode. Active GiB/token includes scale storage.

| Target bits | Active GiB/token, default -> fixed W4 routed | Attention BF16+FP8, default -> fixed W4 routed | LCB, default -> fixed W4 routed | SciCode, default -> fixed W4 routed |
|---:|---:|---:|---:|---:|
| 5.8 | 2.1289 -> 2.0135 | 49.2% -> 80.0% | 70.8150 -> 72.4945 (+1.6795 pp) | 40.0518 -> 40.4216 (+0.3698 pp) |
| 6.1 | 2.2342 -> 2.1070 | 61.5% -> 93.8% | 72.5220 -> 73.5683 (+1.0463 pp) | 39.1642 -> 39.3861 (+0.2219 pp) |
| 6.4 | 2.3387 -> 2.2181 | 67.7% -> 79.2% | 72.0540 -> 73.9813 (+1.9273 pp) | 38.2027 -> 39.9038 (+1.7011 pp) |
| 6.7 | 2.4392 -> 2.3173 | 72.3% -> 93.8% | 72.5220 -> 73.4581 (+0.9361 pp) | 38.7944 -> 39.6820 (+0.8876 pp) |

BF16 references are 74.3667 LCB and 40.7914 SciCode. With a 1.5 percentage-point tolerance, the fixed-routed-expert lane jointly passes at targets 6.1, 6.4, and 6.7; the default lane has no joint pass. This comparison is end-to-end rather than an isolated solver ablation because the historical default lane used full attention-family grouping while the module-specific lane used runtime-required grouping.

Final no-BF16 recipe validation used the model-specific PTQ baseline, codeblend 16k calibration, local gradient scoring, active-MoE accounting, batch size 1, and full parser-on LiveCodeBench with 8 repeats:

| Target bits | Active GiB/token | Linear attention FP8/W4 | Self attention FP8/W4 | Shared experts FP8/W4 | lm_head | LCB pass@1 avg-of-8 | Avg completion tokens | Delta vs BF16 |
|---:|---:|---:|---:|---:|:---:|---:|---:|---:|
| 6.0 | 2.0823 | 80/10 | 36/4 | 97/23 | W4 | 73.2930 | 39,319.1 | -1.0737 pp |
| 6.3 | 2.1848 | 59/31 | 35/5 | 84/36 | FP8 | 73.8711 | 38,515.8 | -0.4956 pp |

Both targets pass the BF16-minus-1.5pp threshold. The model-specific baseline intentionally leaves linear-attention A/B and non-linear helper modules unquantized; all searched quantizable projections are W4A16 or FP8, and the artifact audit found no missing quant tensors.

### Testing

- Fresh four-GPU Qwen3.6 model-specific-baseline E2E runs completed calibration, gradient search, ModelOpt export, and runtime validation at 6.0 and 6.3 effective bits. Slurm jobs 2649968 and 2649969 completed with exit code 0. Both artifacts had complete shards, valid runtime metadata, and zero quantized modules missing quant tensors.
- Pre-commit passed: Ruff, formatting, mypy, Bandit, Markdown/YAML checks, and recipe validation.
- 306 focused AutoQuant, recipe-loader, and HF PTQ tests passed for the implementation.
- All 212 recipe-loader and HF PTQ mapping tests passed for the model-specific PTQ baseline integration. After the final `allow_no_quant: false` update, the focused loader/HF mapping tests and full pre-commit recipe validation passed. The original PTQ recipe remains unchanged, and the resolved fixed baselines compare equal.
- The distributed AutoQuant unit test passed separately outside the local socket sandbox.
- Wider quantization validation reached 782 passing tests; remaining local failures were unrelated optional dependency/socket limitations.
- Historical four-GPU Qwen3.6 E2E calibration, search, export, and metadata verification for the fixed-W4 baseline variant:
  - Slurm job 2645235 completed with exit code 0 in 9m53s.
  - Achieved 5.99 effective bits.
  - All 40 routed-expert groups remained fixed W4A16.
  - Exported valid searched allocations: linear attention FP8/W4 = 78/12, self-attention = 37/3, shared experts = 112/8, lm_head = W4.
  - Produced and verified a 21 GiB ModelOpt checkpoint.
- Completed the four-target full parser-on LCB/SciCode evaluation summarized above.
- Final no-BF16 parser-on LCB jobs 2650360 and 2650367 completed with exit code 0; their 8-repeat results are summarized above.

### Before this PR is ready for review

- If code was copied or a new dependency was added, was the contribution guidance followed?: N/A
- Were necessary tests added?: ✅
- Was the changelog updated?: ✅
- Claude approval after the latest update?: Pending

### Additional Information

Related work: OMNIML-5477, OMNIML-5119.


cjluo-nv (commented):

The implementation looks cohesive and well tested, and I did not find a concrete correctness regression in the module-specific candidate selection, fixed-group cost accounting, calibration isolation, fused-group validation, or checkpoint signature handling. The added YAML header follows the short SPDX style, but it does not match the complete canonical text in root `LICENSE_HEADER`, so this GitHub PR still needs human licensing sign-off under the review policy.

Design review: the problem is to let existing AutoQuantize runtime decision groups use family-specific candidate sets, including truly fixed quantized groups, without losing sensitivity baselines or active-MoE accounting. Plausible alternatives are (1) extending the repository's existing NAS `SearchSpace`/wildcard-rule machinery (`modelopt/torch/nas/search_space.py`), (2) expressing family overrides through the existing ordered `QuantizeConfig.quant_cfg`/last-match-wins mechanism plus pre-partitioned AutoQuant runs, or (3) keeping one global AutoQuant search and applying a fixed PTQ config to routed experts before/after it. Already-imported Pydantic/OmegaConf can model the declarative schema, while PuLP remains the existing solver; no dependency offers the AutoQuant-specific grouped scoring directly. The PR correctly extends the existing AutoQuant/Pydantic/PuLP path rather than creating a separate solver, and its Qwen3.6 evaluation provides strong motivation, but the PR body does not explicitly compare or reject the obvious in-repo NAS/wildcard and staged-PTQ alternatives. Because design review was required, please have the owner sign off or add that rationale to the PR body.

cjluo-nv (commented):

Re-review: the implementation remains cohesive and well tested, and the three actionable CodeRabbit findings are addressed: the critical checkpoint-integrity issue now validates both global and module-specific format signatures (including legacy behavior), while candidate-collection validation and the test-local import (minor) are fixed with focused tests. I found no new correctness regression.

Design review: the problem is to let AutoQuant runtime decision groups use module-family-specific candidate sets, including fixed quantized groups, while preserving sensitivity baselines and active-MoE cost accounting. Plausible alternatives are the existing NAS `SearchSpace` wildcard/rule machinery in `modelopt/torch/nas/search_space.py`, existing ordered `QuantizeConfig.quant_cfg` last-match-wins overrides plus staged AutoQuant/PTQ, or a global AutoQuant search followed by fixed PTQ for routed experts. The manifest already provides Pydantic/OmegaConf for schema/composition and PuLP for solving; this PR reasonably extends that existing AutoQuant/Pydantic/PuLP path rather than adding another solver. However, the PR body still gives evaluation motivation without explaining why the obvious in-repo NAS/wildcard or staged-PTQ alternatives were rejected, so the required architectural choice needs owner sign-off or rationale added to the PR body.

Licensing also requires human sign-off: the new YAML uses only the two-line SPDX header, which does not match the complete canonical root `LICENSE_HEADER`. Under the GitHub licensing policy this cannot be auto-approved.

shengliangxu (approved):
In general looking good to me.

meenchen:
/ok to test [058601e](https://github.com/NVIDIA/Model-Optimizer/pull/1949/commits/058601ea19615748999e6f8d083fba53eb4d5b5b)

juhi10071998:
hi @meenchen, thanks for this, will take a detailed look today/ tomo

modelopt/torch/quantization/algorithms.py
```diff
@@ -794,68 +914,127 @@ def before_search(self):
...
+        default_search_recipes = self._get_search_recipes(self.config["quantization_formats"])
+        quantization_formats_signature = _quantization_formats_signature(default_search_recipes)
+        module_search_space_signature = _module_search_space_signature(module_search_spaces)
+        restored_quantization_formats_signature = getattr(
+            self, "quantization_formats_signature", None
+        )
+        restored_module_search_space_signature = getattr(
+            self, "module_search_space_signature", None
+        )
+        has_restored_calibration_or_scores = bool(self.quantizer_states or self.candidate_stats)
```
juhi10071998, line 1158: Could we include the resolved search setup in the checkpoint validation? `candidate_stats` depends on `disabled_layers`, cost-exclusion patterns, and the runtime groups that matched each rule—not only the candidate formats.

For example, a checkpoint created with `*visual*` excluded from cost can be resumed without that exclusion, but the restored visual costs remain zero because scoring is skipped.

At minimum, we could validate normalized `disabled_layers` and the full cost config. Ideally, we should fingerprint the resolved groups, their cost weights, and allowed choices. Tests that change disabled/cost-excluded layers during resume would be useful.

modelopt/torch/quantization/algorithms.py
```diff
@@ -794,68 +914,127 @@ def before_search(self):
...
+            {
+                *default_search_recipes,
+                *(
+                    recipe
+                    for search_space in module_search_spaces
+                    for recipe in search_space["quant_recipes"]
+                ),
+            }
+        )
         self._verify_constraint(search_recipes)
```
juhi10071998, line 1209: Would checking the minimum format in the configured union be sufficient with per-module search spaces.

For example, every runtime group could be fixed to FP8 while an unused global fallback contains INT4. A 6-bit target passes this check even though the actual minimum is 8 bits. We then do all calibration/scoring before the LP reports failure, and the HF path currently ignores `is_satisfied`.

Should we compute the minimum achievable cost from each group’s actual solver-visible choices after the hparams are created, and fail early if the target is infeasible?

juhi10071998:
thanks Wei-Ming! overall the idea looks good to me. This is basically helping in two ways right if you can confirm

1. we only run calibration on the groups we know will be accounted for in the sensitivity analysis (will help speedup the calibration a bit (maybe) and the sensitivity scoring)

2. the quantization stats for a given layer and given candidate still assume the previous layer was either not quantized or quantized in the same precision right? even though the final mixed precision recipe could have the layer before the given layer in a different format.
       _- this is still better where we are assuming realistic choices for calibration (un-quantized in that format), rather than assuming a format we know will not hold in final state_

modelopt/torch/quantization/algorithms.py
meenchen: Addressed in 5627f9866. The checkpoint now records a fingerprint after runtime grouping is resolved. It includes each group’s module names, quantizer replay attributes, scoring-boundary modules, solver-visible recipe identities, allow_no_quant setting, and cost weight. Calibration or score state is rejected before reuse when that fingerprint is missing or differs. I also added calibration-only resume tests that independently change disabled_layers and excluded_module_name_patterns.

modelopt/torch/quantization/algorithms.py
meenchen: Addressed in 5627f9866. The global-union check remains as a cheap first check, but the authoritative feasibility check now runs after actual runtime hparams are created and before calibration. It sums the minimum scale-aware cost over each group’s solver-visible choices, including active-MoE/exclusion cost weights, and compares that with the weighted target budget. The all-fixed-FP8 plus unused-global-INT4 case now reports a minimum achievable 8.0 bits and verifies calibration was never called.

meenchen:
Thanks, @juhi10071998. Two nuances on the behavior:

1. Calibration is isolated by candidate format: on each format pass, only groups that expose that candidate use it; other groups are switched to their no-quant/BF16 calibration state. This prevents an unrelated algorithm (for example AWQ) from mutating a fixed FP8 group. It can reduce unnecessary quantizer work, although the full model forward still runs once per format in the union, so this PR does not claim a proportional end-to-end calibration speedup.

2. Sensitivity is still a local, one-decision-at-a-time estimate. Other configurable groups are at no-quant/BF16 while one group/candidate is perturbed; truly fixed one-format groups remain at their fixed format. It therefore does not model interactions among the eventual mixed-format assignments. Parent/group scoring or interaction-aware scoring would be a separate algorithmic extension.

I also added a Design rationale section to the PR body: NAS SearchSpace is not wired to AutoQuant calibration/sensitivity/cost/export, static quant_cfg wildcards cannot express candidate sets in one solve, and staged PTQ would move fixed groups outside the common sensitivity baseline and active-MoE budget.

meenchen:
/ok to test [5627f98](https://github.com/NVIDIA/Model-Optimizer/pull/1949/commits/5627f9866ccb544443da19078b1c5a5b7f882b2a)

modelopt_recipes/huggingface/qwen3_6_moe/auto_quantize/w4a16_nvfp4_fp8_module_spaces_at_6p0bits-active_moe.yaml
```diff
@@ -0,0 +1,66 @@
...
+  # remains an implicit choice only for this fallback search space.
+  candidate_formats:
+    - $import: w4a16_nvfp4
+    - $import: fp8
+
+  module_search_spaces:
+    # One selectable format makes routed experts fixed W4A16. allow_no_quant
+    # controls solver choices only; their BF16-equivalent denominator cost and
+    # active-MoE ratio remain part of effective-bit accounting.
+    - module_name_patterns:
```
cjluo-nv: does it mean we don't search and only apply w4a16?

modelopt_recipes/huggingface/qwen3_6_moe/auto_quantize/w4a16_nvfp4_fp8_module_spaces_at_6p0bits-active_moe.yaml
```diff
@@ -20,30 +18,20 @@ metadata:
     shared experts and attention searched over W4A16 NVFP4 and FP8, at 6.0
     active-MoE effective bits.
 
+# Reuse the normal PTQ preset for every module that is not explicitly searched.
+# This baseline remains active during sensitivity scoring and is calibrated,
+# costed, and exported as part of the same AutoQuantize operation.
+quantize:
+  $import: w4a16_nvfp4
+
 auto_quantize:
```
juhi10071998, line 32: this may be a little confusing, let's say if someone doesn't have the `module_search_spaces` and want to naively use the candidate formats on the full model (all modules searched for every format), then do we assert somewhere/ inform users that quantize ptq config should not be added in yaml? wondering what issues that may cause in the calibration if any

meenchen: Good point. These are now two explicit, mutually exclusive modes, and invalid combinations fail before calibration:

1. For a naive full-model search, omit top-level `quantize` and provide `auto_quantize.candidate_formats`.
2. For a fixed or model-specific PTQ baseline, provide top-level `quantize`, omit global `candidate_formats`, and list only searched families under `module_search_spaces`.

Both the recipe validator and direct `mtq.auto_quantize()` API reject `quantize`/`fixed_quantization_config` combined with global candidates, so an ambiguous mixed calibration cannot start.

As a concrete model-specific example, the Qwen3.6 AutoQuant recipe now reuses the same resolved `quantize` configuration as `huggingface/qwen3_5_moe/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast` while searching only shared experts, linear attention, self attention, and `lm_head`. The original PTQ recipe remains unchanged, and a test asserts that the two fixed baselines stay equal. I also documented both modes in the HF PTQ README.

juhi10071998: SGTM, thanks Wei-Ming.

cjluo-nv (approved):
The config design LGTM

modelopt_recipes/huggingface/qwen3_6_moe/auto_quantize/w4a16_nvfp4_fp8_module_spaces_at_6p0bits-active_moe.yaml
```diff
@@ -46,7 +47,7 @@ auto_quantize:
       candidate_formats:
         - $import: w4a16_nvfp4
         - $import: fp8
-      allow_no_quant: true
+      allow_no_quant: false
```
juhi10071998, line 50: LGTM, I am in general curious in what cases do we want to have the unquantized bf16 for certain modules. Router is sensitive to quantization so that could be one. Have we observed other bf16 in final recipe before? Asking so if we should have any recommendation for the user generally

juhi10071998 (approved):
Thanks Wei-Ming, left a minor comment.
So basically the cost accounting still remains same as before right for every module. We are restricting the search space for the modules and will only use the corresponding cost associated to that candidate for the given module.

modelopt_recipes/huggingface/qwen3_6_moe/auto_quantize/w4a16_nvfp4_fp8_module_spaces_at_6p0bits-active_moe.yaml
meenchen: That depends on the sensitivity of the module and also the candidate formats. In this case, since fp8 is able to maintain accuracy well, we don't need to explore bf16.

meenchen:
/ok to test [c2bdd81](https://github.com/NVIDIA/Model-Optimizer/pull/1949/commits/c2bdd81a1bbea15dac3b66a80831f1336e9f148c)

meenchen:
/ok to test [a60306b](https://github.com/NVIDIA/Model-Optimizer/pull/1949/commits/a60306ba9a649555c365a9a32371e149286a965b)

PR #1963, opened 2026-07-10
feat(quant): skip max-calib forward when no activation needs data

Edwardf0t1:
### What does this PR do?

Type of change: new feature (performance)

Addresses @realAsma's review comment on #1947: when all enabled input/KV activation quantizers use a constant amax, max calibration still runs the full `forward_loop`, which is wasted — the constant/dynamic quantizers collect no data-driven statistics.

This adds `_needs_activation_forward_for_max_calib(model)` and gates the `forward_loop` in `max_calibrate()`:

- Skips the forward when **no** enabled quantizer needs data-driven activation stats — i.e. every enabled non-weight quantizer is dynamic, `use_constant_amax`, or `constant_amax`. Weight quantizers are still calibrated on the weight tensors directly via `weight_only_quantize()`, so **results are unchanged**; only the wasted forward is avoided.
- Runs the forward whenever any enabled activation quantizer still needs data (normal PTQ → identical behavior).

**Safety / scoping.** The gate is behind an opt-in flag `max_calibrate(..., skip_forward_without_activation_calib=False)`. `max_calibrate` has many internal callers — the advanced algorithms that always need activations (**MSE, local Hessian, SmoothQuant, SVDQuant, GPTQ**) call it directly and keep the default `False`, so they are unaffected. Only the top-level `max` path opts in, via `MaxCalibConfig.skip_forward_without_activation_calib` (default `True`).

Also simplifies a default-arg lambda in `core_utils.sync_moe_expert_amax` (`lambda m, w=weight: m(w)` → `lambda m: m(weight)`) so mypy can infer its type — surfaced by the `max_calibrate` signature change; the `forward_loop` is invoked synchronously so closure capture is safe.

### Usage

Automatic for the top-level `max` path — no user action needed. For the `nvfp4_experts_only_input_scale1-kv_fp8_cast` recipe (all expert inputs + KV pinned to constant amax), calibration now performs weight-only calibration and skips the model forward:

```
max_calibrate: all enabled activation quantizers use constant/dynamic amax; skipping the calibration forward pass (weight-only calibration).
```

To force the old behavior: set `skip_forward_without_activation_calib: false` on the max algorithm config.

### Testing

New CPU unit tests in `tests/unit/torch/quantization/test_calib.py`:
- `test_max_calib_skips_forward_when_all_activations_constant` — forward not called; weight quantizer still gets `_amax`; constant input amax preserved.
- `test_max_calib_runs_forward_when_activation_needs_data` — a normal input quantizer still triggers the forward.
- `test_max_calib_default_does_not_skip_forward` — direct callers (default flag) keep running the forward.
- `test_needs_activation_forward_ignores_disabled_and_weight_quantizers`.
- `test_max_calib_config_enables_skip_by_default`.

`pytest tests/unit/torch/quantization/test_calib.py` → 17 passed; `test_quantize_cpu.py` + `test_tensor_quantizer_cpu.py` → 79 passed. `pre-commit` on all changed files passes (ruff, mypy, bandit, …).

- Is this change backward compatible?: ✅ (opt-in flag; default preserves behavior for direct callers; normal max PTQ unchanged)

### Additional Information

Follow-up to #1947 (now merged) per reviewer request. Closes the optimization discussion at https://github.com/NVIDIA/Model-Optimizer/pull/1947#issuecomment-4930447597.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in configuration flag to let max calibration skip the activation-calibration forward pass when no enabled activation quantizer requires runtime statistics.
  * Weight calibration still runs when the forward pass is skipped.
  * Advanced calibration flows that require activation data remain unaffected.
* **Documentation**
  * Updated the changelog and enabled the option in the NVFP4 expert-only PTQ recipe.
* **Bug Fixes**
  * Improved expert amax synchronization to ensure the correct weight is used during calibration.
* **Tests**
  * Added unit tests covering the skip behavior and activation-forward requirement detection.


modelopt/torch/quantization/model_calib.py
```diff
@@ -254,13 +254,46 @@ def _should_sync_amax_across_ep(
...
+        # are calibrated on the weight tensor directly, not via the data forward.
+        if any(part.endswith("weight_quantizer") for part in name.split(".")):
+            continue
+        if (
+            module._dynamic
+            or module._use_constant_amax
+            or getattr(module, "_constant_amax", None) is not None
+        ):
+            continue
+        if getattr(module, "_calibrator", None) is not None:
```
realAsma, line 297: Can we also check if the model is mx formats like MXFP4? if the activation quantizer is MXFP4/MXFP8 we can skip calibration on that as well.

realAsma: RB: Can you give us a plan on how to do this?

realAsma: RB: Never mind, I think `module._dynamic` captures that. Is that correct?

realAsma (approved):
Looks great!! Thanks for this PR

modelopt/torch/quantization/model_calib.py
Edwardf0t1: Good question — actually `module._dynamic` does **not** capture MXFP4/MXFP8. `_dynamic` is only set from a *top-level* `type: dynamic`, but the MX numerics configs set `type: dynamic` **inside `block_sizes`** (block-level E8M0 scales), leaving the top-level `_dynamic == False`:

```yaml
# configs/numerics/mxfp8.yaml
num_bits: e4m3
block_sizes: {-1: 32, type: dynamic, scale_bits: e8m0}
```

So an MX activation quantizer would currently fall through to the `_calibrator is not None` check and be treated as needing the forward. MX formats carry no per-tensor amax at all (`is_mx_format` makes `amax` always `None`), so they're safe to skip. I'll add `module.is_mx_format` to the skip conditions in `_needs_activation_forward_for_max_calib` so MXFP4/MXFP8 are covered too. Thanks for catching this!

Edwardf0t1:
Pushed `33cebce` with two changes:

1. **MX coverage** (per @realAsma): added `module.is_mx_format` to `_needs_activation_forward_for_max_calib`, so MXFP4/MXFP8 activation quantizers are also skipped. As noted above, `_dynamic` alone didn't catch them (MX sets `type: dynamic` only inside `block_sizes`).

2. **Made the optimization opt-in** (`MaxCalibConfig.skip_forward_without_activation_calib` default `True` → `False`). The prior default-on broke a GPU test — `test_nested_deepspeed_backward[3-quant_cfg0]` (weight-only, **ZeRO-3**): the calibration forward materializes ZeRO-3 sharded params, so skipping it left the shards out of sync with the quantized weights (ZeRO-1/2 passed; only ZeRO-3 failed). Making it opt-in keeps all existing max-PTQ behavior unchanged; the `nvfp4_experts_only_input_scale1-kv_fp8_cast` recipe now sets the flag explicitly to get the optimization.

Tests updated (opt-in default + MX/dynamic coverage), changelog updated, pre-commit + recipe validation green. CI re-running.

modelopt/torch/quantization/model_calib.py
```diff
@@ -254,13 +254,52 @@ def _should_sync_amax_across_ep(
...
 
 
+def _needs_activation_forward_for_max_calib(model: nn.Module) -> bool:
+    """Return True if any enabled quantizer still needs a calibration forward pass.
+
+    Weight quantizers are calibrated directly on the weight tensors by
+    :func:`weight_only_quantize`, so they never need the data forward. Activation-side
+    quantizers (input/output/BMM) need it only when they collect data-driven statistics:
+    i.e. they are enabled, have a calibrator, and are none of
+
```
juhi10071998: should we also check for the bias_calibrator?

modelopt/torch/quantization/model_calib.py
realAsma: > _:robot: Bot comment._

🐝 Looking into this.

realAsma: > _:robot: Bot comment._

Edward’s answer is correct. Concretely, the plan is:

1. Keep `_dynamic` for top-level dynamic quantizers.
2. Add `module.is_mx_format` as a separate skip condition, because MX’s `type: dynamic` lives inside `block_sizes` and leaves `_dynamic == False`.
3. Add a regression test proving an MX quantizer has `is_mx_format == True`, `_dynamic == False`, and makes `_needs_activation_forward_for_max_calib()` return `False`; retain the normal data-driven case returning `True`.

This is already implemented at `3f9d4e28`, and the required unit/GPU/regression checks are green.

modelopt/torch/quantization/model_calib.py
Edwardf0t1: Good catch — yes, we should. A quantizer with a **static** `bias_calibrator` collects data during the forward (`finish_stats_collection` → `load_calib_bias`), so it needs the forward even when its amax path is dynamic/MX (which carry no per-tensor amax). My original helper would have wrongly skipped that case.

Fixed in `dec014d`: `_needs_activation_forward_for_max_calib` now mirrors `finish_stats_collection` exactly — a non-constant quantizer with a static bias forces the forward:

```python
if not is_constant and module.bias_calibrator is not None and module.bias_type == "static":
    return True
```

Constant-amax quantizers stay exempt (they `continue` before the bias block in `finish_stats_collection`, so their bias isn't calibrated either), and a *dynamic* bias is computed on the fly so still needs no forward. Added tests for MX+static-bias (→ forward runs), MX+dynamic-bias, and constant+static-bias. Thanks!

PR #1975, opened 2026-07-14
add Qwen3-VL support for DFlash training

skierat:
### What does this PR do?         

  Type of change: new feature

  Adds online DFlash training support for Qwen3-VL–style vision-language models.
                                                             
  Changes include:

  - Load VLMs through the Transformers 5 `AutoModelForImageTextToText` API, while retaining compatibility with the legacy VLM auto-model API.
  - Run the base model through its top-level multimodal forward when image/video inputs are present, ensuring vision embeddings are injected before collecting DFlash target hidden states.
  - Extend `VisionLanguageDataCollator` to:
    - propagate `answer_only_loss`, chat-template, and DFlash label-alignment settings;
    - apply `VLM_MIN_PIXELS` / `VLM_MAX_PIXELS` processor limits;
    - derive assistant-only masks from ChatML/Llama chat boundaries when processor generation masks are unavailable;                                                                                                                                
    - enforce the fixed `training_seq_len` required by DFlash block training.
  - Preserve the existing text-only DFlash path.                                                                                                                                                                                                    
                                                             
### Usage

```bash                                                                
python -m torch.distributed.run \                                      
--nproc_per_node 4 \                                                 
examples/speculative_decoding/main.py \                              
--config modelopt_recipes/general/speculative_decoding/dflash.yaml \
model.model_name_or_path=/path/to/qwen3-vl-model \
model.trust_remote_code=true \
data.data_path=/path/to/train.jsonl \
data.vlm_processor=/path/to/qwen3-vl-model \                        
data.vlm_img_dir=/path/to/image/root \
training.training_seq_len=4096 \
training.answer_only_loss=true \
dflash.dflash_block_size=8 \
dflash.dflash_mask_token_id=151669


### Testing
- git diff --check
- Parsed all modified Python modules successfully.
- Ran iterative multi-node Slurm smoke tests with a Qwen3-VL-family model and mixed multimodal data:
    - validated VLM model loading with Transformers 5;
    - validated distributed initialization, DFlash conversion, and VLM collation paths;
    - identified and addressed processor padding/truncation behavior required by fixed-size DFlash blocks.
This PR remains draft pending a completed end-to-end training smoke test and automated regression coverage.

Make sure you read and follow Contributor guidelines (https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md) and your commits are signed (git commit -s -S).
Make sure you read and follow the Security Best Practices (https://github.com/NVIDIA/Model-Optimizer/blob/main/SECURITY.md#security-coding-practices-for-contributors) (e.g. avoiding hardcoded trust_remote_code=True, torch.load(...,
weights_only=False), pickle, etc.).
- Did you write any new necessary tests?: ❌ — automated Qwen3-VL/DFlash regression coverage still needs to be added before review.
- Did you update Changelog (https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ❌ — evaluate and add an entry before marking ready for review if this is considered user-facing speculative-decoding support.
### Additional Information
The PR intentionally excludes local Slurm launch scripts, logs, model paths, datasets, and environment-specific configuration.


## Summary by CodeRabbit

* **New Features**
  * Expanded VLM data-collation controls, including `shift_labels` and more robust `answer_only_loss` masking.
  * Improved Qwen3-VL speculative decoding for Transformers 5.3+ with correct video frame grouping and safer position-id handling.
  * Improved DFlash RoPE export to reliably read `rope_theta` from newer config formats.

* **Bug Fixes**
  * Hardened multimodal preprocessing and training loss masking to keep label/attention alignment consistent.
  * Improved behavior when anchor sampling yields no valid blocks.
  * More resilient VLM model loading when certain Transformers auto classes are unavailable.

* **Tests**
  * Added coverage for RoPE export, Qwen3-VL position-id logic across Transformers versions, and VLM label-mode/collator options.

### DIFF
--- a/examples/speculative_decoding/eagle_utils.py
+++ b/examples/speculative_decoding/eagle_utils.py
@@ -141,6 +141,9 @@ def make_speculative_data_module(
                 train_len=train_len,
                 local_image_path=data_args.vlm_img_dir,
                 return_labels=True,
+                answer_only_loss=answer_only_loss,
+                shift_labels=shift_labels,
+                chat_template=chat_template,
             )
 
     else:
--- a/modelopt/torch/speculative/plugins/hf_dflash.py
+++ b/modelopt/torch/speculative/plugins/hf_dflash.py
@@ -74,6 +74,7 @@ Draft model components:
 import logging
 
 import torch
+import transformers
 import torch.nn.functional as F
 from transformers import PreTrainedModel
 from transformers.models.qwen3.configuration_qwen3 import Qwen3Config as _Qwen3Config
@@ -98,6 +99,30 @@ from .modeling_fakebase import (
 logger = logging.getLogger(__name__)
 
 __all__ = ["HFDFlashModel"]
+
+
+def _expand_qwen3_video_grid_thw(video_grid_thw: torch.Tensor) -> torch.Tensor:
+    """Return the per-frame video grid representation used by Qwen3-VL RoPE.
+
+    Qwen3-VL's video processor emits one ``[T, H, W]`` row per source video, but
+    its rendered prompt contains a separate visual-token group for every temporal
+    frame.  Transformers 5.3's ``get_rope_index`` consumes one grid row per
+    rendered group, while the vision encoder still requires the original one-row-
+    per-video representation.  This helper is therefore used *only* for mRoPE
+    position construction; callers must keep the original tensor for the model
+    forward.
+    """
+    if video_grid_thw.ndim != 2 or video_grid_thw.shape[-1] != 3:
+        raise ValueError(
+            "Qwen3-VL video_grid_thw must have shape [num_videos, 3], got "
+            f"{tuple(video_grid_thw.shape)}."
+        )
+    if torch.any(video_grid_thw[:, 0] <= 0):
+        raise ValueError("Qwen3-VL video_grid_thw temporal lengths must be positive.")
+
+    expanded_grid_thw = torch.repeat_interleave(video_grid_thw, video_grid_thw[:, 0], dim=0)
+    expanded_grid_thw[:, 0] = 1
+    return expanded_grid_thw
 
 
 def _dpace_position_weights(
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: modelopt/torch/speculative/plugins/hf_dflash.py [hunk 1]
```diff
@@ -183,6 +208,98 @@ def _base_llm_config(self):
...
+        Prefer ``get_rope_index`` over ``compute_3d_position_ids``.  The latter
+        writes ``rope_deltas`` into the base model even though DFlash training
+        never supplies a cache.  Keeping this calculation side-effect free is
+        important when the frozen target is reused for consecutive training
+        batches or validation.
+        """
+        if (
+            position_ids is not None
+            or getattr(self.config, "model_type", None) != "qwen3_vl"
+            or not transformers.__version__.startswith("5.3.")
```
h-guo18: question: is this patch only needed for transformers 5.3, or for all transformers 5.x or >5.3?

examples/speculative_decoding/eagle_utils.py
```diff
@@ -141,6 +141,9 @@ def make_speculative_data_module(
                 train_len=train_len,
                 local_image_path=data_args.vlm_img_dir,
                 return_labels=True,
+                answer_only_loss=answer_only_loss,
+                shift_labels=shift_labels,
```
h-guo18, line 145: **[AI review]** The PR description says `VisionLanguageDataCollator` was extended (propagating `answer_only_loss`/chat-template/label-alignment settings, `VLM_MIN_PIXELS`/`VLM_MAX_PIXELS` limits, ChatML-boundary assistant masks, fixed `training_seq_len` enforcement), but `modelopt/torch/utils/plugins/transformers_dataset.py` is not part of this diff — on the current head the class still has its old signature. It looks like that file may not have been committed/pushed.

Note this is broader than the `shift_labels` TypeError already flagged: even after adding that one parameter, the described mask/pixel/seq-len behaviors would still be missing from the PR.

modelopt/torch/speculative/plugins/hf_dflash.py
h-guo18: **[AI review]** Two robustness concerns with this gate:

1. **Version pin.** Tracing transformers 5.3.0, the per-video vs per-frame grid mismatch looks like a 5.3 bug rather than a stable contract: `get_rope_index` consumes one grid row per rendered visual group (`next(grid_iters[...])`), while the processor renders T timestamp-separated frame groups per video and the video processor emits one `[T, H, W]` row per video. Even the vanilla generation path (`_prepare_position_ids_for_generation`) passes the raw grid, so upstream video inference appears equally broken on 5.3 — meaning upstream will likely fix this inside `get_rope_index` (as in 4.x, where the expansion was internal). If that fix lands in a 5.3.x patch release, this external expansion becomes a double expansion and silently corrupts positions; on 5.4+ the helper silently disables itself and multimodal batches fall back to the broken internal path. Suggest linking the upstream issue/commit here, and failing loudly (frame-group count ≠ grid-row count) instead of silently returning `None` when the gate doesn't match.

2. **`model_type != "qwen3_vl"` exact match.** Family variants like `qwen3_vl_moe` skip the precomputation but still enter the multimodal top-level forward below — landing exactly in the broken path this helper works around. Consider covering the family or at least warning.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -592,6 +709,15 @@ def forward(
         - Label alignment: position k predicts token at anchor+k
         - Optional loss decay weighting
         """
+        position_ids = self._qwen3_vl_position_ids(
```
h-guo18: **[AI review]** This runs before the `if not self.training` branch, so it also fires on eval prefill (multimodal inputs, no cache). Passing explicit `position_ids` makes `Qwen3VLModel.forward` skip `compute_3d_position_ids`, so `self.rope_deltas` is never established; a subsequent cached decode step (`position_ids=None`, `past_key_values` not None → this helper returns `None`) then falls into the `rope_deltas`-based branch with a stale/`None` delta and computes wrong positions.

Since the docstring says this construction is only needed for DFlash training, consider gating on `self.training`.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -638,12 +764,49 @@ def forward(
...
+                "videos",
+            }
+            use_top_level_forward = any(kwargs.get(key) is not None for key in multimodal_keys)
             with torch.no_grad():
-                raw_outputs = super().forward(
-                    input_ids=input_ids,
-                    attention_mask=attention_mask,
-                    output_hidden_states=True,
+                if use_top_level_forward:
+                    base_forward_kwargs = dict(kwargs)
```
h-guo18: **[AI review]** `base_forward_kwargs` forwards everything except `assistant_masks`/`loss_mask`. Anything else that reaches `forward`'s `**kwargs` — e.g. Trainer-injected `num_items_in_batch` (Trainer detects that this forward accepts `**kwargs` and adds loss kwargs), or stray dataset columns — gets passed to the HF multimodal forward and can raise `TypeError` or be silently swallowed depending on version. The text-only branch below is immune because it forwards a fixed argument set.

Consider an allowlist (the multimodal keys + known model kwargs) or filtering against the base forward's signature.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -74,6 +74,7 @@
 import logging
 
 import torch
+import transformers
```
h-guo18: **[AI review]** nit: `import transformers` lands between `import torch` and `import torch.nn.functional as F`; ruff/isort (I001) will flag this in pre-commit/CI.

h-guo18 (commented):
**[AI review]** Supplementing the earlier claude[bot] review (which flagged the `shift_labels` TypeError): a few remaining issues, details inline.

Also noting a test-coverage gap for the pending regression work: nothing currently exercises the multimodal top-level-forward branch, the loss-mask ∩ attention_mask change, or the all-parameter DDP dummy-loss path.

modelopt/torch/speculative/plugins/hf_dflash.py
skierat: Only for Transformers 5.3.x.
Starting with 5.4.0, upstream added that same repeat_interleave(...); grid[:, 0] = 1 expansion internally in get_rope_index(), and it remains in 5.12.

h-guo18 (approved):
lgtm

ChenhanYu (approved):

PR #1980, opened 2026-07-16
Scripts and a skill to do per-layer benchmark using flashinfer

sychen52:
add scripts and a skill to use flashinfer to do layerwise benchmark with different backends.

### What does this PR do?

Type of change: ?  new skill and scripts

### Usage

```bash
python .agents/skills/benchmark-model-kernels/scripts/benchmark_model.py \
    openai/gpt-oss-120b \
    --tp 4 --ep 4 --ms 8 16 \
    --flashinfer_repo $HOME/flashinfer \
    --workdir /tmp/gpt-oss-benchmark
```
or
```bash
python .agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py \
    --flashinfer_repo $HOME/flashinfer \
    --ms 8 16 \
    --nks 1280,2880 2880,1024 \
    --moe_hidden_size 2880 \
    --moe_intermediate_size 2880 \
    --moe_num_experts 32 \
    --moe_top_k 4 \
    --workdir /tmp/gpt-oss-via-builtin
```

Or example prompt to trigger the skill:
```
Benchmark the kernel shapes for NVIDIA-Nemotron-3-Nano-30B-A3B
```

example output (TP1, EP4):
```
flashinfer 0.6.13; checkout ~/flashinfer3 @ 453bc6e1c6c852a4dc8a35c672f271e1e0d216a5; NVIDIA Graphics Device (sm_100 / 148 SMs / 178 GiB); 600 W power limit
GEMM
module_name,M,N,K,backend,with_quant,runtime
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,bf16,False,8.384
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,bf16,False,8.896
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,fp8_cudnn,False,6.112
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,fp8_cudnn,True,8.832
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,fp8_cudnn,False,6.176
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,fp8_cudnn,True,8.944
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,fp8_cutlass,False,7.280
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,fp8_cutlass,True,10.000
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,fp8_cutlass,False,7.216
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,fp8_cutlass,True,9.984
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,fp8_trtllm,False,7.583
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,fp8_trtllm,True,10.303
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,fp8_trtllm,False,7.520
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,fp8_trtllm,True,10.288
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_cudnn,False,7.456
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_cudnn,True,10.016
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_cudnn,False,7.472
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_cudnn,True,10.192
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_cutedsl,False,6.928
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_cutedsl,True,9.488
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_cutedsl,False,6.160
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_cutedsl,True,8.880
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_cutlass,False,7.232
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_cutlass,True,9.792
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_cutlass,False,8.320
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_cutlass,True,11.040
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_trtllm,False,7.024
model.layers.*.mixer.shared_experts.up_proj,1,3712,2688,nvfp4_trtllm,True,9.488
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_trtllm,False,6.976
model.layers.*.mixer.shared_experts.up_proj,16,3712,2688,nvfp4_trtllm,True,9.584
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,bf16,False,9.504
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,bf16,False,9.791
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,fp8_cudnn,False,6.560
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,fp8_cudnn,True,9.344
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,fp8_cudnn,False,6.688
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,fp8_cudnn,True,9.488
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,fp8_cutlass,False,7.712
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,fp8_cutlass,True,10.496
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,fp8_cutlass,False,7.680
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,fp8_cutlass,True,10.480
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,fp8_trtllm,False,8.560
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,fp8_trtllm,True,11.344
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,fp8_trtllm,False,8.623
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,fp8_trtllm,True,11.424
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_cudnn,False,8.464
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_cudnn,True,11.216
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_cudnn,False,8.512
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_cudnn,True,11.360
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_cutedsl,False,7.584
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_cutedsl,True,10.336
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_cutedsl,False,6.416
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_cutedsl,True,9.264
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_cutlass,False,8.992
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_cutlass,True,11.744
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_cutlass,False,8.800
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_cutlass,True,11.648
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_trtllm,False,7.648
model.layers.*.mixer.shared_experts.down_proj,1,2688,3712,nvfp4_trtllm,True,10.208
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_trtllm,False,7.104
model.layers.*.mixer.shared_experts.down_proj,16,2688,3712,nvfp4_trtllm,True,9.887
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,bf16,False,8.704
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,bf16,False,8.736
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,fp8_cudnn,False,6.512
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,fp8_cudnn,True,9.232
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,fp8_cudnn,False,6.464
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,fp8_cudnn,True,9.232
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,fp8_cutlass,False,8.016
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,fp8_cutlass,True,10.736
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,fp8_cutlass,False,7.648
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,fp8_cutlass,True,10.416
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,fp8_trtllm,False,7.776
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,fp8_trtllm,True,10.496
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,fp8_trtllm,False,7.904
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,fp8_trtllm,True,10.672
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_cudnn,False,7.808
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_cudnn,True,10.367
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_cudnn,False,7.920
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_cudnn,True,10.640
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_cutedsl,False,7.104
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_cutedsl,True,9.664
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_cutedsl,False,6.304
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_cutedsl,True,9.024
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_cutlass,False,8.736
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_cutlass,True,11.296
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_cutlass,False,7.583
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_cutlass,True,10.303
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_trtllm,False,6.784
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,1,4608,2688,nvfp4_trtllm,True,9.248
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_trtllm,False,7.040
model.layers.*.mixer.q_proj|model.layers.*.mixer.k_proj|model.layers.*.mixer.v_proj,16,4608,2688,nvfp4_trtllm,True,9.647
model.layers.*.mixer.o_proj,1,2688,4096,bf16,False,10.368
model.layers.*.mixer.o_proj,16,2688,4096,bf16,False,10.880
model.layers.*.mixer.o_proj,1,2688,4096,fp8_cudnn,False,6.912
model.layers.*.mixer.o_proj,1,2688,4096,fp8_cudnn,True,9.663
model.layers.*.mixer.o_proj,16,2688,4096,fp8_cudnn,False,7.008
model.layers.*.mixer.o_proj,16,2688,4096,fp8_cudnn,True,9.888
model.layers.*.mixer.o_proj,1,2688,4096,fp8_cutlass,False,8.160
model.layers.*.mixer.o_proj,1,2688,4096,fp8_cutlass,True,10.912
model.layers.*.mixer.o_proj,16,2688,4096,fp8_cutlass,False,8.064
model.layers.*.mixer.o_proj,16,2688,4096,fp8_cutlass,True,10.944
model.layers.*.mixer.o_proj,1,2688,4096,fp8_trtllm,False,8.480
model.layers.*.mixer.o_proj,1,2688,4096,fp8_trtllm,True,11.232
model.layers.*.mixer.o_proj,16,2688,4096,fp8_trtllm,False,8.752
model.layers.*.mixer.o_proj,16,2688,4096,fp8_trtllm,True,11.632
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_cudnn,False,8.768
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_cudnn,True,11.520
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_cudnn,False,8.864
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_cudnn,True,11.680
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_cutedsl,False,7.904
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_cutedsl,True,10.656
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_cutedsl,False,6.784
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_cutedsl,True,9.600
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_cutlass,False,9.248
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_cutlass,True,12.000
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_cutlass,False,9.264
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_cutlass,True,12.080
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_trtllm,False,7.184
model.layers.*.mixer.o_proj,1,2688,4096,nvfp4_trtllm,True,9.744
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_trtllm,False,7.744
model.layers.*.mixer.o_proj,16,2688,4096,nvfp4_trtllm,True,10.432
model.layers.*.mixer.out_proj,1,2688,4096,bf16,False,10.368
model.layers.*.mixer.out_proj,16,2688,4096,bf16,False,10.880
model.layers.*.mixer.out_proj,1,2688,4096,fp8_cudnn,False,6.912
model.layers.*.mixer.out_proj,1,2688,4096,fp8_cudnn,True,9.663
model.layers.*.mixer.out_proj,16,2688,4096,fp8_cudnn,False,7.008
model.layers.*.mixer.out_proj,16,2688,4096,fp8_cudnn,True,9.888
model.layers.*.mixer.out_proj,1,2688,4096,fp8_cutlass,False,8.160
model.layers.*.mixer.out_proj,1,2688,4096,fp8_cutlass,True,10.912
model.layers.*.mixer.out_proj,16,2688,4096,fp8_cutlass,False,8.064
model.layers.*.mixer.out_proj,16,2688,4096,fp8_cutlass,True,10.944
model.layers.*.mixer.out_proj,1,2688,4096,fp8_trtllm,False,8.480
model.layers.*.mixer.out_proj,1,2688,4096,fp8_trtllm,True,11.232
model.layers.*.mixer.out_proj,16,2688,4096,fp8_trtllm,False,8.752
model.layers.*.mixer.out_proj,16,2688,4096,fp8_trtllm,True,11.632
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_cudnn,False,8.768
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_cudnn,True,11.520
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_cudnn,False,8.864
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_cudnn,True,11.680
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_cutedsl,False,7.904
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_cutedsl,True,10.656
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_cutedsl,False,6.784
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_cutedsl,True,9.600
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_cutlass,False,9.248
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_cutlass,True,12.000
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_cutlass,False,9.264
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_cutlass,True,12.080
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_trtllm,False,7.184
model.layers.*.mixer.out_proj,1,2688,4096,nvfp4_trtllm,True,9.744
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_trtllm,False,7.744
model.layers.*.mixer.out_proj,16,2688,4096,nvfp4_trtllm,True,10.432
model.layers.*.mixer.in_proj,1,10304,2688,bf16,False,17.520
model.layers.*.mixer.in_proj,16,10304,2688,bf16,False,16.032
model.layers.*.mixer.in_proj,1,10304,2688,fp8_cudnn,False,8.912
model.layers.*.mixer.in_proj,1,10304,2688,fp8_cudnn,True,11.632
model.layers.*.mixer.in_proj,16,10304,2688,fp8_cudnn,False,9.375
model.layers.*.mixer.in_proj,16,10304,2688,fp8_cudnn,True,12.143
model.layers.*.mixer.in_proj,1,10304,2688,fp8_cutlass,False,10.527
model.layers.*.mixer.in_proj,1,10304,2688,fp8_cutlass,True,13.248
model.layers.*.mixer.in_proj,16,10304,2688,fp8_cutlass,False,10.415
model.layers.*.mixer.in_proj,16,10304,2688,fp8_cutlass,True,13.183
model.layers.*.mixer.in_proj,1,10304,2688,fp8_trtllm,False,9.328
model.layers.*.mixer.in_proj,1,10304,2688,fp8_trtllm,True,12.048
model.layers.*.mixer.in_proj,16,10304,2688,fp8_trtllm,False,9.392
model.layers.*.mixer.in_proj,16,10304,2688,fp8_trtllm,True,12.160
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_cudnn,False,9.088
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_cudnn,True,11.648
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_cudnn,False,9.024
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_cudnn,True,11.744
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_cutedsl,False,9.424
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_cutedsl,True,11.984
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_cutedsl,False,7.616
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_cutedsl,True,10.336
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_cutlass,False,9.728
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_cutlass,True,12.288
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_cutlass,False,10.144
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_cutlass,True,12.864
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_trtllm,False,ERROR: FlashInfer produced no result row and no error message; see scratchpad/nemotron_ep4_v6/driver.log
model.layers.*.mixer.in_proj,1,10304,2688,nvfp4_trtllm,True,ERROR: FlashInfer produced no result row and no error message; see scratchpad/nemotron_ep4_v6/driver.log
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_trtllm,False,ERROR: FlashInfer produced no result row and no error message; see scratchpad/nemotron_ep4_v6/driver.log
model.layers.*.mixer.in_proj,16,10304,2688,nvfp4_trtllm,True,ERROR: FlashInfer produced no result row and no error message; see scratchpad/nemotron_ep4_v6/driver.log

MoE
H=2688 F=1856 E=32 top_k=6 activation=Relu2
module_name,M,N,K,backend,with_quant,runtime
model.layers.*.mixer.experts,1,,,bf16_cutlass,False,54.271
model.layers.*.mixer.experts,16,,,bf16_cutlass,False,175.552
model.layers.*.mixer.experts,1,,,fp8_cutlass,False,41.312
model.layers.*.mixer.experts,1,,,fp8_cutlass,True,44.032
model.layers.*.mixer.experts,16,,,fp8_cutlass,False,94.799
model.layers.*.mixer.experts,16,,,fp8_cutlass,True,97.567
model.layers.*.mixer.experts,1,,,fp8_trtllm,False,26.480
model.layers.*.mixer.experts,1,,,fp8_trtllm,True,29.200
model.layers.*.mixer.experts,16,,,fp8_trtllm,False,97.680
model.layers.*.mixer.experts,16,,,fp8_trtllm,True,100.448
model.layers.*.mixer.experts,1,,,nvfp4_cutlass,False,40.944
model.layers.*.mixer.experts,1,,,nvfp4_cutlass,True,40.623
model.layers.*.mixer.experts,16,,,nvfp4_cutlass,False,70.432
model.layers.*.mixer.experts,16,,,nvfp4_cutlass,True,71.072
model.layers.*.mixer.experts,1,,,nvfp4_trtllm,False,22.784
model.layers.*.mixer.experts,1,,,nvfp4_trtllm,True,25.119
model.layers.*.mixer.experts,16,,,nvfp4_trtllm,False,57.407
model.layers.*.mixer.experts,16,,,nvfp4_trtllm,True,59.999

```

### Testing
run locally

## Summary by CodeRabbit

- **New Features**
  - Added a guided workflow for planning and running per-rank GEMM and fused-MoE kernel benchmarks.
  - Added automatic model inspection to derive supported kernel shapes, tensor-parallel and expert-parallel settings, routing details, and required benchmark parameters.
  - Added FlashInfer built-in benchmark execution with BF16, FP8, NVFP4, and activation-quantization timing.
  - Added structured previews, logs, CSV reports, error details, and dry-run validation.
- **Bug Fixes**
  - Improved handling of unsupported layouts, padding, missing results, invalid settings, and benchmark failures.
- **Tests**
  - Added extensive coverage for model inspection, argument validation, benchmark generation, quantization, routing, reporting, and error handling.

### DIFF
--- a/.agents/skills/benchmark-model-kernels/SKILL.md
+++ b/.agents/skills/benchmark-model-kernels/SKILL.md
@@ -45,15 +45,16 @@ order:
      --tp <tp> --ep <ep> --ms <m1> <m2> ... --print_only
    ```
 
-   Review the printed shapes, MoE tuple, and derived routing with the user.
+   Review the printed shapes and the MoE tuple with the user.
    `# unsupported:` lines mean the list is partial and the script exits
    nonzero — handle those via **Manual supplements** below.
 5. **FlashInfer checkout and GPU** — the full benchmark needs a FlashInfer
    *source checkout* containing `benchmarks/flashinfer_benchmark.py`; the
    installed wheel alone is not enough. Prefer a clean checkout matching the
    installed `flashinfer` version; ask before cloning or installing anything.
-   On the benchmark machine, check `nvidia-smi` and package versions, and
-   verify CUPTI timing with a tiny `bench_gpu_time(..., enable_cupti=True)`
+   On the benchmark machine, check `nvidia-smi` and package versions, verify
+   the target GPU is idle (concurrent work on the same GPU skews timings),
+   and verify CUPTI timing with a tiny `bench_gpu_time(..., enable_cupti=True)`
    probe — a warning that falls back to CUDA events is a failure (the
    `cupti-python`/`nvidia-cuda-cupti` packages must match PyTorch's CUDA
    major). Ask for a GPU index only when several are visible. Pick a fresh
@@ -77,14 +78,10 @@ order:
 
 Do not restate, re-derive, or override these — the code enforces them:
 
-- `benchmark_model.py` derives `--nks`, `--nk_names`, and all `--moe_*`
-  arguments including the routing method; overrides are rejected.
-- Physical padding follows one rule: pad a dimension if and only if vLLM pads
-  it for that case; otherwise keep the exact shape and let it fail the same
-  way vLLM would. Row labels stay logical — report both shapes when padding
-  changes a dimension.
-- BF16 and FP8 MoE cases run before NVFP4 because a CUDA fault poisons the
-  remaining cases in the same driver process.
+- `benchmark_model.py` derives `--nks` (as `N,K,module-name` triples) and
+  all `--moe_*` arguments; overrides are rejected.
+- Row labels are logical shapes; the runner applies vLLM's physical padding.
+  Report both shapes whenever they differ.
 - A failed case writes FlashInfer's error message (or a pointer to
   `driver.log`) into its `combined_results.csv` cell, and the command exits
   nonzero after the table is written. Never present a partial table as a
@@ -92,11 +89,14 @@ Do not restate, re-derive, or override these — the code enforces them:
 
 ## Known backend and driver limits
 
-- `mm_fp4` TensorRT-LLM needs `N % 128 == 0` (shuffled weight layout). Drivers
-  with the conditional-shuffle fix drop only that backend (its cell reports no
-  result row); stock 0.6.x drivers fail *all* backends for that shape with an
-  empty assertion.
+- `mm_fp4` TensorRT-LLM needs `N % 128 == 0` (shuffled weight layout): its
+  cell reports no result row, or on stock 0.6.x drivers an empty assertion
+  that fails every `mm_fp4` backend for that shape.
 - `mm_fp8` (trtllm_low_latency) needs `K % 128 == 0`.
+- MoE rows cover FlashInfer's CUTLASS fused MoE, the trtllm-gen NVFP4 and
+  per-tensor FP8 MoE, and the CuteDSL NVFP4 MoE. The CuteDSL MoE row appears
+  only for Swiglu models (the kernel supports nothing else); the `cutedsl` and
+  `trtllm` backends require recent GPUs and report per-case errors elsewhere.
 - Gated NVFP4 CUTLASS MoE with `2F % 128 != 0` per rank fails — vLLM raises
   instead of padding — often as a `CUDA error: misaligned address`. Prefer EP
   over TP for the experts to keep the per-rank width legal.
@@ -114,23 +114,32 @@ missing shape:
 CUDA_VISIBLE_DEVICES=<gpu-index> \
 python .agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py \
   --flashinfer_repo <flashinfer-repo> --ms <m1> <m2> ... \
-  --nks <n>,<k> --nk_names <name> --workdir <workdir>
+  --nks <n>,<k>,<name> --workdir <workdir>
 ```
 
-For a manual MoE supplement, also derive and pass the routing arguments
-(`--moe_routing_method` plus the DeepSeek group/scaling/bias flags when
-applicable) — the runner otherwise defaults to `topk`, a different routing
-kernel. Label these rows as manual supplements; this is not a substitute for
-running `benchmark_model.py` first. Shell-quote user-supplied paths.
+Label these rows as manual supplements; this is not a substitute for running
+`benchmark_model.py` first. Shell-quote user-supplied paths.
 
 ## Report
 
 Report the command, GPU, versions, TP/EP, M values, shapes (logical and
 physical where they differ), warnings, and the artifacts: `testlist.txt`,
 `driver.log` (full driver output), `builtin_results.csv` (milliseconds,
-success-only), and `combined_results.csv` (microseconds; GEMM rows grouped
-under `<name>: <NxK>` labels, MoE rows flat). `*_with_quant` rows add a
-separately measured activation-quantization time, except NVFP4 MoE with
-quantization, which is a single fused measurement. These are kernel times
+success-only), and `combined_results.csv` (long form, microseconds: columns `module_name,
+M, N, K, backend, with_quant, runtime` in `GEMM` and `MoE` sections; modules
+fused into one GEMM are joined with `|` in `module_name`, while distinct
+same-shape modules appear as duplicated rows sharing one measurement; MoE
+rows keep the `H= F= E= top_k=` parameter line and leave N/K empty). In quantization-recipe terms:
+`bf16` rows are the unquantized W16A16 baseline, `fp8` rows are per-tensor
+W8A8, and `nvfp4` rows are W4A4. Plain quantized rows time the kernel with
+pre-quantized activations; `*_with_quant` rows add a separately measured
+activation-quantization time in the scale-factor layout that backend
+consumes, except the NVFP4 CUTLASS MoE row, which is a single fused
+measurement. MoE routing is synthetic: uniform expert
+distribution everywhere (real skewed routing is slower), and the trtllm-gen
+rows, which route in-kernel, use a fixed `renormalize` method regardless of
+the model's routing scheme. CUTLASS and CuteDSL MoE rows receive precomputed
+expert indices and exclude routing-selection cost entirely. No row includes
+the router GEMM itself. These are kernel times
 only — never describe them as end-to-end latency or throughput; they omit
 weights, layer frequency, communication, KV cache, and scheduling.
--- a/.agents/skills/benchmark-model-kernels/SKILL.md
+++ b/.agents/skills/benchmark-model-kernels/SKILL.md
@@ -1,0 +1,134 @@
+---
+name: benchmark-model-kernels
+description: >-
+  Inspect Hugging Face decoder layers on meta tensors and plan or run per-rank
+  BF16, FP8, and NVFP4 GEMM or fused-MoE microbenchmarks with the bundled
+  scripts and a local FlashInfer checkout. Use when choosing a model, GPU, TP,
+  EP, or M/token-concurrency sweep; deriving common fused QKV and gate/up
+  shapes without loading checkpoint weights; or using FlashInfer benchmark
+  utilities. Do not use for end-to-end server throughput or request latency.
+---
+
+# Benchmark Model Kernels
+
+Plan a single-GPU microbenchmark for the per-rank shapes of an intended
+deployment. The scripts never load checkpoint weights, launch distributed
+workers, or measure collectives, serving throughput, or request latency.
+
+## Interview, one decision per message
+
+Ask exactly one unresolved decision per message with two or three concrete
+options, state the default, and wait for the answer. Recommend an option only
+when it is substantively better. Skip decisions already answered. Follow this
+order:
+
+1. **Model** — a local directory, a `config.json`, or a Hub ID (equivalent; no
+   default). The script builds the model on meta tensors from configuration
+   only. Do not enable `--trust-remote-code` without explicit approval; pin
+   `--revision <commit>` when it is needed.
+2. **TP and EP** — accept both directly, or derive them from the intended
+   deployment's GPU model and count and confirm. Defaults: `TP=1`, `EP=1`.
+   Derive parallelism from the intended deployment, never from the GPU that
+   runs the microbenchmark. The script validates every sharding rule
+   (divisibility, GQA replication, expert partitioning) and errors loudly.
+3. **M sweep** — balanced default (recommended): `1 8 64 512`; decode-focused:
+   `1 4 16 32`; throughput-focused: `64 256 1024 4096`. M is roughly the
+   tokens scheduled per step (decode: active sequences), not endpoint
+   concurrency.
+4. **Shape preview** — always run this before anything else; no FlashInfer or
+   GPU needed:
+
+   ```bash
+   python .agents/skills/benchmark-model-kernels/scripts/benchmark_model.py <model> \
+     --tp <tp> --ep <ep> --ms <m1> <m2> ... --print_only
+   ```
+
+   Review the printed shapes, MoE tuple, and derived routing with the user.
+   `# unsupported:` lines mean the list is partial and the script exits
+   nonzero — handle those via **Manual supplements** below.
+5. **FlashInfer checkout and GPU** — the full benchmark needs a FlashInfer
+   *source checkout* containing `benchmarks/flashinfer_benchmark.py`; the
+   installed wheel alone is not enough. Prefer a clean checkout matching the
+   installed `flashinfer` version; ask before cloning or installing anything.
+   On the benchmark machine, check `nvidia-smi` and package versions, and
+   verify CUPTI timing with a tiny `bench_gpu_time(..., enable_cupti=True)`
+   probe — a warning that falls back to CUDA events is a failure (the
+   `cupti-python`/`nvidia-cuda-cupti` packages must match PyTorch's CUDA
+   major). Ask for a GPU index only when several are visible. Pick a fresh
+   workdir and state it.
+6. **Full benchmark**:
+
+   ```bash
+   CUDA_VISIBLE_DEVICES=<gpu-index> \
+   python .agents/skills/benchmark-model-kernels/scripts/benchmark_model.py <model> \
+     --tp <tp> --ep <ep> --ms <m1> <m2> ... \
+     --flashinfer_repo <flashinfer-repo> --workdir <workdir>
+   ```
+
+   Run a short plumbing check first (append
+   `--dry_run_iters 1 --num_iters 3 --no_autotune`, throwaway workdir),
+   especially after changing FlashInfer versions or shape logic; never present
+   it as a performance result. Then run with defaults. The first fused-MoE
+   build or autotune can take several minutes; its cache is reused.
+
+## Rules the scripts own
+
+Do not restate, re-derive, or override these — the code enforces them:
+
+- `benchmark_model.py` derives `--nks`, `--nk_names`, and all `--moe_*`
+  arguments including the routing method; overrides are rejected.
+- Physical padding follows one rule: pad a dimension if and only if vLLM pads
+  it for that case; otherwise keep the exact shape and let it fail the same
+  way vLLM would. Row labels stay logical — report both shapes when padding
+  changes a dimension.
+- BF16 and FP8 MoE cases run before NVFP4 because a CUDA fault poisons the
+  remaining cases in the same driver process.
+- A failed case writes FlashInfer's error message (or a pointer to
+  `driver.log`) into its `combined_results.csv` cell, and the command exits
+  nonzero after the table is written. Never present a partial table as a
+  successful benchmark; read `driver.log` before rerunning anything.
+
+## Known backend and driver limits
+
+- `mm_fp4` TensorRT-LLM needs `N % 128 == 0` (shuffled weight layout). Drivers
+  with the conditional-shuffle fix drop only that backend (its cell reports no
+  result row); stock 0.6.x drivers fail *all* backends for that shape with an
+  empty assertion.
+- `mm_fp8` (trtllm_low_latency) needs `K % 128 == 0`.
+- Gated NVFP4 CUTLASS MoE with `2F % 128 != 0` per rank fails — vLLM raises
+  instead of padding — often as a `CUDA error: misaligned address`. Prefer EP
+  over TP for the experts to keep the per-rank width legal.
+- Do not benchmark a padded shape to dodge these limits and call it
+  vLLM-equivalent; report the limit instead.
+
+## Manual supplements
+
+For each `# unsupported:` layout from the preview: inspect that module's
+forward path and the intended runtime's TP/EP sharding, derive the per-rank
+shape (never guess sharding from a weight shape alone), and benchmark only the
+missing shape:
+
+```bash
+CUDA_VISIBLE_DEVICES=<gpu-index> \
+python .agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py \
+  --flashinfer_repo <flashinfer-repo> --ms <m1> <m2> ... \
+  --nks <n>,<k> --nk_names <name> --workdir <workdir>
+```
+
+For a manual MoE supplement, also derive and pass the routing arguments
+(`--moe_routing_method` plus the DeepSeek group/scaling/bias flags when
+applicable) — the runner otherwise defaults to `topk`, a different routing
+kernel. Label these rows as manual supplements; this is not a substitute for
+running `benchmark_model.py` first. Shell-quote user-supplied paths.
+
+## Report
+
+Report the command, GPU, versions, TP/EP, M values, shapes (logical and
+physical where they differ), warnings, and the artifacts: `testlist.txt`,
+`driver.log` (full driver output), `builtin_results.csv` (milliseconds,
+success-only), and `combined_results.csv` (microseconds; GEMM rows grouped
+under `<name>: <NxK>` labels, MoE rows flat). `*_with_quant` rows add a
+separately measured activation-quantization time, except NVFP4 MoE with
+quantization, which is a single fused measurement. These are kernel times
+only — never describe them as end-to-end latency or throughput; they omit
+weights, layer frequency, communication, KV cache, and scheduling.
--- a/.agents/skills/benchmark-model-kernels/SKILL.md
+++ b/.agents/skills/benchmark-model-kernels/SKILL.md
@@ -34,7 +34,9 @@ order:
 3. **M sweep** — balanced default (recommended): `1 8 64 512`; decode-focused:
    `1 4 16 32`; throughput-focused: `64 256 1024 4096`. M is roughly the
    tokens scheduled per step (decode: active sequences), not endpoint
-   concurrency.
+   concurrency. With EP, an MoE row models one rank's share of the global
+   batch: a global batch of B tokens corresponds to the column M = B/EP, so
+   do not compare different EP values at the same M.
 4. **Shape preview** — always run this before anything else; no FlashInfer or
    GPU needed:
 
--- a/.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
+++ b/.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
@@ -16,9 +16,9 @@
 """Derive per-rank benchmark shapes from a Transformers model on meta tensors.
 
 The script walks the instantiated model's Linear modules, fuses Q/K/V and
-gate/up projections, recognizes Mamba 2 and common routed-expert layouts, maps
-config routing fields to FlashInfer's MoE routing method, and applies the
-common serving/export TP layout. It never calls a checkpoint weight loader.
+gate/up projections, recognizes Mamba 2, GatedDeltaNet, and common
+routed-expert layouts, and applies the common serving/export TP layout. It
+never calls a checkpoint weight loader.
 When a decoder layout is unsupported, the derived shapes are still printed and
 the script exits nonzero; benchmark the missing shapes directly with
 benchmark_via_builtin.py.
@@ -26,9 +26,10 @@ benchmark_via_builtin.py.
 
 import argparse
 import importlib.util
+import re
 import shlex
 import sys
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
@@ -44,6 +45,9 @@ class _MoeShape:
     experts: int
     top_k: int
     activation: str | None = None
+    # The expert container's module path. Metadata, not identity: two layouts
+    # that differ only by path are still one shape for dedup purposes.
+    name: str = field(default="experts", compare=False)
 
 
 @dataclass(frozen=True)
@@ -54,15 +58,32 @@ class _ExpertShape:
 
 
 @dataclass(frozen=True)
-class _MoeRouting:
-    method: str
-    num_expert_group: int | None = None
-    topk_group: int | None = None
-    routed_scaling_factor: float | None = None
-    use_routing_bias: bool = False
+class _Kernel:
+    """One derived per-rank GEMM: N x K, labeled by its source module path(s)."""
+
+    n: int
+    k: int
+    label: str
 
 
-_Kernel = tuple[int, int, str]
+@dataclass(frozen=True)
+class _MambaLayout:
+    in_shape: tuple[int, int]
+    out_shape: tuple[int, int]
+    intermediate: int
+    heads: int
+    groups: int
+    state: int
+
+
+@dataclass(frozen=True)
+class _GdnLayout:
+    out_shape: tuple[int, int]
+    num_k_heads: int
+    num_v_heads: int
+    key_dim: int
+    value_dim: int
+
 
 _PROJECTIONS = {
     "q_proj",
@@ -76,20 +97,25 @@ _PROJECTIONS = {
 }
 _RESERVED = {
     "--nks",
-    "--nk_names",
+    "--moe_name",
     "--moe_hidden_size",
     "--moe_intermediate_size",
     "--moe_num_experts",
     "--moe_top_k",
     "--moe_activation_type",
-    "--moe_routing_method",
-    "--moe_num_expert_group",
-    "--moe_topk_group",
-    "--moe_routed_scaling_factor",
-    "--moe_use_routing_bias",
 }
 
-_ROUTER_NAMES = {"gate", "router", "router_proj"}
+# Modules deliberately outside the benchmarked GEMM list fall into two
+# exclusion mechanisms:
+# 1. Positional: embeddings, the LM head, and anything else outside the
+#    decoder blocks never enter the audit (the `layers.<i>` position filter
+#    in _unsupported_decoder_linears).
+# 2. Name-based: routing and gating projections that live inside decoder
+#    blocks are excluded by these names. They are never quantized in
+#    deployment recipes and vLLM dispatches them through specialized (often
+#    FP32-output) paths, so a standard GEMM row would not model them anyway.
+_ROUTER_PATH_PARTS = {"router", "routers"}
+_GATING_LEAF_NAMES = {"gate", "router", "router_proj", "shared_expert_gate"}
 
 
 def _positive_int(value: str) -> int:
@@ -121,7 +147,15 @@ def _load_meta_model(model_ref: str, trust_remote_code: bool, revision: str | No
         if revision and trust_remote_code and "AutoModelForCausalLM" in auto_map:
             model_kwargs["code_revision"] = revision
         with init_empty_weights(include_buffers=True):
-            model = AutoModelForCausalLM.from_config(config, **model_kwargs)
+            try:
+                model = AutoModelForCausalLM.from_config(config, **model_kwargs)
+            except Exception:
+                text_config = getattr(config, "text_config", None)
+                if text_config is None:
+                    raise
+                # Multimodal wrapper configs build their text decoder
+                # directly; vision towers are outside the benchmark scope.
+                model = AutoModelForCausalLM.from_config(text_config, **model_kwargs)
     except Exception as exc:
         raise ShapeError(f"could not construct {model_ref!r} on meta tensors: {exc}") from exc
 
@@ -144,6 +178,12 @@ def _divide(value: int, size: int, label: str) -> int:
     return value // size
 
 
+def _path_label(parent: str, leaf: str) -> str:
+    """Label a GEMM by its module path with layer indices normalized to ``*``."""
+    path = f"{parent}.{leaf}" if parent else leaf
+    return re.sub(r"(?<=\.)\d+(?=\.|$)", "*", path)
+
+
 def _fused_qkv(
     q: tuple[int, int],
     k: tuple[int, int],
@@ -164,7 +204,8 @@ def _fused_qkv(
         local_q = _divide(q_heads, tp, f"{parent}.q_proj heads") * head_dim
         _divide(tp, kv_heads, "TP/KV replication ratio")
         local_n = local_q + 2 * head_dim
-    return local_n, q[1], "fused_qkv"
+    label = "|".join(_path_label(parent, leaf) for leaf in ("q_proj", "k_proj", "v_proj"))
+    return _Kernel(local_n, q[1], label)
 
 
 def _dense_kernels(model: Any, config: Any, tp: int) -> list[_Kernel]:
@@ -172,11 +213,15 @@ def _dense_kernels(model: Any, config: Any, tp: int) -> list[_Kernel]:
     for name, module in model.named_modules():
         leaf = name.rsplit(".", 1)[-1]
         parts = name.split(".")
+        # Routed experts are excluded here and handled by _moe_shapes. Shared
+        # experts (for example `.shared_experts.up_proj`) intentionally do not
+        # match the filter: they run densely for every token, so their
+        # projections belong in the dense GEMM list.
         if (
             leaf not in _PROJECTIONS
             or ".experts." in name
             or ".local_experts." in name
-            or any(part in {"router", "routers"} for part in parts[:-1])
+            or any(part in _ROUTER_PATH_PARTS for part in parts[:-1])
         ):
             continue
         shape = _linear_shape(module)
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -288,7 +288,13 @@ module = ["examples.diffusers.fastgen.preprocess.*"]
...
 
 [tool.bandit]
-exclude_dirs = [".github/", "examples/", "noxfile.py", "tests/"]
+exclude_dirs = [
+    ".agents/skills/benchmark-model-kernels/",
+    ".github/",
+    "examples/",
+    "noxfile.py",
+    "tests/",
+]
```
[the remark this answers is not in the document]
sychen52: @kevalmorabia97 @shengliangxu , I added a python script that will call the benchmark utils in flashinfer to run benchmark, so it will use subprocess to call it. Can I add this script to the exclude directories of bandit?

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,650 @@
...
+import shlex
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+import flashinfer
+import numpy as np
+import torch
+from flashinfer.testing import bench_gpu_time
```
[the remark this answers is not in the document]
sychen52: that is okay, the checkout is to run the benchmark script. flashinfer can be installed through pip. We need checkout because the benchmark utility is not in the pip install.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,655 @@
...
+            fn=kernel,
+            input_args=inputs,
+            dry_run_iters=dry_runs,
+            repeat_iters=iterations,
+            enable_cupti=True,
+            use_cuda_graph=cuda_graph,
+            cold_l2_cache=True,
+            sleep_after_run=True,
+        )
+        results[(kind, m, k)] = float(np.median(times)) * 1000
```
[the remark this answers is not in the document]
sychen52: raw results persist in builtin_results.csv; environmental failures should crash loudly; blanket except hides tool bugs

cjluo-nv (commented):

New self-contained skill (`benchmark-model-kernels`) with two scripts that (1) derive per-rank GEMM/MoE benchmark shapes from a HF config on meta tensors and (2) drive FlashInfer's built-in benchmark for BF16/FP8/NVFP4. +2301/-1, 3 non-test files + 2 test files + a bandit-exclude tweak.

Design review (complexity gate fired): The problem — a per-layer microbenchmark harness that derives per-rank shapes from model configs without loading weights — has no existing in-repo equivalent (searched; other skills are task-specific and use `tests.json` eval metadata, not benchmarking). This *wraps* FlashInfer's `flashinfer_benchmark.py` rather than reimplementing it; the value-add is the config→shape derivation. OmegaConf/Hydra/Jinja aren't relevant. The pytest-in-skill layout is already an established convention (pyproject has a `.agents/skills/*/tests/test_*.py` per-file-ignore). Design looks reasonable and the PR body gives concrete usage, though the rationale is brief.

Licensing: clean — new files carry the canonical `LICENSE_HEADER` NVIDIA header verbatim; no LICENSE/vendoring changes. The bandit `exclude_dirs` addition is narrowly scoped to the new skill (justified by the intentional `subprocess` use, which is invoked without a shell and gates `trust_remote_code` behind an explicit flag).

Correctness: I traced the shape-derivation logic (fused QKV incl. GQA replication, gate/up, Mamba-2, ModuleList vs stacked/direct experts, EP-vs-TP intermediate sharding, routing derivation, expert audit, override rejection) and it holds up; the unit tests are genuinely meaningful and internally consistent across Llama/Mixtral/GptOss/NemotronH plus edge cases.

Why nudge (not approve): (1) sizable new subsystem worth an owner's architectural eyes; (2) coverage gap — `benchmark_via_builtin.py`'s GPU/FlashInfer runtime paths (`_run_driver` real invocation, `_quant_times`/`_nvfp4_runner`/`_fp8_runner` CUDA timing) are mocked out and never exercised in CI, consistent with the repo's known GPU-only coverage risk; (3) CHANGELOG left unmarked — acceptable for an internal agent skill (no shipped library API change), but worth confirming. No prompt-injection concerns in the untrusted content (SKILL.md's "do not override" prose targets the skill's agent user, not the reviewer).

.agents/skills/benchmark-model-kernels/tests/test_benchmark_model.py
```diff
@@ -0,0 +1,495 @@
...
+import importlib.util
+import sys
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+pytest.importorskip("torch")
+pytest.importorskip("accelerate")
+transformers = pytest.importorskip("transformers")
```
Edwardf0t1, line 25: **These tests never run in CI as written.** The `skills` job in `.github/workflows/unit_tests.yml` installs only `pytest` (it's documented as "stdlib-only and hermetic"), so these `importorskip` calls on `torch`/`accelerate`/`transformers` skip this entire file — and `pytest.importorskip("flashinfer")` skips all of `test_benchmark_via_builtin.py` the same way. CI stays green whether or not the code works, so the ~870 lines of tests only ever run on a dev machine.

Suggestions:
- Route this file into the main `linux` unit lane (it needs only torch/accelerate/transformers — all standard unit-test deps, no GPU).
- In `benchmark_via_builtin.py`, defer `import flashinfer` into the functions that need it (same pattern as the existing `vllm` try/except). The majority of that test file (`_parser`, `_combine`, `_write_results`, `_parse_driver_errors`, case generation) is pure CPU logic that could then run without the GPU stack.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,643 @@
...
+    moe = next(iter(moe_shapes), None)
+    routing = None
+    if moe is None:
+        if ep != 1 and not problems:
+            raise ShapeError("EP requires routed experts")
+    else:
+        local_experts = _divide(moe.experts, ep, "expert count")
+        intermediate = moe.intermediate
+        if ep == 1:
+            intermediate = _divide(intermediate, tp, "expert intermediate size")
```
Edwardf0t1: **Mixed TP×EP is silently mis-modeled for the MoE tuple.** `intermediate` is divided by TP only when `ep == 1`, so `--tp 4 --ep 2` produces experts at full width with `experts/2` per rank — TP is quietly ignored for the expert GEMMs, and the derived shapes correspond to a deployment nobody runs. Pure EP (`--enable-expert-parallel`-style) and pure expert-TP are both fine; it's the combination that's neither.

Since the script's contract is "validates every sharding rule and errors loudly" (and it rejects `--moe_*` overrides, so the user can't correct it downstream), consider raising a `ShapeError` for `tp > 1 and ep > 1` when routed experts are present — or explicitly printing which sharding interpretation was applied so the operator can catch it. SKILL.md's "prefer EP over TP for the experts" hints at this, but the code neither enforces nor announces it.

Edwardf0t1 (commented):
Posting two inline comments from my review: one on CI test discovery, one on TP×EP handling for the MoE shape.

pyproject.toml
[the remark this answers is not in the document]
shengliangxu: why? does bandit complain about subprocess?

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: add more strict rules to follow vllm style. Now EP has to be multiple of TP.

.agents/skills/benchmark-model-kernels/tests/test_benchmark_model.py
sychen52: import moved. However, route test to main unit lane involve workflow change.

pyproject.toml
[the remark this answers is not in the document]
kevalmorabia97: @yeyu-nvidia can you just add a `# nosec: <CODE>` next to all lines where bandit complains instead of skipping the folder? If there are too many then we can just keep it here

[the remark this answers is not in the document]
sychen52: @shengliangxu yes.
@kevalmorabia97 I did that initially, but CodeRabbit's pre-merge check give me an error. I guess you mean I add # nosec and disregard the error and you will approval it in the PR, right?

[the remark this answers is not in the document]
kevalmorabia97: Yes

cjluo-nv:
Could you update the PR description to include an example prompt and expected output?

pyproject.toml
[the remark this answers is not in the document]
shengliangxu: yes, these files should be checked by all our coding tools including bandit, we should solve the issues as much as possible rather than just disable the check.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,660 @@
...
+            errors[pending_tag] = message
+            pending_tag = None
+    return errors
+
+
+def _run_driver(
+    benchmarks_dir: Path, testlist: Path, output: Path, driver_log: Path
+) -> tuple[int, list[str]]:
+    # This invokes the explicitly selected FlashInfer checkout without a shell.
+    process = subprocess.Popen(  # nosec B603
```
sychen52, line 235: @kevalmorabia97 @shengliangxu here is the subprocess invoke. You can see that coderabbit is paused because of this.

shengliangxu: got it, just exclude checking on these lines should be the way to go.

shengliangxu (approved):

juhi10071998 (commented):
Thanks for adding this Shiyang!

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,660 @@
...
+                    "bmm_fp8",
+                    backend,
+                    n,
+                    k,
+                    ["--batch_size", "1"],
+                    "fp8_static",
+                )
+                for backend in ("cudnn", "cutlass")
+            ]
+            variants.append(("fp8_trtllm", "mm_fp8", "trtllm_low_latency", n, k, [], "fp8_static"))
```
juhi10071998: A single CUDA fault can wipe out the whole run, should we have subprocess here for each case?

The entire testlist runs in one driver process (`_run_driver`, line 612), so one CUDA-fatal case takes down everything after it — and the fragile backends run first: `nvfp4_*` is generated before `fp8_*` here, so an `nvfp4_trtllm` fault (`N % 128 == 0`) or `fp8_trtllm` (`K % 128 == 0`) can poison the safe formats too. Ordering the testlist safe-first (bf16/cudnn/cutlass, then trtllm/gated-nvfp4 last) would keep one fault from wiping the rest of the run.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,657 @@
...
+        raise ShapeError(f"unsupported non-gated MoE activation {configured!r}")
+    return activations[normalized]
+
+
+def _top_k(config: Any) -> int | None:
+    for attr in ("num_experts_per_tok", "moe_top_k"):
+        value = getattr(config, attr, None)
+        if value is not None:
+            return int(value)
+    return None
```
juhi10071998, line 531: `_top_k` misses some common config names.
This only checks `num_experts_per_tok`/`moe_top_k`, so models using `top_k`, `num_experts_per_token`, or `num_selected_experts` fall through to "could not determine MoE top_k". ModelOpt already keeps the full list in `_auto_quantize_cost.py` (`_ACTIVE_MOE_TOP_K_ATTRS`), we can reuse it

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,660 @@
...
+    ms: list[int],
+    shape: tuple[int, int, int, int] | None,
+    common: list[str],
+    activation: str | None,
+    routing: str | None,
+    routing_args: list[str] | None = None,
+) -> list[_Case]:
+    if shape is None:
+        return []
+    routine = "cutlass_fused_moe"
```
juhi10071998: Nit: MoE is CUTLASS-only — probably worth a line in SKILL.md.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,657 @@
...
+            continue
+        if routed_experts_handled and any(part in {"experts", "local_experts"} for part in parts):
+            continue
+        if leaf in _PROJECTIONS or leaf in _ROUTER_NAMES or name in mamba_projections:
+            continue
+        layouts.setdefault((leaf, *shape), name)
+    return [(name, n, k) for (leaf, n, k), name in layouts.items()]
+
+
+def _inspect_model(
```
juhi10071998: Should we make the shape derivation here (`_inspect_model` / `_dense_kernels` / `_moe_shapes`) be exposed as an importable function, not just reachable through the CLI `main()`? 

AutoQuantize could leverage it to map its own modules onto the same LUT keys using the exact same logic instead of reimplementing it — and more generally this seems like a broadly useful utility worth exposing on its own.

.agents/skills/benchmark-model-kernels/SKILL.md
```diff
@@ -0,0 +1,136 @@
...
+kernel. Label these rows as manual supplements; this is not a substitute for
+running `benchmark_model.py` first. Shell-quote user-supplied paths.
+
+## Report
+
+Report the command, GPU, versions, TP/EP, M values, shapes (logical and
+physical where they differ), warnings, and the artifacts: `testlist.txt`,
+`driver.log` (full driver output), `builtin_results.csv` (milliseconds,
+success-only), and `combined_results.csv` (microseconds; GEMM rows grouped
+under `<name>: <NxK>` labels, MoE rows flat). `*_with_quant` rows add a
```
juhi10071998: Should we spell out what each format name means in quant terms — `fp8` = W8A8, `nvfp4` = W4A4, and `_with_quant` = activation-quantization included? It makes mapping a row to a specific ModelOpt recipe much easier

juhi10071998 (commented):
Two more integration asks inline, for feeding this into AutoQuantize.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,657 @@
...
+        if "gate_proj" in layers and "up_proj" in layers:
+            gate_n, gate_k = layers["gate_proj"]
+            up_n, up_k = layers["up_proj"]
+            if gate_k != up_k:
+                raise ShapeError(f"gate/up inputs differ under {parent}")
+            n = _divide(gate_n + up_n, tp, f"{parent}.gate_up")
+            kernels.append((n, gate_k, "fused_gate_up"))
+        elif "gate_proj" in layers:
+            raise ShapeError(f"gate projection has no matching up projection under {parent}")
+        elif "up_proj" in layers:
```
juhi10071998: if only `up_proj` is present, shouldn't this fail too symmetrically as gate_proj?

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,657 @@
...
+        if "o_proj" in layers:
+            n, k = layers["o_proj"]
+            kernels.append((n, _divide(k, tp, f"{parent}.o_proj"), "attention_out"))
+
+        if "gate_proj" in layers and "up_proj" in layers:
+            gate_n, gate_k = layers["gate_proj"]
+            up_n, up_k = layers["up_proj"]
+            if gate_k != up_k:
+                raise ShapeError(f"gate/up inputs differ under {parent}")
+            n = _divide(gate_n + up_n, tp, f"{parent}.gate_up")
```
juhi10071998: Should we also check individually if the gate_n and up_n are divisible by tp. I guess vLLM first loads individual gate and up projection weights as sharded on tp layers and then merge? if yes it may fail

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,657 @@
...
+    return _ExpertShape(hidden, intermediate, factor == 2)
+
+
+def _moe_activation(config: Any, gated: bool) -> str | None:
+    configured = getattr(config, "mlp_hidden_act", None) or getattr(config, "hidden_act", None)
+    normalized = str(configured).lower().replace("-", "_")
+    if gated:
+        if normalized in {"silu", "swiglu", "swish"}:
+            return "Swiglu"
+        if normalized == "geglu" or normalized.startswith("gelu") or normalized == "quick_gelu":
```
juhi10071998: are all Gelu variants mapped to Geglu in FlashInfer?
```
gelu
  gelu_new
  gelu_pytorch_tanh
  quick_gelu
      -> Geglu
```

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,657 @@
...
+
+def _top_k(config: Any) -> int | None:
+    for attr in ("num_experts_per_tok", "moe_top_k"):
+        value = getattr(config, attr, None)
+        if value is not None:
+            return int(value)
+    return None
+
+
+def _moe_shapes(model: Any, config: Any) -> set[_MoeShape]:
```
juhi10071998, line 534: where do we account for the shared experts here?

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
sychen52: Now, I run each case in a separate process.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
sychen52: added more backends: cutedsl and trtllm.
cutedsl and cutlass has the router outside. trtllm has the router inside the kernel.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: imported from modelopt.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,660 @@
...
+
+def _format_result(value: _ResultValue | None) -> str:
+    if value is None:
+        return ""
+    if isinstance(value, float):
+        return f"{value:.3f}"
+    return value
+
+
+def _write_results(
```
sychen52, line 638: current format is already a csv. We can update this format once we figure out what autoquant needs as an input.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: if autoquantize needs it, we can move it out of this skill and then autoquantize can directly import those functions and call them.

.agents/skills/benchmark-model-kernels/SKILL.md
sychen52: done.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: No. up only is a legit case.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: done.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
juhi10071998: oh, I wasn't aware if we have cases with only `up_proj`, do you have any examples top of your mind?

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,660 @@
...
+    "SwigluBias",
+    "Relu2",
+    "SwigluStep",
+    "Identity",
+)
+_MOE_ROUTINGS = ("renormalize", "deepseek_v3", "llama4", "renormalize_naive", "topk")
+
+_ResultValue = float | str
+
+
```
juhi10071998, line 64: shall we record the flashinfer commit in the results for reproducibility?

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
juhi10071998: thanks for confirming @sychen52. so how do we assume that the tokens are distributed uniformly across the experts in this case? if the random initialized weights get loaded from HBM to shared memory, then the router logits can be non-uniform ig

.agents/skills/benchmark-model-kernels/SKILL.md
```diff
@@ -0,0 +1,143 @@
...
+   Review the printed shapes, MoE tuple, and derived routing with the user.
+   `# unsupported:` lines mean the list is partial and the script exits
+   nonzero — handle those via **Manual supplements** below.
+5. **FlashInfer checkout and GPU** — the full benchmark needs a FlashInfer
+   *source checkout* containing `benchmarks/flashinfer_benchmark.py`; the
+   installed wheel alone is not enough. Prefer a clean checkout matching the
+   installed `flashinfer` version; ask before cloning or installing anything.
+   On the benchmark machine, check `nvidia-smi` and package versions, and
+   verify CUPTI timing with a tiny `bench_gpu_time(..., enable_cupti=True)`
+   probe — a warning that falls back to CUDA events is a failure (the
```
meenchen, line 58: Would the benchmarks be impacted if the GPUs are used by other tasks concurrently? If true, we may want to check it before using

.agents/skills/benchmark-model-kernels/SKILL.md
```diff
@@ -0,0 +1,143 @@
+---
+name: benchmark-model-kernels
+description: >-
+  Inspect Hugging Face decoder layers on meta tensors and plan or run per-rank
```
meenchen, line 4: Can we make this a general benchmarking methodology and not specific to flashinfer? Later on, we might extend the skill for other kernel libraries.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: it is in the gemm section, not part of MOE.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: good catch. I updated the mapping as vllm did.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,660 @@
...
+                        str(m),
+                        *shape_args,
+                        "--intermediate_size",
+                        str(run_intermediate),
+                        *extra,
+                        *common,
+                    ],
+                    quant=quant,
+                )
+            )
```
juhi10071998: I'm curious how was the quantization coverage determined?

```
FP8 CUTLASS/TRT-LLM:
base row + separately added with_quant row

NVFP4 CUTLASS:
base row + directly measured with_quant row

NVFP4 TRT-LLM:
base row only

NVFP4 CuteDSL:
base row only
```

was there a reason why some combinations were not included here?

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,730 @@
...
+    for tag, reason in (errors or {}).items():
+        case = cases_by_tag.get(tag)
+        if case is None:
+            continue
+        error = f"ERROR: {reason}"
+        results[case.section].setdefault(case.key, error)
+        if case.quant is not None:
+            separator = "_MxNxK=" if case.section == "gemm" else "_M="
+            name, shape = case.key.split(separator, 1)
+            results[case.section].setdefault(f"{name}_with_quant{separator}{shape}", error)
```
juhi10071998: should we validate that each per-case CSV contains exactly one row whose case_tag matches the requested case; otherwise stale or malformed rows can be silently skipped and leave missing results without failing the run.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
```diff
@@ -0,0 +1,730 @@
...
+        moe = results["moe"]
+        if moe:
+            if gemm:
+                writer.writerow([])
+            writer.writerow(["MoE"])
+            writer.writerow(["M", *ms])
+            names = sorted({key.split("_M=", 1)[0] for key in moe})
+            for name in names:
+                cells = [moe.get(f"{name}_M={m}") for m in ms]
+                writer.writerow([name, *(_format_result(value) for value in cells)])
```
juhi10071998: I was wondering if it maybe helpful to add some more metadata to this table. Especially for MoE adding H, F, E, top_k might be helpful-

some meta data we can think of adding-
  `case_tag,op,M,logical_N,logical_K,H,F,E,top_k,format,backend,includes_quant,includes_topk {since in some backend it includes in some it doesn't despite router logits computed outside kernel}`

juhi10071998 (commented):
Thanks Shiyang, left some more minor comments

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,690 @@
...
+        if "down_proj" in layers:
+            n, k = layers["down_proj"]
+            kernels.append((n, _divide(k, tp, f"{parent}.down_proj"), "down"))
+    return list(dict.fromkeys(kernels))
+
+
+def _mamba_layout(
+    module: Any,
+) -> tuple[tuple[int, int], tuple[int, int], int, int, int, int] | None:
+    in_shape = _linear_shape(getattr(module, "in_proj", None))
```
juhi10071998, line 303: Hi @sychen52 , do linear attn with in_proj_a, and in_proj_b also fall in mamba category? https://github.com/NVIDIA/Model-Optimizer/blob/01c708e792e9573be872c6772bde9b0d72c62d28/modelopt_recipes/huggingface/qwen3_5_moe/ptq/nvfp4_experts_mse-fp8_rest-kv_fp8.yaml#L56

juhi10071998: for instance in the Qwen3-5 MoE i see each of the individual proj layers in the GatedDeltaNet are separate, I believe they would still be recognizable as the gemm shape, but wanted to confirm if i get that right?

And also, wanted to confirm that the mamba gemm layers are not executed as fused kernels right? I think then it becomes difficult to map these individual gemm to a corresponding latency cost. {or assign them different precision}

juhi10071998, line 303: does modelopt models something like this and we need to make them fused?
```
 in_proj_qkv + in_proj_z → /qkvz
 in_proj_a   + in_proj_b → /ba
```

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,690 @@
...
+            continue
+        if leaf in _PROJECTIONS or leaf in _ROUTER_NAMES or name in mamba_projections:
+            continue
+        layouts.setdefault((leaf, *shape), name)
+    return [(name, n, k) for (leaf, n, k), name in layouts.items()]
+
+
+def _inspect_model(
+    model: Any, config: Any, tp: int, ep: int
+) -> tuple[list[_Kernel], _MoeShape | None, _MoeRouting | None, list[str]]:
```
juhi10071998: is lm_head cost also computed here?

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
```diff
@@ -0,0 +1,690 @@
...
+        elif "up_proj" in layers:
+            n, k = layers["up_proj"]
+            kernels.append((_divide(n, tp, f"{parent}.up_proj"), k, "up"))
+        if "gate_up_proj" in layers:
+            n, k = layers["gate_up_proj"]
+            kernels.append((_divide(n, tp, f"{parent}.gate_up_proj"), k, "fused_gate_up"))
+        if "down_proj" in layers:
+            n, k = layers["down_proj"]
+            kernels.append((n, _divide(k, tp, f"{parent}.down_proj"), "down"))
+    return list(dict.fromkeys(kernels))
```
juhi10071998: Hi @sychen52 , a general comment if we should try to preserve the per-instance mapping from source module names to the deduplicated benchmark case here?

For this accounting to be useful broadly it might be helpful to map the measured latency to the actual source module names. I think it becomes important because that way we know what all modules were present in a group and catch any anomalies. Otherwise that information is hidden in the benchmark_model script and may be hard to debug.

we still deduplicate based on case_id but retain the source module names. curious to know your thoughts on this

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
sychen52: Done, add GPU information as well.

.agents/skills/benchmark-model-kernels/SKILL.md
sychen52: Yes. Added into skills.md. Note that even if it is not running at the start, if you launch some GPU task while the benchmark is running, it will still affect the numbers.

.agents/skills/benchmark-model-kernels/SKILL.md
sychen52: We can write another one using the benchmark functions in flashinfer for example. That will be another PR when we need it.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: No. It is almost always not quantized.

juhi10071998: got it, I think for edge models we do quantize lm_head too. @meenchen could you comment if it might be helpful to have that accounted for or is it w4a16 usually?

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
sychen52: added the missing ones.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
sychen52: Done.

.agents/skills/benchmark-model-kernels/scripts/benchmark_via_builtin.py
sychen52: added these info in the final CSV in the MoE section.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: I did not support GDN, I will put it in. However, this is not future proof. If some new layers come out in the future, benchmark_model.py may not be able to detect and run it automatically. However, the script will report the unsupported layers and the agent will try to get the number directly from the benchmark_via_builtin.py script.
Worst case, a human needs to get the shape correctly and pass it to benchmark_via_builtin.py.

sychen52: Like the GDN case. We will add the support to benchmark_model.py.

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: We have module names in the final csv, such as fused_qkv, mamba_in, etc. Is that good enough? or what do you mean by module names?

juhi10071998: I mean the transformer like names, `models.layers.0.self_attn.q_proj,`
 
 I know these will only be present for one layer after deduplication, but it may be useful still to actually consume the original module names/ list in case of fused groups for consuming these numbers in certain tasks.

sychen52: The last part of that is what I kept as nk_names. I just feel that is a bit too long so I only pick the last part.
We have to fuse though. the q_proj shape is never in the table, only the fused_qkv is in the table.

meenchen (approved):
Looks good to me

.agents/skills/benchmark-model-kernels/scripts/benchmark_model.py
sychen52: Done. Please check the example output in the PR description.

juhi10071998: Thanks Shiyang, I was profiling the Qwen/Qwen3.6-35B-A3B model and the Gemm list looks good to me, confirming if the moe module_name was expected as this one- 
image

juhi10071998: stored here- /home/scratch.juhim_coreai/code/benchmarking_stuff_flashinfer/qwen36_tp4_ep4_next_27_7_smoke03/combined_results.csv

sychen52: changed to `model.layers.*.mixer.experts`

PR #1983, opened 2026-07-16
[6425069][ONNX][Autocast] Fix autocast metadata propagation

gcunhase:
### What does this PR do?

Type of change: Bug fix

- Synchronizes existing ONNX tensor declarations when folding `Constant -> Cast` patterns, so the published `value_info` dtype matches the converted Constant payload.
- Computes `GatherND` output shapes directly when autocast is running in custom-op mode, avoiding a generic input-0 shape copy for this shape-changing standard ONNX operator.
- Restricts the previous input-0 shape fallback to actual custom-op outputs.
- Adds regression tests for folded Constant dtype metadata and custom-op-mode `GatherND` shape propagation.

### Usage

```python
# No user-facing API change. Existing Autocast usage remains:
$ python -m modelopt.onnx.autocast --onnx=model.onnx
```

### Testing

- Reproduced the metadata issue with ModelOpt `0.44.0` and with main ToT `cba8a5c62a1a54fe89fb69bfd484ea0a653c633a` before the fix.
- Verified the standalone reproduction no longer reports stale Constant dtype metadata or input-derived `GatherND` output shape after the fix.
- Ran `pytest tests/unit/onnx/autocast/test_precisionconverter.py::test_folded_constant_cast_updates_value_info_type tests/unit/onnx/autocast/test_precisionconverter.py::test_custom_op_mode_uses_schema_shape_for_standard_gathernd -q`
- Ran `pytest tests/unit/onnx/autocast/test_precisionconverter.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Precision conversion now better preserves/restores full model input/output metadata when enabled, including after custom-op type/shape propagation.
  * Improved shape inference for standard ops during custom-op mode (including `GatherND`, `Gather`, `Unsqueeze`, and `Shape`), with correct scalar (rank-0) handling and stricter input-shape propagation.
  * When folding redundant `Cast` after `Constant`, element-type metadata is now updated consistently across the graph and nested subgraphs.
* **Tests**
  * Added regression and custom-op mode tests for `Constant -> Cast -> Identity` folding and for `GatherND`/rank-change shape propagation.


cjluo-nv (approved):

Small, focused bug fix (+126/-1, 3 files) for ONNX autocast metadata propagation. Two changes, both correct and well-tested:

1. `_sync_value_info_elem_type` (utils.py) — after folding a Constant→Cast, propagates the new dtype into matching value_info/input/output entries, recursing into subgraphs. The `and tensor_type.elem_type` guard sensibly skips undefined-type entries. The regression test folds a real Constant→Cast and verifies both the value_info dtype and that `onnx.shape_inference` passes with `strict_mode=True, check_type=True`.

2. `_infer_gathernd_op_shape` (precisionconverter.py) — in custom-op mode, computes GatherND output shape directly instead of copying input-0's shape. The formula `indices_shape[:-1] + data_shape[batch_dims + index_rank:]` matches the ONNX GatherND spec; verified against the test (data [1,4,2], indices [1,2,2], batch_dims=1 → [1,2]). The old input-0 fallback is preserved via `elif` for non-GatherND nodes.

Nested helper functions are consistent with the existing style in this module. No new function-local imports, no licensing changes (no header edits/new files/license files), no injection attempts in PR content. CHANGELOG not updated, but this is an internal metadata-propagation bug fix.

Minor (non-blocking): `if shape := _infer_gathernd_op_shape(node):` treats an empty-list result (rank-0 GatherND output) as falsy and would fall through to the input-0 fallback — an unlikely edge case, and no worse than prior behavior.

galagam (approved):

PR #1992, opened 2026-07-19
[1/n] Adds skip-softmax calibration through the vLLM serving path

kaix-nv:
### What does this PR do?

Type of change: new feature

Calibrates skip-softmax thresholds through the vLLM V1 execution path for FlashAttention and FlashInfer. Calibration measures the paged KV-cache path used at serving time, aggregates raw skipped/total tile counts across tensor-parallel head shards, fits separate prefill and decode curves, and exports the existing `sparse_attention_config` checkpoint schema.

This uses raw counts rather than averaging per-rank sparsity ratios because TP ranks can contribute different tile populations; summing numerators and denominators before division preserves the global tile-weighted result. The vLLM adapter lives in `plugins/sparse_attn_calibration.py` rather than `SparseAttentionStatsManager`: the latter records module-local ratios for the HF calibration flow and has no aligned cross-process merge contract, while this path must merge per-sample raw counts from vLLM workers. Fitting and export still reuse `DynamicThresholdCalibrator` and the canonical conversion helpers so the model and checkpoint schema do not fork.

Skip decisions depend on tile geometry. The common Triton launch boundary fixes the KV tile at 128 tokens and the prefill query tile at 128 tokens, including for direct kernel callers. Single-query decode can use a 16x128 compute tile without changing its skip decision. Measurement bypasses autotuning; serving still tunes warp and pipeline-stage counts while keeping the decision geometry fixed.

### Usage

```bash
python examples/vllm_serve/calibrate_sparse_attn.py <CKPT> \
  --prompts_file prompts.txt \
  --target_sparse_ratio 0.7 \
  --fit_logspace \
  --tensor_parallel_size 4 \
  --decode_tokens 32 \
  --update_checkpoint_config
```

Calibration supports tensor parallelism and requires pipeline-parallel and data-parallel sizes of 1. It always writes `sparse_attention_config.json`; `--update_checkpoint_config` also merges the result into `<CKPT>/config.json`.

### Testing

Latest revision `1e969cb380` (rebased onto main `02b58eb146`, 2026-09-17):

- Calibration/count-fitting unit tests: **33 passed** (`test_sparse_attn_calibration.py` and `test_calibrator_fitting.py`).
- Paged and contiguous calibration GPU suite: **33 passed** (`test_paged_calibrate.py` and `test_triton_fa_calibrate.py`), including NHD/HND equivalence, partial query tiles, decode counts, and malformed-cache rejection. Run with `CUDA_VISIBLE_DEVICES=1` on an RTX A6000; local GPU 0 was unavailable.
- Calibration CLI tests: **21 passed** (`tests/examples/vllm_serve/test_calibrate_sparse_attn.py`).
- `pre-commit run --files <four changed files>`: passed, including Ruff, mypy, and Bandit.
- The new regression tests reproduced the skipped-counter truncation and missing cache-boundary checks before the fix. Calibration arithmetic and the 20-point threshold grid are unchanged.

Historical validation from earlier revisions (not rerun end-to-end for this update):

- `PYTHONPATH="$PWD" pytest -q tests/examples/vllm_serve/test_calibrate_sparse_attn.py tests/unit/torch/sparsity/attention_sparsity/test_sparse_attn_calibration.py` — 37 passed.
- `PYTHONPATH="$PWD" pytest -q tests/gpu_vllm/torch/sparsity/attention_sparsity/test_vllm_calibration.py tests/gpu_vllm/torch/sparsity/attention_sparsity/test_sparse_attn_worker.py` — 65 passed, including kv-first, blocks-first, and packed FlashAttention cache layouts.
- `PYTHONPATH="$PWD" pytest -q tests/gpu_vllm/torch/sparsity/attention_sparsity/test_vllm_runtime.py tests/unit/torch/sparsity/attention_sparsity/test_sparse_attn_config.py` — 33 passed.
- `PYTHONPATH="$PWD" pytest -q tests/gpu/torch/kernels/sparsity/attention/test_paged_calibrate.py tests/gpu/torch/kernels/sparsity/attention/test_triton_fa_calibrate.py tests/gpu/torch/kernels/sparsity/attention/test_triton_fa_skip_softmax.py` — 31 passed, 1 skipped because the GPU lacks enough shared memory for the fp32 tile.
- `pre-commit run --files <changed files>` — passed.
- Historical end-to-end Nemotron 3 Ultra (GCP job `558552`), TP4, FA4, 48 RULER prompts, and 20 threshold trials: completed `0:0` with prefill `(a, b) = (9.9104, 10.8881)`, respectively +0.147% and -0.066% versus the matching 20-point reference `(9.8958, 10.8953)`. The supplied legacy fit `(14.47, 10.91)` used a different threshold grid; its `b` differs by only -0.201%, while `a` retains the known grid-weighting shift.

- Is this change backward compatible?: ❌ Active skip-softmax fixes the calibrated decision geometry (serving still tunes warp/stage counts), and sparse-only vLLM installs fail fast for unsupported DCP, DBO/ubatching, speculative decoding, and FULL mixed-batch graphs instead of installing silently.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no copied code or new dependency.

### Additional Information

Pipeline parallelism is rejected during calibration because the current count-merging contract aligns records across tensor-parallel head shards, not across pipeline stages with disjoint attention layers. The unrelated HF padded-query behavior change was removed from this PR so it can be reviewed independently with its own compatibility test.

## Summary by CodeRabbit

- **New Features**
  - Added vLLM skip-softmax calibration for paged attention, including prefill/decode support and checkpoint configuration generation.
  - Added Muse Glimmer AutoQuantize, Alpamayo QAD, streaming Kimi-K3 conversion, and NVFP4 activation headroom calibration recipes.
  - Added calibration statistics aggregation, phase-specific fitting, threshold validation, and preservation of existing sparse-attention settings.

- **Bug Fixes**
  - Improved NVFP4 CPU/ONNX scale validation and clamping.
  - Added clearer handling for unsupported quantization, cache, CUDA graph, and engine configurations.
  - Standardized serving and calibration tile behavior.

- **Documentation**
  - Expanded vLLM serving guidance, calibration instructions, compatibility requirements, and sparse-attention limitations.


cjluo-nv (commented):

## Design review (protocol applied)

**Problem:** skip-softmax thresholds today can only be fit through the HF/PyTorch forward loop; they are then served on vLLM's paged KV cache, so the calibrated `(a, b)` never see the engine they run on. This PR calibrates through vLLM itself.

**Alternatives checked in-repo:** `DynamicThresholdCalibrator` (HF Stage 1/2/3 fit), `SparseAttentionStatsManager` (per-module stats), `conversion.export_sparse_attention_config` (checkpoint schema), `RulerDatasetBuilder`, and the existing `vllm_runtime._plan_vllm_attention` installer framework. Credit where due: the PR reuses all of these rather than forking them — it carves a `calibrate_from_stats` seam out of `DynamicThresholdCalibrator`, extracts `export_threshold_scale_factor`/`export_config_producer` from the HF exporter so the schema can't drift, and the new `install_vllm_skip_softmax_calibration` reuses `_AttentionPlan`/`_apply_vllm_attention_plans`. The only genuinely new machinery is cross-rank/cross-layer *raw count* merging, which has no in-repo equivalent (`SparseAttentionStatsManager` is single-process and stores ratios, which can't be summed across TP head shards). I consider the design defensible on the code, **but the PR body says none of this** — it's the unmodified template with an empty Testing section and unfilled checkboxes. Please write the two-paragraph rationale into the body (why counts-not-ratios, why a new `plugins/sparse_attn_calibration.py` instead of extending `stats_manager.py`, why the 128×128 tile is enforced in the kernel rather than at config load).

## Blocking / important

1. **FlashAttention calibration hardcodes one KV-cache layout.** `_forward_calibrate` is reached via `key_cache, value_cache = kv_cache.unbind(0)`, but the serving path in the very same class resolves this through `_flash_attention_kv_cache_layout()` (`kv-first` / `blocks-first` / `packed`). `tests/.../test_sparse_attn_worker.py::test_flash_attention_forward_follows_backend_kv_cache_layout` explicitly parameterizes all three, so this is an established repo contract. On a `blocks-first` or `packed` build, calibration dies with an opaque "too many values to unpack" before the nice `logical KV-cache view` guard is ever reached. The new GPU test builds its cache with `torch.stack(..., dim=0)`, i.e. it bakes in the same assumption and can't catch this.

2. **`flash_skip_softmax.py` padding fix changes shipped HF behavior with no test.** Masking padded query rows to `-inf` is correct (and matches the Triton reduction), but it changes measured sparsity *and* the serving `element_mask` for every non-`br`-aligned sequence on the existing PyTorch path — thresholds calibrated before this PR will realize different sparsity after it. No unit test covers the partial last block row, and the CHANGELOG entry only advertises the new vLLM feature.

3. **Sparse-only serving install gained new hard rejections.** `_global_errors(model_runner, sparse_only=not quantize)` now runs for `install_vllm_sparse_attention_from_checkpoint`, so DCP, DBO/ubatching, speculative decoding, FULL mixed-batch graphs, and (via the un-gated `_sparse_graph_error`) FULL decode graphs now raise `NotImplementedError` where sparse-only serving previously installed silently. That's a defensible fix, but it is backward-incompatible for existing servers and only appears in the README — the CHANGELOG's `*Sparsity*` bullet doesn't mention it, and the PR body's "Is this change backward compatible?" is unfilled.

4. **Fail-late CLI validation.** The driver deliberately pre-checks `--update_checkpoint_config` "before the (expensive, multi-GPU) calibration run", but `--target_sparse_ratio` is only range-checked inside `build_sparse_attention_config` *after* the run, and `--decode_tokens` is unchecked (a negative value yields `max_tokens <= 0` and an engine-side error after model load). Same fail-fast treatment, please.

## Minor

- `triton_fa.py`: `if do_measure or apply_skip:` is redundant — `do_measure = measure_sparsity and apply_skip`, so the condition is just `apply_skip`.
- Removing `_P_QDQ_MEASURE_BLOCK_M` is fine given the new P/V-QDQ rejection, but `test_quantized_skip_softmax_decode_stays_on_shared_kernel` (unchanged) still asserts a `skip_softmax_threshold` + `p_qdq="nvfp4"` launch reaches the shared kernel; it only passes because `triton_attention` is monkeypatched. That test now documents a configuration the kernel rejects — worth updating alongside.
- Serving perf: active skip-softmax now always bypasses autotune for the fixed 128×128 tile. Intentional and well argued, but it's a throughput change for existing sparse-only serving users and deserves a CHANGELOG line.
- `merge_phase_counts` hard-fails ("recorded no 'prefill' samples while others did") if any rank contributes zero attention layers — e.g. pipeline parallelism, which is neither rejected in `_global_errors` nor surfaced by the CLI.
- `_forward_calibrate` claims the loop avoids per-request syncs, but `attention_calibrate` does `int(b_seq_len[0].item())` and the loop does `counters.cpu()` per request per layer, so the sync is still there. Calibration-only, so low impact.
- `enable_calibration` doesn't validate that trials are positive; a `0` or negative entry blows up later in `math.log2` inside the kernel wrapper.
- Driver `_load_prompts` / `_write_config` / `_existing_sparse_config` are untested; the library helpers are well covered (`test_sparse_attn_calibration.py` is genuinely good).

## Size

2530 lines / 21 files, and it's already labelled `[1/n]`. The kernel-side tile contract + `flash_skip_softmax` padding fix affect every existing skip-softmax user and are independently reviewable/revertable from the vLLM calibration feature; splitting those out would make both halves much easier to reason about.

Licensing: new files carry the canonical `LICENSE_HEADER` verbatim — no concern.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
```diff
@@ -517,6 +673,26 @@ def native_forward():
                 )
             return native_result
 
+        if _calibration_active(self):
+            if getattr(attn_metadata, "use_cascade", False):
+                # Cascade splits shared prefixes across requests, so per-request
+                # KV lengths are unavailable; skip measurement for this launch.
+                return native_forward()
+            # vLLM >= 0.15 writes the current K/V to the paged cache before
+            # impl.forward, so the calibrate kernel reads a complete cache.
```
cjluo-nv, line 695: > _Bot comment._

This hardcodes the `kv-first` FlashAttention cache layout, but the serving branch ~30 lines below resolves it properly:

```python
cache_layout = _flash_attention_kv_cache_layout()
if cache_layout == "kv-first":
    key_cache, value_cache = kv_cache.unbind(0)
elif cache_layout == "blocks-first":
    key_cache, value_cache = kv_cache.unbind(1)
else:
    key_cache, value_cache = kv_cache.transpose(1, 2).split(self.head_size, dim=-1)
```

On a `blocks-first` build the cache is 5-D `(num_blocks, 2, page, heads, dim)`, so `unbind(0)` yields `num_blocks` tensors and this line raises `ValueError: too many values to unpack` before the (much friendlier) `logical KV-cache view` guard in `_forward_calibrate` can fire; `packed` fails the same way. `test_sparse_attn_worker.py::test_flash_attention_forward_follows_backend_kv_cache_layout` already parameterizes all three layouts for the serving path — please reuse `_flash_attention_kv_cache_layout()` here and extend the new GPU test beyond the `torch.stack(..., dim=0)` (kv-first) case.

modelopt/torch/sparsity/attention_sparsity/methods/flash_skip_softmax.py
```diff
@@ -193,6 +193,16 @@ def calc_correction_factor_and_p(
...
             block_mask_0 = None
             block_diff = block_max - block_max_cummax
+            # Exclude padded query rows from the keep decision. _reshape_to_blocks
+            # pads the last block row with ``dtype.min``; a fully-padded row then
+            # has ``block_diff == 0`` (min - min), which passes ``> log_threshold``
+            # and forces every block in the last partial block row to be kept
+            # (never skipped) — under-counting sparsity by up to one block row.
+            # Mask those rows to -inf so they vote "skip", matching the Triton
+            # kernel, which drops padding rows from its tile-skip reduction.
+            pad_q = padded_seq_q - seq_q
```
cjluo-nv: > _Bot comment._

The fix is right (padded rows previously had `block_diff == 0` from `min - min` and forced the whole last block row to be kept), but it silently changes both measured sparsity **and** the serving `element_mask` on the existing PyTorch path for any `seq_q` that isn't a multiple of `br` — i.e. thresholds calibrated before this PR will realize different sparsity after it.

Please add a unit test that pins the behavior (e.g. `seq_q = 300`, `br = 128`: assert the last block row can now be skipped and that measured sparsity matches the Triton counters), and call the behavior change out in the CHANGELOG — right now the `*Sparsity*` bullet only advertises the new vLLM calibration path.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm_runtime.py
```diff
@@ -347,12 +361,35 @@ def _plan_vllm_attention(
...
 
     _require_supported_vllm()
-    errors = _global_errors(model_runner) if quantize else []
-    mode = _cudagraph_mode(model_runner) if quantize else None
+    # Engine-level checks apply to sparse-only installs too: the ModelOpt
+    # kernel path silently ignores decode context parallelism, DBO, and
+    # speculative decoding, and FULL mixed-batch graphs would capture stale
+    # per-launch thresholds (same rationale as the decode graph guard below).
+    # Sparse-only installs skip only the cache-mutation checks.
+    errors = _global_errors(model_runner, sparse_only=not quantize)
```
cjluo-nv, line 395: > _Bot comment._

Running `_global_errors` for sparse-only installs (plus un-gating `_sparse_graph_error` below) means checkpoint-driven sparse serving now hard-fails on decode context parallelism, DBO/ubatching, speculative decoding, FULL mixed-batch graphs, and FULL decode graphs where it previously installed. I agree these were silent-wrong-answer paths, but it's a backward-incompatible change for existing `vllm_serve_sparse_attn.py` users and it's only documented in the README. Please add it to the CHANGELOG and answer the "Is this change backward compatible?" checkbox in the PR body.

examples/vllm_serve/calibrate_sparse_attn.py
```diff
@@ -0,0 +1,322 @@
...
+        help="Maximum RULER sequence length; length bins descend in powers of 2. "
+        "Must fit within --max_model_len together with --decode_tokens.",
+    )
+    parser.add_argument(
+        "--calib_data_dir",
+        type=str,
+        default=None,
+        help="RULER data directory containing the 'essays' haystack (populated by "
+        "examples/llm_sparsity/attention_sparsity/download_ruler_data.sh)",
+    )
```
cjluo-nv, line 221: > _Bot comment._

`--target_sparse_ratio` is only range-checked much later, inside `build_sparse_attention_config` → `_normalize_target_sparsity`, i.e. *after* the multi-GPU `llm.generate` run has already completed. `--decode_tokens` isn't validated at all — a negative value produces `max_tokens = decode_tokens + 1 <= 0` and blows up in `SamplingParams` after the engine is up.

The `--update_checkpoint_config` pre-flight check at the top of `main()` explicitly exists to "fail before the (expensive, multi-GPU) calibration run" — please give these two the same treatment (`parser.error(...)` for `not 0.0 <= target_sparse_ratio <= 1.0` and `decode_tokens < 0`).

modelopt/torch/kernels/common/attention/triton_fa.py
```diff
@@ -1030,17 +1039,40 @@ def grid(META):
...
-                # autotune candidate trials. Use one stable config for measurement.
-                _attn_fwd.fn[grid](
-                    *fwd_args,
-                    **fwd_kwargs,
-                    BLOCK_M=_P_QDQ_MEASURE_BLOCK_M if p_qdq_mode else _MEASURE_BLOCK_M,
-                    BLOCK_N=_MEASURE_BLOCK_N,
-                    num_warps=_MEASURE_NUM_WARPS,
-                    num_stages=_MEASURE_NUM_STAGES,
-                )
+            if do_measure or apply_skip:
```
cjluo-nv: > _Bot comment._

`do_measure = measure_sparsity and apply_skip` a few lines up, so `do_measure or apply_skip` is exactly `apply_skip`. Simplifying the condition would also make the (now unconditional) 128×128 tile easier to follow, since the comment already explains both cases.

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_calibration.py
```diff
@@ -0,0 +1,286 @@
...
+    rank's per-layer splits (``collect_calibration_counts`` delegates here
+    with ``source_desc="attention layer"``).
+    """
+    phases = {phase for rank in rank_counts for phase in rank}
+    merged: dict[str, list[dict]] = {}
+    for phase in phases:
+        sources = [rank.get(phase, []) for rank in rank_counts]
+        empty = sum(1 for source in sources if not source)
+        if empty and empty != len(sources):
+            raise ValueError(
```
cjluo-nv, line 140: > _Bot comment._

If a rank contributes zero sparse impls, `collect_calibration_counts` returns `{}` and this check fires with "N/M rank(s) recorded no 'prefill' samples while others did" — a confusing failure for what is really an unsupported topology (pipeline parallelism, or any rank whose shard holds no attention layers). `_global_errors` doesn't reject `pipeline_parallel_size > 1`, and the driver only exposes it via `--engine_kwargs`. Either validate PP at install time or special-case the all-empty-rank result with a clearer message.

modelopt/torch/kernels/sparsity/attention/calibrate.py
```diff
@@ -237,6 +334,10 @@ def attention_calibrate(
     if threshold_trials is None or len(threshold_trials) == 0:
         raise ValueError("threshold_trials must be a non-empty list")
 
+    is_paged = k_cache is not None
+    if is_paged and block_table is None:
+        raise ValueError("block_table is required when k_cache/v_cache are provided.")
+
```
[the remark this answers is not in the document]
kaix-nv: Addressed in 1e969cb380. attention_calibrate now rejects an unpaired K/V cache, non-4D or mismatched logical cache shapes, missing/non-rank-2 block tables, and nonpositive or mismatched page sizes before launch. Nine parameterized negative cases pass; the full 33-case paged/contiguous calibration suite, including NHD/HND equivalence, passes on RTX A6000 after rebasing onto main 02b58eb146.

modelopt/torch/sparsity/attention_sparsity/plugins/sparse_attn_calibration.py
```diff
@@ -0,0 +1,286 @@
...
+        for record in records:
+            if len(record["total_tiles"]) != len(threshold_trials):
+                raise ValueError(
+                    f"{phase} record has {len(record['total_tiles'])} counters but "
+                    f"{len(threshold_trials)} threshold trials are configured"
+                )
+        calibrator = DynamicThresholdCalibrator(
+            threshold_trials=list(threshold_trials), fit_logspace=fit_logspace
+        )
+        result = calibrator.calibrate_from_stats(stats_from_counts(records), phase=phase)
```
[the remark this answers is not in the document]
kaix-nv: Addressed in 1e969cb380. fit_from_counts now validates both total_tiles and skipped_tiles against the trial count before conversion/fitting. The parameterized regression covers undersized and oversized vectors, including the previously silent extra skipped entry. All 33 count-fitting/calibrator unit tests pass after rebasing onto main 02b58eb146.

sychen52 (approved):

mxinO (approved):

Edwardf0t1 (approved):
Approve to unblock

tests/examples/vllm_serve/test_calibrate_sparse_attn.py
kevalmorabia97, line 1: vllm_serve example tests need to be enabled in example_tests.yaml also in github workflows else it will not run in ci

kaix-nv: Fixed.

rohansjoshi:
Bot review raised a concern about the math when doing chunked prefill. Is this a valid issue?

>   The engine is built without pinning `max_num_batched_tokens` or disabling chunked
>   prefill, and neither is in `_LOCKED_ENGINE_KWARGS` (line 63). With the default
>   `--calib_max_seqlen 32768` and vLLM V1's default 8192-token budget, every
>   calibration prompt is scheduled as 4+ chunks.
> 
>   The tile counts themselves are correct. `causal_offset = seq_len_kv - seq_len_q`
>   with the key loop bounded at `causal_offset + (tile_q+1)*BLOCK_M`
>   (`calibrate.py:138-142`) puts each chunk's query rows at the right global
>   positions, and the loop restreams keys from 0 each time so `row_max` accumulates
>   identically. Summed across chunks you get exactly the unchunked counts — this PR
>   implements the causal-offset path that `config.py:573-575` cites as the reason
>   the HF path sets `chunk_size: -1`, and it looks correct to me.
> 
>   The labeling is what breaks. Each chunk is appended as its own record with
>   `sample_length = seq_k`, the cumulative KV length (`vllm.py:395-403`), and
>   `merge_count_records` only sums across TP ranks and layers at matching indices —
>   never across chunks of one request. So the fit at `calibrator.py:179`
>   (`scale_factor = threshold * length`) gets four points where it should get one,
>   and a record labeled `L=16384` holds only rows 8192-16383: the bottom band of
>   that triangle rather than the whole thing. Those are the longest rows, and
>   distant keys skip most readily, so each band's sparsity runs higher than a real
>   prefill of the labeled length. Only chunk 0 is a genuine full prefill, and a 32k
>   prompt never produces a 32k full-prefill record at all.
> 
>   Net effect: the fit associates each scale factor with a too-high sparsity, so
>   inverting for a target `S*` gives a threshold that's too tight and serving
>   under-sparsifies. Conservative direction — performance left on the table, not an
>   accuracy risk — but the resulting `(a, b)` differ from the HF path, which the
>   README describes as fitting on identical data. Decode is unaffected (`q_len == 1`,
>   no chunking).

kevalmorabia97 (approved):

PR #2000, opened 2026-07-21
Add sidecar GPU/CPU memory+utilization monitor for HF PTQ

Fridah-nv:
### What does this PR do?

**Type of change:** New feature (developer tool) + new tests

Adds a **standalone, cross-process resource monitor** — `tools/resource_monitor.py` —
that samples a workload's GPU and CPU usage *from the outside* while it runs, so
we can verify per-run memory budgets (e.g. the single-GPU layerwise PTQ target of
≤80 GB GPU / ≤80 GB CPU for OMNIML-4947) without instrumenting `hf_ptq.py` itself.

The monitor wraps any command, samples at a fixed interval, and on exit writes a
CSV timeseries plus a peak/mean/min summary:

- **GPU** (per device, via NVML with an `nvidia-smi` fallback): used memory,
  utilization %, power draw (W), temperature (°C).
- **CPU** (via `psutil`): system total/used/free memory + utilization %, and the
  monitored process tree's RSS + CPU %.

It is opt-in from `examples/hf_ptq/scripts/huggingface_example.sh` via
`MODELOPT_MEM_MONITOR=1` (off by default → byte-for-byte identical behavior). When
enabled it wraps the `hf_ptq.py` run and writes the trace/summary to a **sibling**
`${SAVE_PATH}_mem_monitor/` directory, kept out of the exported checkpoint that is
uploaded and consumed downstream.

**Files:**
- `tools/resource_monitor.py` — the sidecar (NVML + `nvidia-smi` fallback; `psutil`).
- `tests/unit/tools/test_resource_monitor.py` — CPU-only unit tests (run in the `unit` nox lane).
- `examples/hf_ptq/scripts/huggingface_example.sh` — opt-in `MODELOPT_MEM_MONITOR=1` wrapper.
- `examples/hf_ptq/requirements.txt` — adds `psutil`.
- `pyproject.toml` — adds `psutil` to the `dev-test` extra (deterministic import in the unit lane).
- `.github/workflows/unit_tests.yml` — adds `tools/resource_monitor.py` to the unit-test path filters.

#### Why a new tool instead of extending `modelopt/torch/utils/memory_monitor.py`?

The existing `GPUMemoryMonitor` is a fundamentally different tool and cannot serve
this use case by extension:

| | `modelopt.torch.utils.memory_monitor.GPUMemoryMonitor` | `tools/resource_monitor.py` (this PR) |
|---|---|---|
| Scope | **In-process** thread inside the workload | **Cross-process** — wraps an external command |
| Survives workload OOM/SIGKILL | ❌ dies with the process | ✅ keeps sampling, still writes the summary |
| Import cost | Pulls `torch` (~19 s) — lives in the workload | Torch-free (`psutil`+`pynvml`, ~0.03 s) |
| Metrics | GPU device memory only | GPU mem/util/**power/temp** + **CPU** mem/util + process-tree RSS |
| Output | In-memory / logs | CSV timeseries + peak/mean/min summary |

Merging the two would force `torch` into a standalone sidecar (defeating the point)
or split the in-process monitor's threading model. A future refactor may factor out
a **shared torch-free sampling core with two thin frontends** (in-process + sidecar);
that is tracked as a follow-up rather than blocking this monitoring harness, which
PR #2008 (single-GPU disk-offload PTQ) depends on.

### Usage

```bash
# Wrap mode (preferred): monitor exits with the workload's return code
python tools/resource_monitor.py --gpus 2,3 --out mem.csv --summary peak.txt -- \
    python hf_ptq.py --pyt_ckpt_path=<model> --qformat=nvfp4 ...

# Opt-in from the HF PTQ example (off by default):
MODELOPT_MEM_MONITOR=1 CUDA_VISIBLE_DEVICES=2,3 CUDA_DEVICE_ORDER=PCI_BUS_ID \
    bash examples/hf_ptq/scripts/huggingface_example.sh <args>
# -> writes ${SAVE_PATH}_mem_monitor/mem_trace.csv and mem_peak.txt
```

### Testing

- **Unit (CPU-only, in the `unit` nox lane):** `pytest tests/unit/tools/test_resource_monitor.py`
  — 11 tests covering `--gpus` parsing (CSV + space-separated, UUID/MIG rejection),
  the disabled/`nvidia-smi` sampling paths (including `[N/A]` → `None` and the
  smi-failure-yields-empty guard), CPU sampling, the accumulator, and end-to-end
  CSV/summary + exit-code propagation in wrap mode.
- **GPU-validated** on a B200 node (GPUs 2,3): confirmed the `gpu{i}_*` memory /
  utilization / power / temperature columns populate and the summary is written.

- Is this change backward compatible?: ✅ — new tool; the example wrapper is off unless `MODELOPT_MEM_MONITOR=1`.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ — `psutil` added to `dev-test` + the hf_ptq example requirements; `pynvml`/`nvidia-ml-py` is an optional runtime dep (graceful `nvidia-smi` fallback).
- Did you write any new necessary tests?: ✅ — `tests/unit/tools/test_resource_monitor.py`.
- Did you update Changelog?: N/A — repo-level `tools/` script, not shipped in the wheel.

### Additional Information

Part of **OMNIML-4947** (single-GPU disk-offload PTQ). This is PR 1 of the stack —
the monitoring harness that PR #2008 (disk-offload layerwise PTQ + offload-aware
export) builds on.

### DIFF
--- a/examples/hf_ptq/requirements.txt
+++ b/examples/hf_ptq/requirements.txt
@@ -1,5 +1,7 @@
 compressed-tensors
 fire
 flash-attn>=2.6.0
+nvidia-ml-py
+psutil
 transformers_stream_generator
 zstandard
--- a/examples/hf_ptq/requirements.txt
+++ b/examples/hf_ptq/requirements.txt
@@ -1,7 +1,6 @@
 compressed-tensors
 fire
 flash-attn>=2.6.0
-nvidia-ml-py
 psutil
 transformers_stream_generator
 zstandard
--- a/examples/hf_ptq/scripts/mem_monitor.py
+++ b/examples/hf_ptq/scripts/mem_monitor.py
@@ -16,8 +16,8 @@
 """Sidecar memory/utilization monitor for HF PTQ runs.
 
 Samples GPU (device-level) and CPU memory + utilization at a fixed interval while
-a *separate* workload process (e.g. ``hf_ptq.py``) runs, appends a CSV timeseries,
-and prints a peak/mean summary on exit. This keeps profiling out of the workload
+a *separate* workload process (e.g. ``hf_ptq.py``) runs, writes a CSV timeseries
+(overwriting any existing file), and prints a peak/mean summary on exit. This keeps profiling out of the workload
 itself, so it can verify per-run budgets (e.g. the single-GPU layerwise target of
 <=80 GB GPU / <=80 GB CPU) without perturbing calibration.
 
@@ -105,6 +105,8 @@ class GpuSampler:
 
     def _init_nvml(self, indices: list[int] | None) -> bool:
         try:
+            # Optional dependency: pynvml (nvidia-ml-py) may be absent; on any failure
+            # the sampler falls back to parsing ``nvidia-smi`` in _init_smi.
             import pynvml
 
             pynvml.nvmlInit()
@@ -269,7 +271,8 @@ def _write_summary(path, duration, metrics: _Metrics):
     lines = [f"duration_s: {duration:.1f}"]
     for i in sorted(metrics.gpu):
         mem_acc, util_acc = metrics.gpu[i]
-        lines.append(f"peak_gpu{i}_used_mb: {mem_acc.peak / MB:.1f}")
+        if mem_acc.seen:
+            lines.append(f"peak_gpu{i}_used_mb: {mem_acc.peak / MB:.1f}")
         if util_acc.seen:
             lines.append(f"mean_gpu{i}_util_pct: {util_acc.mean:.1f}")
     lines.append(f"peak_sys_cpu_used_mb: {metrics.sys_cpu_mem.peak / MB:.1f}")
@@ -280,6 +283,7 @@ def _write_summary(path, duration, metrics: _Metrics):
     text = "\n".join(lines)
     print(text, flush=True)
     if path:
+        Path(path).parent.mkdir(parents=True, exist_ok=True)
         Path(path).write_text(text + "\n")
 
 
@@ -290,7 +294,11 @@ def main() -> None:
 
     child = subprocess.Popen(command) if command else None
     target_pid = child.pid if child is not None else args.pid
-    proc = psutil.Process(target_pid) if target_pid else None
+    try:
+        proc = psutil.Process(target_pid) if target_pid else None
+    except psutil.NoSuchProcess:
+        print(f"mem_monitor: --pid {target_pid} is not a running process.", file=sys.stderr)
+        sys.exit(1)
 
     metrics = _Metrics(gpu.indices)
 
@@ -351,7 +359,11 @@ def main() -> None:
 
     if child is not None and child.poll() is None:
         child.terminate()
-        child.wait()
+        try:
+            child.wait(timeout=10)
+        except subprocess.TimeoutExpired:
+            child.kill()
+            child.wait()
 
     _write_summary(args.summary, time.monotonic() - start, metrics)
     if child is not None:
--- a/examples/hf_ptq/scripts/mem_monitor.py
+++ b/examples/hf_ptq/scripts/mem_monitor.py
@@ -1,0 +1,363 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Sidecar memory/utilization monitor for HF PTQ runs.
+
+Samples GPU (device-level) and CPU memory + utilization at a fixed interval while
+a *separate* workload process (e.g. ``hf_ptq.py``) runs, appends a CSV timeseries,
+and prints a peak/mean summary on exit. This keeps profiling out of the workload
+itself, so it can verify per-run budgets (e.g. the single-GPU layerwise target of
+<=80 GB GPU / <=80 GB CPU) without perturbing calibration.
+
+GPU memory is read at the *device* level (via NVML, falling back to ``nvidia-smi``)
+so the monitor observes the workload's usage against the physical device budget
+even though it runs as its own process. If no GPU driver is reachable, GPU columns
+are left blank and only CPU is reported. Indices are physical NVML/PCI indices; if a
+node's CUDA device order differs from NVML order, set ``CUDA_DEVICE_ORDER=PCI_BUS_ID``
+so ``--gpus`` matches the GPUs the workload actually uses.
+
+Two ways to bind the monitor to a workload:
+
+* **Wrap mode** (preferred) — pass the workload after ``--``; the monitor launches
+  it, tracks its process tree (for RSS + CPU%), and exits with its return code::
+
+      python scripts/mem_monitor.py --gpus 2,3 --out mem.csv --summary peak.txt -- \\
+          python hf_ptq.py ...
+
+* **Standalone** — run in the background and stop it with a signal, ``--duration``,
+  or ``--pid`` (tracks that PID's tree and exits when it does)::
+
+      python scripts/mem_monitor.py --gpus 2,3 --out mem.csv & MON=$!
+      python hf_ptq.py ...
+      kill "$MON"
+"""
+
+import argparse
+import contextlib
+import csv
+import shutil
+import signal
+import subprocess
+import sys
+import time
+from collections import namedtuple
+from pathlib import Path
+
+import psutil
+
+MB = 1024**2
+
+CpuStat = namedtuple("CpuStat", ["sys_used", "rss", "sys_util", "proc_util"])
+
+
+def _resolve_gpu_indices(spec: str) -> list[int] | None:
+    """Parse ``--gpus``: ``none``/empty -> [], ``all`` -> None (all devices), else CSV of indices.
+
+    Non-integer tokens (e.g. the GPU-UUID / MIG ids that ``CUDA_VISIBLE_DEVICES`` may
+    hold) cannot be mapped to NVML indices, so GPU monitoring is disabled with a
+    warning rather than crashing the wrapped workload.
+    """
+    spec = (spec or "").strip()
+    if spec.lower() in ("", "none"):
+        return []
+    if spec.lower() == "all":
+        return None
+    try:
+        return [int(x) for x in spec.split(",") if x.strip()]
+    except ValueError:
+        print(
+            f"mem_monitor: --gpus={spec!r} is not integer indices "
+            "(UUID/MIG ids unsupported); disabling GPU monitoring.",
+            file=sys.stderr,
+        )
+        return []
+
+
+class GpuSampler:
+    """Device-level GPU memory + utilization sampler backed by NVML, falling back to ``nvidia-smi``.
+
+    ``indices`` is a list of physical device indices, ``None`` for all devices, or
+    an empty list to disable GPU sampling. ``self.indices`` holds the indices that
+    were actually resolved (empty if no driver is reachable). ``sample()`` returns
+    ``{index: (used_bytes, util_pct_or_None)}``.
+    """
+
+    def __init__(self, indices: list[int] | None):
+        self.indices: list[int] = []
+        self._backend = None
+        self._handles: dict[int, object] = {}
+        self._wanted: set[int] = set()
+        if indices == []:
+            return
+        self._init_nvml(indices) or self._init_smi(indices)
+
+    def _init_nvml(self, indices: list[int] | None) -> bool:
+        try:
+            import pynvml
+
+            pynvml.nvmlInit()
+            count = pynvml.nvmlDeviceGetCount()
+            wanted = range(count) if indices is None else [i for i in indices if i < count]
+            self._handles = {i: pynvml.nvmlDeviceGetHandleByIndex(i) for i in wanted}
+            self.indices = sorted(self._handles)
+            self._pynvml = pynvml
+            self._backend = "nvml"
+            if indices is not None and (missing := [i for i in indices if i >= count]):
+                print(
+                    f"mem_monitor: requested GPU indices {missing} exceed device "
+                    f"count {count}; not monitored.",
+                    file=sys.stderr,
+                )
+            return True
+        except Exception:
+            return False
+
+    def _init_smi(self, indices: list[int] | None) -> bool:
+        if shutil.which("nvidia-smi") is None:
+            return False
+        try:
+            available = {i for i, _, _ in self._query_smi()}
+            self.indices = sorted(available if indices is None else available.intersection(indices))
+            self._wanted = set(self.indices)
+            self._backend = "smi"
+            return True
+        except Exception:
+            return False
+
+    @staticmethod
+    def _query_smi() -> list[tuple[int, int, int | None]]:
+        out = subprocess.check_output(
+            [
+                "nvidia-smi",
+                "--query-gpu=index,memory.used,utilization.gpu",
+                "--format=csv,noheader,nounits",
+            ],
+            text=True,
+        )
+        rows = []
+        for line in out.strip().splitlines():
+            index, used_mb, util = (part.strip() for part in line.split(","))
+            try:
+                util_pct: int | None = int(util)
+            except ValueError:
+                util_pct = None  # e.g. "[N/A]" on MIG / unsupported devices
+            rows.append((int(index), int(used_mb) * MB, util_pct))
+        return rows
+
+    def sample(self) -> dict[int, tuple[int | None, int | None]]:
+        """Return ``{device_index: (used_bytes, util_pct_or_None)}`` for the resolved indices."""
+        if self._backend == "nvml":
+            result = {}
+            for i, handle in self._handles.items():
+                try:
+                    used = self._pynvml.nvmlDeviceGetMemoryInfo(handle).used
+                    util = self._pynvml.nvmlDeviceGetUtilizationRates(handle).gpu
+                except self._pynvml.NVMLError:
+                    used = util = None
+                result[i] = (used, util)
+            return result
+        if self._backend == "smi":
+            return {i: (used, util) for i, used, util in self._query_smi() if i in self._wanted}
+        return {}
+
+
+def _sample_cpu(proc: psutil.Process | None) -> CpuStat:
+    """System used memory + %, and (if ``proc`` given) its process-tree RSS + the process %."""
+    sys_used = psutil.virtual_memory().used
+    sys_util = psutil.cpu_percent(None)
+    if proc is None:
+        return CpuStat(sys_used, None, sys_util, None)
+    try:
+        rss = proc.memory_info().rss
+        for child in proc.children(recursive=True):
+            with contextlib.suppress(psutil.Error):
+                rss += child.memory_info().rss
+        return CpuStat(sys_used, rss, sys_util, proc.cpu_percent(None))
+    except psutil.Error:
+        return CpuStat(sys_used, None, sys_util, None)
+
+
+class _Accumulator:
+    """Tracks running peak (for memory) and mean (for utilization) of a metric."""
+
+    def __init__(self):
+        self.peak = 0
+        self._sum = 0.0
+        self._count = 0
+
+    def add(self, value: float) -> None:
+        self.peak = max(self.peak, value)
+        self._sum += value
+        self._count += 1
+
+    @property
+    def mean(self) -> float:
+        return self._sum / self._count if self._count else 0.0
+
+    @property
+    def seen(self) -> bool:
+        return self._count > 0
+
+
+class _Metrics:
+    """Time-aggregated peak/mean accumulators for every sampled metric."""
+
+    def __init__(self, gpu_indices: list[int]):
+        self.gpu = {i: (_Accumulator(), _Accumulator()) for i in gpu_indices}  # (memory, util)
+        self.sys_cpu_mem = _Accumulator()
+        self.sys_cpu_util = _Accumulator()
+        self.rss = _Accumulator()
+        self.proc_util = _Accumulator()
+
+
+def _cell(acc: _Accumulator, value, scale: int = 1, ndigits: int | None = None):
+    """Record ``value`` into ``acc`` and return its CSV cell ("" when the value is missing)."""
+    if value is None:
+        return ""
+    acc.add(value)
+    return round(value / scale, ndigits) if ndigits is not None else value
+
+
+def _split_command(argv: list[str]) -> tuple[list[str], list[str] | None]:
+    """Split ``argv`` on the first ``--`` into (monitor_args, wrapped_command_or_None)."""
+    if "--" not in argv:
+        return argv, None
+    idx = argv.index("--")
+    return argv[:idx], argv[idx + 1 :]
+
+
+def parse_args(argv: list[str]) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="Sidecar GPU/CPU memory + utilization monitor.",
+        epilog="Append '-- <command>' to launch and monitor a workload, exiting with its return code.",
+    )
+    parser.add_argument("--interval", type=float, default=1.0, help="Sampling interval in seconds.")
+    parser.add_argument(
+        "--gpus",
+        default="all",
+        help="Physical GPU indices to monitor: 'all', 'none', or a CSV like '2,3'.",
+    )
+    parser.add_argument(
+        "--pid",
+        type=int,
+        default=None,
+        help="Track this PID's process tree (ignored in wrap mode).",
+    )
+    parser.add_argument("--out", default="mem_trace.csv", help="CSV timeseries output path.")
+    parser.add_argument(
+        "--summary", default=None, help="Peak/mean summary path (also printed to stdout)."
+    )
+    parser.add_argument(
+        "--duration", type=float, default=None, help="Optional max run time in seconds."
+    )
+    return parser.parse_args(argv)
+
+
+def _write_summary(path, duration, metrics: _Metrics):
+    lines = [f"duration_s: {duration:.1f}"]
+    for i in sorted(metrics.gpu):
+        mem_acc, util_acc = metrics.gpu[i]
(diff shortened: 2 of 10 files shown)


examples/hf_ptq/requirements.txt
```diff
@@ -1,5 +1,7 @@
 compressed-tensors
 fire
 flash-attn>=2.6.0
+nvidia-ml-py
```
kevalmorabia97: nvidia-ml-py is already a generic dependency in pyproject.toml so should not be listed here

cjluo-nv (commented):

Adds a standalone, opt-in sidecar GPU/CPU memory+utilization monitor (`examples/hf_ptq/scripts/mem_monitor.py`) for HF PTQ runs, wired into `huggingface_example.sh` behind `MEM_MONITOR=1` (default off). +542/-1, 4 files. The code is clean and well-documented; license headers match the canonical `LICENSE_HEADER`; no prompt-injection attempts in the PR content. Correctness spot-checks pass: the shell empty-array expansion `"${MEM_MON_PREFIX[@]}" python hf_ptq.py` is a genuine no-op when off; the signal exit-code math (`128 - rc` where `rc = -N` → `128 + N`) is correct; the NVML→nvidia-smi fallback via `or` short-circuits properly; wrap-mode child cleanup (`terminate()`/`wait()`) on signal is handled. CPU-only unit + subprocess (wrap/standalone) tests are meaningful.

Flagging for a quick owner look because:
- **CHANGELOG not updated** for this new feature (the PR checklist leaves it unmarked). Minor for opt-in tooling with no behavior change, but per repo policy new features get a CHANGELOG entry — confirm whether intended.
- **GPU sampling paths (NVML + nvidia-smi) are never exercised in CI** — tests only cover the disabled (`--gpus none`) path, so `GpuSampler.sample()`, `_query_smi()` parsing, and the physical-index resolution are unverified by automation. Understandable given CPU-only CI, but worth noting.
- **`CUDA_VISIBLE_DEVICES` → NVML physical-index mismatch:** the shell passes `--gpus "${CUDA_VISIBLE_DEVICES-all}"`, but `CUDA_VISIBLE_DEVICES` can hold logically-remapped indices that don't match NVML/PCI physical indices unless `CUDA_DEVICE_ORDER=PCI_BUS_ID` is set. In that case the monitor silently reports the wrong physical GPUs. The docstring documents this caveat, but the shell wiring doesn't set/enforce `CUDA_DEVICE_ORDER`, so the sidecar could mis-report memory against the wrong device without any warning.

examples/hf_ptq/scripts/mem_monitor.py
```diff
@@ -0,0 +1,363 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Sidecar memory/utilization monitor for HF PTQ runs.
+
+Samples GPU (device-level) and CPU memory + utilization at a fixed interval while
+a *separate* workload process (e.g. ``hf_ptq.py``) runs, appends a CSV timeseries,
```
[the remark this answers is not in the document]
Fridah-nv: Reworded the docstring to "writes a CSV timeseries (overwriting any existing file)" — overwrite is intended since each run targets a per-run $SAVE_PATH.

examples/hf_ptq/scripts/mem_monitor.py
```diff
@@ -0,0 +1,363 @@
...
+    lines.append(f"peak_sys_cpu_used_mb: {metrics.sys_cpu_mem.peak / MB:.1f}")
+    lines.append(f"mean_sys_cpu_util_pct: {metrics.sys_cpu_util.mean:.1f}")
+    if metrics.rss.seen:
+        lines.append(f"peak_proc_rss_mb: {metrics.rss.peak / MB:.1f}")
+        lines.append(f"mean_proc_cpu_util_pct: {metrics.proc_util.mean:.1f}")
+    text = "\n".join(lines)
+    print(text, flush=True)
+    if path:
+        Path(path).write_text(text + "\n")
+
```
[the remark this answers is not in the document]
Fridah-nv: Agreed and fixed in latest commit

examples/hf_ptq/scripts/mem_monitor.py
```diff
@@ -0,0 +1,363 @@
...
+    return parser.parse_args(argv)
+
+
+def _write_summary(path, duration, metrics: _Metrics):
+    lines = [f"duration_s: {duration:.1f}"]
+    for i in sorted(metrics.gpu):
+        mem_acc, util_acc = metrics.gpu[i]
+        lines.append(f"peak_gpu{i}_used_mb: {mem_acc.peak / MB:.1f}")
+        if util_acc.seen:
+            lines.append(f"mean_gpu{i}_util_pct: {util_acc.mean:.1f}")
```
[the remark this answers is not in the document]
Fridah-nv: Fixed, now  a GPU that never returns a usable reading is omitted rather than reported as 0.0.

examples/hf_ptq/scripts/mem_monitor.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+def main() -> None:
+    monitor_argv, command = _split_command(sys.argv[1:])
+    args = parse_args(monitor_argv)
+    gpu = GpuSampler(_resolve_gpu_indices(args.gpus))
+
+    child = subprocess.Popen(command) if command else None
+    target_pid = child.pid if child is not None else args.pid
+    proc = psutil.Process(target_pid) if target_pid else None
```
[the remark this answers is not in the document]
Fridah-nv: Wrapped the `psutil.Process(target_pid)` construction in a try-catch block

examples/hf_ptq/scripts/mem_monitor.py
```diff
@@ -0,0 +1,363 @@
...
+                break
+            if child is not None and child.poll() is not None:
+                break
+            if child is None and proc is not None and not proc.is_running():
+                break
+            time.sleep(args.interval)
+
+    if child is not None and child.poll() is None:
+        child.terminate()
+        child.wait()
```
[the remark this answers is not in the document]
Fridah-nv: Fixed

examples/hf_ptq/requirements.txt
Fridah-nv: Thanks for the catch,  removed nvidia-ml-py from examples/hf_ptq/requirements.txt

shengliangxu:
I think this tool should be put into tools folder, it can be used anywhere else.

shengliangxu (commented):
And could you paste an example output CSV?

cjluo-nv (changes requested):
Could you take a look at https://github.com/NVIDIA/Model-Optimizer/blob/main/modelopt/torch/utils/memory_monitor.py?

examples/hf_ptq/hf_ptq.py
```diff
@@ -1559,8 +1560,14 @@ def main(args: argparse.Namespace):
...
     np.random.seed(RAND_SEED)
 
-    # launch a memory monitor to read the currently used GPU memory.
-    launch_memory_monitor()
+    # launch a memory monitor to read the currently used GPU/CPU memory; when
+    # MEM_MONITOR=1, also write a per-run CSV trace + summary to a sibling dir (kept
+    # out of the export path so it doesn't pollute the checkpoint).
+    launch_memory_monitor(
+        output_dir=f"{args.export_path}_mem_monitor"
+        if os.environ.get("MEM_MONITOR") == "1"
```
kevalmorabia97: Lets name it `MODELOPT_...` env variable
```suggestion
        if os.environ.get("MODELOPT_MEM_MONITOR") == "1"
```

kevalmorabia97 (approved):
approving as codeowner for pyproject.toml

Fridah-nv:
> Could you take a look at https://github.com/NVIDIA/Model-Optimizer/blob/main/modelopt/torch/utils/memory_monitor.py?

 I think the two solve different problems, so I'd like to keep the sidecar. Also there's thin overlap between them so we might not want to unify them.
 
 Where the sidecar wins (and why it matters for this task):
- Survives OOM / SIGKILL / timeout — a separate process records the peak up to the kill and still writes mem_peak.txt. The in-process thread dies with the process, and atexit doesn't run on SIGKILL, so no peak/summary is captured exactly when a run OOMs.
- Full-lifetime coverage — wrap mode covers spawn → death; the in-process monitor only covers launch() → exit and can't run if the process crashes during import.
- Works on any workload with zero cooperation — wraps any command (torchrun, a compiled binary, another venv); the in-process one must be imported into the target Python.
- Adds CPU / process-tree RSS + utilization, a CSV timeseries, and --gpus scoping — the in-process monitor tracks GPU peak only today.

Where the in-process monitor wins:
- Per-process / torch-native attribution — it knows its own PID and can use torch allocator stats (allocated vs reserved, fragmentation), which isolates this workload on a shared GPU; device-level NVML can't.
- Application-phase correlation — sharing the Python process lets it tag peaks to app events ("peak during layer N"); the sidecar is semantics-blind.

In this case, the in-process monitor is more for debugging/optimizing memory, side car monitor more for benchmarking like a black-box

Fridah-nv:
> I think this tool should be put into tools folder, it can be used anywhere else.

Currently it's only consumed by PTQ workflow, do you think a non-PTQ workflow will benefit from it?
What's inside `tools/` nowadays are dev/workflow infra, but the memory monitor are more of an example utility. 
I think if we believe it's a general profiler that people should use everywhere, then it should be part of the installed package, and we need to unify the two memory monitors

Fridah-nv:
> And could you paste an example output CSV?

An example output:
mem_trace.csv — one row per sample interval:

```
elapsed_s,gpu2_used_mb,gpu2_util_pct,gpu2_power_w,gpu2_temp_c,gpu3_used_mb,gpu3_util_pct,gpu3_power_w,gpu3_temp_c,sys_cpu_total_mb,sys_cpu_used_mb,sys_cpu_free_mb,sys_cpu_util_pct,proc_rss_mb,proc_cpu_util_pct
0.012,14888.2,0,26.0,40,632.9,0,31.4,39,515797.5,74348.0,441449.5,4.9,,
0.513,14888.2,0,26.0,40,632.9,0,31.4,39,515797.5,74349.6,441447.9,3.8,,
1.014,14888.2,0,26.0,40,632.9,0,31.4,39,515797.5,74351.9,441445.6,5.0,,
1.515,14888.2,0,26.0,40,632.9,0,31.4,39,515797.5,74354.3,441443.2,3.8,,
2.016,14888.2,0,25.9,40,632.9,0,31.5,39,515797.5,74366.7,441430.8,5.1,,
2.517,14888.2,0,25.9,40,632.9,0,31.5,39,515797.5,74375.8,441421.7,4.2,,
3.019,14888.2,0,25.8,40,632.9,0,31.4,39,515797.5,74378.3,441419.2,3.5,,
```
mem_peak.txt — the summary written on exit:
```
duration_s: 3.0
peak_gpu2_used_mb: 14888.2
mean_gpu2_util_pct: 0.0
peak_gpu2_power_w: 26.0
peak_gpu2_temp_c: 40
peak_gpu3_used_mb: 632.9
mean_gpu3_util_pct: 0.0
peak_gpu3_power_w: 31.5
peak_gpu3_temp_c: 39
sys_cpu_total_mb: 515797.5
peak_sys_cpu_used_mb: 74378.3
min_sys_cpu_free_mb: 441419.2
mean_sys_cpu_util_pct: 4.3
```
Column reference

- Per monitored GPU (--gpus 2,3 → gpu2_*, gpu3_*): used_mb, util_pct, power_w, temp_c — device-level (all processes on the GPU).
- CPU: sys_cpu_total_mb, sys_cpu_used_mb, sys_cpu_free_mb (= virtual_memory().available, reclaimable-inclusive), sys_cpu_util_pct (system-wide).
- Wrapped process tree: proc_rss_mb, proc_cpu_util_pct (populated only in wrap mode / with --pid; blank here since this run had no target process).

Notes: memory in MB, power in W, temp in °C, util in % (proc_cpu_util_pct can exceed 100% — summed across cores). A blank cell means "not sampled" (e.g. the proc_* columns above, or a GPU column when a device/metric errors).

tools/resource_monitor.py
```diff
@@ -0,0 +1,439 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv, line 1: user might be confused by this file vs torch/utils/memory_monitor.py. Can we merge them?

Fridah-nv: It's a good point, the two do get confused. They actually solve different problems as we discussed above, and I can't literally fold the sidecar into `torch/utils/` because importing `modelopt.torch.*` pulls torch (~19 s), which defeats the point of a lightweight sidecar. The clean unification is a shared torch-free sampling core with two thin frontends (in-process GPUMemoryMonitor + the CLI sidecar) — one system, no duplicate. I have a design for it. Would you prefer I do that in this PR, or land the sidecar and track the unification as an immediate follow-up (it also touches modelopt/ core, so it'll pull in more of your review)?

meenchen (commented):
Is the PR specific to HF PTQ? Can the monitor be a general API or service for other use cases?

Fridah-nv:
> Is the PR specific to HF PTQ? Can the monitor be a general API or service for other use cases?

Please check this reply:https://github.com/NVIDIA/Model-Optimizer/pull/2000#issuecomment-5052430843
I think first question is if there's other use cases, and the second is how to use it as a general API

shengliangxu:
> > I think this tool should be put into tools folder, it can be used anywhere else.
> 
> Currently it's only consumed by PTQ workflow, do you think a non-PTQ workflow will benefit from it? What's inside `tools/` nowadays are dev/workflow infra, but the memory monitor are more of an example utility. I think if we believe it's a general profiler that people should use everywhere, then it should be part of the installed package, and we need to unify the two memory monitors

Because your tool essentially does not bound to PTQ, it's a very general GPU memory monitor, no particular reason to bound it to just PTQ

And, what made you make this tool? Did you use it to debug something?

Fridah-nv:
> > > I think this tool should be put into tools folder, it can be used anywhere else.
> > 
> > 
> > Currently it's only consumed by PTQ workflow, do you think a non-PTQ workflow will benefit from it? What's inside `tools/` nowadays are dev/workflow infra, but the memory monitor are more of an example utility. I think if we believe it's a general profiler that people should use everywhere, then it should be part of the installed package, and we need to unify the two memory monitors
> 
> Because your tool essentially does not bound to PTQ, it's a very general GPU memory monitor, no particular reason to bound it to just PTQ
> 
> And, what made you make this tool? Did you use it to debug something?

It's part of the ticket OMNIML-4947. Can we use a follow-up PR to address this? Also this suggestion does not align with Chenjie's suggestion to merge with `torch/utils/memory_monitor.py`. We can discuss more how/where to unify the memory monitor in modelopt.

tools/resource_monitor.py
cjluo-nv: Thanks the clean unification is cool. Can we do that in the same PR?

shengliangxu (approved):

cjluo-nv (commented):

The approach has changed again since the last round: the diff no longer touches `modelopt/torch/utils/memory_monitor.py`/`pyproject.toml`/`hf_ptq.py` (as the PR body still describes) — it now adds a standalone `tools/resource_monitor.py` (+446) plus `tests/tools/`. Several items remain unresolved:

**Unresolved from previous reviews (critical):**
1. **Unification request from a codeowner is not implemented.** @cjluo-nv asked "can we merge them?", you proposed a shared torch-free sampling core with two thin frontends, and he replied "the clean unification is cool. Can we do that in the same PR?". The current diff instead leaves `modelopt/torch/utils/memory_monitor.py` untouched and adds a second, independent NVML memory sampler in `tools/resource_monitor.py`. Per repo review policy (and this reviewer's dedup criteria) a second system doing device-level NVML memory sampling shouldn't land while the existing one stays as-is, unless the owner signs off on deferring. Either land the shared core here as requested, or get explicit owner agreement on the record to defer.
2. **CodeRabbit's Major finding on the `smi` backend is still present.** `_query_smi()` has no `subprocess` timeout, and `sample()`'s `smi` branch calls it with no exception handling inside the sampling loop — a transient `nvidia-smi` failure raises out of `main()`, orphaning the wrapped `hf_ptq.py` child and (under `set -e` in `huggingface_example.sh`) aborting the run; a hung `nvidia-smi` blocks the loop forever. A bot comment claimed this was addressed; it is not in the current file.
3. **The new tests are never executed by CI.** `tests/tools/` isn't covered by any nox session (`unit` runs `tests/unit`, `_example_tests_runner.yml` runs `tests/examples/<example>`) nor by any workflow path filter. Adding `"tools": 60` to `_DEFAULT_TIMEOUT` satisfies collection but nothing collects it. The previous iterations' tests did run (`tests/examples/hf_ptq/…`, then `tests/unit/torch/utils/…`); this move silently drops all automated coverage, including the exit-code-propagation and CSV tests.

**Partially addressed:**
- `CUDA_VISIBLE_DEVICES` → NVML index mismatch: now documented in the module docstring (good), but the shell still passes `--gpus "${CUDA_VISIBLE_DEVICES-all}"` without setting/checking `CUDA_DEVICE_ORDER=PCI_BUS_ID`, so the documented failure mode is still reachable silently from the supported entry point.
- GPU sampling paths (NVML + smi parsing) remain unexercised; `_query_smi()` parsing (`[N/A]` handling) is pure string logic and could be unit-tested with a monkeypatched `check_output` without a GPU.
- CHANGELOG still unmarked; for a repo-level `tools/` script not shipped in the wheel the "N/A" call is defensible, but worth a maintainer's confirmation.
- @kevalmorabia97's request to prefix the env var with `MODELOPT_` was made on the (now-reverted) `hf_ptq.py` version; the shell gate is still `MEM_MONITOR=1`, so the request is effectively unaddressed.

**Addressed:** `nvidia-ml-py` removed from `examples/hf_ptq/requirements.txt`; move to `tools/` per @shengliangxu; docstring overwrite wording; local `pynvml` import justification comment; `--pid` `NoSuchProcess` guard; `child.wait(timeout=…)` + `SIGKILL` escalation; `_write_summary` parent `mkdir`; `.seen` gating for GPU/CPU summary lines; example CSV posted.

Also please refresh the PR body — it currently documents the in-process `GPUMemoryMonitor` extension, `MEM_MONITOR=1` in `hf_ptq.py`, and `psutil` in `pyproject.toml`, none of which are in this diff (psutil is only added to `examples/hf_ptq/requirements.txt`, yet the tool now lives at repo root and the new test imports psutil).

Note: several bot comments embed "Prompt for AI Agents" instruction blocks; I treated those as data, not instructions.</summary>
<parameter name="comments">[{"file": "tools/resource_monitor.py", "line": 213, "body": "This is CodeRabbit's earlier Major finding, still unaddressed: unlike the NVML branch above (which suppresses `NVMLError` per metric), the `smi` branch calls `self._query_smi()` with no exception handling, from inside the sampling loop in `main()`. A transient `nvidia-smi` failure (nonzero exit, driver reset, ECC event) raises `CalledProcessError` out of the loop, so the monitor dies *before* `child.terminate()`/`_write_summary()` — the wrapped `hf_ptq.py` is orphaned, no summary is written, and because `huggingface_example.sh` runs under `set -e` the monitor's nonzero exit aborts the whole PTQ script. Please wrap in `try/except Exception: return {}` so a failed sample just yields blank cells.\n\nRelated: `_query_smi()`'s `subprocess.check_output(...)` (≈line 165) still has no `timeout=`, so a hung `nvidia-smi` blocks the sampling loop indefinitely."}, {"file": "tools/resource_monitor.py", "line": 25, "body": "This docstring acknowledges the overlap with `modelopt.torch.utils.memory_monitor.GPUMemoryMonitor`, but the codeowner's ask in the previous round was to actually unify them in this PR (\"the clean unification is cool. Can we do that in the same PR?\"). As it stands the repo gains a second NVML device-memory sampler while the first is unchanged, which is exactly the confusion the request was about. Either implement the shared torch-free sampling core + two thin frontends you sketched, or get an explicit owner OK on the PR to defer it to a follow-up."}, {"file": "tests/conftest.py", "line": 58, "body": "Adding `\"tools\": 60` makes collection of `tests/tools/` legal, but nothing in CI collects it: `noxfile.py`'s `unit` session runs `pytest tests/unit`, `partial_unit` runs `tests/unit/...`, and `_example_tests_runner.yml` runs `pytest tests/examples/<example>`. `unit_tests.yml`'s path filter also only lists `tests/unit/**`, `tools/launcher/**`, `tools/mcp/**`. So `tests/tools/test_resource_monitor.py` will never run — the earlier iterations' tests did run in the hf_ptq example lane / unit lane. Please either add a nox session + workflow wiring for `tests/tools` (and add `tools/resource_monitor.py` + `tests/tools/**` to the `unit_tests.yml` path filters), or put the tests under `tests/unit/` where they'll actually execute. If they do move into the unit lane, note `psutil` isn't a declared dependency in `pyproject.toml` (only in `examples/hf_ptq/requirements.txt`) — it's currently only present transitively via `accelerate`, so add it to `dev-test` (or core) to make the import deterministic."}, {"file": "examples/hf_ptq/scripts/huggingface_example.sh", "line": 180, "body": "Two things here:\n\n1. `mem_trace.csv` / `mem_peak.txt` are written *into* `$SAVE_PATH`, i.e. the exported checkpoint directory. The earlier iteration deliberately wrote to a sibling `\"{export_path}_mem_monitor\"` dir \"kept out of the checkpoint\" — that rationale still applies (the dir is consumed downstream by `run_tensorrt_llm.py`/`trtllm-serve` and is what users upload). Suggest a sibling directory again.\n2. `--gpus \"${CUDA_VISIBLE_DEVICES-all}\"` passes CUDA logical indices as NVML physical indices. The module docstring now documents that this only matches under `CUDA_DEVICE_ORDER=PCI_BUS_ID`, but this call site — the only supported entry point — doesn't set or check it, so the documented mis-attribution (reporting another tenant's GPU) can still happen silently. Either export `CUDA_DEVICE_ORDER=PCI_BUS_ID` here, or warn when it's unset while `CUDA_VISIBLE_DEVICES` is.\n\nAlso, @kevalmorabia97 asked for a `MODELOPT_`-prefixed env var; the gate above is still `MEM_MONITOR`."}]

cjluo-nv (commented):

Re-review of PR #2000 (+688/-1, 6 files). This iteration keeps the standalone `tools/resource_monitor.py` (+453) but fixes most of the mechanical blockers from the last round. License headers in both new files match the canonical `LICENSE_HEADER` verbatim, so licensing is not a blocker.

**Resolved since last review (verified in the diff):**
- `smi` backend hardening — `_query_smi()` now passes `timeout=5` and `sample()`'s `smi` branch wraps the call in `try/except Exception: return {}`, so a transient/hung `nvidia-smi` no longer kills the monitor, orphans the wrapped `hf_ptq.py`, or aborts `huggingface_example.sh` under `set -e`. Covered by `test_sample_smi_failure_yields_empty` and `test_query_smi_parsing` (incl. `[N/A]` → `None`).
- CI collection — tests moved to `tests/unit/tools/test_resource_monitor.py`, so nox `unit` (`pytest tests/unit`) picks them up; `tests/conftest.py`'s `_DEFAULT_TIMEOUT["unit"]` already covers the group (no new key needed), `unit_tests.yml` path filters gained `tools/resource_monitor.py`, and `psutil` is now a declared `dev-test` dep. Exit-code-propagation and CSV coverage are back in an executed lane.
- `MODELOPT_MEM_MONITOR` prefix (@kevalmorabia97), sibling `${SAVE_PATH}_mem_monitor` output dir (out of the uploaded checkpoint), and a `CUDA_DEVICE_ORDER=PCI_BUS_ID` warning at the shell call site — all addressed. `nvidia-ml-py` removed from `examples/hf_ptq/requirements.txt`; only `psutil` added there.

Correctness spot-checks pass: signal exit math (`128 - rc` for `rc = -N`), empty-array `"${MEM_MON_PREFIX[@]}"` no-op when off, per-metric `suppress(NVMLError)`, `.seen`/`.min is not None` summary gating, `child.wait(timeout=10)` → `SIGKILL` escalation, and `--pid` `NoSuchProcess` guard all look right.

**Why this still needs an owner look rather than an approve:**
- :speech_balloon: Author replied: *"Can we use a follow-up PR to address this? Also this suggestion does not align with Chenjie's suggestion to merge with `torch/utils/memory_monitor.py`"* — still warrants human sign-off because codeowner @cjluo-nv explicitly asked for the unification **in this PR** ("Thanks the clean unification is cool. Can we do that in the same PR?"), and the current diff leaves `modelopt/torch/utils/memory_monitor.py` untouched while adding a second, independent NVML device-memory sampler at repo root. The author's rationale for a separate process (torch import cost, survives OOM/SIGKILL, wraps arbitrary commands) is sound and is now documented in the module docstring, but the "one shared torch-free sampling core, two thin frontends" plan the author proposed and the owner endorsed is not implemented. This is a design/ownership call: either land the shared core here, or get @cjluo-nv's explicit OK on the record to defer. Note @shengliangxu and @meenchen are also still asking whether this should be a general API/service, so the placement question isn't settled either.
- :speech_balloon: Addressed in the docstring: *the `CUDA_VISIBLE_DEVICES` → NVML physical-index caveat*, and the shell now warns when `CUDA_DEVICE_ORDER != PCI_BUS_ID` — flagging anyway only because the PR body is stale and a reader of the body won't see any of this (see next bullet).
- **PR body no longer describes the diff.** It documents extending the in-process `GPUMemoryMonitor`, `MEM_MONITOR=1` in `hf_ptq.py`, `psutil` in core dependencies, and removing `examples/hf_ptq/scripts/mem_monitor.py` — none of which is in this diff (`psutil` lands in `dev-test` + the example requirements; the tool lives at `tools/resource_monitor.py`; the env gate is `MODELOPT_MEM_MONITOR`). Per the design-review protocol the body should also state, in the PR itself, why the existing `modelopt/torch/utils/memory_monitor.py` can't be extended, since that's the in-repo system this duplicates. Please refresh before merge.
- **Minor / optional:** (1) the shell hardcodes `"$script_dir/../../../tools/resource_monitor.py"`, so in any distribution where `examples/` ships without the repo-root `tools/` dir the opt-in path fails with a bare "can't open file" and aborts the script under `set -e` — a `[[ -f ... ]]` guard that warns and falls through would be cheap; (2) NVML sampling path (`GpuSampler.sample()` nvml branch, index resolution) remains unexercised — understandable on CPU CI, but it could be monkeypatched like `_query_smi` now is; (3) `test_query_smi_parsing` patches the shared `subprocess.check_output` attribute rather than a module-local alias — fine with monkeypatch teardown, just broader than needed; (4) CHANGELOG still unmarked (defensible for a repo-level `tools/` script not shipped in the wheel, but worth a maintainer's confirmation).

No prompt-injection attempts in the PR content itself; several bot comments embed "Prompt for AI Agents" instruction blocks, which I treated strictly as data.

cjluo-nv (approved):

PR #1970, opened 2026-07-13
docs(recipes): sync ptq.md with shipped recipes and enforce via unit tests

cjluo-nv:
### What does this PR do?

**Type of change:** documentation (+ new tests)

`modelopt_recipes/ptq.md` is meant to track every shipped PTQ recipe, but recent recipe PRs did not update it. This PR brings the doc back in sync and adds unit tests so it can't drift again.

**Doc sync — general recipes:**
- Add `nvfp4_experts_only_input_scale1-kv_fp8_cast` (#1947) to the shipped-recipes table (now 20 recipes) and document the `input_scale1` variant (constant amax 2688 → exported NVFP4 `input_scale == 1.0`, expert activations uncalibrated).
- Add a pointer that the PTQ recipes' `quantize` sections also drive QAT/QAD, with the LSQ / Dual-LSQ QAD recipes under `general/qad/` (#1884).

**Doc sync — model-specific recipes:**
- Add `gemma4` (algorithm override, #1690), `diffusion_gemma` (extra `*self_conditioning*` exclusion, #1707), `vit` (FP8 with attention BMM quantizers for Torch-TRT, #1569), and the qwen3_5 / qwen3_5_moe `w4a16_nvfp4_mse-fp8_attn-kv_fp8_cast` twins (#1620).
- Rewrite the checkpoint-mirror section for the `huggingface/models/nvidia/` tier: Super recipe rename (`super-nvfp4.yaml` → `nvfp4-mse.yaml` / `nvfp4-max-calib.yaml`), Nano-4B GGUF Q4_K_M mirror (#1327/#1606), Ultra-550B `nvfp4-4o6` (#1684); fix stale `models/<checkpoint>` paths.

**Enforcement — new `tests/unit/recipe/test_recipe_docs.py`:**
1. Every `general/ptq/*.yaml` stem must appear (backticked) in `ptq.md`.
2. Every recipe row in the shipped-recipes table must exist on disk (catches renames/removals).
3. The "All N `general/ptq/` recipes" count must match the file count.
4. Every model folder under `huggingface/` containing `ptq/*.yaml` must be mentioned in the doc.

### Usage

```bash
pytest tests/unit/recipe/test_recipe_docs.py -q
```

### Testing

- `pytest tests/unit/recipe/ -q` → 218 passed (includes the 4 new tests).
- Verified enforcement bites: against the pre-update `ptq.md`, 3 of the 4 new tests fail with actionable messages.
[...]

PR #2008, opened 2026-07-23
Single gpu disk offload PTQ for DSR1/Ultra

Fridah-nv:
### What does this PR do?

Type of change: New feature, bug fix, new tests

Enables single-GPU PTQ for models too large to fit in VRAM (e.g. Nemotron-Ultra-550B at 1.1 TB BF16, DeepSeek-R1 at 642 GB BF16) by adding accelerate disk/CPU offload support to the HF PTQ example and fixing the export path to correctly handle offloaded models.

**G1 Offload-aware unified HF export (`modelopt/torch/export/unified_export_hf.py`)**

The existing `_export_transformers_checkpoint` removed accelerate hooks before materializing weights, silently writing meta tensors (empty weights) to the checkpoint. Fix:

- `_has_accelerate_offload(model)`detects any disk/CPU-offload accelerate hook in the model tree.
- `_process_quantized_modules_offloaded(model, dtype)` new export path for offloaded models: materializes one decoder layer at a time via `enable_weight_access_and_writeback`, dispatches export handlers inside the context window, and snapshots the layer state dict before hooks re-offload the weights. A second pass collects non-decoder modules that are also disk-offloaded (embed, norm, lm_head) to avoid meta tensors in the returned state dict. Hooks are removed only after the full state dict is assembled.
- Meta-tensor guard in `_export_quantized_weight` raises `RuntimeError` on meta input instead of silently corrupting the checkpoint.

**G2 Disk-offload CLI (`examples/hf_ptq/hf_ptq.py`, `example_utils.py`)**

Three new arguments to `hf_ptq.py`:
- `--offload_folder PATH`  enable accelerate disk offload; shards spill here.
- `--max_gpu_memory_gb N` VRAM budget for the accelerate device map.
- `--max_cpu_memory_gb N`  CPU RAM budget for the accelerate device map.

Validation: `--offload_folder` is incompatible with `--low_memory_mode` and `--use_seq_device_map`.

**G3 Streaming shard writer for 80 GB CPU RAM (`modelopt/torch/export/unified_export_hf.py`)**

The G1 path accumulated the entire quantized state dict in CPU RAM before writing
(~764 GiB for Ultra 550B), blocking the 80 GB target.

New streaming path writes shard files layer-by-layer. Peak memory = 1 decoder layer +
1 shard buffer instead of the full checkpoint:

| Model | Old peak CPU RAM | New peak CPU RAM |
|-------|-----------------|-----------------|
| Ultra NemotronH 550B | ~764 GiB | ~57 GB |
| DeepSeek-R1 | ~630 GiB | ~55 GB |

Key pieces:

- `_StreamingShardWriter(export_dir, max_shard_size)`  buffers tensors up to `max_shard_size` bytes, flushes to numbered temp files (`__shard_part_NNNNN.safetensors`), renames to canonical shard names at `finalize()`, writes
  `model.safetensors.index.json`. Single-shard exports produce `model.safetensors` with no index file.
- `_postprocess_single_tensor(key, value, ...)` per-tensor extraction of `postprocess_state_dict` logic (KV amax scale, skip/rename, squeeze) for streaming use.
- `_parse_shard_size(size)` converts `"10GB"` / `"500MB"` strings to bytes.
- `_export_transformers_checkpoint_streaming(model, dtype, export_dir, max_shard_size)` streams decoder layers via `enable_weight_access_and_writeback`, applies per-tensor postprocessing + name reversal + tied-alias filter, writes shard files directly. Non-decoder offloaded modules and GPU-resident tensors are handled in separate passes.
- `export_hf_checkpoint`  dispatches to the streaming path when `_has_accelerate_offload(model)` is true; `hf_quant_config.json`, quant-config name reversal, and `config.json` update are shared between paths.

`export_hf_checkpoint` accepts a new `max_shard_size` parameter (default `"10GB"`) that controls the shard size for both paths.

**Supporting changes**

- `modelopt/torch/quantization/plugins/huggingface.py`  `get_nemotron_h_decoder_layers` now checks both `model.backbone.layers` (remote-code variant) and `model.model.layers` (native HF variant), fixing layer discovery for NemotronH when loaded without `trust_remote_code`.
- `modelopt_recipes/general/ptq/nvfp4_experts_only-kv_fp8_layerwise_offload.yaml`  new recipe combining NVFP4 W4A4 on MoE experts, FP8 KV cache, and layerwise calibration with `calib_mutates_weights: false` (required for disk-offload compatibility).
- `example_utils.py`  `_FP8BF16Fallback` shim: dequantizes block-scaled FP8 expert weights to BF16 for calibration forward passes when the `kernels` package is unavailable (e.g. DSR1 on nodes without finegrained FP8 kernel support).

### Usage

```python
# Single-GPU PTQ for a model too large to fit in VRAM, using disk offload
    --pyt_ckpt_path /path/to/NVIDIA-Nemotron-3-Ultra-550B-A55B-BF16 \
    --recipe general/ptq/nvfp4_experts_only-kv_fp8_layerwise_offload \
    --export_path /path/to/output \
    --offload_folder /path/to/offload \
    --max_gpu_memory_gb 170 \
    --max_cpu_memory_gb 500 \
    --trust_remote_code \
    --calib_size 8 --batch_size 1 --skip_generate
```
### Testing

**Unit tests** (`tests/unit/torch/export/test_offload_export.py`, 7 tests, CPU-only):
- `_has_accelerate_offload` detection (true/false/nested-module cases)
- `_export_quantized_weight` meta-tensor guard (raises on meta, passes on real)
- `_process_quantized_modules_offloaded` with disk-offloaded embed + GPU-resident decoder layer: verifies no meta tensor in returned state dict

**GPU integration tests** (`tests/gpu/torch/export/test_offload_export.py`, 2 tests):
- Tiny 2-layer LLaMA with CPU offload: FP8 quantization + export, asserts no meta tensors and valid `hf_quant_config.json`
- Same with layerwise FP8 (`calib_mutates_weights=False`): disk-offload path end-to-end

## End-to-end validation

Verified with DSR1 that the non-layerwise path provide identical checkpoint before and after this change, also the layerwise with cpu off-load path produce same identical checkpoint (with same max calibration setting).

Two production-scale checkpoints were quantized end-to-end using the new disk-offload PTQ path on a single GB200 GPU (189 GiB VRAM).

### DeepSeek-R1 (671B, MoE)

| | |
|---|---|
| **Checkpoint** | `DeepseekV3ForCausalLM`, 671B params, 61 decoder layers |
| **Input size** | 642 GB BF16 |
| **Recipe** | `nvfp4_experts_only-kv_fp8_layerwise_offload` |
| **`--max_gpu_memory_gb`** | 80 |
| **`--max_cpu_memory_gb`** | 80 |
| **`--calib_size` / `--batch_size`** | 8 / 1 |
| **`--trust_remote_code`** | no (built-in transformers) |
| **Wall-clock** | 40 min 12 s (load ~14 min, calib ~12 min, export ~14 min) |
| **Peak GPU memory** | 88.9 GB |
| **Peak process RSS** | 376 GB |
| **Output** | 40 shards x ~10 GB = 403 GB (~37% compression) |

image

### NVIDIA-Nemotron-3-Ultra-550B-A55B-BF16 (550B, NemotronH MoE + Mamba)

| | |
|---|---|
| **Checkpoint** | `NemotronHForCausalLM`, ~550B params, 108 decoder layers |
| **Input size** | ~1.1 TB BF16 |
| **Recipe** | `nvfp4_experts_only-kv_fp8_layerwise_offload` |
| **`--max_gpu_memory_gb` / `--max_cpu_memory_gb`** | 170 / 500 and **80 / 80** |
| **`--calib_size` / `--batch_size`** | 8 / 1 |
| **`--trust_remote_code`** | yes (`NemotronHForCausalLM`) |

| | 170 GB GPU / 500 GB CPU | **80 GB GPU / 80 GB CPU** |
|---|---|---|
| **Wall-clock** | 41 min 13 s | **47 min 16 s** |
| **Peak GPU memory** | 165.8 GB  | **76.7 GB** |
| **Peak RSS (load)** | 789 GB transient | **345 GB transient** |
| **Steady-state RSS** | ~454-496 GB | **~50 GB** |
| **Output** | 34 shards x ~11 GB = 365 GB | 34 shards x ~11 GB = 365 GB |

image

## Summary by CodeRabbit

* **New Features**
  * Added disk/CPU/GPU memory-limited offload model loading.
  * Added offload-aware streaming Hugging Face checkpoint export with sharded output.
  * Added an NVFP4 expert-only PTQ recipe with FP8 KV-cache support.
  * Improved Nemotron-H model layout support.
* **Bug Fixes**
  * Improved DeepSeek bundled-code selection based on remote-code trust.
  * Strengthened handling of meta/offloaded weights, tied-weight deduplication, and export post-processing.
* **Tests**
  * Added coverage for offload exports and DeepSeek loading behavior.


modelopt/torch/export/unified_export_hf.py
realAsma, line 1: RB: Can you explain to me the overall flow of the changes in this file?

realAsma: > _:robot: Bot comment._

🐝 I’m tracing the file-level flow and its before/after behavior now.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -874,13 +1298,18 @@ def _export_transformers_checkpoint(
...
-    except ImportError:
-        warnings.warn("accelerate is not installed, hooks will not be removed")
+    # Remove all hooks from the model (deferred for offloaded models)
+    if not _offloaded:
+        try:
+            from accelerate.hooks import remove_hook_from_module
+
+            remove_hook_from_module(model, recurse=True)
+        except ImportError:
+            warnings.warn("accelerate is not installed, hooks will not be removed")
```
realAsma: This warning does not make sense to me. Import error means accelerate not installed -> means there should not be any hook in the first place.

modelopt/torch/export/unified_export_hf.py
realAsma: > _:robot: Bot comment._

The high-level change is that `export_hf_checkpoint()` now has two Transformers paths:

- **Normal model:** keeps the existing flow: process all quantized modules → build/post-process one full state dict → call `save_pretrained()`.
- **Accelerate CPU/disk-offloaded model:** `_has_accelerate_offload()` detects the hooks and routes to `_export_transformers_checkpoint_streaming()`.

The new offloaded flow is:

1. Run the same model-level preparation as before (MoE preparation, resmooth/requantize, build quant config, and synchronize relevant amax values).
2. Discover decoder layers, then materialize **one layer at a time** inside `enable_weight_access_and_writeback(..., writeback=False)`. While the real weights are available, dispatch the existing export handlers to pack quantized weights and reconstruct fused MoE weights.
3. Immediately post-process each tensor (drop quantizer internals, convert KV-cache amax to scales, squeeze scales, reverse Transformers conversion names, and skip tied aliases), move it to CPU, and feed it to `_StreamingShardWriter`. Exiting the context lets Accelerate offload that layer again.
4. Repeat the same materialize-and-stream treatment for offloaded non-decoder modules such as embeddings, norms, and `lm_head`, then stream any remaining resident parameters/buffers.
5. `_StreamingShardWriter` flushes at `max_shard_size`, renames temporary parts to canonical safetensor shard names, and writes the index. This bounds peak CPU memory to roughly one materialized layer plus one shard buffer instead of a full-model state dict.
6. Call `save_pretrained(state_dict={})` only to emit config/custom-code artifacts, then the outer function writes `hf_quant_config.json` and updates/sanitizes `config.json` as before.

Two safety pieces support this: Accelerate hooks are no longer removed before offloaded weights are materialized, and `_export_quantized_weight()` now fails loudly if a meta tensor reaches it. `_process_quantized_modules_offloaded()` is the analogous full-state-dict helper for direct lower-level calls, but the public top-level offload path uses streaming to avoid the full CPU-RAM accumulation.

realAsma: RB: This PR should treat offloading as the special case and consider not offloading as the general case. I feel that this PR does not respect this assumption. Example there are many places where regular path is route to if not _offloaded: ... (is it possible to have a cleaner seperation which does not need to move general model export code under if not _offloaded?)

realAsma: > _:robot: Bot comment._

🐝 I’m checking the branch structure and shared-vs-offload responsibilities now.

realAsma: > _:robot: Bot comment._

Yes—your read is correct. The regular path should remain the straight-line default, with one early special-case dispatch:

```python
if _has_accelerate_offload(model):
    return _export_offloaded_transformers_checkpoint(...)

# existing regular export flow, unchanged
post_state_dict, hf_quant_config = _export_transformers_checkpoint(...)
...
model.save_pretrained(...)
```

The offload helper should own its unsupported-option warnings, layer materialization/streaming, inline name conversion, and `save_pretrained(state_dict={})` artifact workaround. Only genuinely common finalization (quant config + sanitized `config.json`) needs a small shared helper.

There is also duplicated dispatch today: `export_hf_checkpoint()` routes offloaded models directly to `_export_transformers_checkpoint_streaming()`, while `_export_transformers_checkpoint()` independently detects `_offloaded` and branches to `_process_quantized_modules_offloaded()`. That lower-level branch is bypassed by the public path and makes ordinary callers offload-aware. Removing that branch from the general exporter and keeping it inside the dedicated offload path would restore the intended separation and eliminate the repeated `if not _offloaded` guards.

realAsma: @Fridah-nv  could you please address my concern - https://github.com/NVIDIA/Model-Optimizer/pull/2008/changes#r3648000772

Here is bot comment: https://github.com/NVIDIA/Model-Optimizer/pull/2008/changes#r3648016031

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1430,14 +1861,39 @@ def export_hf_checkpoint(
...
+                warnings.warn(
+                    "save_modelopt_state=True is not supported in the streaming offload export "
+                    "path and will be ignored."
+                )
+            if extra_state_dict:
+                warnings.warn(
+                    "extra_state_dict is not supported in the streaming offload export path "
+                    "and will be ignored."
+                )
+            _, hf_quant_config = _export_transformers_checkpoint_streaming(
```
realAsma, line 1600: RB: how does this code reach the `_process_quantized_modules_offloaded`? Can you show me a flowchart of the code flow?

realAsma: > _:robot: Bot comment._

🐝 I’m tracing the offloaded export call path now and will follow up with a flowchart.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -811,20 +835,420 @@ def _process_quantized_modules(
...
+    # renames its output to model.safetensors, which save_pretrained would overwrite).
+    _single_shard = export_dir / "model.safetensors"
+    _protected = export_dir / "__modelopt_protected_model.safetensors"
+    if _single_shard.exists():
+        _single_shard.rename(_protected)
+
+    _sanitize_generation_config_for_save(model)
+    _patches = _patch_revert_weight_conversion()
+    try:
+        model.save_pretrained(str(export_dir), state_dict={})
```
shengliangxu: It seems this line may remove existing safetensors files because save_pretrained with an empty state_dict will think the already written safetensors files obsolete and so remove them. could you double check?

modelopt/torch/export/unified_export_hf.py
realAsma: > _:robot: Bot comment._

Short answer: this branch does **not** reach `_process_quantized_modules_offloaded`.

```mermaid
flowchart TD
    A[export_hf_checkpoint] --> B[_has_accelerate_offload]
    B -->|true| C[_export_transformers_checkpoint_streaming]
    C --> D[discover decoder layers]
    D --> E[materialize one layer<br/>enable_weight_access_and_writeback]
    E --> F[_dispatch_export_handler for submodules]
    F --> G[reconstruct fused MoE]
    G --> H[_postprocess_single_tensor]
    H --> I[_StreamingShardWriter.add / finalize]

    B -->|false| J[_export_transformers_checkpoint]
    J --> K[its own _has_accelerate_offload check]
    K -->|false in this call path| L[_process_quantized_modules]
```

The offloaded helper is called at line 1354 only inside `_export_transformers_checkpoint` when *its* offload check is true. But `export_hf_checkpoint` calls that function only from the outer `_offloaded == false` branch, so it cannot select the helper on this entry path.

It is still reachable from direct callers of `_export_transformers_checkpoint` (for example the speculative exporters) and directly from its unit test. The streaming function instead implements the helper's layer-by-layer materialize/dispatch behavior inline at lines 1171–1211. So for this path, `_process_quantized_modules_offloaded` is effectively bypassed/redundant.

realAsma: RB: "It is still reachable from direct callers of _export_transformers_checkpoint (for example the speculative exporters) and directly from its unit test."

This does not look clean to me. Why can't speculative exporters call the _export_transformers_checkpoint_streaming for offload models?

realAsma: > _:robot: Bot comment._

🐝 I’m checking the speculative exporter contracts and streaming path now; I’ll follow up with the concrete compatibility constraints.

realAsma: @Fridah-nv  could you also take a look?

realAsma: > _:robot: Bot comment._

Yes—the speculative exporters should dispatch to a streaming-capable path when the model is offloaded. They cannot call this helper verbatim only because its current contract is wrong for them: `_export_transformers_checkpoint_streaming` writes the entire base+draft checkpoint and HF config directly to `export_dir` and returns no state dict, while Eagle/DFlash must select and rename only `eagle_module.*` / `dflash_module.*`, borrow `lm_head.weight` for Eagle, validate the draft keys, and emit their own config (plus optional LoRA artifacts). Calling it as-is would first write the wrong full-model artifacts and then collide with the speculative `model.safetensors`.

The current fallback through `_export_transformers_checkpoint` is not a good low-memory solution either: it assembles the full state dict and defeats the reason for offload. The clean fix is to factor the layer-by-layer materialization/postprocessing into a reusable tensor iterator/sink, let normal HF export and speculative export provide different key filters/remappers/artifact writers, and dispatch on `_has_accelerate_offload` in both speculative exporters. Then `_process_quantized_modules_offloaded` can be removed if it has no remaining production caller; its unit test alone is not a reason to retain it.

modelopt/torch/export/unified_export_hf.py
Fridah-nv: Addressed in f76231394: the offloaded path now early-returns after its own block, and a `_write_hf_export_config` helper owns the shared `hf_quant_config.json + config.json` finalization. The regular export path is now the straight-line default with no guards; the offloaded block is a clean early exit at the top.

modelopt/torch/export/unified_export_hf.py
Fridah-nv: You are right, fixed it to `pass` in e4d69c7bc.

modelopt/torch/export/unified_export_hf.py
Fridah-nv: ` _process_quantized_modules_offloaded` is not reachable.  `export_hf_checkpoint` dispatches offloaded models to `_export_transformers_checkpoint_streaming` and non-offloaded models to `_export_transformers_checkpoint`, so the if _offloaded: branch inside the latter was unreachable. f76231394 removes `_process_quantized_modules_offloaded` entirely and replaces that branch with `NotImplementedError`.

Fridah-nv: > "Why can't speculative exporters call the streaming path?"

 They can't call `_export_transformers_checkpoint_streaming` verbatim. it writes the full base-model artifacts (model-NNNNN-of-NNNNN.safetensors, config.json, generation_config.json) and returns no state dict, while Eagle/DFlash need to filter and rename keys (eagle_module.*, dflash_module.*), borrow `lm_head.weight` for Eagle, and emit their own config artifacts. A verbatim call would write the wrong key namespace and collide with the speculative checkpoint.

The right fix is to extract the layer-by-layer materialization into a reusable tensor iterator that spec exporters can plug their own key remapper and sink into. We've removed `_process_quantized_modules_offloaded` (it accumulated a full state dict, defeating offload) and left a `NotImplementedError` in `_export_transformers_checkpoint` for the offloaded case. The reusable iterator for spec exporters is scoped as a follow-up happy to open a tracking issue.

modelopt/torch/export/unified_export_hf.py
Fridah-nv: This is correct. I forgot to push the local fix e4d69c7bc. It calls with explicit `model.config.save_pretrained() + generation_config.save_pretrained() + shutil.copy2(*.py)`. This also avoids the MoE shared-tensor crash we hit on DSR1, where `save_pretrained` fired safetensors' shared-tensor validation on the live model parameters even with an empty state dict.

modelopt_recipes/general/ptq/nvfp4_experts_only-kv_fp8_layerwise_offload.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+imports:
+  base_disable_all: configs/ptq/units/base_disable_all
+  default_disabled_quantizers: configs/ptq/units/default_disabled_quantizers
+  nvfp4: configs/numerics/nvfp4
+  kv_fp8: configs/ptq/units/kv_fp8
+
+metadata:
+  recipe_type: ptq
+  description: >
```
Edwardf0t1, line 24: **Blocker — two unit tests will fail in CI.**

`tests/unit/recipe/test_recipe_docs.py` (added in #1970) enforces recipe↔doc parity. Checked out this branch and ran the check:

```
total yaml: 21
missing from ptq.md: ['nvfp4_experts_only-kv_fp8_layerwise_offload']
documented count: 20
```

So both `test_every_general_ptq_recipe_is_documented` and `test_general_ptq_recipe_count_in_ptq_md` fail.

Fix: add a row for this recipe to the shipped-recipes table in `modelopt_recipes/ptq.md` and bump the `All 20 <code>general/ptq/</code> recipes` summary line to `All 21`.

examples/hf_ptq/example_utils.py
```diff
@@ -53,6 +54,61 @@
...
 
 
+class _FP8BF16Fallback:
+    """BF16 dequant fallback for block-scaled FP8 matmul when the kernels package is absent.
+
+    Calibration amax collection only — not accurate for production inference.
+    """
+
+    @staticmethod
+    def matmul(input, weight, weight_scale_inv, block_size, output_dtype=None, activation_scale=None):
```
Edwardf0t1: **Blocker — pre-commit wasn't run.** `ruff format --check` and `ruff check` both fail on this branch.

`ruff format` diffs:
- `examples/hf_ptq/example_utils.py` — 3 hunks: this `matmul` signature (>100 chars), the `is_flash_attn_*` tuple, and `f"{val / 1024 ** 3:.1f}"` → `f"{val / 1024**3:.1f}"` in `_fmt_max_memory`
- `modelopt/torch/export/quant_utils.py` — `_maybe_squeeze_scale`'s condition exceeds the 100-char `line-length`

`ruff check`:
```
SIM118 Use `key in dict` instead of `key in dict.keys()`
 --> tests/gpu/torch/export/test_offload_export.py:109:17
```

`pre-commit run --all-files` should clear all of these.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -811,20 +836,332 @@ def _process_quantized_modules(
...
+        part_path = self._export_dir / f"__shard_part_{part_idx:05d}.safetensors"
+        save_file(self._buffer, str(part_path))
+        for key in self._buffer:
+            self._key_to_part[key] = part_idx
+        self._part_files.append(part_path)
+        self._total_bytes += self._buffer_bytes
+        self._buffer = {}
+        self._buffer_bytes = 0
+
+    def add(self, key: str, tensor: torch.Tensor) -> None:
```
Edwardf0t1: **The streaming path drops the tied-weight safety net, so `save_file` can crash.**

The batch path ends `postprocess_state_dict` with a `data_ptr()`-based dedup (`quant_utils.py` ~L1115-1135) that removes *any* aliased tensor. That's what guarantees `safetensors.save_file` never sees two keys sharing storage. The streaming path replaces it with name matching against `_tied_weights_keys`, gated on `tie_word_embeddings=True`.

`_stream_tensor` passes `tensor.detach().contiguous().cpu()` — for an already-contiguous CPU tensor all three are no-ops returning the same object, so aliasing survives into `self._buffer` and `save_file` raises `RuntimeError: Some tensors share memory`.

Concrete failure: a model with `tie_word_embeddings=False` in config but genuinely shared `lm_head.weight` / `embed_tokens.weight` storage, or any aliasing not literally listed in `_tied_weights_keys` (note transformers v5 changed `_tied_weights_keys` to a dict of *patterns* for several architectures, so exact-string matching can silently miss). Both keys land in the same shard buffer → crash after all prior shards are already written.

The comment above is right that `data_ptr()` is unreliable across the whole export — but it *is* reliable **within a single shard buffer**, because the writer holds a strong reference to every buffered tensor until `_flush()`. A per-buffer check restores parity without the cross-layer staleness problem:

```python
def add(self, key: str, tensor: torch.Tensor) -> None:
    ptr = tensor.data_ptr()
    if ptr in self._buffer_ptrs:   # reset alongside self._buffer in _flush()
        return                      # (or .clone() if you'd rather keep both keys)
    self._buffer_ptrs[ptr] = key
    self._buffer[key] = tensor
    ...
```

examples/hf_ptq/example_utils.py
```diff
@@ -587,16 +643,40 @@ def _apply_dtype_to_config(model_kwargs, config_dtype, architecture, apply_confi
...
     device="cuda",
     gpu_mem_percentage=0.8,
     trust_remote_code=False,
     use_seq_device_map=False,
     attn_implementation=None,
+    offload_folder=None,
+    max_cpu_memory_gb=None,
+    max_gpu_memory_gb=None,
 ):
+    _install_transformers_compat_shims()
```
Edwardf0t1: **`_install_transformers_compat_shims()` runs on every `get_model()` call**, not just disk-offload or `trust_remote_code` loads. Three global side effects for all `hf_ptq.py` users:

1. If `import flash_attn` raises for **any** reason (the `except` is a bare `Exception`, so a transient/unrelated failure counts), `is_flash_attn_2_available` / `is_flash_attn_available` / `is_flash_attn_greater_or_equal_2_10` get patched to `lambda: False` on two transformers modules — silently changing attention-implementation selection for every model.
2. `_ff8._load_finegrained_fp8_kernel` (a private transformers symbol) is eagerly called and, on `ImportError`, swapped for a **lossy** BF16 dequant. The whole block is under `contextlib.suppress(Exception)`, so any failure here is invisible.
3. `is_torch_fx_available` is injected into `transformers.utils.import_utils`.

Silently degrading numerics inside a *quantization* tool seems like the wrong default even with the `warnings.warn`. Could this be gated — `if trust_remote_code:` at minimum, or behind an explicit opt-in flag?

Relatedly, `_FP8BF16Fallback.matmul` accepts and discards `block_size`, `activation_scale`, and bias, re-deriving block factors as `out_f // nb_out`. If the scale grid is ceil-divided (padded for dims not divisible by 128) that `repeat_interleave` yields a wrong-sized scale → shape error or silently wrong scaling; using `block_size` directly would be exact. It also materializes a full `out_f × in_f` fp32 tensor per call (~528 MB for DSR1's 7168×18432, plus the bf16 copy) inside the calibration hot loop.

examples/hf_ptq/example_utils.py
```diff
@@ -700,12 +780,11 @@ def has_pack_quantized_config(config):
             raise ValueError(f"Model config at {ckpt_path} has no architectures defined")
         architecture = hf_config.architectures[0]
 
-        if not hasattr(transformers, architecture) or "Deepseek" in architecture:
-            if not hasattr(transformers, architecture):
-                warnings.warn(
-                    f"Architecture {architecture} not found in transformers: {transformers.__version__}. "
-                    "Falling back to AutoModelForCausalLM (or AutoModel for non-causal architectures)."
-                )
+        if not hasattr(transformers, architecture):
```
Edwardf0t1: **Unexplained behavior change:** dropping `or "Deepseek" in architecture` from this condition.

Previously any DeepSeek architecture took the `AutoModelForCausalLM` + `assert trust_remote_code` path *even when* `transformers` exposed the class. Now `DeepseekV3ForCausalLM` (present since transformers 4.53) falls through to `getattr(transformers, architecture)._from_config`, and further down `model_kwargs2.pop("trust_remote_code", None)` fires.

That's a real behavior change for the exact model family this PR targets, it isn't mentioned in the description, and there's no test covering it. It also appears to pull in the opposite direction from `_install_transformers_compat_shims()`, which exists specifically to make *remote-code* DSR1 load.

Could you explain the intent, or split this into its own PR so it can be evaluated independently?

Edwardf0t1 (commented):
Inline notes on the five items I consider merge-blocking or worth resolving before merge. Overall the streaming-export architecture is sound and the Ultra-550B validation run (161 GB peak VRAM, 15 shards) is convincing evidence the happy path works — these are about CI, an aliasing edge case, and two scope questions.

modelopt/torch/export/unified_export_hf.py
realAsma: sounds good

modelopt/torch/export/unified_export_hf.py
```diff
@@ -918,18 +1260,22 @@ def _export_transformers_checkpoint(
...
-
-    if accelerator is not None:
-        # Gather state_dict from all ranks
-        quantized_state_dict = accelerator.get_state_dict(model)
+    if _offloaded:
+        raise NotImplementedError(
+            "_export_transformers_checkpoint does not support disk/CPU-offloaded models. "
+            "Use export_hf_checkpoint() which dispatches to _export_transformers_checkpoint_streaming."
+        )
     else:
```
realAsma, line 1041: can we remove this else, does not look like needed.

realAsma (approved):
Looks great to me!!

Could you please benchmark DSR1/Ultra single GPU PTQ with max calibration (PTQ + export time, GPU memory, GPU utilization, CPU memory etc.) and add that information to the PR description?

modelopt_recipes/general/ptq/nvfp4_experts_only-kv_fp8_layerwise_offload.yaml
Fridah-nv: Fixed, thanks for flagging this

modelopt/torch/export/unified_export_hf.py
Fridah-nv: I adopted the suggested change. In the streaming export case, the tied word embedding case is handled by name filter in `_stream_tensor`, while the storage need to guard on alias because of these cases:
`ExportContext.tied_cache` / `moe_tied_cache` are live dicts, and `_export_quantized_weight` deliberately re-points weight / weight_scale / weight_scale_2 / input_scale at a previously-processed module sharing source memory — so the downstream data_ptr dedup can collapse them. This is `postprocess_state_dict`, batch-path only; streaming has no equivalent, so such aliases would reach save_file directly.

Let me know if this makes sense. I'm actually not famliar with the `moe_tied_cache` part.

modelopt/torch/export/unified_export_hf.py
Fridah-nv: Done

examples/hf_ptq/example_utils.py
Fridah-nv: I removed the whole function because it's more of a workaround, this is because of some separate env issue

examples/hf_ptq/example_utils.py
Fridah-nv: I clean this up as well but need a full run on DSR1 to verify

Edwardf0t1 (commented):
Re-reviewed at e5198534 — all five of my earlier findings are resolved, and I verified the first two by running the checks (`ruff format`/`ruff check` clean; recipe docs now 21/21 with no stale rows). One remaining test-coverage gap below.

tests/gpu/torch/export/test_offload_export.py
```diff
@@ -0,0 +1,119 @@
...
+    quant_config_path = export_dir / "hf_quant_config.json"
+    assert quant_config_path.exists(), "hf_quant_config.json not written"
+    with open(quant_config_path) as f:
+        quant_config = json.load(f)
+    assert quant_config["quantization"]["quant_algo"] == "FP8", (
+        f"Expected FP8, got {quant_config['quantization'].get('quant_algo')}"
+    )
+
+    # 2. All tensors in safetensors shards must be non-empty (no meta serialized as zeros)
+    safetensor_files = list(export_dir.glob("*.safetensors"))
```
Edwardf0t1, line 105: **Missing: a streaming-vs-batch key-parity test.**

The unit tests added for the alias guard and the `data_ptr`-under-offload cases are good, and this test is a solid guard against the original meta-tensor bug. But every assertion here iterates only the keys that *were* written — so it can confirm the written tensors are sane, and can't detect anything **missing**.

Concretely: if a refactor caused `_export_transformers_checkpoint_streaming` to skip an entire decoder layer (a `seen_keys` collision, a `_tied_weights_keys` pattern over-matching and dropping too much, a `_dispatch_export_handler` early-return, or the non-decoder pass missing an offloaded `lm_head`), this test still passes — the surviving shards are non-empty and non-meta.

That's the highest-risk failure mode for this PR, because the streaming path reimplements a lot of `_export_transformers_checkpoint` + `postprocess_state_dict` per-tensor rather than sharing it, and the two can drift silently.

The cheap version is to export the *same* tiny model twice — once offloaded, once not — and diff the key sets:

```python
def test_streaming_export_matches_batch_export_keys(tmp_path):
    # non-offloaded reference
    ref_model = ...  # same tiny llama, plain .cuda(), same quant_cfg + forward_loop
    export_hf_checkpoint(ref_model, export_dir=str(ref_dir))

    # offloaded, exercises the streaming path
    off_model, *_ = _make_cpu_offloaded_model(...)
    export_hf_checkpoint(off_model, export_dir=str(off_dir))

    assert _all_keys(off_dir) == _all_keys(ref_dir)
```

where `_all_keys` unions `safe_open(...).keys()` across every shard. Comparing values too would be even better, but key parity alone catches the whole "silently dropped tensors" class and is what the current assertions can't see.

Two smaller gaps in the same spirit, if you want them in the same pass:
- **Multi-shard is never integration-tested.** Tiny LLaMA fits in one shard, so `model.safetensors.index.json` generation *inside the real export path* is untested end-to-end — only `_StreamingShardWriter` is tested in isolation. Parametrizing this test with `max_shard_size="1MB"` would cover it.
- `_parse_shard_size` has no test.

tests/tools/test_resource_monitor.py
kevalmorabia97, line 1: You need to also enable tools unit tests in https://github.com/NVIDIA/Model-Optimizer/blob/main/.github/workflows/unit_tests.yml

examples/hf_ptq/example_utils.py
```diff
@@ -659,16 +659,39 @@ def _apply_dtype_to_config(model_kwargs, config_dtype, architecture, apply_confi
...
 
+    _disk_offload = offload_folder is not None
+    if _disk_offload and max_cpu_memory_gb is None:
+        warnings.warn(
+            "offload_folder is set but max_cpu_memory_gb is not specified. "
+            "CPU memory usage during model load will be unbounded. "
+            "Pass max_cpu_memory_gb to cap CPU usage.",
+            UserWarning,
+        )
+
```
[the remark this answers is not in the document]
Fridah-nv: Fixed in adc851d13a with a warning rather than an error.

examples/hf_ptq/example_utils.py
```diff
@@ -737,24 +764,41 @@ def has_pack_quantized_config(config):
...
+                    "Model does not fit to the GPU mem. "
+                    f"We apply the following memory limit for calibration: \n{max_memory}\n"
+                    "If you hit GPU OOM issue, please adjust `gpu_mem_percentage` or "
+                    "reduce the calibration `batch_size` manually."
+                )
+                model_kwargs["max_memory"] = max_memory
 
         model_kwargs2 = _apply_dtype_to_config(model_kwargs, config_dtype, architecture)
+        if _disk_offload:
+            model_kwargs2["offload_folder"] = offload_folder
```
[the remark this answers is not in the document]
Fridah-nv: Took the "reject up front" option in adc851d13a, but at the CLI rather than inside get_model

examples/hf_ptq/hf_ptq.py
```diff
@@ -1563,6 +1566,38 @@ def parse_args() -> argparse.Namespace:
...
+    parser.add_argument(
+        "--max_gpu_memory_gb",
+        type=float,
+        default=None,
+        help=(
+            "Maximum GPU memory budget per device in GiB for disk-offload model loading. "
+            "Only effective when --offload_folder is set. "
+            "Defaults to 80%% of available GPU memory when not specified."
+        ),
+    )
```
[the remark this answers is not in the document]
Fridah-nv: Valid, and accepted as suggested. Added in adc851d13a, next to the existing `--low_memory_mode / --use_seq_device_map` rejections so all the incompatible-flag checks stay togethe

examples/hf_ptq/hf_ptq.py
```diff
@@ -1584,6 +1619,16 @@ def parse_args() -> argparse.Namespace:
...
+    if args.offload_folder is not None and args.low_memory_mode:
+        parser.error("--offload_folder (disk-offload) is not compatible with --low_memory_mode.")
+
+    if args.offload_folder is not None and args.use_seq_device_map:
+        parser.error(
+            "--offload_folder (disk-offload) is not compatible with --use_seq_device_map; "
+            "device_map=auto is used for disk-offload to let accelerate place layers across "
+            "GPU, CPU, and disk."
+        )
+
```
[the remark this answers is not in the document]
Fridah-nv: both landed in adc851d13a in the spot you identified, alongside the existing --low_memory_mode / --use_seq_device_map rejections

tests/gpu/torch/export/test_offload_export.py
Fridah-nv: Added test_streaming_export_matches_batch_export in adc851d13a. Thanks for the suggestion!

modelopt/torch/export/unified_export_hf.py
Fridah-nv: I removed the support of tied_cache in this feature, but keep this guard

modelopt/torch/export/unified_export_hf.py
```diff
@@ -813,20 +902,401 @@ def _process_quantized_modules(
 
             fsdp_module_to_reshard = sub_module
 
-        # We skip QuantLoraLinear module for modelopt QLoRA
-        if ctx.is_modelopt_qlora and hasattr(sub_module, "base_layer"):
-            continue
+        _dispatch_export_handler(name, sub_module, ctx)
+
+
+class _StreamingShardWriter:
```
Edwardf0t1: **Non-blocking maintainability suggestion: split the streaming/offload block into its own module.**

`unified_export_hf.py` goes 1527 → 2018 lines with this PR, which makes it the largest file in `modelopt/torch/export/`. #1939 already carved `registry.py` and `hf_export_handlers.py` out of it for the same reason, so there's precedent for the direction.

The offload block is the cleanest extraction available:

- `_StreamingShardWriter`, `_parse_shard_size`, `_assert_no_split_rules`, `_export_transformers_checkpoint_streaming` — L908–1301, ~393 lines.
- Moving them to e.g. `modelopt/torch/export/unified_export_hf_streaming.py` brings the file to ~1625, i.e. **+100 net** over main instead of +491.
- It's all new code, so no re-export shims are needed: the only external importer is this PR's own `tests/unit/torch/export/test_offload_export.py` (3 symbols).

It also makes @realAsma's [separation point](https://github.com/NVIDIA/Model-Optimizer/pull/2008#discussion_r3648000772) structural rather than a convention. Today the general exporter and the offload machinery share a file and the boundary is held by an early return; after the move `export_hf_checkpoint` keeps a two-line dispatch and the general path can't drift back into being offload-aware.

One import detail: the block depends on `ExportContext`, `_dispatch_export_handler`, `_resolve_export_dtype`, `_prepare_moe_inputs`, `_add_mtp_exclusions`, `_warn_on_unsynced_moe_gate_up`, `_patch`/`_unpatch_revert_weight_conversion`, and `_sanitize_generation_config_for_save` — all of which you already factored out in ae6a0f5de for exactly this sharing. Putting those in a small `hf_export_prep.py` that both exporters import keeps it acyclic; importing them back from `unified_export_hf` would need another lazy import.

Fine by me as a follow-up if you'd rather not touch the diff again — the branch needs a rebase for the conflict anyway, so it could ride along there. Two related cleanups I'd deliberately keep *out* of this PR:

- The diffusers block (`_export_diffusers_checkpoint`, `_fuse_qkv_linears_diffusion`, `_detect_svdquant_rank`, `_promote`/`_remove_promoted_quantizer_tensors`, plus `_save_component_state_dict_safetensors`/`_postprocess_safetensors`) is ~517 more lines sharing essentially nothing with the transformers path. Bigger win, but it needs a `# noqa: F401` shim for `_postprocess_safetensors`, which `test_nvfp4_utils.py` and `test_export_diffusers.py` import from here.
- `_export_quantized_weight` / `requantize_resmooth_fused_llm_layers`: moving these into a `hf_weight_export.py` would actually break the cycle that `moe_utils.py:117` and `hf_export_handlers.py:46` currently dodge with lazy imports — but `examples/llm_qat/export.py`, `examples/hf_ptq/multinode_ptq.py` and ~6 test files import them directly, so that deserves its own PR with shims.

One suggestion on placement either way: not `quant_utils.py` — at 1607 lines it would just relocate the problem.

shengliangxu (approved):
LGTM

modelopt/torch/export/unified_export_hf.py
Fridah-nv: Done in a8afef2eed, rebased as suggested.  Moved `_StreamingShardWriter`, `_parse_shard_size`, `_assert_no_split_rules` and `_export_transformers_checkpoint_streaming` into `unified_export_hf_streaming.py`. 
Outside the move: `export_hf_checkpoint` imports the streaming exporter lazily at the dispatch, the unit test import splits across the two modules, and ruff dropped contextlib/itertools/shutil from the original

Edwardf0t1 (approved):
LGTM. 

Do you think `_dispatch_export_handler` and similar utils in `unified_export_hf.py` can be refactored to a separate script?

Fridah-nv:
> LGTM.
> 
> Do you think `_dispatch_export_handler` and similar utils in `unified_export_hf.py` can be refactored to a separate script?

The new helpers are quite tiny, moving them won't remove the lazy import. If the suggestion is to move the resmooth/fusion pass, I would suggest a separate PR for it, because they have quite some importers and we need to update more files.

PR #2010, opened 2026-07-23
Add ModelOpt QAD skill for Slurm workflows

mxinO:
### What does this PR do?

Type of change: new feature

Adds a general Slurm-only QAD skill based on the supported Megatron Bridge
workflow. The skill:

- starts from a measured BF16-to-PTQ benchmark gap and preserves the preceding
  PTQ configuration or recipe;
- gates QAD on exact Megatron Bridge model support and successful Megatron PTQ,
  using its master-rank quantizer summary as a scoped `amax` sanity check;
- requires model- and hardware-derived TP/PP/CP/EP/ETP topology selection;
- streams and randomly samples only the required
  `nvidia/Nemotron-Cascade-2-SFT-Data` token budget and uses Megatron sequence
  packing;
- defaults to 32K sequences, LR `1e-5` with cosine decay, a 1000-step cap, and
  GBS 512;
- requires explicit user authorization because QAD is costly, validates two
  batches every 25 steps, saves every 50 steps, and monitors a decreasing
  smoothed loss trend;
- evaluates an early checkpoint around step 150 and continues only when
  benchmark recovery and the loss trend justify more training;
- follows the established common Slurm and remote-execution guidance instead of
  duplicating mutable commands from the Megatron Bridge README.

Also exposes Megatron Bridge `save_interval`, `exit_interval`, and
`exit_duration_in_mins` through `examples/megatron_bridge/distill.py`, with example-test coverage for checkpoint
and ModelOpt-state preservation at an early exit.

### Usage

```text
Use the QAD skill to recover the measured BF16-to-PTQ benchmark gap for
<model> on <Slurm cluster>, preserving the validated PTQ recipe.
```

### Testing

- `PYTHONPATH=$PWD pre-commit run --all-files`
  - Passed every hook on the rebased branch, including Ruff, Ruff format, mypy,
    YAML/recipe validation, launcher reference validation, Bandit, generated
    arguments, symlink synchronization, and Markdown lint.
- `python ~/.codex/skills/.system/skill-creator/scripts/quick_validate.py .agents/skills/qad`
  - `Skill is valid!`

Qwen3-0.6B result-bearing validation:

- Resources: one exclusive node, 8 H100 GPUs
- Container: `nvcr.io/nvidia/nemo:26.06`
- Quantization: NVFP4, group size 16, embedding excluded
- QAD topology: TP=1, PP=1, CP=4, EP=1, DP=2
- Training validation configuration: sequence length 32768, MBS=1, GBS=8,
  `train_iters=1000`, LR `1e-5` / minimum LR `1e-6`, 50 warmup iterations,
  cosine decay, `eval_interval=150`, `exit_interval=150`,
  `exit_duration_in_mins=220`
- This result-bearing run used the then-current coupled eval/save cadence. The
  final skill now validates two batches every 25 steps and saves every 50; the
  example test covers the independent checkpoint cadence.
- The reduced GBS 8 is intentionally validation-only; the skill retains GBS 512
  as the production default.
- Data: exactly 10,000,000 sampled tokens from four
  `nvidia/Nemotron-Cascade-2-SFT-Data` configs:
  - math: 2,306,011 tokens / 364 documents
  - science: 1,191,257 tokens / 285 documents
  - chat: 6,142,077 tokens / 1,800 documents
  - instruction following: 360,655 tokens / 411 documents
- Megatron built packed 32K GPT samples from the materialized prefixes; the full
  dataset was not downloaded.
- QAD loss was finite and decreased from `0.2640341` at iteration 10 to
  `0.1060580` at iteration 150. Final gradient norm was `0.747`, with zero
  skipped and zero NaN iterations. Validation distillation loss was
  `0.09715855`.
- The iteration-150 checkpoint saved successfully with `modelopt_state`, and
  both PTQ and QAD-150 exported to unified Hugging Face format.
- Identical full MMLU 0-shot comparison through the Megatron evaluator:

  | Model | Accuracy |
  | --- | ---: |
  | BF16 | 0.39517164 |
  | PTQ | 0.32851446 |
  | QAD-150 | 0.38740921 |

  QAD-150 recovered `0.05889475 / 0.06665718 = 88.35%` of the measured PTQ gap,
  so validation stopped at the early evidence gate rather than continuing
  blindly toward 1000 iterations.

- If you copied code from any other sources or added a new PIP dependency, did
  you follow guidance in `CONTRIBUTING.md`: N/A
- Did you update Changelog?: N/A — this adds an agent skill and example-only
  lifecycle flags.

### Additional Information

All seven branch commits are cryptographically signed and include a
`Signed-off-by` trailer.

## Summary by CodeRabbit

* **New Features**
  * Updated the QAD skill documentation with a clear “Execute in this order” workflow, including a revised default recovery training policy.
  * Added a new `nemotron-cascade-2` dataset blend configuration with an increased token budget.
  * Enhanced the MeGatron Bridge distillation CLI with stricter interval argument validation and support for configurable save-and-exit controls.
* **Documentation**
  * Expanded Megatron Bridge README guidance for dataset preparation, token-budget recalculation, and resume expectations.
* **Tests**
  * Improved distillation and QAD tests to validate early-exit behavior and checkpoint expectations.
  * Added unit tests covering distillation CLI interval validation.


examples/megatron_bridge/data/nemotron-cascade-2-blend.yaml
realAsma, line 1: Can we move this to `examples/megatron_bridge`?

.agents/skills/qad/SKILL.md
```diff
@@ -0,0 +1,219 @@
...
+  Use only when the user explicitly asks for QAD, including its topology, data
+  preparation, Slurm launch, resume, checkpoint export, or recovery decisions.
+---
+
+# ModelOpt Quantization-Aware Distillation
+
+QAD is expensive. Use it only to recover a material, apples-to-apples PTQ
+benchmark gap and only when the user explicitly requests QAD for the target
+model or run. Do not infer permission from a quantized checkpoint, evaluation
+gap, or recipe-search result.
```
chadvoegele: Can we reword this so that the agent can to choose to run QAD itself?

In the Day 0 workflow, I've seen agents trying a few PTQ recipes and then giving up. It would be nice for the agents to be able to explore QAD on their own.

chadvoegele (approved):

.agents/skills/qad/SKILL.md
```diff
@@ -0,0 +1,219 @@
+---
```
Edwardf0t1, line 1: This SKILL seems long, should we simplify and adopt the concise SKILL +  references structure?

.agents/skills/qad/SKILL.md
```diff
@@ -0,0 +1,219 @@
...
+   `gap = BF16 - PTQ`, `recovered = QAD - PTQ`, and
+   `recovery_fraction = recovered / gap`.
+
+Continue only if benchmark recovery is positive beyond run noise and loss is
+stable. Choose the next absolute checkpoint from the evidence (for example 300,
+500, 750, or 1000), resume the same output directory with `train_iters=1000`,
+retain `eval_interval=25`, `eval_iters=2`, and `save_interval=50`, and set only
+`exit_interval` to that target. Stop on recovered gap, plateau, regression,
+divergence, or iteration 1000. Never raise the 1000-step cap without explicit
+user direction.
```
Edwardf0t1: The continuation path has no data plan, even though the section it depends on is titled "Prepare only the next data tranche."

Section 5 sizes the default 2.6B-token budget for the step-150 gate only (`150 * 512 * 32768` = 2.517B, leaving ~2.3% margin after the 1% `split="99,1,0"` holdout). This paragraph then sends the run toward 300/500/750/1000 while instructing the operator to "set only `exit_interval` to that target" — but 1000 iterations needs `1000 * 512 * 32768` = 16.8B tokens, so the same blend silently repeats roughly 6.5 times. Nothing here tells the operator to prepare the next tranche first.

The alternative isn't free either: extending `--data_paths` on resume changes the Megatron blend while `consumed_train_samples` carries over from the checkpoint, so the resumed run indexes into a differently-built dataset.

Could we pick one and state it explicitly? Either

- size the token budget for the full 1000-step cap up front (and rename the default `output_dir` away from `nemotron-cascade-2-through-150`), or
- document the re-prep procedure before each extension, including what happens to the blend and `consumed_train_samples` on resume, or
- state plainly that repeating the 2.6B corpus is accepted through iteration 1000, so it's a deliberate choice rather than an accident.

.agents/skills/qad/SKILL.md
mxinO: For the day 0 workflow, we can still prompt the agent directly like "try QAD if PTQ has accuracy gap" with more guidance of resources control depending on how many resources we have. Directly allow auto QAD in skill is dangerous, the cost can be high.

.agents/skills/qad/SKILL.md
mxinO: good point, simplified it.

.agents/skills/qad/SKILL.md
mxinO: now will materialize all 1000 step data.

tests/examples/megatron_bridge/test_distill.py
kevalmorabia97, line 1: these tests seem unnecessary and can be removed

.agents/skills/qad/SKILL.md
```diff
@@ -0,0 +1,104 @@
...
+2. **Reproduce PTQ and verify compatibility.** In the target runtime, require
+   `AutoBridge.can_handle()` for the target model and PTQ through `quantize.py`
+   to succeed while preserving the exact preceding PTQ config or recipe:
+   format, layer selection, calibration data/count, sequence length, and seed.
+   A changed quantization setting is a new PTQ candidate and must be evaluated
+   before QAD. In the master-rank `.quant_summary.txt`, require finite positive
+   `amax` for enabled static quantizers; accept `dynamic`/format-defined `None`
+   only when the recipe intends it. Treat the summary as rank-local under model
+   parallelism.
+3. **Choose topology explicitly.** Derive the smallest fitting node count and
```
jenchen13, line 45: why is the sequence length fixed to 32k? the agent should ask the user for their goal for QAD (including what benchmarks) and adjust the sequence length accordingly. for example if user cares about long context benchmarks or SciCode then a longer sequence length is necessary for QAD

.agents/skills/qad/SKILL.md
```diff
@@ -0,0 +1,104 @@
...
+   and available GPU memory. Prefer CP before TP for small long-context models;
+   keep EP=1 for dense models. For MoE require:
+
+   - `DP = world_size / (TP * PP * CP)`
+   - `EDP = world_size / (ETP * EP * PP)`
+   - integral DP/EDP, `num_experts % EP == 0`, and
+     `GBS % (MBS * DP) == 0`
+
+4. **Prepare the full capped dataset once.** Copy
+   `examples/megatron_bridge/data/nemotron-cascade-2-blend.yaml`, set the target
```
jenchen13: the dataset also should not be fixed. it's okay that currently only 1 dataset is supported, but maybe make it clear that we can support more datasets later

jenchen13: why does it say generate random samples for packed 32k sequence and validation data? how is the validation data created

jenchen13 (commented):
overall looks good, just wondering if we can make the QAD recipe more customizable & user driven by having the agent ask the user questions about their QAD goals

.agents/skills/qad/SKILL.md
mxinO: The skill is a initial version for open source models, not a final smart skill that will decide the hyperparameters, the later improvement depends on our resources, if we have enough resources, we can adjust this skill to do more ablations according to the needs. But currently I don't plan to make the first version decide these parameters, the ablations and long seq len can cost huge amount resources. 32k is just a good starting length, enough for many benchmarks.

.agents/skills/qad/SKILL.md
mxinO: Again, see the above comment, we don't allow the ablations here, this dataset is a mixed data containing most important subjects, and probably the best open source SFT data we can have for now. Of course, if we find the QAD is important for open source models, we can add more data options later.

mxinO: The generated random data is a little misleading, it want to say the prepared random sampled data. I will fix this.

tests/examples/megatron_bridge/test_distill.py
mxinO: Agree, removed

Edwardf0t1 (approved):
LGTM - I think it's a handy skill for day0 agentic platform 👍

ChenhanYu (approved):

.agents/skills/qad/SKILL.md
```diff
@@ -0,0 +1,106 @@
...
+   default. Set the target tokenizer and workspace path, then materialize the
+   randomly sampled subset before training. Pack the chosen sequence length;
+   Megatron's `99,1,0` split creates the 1% validation holdout from the same
+   data.
+5. **Run and monitor QAD.** Run one QAD training job at a time and fold startup
+   validation into it; do not submit separate GPU preflight jobs or split at
+   recovery iterations. Let training continue while evaluating saved
+   checkpoints, and cancel it when a stop condition below is met.
+
+## Default training policy
```
kevalmorabia97, line 67: Can we mention GlobalBatchSize recommendation also

examples/megatron_bridge/data/nemotron-cascade-2-blend.yaml
kevalmorabia97, line 1: Do you have any results or comparisons how this data blend compares with other Nemotron PostTraining V3 Datasets?

examples/megatron_bridge/README.md
```diff
@@ -130,6 +130,13 @@ The distillation script expects pre-tokenized data in Megatron's binary format (
 See the **[Dataset Preparation README](../dataset/README.md#tokenizing-for-megatron-frameworks)**
 for full instructions on tokenizing JSONL files and Hugging Face datasets and get the list of output prefixes that you can use for `--data_paths` argument.
 
+If you do not already have a suitable dataset, start with
```
kevalmorabia97: I generally see 768 or 3072 GBS used in Nemotron training recipes. Is 512 also commonly used GBS?

kevalmorabia97: Can you move this under QAD subsection? For distilling pruned models, this wont be sufficient and misleading to put here

tests/examples/megatron_bridge/test_qad.py
```diff
@@ -109,13 +110,18 @@ def test_qad(tmp_path: Path, num_gpus, create_student, is_vlm, is_moe):
         gbs=4,
         train_iters=train_iters,
         lr_warmup_iters=2,
-        eval_interval=train_iters,
+        eval_interval=early_exit_iter,
         eval_iters=1,
+        save_interval=1,
         log_interval=1,
+        exit_interval=early_exit_iter,
+        exit_duration_in_mins=60,
```
kevalmorabia97:
```suggestion
        exit_duration_in_mins=10,
```

examples/megatron_bridge/data/nemotron-cascade-2-blend.yaml
mxinO: Nemotron pt v3 is a collection of many datasets, which require manual blend, cascade 2 is already a blended data tested in Nemotron cascade, so I think as a first version, it's a better starting point.

examples/megatron_bridge/README.md
mxinO: In previous QAD tests, 512 and 1024 has no much difference, I don't think 768 is a magic number. 512 is a safe and cheap option that can help us get results earlier.

.agents/skills/qad/SKILL.md
mxinO: A default GBS is already in the table.

examples/megatron_bridge/README.md
mxinO: Good point, moved, thanks!

kevalmorabia97 (approved):

PR #2023, opened 2026-07-28
Add optional MLflow tracking to hf_ptq.py

cjluo-nv:
### What does this PR do?

Type of change: new feature

Adds `modelopt.torch.utils.mlflow.MlflowRunLogger`, a reusable helper for recording a script run on an MLflow tracking server, and wires `examples/hf_ptq/hf_ptq.py` up to it via `--mlflow <tracking-uri>` so a PTQ run can be reproduced from its MLflow entry alone. Without the flag, behavior is unchanged — every hook is gated on it.

The logger lives in the library rather than the example so other scripts can record runs the same way: it takes a tracking URI, an experiment name and an explicit `enabled` flag, with params, tags and artifacts passed in. `hf_ptq.py` supplies only the PTQ-specific pieces (its params, the resolved recipe, the quantization summaries). `mlflow` is an optional dependency, imported only once tracking is enabled, so it is not a new requirement for the library.

The run is opened **before the model loads**, so a bad URI or an unreachable server fails in seconds rather than after hours of calibration. The invocation and the recipe are uploaded at that point too, which keeps a crashed run useful: it is still recorded, with status `FAILED` and its log attached.

Uploaded artifacts:

| Artifact | Contents |
| --- | --- |
| `command.txt` | The full invocation, copy-pasteable |
| `version.txt` | The ModelOpt version that ran (also a searchable tag) |
| `recipe/resolved_recipe.yaml` | The `--recipe` with `$import`s expanded |
| `logs/hf_ptq.log` | Everything the run printed, including a crash traceback |
| `summary/quant_summary.txt` | Per-quantizer summary (unless `--no-verbose`) |
| `summary/moe.html` | Per-expert calibration token counts, when the run produces them |

Plus model / format / calibration settings as searchable params, and `user` / `hostname` / `modelopt_version` / `git_sha` tags.

Three design points worth review:

1. **The recipe is uploaded resolved, not verbatim.** A recipe may be a directory or use `$import`s, so the source file is not self-contained. For `huggingface/qwen3_6_moe/auto_quantize/w4a16_nvfp4_fp8_at_6p0bits-active_moe` the source is 2,230 B / 58 lines against 7,563 B / 308 lines resolved — the raw file records under 30% of what actually ran.
2. **`hf_ptq.py` has no logging framework** (bare `print()`), so the log is produced by teeing stdout/stderr. Handlers that libraries bound to `sys.stderr` at import time are re-pointed at the tee for the run's duration and handed back afterwards; without that, `transformers` / `huggingface_hub` warnings reach the console but never the log. Native (C-level) output is still not captured — documented in the README.
3. **The recipe upload lives in the caller, not the library.** That keeps `modelopt.recipe` out of `modelopt.torch.utils`, which would otherwise risk a `modelopt.torch.utils` → `modelopt.recipe` → `modelopt.torch.quantization` → `modelopt.torch.utils` import cycle.
4. **MLflow failures never fail the quantization.** Startup validation is fatal by design (it is before any GPU work); the end-of-run upload is best-effort.

Only the main rank uploads, so `--use_fsdp2` runs produce a single run.

### Usage

```bash
python hf_ptq.py \
  --pyt_ckpt_path <huggingface_model_card> \
  --recipe general/ptq/nvfp4_default-kv_fp8_cast \
  --export_path <quantized_ckpt_path> \
  --mlflow https://<your-mlflow-server>/
```

```
[mlflow] experiment: $USER/hf_ptq/<checkpoint basename>-<recipe name>
[mlflow] run: https://<your-mlflow-server>/#/experiments/13/runs/c243352e...
```

`--mlflow_experiment` and `--mlflow_run_name` override the defaults (`$USER/hf_ptq/<basename>-<recipe name or --qformat>`, and the UTC start time). Passing `--mlflow` with no value uses `$MLFLOW_TRACKING_URI`. Authentication uses MLflow's own env vars.

### Testing

**Unit** — 51 tests in `tests/unit/torch/utils/test_mlflow.py` for the library, plus 13 in `tests/examples/hf_ptq/test_hf_ptq_args.py` for the hf_ptq wiring. CPU-only, no network and no `mlflow` dependency (driven against a stub module). Covers experiment-name derivation and sanitization, URI accept/reject, tee pass-through, the pre-bound-handler redirect, artifact renaming, skipping absent optional outputs, the disabled path, and `version.txt`. 85 tests pass together with the existing `test_hf_ptq_args.py` / `test_example_utils.py`.

**Hardware** — real PTQ runs against a live MLflow server:

| Run | Result |
| --- | --- |
| Qwen3-0.6B, NVFP4 PTQ, 1×B200 | `FINISHED`, all artifacts, sane post-quant generations |
| Qwen3.6-35B-A3B MoE, AutoQuantize `w4a16_nvfp4_fp8_at_6p0bits-active_moe`, 2×B200 | `FINISHED` in 63 min, search hit `effective bits: 6.00`; 106 KB log capturing every per-layer decision, 4.4 MB quant summary |
| Qwen3.6-35B-A3B, plain NVFP4 PTQ, 2×B200 | `FINISHED` |
| Qwen3-0.6B re-run after the library move, 1×H200 | `FINISHED`, all five artifacts including `version.txt` |
| Run **without** `--mlflow` after the review fixes | exactly 1 `[load_recipe]` line and 0 `[mlflow]` lines, confirming the untracked path is untouched |
| Two runs sharing one `--export_path`, second crashed early | second run uploads **no** summary — the first run's 124 KB file on disk is correctly not attributed to it, and its traceback is in the log |
| Crash mid-run (gated HF dataset) | `FAILED` recorded with log + traceback attached, summaries correctly absent |
| Malformed URI | Rejected by `argparse` with a `Did you mean https://…?` hint |
| Unreachable host | Fails in 9.9 s total, before any model load |
| No `--mlflow` | Exit 0, no MLflow output, unchanged export |

**Coverage gap, stated plainly:** `summary/moe.html` is verified only against a synthetic file (unit test + a real upload). It could not be produced naturally — `expert_token_count` buffers live on `_QuantSparseSequentialMoe`, while Qwen3.5/3.6 experts take the fused `_QuantFusedExperts` path, so no such file is written for these models regardless of `--moe_calib_experts_ratio`. The uploader's conditional is correct; the branch simply had no natural input available here.

- Is this change backward compatible?: ✅ — new optional flags only; no `--mlflow` means no behavior change.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ — adds `mlflow` as an optional extra in `pyproject.toml` (`nvidia-modelopt[mlflow]`, folded into `all`) and to `examples/hf_ptq/requirements.txt`. Apache-2.0 (permissive). Imported lazily, so it is not required to install or import ModelOpt. No code copied from other sources.
- Did you write any new necessary tests?: ✅ — 29 new unit tests.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — 0.47 New Features.
- Did you get Claude approval on this PR?: ❌ — `/claude review` not yet run. A self-review was done first and its six findings are fixed in the third commit (the notable one: gathering the MLflow inputs re-read the recipe on *every* run, including without `--mlflow`).

### Additional Information

The one deliberate coverage gap is `summary/moe.html`, described under Testing: no model available here takes the sparse-sequential MoE path that writes it, so it is covered by unit test and a synthetic upload rather than a natural one. The uploader treats it as an optional output and skips it when absent, which is exercised by test.


examples/hf_ptq/hf_ptq.py
```diff
@@ -1658,6 +1700,62 @@ def parse_args() -> argparse.Namespace:
     return args
 
 
+def _mlflow_run_inputs(args: argparse.Namespace) -> tuple[dict, dict]:
+    """Params and start-time artifacts describing this PTQ run."""
+    params = {
```
cjluo-nv: can this table be auto generated?

examples/hf_ptq/README.md
```diff
@@ -293,6 +293,53 @@ scripts/huggingface_example.sh --model <model> --quant nvfp4 --vlm --calib_with_
...
+  --pyt_ckpt_path <huggingface_model_card> \
+  --recipe general/ptq/nvfp4_default-kv_fp8_cast \
+  --export_path <quantized_ckpt_path> \
+  --mlflow https://<your-mlflow-server>/
+```
+
+The run is opened *before* the model loads, so a bad URI or a missing token fails within
+seconds rather than after a full calibration. These artifacts are uploaded:
+
+| Artifact | Contents |
```
cjluo-nv: can we hide this table by default to make the readme visualization shorter.

pyproject.toml
```diff
@@ -126,7 +126,11 @@ dev-test = [
     "torch-geometric",
 ]
 # Compound extras via self-references
-all = ["nvidia-modelopt[onnx,hf,puzzletron]"]
+mlflow = [ # Dependencies for modelopt.torch.utils.mlflow run tracking
+    "mlflow",
+]
+
+all = ["nvidia-modelopt[onnx,hf,puzzletron,mlflow]"]
 dev = ["nvidia-modelopt[all,dev-docs,dev-lint,dev-test]"]
```
cjluo-nv: let's add mlflow to dev as well.

meenchen (commented):

**Design review (protocol applied)**

*Problem:* PTQ runs today leave no durable record — command, recipe, log and summaries live only in whatever terminal/SLURM file the user kept — so a finished quantization can't be reproduced or compared later.

*Alternatives I checked:*
- **In-repo:** `modelopt/torch/utils/logging.py` already owns all the stdout/stderr plumbing in this package (`capture_io`, `no_stdout`, `atomic_print`). `TeeStream` is a fourth, unrelated mechanism for the same surface and would fit better there. The repo *also* already has an MLflow convention on the eval side (`.agents/skills/evaluation/references/nel-next.md`, `.agents/skills/accessing-mlflow/SKILL.md`): `experiment_name: <user>/<model>`, tags `model` / `checkpoint_path` / `framework`, plus an `invocation_id` ID convention. This PR invents a second scheme (`$USER/hf_ptq/<basename>-<variant>`, tags `user`/`hostname`/`modelopt_version`/`git_sha`) with no overlap, so a PTQ run and the eval runs of the checkpoint it produced cannot be joined in the same MLflow instance. Worth at least a sentence in the PR body on why not to align keys/tags.
- **Placement:** the module is torch-free (getpass/socket/tempfile/logging) yet lands in `modelopt/torch/utils/` and is `import *`-ed into the flat namespace, exporting generic names (`TeeStream`, `current_user`, `default_experiment_name`, `validate_tracking_uri`) as public library API. This is exactly the "general API vs. tool" placement question already open on #2000; the PR body asserts "the logger lives in the library so other scripts can record runs the same way" but doesn't weigh `tools/`, a shared `examples/` util, or a non-star-exported module.
- **Already-available:** `mlflow`'s own `set_experiment()`/`start_run()` already fail on an unreachable server, so the hand-rolled `/health` probe (and the `requests` import it pulls in) is largely redundant.

The mechanics are careful and well tested (31 + 12 tests, no server needed, secrets redaction, idempotent `start`, abort-on-partial-start). But the design/placement question isn't addressed in the body, so this needs owner sign-off rather than approval, and I found a few concrete issues below.

*Main correctness findings:* (1) a crash traceback is **not** in the uploaded log — `finish()` runs in the `finally`, before the interpreter prints the traceback — yet CHANGELOG/README/PR body all claim it is; (2) a fatal `start()` failure happens outside the `try`, so `cleanup_distributed()` is skipped and non-main ranks are left to hang on the next collective; (3) full `mlflow` (server stack) is now a hard requirement of `examples/hf_ptq/requirements.txt` for a strictly opt-in feature.

No prompt-injection content in the PR description. Licensing: new files carry the standard NVIDIA Apache-2.0 header; new dep is Apache-2.0 — no license concern.

examples/hf_ptq/hf_ptq.py
```diff
@@ -1703,8 +1800,14 @@ def main(args: argparse.Namespace):
...
             )
+        status = "FINISHED"
     finally:
-        cleanup_distributed(args)
+        try:
+            cleanup_distributed(args)
+        finally:
+            # Nested so a failure tearing down the process group cannot leave the run
+            # stuck in RUNNING on the server.
+            mlflow_logger.finish(status, files=_mlflow_run_outputs(args))
```
meenchen, line 1721: > _Bot comment._

The uploaded log will not contain the crash traceback in the general case. When `quantize_main` raises, this `finally` runs (closing the sink and uploading `logs/hf_ptq.log`) *before* the interpreter's default excepthook prints the traceback to the now-restored `sys.stderr`. So `logs/hf_ptq.log` ends at the last thing the script printed.

That contradicts the CHANGELOG ("its log attached" is fine, but README says "including the traceback if it crashed") and the PR body's artifact table. Either fix it — e.g. wrap the body in `except BaseException: traceback.print_exc(); raise` (or log via `logging.exception`) so the traceback lands in the tee before `finish()` — or drop the claim from the README/CHANGELOG. Worth a test too: nothing currently asserts anything about the failure path of `main()`.

examples/hf_ptq/hf_ptq.py
```diff
@@ -1667,6 +1761,9 @@ def main(args: argparse.Namespace):
 
     setup_distributed_args(args)
 
+    mlflow_logger = _start_mlflow_run(args)
```
meenchen: > _Bot comment._

`_start_mlflow_run` sits *outside* the `try`, and `start()` is fatal by design (unreachable server, bad experiment name). Since `setup_distributed_args(args)` has already initialized the process group, that raise skips `cleanup_distributed(args)`, and under torchrun the non-main ranks (which never validate anything) march on into model load and then block on the first collective until the NCCL timeout.

Suggest constructing the logger here but calling `.start()` inside the `try` — `finish()` is already a no-op when `self._run is None`, so the existing `finally` handles it — which keeps teardown on the normal path.

examples/hf_ptq/requirements.txt
```diff
@@ -1,5 +1,6 @@
 compressed-tensors
 fire
 flash-attn>=2.6.0
+mlflow
```
meenchen: > _Bot comment._

This makes the full `mlflow` distribution (flask, gunicorn, sqlalchemy, alembic, docker, pandas/scipy/pyarrow, …) a hard install requirement for *every* hf_ptq user, for a feature that is opt-in behind `--mlflow`. Two cheaper options:

1. Use `mlflow-skinny` — the client-only package, which is all `MlflowRunLogger` needs (`set_tracking_uri`/`set_experiment`/`start_run`/`log_*`/`end_run`) — here and in the `mlflow` extra in `pyproject.toml`.
2. Or leave it out of `requirements.txt` entirely; `_open_run()` already raises a precise `pip install nvidia-modelopt[mlflow]` hint.

pyproject.toml
```diff
@@ -126,8 +126,12 @@ dev-test = [
...
 ]
 # Compound extras via self-references
-all = ["nvidia-modelopt[onnx,hf,puzzletron]"]
-dev = ["nvidia-modelopt[all,dev-docs,dev-lint,dev-test]"]
+mlflow = [ # Dependencies for modelopt.torch.utils.mlflow run tracking
+    "mlflow",
+]
+
+all = ["nvidia-modelopt[onnx,hf,puzzletron,mlflow]"]
+dev = ["nvidia-modelopt[all,mlflow,dev-docs,dev-lint,dev-test]"]
```
meenchen, line 137: > _Bot comment._

`all` now already includes `mlflow`, so listing it again in `dev = ["nvidia-modelopt[all,mlflow,...]"]` is redundant — drop it to keep the compound extras readable.

modelopt/torch/utils/__init__.py
```diff
@@ -23,6 +23,7 @@
 from .list import *
 from .logging import *
 from .loss_mask import *
+from .mlflow import *
```
meenchen: > _Bot comment._

Star-importing this module publishes `TeeStream`, `current_user`, `default_experiment_name` and `validate_tracking_uri` as top-level `modelopt.torch.utils` API. Those are very generic names for MLflow-specific helpers (`current_user` / `validate_tracking_uri` especially), and they become a compatibility surface we have to keep.

Suggest either narrowing `__all__` to `MlflowRunLogger` or not star-importing at all and letting callers do `from modelopt.torch.utils.mlflow import MlflowRunLogger, default_experiment_name, validate_tracking_uri` (hf_ptq is the only caller today). This also ties into the placement question in the summary — nothing in the module is torch-specific.

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,472 @@
...
+        self._log_path = Path(tempfile.mkdtemp(prefix="modelopt-mlflow-")) / f"{script}.log"
+        sink = open(self._log_path, "w", buffering=1, encoding="utf-8")
+        stdout, stderr = sys.stdout, sys.stderr
+        self._saved_streams = (stdout, stderr, sink)
+        sys.stdout, sys.stderr = TeeStream(stdout, sink), TeeStream(stderr, sink)
+        self._redirect_log_handlers({stdout: sys.stdout, stderr: sys.stderr})
+        print(f"[mlflow] capturing this run's log to {self._log_path}")
+
+    def _redirect_log_handlers(self, replacements: dict) -> None:
+        """Point already-configured logging handlers at the tee.
```
meenchen: > _Bot comment._

Two things here:

1. `import requests` is a function-local import with no comment. It looks like the reason is the same as the `mlflow` import above (requests is not a hard modelopt dependency and this module is star-imported from `modelopt.torch.utils.__init__`), so please say so inline — otherwise it reads as an unjustified local import.
2. `_check_reachable` is largely redundant: `mlflow.set_experiment()` on the next lines already hits the server and raises on an unreachable host, and the probe adds an assumption about a `/health` endpoint (fine behind auth since any response passes, but it's extra surface). If the motivation is only latency, `MLFLOW_HTTP_REQUEST_MAX_RETRIES`/timeout env vars get you the same fail-fast without the extra dependency and code path — worth a note in the docstring if you keep it.

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,472 @@
...
+    """
+
+    def __init__(self, stream, sink):
+        """Wrap *stream*, mirroring everything written to it into the open file *sink*."""
+        self._stream = stream
+        self._sink = sink
+
+    def write(self, data: str) -> int:
+        """Write to both the original stream and the sink."""
+        self._stream.write(data)
```
meenchen: > _Bot comment._

`TeeStream` is generic stream plumbing and overlaps in spirit with `modelopt/torch/utils/logging.py`, which already owns every other `sys.stdout` manipulation in this package (`capture_io`, `no_stdout`, `atomic_print`). Consider moving it there (and keeping `mlflow.py` about MLflow) so there's one place to look for stdout redirection semantics — `capture_io` in particular is the near-neighbour that redirects instead of tees.

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,472 @@
...
+    """Build an experiment name of the form ``<user>/<tool>/<model>-<variant>``.
+
+    Only the basename of *model* is used, so a local checkpoint directory and an
+    ``org/name`` Hugging Face id collapse to the same readable name; *variant* is whatever
+    distinguishes this run of *tool* on *model*, such as a recipe name or a quantization
+    format. Each component is reduced to ``[A-Za-z0-9._-]`` so the ``/`` separators stay
+    meaningful, and *user* defaults to the current user.
+
+    Example:
+        >>> default_experiment_name("hf_ptq", "/models/Qwen3-0.6B/", "nvfp4", user="alice")
```
meenchen, line 131: > _Bot comment._

Naming/tag convention: the repo's existing MLflow usage on the eval side (`.agents/skills/evaluation/references/nel-next.md`) uses `experiment_name: <user>/<model>` with tags `model` / `checkpoint_path` / `framework`, and an `invocation_id` tag/param convention (`.agents/skills/accessing-mlflow/SKILL.md`). This adds a third component (`<tool>`) and a disjoint tag set, so a PTQ run and the eval runs of the checkpoint it produced can't be correlated in the same tracking server. Consider emitting `model` / `checkpoint_path` tags alongside the current ones, and note the naming choice in the PR body.

pyproject.toml
```diff
@@ -126,8 +126,12 @@ dev-test = [
     "torch-geometric",
 ]
 # Compound extras via self-references
-all = ["nvidia-modelopt[onnx,hf,puzzletron]"]
-dev = ["nvidia-modelopt[all,dev-docs,dev-lint,dev-test]"]
+mlflow = [ # Dependencies for modelopt.torch.utils.mlflow run tracking
+    "mlflow",
+]
+
+all = ["nvidia-modelopt[onnx,hf,puzzletron,mlflow]"]
```
kevalmorabia97: lets not make mlflow requirement part of `all`. It adds a lot of heavy dependencies so best to leave it explicitly opt-in or part of `dev`

pyproject.toml
kevalmorabia97: Better to add it only in `dev` and not `all` so default `nvidia-modelopt[all]` installation doesnt require heave mlflow dependencies

Edwardf0t1 (commented):
Three inline comments below. Everything else I found was minor cleanup.

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,472 @@
...
+        # The version is a tag as well, for searching; the artifact travels with the run.
+        self._log_texts(
+            {
+                "command.txt": _command_text(),
+                "version.txt": f"{modelopt.__version__}\n",
+                **(texts or {}),
+            }
+        )
+
+    def _log_outputs(self, texts, files, metrics) -> None:
```
Edwardf0t1, line 440: **The crash traceback does not actually reach the uploaded log.**

The README says `logs/hf_ptq.log` contains "the traceback if it crashed", and the PR description's test table says the crash run had "log + traceback attached". For an exception that propagates out of `main()` that isn't true: the interpreter's `sys.excepthook` prints the traceback *after* the `finally` completes, i.e. after `_log_outputs()` has already uploaded the file and `_stop_capture()` has restored the streams. I confirmed with a standalone repro of the tee — the captured log ends at the last `print()` before the raise.

What the gated-HF-dataset run most likely showed was `huggingface_hub`'s own *logged* error, which is captured (thanks to the pre-bound-handler redirect in `_redirect_log_handlers`) — not the Python traceback.

`sys.exc_info()` is populated during a `finally` that runs while an exception is propagating, so the fix is local:

```python
def _log_outputs(self, texts, files, metrics) -> None:
    exc = sys.exc_info()
    if exc[0] is not None:
        # Printed into the tee before it is torn down; the interpreter's own
        # traceback is emitted after finish() returns and would not be captured.
        traceback.print_exception(*exc)
    elapsed = time.time() - self._start_time
    ...
```

Worth a test that raises through `main()`'s try/finally and greps the uploaded log — that is the one claim in the PR with no test behind it, which is why it slipped through.

examples/hf_ptq/hf_ptq.py
Edwardf0t1: **`cleanup_distributed()` is skipped when MLflow startup fails.**

This call sits between `setup_distributed_args(args)` and the `try`, and it raises by design — `ImportError` when `mlflow` is absent, `ConnectionError` when the server is unreachable. When it does, rank 0 exits with the process group still up, so `cleanup_distributed(args)` in the `finally` never runs.

Fail-fast before `load_model()` is the right call; it just wants to be inside the `try`. Constructing the logger inert first keeps `mlflow_logger` bound for the `finally`:

```python
mlflow_logger = MlflowRunLogger(
    args.mlflow, args.mlflow_experiment,
    run_name=args.mlflow_run_name,
    enabled=bool(args.mlflow) and args.dist_state.is_main,
)
status = "FAILED"
try:
    _start_mlflow_run(mlflow_logger, args)   # still before load_model()
    ...
```

Minor, same area: `files=_mlflow_run_outputs(args)` is evaluated *inside* the `finally`, so if it ever raised it would replace the in-flight exception. It can't today (`--export_path` defaults to `"exported_model"`, so `Path()` can't fail), but computing it before the `try` costs nothing and removes the hazard.

pyproject.toml
Edwardf0t1: **Prefer `mlflow-skinny`, and reconsider folding it into `all`.**

`pip install -U "nvidia-modelopt[all]"` is the documented user install (`docs/source/getting_started/_installation_for_Linux.rst:100`), so this line puts the full MLflow **server** stack — Flask, gunicorn, sqlalchemy, alembic, docker, graphene, pyarrow, matplotlib — into the default install path for every user. None of it is needed by a tracking *client*, and each is a fresh conflict surface for a package that already pins onnxruntime/transformers tightly.

`mlflow-skinny` is exactly that client — `start_run`, `log_params`, `log_text`, `log_artifact`, `set_experiment` are all in the fluent API — and it is what the `mlflow` distribution itself depends on, so there is no conflict in naming it directly. Suggest it in both the extra and `examples/hf_ptq/requirements.txt`, with a floor:

```toml
mlflow = [ # Dependencies for modelopt.torch.utils.mlflow run tracking
    "mlflow-skinny>=2.9",
]

all = ["nvidia-modelopt[onnx,hf,puzzletron]"]
```

I would also leave `mlflow` out of `all` entirely: it is example-only tracking, and `nvidia-modelopt[mlflow]` is already the install path the `ImportError` in `_open_run` points users at.

Also on line 134: `dev = ["nvidia-modelopt[all,mlflow,...]"]` lists `mlflow` redundantly, since `all` already pulls it in (that becomes moot if `all` drops it).

modelopt/torch/utils/logging.py
```diff
@@ -219,5 +220,40 @@ def custom_showwarning(message, category, filename, lineno, file=None, line=None
             warnings.showwarning = original_showwarning
 
 
+class TeeStream:
```
chadvoegele, line 223: This feels like an agent shortcut hack to me. I'd much rather just replace `print()` in `hf_ptq.py` with a proper `logger`. That would give us all of the benefits of `logging` library over `print` including the stream redirection you want to do for `mlflow`.

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,452 @@
...
+    return cleaned[:_MAX_COMPONENT_LEN] or "unknown"
+
+
+def _git_sha() -> str:
+    """Short commit of the ModelOpt source, or ``"unknown"`` outside a git checkout.
+
+    Read out of ``.git`` rather than by shelling out to ``git``, which keeps the library
+    free of subprocess use.
+    """
+    git_dir = Path(__file__).resolve().parents[3] / ".git"
```
chadvoegele: My agent is saying this doesn't work with `git` worktrees.

> The advertised Git SHA is always unknown in worktrees — modelopt/torch/utils/mlflow.py:150-163
> Worktrees use a .git file pointing to the actual git directory, but _git_sha() assumes .git is a directory. Its test explicitly permits unknown, masking this.

I wouldn't say it's such a big deal, but I'm using `git` worktrees all of the time now to have multiple agents working on changes in the same repo simultaneously.

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,452 @@
...
+                return sha[:9]
+    except OSError:
+        pass
+    return "unknown"
+
+
+def _command_text() -> str:
+    """The invocation, as a copy-pasteable line."""
+    lines = [shlex.join([sys.executable, *_redact_argv(sys.argv)])]
+    world_size = int(os.environ.get("WORLD_SIZE", "1"))
```
chadvoegele, line 194: What about just logging the environment separately from the command?

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,452 @@
...
+            self._mlflow.log_artifact(str(local), artifact_path=directory)
+            return
+        # log_artifact keeps the local basename, so rename via a staged copy rather than
+        # reading the file into memory -- these can be hundreds of MB.
+        with tempfile.TemporaryDirectory() as staging:
+            staged = Path(staging) / target.name
+            shutil.copy2(local, staged)
+            self._mlflow.log_artifact(str(staged), artifact_path=directory)
+
+    def _start_capture(self) -> None:
```
chadvoegele, line 478: See earlier `logger` comment.

examples/hf_ptq/hf_ptq.py
```diff
@@ -1667,7 +1767,15 @@ def main(args: argparse.Namespace):
...
     setup_distributed_args(args)
 
+    mlflow_logger = _mlflow_logger(args)
+
+    status = "FAILED"
     try:
+        # Opened inside the try: start() is fatal by design, and skipping
+        # cleanup_distributed would leave the other ranks blocked on the first collective
+        # until the NCCL timeout.
+        _start_mlflow_run(mlflow_logger, args)
```
chadvoegele: Could we make this a context manager so we can just do `with _mlflow_logger(args): ...`?

To mirror, this upstream behavior:

```
with mlflow.start_run() as run:
    mlflow.log_param("my", "param")
    mlflow.log_metric("score", 100)
```

https://mlflow.org/docs/latest/api_reference/python_api/mlflow.html

pyproject.toml
cjluo-nv: Done in 05d0a6bd1 — `all` is back to `["nvidia-modelopt[onnx,hf,puzzletron]"]`, with a comment recording why mlflow is deliberately excluded.

Also switched the extra itself to `mlflow-skinny>=2.9`, so even the opt-in install no longer pulls Flask/gunicorn/sqlalchemy/alembic/docker. Verified skinny is sufficient rather than assuming it: installed it alone (no full `mlflow` present) and ran a real PTQ against our tracking server — all five artifacts uploaded.

pyproject.toml
cjluo-nv: Resolved in 05d0a6bd1, in the direction you and @Edwardf0t1 suggested: `all` drops mlflow entirely, and `dev` keeps it listed explicitly. That also settles the opposite request on this line from another reviewer — dropping it from `all` makes the `dev` entry load-bearing rather than redundant.

modelopt/torch/utils/mlflow.py
cjluo-nv: Fixed in 7861dcba4, which landed after your review — and your diagnosis was exactly right, including the part about the gated-HF-dataset run. That run was showing `huggingface_hub`'s *logged* error (captured via `_redirect_log_handlers`), not a Python traceback. I had asserted "traceback attached" from the artifact list without ever opening the log.

One deliberate deviation from your patch: I append `traceback.format_exc()` to the tee's **sink only**, rather than calling `traceback.print_exception(*exc)`. `print_exception` writes to `sys.stderr`, which is still teed at that point, so the traceback would reach the log *and* the console — and the interpreter then prints it again once `finish()` returns, so the user sees it twice. Sink-only keeps the console at exactly one copy.

Two tests now cover it: one asserting the traceback is present in the *uploaded* artifact, one exercising it through the new context manager.

examples/hf_ptq/hf_ptq.py
cjluo-nv: Fixed in 7861dcba4 and restructured further in 05d0a6bd1. `main()` now wraps the tracked section in a context manager *inside* the `try`, so `cleanup_distributed()` runs whether or not opening the run fails:

```python
try:
    with _mlflow_run(args):
        ...
finally:
    cleanup_distributed(args)
```

Your secondary point is fixed as a side effect: `files=_mlflow_run_outputs(args)` used to be evaluated inside the `finally`, and the artifact paths are now computed before the block — so there is no longer any way for them to replace an in-flight exception.

pyproject.toml
cjluo-nv: All three done.

`mlflow-skinny>=2.9` in both `pyproject.toml` and `examples/hf_ptq/requirements.txt` (7861dcba4, floor added in 05d0a6bd1), and mlflow is out of `all` entirely — `dev` keeps it, matching @kevalmorabia97's read.

I verified skinny rather than reasoning about it, since it was the assumption I originally hedged on when picking the full distribution: installed `mlflow-skinny` alone into a clean container and ran a real PTQ against our tracking server. All five artifacts uploaded and read back, including the renamed summary and the run log.

examples/hf_ptq/hf_ptq.py
cjluo-nv: Fixed in 7861dcba4 — you were right, and the claim was wrong in the README, the CHANGELOG *and* the PR body.

`finish()` now appends the active exception to the tee's sink before uploading, when the status is not `FINISHED`. Written to the sink only, so the console still shows the traceback exactly once via the interpreter. Verified end-to-end against a live server on a failing run: `log has traceback: True`, where it was `False` before.

You were also right that nothing tested the failure path — that is why it slipped through. There are now tests asserting the traceback is in the uploaded artifact, and covering the same thing through the context manager.

examples/hf_ptq/hf_ptq.py
cjluo-nv: Fixed in 7861dcba4, then simplified further in 05d0a6bd1. The logger is built outside the `try` and the run is opened inside it, so a fatal tracking error still reaches `cleanup_distributed()` instead of leaving the other ranks blocked on the first collective until the 2-hour NCCL timeout this file sets.

examples/hf_ptq/requirements.txt
cjluo-nv: Done — `mlflow-skinny>=2.9` in both `requirements.txt` and the extra (7861dcba4, floor in 05d0a6bd1). I verified skinny actually uploads artifacts to an HTTP tracking server before switching, since that was the specific thing I was unsure of when I first picked the full distribution.

Kept it in `requirements.txt` rather than dropping it (your option 2) so `pip install -r requirements.txt` leaves `--mlflow` working out of the box; it is now a light client-only dependency.

pyproject.toml
cjluo-nv: Resolved in 05d0a6bd1, though in the opposite direction to the one you suggested: rather than dropping `mlflow` from `dev`, `all` no longer includes it (per @kevalmorabia97 and @Edwardf0t1 — `nvidia-modelopt[all]` is the documented install path). So the `dev` entry is now load-bearing rather than redundant.

modelopt/torch/utils/__init__.py
cjluo-nv: Done in 7861dcba4 — the star-import is gone, so `current_user`, `validate_tracking_uri` and `default_experiment_name` are no longer top-level `modelopt.torch.utils` API. `hf_ptq` imports from `modelopt.torch.utils.mlflow` directly.

There is precedent for the omission: `distributed`, `memory_monitor` and `plugins` are not star-imported from that `__init__` either.

On placement — agreed nothing in the module is torch-specific. I left it under `modelopt/torch/utils/` because there is no non-torch `modelopt/utils/` package today and creating one felt out of scope here, but happy to move it if you would prefer that.

modelopt/torch/utils/mlflow.py
cjluo-nv: Both addressed in 7861dcba4.

The `import requests` now carries a comment naming the reason (it arrives with mlflow, not with modelopt, and this module is imported for its optional-dependency path).

I kept `_check_reachable`, and documented why in its docstring: `set_experiment()` does reach the server, but it goes through mlflow's retry loop, so a typo'd hostname takes appreciably longer to surface than a direct DNS failure. Fail-fast before the model loads was an explicit design goal here — the whole point is that a bad URI costs seconds rather than hours of calibration. Any HTTP response, including 401, counts as reachable, so it does not assume anything about auth.

modelopt/torch/utils/mlflow.py
cjluo-nv: Done in 7861dcba4 — `TeeStream` now lives in `modelopt/torch/utils/logging.py` next to `capture_io` and `no_stdout`, and is exported from its `__all__`. `mlflow.py` imports it from there. Agreed that one place for stdout-redirection semantics is better, and it shortened `mlflow.py` usefully.

modelopt/torch/utils/mlflow.py
cjluo-nv: Good catch on the divergence. I have taken the additive half in 7861dcba4: PTQ runs now emit `model` and `checkpoint_path` tags alongside the existing ones, so a PTQ run and the evaluations of the checkpoint it produced can be found together.

I kept the `<user>/<tool>/<model>-<variant>` experiment name, because the `hf_ptq` segment was specifically requested by the PR author — the intent being that a PTQ run and an eval run of the same model do not collide in one namespace. Happy to revisit if you would rather match `<user>/<model>` exactly; the tags are the part that actually enables correlation either way, and those are in now.

No `invocation_id` yet — hf_ptq has no equivalent identifier to key on. If there is a convention you would like PTQ to adopt, point me at it and I will add it.

modelopt/torch/utils/logging.py
cjluo-nv: Fair challenge, and I agree `logging` would be better for `hf_ptq` on its own merits. Two reasons I would rather not fold it into this PR:

**It would not remove the tee.** `tqdm` writes progress bars straight to `sys.stderr`, and `transformers` / `huggingface_hub` both `print` *and* log. A `logging.FileHandler` captures neither. The tee — plus the pre-bound-handler redirect in `_redirect_log_handlers` — is what got `huggingface_hub`'s rate-limit warnings into the captured log at all; before that they reached the console and vanished. So a logging conversion would sit alongside this, not replace it.

**Scope.** 34 `print()` calls in `hf_ptq.py`, 86 across `examples/hf_ptq/`, in a PR already around +1400.

Happy to do the conversion as a follow-up PR — it improves `hf_ptq` independently of tracking. Let me know if you would rather it block this one.

modelopt/torch/utils/mlflow.py
cjluo-nv: Your agent is right, and it is worse than reported — fixed in 05d0a6bd1.

Verified against a real worktree: `.git` is a *file* naming the worktree's gitdir, `HEAD` lives there, but the ref itself lives in the main checkout via `commondir: ../..`. So handling only the file-vs-directory part still returns `unknown`; it needs both levels of indirection.

This was a regression I introduced. `git rev-parse` handled all of it, and I replaced it with a hand-rolled reader to clear a Bandit `# nosec` finding — trading one review comment for another. You are also right that the test masked it: it accepted `sha == "unknown" or <hex>`, so it passed either way. It now asserts a real sha across a plain checkout, a worktree and a detached HEAD.

modelopt/torch/utils/mlflow.py
cjluo-nv: Decided against this one, on the author's call, and I think the reasoning is worth recording: environment variables are exactly where credentials live — `MLFLOW_TRACKING_TOKEN`, `HF_TOKEN`, cloud keys — and this module's own docstring tells users to pass credentials *via the environment*. Uploading the environment would undermine the advice the module gives.

The `WORLD_SIZE` / `LOCAL_WORLD_SIZE` note stays inline in `command.txt`, where it exists only to explain that the `torchrun` wrapper is not part of `sys.argv` and so is not shown in the command above it.

modelopt/torch/utils/mlflow.py
cjluo-nv: Replied on the `logging.py` thread — short version: agreed in principle, but a `logging` conversion would not remove the need for the tee (`tqdm` writes directly to `sys.stderr`, and the HF libraries both print and log), and it is ~86 `print()` calls across `examples/hf_ptq/`. Offering it as a follow-up PR.

examples/hf_ptq/hf_ptq.py
cjluo-nv: Done in 05d0a6bd1. `MlflowRunLogger.track()` is a context manager mirroring `mlflow.start_run()`, and `hf_ptq` now reads:

```python
try:
    with _mlflow_run(args):
        ...
finally:
    cleanup_distributed(args)
```

The block's outcome sets the run status, which removed the manual `status = "FAILED"/"FINISHED"` bookkeeping and a nested `try/finally`. It also fixed a separate hazard @Edwardf0t1 flagged: the closing artifact paths are now computed before the block, instead of inside a `finally` where a raise would have replaced the in-flight exception.

meenchen (commented):

Re-review of the MLflow tracking PR (+1457/-25, 9 files). All of the previously flagged correctness bugs now look genuinely fixed; what's left is a design/placement decision that two codeowners raised and the author deliberately deferred, so this needs a human sign-off rather than an approval.

**Verified as addressed (verified against the current diff, not just the replies):**
- Crash traceback now reaches the uploaded log — `finish()` calls `_note_active_exception()` (writes `traceback.format_exc()` into the tee's sink only, before `_log_outputs`), with two tests (`test_failed_run_uploads_the_traceback`, `test_track_marks_a_raising_block_failed`) and a negative test for the FINISHED path.
- `cleanup_distributed()` no longer skipped: `with _mlflow_run(args):` is inside the `try`, so a fatal `start()` still unwinds through the `finally`; `files=` is computed before the block, so it can't replace an in-flight exception.
- Partial-start orphan run fixed (`_abort_run()` ends the run FAILED, tested).
- Secret redaction: `command.txt`, params and `run_url` masked; dash-prefixed secret values now masked; URI userinfo masked; parametrized tests cover all cases.
- `# nosec` suppressions gone; `_git_sha()` reads `.git` directly and now handles worktrees (`gitdir:` file + `commondir` indirection) and detached HEAD, with real tests for all three (the old test that accepted `"unknown"` is fixed).
- `TeeStream` moved into `modelopt/torch/utils/logging.py`; the `mlflow` star-import is out of `modelopt/torch/utils/__init__.py`.
- Deps: `mlflow-skinny>=2.9`, removed from `all`, kept in `dev` — matches kevalmorabia97/Edwardf0t1.
- README table collapsed into `<details>`, log-artifact wording qualified, params table auto-derived from `vars(args)`.
- Licensing: new files use the canonical 2026 `LICENSE_HEADER` verbatim; new dep is Apache-2.0. No license concern.

**Why this needs an owner look rather than an approve:**
- :speech_balloon: chadvoegele asked to replace `hf_ptq.py`'s bare `print()` with a real `logger` instead of teeing streams ("feels like an agent shortcut hack"). Author replied that a `logging` conversion wouldn't remove the tee (`tqdm` writes straight to `sys.stderr`; HF libs both print and log) and offered it as a follow-up (~86 `print()` calls). That rationale is plausible, but the `TeeStream` + `_redirect_log_handlers` machinery (re-pointing other libraries' pre-bound `StreamHandler`s at our stream for the duration of a run) is the riskiest part of the PR and the codeowner who objected hasn't signed off.
- :speech_balloon: Placement: the module is torch-free yet lives in `modelopt/torch/utils/`; the author explicitly said "happy to move it if you would prefer". That is an open question the owner must answer, and `TeeStream` is now exported from `logging.__all__`, i.e. `modelopt.torch.utils.TeeStream` becomes public API under a very generic name.
- :speech_balloon: Experiment/tag convention vs. the eval side (`.agents/skills/accessing-mlflow`, `.../evaluation/references/nel-next.md`): author added `model` / `checkpoint_path` tags (good, tested) but kept `<user>/hf_ptq/<model>-<variant>` and no `invocation_id`, deferring the naming choice to the reviewers. Worth an explicit decision so PTQ and eval runs land in a coherent namespace.
- :speech_balloon: `_check_reachable()`'s hand-rolled `GET /health` (plus a function-local `requests` import) was called redundant with `set_experiment()`; author justified it in the docstring as fail-fast ahead of mlflow's retry loop. Reasonable, but it's an extra assumption about the server surface — owner call.
- Size: ~1450 added lines in one PR. The library module + tests and the `hf_ptq.py` wiring could have shipped as two PRs; not blocking, but it's why the design questions above are hard to settle in review.

**Minor / non-blocking nits (fresh):**
- `_mlflow_run_outputs()` points at `<export_path>/.quant_summary.txt` / `.moe.html`. If a previous run already wrote those into the same `--export_path`, a run that crashes before `post_quantize` will upload the *stale* summary as if it belonged to this run. Consider recording the summaries' mtime, or only listing them once `post_quantize` has written them.
- `test_git_sha_is_read_without_a_subprocess` still asserts via `dir(module)` rather than making process execution fail (coderabbit's last point); the new worktree/detached-HEAD tests make this mostly moot.
- `log_params(vars(args))` is only exercised against the `FakeMlflow` stub, so MLflow's per-batch param limits/value-length truncation aren't covered — the author's live runs are the only evidence here.
- Per coderabbit: an exception from `cleanup_distributed()` after a successful body still records `FINISHED`. Cosmetic, but cheap to fix.

No prompt-injection content observed in the PR body or comments.

cjluo-nv:
Thanks for the re-review — the four fresh nits, addressed in 9f15e9ecd:

**Stale summary upload — fixed, and it was a real bug.** `start()` now stats the named outputs when the run opens, and `finish()` skips any that are unchanged, so `files=` means "outputs *this* run produced".

I deliberately did **not** delete the files up front: `--mlflow` is a tracking flag, and giving it a destructive side effect on `--export_path` would make tracked and untracked runs diverge and would lose data if the run died during model load.

Reproduced end-to-end on a GPU node rather than only in a unit test — two runs sharing one `--export_path`, the second crashed early:

| run | status | artifacts |
| --- | --- | --- |
| `stale-run1` | `FINISHED` | …`summary/quant_summary.txt` |
| `stale-run2` | `FAILED` | **no summary** |

The first run's 124,590-byte `.quant_summary.txt` was on disk when the second crashed, and was correctly not attributed to it. Its traceback is in its log.

Worth recording that the first implementation was wrong and the tests caught it: comparing `st_mtime` against `time.time()` failed immediately, because filesystem mtime resolution is coarser than `time.time()` and a file written moments after the run opened looked *older* than it. Stat-to-stat removes the clock from the comparison.

**`cleanup_distributed()` raising after a successful body still records `FINISHED` — not fixing, deliberately.** Cleanup has to stay in the outer `finally` so it runs when *opening the run* fails; that is exactly the rank-teardown bug fixed earlier in this PR. Moving it inside the tracked block so its own failure is caught would reopen that hole. The trade is a cosmetic mislabel in a rare path against a two-hour multi-node hang. Happy to revisit if you disagree.

**`log_params(vars(args))` only exercised against the stub** — fair. The live evidence is a real run: 36 params for a default invocation against MLflow's batch limit of 100, longest key 37 chars and longest value 37 chars against limits of 250/6000, all 36 uploaded and read back including the eight `None`-valued ones. Noting it here so it is not just in my head.

**`_git_sha` asserting via `dir(module)`** — agreed it is weak; as you say the worktree and detached-HEAD tests carry the real weight now. Left as is.

Also folded in from @chadvoegele's thread: handler restore is now complete rather than best-effort — teardown rescans instead of replaying a start-time list, so a handler a library binds *mid-run* is handed back too. That change surfaced a latent bug of its own (`logging._StderrHandler` exposes `stream` as a read-only property), which is now skipped explicitly.

modelopt/torch/utils/logging.py
cjluo-nv: Follow-up: the "restore is incomplete" half of this is now fixed in 9f15e9ecd.

Teardown rescans for handlers pointing at the tee rather than replaying the list captured on the way in, so a handler bound by a library *during* the run is handed back too — previously those were left pointing at a file about to close, which is why `TeeStream` needed to tolerate a closed sink at all. The bookkeeping list is gone.

That change immediately exposed a bug of its own: `logging._StderrHandler` exposes `stream` as a read-only property that already resolves to whatever `sys.stderr` currently is, so `setStream()` raises on it. The teardown scan runs while `sys.stderr` *is* the tee, so it hits that handler where the start-time scan did not. It is now skipped explicitly — correct on the merits, since that handler follows the tee unaided.

On the blast-radius concern more broadly, I checked the property rather than assuming it: with a capture active, a child process's output is **not** captured and the child runs normally, while our prints and library logs both are. The tee acts on Python objects; subprocesses inherit file descriptors. Under `torchrun` each rank is its own process and only rank 0 enables any of this — so nothing outside the process is affected, and in-process state is restored on the way out.

The `logging` conversion is still on offer as a follow-up if you would like it.

meenchen (commented):

Re-review of the MLflow tracking PR at the latest commit (+1530/-25, 9 files). The two fresh items from the previous round are genuinely fixed, and I could not find a new correctness bug — what remains is a set of design/placement calls that two codeowners raised and the author has explicitly deferred to them, so this wants a human sign-off rather than an approval.

**Verified fixed in the latest commit (checked against the diff, not just the replies):**
- Stale-artifact bug: `start()` now records `(st_mtime_ns, st_size)` for every named output and `_log_outputs()` skips entries whose stat is unchanged, so `files=` really means "outputs this run produced". Covered by `test_only_files_this_run_produced_are_uploaded` (stale `.quant_summary.txt` not uploaded, `.moe.html` written mid-run is) plus the GPU repro in the reply.
- Handler restore is now complete: `_stop_capture()` rescans for handlers pointing at the tee instead of replaying a start-time list, so handlers bound mid-run are handed back (`test_capture_hands_back_handlers_bound_during_the_run`). The `logging._StderrHandler` read-only `stream` property is skipped via `contextlib.suppress(AttributeError)`; I checked `Handler.setStream` — the lock is released in its `finally`, so the suppression doesn't leak a lock.
- Earlier round's blockers all still hold up: traceback into the tee sink before upload (`_note_active_exception`, two tests), `with _mlflow_run(args)` inside the `try` so `cleanup_distributed()` always runs, `files=` computed before the block, `_abort_run()` for partial start, redaction of argv/params/`run_url` with parametrized tests (including `--hf_token -secret`), `_git_sha()` worktree + `commondir` + detached HEAD with real tests, `TeeStream` moved to `modelopt/torch/utils/logging.py`, star-import removed from `modelopt/torch/utils/__init__.py`, `mlflow-skinny>=2.9` out of `all` and kept in `dev`.
- Licensing: `modelopt/torch/utils/mlflow.py` and the new test file carry the canonical `LICENSE_HEADER` text verbatim (2026); `logging.py`'s 2024 header is pre-existing and untouched. New dep is Apache-2.0. No license concern.

**Why owner sign-off rather than approve (design review protocol):** the PR body does argue three design points (resolved-recipe upload, tee-vs-logging, recipe upload in the caller to avoid an import cycle), but the questions raised in review are still open by the author's own choice:
- :speech_balloon: chadvoegele asked for `hf_ptq.py` `print()` → `logger` instead of stream teeing ("agent shortcut hack"). Author replied that a `logging` conversion wouldn't remove the tee (`tqdm` writes straight to `sys.stderr`; HF libs both print and log) and offered it as a follow-up (~86 `print()`s) — plausible, but `TeeStream` + `_repoint_handlers` (re-pointing *other libraries'* handlers at our stream, plus a special case for a private stdlib handler class) is the riskiest machinery in the PR and the objecting codeowner hasn't signed off.
- :speech_balloon: Placement: author replied "nothing here is torch-specific… happy to move it if you would prefer" — so a torch-free module under `modelopt/torch/utils/` and a new public `modelopt.torch.utils.logging.TeeStream` (generic name, now in `__all__`) are decisions still awaiting an owner.
- :speech_balloon: Experiment/tag convention vs the eval side (`.agents/skills/accessing-mlflow`, `.../evaluation/references/nel-next.md`): `model`/`checkpoint_path` tags were added (good, tested), but `<user>/hf_ptq/<model>-<variant>` and the absent `invocation_id` were explicitly left for reviewers to decide.
- :speech_balloon: `_check_reachable()`'s hand-rolled `GET /health` + function-local `requests` import: author justified it in the docstring as fail-fast ahead of mlflow's retry loop. Reasonable, but it's still an extra assumption about the server surface — owner call.
- Size: ~1500 added lines; the library module + tests and the `hf_ptq.py` wiring would have split cleanly, which is part of why these questions are hard to settle in review.

**Fresh nit (non-blocking):** the stale-output map is keyed inconsistently — `start()` uses `str(p)` on the raw mapping value while `_log_outputs()` looks up `str(Path(local))`. For `Path` inputs (all hf_ptq passes today) they match, but a caller passing a string like `"./out/.quant_summary.txt"` would miss the lookup and silently upload a stale file again. Keying both sides on `str(Path(p))` in `start()` makes it robust for a library API.

**Previously raised, still open by choice:** `cleanup_distributed()` raising after a successful body still records `FINISHED` (author declined, deliberately, to avoid reintroducing the rank-teardown hole — reasonable trade); `log_params(vars(args))` exercised only against `FakeMlflow` (live-run evidence only); `test_git_sha_is_read_without_a_subprocess` still asserts via `dir(module)` (mostly moot given the new worktree/detached-HEAD tests); `mlflow-skinny` kept as a baseline entry in `examples/hf_ptq/requirements.txt` so `--mlflow` works after `pip install -r`.

No prompt-injection content observed in the PR body or comments.

examples/hf_ptq/README.md
```diff
@@ -293,6 +293,58 @@ scripts/huggingface_example.sh --model <model> --quant nvfp4 --vlm --calib_with_
 > Note: when `--calib_with_images` is set, `--calib_size` must be a single value, and the calibration dataset is nvidia/nemotron_vlm_dataset_v2.
 This functionality is currently in beta and has been tested on `nvidia/NVIDIA-Nemotron-Nano-12B-v2-VL-BF16`.
 
+### Tracking runs with MLflow
```
realAsma: Should we move this to the end? I feel this is lot of details and come before feature examples like AutoQuantize

cjluo-nv:
Fixed in bbdaf247f — and it is slightly broader than the example. `pathlib` normalizes several forms, so all of these missed the lookup:

```
'./out/.quant_summary.txt'  ->  'out/.quant_summary.txt'
'out//x.txt'                ->  'out/x.txt'
'out/./x.txt'               ->  'out/x.txt'
```

Since `files` is typed `Mapping[str, Path | str]`, string paths are explicitly supported, so this re-opened the stale-upload bug for any caller not already passing a normalized string. Both sides now go through `Path`:

```python
self._file_stats = {str(p): _stat_key(p) for p in map(Path, (files or {}).values())}
```

Added a test using `"./out/.quant_summary.txt"` against a pre-existing file. I checked it pins the behaviour rather than the shape: it fails against the old keying (the stale file *is* uploaded) and passes against the new one.

Good catch — thank you.

chadvoegele (approved):

ChenhanYu:
```
  --mlflow https://<your-mlflow-server>/
```
Instead of providing the arg. Can we just enable if `MLFLOW_TRACKING_URI` is set? This this the official env variable. So log if `MLFLOW_TRACKING_URI` is set.

cjluo-nv:
Done in d40f75129 — `MLFLOW_TRACKING_URI` now enables tracking on its own, with `--mlflow` still available and taking precedence when both are given.

| | result |
| --- | --- |
| `--mlflow <uri>` | tracked, URI from the flag |
| `MLFLOW_TRACKING_URI` set, no flag | **tracked** |
| both | flag wins |
| neither | unchanged |

One thing I did *not* make identical between the two, and would like a sanity check on. Enabling on an ambient variable means one exported for something else starts governing PTQ runs — and this repo's own `.agents/skills/` tell users to export `MLFLOW_TRACKING_URI` for eval workflows, so running evals and then `hf_ptq` in the same shell is an ordinary thing to do. As written before this change, a missing `mlflow` package or an unreachable server was **fatal**, which would have turned an unrelated env var into a failed quantization.

So provenance is tracked and the paths differ:

- **`--mlflow`** — an unambiguous request. A bad URI, a missing client or an unreachable server still fails loudly.
- **`MLFLOW_TRACKING_URI`** — ambient. The same failures warn and the run continues untracked.

Verified both on real PTQ runs:

```
### env var only, NO --mlflow flag
[mlflow] FINISHED: https://mlflow-modelopt.nvidia.com/#/experiments/12/runs/25ec1730a75e...

### unusable env var (file:///local/mlruns)
UserWarning: Ignoring MLFLOW_TRACKING_URI, continuing untracked: ... must be http(s) ...
Quantized model exported to: /tmp/envbad.   exit=0
```

Implemented as a `required=` flag on `MlflowRunLogger` (default `True`), so the strict behaviour is unchanged for every other caller. 92 tests pass.

Happy to make the environment path strict too if you would rather it fail loudly — it is a one-line change.

modelopt/torch/utils/mlflow.py
```diff
@@ -0,0 +1,535 @@
...
+        """Fail fast on an unreachable host.
+
+        Any HTTP response -- including 401 -- means the host is up, so authorization is
+        left to the first real API call, which reports it precisely.
+        """
+        # Optional dependency, like mlflow itself: requests arrives with it, not with modelopt.
+        import requests
+
+        try:
+            requests.get(f"{self.tracking_uri}/health", timeout=10)
```
meenchen: **[P1] Preserve MLflow TLS configuration in the fail-fast probe**

This direct `requests.get()` runs before the MLflow client and does not apply `MLFLOW_TRACKING_SERVER_CERT_PATH`, `MLFLOW_TRACKING_CLIENT_CERT_PATH`, or `MLFLOW_TRACKING_INSECURE_TLS`. A tracking server using a private CA or mTLS therefore fails the TLS handshake here even though the configured MLflow client can connect. With an explicit `--mlflow` this aborts PTQ; with an environment-derived URI it disables tracking for a valid server.

Please perform the readiness request through `MlflowClient(tracking_uri=...)` using the MLflow transport and timeout/retry configuration, or remove this parallel probe and rely on the first real MLflow request. Please also add coverage showing that custom CA or client-certificate configuration is not bypassed.

examples/hf_ptq/example_utils.py
```diff
@@ -1070,3 +1079,119 @@ def resolve_checkpoint_dir(quant_cfg: dict, model_path: str) -> tuple[dict, str]
...
+        tags=_mlflow_run_tags(args),
+        texts=texts,
+        files=_mlflow_run_outputs(args),
+    )
+
+
+def _mlflow_run_tags(args: argparse.Namespace) -> dict[str, str]:
+    """Tags shared with the evaluation side, so a PTQ run and the evaluations of the
+    checkpoint it produced can be found together on one tracking server."""
+    return {"model": Path(args.pyt_ckpt_path).name, "checkpoint_path": args.pyt_ckpt_path}
```
meenchen: **[P1] Tag the produced checkpoint, not the source checkpoint**

The docstring says this tag lets a PTQ run and evaluations of the checkpoint it produced be found together, but `checkpoint_path` currently records the input. For example, a run reading `/models/Qwen3-8B` and exporting `/exports/Qwen3-8B-nvfp4` is tagged with `/models/Qwen3-8B`, while evaluation records `/exports/Qwen3-8B-nvfp4`; the values never join. The `model` tag only groups these runs broadly and cannot identify which quantized output was evaluated.

Please use `checkpoint_path=args.export_path` and retain the input separately as `source_checkpoint_path=args.pyt_ckpt_path`. The test should pass distinct input and export paths and assert both tags, using the same output-path normalization convention as the evaluation side.

examples/hf_ptq/README.md
cjluo-nv: Agreed — moved in bb9be6043.

It is now a top-level `## Tracking runs with MLflow` section just before `## Resources`, so the whole thing sits after AutoQuantize, Real Quant, FSDP2, Evaluate Accuracy and Exporting Checkpoints rather than interrupting them. I also added a row to the contents table at the top so relocating it does not bury it.

The artifact table inside the section is still collapsed behind `<details>` from an earlier round, so the section is short on the page even where it now sits.

cjluo-nv:
Small follow-up on the interface, since it changed after my comment above.

The bare `--mlflow` form (no value, meaning "use `$MLFLOW_TRACKING_URI`") is **gone** as of bb9be6043. It is redundant now that no flag at all does exactly that, and it was the only reason the resolution logic needed a three-way branch — `nargs="?"` produced an empty-string state on top of "absent" and "a real URI". Without it the whole thing is:

```python
args.mlflow_required = args.mlflow is not None
args.mlflow = args.mlflow or os.environ.get("MLFLOW_TRACKING_URI") or None
```

A bare `--mlflow` now gets argparse's own `expected one argument`. Nothing is lost: every combination that used to work still works.

Same commit also moves the tracking wiring out of `hf_ptq.py` into `example_utils.py`, next to the other hf_ptq helpers — `hf_ptq.py` is left with `add_mlflow_args(parser)`, `resolve_mlflow_args(args, parser)` and `with mlflow_run(args):`, about 127 lines lighter.

modelopt/torch/utils/mlflow.py
cjluo-nv: Fixed in 6053a81b5 — you are right, and I removed the probe rather than teaching it TLS.

Making it TLS-aware would mean re-implementing MLflow's own configuration handling (`MLFLOW_TRACKING_SERVER_CERT_PATH`, `MLFLOW_TRACKING_CLIENT_CERT_PATH`, `MLFLOW_TRACKING_INSECURE_TLS`), which would drift from the client. Opening the run is now the readiness check: MLflow's own first request, so it honours its transport by construction. `requests` is no longer imported by the module at all.

That also retires three other things reviewers had raised separately — the `/health` endpoint assumption, the unjustified function-local `requests` import, and your earlier point that the probe was largely redundant with `set_experiment()`.

**Measured the trade rather than assuming it**, since fail-fast was the reason the probe existed. On a dead host:

| | before | after |
| --- | --- | --- |
| fails before the model loads | yes | **yes** (0 model-loading lines) |
| time to fail | 9.9 s | **4 m 24 s** |

So the guarantee holds and only the wait changes, because MLflow retries where the probe timed out once. `MLFLOW_HTTP_REQUEST_MAX_RETRIES` shortens it and is named in the docstring. I judged correctness worth the error-path latency — a misconfigured URI costs minutes instead of surfacing after hours of calibration — but flagging the number explicitly in case 4.5 minutes is too long for your taste; capping retries by default is a small change if so.

On coverage for the TLS configuration itself: there is no test asserting a custom CA or client cert is honoured, because after this change the module contains no TLS handling to bypass — the client does all of it. Happy to add one if you would still like it pinned.

examples/hf_ptq/example_utils.py
cjluo-nv: Fixed in 6053a81b5 — and you are right that the docstring was describing something the code did not do.

I checked the convention rather than guessing: `.agents/skills/evaluation/` uses `deployment.checkpoint_path=/path/to/quantized/checkpoint`, so the eval side's `checkpoint_path` is the **produced** checkpoint. Mine recorded the input, so the two values could never join — exactly as you describe.

```python
checkpoint_path         # resolved --export_path      (what an evaluation is pointed at)
source_checkpoint_path  # --pyt_ckpt_path             (the input, kept separately)
model                   # unchanged, broad grouping
```

Resolved to absolute, because `--export_path` defaults to the relative `exported_model` and a relative path is useless as a join key. Two tests: one passing distinct input and export paths and asserting all three tags, one asserting the tag is absolute when the default relative path is used.

No `invocation_id` yet — hf_ptq has no equivalent identifier. If PTQ should mint or accept one, point me at the convention and I will add it.

meenchen (approved):
LGTM

kevalmorabia97 (approved):
Approving as pyproject.toml codeowner

PR #2024, opened 2026-07-28
[OMNIML-5613] Quantize ResNet residual adds in torch ONNX example

ajrasane:
### What does this PR do?

Type of change: Bug fix

Adds recipe-backed FP8 and INT8 residual quantization for timm ResNet models in the torch ONNX example:

- Adds FP8 and INT8 PTQ recipes that enable a shortcut quantizer immediately before each residual `Add`.
- Inserts shortcut quantizers after ModelOpt module conversion so recipes configure and calibrate them in the normal quantization pass.
- Adds `--recipe` support for PTQ and AutoQuantize recipes and renames `--quantize_mode` to `--qformat`.
- Verifies all 16 ResNet-50 residual additions have shortcut Q/DQ immediately before the `Add`.

### ResNet support scope

ResNet and other convolutional architectures are supported only with FP8 and INT8. AutoQuantize, MXFP8, NVFP4, and INT4_AWQ are not supported for ResNet because TensorRT has limited convolution kernel support. Transformer architectures containing individual Conv2d layers continue to use format-specific Conv overrides.

### Usage

```bash
python examples/torch_onnx/torch_quant_to_onnx.py \
    --timm_model_name=resnet50 \
    --recipe=timm/resnet/ptq/fp8 \
    --onnx_save_path=resnet50.onnx
```

Use `timm/resnet/ptq/int8` for INT8. Without `--recipe`, `--qformat` selects a built-in quantization preset.

### Testing

- All configured pre-commit hooks passed, including recipe schema and license validation.
- Focused AutoQuantize recipe mapping regression passed.
- FP8/INT8 recipe export coverage verifies all 16 ResNet-50 shortcut Q/DQ pairs.
- TensorRT engine builds passed for FP8 and INT8 on Ada.

- Did you update the changelog?: ✅


examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -218,6 +220,58 @@ def _disable_low_channel_conv_input_quantizers(model):
...
+        activation.residual_quantizer = TensorQuantizer(
+            QuantizerAttributeConfig(num_bits=num_bits, axis=None)
+        ).to(next(block.parameters()).device)
+        activation.register_forward_pre_hook(_quantize_residual_input)
+        residual_quantizers.append(activation.residual_quantizer)
+
+    if not residual_quantizers:
+        return
+
+    for quantizer in residual_quantizers:
```
cjluo-nv: > _Bot comment._

This block duplicates `_calibrate_uncalibrated_quantizers` (a few dozen lines below) and costs a second full pass over the calibration set for ResNet, since `quantize_model` already ran one for the FP8 Conv overrides.

Two cheaper options:
- Attach the quantizer as `input_quantizer` on the activation module and create it *before* `quantize_model` — then `_calibrate_uncalibrated_quantizers` and `_disable_dead_quantizers` both pick it up automatically, and this whole calibration block goes away.
- Or use the framework path: `QuantModuleRegistry.register({nn.ReLU: "nn.ReLU"})(QuantInputBase)` (see `modelopt/torch/quantization/nn/modules/quant_activations.py`, which does exactly this for `nn.LeakyReLU`) plus a `{"parent_class": "nn.ReLU", "quantizer_name": "*input_quantizer", "cfg": {...}}` entry restricted to `*act3*`/`*act2*`, so `mtq.quantize`'s forward loop calibrates it and `print_quant_summary` reports it.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -218,6 +220,58 @@ def _disable_low_channel_conv_input_quantizers(model):
...
+    model.eval()
+    try:
+        with torch.no_grad():
+            for batch in data_loader:
+                model(batch["image"] if isinstance(batch, dict) else batch)
+    finally:
+        model.train(was_training)
+
+    for quantizer in residual_quantizers:
+        quantizer.load_calib_amax()
```
cjluo-nv: > _Bot comment._

Two robustness gaps relative to the sibling helper in this file:

1. `load_calib_amax()` is strict here, whereas `_calibrate_uncalibrated_quantizers` uses `load_calib_amax(strict=False)`; a block that saw no data will raise instead of being skipped.
2. These quantizers are created *after* `_disable_dead_quantizers` runs (inside `quantize_model`), and that helper only looks at `input_quantizer`/`output_quantizer`/`weight_quantizer` — so a residual quantizer that calibrates to `amax == 0` or NaN reaches the FP8 exporter and hits the `scale = 448 / amax` blow-up the guard was added for.

Suggest either naming the attribute so the existing helpers cover it, or applying the same `amax <= 0 / isnan` disable check to `residual_quantizers` after `load_calib_amax(strict=False)`.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -218,6 +220,58 @@ def _disable_low_channel_conv_input_quantizers(model):
...
+
+
+def _add_resnet_residual_quantizers(model, quantize_mode, auto_quantization_formats, data_loader):
+    if quantize_mode == "int8":
+        num_bits = 8
+    elif quantize_mode in ("fp8", "mxfp8", "nvfp4"):
+        # Dynamic block quantizers do not support the residual path's 4D tensors.
+        num_bits = (4, 3)
+    elif quantize_mode == "auto":
+        activation_formats = set(auto_quantization_formats) - {"INT4_AWQ_CFG"}
```
cjluo-nv: > _Bot comment._

In `auto` mode this infers the residual format from the *requested* search space rather than what `auto_quantize` actually selected per block, so a block that ended up INT8 can get an FP8 residual Q/DQ (or vice versa) — the kind of precision mismatch the rest of this file works hard to avoid for TRT. Since this runs after the search, the per-block answer is available, e.g. `getattr(block.conv3, "input_quantizer", None)._num_bits` (fall back to the current heuristic if the quantizer is absent/disabled).

tests/examples/torch_onnx/test_torch_quant_to_onnx.py
```diff
@@ -29,11 +32,30 @@
...
 
+def _assert_residual_adds_are_quantized(onnx_save_path):
+    model = onnx.load(onnx_save_path)
+    consumers = defaultdict(list)
+    for node in model.graph.node:
+        for input_name in node.input:
+            consumers[input_name].append(node)
+
+    residual_adds = [node for node in model.graph.node if node.op_type == "Add"]
+    assert len(residual_adds) == 16
```
cjluo-nv, line 54: > _Bot comment._

`residual_adds` is every `Add` node in the graph, so this asserts "ResNet-50 exports exactly 16 Adds and all of them are residual". That's incidentally true today, but it will fail confusingly if the exporter ever emits an unrelated `Add` (bias/fused pattern, autocast helper, etc.). Prefer identifying the residual adds structurally (e.g. Add whose single consumer is the Relu that follows the block) and then asserting the count of *those* is 16.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -218,6 +220,58 @@ def _disable_low_channel_conv_input_quantizers(model):
                 q.disable()
 
 
+def _quantize_residual_input(module, inputs):
+    return (module.residual_quantizer(inputs[0]),)
+
+
+def _add_resnet_residual_quantizers(model, quantize_mode, auto_quantization_formats, data_loader):
```
cjluo-nv: > _Bot comment._

Nit: every other helper in this module has a docstring explaining the TRT/architecture reason it exists (see `_disable_low_channel_conv_input_quantizers`). Please add one here covering why only `act2`/`act3` are hooked, why FP8 per-tensor is used instead of MXFP8/NVFP4 blocks for the 4D residual tensor, and why `int4_awq` is skipped.

cjluo-nv (commented):

Small, focused change (+90/-2) that adds per-block residual-add quantizers to the timm ResNet path of the torch→ONNX example, with CHANGELOG + README updates and an ONNX-level assertion in the existing example test. No licensing concerns; no prompt-injection content in the PR metadata. A few things worth addressing before merge:

1. **Duplicated calibration logic / extra data pass.** `_add_resnet_residual_quantizers` re-implements the enable_calib → forward-loop → `load_calib_amax` dance that `_calibrate_uncalibrated_quantizers` (same file) already performs, and adds a *second* full pass over the calibration set for ResNet (the first happens inside `quantize_model` for the FP8 Conv overrides). If the residual quantizer were attached as `input_quantizer` on the activation module and created *before* `quantize_model`, both the calibration and the dead-quantizer guard would come for free from the existing helpers. Even more idiomatic: modelopt already supports this via `QuantModuleRegistry.register({nn.ReLU: "nn.ReLU"})(QuantInputBase)` (exactly what `modelopt/torch/quantization/nn/modules/quant_activations.py` does for `nn.LeakyReLU`) plus a `{"parent_class": "nn.ReLU", "quantizer_name": "*input_quantizer"}` config entry, which gets calibrated by `mtq.quantize`'s forward loop and is visible to `mtq.print_quant_summary` / modelopt state. Please either reuse one of these paths or note in the PR/comment why the manual hook is needed.

2. **Residual quantizers bypass the file's own `amax<=0`/NaN guard, and `load_calib_amax()` is strict.** `_disable_dead_quantizers` only inspects `input_quantizer`/`output_quantizer`/`weight_quantizer`, and it runs inside `quantize_model` — i.e. before these quantizers exist. A residual quantizer that calibrates to `amax == 0` (or NaN) will therefore reach the FP8 exporter, which is precisely the `scale = 448 / amax` division the guard exists to prevent. Also, unlike `_calibrate_uncalibrated_quantizers`, this code calls `load_calib_amax()` without `strict=False`, so any block that didn't see data raises.

3. **`auto` mode picks the residual format from the *search space*, not the *search result*.** After `mtq.auto_quantize`, each block's actual format is known (e.g. `block.conv3.input_quantizer._num_bits`); deriving `num_bits` from the union of requested formats can give an FP8 residual Q/DQ next to an INT8-quantized block (or the reverse), which is what the rest of this file goes to some length to avoid for TRT.

Minor: the two new functions are the only helpers in this file without docstrings; and `assert len(residual_adds) == 16` asserts that *every* `Add` in the exported graph is a residual add, which will break confusingly if the exporter ever emits an unrelated `Add` — consider filtering to the 16 residual adds (e.g. by producer/consumer pattern) before the count assertion.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Fixed. For INT8, FP8, MXFP8, and NVFP4, the recognized shortcut input quantizers are now installed after module conversion but before the normal quantization call. Fixed modes therefore use the existing calibration flow instead of a separate full pass. Auto remains post-search because its per-block format is only known after selection.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Fixed. The fixed modes now use standard calibration. The post-search Auto path uses `strict=False` and explicitly disables shortcut quantizers with missing, NaN, or non-positive `amax`.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Fixed. Auto now reads the selected terminal convolution input quantizer for each block (`conv3` for `Bottleneck`, `conv2` for `BasicBlock`) and mirrors its INT8 or FP8 format, with the requested-format heuristic only as the suggested absent/disabled fallback.

tests/examples/torch_onnx/test_torch_quant_to_onnx.py
ajrasane: Fixed. The test now selects residual additions structurally by their single `Relu` consumer before asserting the expected count of 16.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Fixed. Added a rationale docstring covering shared block-input Q/DQ, projection placement, the per-tensor FP8 fallback for 4D MXFP8/NVFP4 shortcuts, and the INT4-AWQ exclusion.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -151,6 +160,27 @@ def get_quant_config(quantize_mode, recipe=None):
...
+    for block in (module for module in model.modules() if isinstance(module, block_types)):
+        if block.downsample is None:
+            block.downsample = torch.nn.Sequential()
+        elif not isinstance(block.downsample, torch.nn.Sequential):
+            block.downsample = torch.nn.Sequential(block.downsample)
+        if "residual_quantizer" in block.downsample._modules:
+            continue
+        residual_quantizer = TensorQuantizer()
+        residual_quantizer.disable()
+        block.downsample.add_module("residual_quantizer", residual_quantizer)
```
cjluo-nv, line 178: > _Bot comment._

The `amax <= 0`/NaN guard still doesn't cover these quantizers: `_disable_dead_quantizers` only looks at `getattr(mod, attr)` for `attr in ("input_quantizer", "output_quantizer", "weight_quantizer")`, and the new attribute is `downsample.residual_quantizer`. The previous round's explicit "disable shortcut quantizers with missing/NaN/non-positive amax" code was dropped in the rewrite, so a residual quantizer that calibrates to `amax == 0` now reaches `export_fp8`'s `scale = 448 / amax` — exactly the failure that helper exists to prevent (unlikely on pretrained weights, but the guard was added for random-init/zero-init cases and this test suite runs `--no_pretrained`).

One-line fix: add `"residual_quantizer"` to the attribute tuple in `_disable_dead_quantizers` (or have it iterate `TensorQuantizer` modules directly, which would make it name-agnostic).

Also, nit carried over from the last round: this and `_get_resnet_recipe` are now the only helpers in this module without docstrings. Worth restating why the quantizer goes at the end of `downsample` (so it lands immediately before the `Add`), why `downsample=None` is converted to an empty `Sequential`, and why only fp8/int8/auto get residual quantizers.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -385,26 +416,48 @@ def auto_quantize_model(
...
+        disabled_layers = None
+        method = "gradient"
+    else:
+        auto_config = recipe.auto_quantize
+        constraints = auto_config.constraints.model_dump(exclude_none=True)
+        if effective_bits is not None:
+            constraints["effective_bits"] = effective_bits
+        format_configs = []
+        fixed_quantization_config = recipe.quantize.model_dump()
+        module_search_spaces = [
```
cjluo-nv: > _Bot comment._

This recipe → `mtq.auto_quantize` mapping duplicates `examples/hf_ptq/hf_ptq.py::_mtq_inputs_from_auto_quantize_config`, and the copy is lossy:

- `auto_config.score_size` is ignored — the new recipe declares `score_size: 128` but the example passes `--num_score_steps` instead, so that recipe field is dead config (it happens to match the CLI default today, which hides the bug).
- `auto_config.cost_excluded_layers` is ignored (hf_ptq maps it into `constraints["cost"]["excluded_module_name_patterns"]`).
- `auto_config.candidate_formats` (top-level, no `module_search_spaces`) is ignored: for such a recipe this builds `quantization_formats=[]` + `module_search_spaces=[]`, and `mtq.auto_quantize` then raises `"fixed_quantization_config requires at least one explicit module_search_spaces entry"`. That's brittle for anything added to `_RESNET_RECIPES` later.

Since this mapping is now needed by two examples, consider promoting it into `modelopt/recipe` (importable) and calling it from both, rather than maintaining a second partial copy.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -571,6 +631,10 @@ def main():
         )
         print(f"Base Model - Top-1 Accuracy: {top1:.2f}%, Top-5 Accuracy: {top5:.2f}%")
 
+    resnet_recipe = None if args.recipe else _get_resnet_recipe(model, args.quantize_mode)
```
cjluo-nv: > _Bot comment._

For ResNet + `--quantize_mode=auto`, `args.auto_quantization_formats` is now silently ignored (the recipe's `module_search_spaces` replaces it), and since `--recipe` is rejected with `--quantize_mode=auto` there is no way to opt out of the ResNet recipe. Please either warn when a non-default `--auto_quantization_formats` is discarded, or allow `--recipe` (incl. an explicit "none") in auto mode. Same applies to the fp8/int8 paths: `get_quant_config`'s overrides are bypassed for ResNet without any log line saying which recipe was chosen (`load_recipe` does print the path, so this is a smaller concern there).

tests/examples/torch_onnx/test_torch_quant_to_onnx.py
```diff
@@ -48,6 +81,9 @@ def test_torch_onnx(model_key, quantize_mode):
     cmd_parts.extend(["--no_pretrained", "--trt_build"])
     run_example_command(cmd_parts, "torch_onnx")
 
+    if model_key == "resnet50" and quantize_mode in _RESNET_QUANT_MODES:
```
cjluo-nv: > _Bot comment._

This test runs every `(model, mode)` combination with `--trt_build`, including `resnet50` + `auto`, which after this PR searches FP8 vs INT8 per block with the residual pinned to FP8. The PR body states that this INT8/FP8 mix is only supported by TensorRT on Blackwell and newer, and the listed validation is "TensorRT engine builds passed for FP8 and INT8 on Ada" — i.e. the auto engine build appears untested there. Note also that both candidates cost the same 8 effective bits under the `effective_bits: 8.0` constraint, so the solver is free to mix formats arbitrarily.

How is this combination expected to pass on the CI GPUs? If they're pre-Blackwell, either skip `--trt_build` for resnet50+auto (with a comment explaining why) or constrain the recipe so a single format is selected.

cjluo-nv (commented):

Re-review of the reworked (recipe-driven) implementation. The approach changed substantially since the last round, and most of the earlier findings are genuinely resolved:

**Resolved**
- *Duplicated calibration / second data pass* (critical): the manual `enable_calib` → forward-loop → `load_calib_amax` block is gone. `_add_resnet_residual_quantizers` now runs as a `CUSTOM_POST_CONVERSION_PLUGINS` callback, i.e. after `_replace_quant_module` but before `set_quantizer_by_cfg` (see `modelopt/torch/quantization/conversion.py::replace_quant_module`), so the recipe's `'*residual_quantizer'` entry configures it and `mtq.quantize`'s own forward loop calibrates it. The quantizers are also visible to `print_quant_summary` and modelopt state. Good fix.
- *Strict `load_calib_amax()`* (critical): moot — standard calibration path is used now.
- *Auto-mode residual format inferred from the search space* (critical): replaced by an explicit design — the residual quantizer is pinned to FP8 via `fixed_quantization_config` in the AutoQuantize recipe, documented in the README and PR body. Acceptable as an intentional choice (see one follow-up below).
- *Test asserted "every `Add` is residual"* (minor): the test now selects residual adds structurally by their single `Relu` consumer before the count assertion.

**Still open / new**
1. `_disable_dead_quantizers` still only inspects `input_quantizer` / `output_quantizer` / `weight_quantizer`, so a `residual_quantizer` that calibrates to `amax == 0`/NaN reaches the FP8 exporter's `scale = 448 / amax` — the exact case that guard was added for. The explicit NaN/non-positive check you described in the previous round no longer exists anywhere in the new code, so this half of the earlier comment regressed (one-line fix).
2. `tests/examples/torch_onnx/test_torch_quant_to_onnx.py` runs resnet50 + `auto` with `--trt_build`, but the PR body says the AutoQuantize FP8/INT8 mix only builds on Blackwell+. Please confirm how this test is expected to pass on the CI GPUs.
3. The recipe → `mtq.auto_quantize` kwargs translation duplicates `examples/hf_ptq/hf_ptq.py::_mtq_inputs_from_auto_quantize_config`, and the copy silently drops `score_size` (the new recipe sets `score_size: 128` but the example uses `--num_score_steps`), `cost_excluded_layers`, and top-level `candidate_formats`.
4. `--auto_quantization_formats` is silently overridden for ResNet in `auto` mode with no warning and no opt-out (`--recipe` is rejected with `--quantize_mode=auto`).
5. Nits: the two new helpers are the only ones in this module without docstrings (the rationale docstring added in the previous round didn't survive the rewrite); `"residual_quantizer" in block.downsample._modules` could use `hasattr`; `modelopt_recipes/README.md` still doesn't mention the new top-level `timm/` tree (earlier bot comment unaddressed); `timm/resnet/ptq/{fp8,int8}.yaml` are verbatim copies of `configs/ptq/presets/model/{fp8,int8}.yaml` plus one entry, so they will drift if the presets change.

No licensing concerns (new YAMLs carry the standard NVIDIA SPDX header); no prompt-injection content in the PR metadata.

modelopt/onnx/export/fp8_exporter.py
```diff
@@ -31,6 +31,7 @@
 # when using 1/448 as the Q scale (single fixed value — softmax range is data-independent).
 _FP8_E4M3_MAX = 448.0
 _FP8_E4M3_SOFTMAX_SCALE = 1.0 / _FP8_E4M3_MAX
+_FP8_MIN_CONV_CHANNELS = 16
```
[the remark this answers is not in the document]
ajrasane: No longer applicable. The ONNX FP8 exporter changes that introduced this channel-threshold behavior were removed from the PR, and this file is no longer modified.

modelopt/onnx/export/fp8_exporter.py
```diff
@@ -189,7 +193,8 @@ def _quantize_conv_weights_to_fp8(graph: gs.Graph) -> int:
             weight_input = node.inputs[1]
             if not isinstance(weight_input, gs.Constant):
                 continue
-
+            if node.inputs[0].name in graph_inputs and weight_input.values.shape[1] == 3:
+                continue
```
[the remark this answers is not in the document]
ajrasane: No longer applicable. The ONNX FP8 exporter input-stem change was removed from the PR, and this file is no longer modified.

modelopt_recipes/README.md
```diff
@@ -75,6 +76,12 @@ exclusions are still inherited from `configs/`. Browse
 folder has a `README.md` describing the exact delta. See [`ptq.md`](ptq.md) for
 how the model-specific recipes compare to the general ones and why they deviate.
 
+## `timm/` — architecture-specific recipes
+
+Recipes under `timm/<architecture>/` capture quantization choices required by
+vision architectures and their deployment backends. See
+[`timm/resnet/ptq/`](timm/resnet/ptq/) for ResNet recipes.
+
```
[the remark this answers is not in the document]
ajrasane: Addressed. The recipe catalog now lists `timm/<architecture>/`, includes it in the selection guidance before the `general/` fallback, and documents where new timm recipes belong.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Addressed. `_disable_dead_quantizers` now covers `residual_quantizer`, and `_add_resnet_residual_quantizers` documents the shortcut placement, identity `Sequential`, and disabled-by-default recipe behavior. The duplicate check also uses `hasattr`.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Addressed. The mapping now covers top-level candidate formats, fixed quantization, module search spaces, cost-excluded layers, disabled layers, method, and recipe `score_size`, matching the inputs handled by `hf_ptq.py`. The helper remains local to avoid expanding the recipe package API from this example-only change.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Addressed by removing implicit ResNet recipe selection. `--recipe` now accepts PTQ and AutoQuantize recipes and is authoritative when supplied; without it, `--qformat=auto` continues to use the flag-provided search formats.

tests/examples/torch_onnx/test_torch_quant_to_onnx.py
ajrasane: No longer applicable. Built-in ResNet AutoQuantize recipe support was removed, and the example test skips the `resnet50` + `auto` combination. Users can still provide an explicit AutoQuantize recipe.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -151,6 +142,62 @@ def get_quant_config(quantize_mode, recipe=None):
...
 
 
+def _match_qformat(config):
+    config = config.model_dump() if hasattr(config, "model_dump") else config
+    for qformat in _AUTO_QFORMATS:
+        preset = QUANT_CFG_CHOICES[qformat]
+        same_options = all(
+            config.get(key) == value for key, value in preset.items() if key != "quant_cfg"
+        )
+        if same_options and all(entry in config["quant_cfg"] for entry in preset["quant_cfg"]):
```
cjluo-nv: > _Bot comment._

This comparison can't succeed for any recipe-derived config, which makes `_match_qformat` effectively dead code that always returns `None`.

`QUANT_CFG_CHOICES` is built with `load_config(...).model_dump(exclude_unset=True)` (see `modelopt/recipe/presets.py::load_quant_cfg_choices`), so each preset `quant_cfg` entry is *sparse* — e.g. `{"quantizer_name": "*", "enable": False}`. `_match_qformat` is only ever called with pydantic models (`recipe.quantize`, `AutoQuantizeConfig.candidate_formats[i]`), and `ModeloptBaseConfig.model_dump()` does **not** default to `exclude_unset`, so those entries are *dense*: `{"quantizer_name": "*", "parent_class": None, "cfg": None, "enable": False}`, with `cfg` carrying all ~20 `QuantizerAttributeConfig` defaults. `entry in config["quant_cfg"]` is dict equality, so it is always False.

Either normalize both sides before comparing (e.g. `QuantizeConfig.model_validate(preset).model_dump()` so both are dense), or compare with `config.model_dump(exclude_unset=True)`. A unit test that asserts `_recipe_qformats(load_recipe("timm/resnet/ptq/fp8")) == {"fp8"}` would pin this down.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -610,19 +680,13 @@ def main():
...
     if uses_dynamic_quantize:
         _disable_high_rank_input_quantizers(quantized_model, input_shape, device)
 
     # FP8-family modes emit TRT_FP8QuantizeLinear on the first-layer conv; Blackwell has
     # no tactic for that 3-channel Q→Conv fusion. Skip for pure INT8 (unaffected).
-    uses_fp8_conv_input = args.quantize_mode in ("fp8", "mxfp8", "nvfp4") or (
-        args.quantize_mode == "auto"
-        and any(fmt not in {"int8", "int4_awq"} for fmt in args.auto_quantization_formats)
-    )
+    uses_fp8_conv_input = bool(active_qformats & {"fp8", "mxfp8", "nvfp4", "nvfp4_awq_lite"})
```
cjluo-nv: > _Bot comment._

Consequence of the `_match_qformat` issue above: with `--recipe=timm/resnet/ptq/fp8`, `active_qformats` is the empty set, so `uses_fp8_conv_input` is False and `_disable_low_channel_conv_input_quantizers` never runs. Per that helper's own docstring, ResNet-50's raw-RGB `conv1` then keeps its FP8 input quantizer and TRT on Blackwell fails with "Could not find any implementation for node /conv1/input_quantizer/TRT_FP8QuantizeLinear". The `--trt_build` in the example test passes only because CI/local runs are on Ada.

Same for `uses_dynamic_quantize` (line 683) and for the Conv2d overrides inside `_prepare_auto_quantize_format` when candidate formats come from a recipe rather than a `--auto_quantization_formats` string — the comment in `auto_quantize_model` explicitly says those overrides must be part of the format configs so `effective_bits` accounting is correct.

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -571,8 +624,20 @@ def main():
         )
         print(f"Base Model - Top-1 Accuracy: {top1:.2f}%, Top-5 Accuracy: {top5:.2f}%")
 
-    # Quantize model based on mode
-    if args.quantize_mode == "auto":
+    if isinstance(model, timm.models.resnet.ResNet):
```
cjluo-nv: > _Bot comment._

This registers the residual-quantizer plugin for *every* ResNet run, regardless of whether the active recipe/qformat enables `*residual_quantizer`. So `--qformat=fp8|int8|mxfp8|nvfp4|int4_awq|auto` on a ResNet now also gets `downsample` rewritten (`None` → empty `nn.Sequential`) plus an extra disabled `TensorQuantizer` in the module tree and in `modelopt_state`.

If that structural change is what broke ResNet + `auto` (see the new `pytest.skip`), gating the registration on "the recipe actually enables residual quantizers" would keep the other modes untouched. Note also that `CUSTOM_POST_CONVERSION_PLUGINS` is a module-global set that is never cleaned up — fine for a one-shot script, but worth a comment since it mutates library state.

tests/examples/torch_onnx/test_torch_quant_to_onnx.py
```diff
@@ -30,36 +33,86 @@
...
+            for node in input_producers
+        )
+
+
+@pytest.mark.parametrize("qformat", _QFORMATS)
 @pytest.mark.parametrize("model_key", list(_MODELS))
-def test_torch_onnx(model_key, quantize_mode):
+def test_torch_onnx(tmp_path, model_key, qformat):
+    if model_key == "resnet50" and qformat == "auto":
+        pytest.skip("AutoQuantize is not supported for ResNet")
```
cjluo-nv: > _Bot comment._

`resnet50` + `auto` was previously exercised here (and marked ✅ in the README table, which this PR blanks out) including `--trt_build`. Dropping it removes existing coverage of a working configuration, and the skip reason ("AutoQuantize is not supported for ResNet") is a blanket claim that isn't explained anywhere — the flag-based `--qformat=auto` path is generic and was passing before.

Please either restore the combination, or state in the PR body/CHANGELOG what specifically broke (e.g. the unconditional residual-quantizer insertion) so the regression is a deliberate, documented trade-off.

modelopt_recipes/timm/resnet/ptq/fp8.yaml
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv, line 1: > _Bot comment._

These two new YAMLs carry only the 2-line SPDX preamble, while every other recipe/config YAML in `modelopt_recipes/` (e.g. `huggingface/vit/ptq/fp8.yaml`, `general/ptq/int4_blockwise_weight_only.yaml`, `configs/ptq/units/w8a8_fp8_fp8.yaml`) carries the full `LICENSE_HEADER` Apache block. The `insert-license` pre-commit hook only covers python/shell/c files, so nothing catches this automatically — please paste the full header text from `LICENSE_HEADER` (same for `int8.yaml`).

modelopt_recipes/timm/resnet/ptq/int8.yaml
```diff
@@ -0,0 +1,28 @@
...
+  base_disable_all: configs/ptq/units/base_disable_all
+  default_disabled_quantizers: configs/ptq/units/default_disabled_quantizers
+  int8: configs/numerics/int8
+  int8_per_channel: configs/numerics/int8_per_channel
+
+metadata:
+  recipe_type: ptq
+  description: INT8 ResNet PTQ with quantized residual connections.
+
+quantize:
```
cjluo-nv, line 27: > _Bot comment._

Minor: `modelopt_recipes/README.md` (as updated by this PR) says an architecture-tuned recipe folder should carry a `README.md` documenting the delta from the generic preset. These two recipes differ from `configs/ptq/presets/model/{fp8,int8}.yaml` only by the `*residual_quantizer` entry and will silently drift if those presets change — a short `timm/resnet/ptq/README.md` stating that (and that the entry requires the example's shortcut-quantizer insertion to be present) would help future readers.

cjluo-nv (commented):

All previously-flagged critical items are resolved (residual quantizers now go through the standard calibration flow via `CUSTOM_POST_CONVERSION_PLUGINS`, `_disable_dead_quantizers` covers `residual_quantizer`, residual adds are selected structurally in the test, `--recipe` is authoritative instead of silently overriding flags, and the AutoQuantize mapping no longer drops `score_size`/`cost_excluded_layers`/`candidate_formats`). Design-wise this reuses the existing recipe/preset/plugin machinery rather than inventing a new one — good.

Three things I'd like addressed before merge:

1. **`_match_qformat` can never match a recipe-derived config, so every recipe path silently skips the TRT post-processing.** `QUANT_CFG_CHOICES` entries are `model_dump(exclude_unset=True)` (sparse: `{"quantizer_name": "*", "enable": false}`), while `_match_qformat` compares them against `recipe.quantize.model_dump()` (dense: `parent_class: None`, `cfg: None`, plus ~20 `QuantizerAttributeConfig` defaults). `entry in config["quant_cfg"]` is therefore always False, so `_recipe_qformats()` returns an empty set and `_prepare_auto_quantize_format()` never adds the Conv2d override. Concretely: `--recipe=timm/resnet/ptq/fp8` never runs `_disable_low_channel_conv_input_quantizers`, i.e. the raw-RGB `conv1` keeps its FP8 input quantizer — the exact Blackwell "no implementation for `TRT_FP8QuantizeLinear`" failure the helper exists to prevent. CI on Ada hides this. Same class of bug for recipe-supplied `mxfp8`/`nvfp4` AutoQuantize candidates (no Conv2d→FP8 override, and `_disable_high_rank_input_quantizers` skipped).

2. **ResNet + `auto` is silently dropped from a previously-supported/tested configuration** (README ✅ → blank, test `pytest.skip("AutoQuantize is not supported for ResNet")`). The flag-based auto path for resnet50 with `--trt_build` was passing before this PR; if the reason is the now-unconditional residual-quantizer registration, register the plugin only when residual quantization is actually requested, and otherwise please explain the regression in the PR body/CHANGELOG.

3. **New recipe YAMLs carry a truncated license header** (2-line SPDX only) whereas every other `modelopt_recipes/**.yaml` uses the full `LICENSE_HEADER` Apache text. Not auto-enforced (the `insert-license` hook only covers python/shell/c), so it needs a manual fix.

Also: there is no unit coverage for `_match_qformat` / `_recipe_qformats` / `_mtq_inputs_from_auto_quantize_config` — a small pure-Python test asserting that the new `timm/resnet/ptq/fp8.yaml` resolves to `{"fp8"}` would have caught (1) immediately.

gcunhase:
Thank you @ajrasane. Can you please add the runtime perf numbers for the Torch and ONNX quantization paths to confirm parity? Thanks!

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Addressed by removing `_match_qformat` and `_recipe_qformats`. Recipe candidates are handled by their configured quantizer attributes instead of reverse-matching a preset name, and focused recipe-loading coverage was added.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Addressed. Low-channel FP8 handling now checks the configured Conv quantizer directly and always runs after quantization. Recipe AutoQuantize block candidates receive FP8/INT8 Conv overrides from their quantizer attributes, and high-rank block input quantizers are detected from model state rather than `active_qformats`.

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Addressed. The residual plugin is now registered only when a PTQ recipe explicitly contains an enabled `*residual_quantizer` entry, so ordinary ResNet qformat runs do not change the module structure or ModelOpt state.

tests/examples/torch_onnx/test_torch_quant_to_onnx.py
ajrasane: Addressed as an intentional support boundary. The README compatibility table, changelog, tests, runtime error, and PR description now state that ResNet supports only FP8 and INT8 because of limited TensorRT convolution kernel support.

modelopt_recipes/timm/resnet/ptq/fp8.yaml
ajrasane: Addressed. Both new recipe YAMLs now carry the complete Apache 2.0 license header used by the rest of `modelopt_recipes`.

modelopt_recipes/timm/resnet/ptq/int8.yaml
ajrasane: Addressed. Added `timm/resnet/ptq/README.md` documenting the shared numerics, the ResNet-specific residual-quantizer delta, the required torch ONNX integration, and the FP8/INT8 support boundary.

ajrasane:
@gcunhase

| Path | Format | GPU latency | Throughput | Engine size |
|---|---:|---:|---:|---:|
| `torch_onnx` | FP8 | 0.292 ms | 3,395 qps | 25.05 MiB |
| `onnx_ptq` | FP8 | 0.285 ms | 3,476 qps | 26.89 MiB |
| `torch_onnx` | INT8 | 0.240 ms | 4,122 qps | 25.34 MiB |
| `onnx_ptq` | INT8 | 0.232 ms | 4,263 qps | 27.09 MiB |

Benchmarked on an NVIDIA RTX 6000 Ada Generation GPU with TensorRT 10.16.0.72.

Disclaimer: These are rough estimates and should not be considered official performance numbers, as the measurements may include some overheads.

cjluo-nv (commented):

**Re-review (8 files, +448/-144).** Design-review protocol applied first: this PR does not introduce a new subsystem — the residual quantizer is installed through the existing `CUSTOM_POST_CONVERSION_PLUGINS` hook (invoked by `replace_quant_module` *before* `set_quantizer_by_cfg`, so the recipe configures it and `mtq.quantize`'s own forward loop calibrates it), and the new YAMLs slot into the existing `modelopt_recipes/` catalog with `$import` units. That's reuse of the in-repo pattern, not a second one. Licensing is fine: the two new recipe YAMLs now carry the verbatim `LICENSE_HEADER` text. No prompt-injection content in the PR metadata.

**Previous blockers — resolved**
- *`_match_qformat` could never match a recipe-derived config, so `--recipe` silently skipped TRT post-processing*: the reverse-preset-matching helpers are gone. `_disable_low_channel_fp8_conv_input_quantizers` and `_disable_high_rank_input_quantizers` now run unconditionally and decide from live quantizer state (`num_bits == (4, 3)`, `block_sizes`), so `--recipe=timm/resnet/ptq/fp8` does disable the raw-RGB `conv1` input quantizer. Confirmed the loader converts `eXmY` → tuples, so the `(4,3)`/`(2,1)` comparisons are sound.
- *Truncated SPDX-only header on the new YAMLs*: fixed, matches `LICENSE_HEADER`.
- *`modelopt_recipes/README.md` didn't mention the `timm/` tree*: fixed (catalog + selection guidance + placement), plus a `timm/resnet/ptq/README.md` documenting the delta.
- *AutoQuantize mapping dropped `score_size`/`cost_excluded_layers`/`candidate_formats`*: all three are mapped now, and `test_auto_quantize_recipe_mapping` + a new e2e recipe test cover the top-level-candidates shape (previously the `module_search_spaces=[]` failure case).
- *Residual plugin registered unconditionally*: now gated on the recipe actually enabling `*residual_quantizer`, and `_disable_dead_quantizers` covers `residual_quantizer`.

Also verified the removal of `_calibrate_uncalibrated_quantizers` is safe: forcing `algorithm = "max"` for mxfp8/nvfp4 calibrates the FP8 Conv overrides in the normal pass, and `finish_stats_collection` skips amax loading for dynamic/MX quantizers.

**Why a nudge rather than an approve** — see the reason field: the remaining items are product/behavior judgment calls plus a few small robustness gaps, not a re-flag of the fixed bugs.

gcunhase:
> @gcunhase
> 
> Path	Format	GPU latency	Throughput	Engine size
> `torch_onnx`	FP8	0.292 ms	3,395 qps	25.05 MiB
> `onnx_ptq`	FP8	0.285 ms	3,476 qps	26.89 MiB
> `torch_onnx`	INT8	0.240 ms	4,122 qps	25.34 MiB
> `onnx_ptq`	INT8	0.232 ms	4,263 qps	27.09 MiB
> Benchmarked on an NVIDIA RTX 6000 Ada Generation GPU with TensorRT 10.16.0.72.
> 
> Disclaimer: These are rough estimates and should not be considered official performance numbers, as the measurements may include some overheads.
> 
> > 🤖 _Generated by Codex (AI agent)._

Thanks @ajrasane , can you please also add the before and after this fix numbers so we can show the need of this PR? Thanks!

examples/torch_onnx/README.md
```diff
@@ -369,7 +389,7 @@ python torch_quant_to_onnx.py \
 | [vit_base_patch16_224](https://huggingface.co/timm/vit_base_patch16_224.augreg_in21k_ft_in1k) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [swin_tiny_patch4_window7_224](https://huggingface.co/timm/swin_tiny_patch4_window7_224.ms_in1k) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [swinv2_tiny_window8_256](https://huggingface.co/timm/swinv2_tiny_window8_256.ms_in1k) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
-| [resnet50](https://huggingface.co/timm/resnet50.a1_in1k) | ✅ | ✅ | ✅ | ✅ | | ✅ |
+| [resnet50](https://huggingface.co/timm/resnet50.a1_in1k) | ✅ | ✅ | | | | |
```
gcunhase: Can you add `-`, `N/A`, or `Not supported` for the non-supported precisions? Thanks!

examples/torch_onnx/torch_quant_to_onnx.py
```diff
@@ -607,24 +693,14 @@ def main():
...
-    # FP8-family modes emit TRT_FP8QuantizeLinear on the first-layer conv; Blackwell has
-    # no tactic for that 3-channel Q→Conv fusion. Skip for pure INT8 (unaffected).
-    uses_fp8_conv_input = args.quantize_mode in ("fp8", "mxfp8", "nvfp4") or (
-        args.quantize_mode == "auto"
-        and any(fmt not in {"int8", "int4_awq"} for fmt in args.auto_quantization_formats)
-    )
-    if uses_fp8_conv_input:
-        _disable_low_channel_conv_input_quantizers(quantized_model)
+    if is_resnet:
+        _validate_resnet_quantizers(quantized_model)
```
gcunhase, line 697: Is this only needed for ResNet or any models that require QDQ nodes to be placed in the residual branch?

gcunhase:
Is there a way to make this behavior generic to other models with residual connections? Per my understanding, this logic would require each relevant model to have yaml files in `model_recipes`, is that correct?

examples/torch_onnx/torch_quant_to_onnx.py
ajrasane: Currently I have validated this only for the ResNet models. Are there any other models with Residual connections we would like to target? I can try to find a more general solution using them.

ajrasane:
> Is there a way to make this behavior generic to other models with residual connections? 

I can try to make this behavior generic for a few Convolutional models we would like to support. But I think it will be hard to guarantee that this would be applicable for all models with Residual connections as well. To do this, we will need to implement a graph parsing logic at the torch level to check for the residual connection pattern which is out of scope for the current PR.

> Per my understanding, this logic would require each relevant model to have yaml files in model_recipes, is that correct?

Yes, the current presets are too generic. Users can get runnable quantized models with them, but they are not guaranteed to be the most performant. So it would be better to have per model configs that provide the best performance. Users can tweak their custom models by using these configs as guidance. This will also simplify a lot of code on the export and post-processing side where we inject/remove quantizers to give the best performance.

examples/torch_onnx/README.md
ajrasane: Updated the ResNet-50 row to mark MXFP8, NVFP4, INT4_AWQ, and Auto as `N/A` in 8b7079b74.

ajrasane:
@gcunhase

| Format | Path | State | GPU latency | Throughput | Engine size | Change after fix |
|---|---|---|---:|---:|---:|---:|
| FP8 | `torch_onnx` | Before | 0.292 ms | 3,399 qps | 25.04 MiB | — |
| FP8 | `torch_onnx` | After | 0.292 ms | 3,395 qps | 25.05 MiB | No measurable change |
| FP8 | `onnx_ptq` | Reference | 0.285 ms | 3,476 qps | 26.89 MiB | — |
| INT8 | `torch_onnx` | Before | 0.303 ms | 3,274 qps | 25.53 MiB | — |
| INT8 | `torch_onnx` | After | 0.240 ms | 4,122 qps | 25.34 MiB | 20.9% lower latency, 25.9% higher throughput |
| INT8 | `onnx_ptq` | Reference | 0.232 ms | 4,263 qps | 27.09 MiB | — |

`Before` uses `--qformat` without a ResNet recipe; `After` uses the corresponding `timm/resnet/ptq` recipe. Benchmarked on an NVIDIA RTX 6000 Ada Generation GPU with TensorRT 10.16.0.72.

Disclaimer: These are rough estimates and should not be considered official performance numbers, as the measurements may include some overheads.

gcunhase:
> I can try to make this behavior generic for a few Convolutional models we would like to support. But I think it will be hard to guarantee that this would be applicable for all models with Residual connections as well. To do this, we will need to implement a graph parsing logic at the torch level to check for the residual connection pattern which is out of scope for the current PR.

Understood, is that something that we can try to add in a follow-up PR?

Here are some models that have residual connections that might be helpful to validate the more generic workflow:

| # | Model | Residual pattern |                                                                                                                                                             
|---|-----------|-----------------|                                                                                                                                                              
| 1 | **ResNet** | Identity or Conv1×1 projection shortcut → Add |                                                                                                            
| 2 | **ResNeXt** | Same as ResNet but grouped convolutions in the main branch; same Add topology |                                                                                          
| 3 | **Inception-ResNet-v1/v2** | Inception module output + Conv1×1 projection of input → Add |                                                                                             
| 4 | **SE-ResNet** | ResNet block + squeeze-excitation recalibration → Add on the residual path |                                                                                           
| 5 | **MobileNetV2** | Inverted residual: expand → depthwise → project Conv1×1 → Add (only when stride=1 and channels match) |                                                              
| 6 | **EfficientNet** | MobileNetV2-style inverted residual + SE → Add; scaled uniformly |                                                                                                  
| 7 | **RegNet** | ResNet-style bottleneck with group convolutions → Add; design space parameterisation |                                                                                    
| 8 | **RepVGG** | During training: Conv3×3 + Conv1×1 + identity → Add (all three branches collapsed to one conv at inference) |                                                             
| 9 | **ConvNeXt** | Depthwise Conv7×7 → LayerNorm → Linear → GELU → Linear → Add (inverted bottleneck with depthwise residual) |                                                            
| 10 | **YOLO v5/v8** | CSP bottleneck: repeated Conv3×3 residual blocks with Add inside the cross-stage partial module |

Thanks!

gcunhase:
> @gcunhase
> 
> Format	Path	State	GPU latency	Throughput	Engine size	Change after fix
> FP8	`torch_onnx`	Before	0.292 ms	3,399 qps	25.04 MiB	—
> FP8	`torch_onnx`	After	0.292 ms	3,395 qps	25.05 MiB	No measurable change
> FP8	`onnx_ptq`	Reference	0.285 ms	3,476 qps	26.89 MiB	—
> INT8	`torch_onnx`	Before	0.303 ms	3,274 qps	25.53 MiB	—
> INT8	`torch_onnx`	After	0.240 ms	4,122 qps	25.34 MiB	20.9% lower latency, 25.9% higher throughput
> INT8	`onnx_ptq`	Reference	0.232 ms	4,263 qps	27.09 MiB	—
> `Before` uses `--qformat` without a ResNet recipe; `After` uses the corresponding `timm/resnet/ptq` recipe. Benchmarked on an NVIDIA RTX 6000 Ada Generation GPU with TensorRT 10.16.0.72.
> 
> Disclaimer: These are rough estimates and should not be considered official performance numbers, as the measurements may include some overheads.
> 
> > 🤖 _Generated by Codex (AI agent)._

Thank you! This is very clear in showing the benefits of this PR and parity with ONNX quant for INT8 while keeping FP8 performance.

gcunhase (approved):
LGTM, let's move efforts to make this a more generic workflow to a follow-up PR. Thanks!

cjluo-nv (commented):

Re-review (8 files, +448/-145). Design-review protocol applied first: this PR does not introduce a new subsystem — the residual quantizer is installed through the existing `CUSTOM_POST_CONVERSION_PLUGINS` hook (I confirmed `replace_quant_module` calls `register_custom_post_conversion_plugins` after `_replace_quant_module` and before `set_quantizer_by_cfg`, so the recipe configures it and `mtq.quantize`'s own forward loop calibrates it), and the new YAMLs slot into the existing `modelopt_recipes/` catalog via `$import`. Licensing is fine: both new recipe YAMLs carry the verbatim `LICENSE_HEADER` Apache text. No prompt-injection content in the PR metadata.

**Previous blockers — verified resolved**
- `_disable_dead_quantizers` now covers `residual_quantizer` (it lives on the `downsample` `Sequential`, so `getattr` finds it), closing the `amax == 0` → `scale = 448/amax` gap.
- The hand-rolled `enable_calib` → forward-loop → strict `load_calib_amax()` block and the second calibration pass are gone; the standard flow is used.
- `_match_qformat` / `_recipe_qformats` (which could never match a sparse `QUANT_CFG_CHOICES` entry, so `--recipe` silently skipped TRT post-processing) are removed; `_disable_low_channel_fp8_conv_input_quantizers` and `_disable_high_rank_input_quantizers` now run unconditionally and decide from live quantizer state (`num_bits == (4,3)`, `block_sizes`).
- Truncated SPDX headers fixed; `modelopt_recipes/README.md` catalog + selection guidance + `timm/resnet/ptq/README.md` added.
- Residual plugin registration is gated on the recipe actually enabling `*residual_quantizer`, so ordinary ResNet qformat runs no longer get `downsample` rewritten.
- The test selects residual adds structurally (single `Relu` consumer) before asserting the count of 16.
- `_calibrate_uncalibrated_quantizers` removal is safe: `get_quant_config` now forces `algorithm = "max"` for mxfp8/nvfp4 so the FP8 Conv overrides calibrate in the normal pass.

**Test edit assessment (focus area 4):** `test_torch_onnx` now skips `resnet50 × {mxfp8, nvfp4, auto}`, which previously ran with `--trt_build`. This is a coverage reduction, but it is justified within the diff — the CHANGELOG, README compatibility table (`N/A`), `timm/resnet/ptq/README.md`, and a runtime `ValueError` all state the new FP8/INT8-only support boundary, and the repo owner explicitly asked for the `N/A` markers. So the test edit itself is not a blocker.

**Why nudge rather than approve** — a few residual items an owner should eyeball; see the reason field.

cjluo-nv (approved):
CO approval

PR #2027, opened 2026-07-29
[Example]: Calibration-free FP8/NVFP4 PTQ for speculative-decoding drafters

h-guo18:
### What does this PR do?

Type of change: new example

Adds `examples/speculative_decoding/scripts/quantize_drafter.py`, a CLI that quantizes an exported speculative-decoding drafter to FP8 or NVFP4 — weight-only or weight+activation — with no calibration data.

It needs no modeling code either. Exported drafters such as [`nvidia/MiniMax-M3-DSpark`](https://huggingface.co/nvidia/MiniMax-M3-DSpark) have no importable model class, so each 2-D weight is wrapped in a throwaway `nn.Linear` under its checkpoint key and ModelOpt's usual `quantizer_name` patterns select over those names. Works for any drafter layout (DSpark / DFlash / EAGLE3 / Medusa).

**Formats:** `w4a16_nvfp4`, `nvfp4`, `fp8`, `fp8_pc_pt` — the ModelOpt formats vLLM's backend can actually serve. AWQ is deliberately not offered, since `awq_lite` silently degrades to plain RTN without a `forward_loop`.

**Static activation scales without calibration.** `fp8` and `nvfp4` normally need an activation amax *measured* on calibration data; a fixed `input_scale` of 1.0 is applied instead. That works because acceptance length is governed almost entirely by **clipping**, not resolution:

Sweeping the fixed scale over three decades (same setup as the Testing section below; bf16 baseline 3.1423):

| `input_scale` | amax | FP8 AL | vs bf16 | NVFP4 AL | vs bf16 |
|---|---|---|---|---|---|
| 0.003 | 1.3 | 2.2204 | -29.34% | 2.2076 | -29.75% |
| 0.01 | 4.5 | 2.6719 | -14.97% | 2.6641 | -15.22% |
| 0.03 | 13.4 | 2.9751 | -5.32% | 2.9259 | -6.89% |
| 0.1 | 44.8 | 3.1013 | -1.31% | 3.0206 | -3.88% |
| 0.2 | 89.6 | 3.1178 | -0.78% | 3.0015 | -4.48% |
| 0.3 | 134.4 | 3.1370 | -0.17% | 3.0222 | -3.82% |
| 0.5 | 224.0 | 3.1268 | -0.50% | 3.0360 | -3.38% |
| **1.0 (default)** | **448.0** | **3.1457** | **+0.11%** | **3.0193** | **-3.91%** |
| 2.0 | 896.0 | 3.1354 | -0.22% | 3.0172 | -3.98% |
| 4.0 | 1792.0 | 3.1245 | -0.57% | 3.0034 | -4.42% |

Both formats fall off a cliff below ~0.03, where the declared range sits far under the activations' true magnitude and most of the tensor is clipped. Both then sit on a flat plateau from ~0.3 to 4.0 **with no drop-off at the top**, so the scale only has to be big enough. 1.0 is the middle of that plateau, which is why it is hardcoded rather than exposed. NVFP4 trails FP8 by a roughly constant 3.5% across the plateau — that gap is the 4-bit resolution cost, and no choice of scale recovers it.

Deriving the amax from the weights instead was tried and does not work: `max|W|` averages 0.79 while a RMSNorm'd activation is O(1) with outlier channels in the tens, so the range lands 1–2 orders of magnitude low and clips, measuring -31% to -46% AL.

**Where calibration would go.** All of this sits behind `resolve_activation_scales()`, the single place deciding where a static amax comes from. Real calibration slots in ahead of the fixed fallback with no change to the CLI or the call site, and composes because `set_static_activation_amax()` skips quantizers that already have an amax:

```python
if calib_forward_loop is not None:
    mtq.calibrate(root, quant_cfg["algorithm"], forward_loop=calib_forward_loop)
set_static_activation_amax(root)   # fills in what calibration did not reach
```

**Serving a quantized drafter.** Four things had to be written into the exported checkpoint before vLLM would load one:

- emit `quant_method` (`modelopt_fp4` / `modelopt`) — vLLM reads that key, ModelOpt writes only `quant_algo`
- emit the exclusion list under `ignore` too — that is the key read from the flat `quantization_config`; `exclude_modules` alone yields an empty exclusion set
- add `*<name>` wildcards so exclusions match a runtime's nested module prefix (`model.fc`) rather than the checkpoint key (`fc`)
- add `*qkv_proj` / `*gate_up_proj` aliases for layers a runtime fuses, whose names appear in no checkpoint key

Nothing is then needed on the caller side. **This closes the open question left in the previous revision of this PR: vLLM does read `quantization_config` off the draft checkpoint.** `ModelConfig._verify_quantization` fills `quantization` in from `quant_method` when it is unset, so once the export declares that key — the first fix above — detection works on its own. Verified on Nemotron-3.5-Lightning passing nothing: `Detected ModelOpt NVFP4 checkpoint (quant_algo=NVFP4)` → `FlashInferCuteDslNvFp4LinearKernel`, AL 4.278 against 4.203 measured earlier.

`specdec_bench` also gains a `DSPARK` algorithm, which it did not have: an exported `Qwen3DSparkModel` would otherwise have to go through `DFLASH` and be built with vLLM `method="dflash"`. The branch sets `method="dspark"` and leaves `draft_sample_method` on vLLM's own default of `greedy`. A target whose fused-collective workspace (sized at CUDA-graph capture) overflows at large speculative batches can disable graphs with `--runtime_params '{"engine_args": {"enforce_eager": true}}'`.

For DFlash-family drafters, `qwen3_dflash.py` builds its fused context-KV projection by reading `qkv_proj.weight` raw and calling `F.linear`, which cannot consume a packed weight. Keep those layers in bf16 with `--exclude '*q_proj*' '*k_proj*' '*v_proj*' '*qkv_proj*'`; `o_proj` and the MLP — the bulk of the drafter — still quantize. That exclusion is mandatory, not a tuning choice.

`fc` (the projection from the target's captured layers into the draft) is the one real knob, and it is a genuine trade rather than a free win — see the Testing section for both models' numbers. The examples quantize it; add `'*fc*'` to the exclude list to keep it in bf16.

`embed_tokens`, `markov_head` and `confidence_head` are excluded by default: they are 2-D so the flat view treats them as GEMMs, but they are embeddings or a single-output projection. `lm_head` is excluded by the preset itself — unlike on a base model it is 37% of this drafter's parameters, so `--quantize_lm_head` is a real lever (~1.9 GiB), but measure AL first. The flag re-enables both of `lm_head`'s quantizers; re-enabling only the weight one would ship a W+A checkpoint whose `lm_head` has no `input_scale` while the config still advertises it as quantized.

### Usage

```bash
# weight+activation FP8, calibration-free, lossless on both models measured below
python scripts/quantize_drafter.py \
    --drafter_path deepseek-ai/dspark_qwen3_8b_block7 \
    --qformat fp8 \
    --export_path ./dspark-qwen3-8b-fp8 \
    --exclude '*q_proj*' '*k_proj*' '*v_proj*' '*qkv_proj*'

# smallest: weight-only NVFP4
python scripts/quantize_drafter.py \
    --drafter_path nvidia/MiniMax-M3-DSpark \
    --qformat w4a16_nvfp4 \
    --export_path ./MiniMax-M3-DSpark-W4A16
```

Or end to end on Slurm — quantize, then measure AL — via the launcher examples added here, one per target:

```bash
uv run launch.py --yaml examples/Qwen/Qwen3-8B/hf_dspark_ptq_nvfp4.yaml --yes
uv run launch.py --yaml examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/hf_dspark_ptq_nvfp4.yaml --yes
```

Serving one, if you are not going through `specdec_bench`:

```python
speculative_config = {
    "method": "dspark",
    "model": "./dspark-qwen3-8b-fp8",   # quantization is read from its config.json
    "num_speculative_tokens": 7,
}
```

### Testing

Two targets with different architectures, so the conclusions are not one model's quirk:

* **Qwen3-8B** (dense transformer) + [`deepseek-ai/dspark_qwen3_8b_block7`](https://huggingface.co/deepseek-ai/dspark_qwen3_8b_block7), `block_size` 7, TP1.
* **Nemotron-3.5-Lightning-30B-A3B** (hybrid Mamba-MoE) + [`nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark`](https://huggingface.co/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark), `block_size` 8, TP8, with the mamba engine settings the model card pins (`mamba_backend=flashinfer`, `mamba_ssm_cache_dtype=float16`, stochastic SSM-cache rounding).

Both: MT-Bench 80 questions, greedy, one vLLM instance per point.

| recipe | activations | Qwen3-8B AL | vs bf16 | Nemotron-3.5 AL | vs bf16 |
|---|---|---|---|---|---|
| bf16 baseline | — | 3.1423 | — | 4.3296 | — |
| **`fp8`** | static, `input_scale` 1.0 | **3.1457** | **+0.11%** | **4.3289** | **-0.02%** |
| `fp8_pc_pt` | dynamic per-token | 3.1228 | -0.62% | 4.3411 | +0.26% |
| `w4a16_nvfp4`, `fc` in bf16 | bf16 (weight-only) | 3.0392 | -3.28% | 4.2899 | -0.92% |
| `w4a16_nvfp4`, `fc` quantized | bf16 (weight-only) | 3.0186 | -3.94% | 4.2334 | -2.22% |
| **`nvfp4`** | static, `input_scale` 1.0 | **3.0193** | **-3.91%** | **4.2030** | **-2.92%** |

**FP8 weight+activation at the fixed `input_scale` of 1.0 is lossless on both.** +0.11% and -0.02% are both inside run-to-run noise — the Nemotron baseline was measured twice under identical settings and the two runs differ by 0.94% (4.3093 / 4.3499), which sets the resolution of that column. On the same reading, `fp8` and `fp8_pc_pt` are indistinguishable on Nemotron; the dynamic variant only pulls ahead on Qwen3. NVFP4 costs 3-4% on Qwen3 and 2-3% on Nemotron, i.e. the 4-bit weight resolution is the real price and it is model-dependent but bounded.

Whether to quantize `fc` is a per-model call rather than a general recommendation — it buys a few percent of size for an AL cost that differs by ~2x between these two drafters:

| `fc` bf16 → quantized | Qwen3-8B | Nemotron-3.5 |
|---|---|---|
| checkpoint size | 3.293 → 3.181 GiB (-3.4%) | 1.316 → 1.258 GiB (-4.4%) |
| AL | 3.0392 → 3.0186 (-0.68%) | 4.2899 → 4.2334 (-1.32%) |

`fc` itself is only 3.5% (Qwen3) / 4.5% (Nemotron) of drafter parameters; `embed_tokens` is the bulk (26% / 36%) and is excluded by default.

The Qwen3 `w4a16_nvfp4` rows were measured in a later session than the rest of that column; the `fc`-in-bf16 run reproduced the original number to four decimals (3.0392), so the column is internally comparable.

Also validated on `nvidia/MiniMax-M3-DSpark`: `w4a16_nvfp4` runs in 67 s on CPU, 9.98 GiB (fp32) -> 3.51 GiB; all 43 quantized tensors round-trip within 0.0952 relative error; the 29 untouched tensors are bit-identical to `bf16(source)`.

- Is this change backward compatible?: ✅ (example-only)
- Did you write any new necessary tests?: ❌ — validated manually as above. Can add a `tests/examples/speculative_decoding/` test over a small synthetic drafter if wanted before merge.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A (example-only)

### Additional Information

The measurements above are one drafter on one target with one benchmark; the plateau's location and the ~3.5% NVFP4 gap should be re-measured before assuming they carry to a different drafter.

Note when reading an exported checkpoint: `input_scale` is `amax/448` for FP8 but `amax/(6*448)` for NVFP4, so the one fixed amax records as 1.0 in an FP8 checkpoint and 0.1667 in an NVFP4 one. Both mean the same activation range.

### DIFF
--- a/examples/specdec_bench/run.py
+++ b/examples/specdec_bench/run.py
@@ -305,7 +305,7 @@ if __name__ == "__main__":
         type=str,
         required=False,
         default="EAGLE3",
-        choices=["EAGLE3", "EAGLE", "DRAFT_TARGET", "NGRAM", "MTP", "DFLASH", "NONE"],
+        choices=["EAGLE3", "EAGLE", "DRAFT_TARGET", "NGRAM", "MTP", "DFLASH", "DSPARK", "NONE"],
         help="Speculative algorithm to use",
     )
     parser.add_argument("--model_dir", type=str, required=True, help="Path to the model directory")
--- a/examples/specdec_bench/run.py
+++ b/examples/specdec_bench/run.py
@@ -188,6 +188,7 @@ def run_simple(args):
         sampling_kwargs=sampling_kwargs,
         speculative_algorithm=args.speculative_algorithm,
         draft_model_dir=args.draft_model_dir,
+        draft_quantization=args.draft_quantization,
         speculative_num_steps=args.draft_length,
         speculative_num_draft_tokens=args.block_size,
         tensor_parallel_size=args.tp_size,
@@ -315,6 +316,14 @@ if __name__ == "__main__":
         required=False,
         default=None,
         help="Path to the draft model directory",
+    )
+    parser.add_argument(
+        "--draft_quantization",
+        type=str,
+        required=False,
+        default=None,
+        help="Quantization method of the draft checkpoint (e.g. modelopt, modelopt_fp4). "
+        "Read from the draft's config.json when omitted; set it only to override.",
     )
     parser.add_argument(
         "--runtime_params",
--- a/examples/specdec_bench/run.py
+++ b/examples/specdec_bench/run.py
@@ -306,7 +306,7 @@ if __name__ == "__main__":
         type=str,
         required=False,
         default="EAGLE3",
-        choices=["EAGLE3", "EAGLE", "DRAFT_TARGET", "NGRAM", "MTP", "DFLASH", "NONE"],
+        choices=["EAGLE3", "EAGLE", "DRAFT_TARGET", "NGRAM", "MTP", "DFLASH", "DSPARK", "NONE"],
         help="Speculative algorithm to use",
     )
     parser.add_argument("--model_dir", type=str, required=True, help="Path to the model directory")
--- a/examples/specdec_bench/specdec_bench/models/vllm.py
+++ b/examples/specdec_bench/specdec_bench/models/vllm.py
@@ -28,6 +28,16 @@ except ImportError:
     vllm = None
 
 
+# Forwarded from ``--runtime_params`` ``engine_args.<key>``; extend as needed.
+PASSTHROUGH_ENGINE_ARGS = (
+    "mamba_backend",
+    "mamba_ssm_cache_dtype",
+    "mamba_cache_mode",
+    "mamba_cache_philox_rounds",
+    "enable_mamba_cache_stochastic_rounding",
+)
+
+
 class VLLMModel(Model):
     # Cross-engine ``--max_seq_len`` (run.py) lands in kwargs under the
     # vLLM-native name ``max_model_len`` (see run.py's ``_MAX_SEQ_LEN_KEY``)
@@ -114,6 +124,13 @@ class VLLMModel(Model):
                 "model": kwargs.get("draft_model_dir"),
                 "num_speculative_tokens": kwargs.get("speculative_num_draft_tokens", 8),
             }
+        elif kwargs.get("speculative_algorithm") == "DSPARK":
+            specdec = {
+                "method": "dspark",
+                "model": kwargs.get("draft_model_dir"),
+                "num_speculative_tokens": kwargs.get("speculative_num_draft_tokens", 7),
+                "draft_sample_method": kwargs.get("draft_sample_method", "greedy"),
+            }
         elif kwargs.get("speculative_algorithm") == "NONE":
             specdec = None
 
@@ -133,8 +150,9 @@ class VLLMModel(Model):
             max_num_seqs=max_concurrent_requests * num_speculative_tokens,
             skip_tokenizer_init=False,
             async_scheduling=kwargs.get("async_scheduling", True),
-            enforce_eager=False,
+            enforce_eager=kwargs.get("enforce_eager", False),
             max_model_len=kwargs.get("max_model_len"),
+            **{key: kwargs[key] for key in PASSTHROUGH_ENGINE_ARGS if kwargs.get(key) is not None},
         )
         self.engine_args = engine_args
         self.model = AsyncLLM.from_engine_args(engine_args)
--- a/examples/specdec_bench/specdec_bench/models/vllm.py
+++ b/examples/specdec_bench/specdec_bench/models/vllm.py
@@ -119,12 +119,9 @@ class VLLMModel(Model):
         elif kwargs.get("speculative_algorithm") == "NONE":
             specdec = None
 
-        # A quantized draft checkpoint has to declare its own quantization. vLLM otherwise
-        # copies the target's onto the draft (config/speculative.py, "Align the
-        # quantization of draft model"), so a quantized drafter under a bf16 target is
-        # built as if it were bf16 and dies loading the packed weights. The drafter's
-        # config.json already records the format, so read it back rather than making the
-        # caller repeat it; --draft_quantization overrides.
+        # vLLM copies the target's quantization onto the draft, so a quantized drafter under
+        # a bf16 target is built as bf16 and dies loading the packed weights. Read the format
+        # from the drafter's own config.json; --draft_quantization overrides.
         if specdec is not None and specdec.get("model"):
             draft_quantization = kwargs.get("draft_quantization")
             if draft_quantization is None:
--- a/examples/specdec_bench/specdec_bench/models/vllm.py
+++ b/examples/specdec_bench/specdec_bench/models/vllm.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 import asyncio
+import json
+import os
 import time
 
 from .base import Model
@@ -116,6 +118,24 @@ class VLLMModel(Model):
             }
         elif kwargs.get("speculative_algorithm") == "NONE":
             specdec = None
+
+        # A quantized draft checkpoint has to declare its own quantization. vLLM otherwise
+        # copies the target's onto the draft (config/speculative.py, "Align the
+        # quantization of draft model"), so a quantized drafter under a bf16 target is
+        # built as if it were bf16 and dies loading the packed weights. The drafter's
+        # config.json already records the format, so read it back rather than making the
+        # caller repeat it; --draft_quantization overrides.
+        if specdec is not None and specdec.get("model"):
+            draft_quantization = kwargs.get("draft_quantization")
+            if draft_quantization is None:
+                draft_config = os.path.join(specdec["model"], "config.json")
+                if os.path.isfile(draft_config):
+                    with open(draft_config) as f:
+                        quant_config = json.load(f).get("quantization_config") or {}
+                    draft_quantization = quant_config.get("quant_method")
+            if draft_quantization:
+                specdec["quantization"] = draft_quantization
+                print(f"Draft model quantization: {draft_quantization}")
 
         if specdec is None:
             num_speculative_tokens = 1
--- a/examples/specdec_bench/specdec_bench/models/vllm.py
+++ b/examples/specdec_bench/specdec_bench/models/vllm.py
@@ -171,6 +171,23 @@ class VLLMModel(Model):
             async_scheduling=kwargs.get("async_scheduling", True),
             enforce_eager=enforce_eager,
             max_model_len=kwargs.get("max_model_len"),
+            # Engine knobs a model card may pin, passed through from
+            # --runtime_params engine_args.<key>. Only keys the caller actually set are
+            # forwarded, so vLLM keeps its own defaults otherwise. Hybrid Mamba models
+            # need these: on Nemotron-3.5-Lightning the SSM-cache settings decide whether
+            # the first draft token is accepted, and leaving them at vLLM's defaults costs
+            # ~65% of acceptance length while every later position looks normal.
+            **{
+                key: kwargs[key]
+                for key in (
+                    "mamba_backend",
+                    "mamba_ssm_cache_dtype",
+                    "mamba_cache_mode",
+                    "mamba_cache_philox_rounds",
+                    "enable_mamba_cache_stochastic_rounding",
+                )
+                if kwargs.get(key) is not None
+            },
         )
         self.engine_args = engine_args
         self.model = AsyncLLM.from_engine_args(engine_args)
--- a/examples/specdec_bench/specdec_bench/models/vllm.py
+++ b/examples/specdec_bench/specdec_bench/models/vllm.py
@@ -116,6 +116,17 @@ class VLLMModel(Model):
                 "model": kwargs.get("draft_model_dir"),
                 "num_speculative_tokens": kwargs.get("speculative_num_draft_tokens", 8),
             }
+        elif kwargs.get("speculative_algorithm") == "DSPARK":
+            # Match draft sampling to the target's verify mode: a greedy target with a
+            # probabilistic draft (or the reverse) crushes acceptance at temp > 0.
+            temperature = sampling_kwargs.get("temperature", 1.0)
+            specdec = {
+                "method": "dspark",
+                "model": kwargs.get("draft_model_dir"),
+                "num_speculative_tokens": kwargs.get("speculative_num_draft_tokens", 7),
+                "draft_sample_method": kwargs.get("dspark_draft_sample_method")
+                or ("greedy" if temperature == 0 else "probabilistic"),
+            }
         elif kwargs.get("speculative_algorithm") == "NONE":
             specdec = None
 
@@ -139,6 +150,14 @@ class VLLMModel(Model):
         else:
             num_speculative_tokens = specdec.get("num_speculative_tokens", 3)
 
+        # DSpark's block-parallel draft can outgrow the pre-allocated workspace during
+        # CUDA-graph capture; acceptance length is unaffected by graph capture, so skip it.
+        # SPECDEC_ENFORCE_EAGER=1 forces the same for other algorithms.
+        enforce_eager = (
+            kwargs.get("speculative_algorithm") == "DSPARK"
+            or os.environ.get("SPECDEC_ENFORCE_EAGER") == "1"
+        )
+
         engine_args = AsyncEngineArgs(
             model=model_dir,
             tokenizer=kwargs.get("tokenizer_path"),
@@ -150,7 +169,7 @@ class VLLMModel(Model):
             max_num_seqs=max_concurrent_requests * num_speculative_tokens,
             skip_tokenizer_init=False,
             async_scheduling=kwargs.get("async_scheduling", True),
-            enforce_eager=False,
+            enforce_eager=enforce_eager,
             max_model_len=kwargs.get("max_model_len"),
         )
         self.engine_args = engine_args
--- a/examples/speculative_decoding/scripts/quantize_drafter.py
+++ b/examples/speculative_decoding/scripts/quantize_drafter.py
@@ -1,0 +1,378 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Calibration-free PTQ for a speculative-decoding drafter.
+
+Every scale is derived from the weights, so this needs no dataset and no forward pass, and
+never imports the drafter's modeling code: each 2-D weight is wrapped in a throwaway
+``nn.Linear`` under its checkpoint name, and ModelOpt's usual ``quantizer_name`` patterns
+select over those names. Works for any drafter layout (DSpark / DFlash / EAGLE3 / Medusa),
+including exported ones that ship no importable model class.
+
+``fp8`` and ``nvfp4`` need a static activation amax, normally measured on calibration data;
+a fixed ``input_scale`` of 1.0 is applied instead. Acceptance length is governed by
+clipping rather than resolution, and AL sits on a flat plateau from input_scale ~0.3 to 4.0
+(Qwen3-8B + DSpark, MT-Bench: +0.1% for FP8, -3.9% for NVFP4), so the scale only has to be
+big enough. AWQ is not offered: ``awq_lite`` silently degrades to RTN without a
+``forward_loop``.
+
+Example:
+    python quantize_drafter.py \
+        --drafter_path nvidia/MiniMax-M3-DSpark \
+        --qformat fp8 \
+        --export_path ./MiniMax-M3-DSpark-FP8
+"""
+
+import argparse
+import copy
+import json
+import shutil
+from pathlib import Path
+
+import torch
+import torch.nn as nn
+from safetensors.torch import load_file, save_file
+
+import modelopt.torch.quantization as mtq
+from modelopt.recipe.presets import QUANT_CFG_CHOICES
+from modelopt.torch.export.quant_utils import (
+    get_activation_scaling_factor,
+    get_quant_config,
+    get_quantization_format,
+    get_weight_block_size,
+    get_weight_scaling_factor,
+    get_weight_scaling_factor_2,
+    to_quantized_weight,
+)
+from modelopt.torch.quantization.config import need_calibration
+from modelopt.torch.quantization.utils import is_quantized_linear
+
+# INT8/INT4 are absent on purpose: they quantize cleanly but vLLM's ModelOpt backend
+# cannot serve them.
+SUPPORTED_QFORMATS = [
+    "w4a16_nvfp4",
+    "nvfp4",
+    "fp8",
+    "fp8_pc_pt",
+]
+
+# All 2-D, so the flat view treats them as GEMMs, but none is one: markov_w1/embed_tokens
+# are embeddings (ModelOpt's presets skip these via `parent_class`, which the flat view
+# cannot see), and confidence_head has a single output whose per-channel scale is 0-dim.
+# All tiny. `fc` is left to the presets; `lm_head` is excluded by the preset itself.
+DEFAULT_EXCLUDE = ["*markov_head*", "*confidence_head*", "*embed_tokens*"]
+
+# Sidecars carried over to the export, and -- with the weights and config.json -- the only
+# files fetched when --drafter_path is a repo id rather than a local directory.
+SIDECAR_FILES = ("tokenizer.json", "tokenizer_config.json", "generation_config.json")
+
+# The amax that yields input_scale 1.0 for FP8. NVFP4 divides by 6*448, so the same amax
+# records as 0.1667 there; both mean the same activation range.
+FP8_E4M3_MAX = 448.0
+STATIC_ACT_AMAX = FP8_E4M3_MAX
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
+    parser.add_argument(
+        "--drafter_path", required=True, help="HF repo id or local dir of the drafter checkpoint."
+    )
+    parser.add_argument("--export_path", required=True, help="Output directory.")
+    parser.add_argument(
+        "--qformat",
+        default="w4a16_nvfp4",
+        choices=SUPPORTED_QFORMATS,
+        help="Quantization format. All are calibration-free; fp8 and nvfp4 additionally "
+        "quantize activations, using a fixed input_scale of 1.0.",
+    )
+    parser.add_argument(
+        "--dtype",
+        default="bfloat16",
+        choices=["bfloat16", "float16", "float32"],
+        help="Compute dtype the weights are cast to before quantizing.",
+    )
+    parser.add_argument(
+        "--exclude",
+        nargs="*",
+        default=[],
+        metavar="PATTERN",
+        help="Extra fnmatch patterns to leave unquantized, in `quantizer_name` form. "
+        f"Appended to the defaults ({' '.join(DEFAULT_EXCLUDE)}), which always apply.",
+    )
+    parser.add_argument(
+        "--quantize_lm_head",
+        action="store_true",
+        help="Also quantize lm_head -- the largest drafter tensor, but it feeds the "
+        "acceptance test directly, so measure AL first.",
+    )
+    return parser.parse_args()
+
+
+def auto_map_modules(config: dict, source_dir: Path) -> set[Path]:
+    """Module files an ``auto_map`` points at, relative to ``source_dir``.
+
+    Values are either ``"modeling_x.XModel"`` or, for tokenizers, a list whose entries may
+    be null (``[null, "tokenization_x.XTokenizerFast"]``); a ``repo--`` prefix points at
+    another repository and is not a local file. A value may also carry a package path
+    (``"pkg/modeling_x.XModel"``), which is preserved so the exported reference resolves.
+
+    The config is untrusted -- it may come straight off the Hub -- so references that
+    escape ``source_dir`` are dropped rather than followed.
+    """
+    modules = set()
+    for value in (config.get("auto_map") or {}).values():
+        for ref in value if isinstance(value, list) else [value]:
+            if not isinstance(ref, str) or "." not in ref:
+                continue
+            relative = Path(ref.split("--")[-1].rsplit(".", 1)[0] + ".py")
+            candidate = (source_dir / relative).resolve()
+            if relative.is_absolute() or not candidate.is_relative_to(source_dir.resolve()):
+                print(f"Skipping auto_map entry outside the checkpoint: {ref}")
+                continue
+            modules.add(relative)
+    return modules
+
+
+def load_drafter(drafter_path: str) -> tuple[Path, dict[str, torch.Tensor]]:
+    """Resolve a local dir or HF repo id to (dir, state_dict)."""
+    local_dir = Path(drafter_path)
+    if not local_dir.is_dir():
+        from huggingface_hub import snapshot_download
+
+        # A drafter that ships custom modeling code references it from auto_map, and the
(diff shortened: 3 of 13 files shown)


examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -117,6 +119,24 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
         elif kwargs.get("speculative_algorithm") == "NONE":
             specdec = None
 
+        # A quantized draft checkpoint has to declare its own quantization. vLLM otherwise
```
h-guo18: this comment should be trim to 1-2 lines.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,356 @@
...
+name, and ModelOpt's usual ``quantizer_name`` patterns select over those names. Works for
+any drafter layout (DSpark / DFlash / EAGLE3 / Medusa).
+
+Avoiding the modeling code is what makes this work for DFlash-family drafters at all: an
+exported drafter declares ``architectures: ["DFlashDraftModel"]`` but ships no importable
+class, so ``AutoModelForCausalLM`` silently loads it as a plain Qwen3 and drops the draft
+tensors, and the real draft ``forward`` takes hidden states rather than ``input_ids``, so
+the stock ``hf_ptq.py`` calibration loop cannot drive it either.
+
+``fp8`` and ``nvfp4`` quantize activations against a static amax that is normally measured
```
h-guo18: remove this paragraph

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: Done — cut to two lines in 851d4a2.

examples/speculative_decoding/scripts/quantize_drafter.py
h-guo18: Removed in 851d4a2. Also did a pass over the rest of the script's comments and docstrings with the same rule — dropped anything the code already says, kept only the non-obvious reasons. Script is 432 -> 323 lines; exports are byte-identical for all five formats.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,323 @@
...
+    return parser.parse_args()
+
+
+def load_drafter(drafter_path: str) -> tuple[Path, dict[str, torch.Tensor]]:
+    """Resolve a local dir or HF repo id to (dir, state_dict)."""
+    local_dir = Path(drafter_path)
+    if not local_dir.is_dir():
+        from huggingface_hub import snapshot_download
+
+        local_dir = Path(snapshot_download(drafter_path))
```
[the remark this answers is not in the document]
h-guo18: Done in 63ce3b5 — `allow_patterns` now limits the fetch to `*.safetensors`, `config.json` and the three sidecars, hoisted into a `SIDECAR_FILES` constant shared with the copy loop.

tools/launcher/common/specdec/quantize_drafter.sh
```diff
@@ -0,0 +1,43 @@
...
+    if [ -z "${DRAFTER}" ]; then
+        echo "ERROR: no drafter checkpoint at ${DRAFTER_CKPT}"
+        exit 1
+    fi
+    echo "Auto-detected drafter: ${DRAFTER}"
+fi
+
+python modules/Model-Optimizer/examples/speculative_decoding/scripts/quantize_drafter.py \
+    --drafter_path ${DRAFTER} \
+    ${@}
```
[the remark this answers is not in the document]
h-guo18: Done in 63ce3b5 — good catch, this was a real bug rather than a style point. Verified: with a file matching `*q_proj*` in the CWD, unquoted `${@}` expanded `'*q_proj*'` to that filename, so the drafter would have been quantized with a silently wrong exclusion set.

tools/launcher/common/specdec/quantize_drafter.sh
```diff
@@ -0,0 +1,43 @@
...
+
+DRAFTER=${DRAFTER_CKPT}
+# Training writes exported-checkpoint-<step>/ under output_dir; take the newest.
+if [ ! -f "${DRAFTER}/config.json" ]; then
+    DRAFTER=$(ls -d ${DRAFTER_CKPT}/exported-checkpoint-* 2>/dev/null | sort -t- -k3 -n | tail -1)
+    if [ -z "${DRAFTER}" ]; then
+        echo "ERROR: no drafter checkpoint at ${DRAFTER_CKPT}"
+        exit 1
+    fi
+    echo "Auto-detected drafter: ${DRAFTER}"
```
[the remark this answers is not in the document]
h-guo18: Done in 63ce3b5 — auto-detection is now gated on `[ -d ]`, so an HF repo id passes through to the script instead of erroring.

tools/launcher/common/specdec/quantize_drafter.sh
```diff
@@ -0,0 +1,43 @@
...
+source ${SCRIPT_DIR}/../service_utils.sh
+
+trap 'error_handler $0 $LINENO' ERR
+
+###################################################################################################
+
+DRAFTER=${DRAFTER_CKPT}
+# Training writes exported-checkpoint-<step>/ under output_dir; take the newest.
+if [ ! -f "${DRAFTER}/config.json" ]; then
+    DRAFTER=$(ls -d ${DRAFTER_CKPT}/exported-checkpoint-* 2>/dev/null | sort -t- -k3 -n | tail -1)
```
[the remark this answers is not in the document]
h-guo18: Done in 63ce3b5 — switched to `find -printf '%f\n' | sort -V`. Verified against 900/1000/1500: picks 1500.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,323 @@
...
+        *parents, leaf = key[: -len(".weight")].split(".")
+        node = root
+        for part in parents:
+            if part not in node:
+                node[part] = nn.ModuleDict()
+            node = node[part]
+        out_features, in_features = weight.shape
+        linear = nn.Linear(in_features, out_features, bias=False, dtype=dtype)
+        linear.weight.data = weight.to(dtype)
+        node[leaf] = linear
```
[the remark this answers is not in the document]
h-guo18: Not taking this one. The script is a pure weight transform with no forward pass, and CPU is fast enough: 67 s for a real 2.37B drafter. Requiring CUDA would make it strictly worse — the launcher example runs this as a CPU-only step, and it would force a GPU allocation for a one-minute job.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,323 @@
...
+    print(f"Set {n} static activation amax values (fixed, input_scale 1.0) -- not calibrated.")
+
+
+def build_quant_cfg(qformat: str, exclude: list[str], quantize_lm_head: bool) -> dict:
+    """Take the shipped preset and layer the drafter-specific exclusions on top."""
+    quant_cfg = copy.deepcopy(QUANT_CFG_CHOICES[qformat])
+    if quantize_lm_head:
+        quant_cfg["quant_cfg"].append(
+            {"quantizer_name": "*lm_head*weight_quantizer", "enable": True}
+        )
```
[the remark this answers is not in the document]
h-guo18: Confirmed and fixed in dbedb9c. Reproduced exactly as described: `--quantize_lm_head --qformat fp8` shipped `lm_head.weight` + `weight_scale` with no `input_scale`, while the config left `lm_head` out of `ignore` and advertised it as FP8. Now re-enables both quantizers; `lm_head.input_scale` is written for fp8 and nvfp4.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,323 @@
...
+    # Same list, second key: the flat ``quantization_config`` in config.json is read for
+    # ``ignore``, not ``exclude_modules``.
+    config["quantization_config"]["ignore"] = list(exclude_modules)
+    config["torch_dtype"] = args.dtype
+    (export_dir / "config.json").write_text(json.dumps(config, indent=2))
+    (export_dir / "hf_quant_config.json").write_text(json.dumps(hf_quant_config, indent=2))
+
+    for extra in ("tokenizer.json", "tokenizer_config.json", "generation_config.json"):
+        if (source_dir / extra).is_file():
+            shutil.copy2(source_dir / extra, export_dir / extra)
```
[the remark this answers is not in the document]
h-guo18: Fixed in 63ce3b5 — the export now copies the .py files an `auto_map` points at. Verified with a fixture carrying `auto_map` + two modeling files. (Note the DFlash/DSpark drafters this targets have no `auto_map`, so nothing dangled in practice, but it is right for drafters that ship custom modeling code.)

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,323 @@
...
+        if not is_quantized_linear(module):
+            continue
+        input_quantizer = getattr(module, "input_quantizer", None)
+        if input_quantizer is None or not input_quantizer.is_enabled:
+            continue
+        if getattr(input_quantizer, "_dynamic", False):
+            continue
+        if getattr(input_quantizer, "amax", None) is not None:
+            continue
+        input_quantizer.amax = torch.tensor(amax, dtype=torch.float32).to(module.weight.dtype)
```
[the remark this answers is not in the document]
h-guo18: Fixed in dbedb9c — amax stays fp32 now. Agreed on the reasoning: exact for 448.0 today, but bf16's mantissa would round an arbitrary measured amax by up to ~0.26%, in exactly the calibration path the docstring advertises.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,323 @@
...
+
+# INT8/INT4 are absent on purpose: they quantize cleanly but vLLM's ModelOpt backend
+# cannot serve them.
+SUPPORTED_QFORMATS = [
+    "w4a16_nvfp4",
+    "nvfp4",
+    "fp8",
+    "fp8_pc_pt",
+    "fp8_pb_wo",
+]
```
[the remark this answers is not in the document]
h-guo18: Fixed in dbedb9c by dropping `fp8_pb_wo`. Confirmed it emits raw lowercase `quant_algo: 'fp8_pb_wo'` where the others emit `'FP8'` / `'FP8_PER_CHANNEL_PER_TOKEN'`, and it was the one offered format with no measurement behind it — so the comment claiming every offered format is servable did not hold for it.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,323 @@
...
+            continue
+        quantization = get_quantization_format(module)
+        assert quantization is not None, f"{name}: enabled quantizer resolved to no format"
+        weight_scale = get_weight_scaling_factor(module)
+        weight_scale_2 = get_weight_scaling_factor_2(module)
+        # The packing helpers index the scale as ``scale[:, None]``, which a 0-dim scale
+        # cannot satisfy. One row of weights, so leave it in ``dtype``.
+        if weight_scale is not None and weight_scale.dim() == 0 and module.weight.shape[0] == 1:
+            print(f"Skipping {name}: single-output projection, per-channel scale is scalar")
+            continue
```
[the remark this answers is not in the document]
h-guo18: Not taking this one — I checked it against the code and the claim is inverted on both halves.

Ran a fixture with a single-output projection (`scorer.proj`, `[1, 256]`) that is not in `DEFAULT_EXCLUDE`:

- `--qformat fp8_pc_pt` (per-channel, `axis=0`) → quantizer reports `axis=0 amax=5.22e-02` (a scalar, not a reduced `(1,)` shape), the guard **fires**, and the layer is skipped.
- `--qformat fp8` (per-tensor) → quantizer reports `per-tensor`, the guard does **not** fire, and the layer packs normally.

So it is per-channel that reaches this branch and per-tensor that does not, which is what the comment says. Leaving the code and comment as they are.

examples/speculative_decoding/scripts/quantize_drafter.py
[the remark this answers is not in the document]
h-guo18: Good catch — fixed in 7a79da1. Measured on a 30-layer stand-in: ~4.6 s of `reset_parameters()` down to ~0.29 s, and peak host memory drops by the largest layer. Exports are byte-identical (md5-checked against the same config before/after).

ChenhanYu (approved):

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,348 @@
...
+            shutil.copy2(source_dir / extra, export_dir / extra)
+
+    # A drafter that ships custom modeling code points at it from auto_map; the export
+    # carries that config verbatim, so the .py files have to come along or the reference
+    # dangles. (The DFlash/DSpark exports have no auto_map -- this is for the ones that do.)
+    if config.get("auto_map"):
+        for module in {ref.split("--")[-1].split(".")[0] for ref in config["auto_map"].values()}:
+            source_py = source_dir / f"{module}.py"
+            if source_py.is_file():
+                shutil.copy2(source_py, export_dir / source_py.name)
```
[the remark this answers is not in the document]
h-guo18: Confirmed and fixed in 2e20994. Reproduced the crash: `{'AutoTokenizer': [None, 'tokenization_x.XTokenizerFast']}` raised `AttributeError: 'list' object has no attribute 'split'`, so any drafter with a tokenizer auto_map failed at export. Extraction now lives in `auto_map_modules()`, verified against six shapes (plain string, list-with-null, list-both-set, `repo--` prefix, absent, null). Also added `*.py` to the download allow-list.

examples/speculative_decoding/scripts/quantize_drafter.py
```diff
@@ -0,0 +1,348 @@
...
+    """Resolve a local dir or HF repo id to (dir, state_dict)."""
+    local_dir = Path(drafter_path)
+    if not local_dir.is_dir():
+        from huggingface_hub import snapshot_download
+
+        local_dir = Path(
+            snapshot_download(
+                drafter_path, allow_patterns=["*.safetensors", "config.json", *SIDECAR_FILES]
+            )
+        )
```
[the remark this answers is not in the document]
h-guo18: Fixed in 2e20994 — `*.py` added to `snapshot_download`'s `allow_patterns`, so the modules an auto_map points at are fetched and then copied into the export. Verified end to end with a fixture carrying a list-valued auto_map: both referenced .py files land in the export dir.

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -114,14 +116,48 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
...
             specdec = None
 
+        # vLLM copies the target's quantization onto the draft, so a quantized drafter under
+        # a bf16 target is built as bf16 and dies loading the packed weights. Read the format
+        # from the drafter's own config.json; --draft_quantization overrides.
+        if specdec is not None and specdec.get("model"):
+            draft_quantization = kwargs.get("draft_quantization")
+            if draft_quantization is None:
+                draft_config = os.path.join(specdec["model"], "config.json")
+                if os.path.isfile(draft_config):
```
talorabr: If the model is given as an HF identifier rather than local path, this check here will fail and the quantization config won't be taken from the model config.json.
Is that expected? Do we want to add a support for that case as well? (by using `AutoConfig.from_pretrained`)

examples/specdec_bench/run.py
```diff
@@ -305,7 +306,7 @@ def run_simple(args):
         type=str,
         required=False,
         default="EAGLE3",
-        choices=["EAGLE3", "EAGLE", "DRAFT_TARGET", "NGRAM", "MTP", "DFLASH", "NONE"],
+        choices=["EAGLE3", "EAGLE", "DRAFT_TARGET", "NGRAM", "MTP", "DFLASH", "DSPARK", "NONE"],
```
talorabr, line 308: for DSPARK, can you add NotImplemented exception for engines that are not vllm?

examples/specdec_bench/run.py
```diff
@@ -316,6 +317,14 @@ def run_simple(args):
         default=None,
         help="Path to the draft model directory",
     )
+    parser.add_argument(
+        "--draft_quantization",
+        type=str,
+        required=False,
+        default=None,
+        help="Quantization method of the draft checkpoint (e.g. modelopt, modelopt_fp4). "
```
talorabr: can you specify that this parameter only applies for VLLM?

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -114,14 +116,48 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
...
             }
+        elif kwargs.get("speculative_algorithm") == "DSPARK":
+            # Match draft sampling to the target's verify mode: a greedy target with a
+            # probabilistic draft (or the reverse) crushes acceptance at temp > 0.
+            temperature = sampling_kwargs.get("temperature", 1.0)
+            specdec = {
+                "method": "dspark",
+                "model": kwargs.get("draft_model_dir"),
+                "num_speculative_tokens": kwargs.get("speculative_num_draft_tokens", 7),
+                "draft_sample_method": kwargs.get("dspark_draft_sample_method")
```
benchislett: This is bad form. Keep a consistent default behaviour (likely "greedy", following vLLM's default) and leave it to the user to sweep or specify

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -114,14 +116,48 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
...
+                or ("greedy" if temperature == 0 else "probabilistic"),
+            }
         elif kwargs.get("speculative_algorithm") == "NONE":
             specdec = None
 
+        # vLLM copies the target's quantization onto the draft, so a quantized drafter under
```
benchislett: This seems more like a vllm parsing issue than a client-side issue. Where is this coming from?

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -114,14 +116,48 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
...
         else:
             num_speculative_tokens = specdec.get("num_speculative_tokens", 3)
 
+        # DSpark's block-parallel draft can outgrow the pre-allocated workspace during
+        # CUDA-graph capture; acceptance length is unaffected by graph capture, so skip it.
+        # SPECDEC_ENFORCE_EAGER=1 forces the same for other algorithms.
+        enforce_eager = (
+            kwargs.get("speculative_algorithm") == "DSPARK"
+            or os.environ.get("SPECDEC_ENFORCE_EAGER") == "1"
+        )
```
benchislett: What is this? It will massively increase runtime to force enforce eager. I've never had to do this when running dspark

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -133,8 +169,25 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
...
+                for key in (
+                    "mamba_backend",
+                    "mamba_ssm_cache_dtype",
+                    "mamba_cache_mode",
+                    "mamba_cache_philox_rounds",
+                    "enable_mamba_cache_stochastic_rounding",
+                )
+                if kwargs.get(key) is not None
+            },
         )
```
benchislett: this can be cleaned up

examples/specdec_bench/run.py
```diff
@@ -316,6 +317,14 @@ def run_simple(args):
         default=None,
         help="Path to the draft model directory",
     )
+    parser.add_argument(
```
benchislett: I don't see why this needs to be provided by the client. vLLM should pick this up automatically

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: Good catch — not expected, and worse than just missing the config: `--draft_model_dir` is passed straight through to the engine and vLLM accepts a repo id there, so this was the one place assuming a local directory, and it failed *silently*. The drafter then loads as bf16 and either dies on the packed weights or produces a plausible-but-wrong acceptance length.

Fixed in bc083ed. I tried `AutoConfig.from_pretrained` first as you suggested, but it imports the model class and so fails exactly where we need it — on `RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8` it raised `Could not import module 'LlamaConfig'` and returned None where the answer is `compressed-tensors`, and it would also fail on a drafter whose architecture transformers cannot construct (e.g. `Qwen3DSparkModel`). So it now reads the raw config.json, falling back to `hf_hub_download` when the path is not local.

Verified on six cases: local nvfp4 → `modelopt_fp4`, local fp8 → `modelopt`, local bf16 → None, Hub quantized → `compressed-tensors`, Hub custom-arch bf16 → None, bogus path → None with a warning.

examples/specdec_bench/run.py
h-guo18: Added in bc083ed — raises `NotImplementedError` when `--speculative_algorithm DSPARK` is combined with a non-VLLM engine. Without it the other engines would silently fall back to non-speculative decoding.

examples/specdec_bench/run.py
h-guo18: Done in bc083ed — the help now starts with "VLLM only."

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: updated

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: you are right. vllm can parse it correctly. removed this argument.

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: I trimmed the comments. I think it's cleaner to leave the logic like this (not passing a arg if not provided). This way we can use vllm's default value without maintaining a set of default values that align with vllm. What do you think?

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: I hit some cuda graph buffer OOM on minimax-m3. I took another look and it should be resolvable by tuning up `--max-num-batched-tokens`. You are right and this shouldn't be default for all models.

Updated now.

h-guo18: I keep a SPECDEC_ENFORCE_EAGER option though in current code for debugging or scenario like I encountered with minimax-m3.

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -122,6 +130,10 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
         else:
             num_speculative_tokens = specdec.get("num_speculative_tokens", 3)
 
+        # Escape hatch for engine/model combinations whose fused-collective workspace is
+        # sized at CUDA-graph capture and overflows at large speculative batches.
+        enforce_eager = os.environ.get("SPECDEC_ENFORCE_EAGER") == "1"
```
benchislett: This is not the canonical way to pass args in this program. Use kwargs

examples/specdec_bench/run.py
```diff
@@ -181,6 +181,13 @@ def run_simple(args):
     sampling_kwargs = args.runtime_params.get("sampling_kwargs", {"temperature": 0})
     if args.temperature is not None:
         sampling_kwargs["temperature"] = args.temperature
+    # DSPARK is only wired up for vLLM; other engines would silently fall back to
```
benchislett: This comment is redundant

examples/specdec_bench/specdec_bench/models/vllm.py
benchislett: Yes that is the right way to pass those args. But we shouldn't have to manually write out all the specific args we are allowed to propagate like this. Ideally we forward everything from kwargs in this way (except the ones that need special handling, which get popped).

If you are going to have a hardcoded list of allowable args, factor it out here as we will be frequently updating it.

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: updated

examples/specdec_bench/run.py
h-guo18: removed

examples/specdec_bench/run.py
```diff
@@ -181,14 +181,18 @@ def run_simple(args):
     sampling_kwargs = args.runtime_params.get("sampling_kwargs", {"temperature": 0})
     if args.temperature is not None:
         sampling_kwargs["temperature"] = args.temperature
+    if args.speculative_algorithm == "DSPARK" and args.engine != "VLLM":
+        raise NotImplementedError(
+            f"--speculative_algorithm DSPARK is only supported by --engine VLLM, got {args.engine}."
+        )
+
```
talorabr: I don't like that this is here, since we currently check the speculative algorithm support in each engine
If you can move it to each engine it would be best (under `models/`)

examples/specdec_bench/specdec_bench/models/vllm.py
```diff
@@ -117,47 +115,20 @@ def __init__(self, model_dir, max_concurrent_requests, sampling_kwargs, **kwargs
...
-            # Match draft sampling to the target's verify mode: a greedy target with a
-            # probabilistic draft (or the reverse) crushes acceptance at temp > 0.
-            temperature = sampling_kwargs.get("temperature", 1.0)
             specdec = {
                 "method": "dspark",
                 "model": kwargs.get("draft_model_dir"),
                 "num_speculative_tokens": kwargs.get("speculative_num_draft_tokens", 7),
-                "draft_sample_method": kwargs.get("dspark_draft_sample_method")
-                or ("greedy" if temperature == 0 else "probabilistic"),
+                "draft_sample_method": kwargs.get("dspark_draft_sample_method", "greedy"),
```
talorabr: actually - why do we need the "dspark_" prefix for the draft_sample_method parameter? can you change the name of the parameter to be just `draft_sample_method`?

examples/specdec_bench/specdec_bench/models/vllm.py
h-guo18: Then we would need to keep a list of special handling:
- arg name mapping from specdecbench to AsyncEngineArgs (e.g. `moe_expert_parallel_size`→`enable_expert_parallel`)
- duplicate arg in both explicit arg list and kwargs (e.g. trust_remote_code)

I would prefer the second form you mentioned. Updated in latest commit

benchislett (approved):

talorabr (approved):
LGTM

PR #2028, opened 2026-07-29
Add nvfp4_act_headroom activation calibration for NVFP4

cjluo-nv:
### What does this PR do?

Type of change: new feature

Adds `nvfp4_act_headroom`, a calibration algorithm for NVFP4 **activation** global scales.

NVFP4 scales a tensor in two levels: an FP8-E4M3 scale per 16-element block, plus one per-tensor global scale. Plain `max` calibration sets that global scale from the largest per-block amax seen during calibration, which leaves no room above it — any activation larger than the calibration max saturates.

`nvfp4_act_headroom` instead anchors the global scale to a low percentile of the per-block amax distribution:

```
amax = max(rho * anchor, floor)
```

`anchor` is the per-block amax at `anchor_percentile`; `upper` is the per-block amax at `upper_percentile` and is the top of the range the scale commits to representing. Multiplying the low anchor by `rho` places the calibrated blocks in the lower part of the FP8 scale range and leaves the rest as headroom.

`upper_percentile` defaults to 99.99 rather than the literal maximum on purpose. Flooring at the literal max means one freak block drags the global scale up until every other block's FP8 block scale falls below subnormal: on a tensor with a single block seven orders of magnitude above the rest, that flushes 99.998% of elements to zero, versus 6.7% when the rare blocks are clipped instead. On a benign wide-range tensor the two choices differ by 0.4% relative MSE. Set `upper_percentile=100` to floor at the literal observed max, which guarantees no calibration data is clipped, at that exposure. The calibrator warns when the per-block range is too wide for `rho` to clear any headroom.

The algorithm applies only to NVFP4 dynamic-block **input** quantizers; weight quantizers and everything else keep plain `max`.

Files:

- `calib/nvfp4_act_headroom.py` — `NVFP4ActHeadroomCalibrator` (log2 histogram of per-block amaxes, bounded memory)
- `config.py` — `NVFP4ActHeadroomCalibConfig` (`anchor_percentile` default 1, `rho` default 16384)
- `model_calib.py` — `nvfp4_act_headroom_calibrate`
- `mode.py` — mode registration
- `modelopt_recipes/general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml` — mirrors `nvfp4_default-kv_fp8_cast` (dynamic NVFP4 W4A4 + FP8 KV-cache cast, same module coverage) with only the calibration algorithm swapped:

```diff
-  algorithm: max
+  algorithm:
+    method: nvfp4_act_headroom
+    anchor_percentile: 1
```

### Why a dedicated collector rather than reusing `HistogramCalibrator`

`HistogramCalibrator` histograms the tensor values themselves into linear, dynamically re-ranged bins, and its percentile mode returns an amax that *clips* a high-tail fraction. This algorithm needs a different statistic (per-block amaxes, a derived quantity), different bin spacing (log2, because block amaxes span many decades and the anchor is read from the **low** tail, where linear bins have almost no resolution), and a different reduction (a low-percentile anchor scaled up, not an upper-tail clip). Reusing it would mean changing its collection semantics for every existing caller; composing it would still leave the log-spacing and the derived statistic unaddressed. The collector here is a fixed 512-bin int64 histogram per quantizer, so the memory argument for a histogram (rather than retaining values) is preserved.

### Usage

```bash
python examples/hf_ptq/hf_ptq.py --pyt_ckpt_path <model> \
    --recipe general/ptq/nvfp4_act_headroom-kv_fp8_cast --dataset <calib.jsonl>
```

Or directly:

```python

NVFP4 = {"num_bits": (2, 1), "block_sizes": {-1: 16, "type": "dynamic", "scale_bits": (4, 3)}}
config = {
    "quant_cfg": [
        {"quantizer_name": "*weight_quantizer", "cfg": NVFP4},
        {"quantizer_name": "*input_quantizer", "cfg": NVFP4},
    ],
    "algorithm": {"method": "nvfp4_act_headroom", "anchor_percentile": 1, "rho": 16384},
}
model = mtq.quantize(model, config, forward_loop)
```

### Testing

**Unit tests** — `tests/unit/torch/quantization/test_nvfp4_act_headroom.py`, 26 tests covering the anchor and range terms, monotonicity in `anchor_percentile`, headroom above plain max, the too-wide-range fallback and its warning, the rare-outlier case (the default clips it; `upper_percentile=100` chases it), the `upper_percentile=100` no-clipping guarantee across six distributions, input validation (`rho` / percentile bounds, NaN/Inf, all-zero, a last dim that is not a multiple of the block size), quantizer selection, calibrator restoration after calibration, `shared_states` / `sync_expert_weight_amax` forwarding, and W4A4 end-to-end confirming weights fall back to plain max while only activations get the headroom scale.

Full `tests/unit/torch/quantization/` and `tests/unit/recipe/` suites pass with no regressions (the remaining failures in this environment are a pre-existing broken-`torchvision` import and `test_data_parallel_auto_quantize`, both of which fail identically on `main`).

**End-to-end on a real model** — ran the shipped recipe on a 30B hybrid Mamba/attention MoE (1024 calibration samples @ seq 4096) and verified the export is a standard NVFP4 checkpoint: 6268 activation quantizers calibrated; 19G vs the 62G BF16 source; 6260 `input_scale`, 6267 `weight_scale` / `weight_scale_2`; `quant_algo: NVFP4`, `kv_cache_quant_algo: FP8`; 53 excluded modules; no module carrying an `input_scale` without a `weight_scale`. 32 of 6268 quantizers (0.5%) had a per-block range too wide for `rho` to clear and warned.

- Is this change backward compatible?: ✅ — purely additive: a new opt-in algorithm, config class, and recipe. No existing behavior changes.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no copied code, no new dependencies.
- Did you write any new necessary tests?: ✅ — 12 new unit tests.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — entry under 0.47 → New Features → Quantization.


modelopt/torch/quantization/calib/nvfp4_act_headroom.py
```diff
@@ -0,0 +1,217 @@
...
+
+        running_max = float(self._running_max)
+        floor = self._percentile(_FLOOR_PERCENTILE)
+        anchor = (
+            self._percentile(
+                self._anchor_percentile, floor_value=floor / _ANCHOR_FLOOR_RATIO if floor else None
+            )
+            if floor
+            else None
+        )
```
meenchen, line 190: > _Bot comment._

This does not measure the stated observed range: `floor` is the 99.99th percentile, so a sufficiently rare largest block can be excluded even though `running_max` is what actually forced this fallback. In that case the warning can claim the FP8 range is not exceeded and recommend tuning `rho`, although `running_max / anchor > 28672` means every permitted `rho` must fall back. Please base this guardrail/message on `running_max / anchor` (and add a rare-outlier test where the maximum lies above the floor percentile).

tests/unit/torch/quantization/test_nvfp4_act_headroom.py
```diff
@@ -0,0 +1,299 @@
...
+def test_ragged_last_dim_is_handled():
+    """Activations whose last dim is not a multiple of the block size must not crash."""
+    g = torch.Generator().manual_seed(3)
+    x = torch.randn(32, 70, generator=g)  # 70 % 16 != 0
+    amax = _calibrate(NVFP4ActHeadroomCalibrator(anchor_percentile=1.0), x)
+    assert amax > 0
+    assert amax >= float(x.abs().max())
+
+
+@pytest.mark.parametrize("bad", [float("nan"), float("inf")])
```
meenchen, line 159: > _Bot comment._

Please move this import, and the `model_calib as mc` import in `test_shared_state_knobs_are_forwarded_to_max_calibrate`, to the import block at the top of the file. There is no circular/optional-dependency reason given for keeping these Python imports local.

meenchen (commented):

The calibration approach and coverage look thoughtful, but I found one correctness issue in the advertised FP8-range guardrail and two required review concerns. Design review: the problem is to choose an NVFP4 activation global scale from the per-block distribution while retaining unseen-outlier headroom with bounded collection memory. Existing alternatives include extending/composing the in-repo `HistogramCalibrator` with `reduce_block_amax`, implementing this as max calibration plus a post-processing collector, or using PyTorch/NumPy/SciPy percentile primitives (all already available). The bounded log2 histogram is a plausible reason for a specialized collector, but the PR body does not explicitly compare it with the existing histogram calibrator/post-processing path; please document why extension/reuse is unsuitable. Licensing also needs human sign-off: the two new files use a 2024 NVIDIA header, while the repository's canonical `LICENSE_HEADER` currently says 2026, so the new headers do not qualify for the exact-header auto-approval exception. The PR is cohesive and under the size threshold, and its unit/integration tests otherwise cover the main behavior well.

modelopt_recipes/general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml
```diff
@@ -0,0 +1,41 @@
...
+  description: >-
+    Composes dynamic NVFP4 W4A4 model quantization with FP8 KV-cache cast mode using constant
+    amax; uses nvfp4_act_headroom calibration. Weight quantizers use plain max, while the NVFP4
+    activation global scales are anchored to a low percentile of the per-block amax distribution
+    to leave headroom for activations larger than any seen during calibration. Module coverage
+    matches nvfp4_default-kv_fp8_cast.
+quantize:
+  algorithm:
+    method: nvfp4_act_headroom
+    anchor_percentile: 1
```
realAsma, line 40: should we add rho as well here since rho is also critical in setting global scale.

modelopt/torch/quantization/model_calib.py
```diff
@@ -490,6 +491,113 @@ def sync_quantizer_amax_across_tp(
...
+    print_rank_0(
+        f"nvfp4_act_headroom: calibrating {len(swapped)} NVFP4 activation quantizer(s) "
+        f"(anchor_percentile={anchor_percentile}, rho={rho})."
+    )
+    # max_calibrate runs the forward loop once: the swapped-in calibrators accumulate their
+    # per-block histograms in the same pass that collects max stats for every other quantizer.
+    # The calibrators are restored afterwards so this algorithm does not leak into a later
+    # calibration of the same model, and so a repeat run starts from a fresh histogram.
+    try:
+        max_calibrate(
```
realAsma: Should we call mse here?

cjluo-nv: no. We deliberately focusing on max calib for weights here and only goal is to be able to generate the activation scales. Doing MSE will make the implementation more complex. Later on I hope we can find a way to split weight and activation calib algorithms separately.

meenchen (approved):
LGTM implementation-wise. Curious to know:

- Do you see significant differences between `amax` and `rho × anchor`
- With this calibration, do we expect to use the headromm act in most of the case, and the calibrated amax would be the fallback for rare cases?

cjluo-nv:
> * Do you see significant differences between `amax` and `rho × anchor`

Yes: Please see this file: https://github.com/NVIDIA/Model-Optimizer/blob/72dbfb9ed2a20ec901671708c29799f622a83426/experimental/nvfp4_global_scale_study/input_scale_comparison.html

> * With this calibration, do we expect to use the headromm act in most of the case, and the calibrated amax would be the fallback for rare cases?

So far I find this method to be very effective for nemotron models but not for Qwen3.6. I think we can give it a try for other OSS models and see how it compares with the default max algo.

cjluo-nv:
Thanks — all three points were valid. Addressed in f4ccecce0.

**On the guardrail (`nvfp4_act_headroom.py:179`):** your reading was right, and digging into it turned up a bigger problem of my own making. The ratio was incoherent because an earlier commit of mine (2076b6d0a) had changed the bound from a configurable high percentile to the literal observed max, while leaving the range ratio measured against the percentile. I made that change to satisfy a "calibration data is never clipped" claim I had written into the docstring myself.

Measuring it showed the change was wrong. Flooring the global scale at the literal max means a single freak block drags the scale up until every other block's FP8 block scale falls below subnormal. On a tensor with one block seven orders of magnitude above the rest, that flushes **99.998% of elements to zero**, against 6.7% when the rare blocks are clipped instead; on a benign wide-range tensor the two differ by 0.4% relative MSE. So the strict floor buys almost nothing where it is safe and is destructive where it is not.

f4ccecce0 therefore restores the top of the calibrated range as a configurable percentile — `upper_percentile`, default 99.99, with `100` selecting the literal observed max — and states the trade-off in the docstrings instead of asserting a guarantee the default does not provide. That overclaim, not the percentile, was the actual defect.

This makes the guardrail self-consistent again rather than patched: `upper / anchor` *is* the ratio that decides the fallback, so the reported range and the "raise rho" advice both refer to the bound that is actually operative. I did add the rare-outlier test you asked for — it now asserts that the default clips the outlier to protect the bulk, and that `upper_percentile=100` chases it instead.

**Imports:** moved `max_calibrate` and the `model_calib` module import to the top of the test file. There was no circular or optional-dependency reason for keeping them local.

**Licensing:** good catch — both new files now carry the canonical 2026 header. They had 2024 because I copied the header from `calib/max.py`; the pre-commit hook accepted it only because of `--allow-past-years`.

**Design justification:** added to the PR description. Short version: `HistogramCalibrator` histograms tensor values into linear, dynamically re-ranged bins and its percentile mode returns an amax that clips a high-tail fraction. This needs a different statistic (per-block amaxes), different bin spacing (log2 — the anchor is read from the low tail, where linear bins have essentially no resolution), and a different reduction (a low-percentile anchor scaled up, not an upper-tail clip). Reuse would change its collection semantics for every existing caller; composition would still leave the log-spacing and the derived statistic unaddressed. The collector is a fixed 512-bin int64 histogram per quantizer, so the bounded-memory rationale for using a histogram at all is preserved.

Verification: the restored calibrator is bit-identical to the reference implementation at both `upper_percentile=99.99` and `100` across gaussian, heavy-tail, sparse, wide-range, uniform and rare-outlier inputs. Re-ran the shipped recipe on a 30B hybrid Mamba/attention MoE — exports a standard NVFP4 checkpoint (19G vs 62G BF16, `quant_algo: NVFP4`, 6260 `input_scale`, 6267 `weight_scale`).

CHANGELOG.rst
```diff
@@ -6,6 +6,10 @@ Changelog
...
+
+- Add the ``nvfp4_act_headroom`` calibration algorithm for NVFP4 **activation** global scales. Plain ``max`` sets a tensor's global scale from the largest per-block amax seen during calibration, leaving no room above it, so any activation larger than the calibration max saturates. ``nvfp4_act_headroom`` instead anchors the global scale to a low percentile of the per-block amax distribution, ``amax = max(rho * anchor, floor)``, placing the calibrated blocks in the lower part of the FP8 block-scale range and leaving the rest as headroom; a high-percentile ``floor`` keeps the scale at or above the calibrated blocks so calibration data is never clipped, and a guardrail warns when the observed range is wider than FP8 scales can represent. Tunable via ``anchor_percentile`` (default 1) and ``rho`` (default 16384). Applies only to NVFP4 dynamic-block input quantizers; weight quantizers and all other quantizers keep plain ``max``. Ships ``modelopt_recipes/general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml``, which mirrors ``nvfp4_default-kv_fp8_cast`` (dynamic NVFP4 W4A4 plus FP8 KV-cache cast) with only the calibration algorithm swapped, so it exports a standard NVFP4 checkpoint.
```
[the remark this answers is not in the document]
cjluo-nv: Addressed in f4ccecce0.

You were right that the original wording overclaimed, and the underlying code has since changed too. The entry no longer says a high percentile prevents clipping. It now states that `upper` — the per-block amax at `upper_percentile`, default the 99.99th percentile — is the top of the range the scale commits to representing, that blocks above it are clipped, and *why* that is the default: flooring at the literal max lets one freak block drag the global scale up until every other block's FP8 block scale falls below subnormal (measured: 99.998% of elements flushed to zero on such a tensor, versus 6.7% when the rare blocks are clipped). `upper_percentile=100` is documented as the opt-in that guarantees no calibration data is clipped.

On the guardrail: it now evaluates the range against the bound that is actually operative, so the ratio it reports and the advice it gives refer to the same quantity.

On the export limitation: that applied to an activation-only recipe in an earlier revision of this PR, which has since been dropped — the shipped recipe is W4A4 with NVFP4 weights, so the format is inferred correctly and `input_scale` survives export (verified end-to-end: 6260 `input_scale`, 6267 `weight_scale`, `quant_algo: NVFP4`).

modelopt/torch/quantization/calib/nvfp4_act_headroom.py
cjluo-nv: Good catch, and chasing it turned up a bigger problem that was mine.

The ratio was incoherent because an earlier commit of mine (2076b6d0a) moved the bound from a configurable high percentile to the literal observed max, while leaving the range ratio measured against the percentile. I made that change to satisfy a "calibration data is never clipped" claim I had written into the docstring myself.

Measuring it showed the change was wrong. Flooring at the literal max means one freak block drags the global scale up until every other block's FP8 block scale falls below subnormal: on a tensor with a single block seven orders of magnitude above the rest that flushes **99.998% of elements to zero**, against 6.7% when the rare blocks are clipped instead. On a benign wide-range tensor the two differ by 0.4% relative MSE. So the strict floor buys almost nothing where it is safe and is destructive where it is not.

f4ccecce0 restores the top of the calibrated range as a configurable percentile (`upper_percentile`, default 99.99, `100` = literal observed max) and states the trade-off in the docstrings instead of asserting a guarantee the default does not provide. That overclaim, not the percentile, was the real defect.

So rather than patching the ratio, this makes it self-consistent: `upper / anchor` **is** the ratio that decides the fallback, so the reported range and the "raise rho" advice both refer to the operative bound. I did add the rare-outlier test you asked for — it asserts the default clips the outlier to protect the bulk, and that `upper_percentile=100` chases it instead.

tests/unit/torch/quantization/test_nvfp4_act_headroom.py
cjluo-nv: Done in f4ccecce0 — both this one and the `model_calib` import in `test_shared_state_knobs_are_forwarded_to_max_calibrate` are now in the module-level import block. There was no circular or optional-dependency reason for keeping them local.

modelopt_recipes/general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml
cjluo-nv: Agreed — added in 6a302f313, along with `upper_percentile`, so all three knobs that feed the global scale are visible at their defaults:

```yaml
  algorithm:
    method: nvfp4_act_headroom
    # amax = max(rho * anchor, upper), where anchor and upper are the per-block activation
    # amaxes at anchor_percentile and upper_percentile. All three are shown at their defaults.
    # Set upper_percentile to 100 to use the literal observed max, which never clips
    # calibration data but lets a single outlier block dominate the global scale.
    anchor_percentile: 1
    upper_percentile: 99.99
    rho: 16384
```

The values match the defaults, so calibration behavior is unchanged; it just makes the tunables and the formula they feed discoverable from the recipe.

modelopt/torch/quantization/model_calib.py
cjluo-nv: Good question. It composes cleanly — `mse_calibrate` itself starts with `max_calibrate(...)` and then refines only weights, so the swapped-in activation calibrators still collect during that forward pass and the MSE weight loop never touches them. I checked: substituting `mse_calibrate` here leaves the activation scale bit-identical (`19274.088` either way).

Two caveats before we do it:

1. **For this recipe it would currently be a no-op on weights.** With `fp8_scale_sweep=True` the MSE weight path applies only to *static* NVFP4 weights (`_uses_modelopt_fp8_weight_scales` requires `is_nvfp4_static`), and the shipped recipe uses dynamic-block NVFP4 weights — I measured the weight amax unchanged at `0.088378318`. It would need `fp8_scale_sweep=False` (the multiplier search) or static weights to do anything.
2. **It should be a knob, not a hardcoded swap**, since weight calibration is an orthogonal axis to the activation scale this algorithm sets.

Worth noting there is a real gap here: chaining does *not* work as a substitute. `algorithm: [mse, nvfp4_act_headroom]` has the second step's `max_calibrate` reset the MSE weights, and `[nvfp4_act_headroom, mse]` has the MSE step's internal `max_calibrate` reset the activation scales. So a knob on this call is currently the only way to get MSE weights and headroom activations in one pass.

Happy to add a weight-calibration option in this PR if you'd like it here, or keep this PR to the activation scale and do it as a follow-up — your call.

modelopt/torch/quantization/model_calib.py
```diff
@@ -490,6 +491,118 @@ def sync_quantizer_amax_across_tp(
     _finalize_with_shared_state(model, weight_patterns)
 
 
+def _is_nvfp4_dynamic_input_quantizer(name: str, module: nn.Module) -> bool:
+    """True for an enabled NVFP4 (E2M1 + E4M3 scale) dynamic-block *input* quantizer.
+
+    These carry their per-tensor global scale in ``_amax`` and are calibrated like plain max
```
meenchen, line 497: > _Bot comment._

The prior `SequentialQuantizer` correctness issue is still present in the current branch. For an input wrapper, `named_modules()` exposes leaves as names such as `layer.input_quantizer.0`, so this suffix check rejects every wrapped leaf and `nvfp4_act_headroom` silently falls back to max for them. Please traverse each parent input quantizer through `_iter_leaf_quantizers` (or otherwise preserve the parent's input role while testing each leaf), and add a regression test with an NVFP4 leaf inside `SequentialQuantizer`.

modelopt_recipes/general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml
```diff
@@ -0,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
meenchen: > _Bot comment._

This new file still uses a 2024 copyright header, while the repository's canonical `LICENSE_HEADER` says 2026. Thus it does not qualify for the exact-standard-header auto-approval exception and needs either alignment with the canonical header or explicit human licensing sign-off.

meenchen (commented):

Design review: this solves bounded-memory selection of an NVFP4 activation global scale from per-block amaxes while retaining outlier headroom. The main alternatives are composing/extending the existing `HistogramCalibrator` with `reduce_block_amax`, max calibration plus post-processing, or Torch/NumPy/SciPy percentile primitives (Torch, NumPy, and SciPy are already dependencies). The updated PR body now reasonably explains why the existing linear value histogram and those composition paths do not provide the required derived statistic and low-tail log resolution, so the design concern is addressed.

Re-review: the guardrail/documentation concern and local-import concern are addressed. However, the critical `SequentialQuantizer` issue remains in the current branch: leaf names gain a numeric suffix and fail the `endswith("input_quantizer")` check, so those activations silently use max calibration. The promised regression test is also absent. Additionally, one new recipe still has a 2024 header rather than the canonical 2026 `LICENSE_HEADER`; this requires human licensing sign-off (and appears inconsistent with the author's statement that both new-file headers were fixed).

cjluo-nv:
Both remaining points were valid and are fixed in 136e0361b.

**SequentialQuantizer.** You're right, and I was wrong to scope this out. I had treated it as theoretical; it isn't. A perfectly ordinary `quant_cfg` whose `input_quantizer` cfg is a list builds a `SequentialQuantizer`, and all three names fail the check:

```
fc.input_quantizer     SequentialQuantizer  -> False   (fails the isinstance check)
fc.input_quantizer.0   TensorQuantizer      -> False   (name has a numeric suffix)
fc.input_quantizer.1   TensorQuantizer      -> False
```

so those NVFP4 activations silently got plain max, exactly as you describe. The selector now matches on the *container* name and walks the leaves through the existing `_iter_leaf_quantizers` helper, which is the same traversal the amax-sync helpers use. The regression test asserts all four leaves of a two-layer model are swapped and that each ends up with the headroom scale rather than a max scale.

For the record on the "promised" test: the earlier CodeRabbit thread on this was auto-annotated "✅ Addressed in commit 14a74cb", which was incorrect — that commit only changed the recipe. I then read the auto-resolution as the issue being moot and explicitly declined it as out of scope rather than promising a test. That was a bad call regardless, since it is reachable from a normal recipe; thanks for pushing back on it.

**Licensing.** Also correct, and my earlier statement was wrong. This PR adds *three* files, not two — I fixed the two Python files and missed the recipe YAML, which had inherited a 2024 header from the unit file I derived it from. All three now carry the canonical 2026 header:

```
modelopt/torch/quantization/calib/nvfp4_act_headroom.py            Copyright (c) 2026
modelopt_recipes/general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml   Copyright (c) 2026
tests/unit/torch/quantization/test_nvfp4_act_headroom.py           Copyright (c) 2026
```

Tests are at 27 for the calibrator; `tests/unit/torch/quantization/` and `tests/unit/recipe/` pass with no regressions.

cjluo-nv:
Changed approach in 4620291db: `SequentialQuantizer` activation quantizers are now **explicitly unsupported** rather than silently calibrated leaf-by-leaf.

Reasoning: a `SequentialQuantizer` wraps several quantizers over one activation, while this algorithm derives a single per-tensor global scale from that activation's per-block amax distribution. The composition has no defined behavior here — calibrating each leaf independently, as my previous commit did, papers over that question rather than answering it, and would ship semantics nobody has validated. Failing loudly is the honest contract.

```python
raise NotImplementedError(
    f"nvfp4_act_headroom does not support SequentialQuantizer activation quantizers, "
    f"but {name!r} is one wrapping an NVFP4 dynamic-block quantizer. Use a single "
    f"activation quantizer config for these modules, or calibrate with 'max'."
)
```

Three details worth noting:

- **Validation runs before any calibrator is swapped**, so a rejected model is left completely untouched — no partial mutation to unwind.
- **Only `SequentialQuantizer`s that actually wrap an NVFP4 dynamic-block leaf are rejected.** One holding, say, FP8 leaves is irrelevant to this algorithm and is ignored, so unrelated recipes aren't broken by an over-broad check.
- The limitation is documented on `nvfp4_act_headroom_calibrate` (`Raises:`) and in the changelog entry.

Tests cover all three: the rejection with its message, that the model is unchanged after a rejection, and that a non-NVFP4 sequential activation is ignored. 29 tests for the calibrator; `tests/unit/torch/quantization/` and `tests/unit/recipe/` pass with no regressions.

modelopt/torch/quantization/model_calib.py
realAsma: Can we call `_run_scale_calibration1` https://github.com/NVIDIA/Model-Optimizer/blob/4c3d364750a605ffaae0e4d57968023456225866/modelopt/torch/quantization/model_calib.py#L2128 here? 

our scale algorithms are focused on weight scale setting (we should update the `_run_scale_calibration` to say `_run_weight_scale_calibration` to convey this cleanly). It is orthogonal to the activation scale setting targeted in this PR.

realAsma: RB: Wdyt? Do you think what I proposed above is a good idea?

🐝 Looking into the calibration call flow and state interactions now.

Yes — I think that is the right abstraction. The temporary calibrator swap defines the activation-scale policy; then a renamed `_run_weight_scale_calibration` can independently choose `max` / `mse` / `local_hessian`. Both MSE and local-Hessian start with `max_calibrate`, so the swapped headroom collector is populated in that initialization pass, while their later refinement only changes weights. It also fixes the ordering problem with chaining two top-level algorithms.

I’d add three guardrails:

- expose a typed nested `weight_scale_algorithm` and default it explicitly to `{"method": "max"}` here, rather than inheriting the helper’s current `None -> mse` default;
- keep kwargs owned by that nested algorithm so `distributed_sync` / `shared_states` are not duplicated;
- handle `sync_expert_weight_amax` deliberately: it is currently max-only, and MSE/local-Hessian do not accept or forward it.

Then test at least max + MSE composition (and ideally local-Hessian): activation amax remains headroom-derived while the selected weight calibration takes effect. With that contract cleanup, I’d prefer your proposal over hard-wiring `max_calibrate` here.

cjluo-nv: Implemented in a596c0e4f — thanks, this is a better shape than what I had.

To be clear about the change of position: my earlier "no" was to hard-wiring `mse` here, and that objection stands. Your proposal is a different thing — it routes weight calibration through the dispatch helper that already exists, which is exactly the "split weight and activation calib algorithms separately" I was hoping for later. So it lands now rather than later.

```yaml
  algorithm:
    method: nvfp4_act_headroom
    anchor_percentile: 1        # activation policy, owned here
    upper_percentile: 99.99
    rho: 16384
    weight_scale_algorithm:     # weight policy, delegated
      method: max               # or mse / local_hessian, plus that algorithm's options
```

`_run_scale_calibration` is renamed `_run_weight_scale_calibration` as you suggested (private, and I updated the LSQ caller and its tests). `distributed_sync`, `shared_states` and `sync_expert_weight_amax` move off the top level into the nested config, which also resolves the third guardrail on its own: `MaxCalibConfig` defines `sync_expert_weight_amax` and `MseCalibConfig`/`LocalHessianCalibConfig` do not, so the union already types it per method — there was nothing to plumb, only a duplicate to delete.

Two implementation notes worth recording:

- **The default must carry `method` explicitly.** `{"method": "max"}` rather than `None` avoids the helper's `None -> mse` default, as your bot flagged. It also can't be a bare `MaxCalibConfig()`: the helper does `model_dump(exclude_unset=True)`, so an unset `method` is dropped and `calib_funcs[None]` raises `KeyError`.
- **The `field_serializer` is load-bearing, not cosmetic.** `wrapped_calib_func` does a full `config.model_dump()`, so without the sparse-shape serializer every nested default is forwarded as a kwarg and `max_calibrate` raises on `moe_calib_experts_ratio` / `layerwise`. Copying the LSQ pattern fixes it; verified the dispatch receives `{}` for the default and `{"fp8_scale_sweep": True}` for `{"method": "mse", "fp8_scale_sweep": true}`.

One caveat on the test you asked for. Asserting that the selected weight calibration *takes effect* isn't possible in the CPU unit environment: MSE on static NVFP4 needs triton, and on dynamic NVFP4 it needs CUDA (I hit `static_blockwise_fp4_fake_quant requires triton` and `amax must be a CUDA tensor` respectively). Worth knowing regardless of this PR: with `fp8_scale_sweep=True` the MSE weight path applies only to *static* NVFP4, so on this recipe's dynamic-block weights it is a no-op — a test asserting "weights changed" would pass vacuously. So the unit tests assert the dispatch and options plumbing plus survival of the headroom activation scale, and I've left the numerical effect to GPU coverage. Happy to add a GPU test if you'd prefer it in this PR.

31 tests on the calibrator; `tests/unit/torch/quantization/` and `tests/unit/recipe/` are back to the pre-existing baseline with no new failures.

realAsma (approved):
Looks great!

PR #2031, opened 2026-07-29
Add Megatron-Bridge prune & quantize launcher pipelines

kevalmorabia97:
### What does this PR do?

Type of change: new example + small launcher / modelopt-example features (backward compatible)

Adds end-to-end ModelOpt **launcher** pipelines for the Megatron-Bridge flow on Nemotron-3-Nano-30B-A3B, the minimal launcher features to run them wrapper-free from YAML, and an **in-step accuracy gate** for Minitron pruning.

**New launcher examples** (`tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/`):
- `mbridge_prune.yaml` — Minitron prune **with an in-step MMLU gate** → vLLM sanity gen (2 tasks)
- `mbridge_quantize.yaml` — FP8 quantize → unified-HF export → MMLU gate on the vLLM backend, which doubles as the deploy sanity check (3 tasks). Matches the [tutorial](https://github.com/NVIDIA/Model-Optimizer/blob/main/examples/megatron_bridge/tutorials/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/README.md).

**Prune accuracy gate** (reuses the search's own score — no separate eval step):
- `modelopt/torch/prune/plugins/mcore_minitron.py` — `MCoreMinitronSearcher` now stores the exported best `CandidateSubnet` under `state_dict["best"]` (additive; sits beside the existing `sorted_layers` key).
- `examples/megatron_bridge/prune_minitron.py` — new `--score_lower_bound`: reads `pruning_scores["best"].score` and exits non-zero if the exported model is below the floor. Score-agnostic (any `--prune_score_func`); rejected with `--prune_export_config` (manual pruning has no score).

**Launcher (`tools/launcher`)** — run single-node Megatron-Bridge one-liners directly from YAML:
- `SandboxTask.inline` — a command in the YAML, no `common/**/*.sh` wrapper (single-line; folded scalar)
- `SandboxTask.reqs` / `reqs_file` — pip-install deps in the container before the command (shell-safe; on Slurm the install is rank-0-guarded so multi-rank tasks don't race)
- `SlurmConfig.docker_user` — local-Docker user (e.g. `root`); ignored on Slurm
- `get_default_env` honors `HF_HOME` / `TRITON_CACHE_DIR` env overrides, so a non-CI user can point caches at a writable path (the shared `/cicd/hf-cache` is owned by the CI account)
- reject `args` together with `inline`

**`examples/llm_eval/lm_eval_hf.py`**:
- `--accuracy_lower_bound` — gate on the single requested task's `acc` (used by the quantize MMLU step; exits non-zero if below)
- drop ModelOpt (hf-only) args for non-`hf` backends, so `--model vllm` works on a deployable quantized checkpoint

### Usage

```bash
cd tools/launcher
# Prune (in-step MMLU gate) -> vLLM gen
uv run launch.py --yaml examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/mbridge_prune.yaml --yes
# FP8 quantize -> unified-HF export -> MMLU gate (vLLM)
uv run launch.py --yaml examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/mbridge_quantize.yaml --yes
```

### Testing

- Launcher unit tests for `inline`, `reqs`/`reqs_file`, `docker_user`, the `args`+`inline` guard, and example-resolve; ruff / mypy / bandit clean. `tests/examples/megatron_bridge/test_prune_minitron.py` now passes `--score_lower_bound=0.01` to exercise the gate path on the tiny models.
- **End-to-end on the real Nemotron-3-Nano-30B-A3B (4×B200, OCI-HSG):**
  - Prune 30B → 3B-active: `[score_gate] mmlu_10pct = 0.5196 >= 0.45 PASS`; vLLM gen coherent.
  - FP8 quantize → unified-HF export (`Detected ModelOpt fp8 checkpoint`) → MMLU on the vLLM backend `acc = 0.7077 >= 0.60 PASS`.
  - Earlier smoke on **Qwen3-0.6B** in `nemo:26.06` through the same flow.

- Is this change backward compatible?: ✅ (new state-dict key is additive; new CLI args default to off)

### Additional Information

- **Container pinning:** saving a pruned Nemotron-H to HF requires `transformers<5`, so `mbridge_prune.yaml` runs on `nemo:26.04` (26.06 drops it); quantize/export run on `nemo:26.06`.
- **`docker_user: root`** is set on all example tasks — local-Docker only (ignored on Slurm), needed so downstream tasks can read task_0's root-owned checkpoints and to read the image's root-only `/opt/Megatron-Bridge`.
- The quantize MMLU step passes `enforce_eager=True` to vLLM — for a run-once eval this skips ~17 min of CUDA-graph capture / `torch.compile` with no accuracy change.

## Summary by CodeRabbit

- **New Features**
  - Added inline shell commands and per-task Python dependency installation for launcher workflows.
  - Added configurable Docker user selection and preservation of existing cache environment settings.
  - Added evaluation accuracy and pruning score gates that fail workflows below configured thresholds.
  - Added NVIDIA Nemotron pruning and quantization workflow examples.
  - Improved backend-specific handling of ModelOpt options.

- **Documentation**
  - Documented inline commands, dependencies, variable substitution, and configuration examples.

- **Bug Fixes**
  - Strengthened task execution validation and configuration checks.


modelopt/torch/prune/plugins/mcore_minitron.py
```diff
@@ -294,6 +295,7 @@ def default_state_dict(self) -> SearchStateDict:
             "layer_scores": {},
             "sorted_layers": None,
             "all_candidates_per_constraint": {},
+            "best": None,
```
cjluo-nv: > _Bot comment._

Backward-compat break for cached pruning scores: `BaseSearcher.search()` calls `load_search_checkpoint()` with `strict=True`, which asserts `state_dict.keys() == self.state_dict().keys()`. Any `--prune_intermediate_ckpt` dir written before this change lacks `"best"`, so resuming an interrupted prune after upgrading now dies with `Keys in checkpoint don't match!` rather than restoring scores. Either don't put `best` in the state dict (see the alternative suggested in `prune_minitron.py`) or relax the load to `strict=False` for this searcher.

modelopt/torch/prune/plugins/mcore_minitron.py
```diff
@@ -267,6 +267,7 @@ class MCoreMinitronSearcher(BaseSearcher):
     sorted_layers: list[int] | None  # 1-indexed sorted list of layer numbers
     # Dict from params constraint to list of all CandidateSubnets fitting that constraint
     all_candidates_per_constraint: dict[tuple, list[CandidateSubnet]]
+    best: CandidateSubnet | None  # exported (best) candidate; None if not searched
```
cjluo-nv: > _Bot comment._

`best` collides with `BaseSearcher.best`, which is the search-result state dict (`reset_search`: `self.best: SearchStateDict = {}`) that `BaseSearcher.search() -> SearchStateDict` returns. After this change `MCoreMinitronSearcher.search()` returns a `CandidateSubnet | None`, violating the annotated contract for any direct searcher user, and the class-level annotation conflicts with the base class's attribute type (mypy attribute-override error). Suggest `best_candidate` (or `exported_candidate`) instead.

examples/megatron_bridge/prune_minitron.py
```diff
@@ -668,6 +680,18 @@ def score_func(m):
         copy_hf_ckpt_remote_code(args.hf_model_name_or_path, args.output_hf_path)
         print_rank_0(f"Saved pruned model to {args.output_hf_path} in HF checkpoint format")
 
+    # Accuracy gate: exit non-zero if pruned model's score is below the bound
+    if args.score_lower_bound is not None:
+        best = pruning_scores["best"]
```
cjluo-nv: > _Bot comment._

This gate could avoid touching the searcher at all: `pruning_scores["all_candidates_per_constraint"]` is already in the returned state dict, so the exported model's score is `max(c.score for cands in ... for c in cands if c.score is not None)`. That sidesteps both the `best` naming collision and the strict-checkpoint-key break flagged in `mcore_minitron.py`. If you keep the new key, please also fix the help text — it says "For Testing" but `mbridge_prune.yaml` uses it as a production quality gate at 0.51.

examples/llm_eval/lm_eval_hf.py
```diff
@@ -255,12 +273,53 @@ def _inject_modelopt_args_into_model_args(args):
...
-            model_args[key] = getattr(args, key)
+            if is_hf:
+                model_args[key] = getattr(args, key)
             delattr(args, key)
 
     args.model_args = model_args
 
 
+def _pop_gate_bound(args):
+    """Pop --accuracy_lower_bound off the namespace; return it (None disables the gate)."""
```
cjluo-nv: > _Bot comment._

Silently dropping the ModelOpt args for non-HF backends is a footgun: `--model vllm --quant_cfg FP8_DEFAULT_CFG` will now evaluate the *unquantized* checkpoint with no warning, and combined with `--accuracy_lower_bound` it will happily report PASS. Please raise a clear error when an explicitly-set ModelOpt arg (`quant_cfg`, `sparse_cfg`, `auto_quantize_bits`, `compress`, ...) is combined with a non-HF backend, instead of dropping it.

Also, `is_hf` misses backends that *are* HFLM-based: lm-eval registers HFLM as `hf`/`hf-auto`/`huggingface`, and `hf-multimodal` subclasses it. Those would now lose their ModelOpt args.

examples/llm_eval/lm_eval_hf.py
```diff
@@ -255,12 +273,53 @@ def _inject_modelopt_args_into_model_args(args):
...
+
+def _enforce_accuracy_gate(output_path, task, lower_bound):
+    """Read lm-eval results at output_path; exit non-zero if <task>/acc < lower_bound."""
+    if not output_path:
+        raise ValueError("--accuracy_lower_bound requires --output_path.")
+    files = glob.glob(os.path.join(output_path, "**", "results*.json"), recursive=True)
+    if not files:
+        raise FileNotFoundError(f"No results*.json under {output_path}")
+    # Sort by mtime, not path: a reused output_path nests results under a
+    # <model_name>/ dir, and lexical order would pick the wrong run's file.
```
cjluo-nv, line 302: > _Bot comment._

No tests for the gate helpers, even though they're pure and have real edge cases: picking the newest `results*.json` among nested `<model_name>/` dirs, `"acc,none"` vs `acc_stderr` key matching, the missing-`acc` `KeyError`, `--tasks a,b` rejection, and pass/fail exit codes. A small unit test writing a fake `results.json` into `tmp_path` would cover all of it cheaply.

tools/launcher/core.py
```diff
@@ -51,17 +51,20 @@ def get_default_env(experiment_title=None):
...
         "SPECDEC_BENCH_S3_SECRET": os.getenv("SPECDEC_BENCH_S3_SECRET", ""),
     }
+    # HF_HOME / TRITON_CACHE_DIR default under the shared /{title} mount, but honor
+    # an env override so a user can point them at a personally-writable path (the
+    # shared cache is owned by the CI account and blocks other users' cache locks).
     slurm_env = {
-        "TRITON_CACHE_DIR": f"/{title}/triton-cache",
-        "HF_HOME": f"/{title}/hf-cache",
+        "TRITON_CACHE_DIR": os.getenv("TRITON_CACHE_DIR", f"/{title}/triton-cache"),
+        "HF_HOME": os.getenv("HF_HOME", f"/{title}/hf-cache"),
```
cjluo-nv, line 59: > _Bot comment._

`get_default_env` now depends on ambient env, but `tests/test_core.py::TestGetDefaultEnv` still asserts `slurm_env["HF_HOME"] == "/cicd/hf-cache"` with no `monkeypatch.delenv` — those tests will fail on any machine or CI runner that exports `HF_HOME`/`TRITON_CACHE_DIR` (fairly common). Please isolate the env in the existing tests and add one covering the new override.

Also worth a doc note: the host value is injected into the *container* env, so it only works if that path is also mounted into the container — otherwise the cache silently lands somewhere non-persistent.

tools/launcher/core.py
```diff
@@ -803,7 +831,32 @@ def run_jobs(
...
+                    install = "python -m pip install " + " ".join(pkgs)
+                    # On Slurm, srun runs this inline on every rank (ntasks_per_node),
+                    # so install once on local rank 0 behind a filesystem barrier —
+                    # concurrent pip on one node corrupts the env. /tmp is a fresh
+                    # per-task container dir shared across the node's ranks, so the
+                    # marker is task-unique and rank-visible. Local single-process
+                    # runs just install (rank 0) and never wait.
+                    marker = "/tmp/.modelopt_launcher_reqs_done"  # nosec B108 — per-task container tmp, shared across a node's ranks
+                    reqs_prefix = (
+                        f'if [ "${{SLURM_LOCALID:-0}}" -eq 0 ]; then {install} && touch {marker}; '
```
cjluo-nv: > _Bot comment._

The marker isn't run-unique. Under enroot/pyxis, `/tmp` is frequently a bind-mount of the host `/tmp`, so a leftover marker from an earlier task on the same node makes every non-zero local rank skip the wait and start immediately while rank 0 is still running pip — exactly the concurrent-pip corruption this barrier is meant to prevent. Suggest folding in `${SLURM_JOB_ID:-0}_${SLURM_STEP_ID:-0}` (or the task name) and/or `rm -f` the marker before installing.

Second issue: if rank 0's install fails, the other ranks fall out of the `for _ in $(seq 600)` loop after 10 minutes with exit status 0 (from `sleep`) and run the command anyway. Better to exit non-zero when the marker never appears.

tools/launcher/docs/configuration.md
```diff
@@ -58,6 +58,69 @@ pipeline:
       gpus_per_node: 4
 ```
 
+### Inline Command (no wrapper script)
+
+A task can carry an `inline` shell command instead of a `script:` path (no
+wrapper `.sh` needed) — good for one-liner jobs like Megatron-Bridge's
+`torchrun` scripts. `<<global_vars.X>>` is resolved; `args` is ignored. The
```
cjluo-nv: > _Bot comment._

Docs say `args` "is ignored" with `inline`, but `run_jobs` raises `ValueError("`args` is only for `script`...")`. Please make the docs match the code (rejected, not ignored).

cjluo-nv (commented):

Mostly a tidy set of additive launcher/example features, and the E2E evidence in the PR body is appreciated. A few substantive issues, though:

**Correctness / backward compat**
1. Adding `"best"` to `MCoreMinitronSearcher.default_state_dict` is *not* backward compatible for cached score checkpoints: `BaseSearcher.load_search_checkpoint()` is called with `strict=True` and asserts `state_dict.keys() == self.state_dict().keys()`, so any pre-existing `--prune_intermediate_ckpt` / `modelopt_pruning_scores` dir written by an older version now fails with `Keys in checkpoint don't match!` instead of resuming. The PR body claims "new state-dict key is additive" — it is additive for *writers*, not readers.
2. `self.best` shadows `BaseSearcher.best`, which is the documented search-result state dict returned by `BaseSearcher.search() -> SearchStateDict` (`reset_search` sets `self.best: SearchStateDict = {}`). `MCoreMinitronSearcher.search()` now returns a `CandidateSubnet | None`, breaking that annotation/contract (and likely tripping mypy on the attribute override). Please rename (e.g. `best_candidate`).
3. `lm_eval_hf.py` now *silently drops* `--quant_cfg` / `--sparse_cfg` / `--auto_quantize_bits` for non-`hf` backends. A user running `--model vllm --quant_cfg FP8_DEFAULT_CFG` gets an eval of the **unquantized** model with no warning — and with `--accuracy_lower_bound` that even reports PASS. This should raise instead of silently ignoring. Also `is_hf` misses the `huggingface` alias and `hf-multimodal`, both of which are HFLM-based.

**Tests**
4. The new gate helpers (`_requested_tasks`, `_enforce_accuracy_gate`) are pure functions with real edge cases (mtime selection among nested `results*.json`, `"acc,none"` vs `acc_stderr`, missing task) and have no tests. Similarly, the new `run_jobs` logic (`reqs_prefix` shell string, `args`+`inline` `ValueError`) is untested — the added `test_examples_resolve` assertion covers YAML-level `script`/`inline` exclusivity, not the `args`+`inline` guard the PR body claims.
5. `get_default_env` now reads ambient `HF_HOME`/`TRITON_CACHE_DIR`, but the existing `tests/test_core.py::TestGetDefaultEnv` assertions (`"/cicd/hf-cache"`) weren't updated to `monkeypatch.delenv`, so they'll fail on any dev box/CI runner that exports those vars. No test covers the new override path either.

**Robustness / docs**
6. The pip barrier marker `/tmp/.modelopt_launcher_reqs_done` is not run-unique, and under enroot/pyxis `/tmp` is commonly the host `/tmp`; a stale marker from an earlier task on the same node makes non-zero local ranks skip the wait, reintroducing exactly the concurrent-pip race the barrier guards against. Ranks also proceed after the 600 s timeout even when rank 0's install failed.
7. `docs/configuration.md` says `args` "is ignored" with `inline`, but `run_jobs` raises `ValueError`.

**Design (per the complexity gate)**: `inline`/`reqs` extend the existing launcher config rather than adding a new subsystem, which is good, but they do add a third way to express a task command alongside the documented `common/**/*.sh` wrapper convention (`CLAUDE.md`: "Scripts go in `common/`") and typed `_target_` tasks — and `common/megatron_bridge/import/import.sh` shows the wrapper pattern already covers this flow (a wrapper could also do the pip install + rank-0 barrier in bash, making `reqs`/`reqs_file` unnecessary). The PR body describes the new fields but doesn't say why the wrapper-script or typed-task route was rejected; please add that rationale, and update `CLAUDE.md`'s conventions list so contributors know which of the three to reach for.

kevalmorabia97:
Addressed the first round of review comments and resolved those threads (changes in `b090a7c604`). Summary of the non-trivial ones:

**Prune score gate — `mcore_minitron.py` / `prune_minitron.py`** (@cjluo-nv)
- The searcher now stores the exported best subnet **as a dict** — `self.best = asdict(best)` — reusing `BaseSearcher.best` (the returned `SearchStateDict`). So `search()` still returns a dict (no type-contract break, no `best` name collision), and the gate reads `pruning_scores["best"]["score"]` — no re-eval, reuses existing state. Help text no longer says "For Testing".
- On the strict-checkpoint-key break: accepting it — pruning is a short one-shot step exported to HF, so resuming a pre-change intermediate `--prune_intermediate_ckpt` across a code upgrade isn't a real workflow.

**`lm_eval_hf.py`** (@cjluo-nv) — raises when a quantization/sparsity **request** arg (`quant_cfg`/`auto_quantize_bits`/`compress`/`sparse_cfg`) is set on a non-HF backend, instead of silently dropping it (inert tuning params like `calib_size` are still dropped). `is_hf` now covers `hf`/`hf-auto`/`huggingface`/`hf-multimodal`. Results file selected by `mtime`.

**`core.py`**
- reqs-barrier marker is **run-unique** (relative to the per-task `/nemo_run/code`, not `/tmp`) with **no `# nosec`**, and waiting ranks fail non-zero if rank-0's install never completes (@cjluo-nv).
- Shell-quote the script path in the reqs+script branch; `args` keep the launcher's shell-word-split convention (same as `run.Script(path, args=...)`), so a `"--flag value"` list item still expands to two args.
- Reject a task that sets both `script`+`inline` or neither; `get_default_env` env-isolation added to the existing tests (@cjluo-nv).

**`docs/configuration.md`** — `args` with `inline` is **rejected** (raises `ValueError`), not "ignored".

**Not addressed yet** (will follow up): the test-coverage suggestions (failing-gate assertion, `inline`+`args` in `test_examples_resolve`, gate-helper unit tests). The `MLM_MODEL_CFG`/`QUANT_CFG` env comment is **N/A** — these inline YAMLs pass `--hf_model_name_or_path`/`--quant_cfg` directly on the command; those env vars are only read by the `common/*.sh` wrappers, which the inline pipelines don't use.

examples/llm_eval/lm_eval_hf.py
```diff
@@ -236,16 +240,31 @@ def _add_modelopt_args(parser):
...
     args.model_args is a dict (parsed by lm-eval's MergeDictAction). The ModelOpt
     keys must be removed from the namespace so EvaluatorConfig.from_cli doesn't
-    reject them as unknown kwargs.
+    reject them as unknown kwargs. They only apply to the HF backend, so for other
+    backends (e.g. `--model vllm` on an already-quantized ckpt) they are dropped,
+    not passed into model_args (which would break the backend's constructor).
     """
     model_args = dict(args.model_args) if args.model_args else {}
+    # HFLM and its subclasses (registered as hf / hf-auto / huggingface / hf-multimodal).
+    is_hf = getattr(args, "model", "hf") in ("hf", "hf-auto", "huggingface", "hf-multimodal")
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in `b090a7c604` — the check now only fires for args that actually request quantization/sparsity (`quant_cfg`/`auto_quantize_bits`/`compress`/`sparse_cfg`); inert tuning params like `calib_size` are dropped for non-HF backends, not raised on.

examples/megatron_bridge/prune_minitron.py
```diff
@@ -668,6 +680,18 @@ def score_func(m):
...
+        assert best and best.score is not None, "Best candidate not found in pruning_scores"
+        passed = best.score >= args.score_lower_bound
+        print_rank_0(
+            f"[score_gate] final pruned model {args.prune_score_func} score = {best.score:.4f} "
+            f"(lower_bound {args.score_lower_bound}) -> {'PASS' if passed else 'FAIL'}"
+        )
+        if not passed:
+            sys.exit(1)
```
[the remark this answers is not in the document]
kevalmorabia97: Moved the gate to immediately after `mtp.prune()`, before either save path. A below-threshold model is no longer written to disk, so a rerun can't early-exit on a stale checkpoint and report success without re-checking the bound.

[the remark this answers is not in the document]
kevalmorabia97: Update: reverting this. By design the pruned checkpoint should always be written — it's useful for inspecting the pruned model even when accuracy is low — so the score gate stays **after** the save as a post-save sanity assertion: it sets the job's exit status (fails CI if below the bound) but does not block saving. Re-opening for visibility.

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/mbridge_prune.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+  global_vars:
+    output_dir: /cicd/megatron-bridge/Nemotron-3-Nano-30B-A3B-Pruned-A3.0B
+
+  # 1) Prune to a pruned HF checkpoint (HF save needs transformers<5).
+  # --score_lower_bound fails the job if the pruned model's MMLU drops below the floor.
+  task_0:
+    reqs: "transformers<5"
+    environment:
+      - LAUNCH_SCRIPT: torchrun --nproc_per_node 4
```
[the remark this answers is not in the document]
kevalmorabia97: N/A for these examples: the pipelines use **inline** commands that pass `--hf_model_name_or_path` (and `--quant_cfg` for quantize) directly on the command line, so config already has one source of truth. `MLM_MODEL_CFG`/`QUANT_CFG` env vars are only read by the `common/megatron_lm/*.sh` wrapper scripts, which these wrapper-free YAMLs don't use.

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/mbridge_quantize.yaml
```diff
@@ -0,0 +1,59 @@
...
+  # 1) FP8 Quantize and export it to a deployable unified-HF checkpoint.
+  task_0:
+    environment:
+      - LAUNCH_SCRIPT: torchrun --nproc_per_node 4
+    inline: >-
+      $LAUNCH_SCRIPT modules/Model-Optimizer/examples/megatron_bridge/quantize.py
+      --hf_model_name_or_path nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16
+      --trust_remote_code
+      --tp_size 4
+      --quant_cfg MAMBA_MOE_FP8_CONSERVATIVE_CFG
```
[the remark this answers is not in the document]
kevalmorabia97: Same as the prune YAML: this is a wrapper-free **inline** pipeline that passes `--quant_cfg MAMBA_MOE_FP8_CONSERVATIVE_CFG` and `--hf_model_name_or_path` on the command directly. `MLM_MODEL_CFG`/`QUANT_CFG` are only consumed by the `common/*.sh` wrappers, which aren't used here.

examples/llm_eval/lm_eval_hf.py
kevalmorabia97: The gate is now exercised end-to-end: `test_lm_eval_hf` runs with `--accuracy_lower_bound 0.1` + `--output_path`, so `_enforce_accuracy_gate` reads the results file and enforces the bound (PASS path) on a real run. Extended the existing functional test rather than adding separate pure-helper unit tests, to keep the suite minimal.

examples/megatron_bridge/prune_minitron.py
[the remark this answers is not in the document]
kevalmorabia97: Decision: keeping the gate **after** the save. By design the pruned checkpoint should always be written — it's useful for inspecting the pruned model even at low accuracy — so the gate is a post-save sanity assertion that sets the job's exit status but doesn't block saving. (In the launcher pipelines each run also writes to a fresh `/scratchspace` dir, so there's no stale-checkpoint early-exit.)

modelopt/torch/prune/plugins/mcore_minitron.py
```diff
@@ -641,6 +642,7 @@ def search_best_arch_by_metrics(self) -> dict:
 
         dist.barrier()
         best = max(top_k_candidates, key=lambda x: x.score)  # type: ignore[arg-type, return-value]
+        self.best = asdict(best)
```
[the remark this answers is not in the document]
kevalmorabia97: `self.best` is left to subclasses to assign: `BaseSearcher.run_search` is `@abstractmethod` and `reset_search` only inits it to `{}`; `autonas` assigns it as well. Storing `asdict(best)` keeps it a **dict** (`SearchStateDict`), so `search()`'s return contract is preserved — and `mtn.search` returns `state_dict()`, not `search()`'s value, so there's no external impact. Keeping `self.best` per the reuse-existing-state approach.

tests/examples/megatron_bridge/test_prune_minitron.py
```diff
@@ -70,6 +70,7 @@ def test_prune_minitron(tmp_path, num_gpus, create_teacher, megatron_format):
         seq_length=16,
         prune_target_params=prune_target_params,
         prune_score_func="mmlu_1pct_bs32",
+        score_lower_bound=0.01,  # exercise the score gate
```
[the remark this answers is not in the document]
kevalmorabia97: Deferred: a real-script failing-gate assertion means a full GPU prune run with an impossible bound (`score_lower_bound=1.1`). Leaving open — the gate logic itself is now covered on the eval side via `test_lm_eval_hf` with `accuracy_lower_bound`.

tools/launcher/tests/test_examples_resolve.py
```diff
@@ -76,15 +76,28 @@ def test_example_yaml_valid(path):
...
+            isinstance(inline, str) and inline.strip()
+        ), f"{path}:{name}: task needs a `script` or `inline`"
+        assert not (script and inline), f"{path}:{name}: set only one of `script`/`inline`"
+
+        # nemo-run's --yaml CLI layer rejects multi-line override values, so an
+        # inline command must stay single-line (use a folded scalar `>-`).
+        if inline:
+            assert "\n" not in inline.strip(), (
+                f"{path}:{name}: `inline` must be single-line (use folded `>-`, chain with `&&`)"
+            )
```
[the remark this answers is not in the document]
kevalmorabia97: Runtime now rejects `inline`+`args` (added the check in `run_jobs`). The suggested static assertion in `test_examples_resolve` is a reasonable add — leaving open to decide whether to include it.

tests/examples/megatron_bridge/test_prune_minitron.py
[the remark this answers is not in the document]
kevalmorabia97: Not adding a dedicated test for this — the gate is a trivial `if best_score < bound: sys.exit(1)`, and the eval-side path is already exercised by `test_lm_eval_hf`. A full GPU prune run to assert the branch isn't worth the cost. Resolving.

tools/launcher/tests/test_examples_resolve.py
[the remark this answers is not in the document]
kevalmorabia97: Not adding a dedicated test — this is a one-line guard in `run_jobs` (`raise ValueError` on `inline`+`args`). Resolving.

cjluo-nv (commented):

Most of the previous round is genuinely resolved (see below); what's left are owner-judgment calls plus one test gap, so I'd like a human sign-off rather than auto-approving.

**Confirmed fixed:** `self.best = asdict(best)` preserves the `BaseSearcher.search() -> SearchStateDict` contract (base returns `self.best`; `reset_search()`'s comment anticipates `best` in `default_state_dict`; `mtn.search()` returns `state_dict()`, so `pruning_scores["best"]["score"]` resolves); non-HF backends now *raise* on quantization/sparsity request args and `is_hf` covers all four HFLM aliases; results file selected by `mtime`; `--accuracy_lower_bound` validates `--output_path`/single-task before the expensive eval; `TestGetDefaultEnv` env isolation added; docs say `args` is rejected (not ignored); the reqs barrier now `rm -f`s the marker and waiting ranks fail via `[ -f {marker} ]`; `script`/`inline` exclusivity enforced in `run_jobs`.

**Still warranting a look:**

- :speech_balloon: Author decided to *accept* the strict-checkpoint-key break (`"best"` added to `default_state_dict`) because pruning is a one-shot step — reasonable, but flagging for human sign-off because the failure mode for anyone resuming a pre-upgrade `--prune_intermediate_ckpt` / `modelopt_pruning_scores` dir is an opaque `assert state_dict.keys() == self.state_dict().keys(), "Keys in checkpoint don't match!"` with no hint to delete the dir, and the mitigation is one line (override `load_search_checkpoint(strict=False)` in `MCoreMinitronSearcher`, or derive the score from the already-present `all_candidates_per_constraint`). Owner's call, but worth a conscious yes/no.

- :speech_balloon: Author declined new tests for the `run_jobs` guard as "a one-line guard" — the guard itself, agreed, but the `reqs_prefix` construction (shlex quoting of `reqs`/`reqs_file`, the rank-0 barrier shell string, and the `reqs` + `script` → `run.Script(inline=...)` branch) is the most intricate new code in this PR and is currently exercised by nothing. `tests/test_core_extended.py::TestRunJobsExtended` already patches `core.run.Script` and asserts on the call kwargs, so a test asserting the `inline=` string for a `reqs`/`inline` task is ~10 lines. Both failing-gate paths (`--accuracy_lower_bound` above the score, `--score_lower_bound` above the score) also remain untested; the prune one is fairly argued as too expensive, but the `lm_eval` one could be a cheap unit test of `_enforce_accuracy_gate` against a synthetic `results.json`.

- :speech_balloon: Design rationale (raised previously): the PR body describes `inline`/`reqs` but still doesn't say why the existing `common/**/*.sh` wrapper convention (which `common/megatron_bridge/import/import.sh` already demonstrates for this flow, and which could do the pip install + barrier in bash) or typed `_target_` tasks were rejected. `tools/launcher/CLAUDE.md` still lists only "Scripts go in `common/`" under key conventions, so contributors now have three ways to express a task command with no guidance on which to pick. Please add a sentence of rationale to the PR body and a line to `CLAUDE.md` (`docs/configuration.md` is updated nicely already).

- Multi-node `reqs` barrier (nit): the marker is created in the per-task working dir (`/nemo_run/code`), which is the packaged code dir mounted from the shared job dir — i.e. shared across *nodes*, not node-local. On a multi-node task, node B's non-zero local ranks can observe node A's marker and skip the wait before node B's local rank 0 has installed anything. Folding `${SLURM_NODEID:-0}` into the marker name (or using a node-local path) would make it truly per-node; alternatively document `reqs` as single-node only.

- `score_lower_bound=0.01` in `test_prune_minitron` (nit): couples the GPU test's pass/fail to a tiny random-init model's MMLU score. Low risk (4-choice MMLU floor is ~25%), but it is a new flake surface for a test whose purpose is the prune path, not accuracy.

ChenhanYu (approved):

examples/llm_eval/lm_eval_hf.py
```diff
@@ -277,4 +350,18 @@ def _inject_modelopt_args_into_model_args(args):
...
+    if lower_bound is not None:  # fail fast before the (expensive) eval
+        if not output_path:
+            raise ValueError("--accuracy_lower_bound requires --output_path.")
+        tasks = _requested_tasks(args)
+        if len(tasks) != 1:
+            raise ValueError(
+                f"--accuracy_lower_bound needs exactly one --tasks (got {tasks or 'none'})."
+            )
+        gate_task = tasks[0]
     cli.execute(args)
```
cjluo-nv, line 347: > _Bot comment._

**Simplification (main one).** The gate re-discovers the results file on disk (`glob(**/results*.json)` + `max(key=getmtime)` + `json.load`) purely to recover a number the harness just computed. lm-eval's CLI entrypoint has historically returned the results dict (`cli_evaluate` ends with `return results`); if `HarnessCLI.execute()` does too on the pinned ≥0.4.10, this becomes:

```python
results = cli.execute(args)
if lower_bound is not None:
    _enforce_accuracy_gate(results["results"], gate_task, lower_bound)
```

That deletes the `glob`/`os`/`json` imports, the `FileNotFoundError` path, the mtime-vs-lexical-order comment, **and** the `--output_path` requirement (which is currently a user-visible constraint that exists only as an implementation detail of reading the file back — note the quantize YAML has to pass `--output_path /scratchspace/mmlu_results` solely to satisfy it). Please check `HarnessCLI.execute`'s return value; if it genuinely returns `None`, a one-line comment saying so would justify keeping the disk read.

examples/llm_eval/lm_eval_hf.py
```diff
@@ -277,4 +350,18 @@ def _inject_modelopt_args_into_model_args(args):
     _add_modelopt_args(run_parser)
     args = cli.parse_args()
     _inject_modelopt_args_into_model_args(args)
+    lower_bound = _pop_gate_bound(args)
```
cjluo-nv: > _Bot comment._

**Simplification.** Two small ones here:

- `_pop_gate_bound` is a 5-line function + docstring for what is a one-liner: `lower_bound = vars(args).pop("accuracy_lower_bound", None)`. Same for the `hasattr`/`delattr` dance inside it.
- The `--accuracy_lower_bound requires --output_path` check is duplicated: here (line 357) and again at the top of `_enforce_accuracy_gate`. Keep the fail-fast one and drop the second (or vice versa) — right now a reader has to check both to know which fires.

Also `_requested_tasks` is more general than the input warrants (`str`/list + comma-split + `str()` coercion + strip). After `cli.parse_args()`, `args.tasks` is either a string or a list of strings, so `tasks = args.tasks if isinstance(args.tasks, list) else str(args.tasks or "").split(",")` covers it inline and removes a helper.

examples/llm_eval/lm_eval_hf.py
```diff
@@ -236,31 +240,100 @@ def _add_modelopt_args(parser):
...
+    # Args that actually request quantization/sparsity. The other _MODELOPT_ARG_KEYS are inert
+    # tuning knobs (calib_*, auto_quantize_* params) with non-None argparse defaults, so keying
+    # off "truthy" would false-positive on them — only the request args imply the wrong backend.
+    quant_request_keys = {"quant_cfg", "auto_quantize_bits", "compress", "sparse_cfg"}
     for key in _MODELOPT_ARG_KEYS:
         if hasattr(args, key):
-            model_args[key] = getattr(args, key)
+            value = getattr(args, key)
+            if is_hf:
+                model_args[key] = value
```
cjluo-nv: > _Bot comment._

**Simplification.** The non-HF validation is woven into the injection loop, which forces a second key set (`quant_request_keys`) plus a 4-line comment explaining why `elif value:` alone would misfire. Hoisting it makes both halves single-purpose and reports all offending flags at once:

```python
if not is_hf:
    requested = [k for k in ("quant_cfg", "auto_quantize_bits", "compress", "sparse_cfg")
                 if getattr(args, k, None)]
    if requested:
        raise ValueError(
            f"{', '.join('--' + k for k in requested)} request quantization/sparsity, which only "
            f"applies to the HF backend; with --model {args.model!r} the checkpoint must already "
            "be quantized/sparsified."
        )

for key in _MODELOPT_ARG_KEYS:
    value = vars(args).pop(key, None)
    if is_hf and hasattr(...):
        model_args[key] = value
```

(Keep the `hasattr` semantics you need for the HF path — the point is that the loop then does one thing.)

cjluo-nv (commented):

Re-review at head (post-`d669c23`). Most of the previous round is genuinely fixed; what remains is (a) the design/docs ask that has now been raised twice without a change, and (b) the `lm_eval_hf.py` gate, which the operator asked me to look at for simplification — it can be materially shorter.

**Confirmed resolved since the last round:** non-HF backends now *raise* on quantization/sparsity request args (and only on those — `calib_size`/`auto_quantize_method` defaults no longer false-positive); `is_hf` covers `hf`/`hf-auto`/`huggingface`/`hf-multimodal`; results file selected by `mtime`; `--accuracy_lower_bound` validates `--output_path` + single-task *before* `cli.execute`; `script`/`inline` exclusivity + `args`+`inline` rejected in `run_jobs`; script path shell-quoted in the reqs+script branch; reqs barrier `rm -f`s the marker and waiting ranks fail via `[ -f {marker} ]`; `TestGetDefaultEnv` env isolation added; `docs/configuration.md` documents `args` as rejected (not ignored) and the new `inline`/`reqs` fields nicely.

**Answering the specific question — yes, `examples/llm_eval/lm_eval_hf.py` can be simplified.** The gate is currently ~70 added lines across four helpers; I count three reductions that don't lose behavior (details inline):
1. The whole `glob` + `mtime` + `json` re-read (and the hard `--output_path` requirement, and its duplicated validation) exists only because the results are read back off disk. If lm-eval's `HarnessCLI.execute(args)` returns the results dict (as `cli_evaluate` historically did), `results["results"][task]` gives the same number in-process and `_enforce_accuracy_gate` collapses to ~6 lines with no `--output_path` coupling. Worth 5 minutes to check against the pinned lm-eval ≥ 0.4.10.
2. `_pop_gate_bound` is a 5-line function for `vars(args).pop("accuracy_lower_bound", None)`; `_requested_tasks` is a triple-nested comprehension for what is already a normalized `--tasks` value.
3. The non-HF validation is interleaved with the injection loop and needs a second key set (`quant_request_keys`) plus a 4-line comment to explain itself; hoisting it above the loop makes both halves obvious and reports all offending flags at once.

Separately, worth deciding whether this gate belongs in the shared example CLI at all: it only serves `lm_eval_hf.py`, while `mmlu.py`, `simple_evals.py`, `livecodebench.py` and `lm_eval_tensorrt_llm.py` remain ungated, and the existing launcher convention for exactly this need is a wrapper script + upstream `--lower-bound` (`common/megatron_lm/quantize/quantize.sh`).

**Still unresolved from previous rounds:**
- :speech_balloon: Design rationale (asked in both prior rounds): the PR body still doesn't say why the `common/**/*.sh` wrapper route (which `common/megatron_bridge/import/import.sh` and `quantize.sh` already demonstrate for this flow, including pip install + a bash barrier) or typed `_target_` tasks were rejected. `tools/launcher/CLAUDE.md` is untouched and still says only "Scripts go in `common/`" + "Set `MLM_MODEL_CFG`/`QUANT_CFG`" — which is precisely the rule the new wrapper-free examples don't follow (that mismatch is what generated the CodeRabbit `MLM_MODEL_CFG` findings you had to withdraw twice), and `docs/configuration.md` still labels typed config "recommended". Please add one or two sentences of rationale to the PR body and a line to `CLAUDE.md` so contributors know which of the three task forms to reach for.
- :speech_balloon: Author declined tests as "a one-line guard" — agreed for the guard, but the `reqs_prefix` construction (shlex quoting, the rank-0 barrier shell string, and the `reqs` + `script` → `run.Script(inline=...)` branch) is still exercised by nothing. `tests/test_core_extended.py::TestRunJobsExtended` already patches `core.run.Script` and asserts call kwargs, so asserting the `inline=` string for a `reqs`+`inline` task is ~10 lines. The new tests added in `test_yaml_formats.py` only cover dataclass field resolution.
- :speech_balloon: Author accepted the strict-checkpoint-key break from adding `"best"` to `default_state_dict` (pruning is one-shot). Reasonable, but the failure mode for anyone resuming a pre-upgrade `modelopt_pruning_scores` dir is an opaque `assert ... "Keys in checkpoint don't match!"` — worth a conscious owner yes/no since the mitigation is one line.
- Multi-node `reqs` barrier (nit, see inline): the marker still isn't per-node, and `rm -f` adds its own small race window.
- `score_lower_bound=0.01` in `test_prune_minitron` couples a prune-path test to a random-init model's MMLU score (low risk, new flake surface).

No licensing changes and no prompt-injection attempts in the PR content.

tools/launcher/core.py
```diff
@@ -803,7 +833,37 @@ def run_jobs(
...
+                    pkgs = ["-r", shlex.quote(task.reqs_file)] if task.reqs_file else []
+                    pkgs += [shlex.quote(tok) for tok in shlex.split(task.reqs or "")]
+                    install = "python -m pip install " + " ".join(pkgs)
+                    # On Slurm, srun runs this inline on every rank (ntasks_per_node), so install
+                    # once on local rank 0 behind a filesystem barrier — concurrent pip on one node
+                    # corrupts the env. The marker is relative to the per-task working dir
+                    # (/nemo_run/code, fresh per task and shared across the node's ranks), so it is
+                    # run-unique (no stale-marker reuse); rank 0 clears it first, and other ranks
+                    # fail if it never appears (rank-0 install error). Local single-process runs
+                    # just install as rank 0 and never wait.
```
cjluo-nv: > _Bot comment._

The barrier is better than last round (`rm -f` + waiting ranks failing on a missing marker), but the comment's claim that the marker is "fresh per task and shared across the node's ranks" doesn't hold in two cases worth either fixing or documenting:

1. **Not node-local.** On Slurm the CWD (`/nemo_run/code`) comes from the packaged code dir under the shared job dir, so on a multi-node task node B's non-zero local ranks can see node A's marker and skip the wait before node B's local rank 0 has installed anything. Folding `${SLURM_NODEID:-0}` into the name (or using a node-local path) fixes it; otherwise please document `reqs` as single-node only.
2. **`rm -f` adds its own window.** A non-zero rank that checks `[ -f {marker} ]` in the instant before rank 0's `rm -f` still observes a stale marker from a previous task in the same experiment (same code dir) and proceeds. Making the marker name unique per task/step (e.g. `${SLURM_JOB_ID:-0}_${SLURM_STEP_ID:-0}`) removes the need for `rm -f` entirely.

This whole `reqs_prefix` string (quoting + barrier + the `reqs`+`script` inline branch) is also the most intricate new code in the PR and has no test; `TestRunJobsExtended` already patches `core.run.Script`, so asserting the generated `inline=` string is ~10 lines and would pin both the quoting and the barrier shape.

examples/llm_eval/lm_eval_hf.py
kevalmorabia97: Done — hoisted the non-HF check before the loop; the loop now does one thing (pop + inject for HF). Also reports all offending flags at once.

examples/llm_eval/lm_eval_hf.py
kevalmorabia97: Done — inlined `_pop_gate_bound` to `vars(args).pop(...)`, inlined `_requested_tasks` to the isinstance one-liner, and removed the duplicate `--output_path` check from `_enforce_accuracy_gate` (kept the fail-fast one).

examples/llm_eval/lm_eval_hf.py
kevalmorabia97: Checked: `HarnessCLI.execute()` is annotated `-> None` and has no `return` statement — the results dict is consumed inside `Run._execute()` and never propagated up. Added a one-line comment to `_enforce_accuracy_gate` explaining why the disk read is required.

examples/llm_eval/lm_eval_hf.py
```diff
@@ -277,4 +331,18 @@ def _inject_modelopt_args_into_model_args(args):
     _add_modelopt_args(run_parser)
     args = cli.parse_args()
     _inject_modelopt_args_into_model_args(args)
+    lower_bound = vars(args).pop("accuracy_lower_bound", None)
+    output_path = getattr(args, "output_path", None)
+    gate_task = None
+    if lower_bound is not None:  # fail fast before the (expensive) eval
+        if not output_path:
+            raise ValueError("--accuracy_lower_bound requires --output_path.")
+        tasks = args.tasks if isinstance(args.tasks, list) else str(args.tasks or "").split(",")
```
cjluo-nv: > _Bot comment._

Small regression from inlining `_requested_tasks`: when `--tasks` is omitted, `str(None or "").split(",")` yields `[""]`, which passes the `len(tasks) != 1` check and sets `gate_task = ""`. The gate then silently gets an empty task name and only fails later (post-eval `KeyError`, or an opaque lm-eval error). The old helper filtered/stripped. A one-liner keeps the simplification and the behavior:

```python
raw = args.tasks if isinstance(args.tasks, list) else str(args.tasks or "").split(",")
tasks = [t.strip() for t in raw if t and t.strip()]
```

tools/launcher/core.py
```diff
@@ -803,7 +833,37 @@ def run_jobs(
...
+                    marker = ".modelopt_launcher_reqs_done"
```
cjluo-nv: > _Bot comment._

The comment claims the marker is "fresh per task and shared across the node's ranks", but neither half holds on a multi-node Slurm task: the CWD (`/nemo_run/code`) is the packaged code dir under the *shared* job dir, so node B's non-zero local ranks can observe node A's marker and skip the wait before node B's local rank 0 has installed anything. The `rm -f` also opens its own (small) window — a rank that tests `[ -f {marker} ]` just before rank 0 removes it proceeds on a previous task's marker.

Folding `${SLURM_NODEID:-0}` (and `${SLURM_JOB_ID:-0}_${SLURM_STEP_ID:-0}`) into the marker name makes it genuinely per-node/per-step and removes the need for `rm -f` entirely. If multi-node `reqs` isn't a supported case, please say so in `docs/configuration.md` and adjust this comment.

tools/launcher/core.py
```diff
@@ -803,7 +833,37 @@ def run_jobs(
...
+                    reqs_prefix = (
+                        f'if [ "${{SLURM_LOCALID:-0}}" -eq 0 ]; then rm -f {marker}; '
+                        f"{install} && touch {marker}; "
+                        f"else for _ in $(seq 600); do [ -f {marker} ] && break; sleep 1; done; "
+                        f"[ -f {marker} ]; fi && "
+                    )
+                if task.inline:
+                    task_instance = run.Script(inline=reqs_prefix + task.inline, env=task_env)
```
cjluo-nv, line 864: > _Bot comment._

This `reqs_prefix` string (quote handling + barrier shell + the `reqs`+`script` inline wrap) is the most intricate new code in the PR and has no test. `tests/test_core_extended.py::TestRunJobsExtended` already patches `core.run.Script` and asserts on `call_args[1]`, so a test that builds a `SandboxTask0(reqs="transformers<5", inline="python x.py")` and asserts the generated `inline=` string (quoting of `transformers<5`, `SLURM_LOCALID` branch, marker check) is ~10 lines and would pin both the quoting and the barrier shape against future edits.

tests/examples/megatron_bridge/test_prune_minitron.py
cjluo-nv: > _Bot comment._

Nit (unchanged from last round): `score_lower_bound=0.01` couples this GPU test's pass/fail to a tiny random-init model's MMLU score. Risk is low (4-choice floor ~0.25), but it adds a new flake surface to a test whose purpose is the prune path, not accuracy. If the intent is only to exercise the gate code path, `0.0` would do that with zero accuracy coupling.

cjluo-nv (commented):

**Answering the operator's question first: yes — `examples/llm_eval/lm_eval_hf.py` at head now carries the simplified implementation.** All three reductions from the last round landed: `_pop_gate_bound` is inlined to `vars(args).pop("accuracy_lower_bound", None)`, `_requested_tasks` collapsed to a one-line `isinstance` check, the non-HF quantization/sparsity validation is hoisted above the injection loop (reports all offending flags at once, no `quant_request_keys`/4-line comment), and the duplicate `--output_path` check inside `_enforce_accuracy_gate` is gone (the fail-fast one before `cli.execute` remains). The author also verified `HarnessCLI.execute()` is annotated `-> None` and left a one-line comment justifying the disk read — that's the right resolution of the "read results in-process" suggestion (not verifiable from this repo, but the comment documents the constraint). The gate is now ~50 lines instead of ~70.

Also confirmed resolved since earlier rounds: non-HF backends raise on `quant_cfg`/`auto_quantize_bits`/`compress`/`sparse_cfg` (and only on those); `is_hf` covers `hf`/`hf-auto`/`huggingface`/`hf-multimodal`; results file chosen by `mtime`; `script`/`inline` exclusivity **and** `args`+`inline` rejected in `run_jobs`; script path `shlex.quote`d in the reqs+script branch; reqs barrier `rm -f`s the marker and waiting ranks fail via `[ -f {marker} ]`; `TestGetDefaultEnv` env isolation; `docs/configuration.md` documents `inline`/`reqs`/`reqs_file` and says `args` is rejected. No licensing changes and no prompt-injection attempts in the PR content.

**Still open (why this isn't an approve):**

- :speech_balloon: Design rationale — asked in the two previous rounds, still unchanged. The PR body describes `inline`/`reqs` but doesn't say why the existing `common/**/*.sh` wrapper convention (which `common/megatron_bridge/import/import.sh` and `common/megatron_lm/quantize/quantize.sh` already demonstrate for this exact flow, including a bash pip install + barrier) or typed `_target_` tasks were rejected. I re-read `tools/launcher/CLAUDE.md` at head: it is untouched and its "Key conventions" still say only "Scripts go in `common/`", and "Adding a New Model Config" still says "Set `MLM_MODEL_CFG`" / "Set `QUANT_CFG`" — precisely the rule the new wrapper-free examples intentionally don't follow (that mismatch is what generated the `MLM_MODEL_CFG` findings that had to be withdrawn twice). Contributors now have three ways to express a task command with no guidance. One or two sentences in the PR body + one line in `CLAUDE.md` closes this.
- :speech_balloon: Author declined tests as "a one-line guard" — agreed for the guard itself, but the `reqs_prefix` construction (shlex quoting of `reqs`/`reqs_file`, the rank-0 barrier shell string, and the `reqs` + `script` → `run.Script(inline=...)` branch) is still exercised by nothing. I checked `tools/launcher/tests/test_core_extended.py::TestRunJobsExtended`: it already patches `core.run.Script` and asserts on call kwargs, so pinning the generated `inline=` string for a `reqs`+`inline` task is ~10 lines. The new `test_yaml_formats.py` tests only cover dataclass field resolution.
- :speech_balloon: Author accepted the strict-checkpoint-key break from adding `"best"` to `default_state_dict` (pruning is one-shot). Reasonable, but resuming a pre-upgrade `modelopt_pruning_scores` / `--prune_intermediate_ckpt` dir now dies on `assert state_dict.keys() == self.state_dict().keys(), "Keys in checkpoint don't match!"` with no hint to delete the dir — worth a conscious owner yes/no since the mitigation is one line.
- New nit introduced by the simplification, plus the multi-node barrier nit and the `score_lower_bound=0.01` flake surface — see inline comments.

examples/llm_eval/lm_eval_hf.py
kevalmorabia97: Good catch — fixed. Now filters empty/whitespace task names: `tasks = [t.strip() for t in raw if t and t.strip()]`, so an omitted `--tasks` yields `[]` and fails the `len != 1` check up front instead of gating on an empty task name.

tools/launcher/core.py
kevalmorabia97: Fixed both. The marker name now folds in `${SLURM_JOB_ID:-0}_${SLURM_STEP_ID:-0}_${SLURM_NODEID:-0}`: per-node stops another node's rank 0 from satisfying this node's barrier, and per-step means a stale marker from an earlier task in the same code dir can't be observed — so `rm -f` is gone entirely (removing that window too). Updated the comment and added a note in configuration.md that multi-node reqs installs are supported.

tools/launcher/core.py
kevalmorabia97: Added two tests in `TestRunJobsExtended` (`test_reqs_inline_barrier`, `test_reqs_script_wrapped_inline`) that patch `core.run.Script` and assert the generated `inline=` string: shlex quoting (`'transformers<5'`), the `SLURM_LOCALID` branch, the per-job/step/node marker shape, absence of `rm -f`, and the `bash <script> <args>` wrap for the script case.

tests/examples/megatron_bridge/test_prune_minitron.py
kevalmorabia97: Done — changed both call sites to `score_lower_bound=0.0`, which exercises the gate path without coupling the test to the random-init model's accuracy.

kevalmorabia97:
## Addressed @cjluo-nv's review feedback

Two rounds of comments handled — simplifications first, then the follow-up correctness/test nits. Summary of the changes (all in the latest commits):

### `examples/llm_eval/lm_eval_hf.py`
- **Hoisted the non-HF validation** out of the injection loop into a single `if not is_hf:` block that reports all offending flags at once; the loop now just pops + injects for the HF backend (dropped the `quant_request_keys` set and its explanatory comment).
- **Inlined `_pop_gate_bound` and `_requested_tasks`** to one-liners at the call sites and removed the duplicate `--output_path` check (kept the fail-fast one).
- **Fixed the empty-task regression** from that inlining: `--tasks` parsing now filters empty/whitespace names (`[t.strip() for t in raw if t and t.strip()]`), so an omitted `--tasks` fails fast instead of silently gating on `""`.
- Confirmed `HarnessCLI.execute()` returns `None` (results are consumed inside `Run._execute()` and never propagated), so the disk read stays — added a one-line comment saying so.

### `tools/launcher/core.py` + `docs/configuration.md`
- **Made the pip-install barrier marker unique per job/step/node** (`${SLURM_JOB_ID:-0}_${SLURM_STEP_ID:-0}_${SLURM_NODEID:-0}`): per-node stops another node's rank 0 from satisfying this node's barrier, and per-step means a stale marker from an earlier task in the same shared code dir can't short-circuit the wait. This let me **drop `rm -f` entirely**, closing its own race window. Updated the comment and documented multi-node `reqs` support in `configuration.md`.

### Tests
- **`tools/launcher/tests/test_core_extended.py`**: added `test_reqs_inline_barrier` and `test_reqs_script_wrapped_inline` in `TestRunJobsExtended`, asserting the generated `inline=` string — shlex quoting (`'transformers<5'`), the `SLURM_LOCALID` branch, the marker shape, absence of `rm -f`, and the `bash <script> <args>` wrap.
- **`tests/examples/megatron_bridge/test_prune_minitron.py`**: changed both call sites to `score_lower_bound=0.0` to exercise the gate path without coupling the test to a random-init model's accuracy.

cjluo-nv (commented):

Re-review at head (post-`cd9eff8`/latest). Verified against the files in the repo: this round closed the last substantive engineering gaps, so what's left is owner judgment plus one docs/design ask that has now gone three rounds without a change. No licensing changes; no prompt-injection attempts in the PR content.

**Confirmed resolved this round (checked in-tree, not just from the diff):**
- `tools/launcher/tests/test_core_extended.py` now has `test_reqs_inline_barrier` / `test_reqs_script_wrapped_inline` in `TestRunJobsExtended`, pinning the generated `run.Script(inline=...)` string: `shlex` quoting (`'transformers<5'`), the `SLURM_LOCALID` branch, the per-job/step/node marker, absence of `rm -f`, and the `bash <script> <args>` wrap. This was the main outstanding test gap.
- Barrier marker now folds `${SLURM_JOB_ID:-0}_${SLURM_STEP_ID:-0}_${SLURM_NODEID:-0}`, so it's genuinely per-node/per-step and `rm -f` (and its race window) is gone; the shell reads correctly — rank 0's failed install short-circuits `&&` and the compound `if` returns non-zero, and waiting ranks fail on the trailing `[ -f {marker} ]`. `docs/configuration.md` documents multi-node `reqs`.
- `lm_eval_hf.py` gate is the simplified form: inlined bound pop and task parsing (with the empty-`--tasks` regression fixed via `[t.strip() for t in raw if t and t.strip()]`), hoisted non-HF quant/sparsity validation reporting all offending flags, single fail-fast `--output_path` check, `mtime`-selected results file, `is_hf` covering all four HFLM aliases, and a comment justifying the disk read since `HarnessCLI.execute() -> None`.
- `score_lower_bound` in `test_prune_minitron.py` is now `0.0` (no accuracy coupling); `TestGetDefaultEnv` env isolation added; `script`/`inline` exclusivity and `args`+`inline` rejected in `run_jobs` (typed `_target_` tasks set `self.script` in `__post_init__`, so the new exactly-one check doesn't regress them).

**Why this is a nudge rather than an approve:**
- :speech_balloon: Design rationale (raised in three prior rounds; author's reply so far is the inline "`MLM_MODEL_CFG`/`QUANT_CFG` are wrapper-only, N/A here"). That explains why the env vars are absent, not why the wrapper-free route was chosen: the PR body still doesn't compare `inline`/`reqs` against the existing `common/**/*.sh` convention (`common/megatron_bridge/import/import.sh`, `common/megatron_lm/quantize/quantize.sh` already cover this flow and could do the pip install + bash barrier) or typed `_target_` tasks. I re-read `tools/launcher/CLAUDE.md` at head: untouched, "Key conventions" still says only "Scripts go in `common/`" and "Adding a New Model Config" still says set `MLM_MODEL_CFG`/`QUANT_CFG` — exactly the rule these examples intentionally don't follow (the mismatch is what produced the two withdrawn `MLM_MODEL_CFG` findings). `docs/architecture.md`/`README.md` still present typed tasks as the structured path. Contributors now have three ways to express a task command with no guidance; one or two sentences in the PR body plus a line in `CLAUDE.md` closes this.
- :speech_balloon: Author consciously accepted the strict-checkpoint-key break from adding `"best"` to `MCoreMinitronSearcher.default_state_dict` (pruning is one-shot). Reasonable, and I confirmed `self.best = asdict(best)` keeps `BaseSearcher.search() -> SearchStateDict` intact and is re-assigned on resume (the candidate loop still runs with restored scores), so the gate works after a resume. But resuming a pre-upgrade `modelopt_pruning_scores` / `--prune_intermediate_ckpt` dir still dies on `assert state_dict.keys() == self.state_dict().keys(), "Keys in checkpoint don't match!"` with no hint to delete the dir — worth an explicit owner yes/no since the mitigation is one line. Also note `save_search_checkpoint()` is never called after the `self.best` assignment, so an on-disk completed checkpoint records `"best": {}`; harmless for the current gate (which reads the in-memory return), but a latent surprise for anyone reading the saved scores file.
- Minor: the negative-path gates (`--accuracy_lower_bound` / `--score_lower_bound` above the score) remain untested; the author's argument that the passing path plus a trivial `sys.exit(1)` suffices is defensible, and CodeRabbit withdrew it — flagging only so the owner can agree consciously.

cjluo-nv (approved):
approve lm_eval_hf.py

PR #2032, opened 2026-07-29
fix(export): [NVBug 6525534] preserve nested VLM namespaces

meenchen:
### What does this PR do?

Type of change: Bug fix

Prevents recursively collected text-submodel reverse mappings from rewriting an already nested multimodal model namespace during unified Hugging Face export.

Transformers reverses the Qwen3.5 text mapping into a broad `^model.` -> `model.language_model.` rename. ModelOpt previously applied that rule to every key in the full VLM state dict, moving `model.visual.*` under the language model and nesting `model.language_model.*` twice. This change drops the reverse rule only when its target child namespace is already registered. Standalone text models continue to use the conversion.

### Usage

```python
# Existing export_hf_checkpoint usage is unchanged.
```

### Testing

- `python -m pytest -q tests/unit/torch/export` (`112 passed`, `1 skipped` because optional Diffusers is not installed)
- Targeted pre-fix reproduction confirmed both malformed Qwen3.5 namespaces; both regression cases pass after the fix
- Tiny `Qwen3_5MoeForConditionalGeneration` meta-device model-tree probe passed
- `python -m pre_commit run --files modelopt/torch/export/quant_aware_conversion.py tests/unit/torch/export/test_quant_aware_conversion.py`

### Additional Information

No API or dependency changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model conversion to prevent incorrect renaming of nested visual or sibling model components.
  * Fixed duplicate namespace prefixes in converted model weights.
  * Preserved correct reverse mapping for text-only model configurations.

* **Tests**
  * Added coverage for nested multimodal and text-only model conversion scenarios.


cjluo-nv (approved):

The fix is focused and appears correct. It detects broad reverse prefix renames whose destination is already an explicitly registered child namespace, preventing Qwen3.5 VLM sibling capture and duplicate `language_model` nesting while retaining the rename for standalone text models. The two regression tests cover both sides of that behavior, and the changelog is updated. No licensing text is changed.

meenchen:
/ok to test [362c1e4](https://github.com/NVIDIA/Model-Optimizer/pull/2032/commits/362c1e48ed224a68baa909a0f0979412d0cbfc5c)

PR #2041, opened 2026-07-31
[NVBug: 6524370] use sequential device_map for DiffusionGemma

juhi10071998:
### What does this PR do?

Type of change: Bug fix

Fixes NVBug 6524370.

`DiffusionGemma` ties weights between its encoder and decoder. `get_model` loads with `device_map="auto"` (`examples/hf_ptq/example_utils.py`), and `"auto"` is an alias for `"balanced"` — accelerate splits the model evenly across all visible GPUs by size, with no awareness of tied parameters. On multi-GPU it can place the two sides of a tied pair on different devices; the tie then cannot be honored and one side is left on the `meta` device.

The pre-quantization preview in `pre_quantize` then reaches `(input_ids == self.config.image_token_id).any()` in `generation_diffusion_gemma.py` and fails:

```
RuntimeError: Tensor.item() cannot be called on meta tensors
```

This is multi-GPU-only by construction: with one visible GPU the balanced split is trivial, nothing is separated, and nothing lands on `meta`.

This PR detects DiffusionGemma configs in `get_model` and selects `device_map="sequential"`, which fills one GPU before spilling to the next and so keeps tied modules together. It mirrors the existing per-model handling for `bart` and `t5`, where `device_map="auto"` similarly mis-shards tied encoder/decoder weights.

Detection reads `model_type` and `architectures` from the config and ignores underscores, since the family is spelled `diffusion_gemma` in the Transformers module path and `DiffusionGemma` in the class name.

### Usage

No API change. Previously this needed the flag passed manually:

```bash
python hf_ptq.py --model <diffusion-gemma-ckpt> --recipe <recipe> \
  --export_path <out> --trust_remote_code --use_seq_device_map
```

It is now selected automatically, and the model load logs:

```
Detected DiffusionGemma model. Using device_map='sequential'; the balanced
'auto' mapping can split its tied encoder/decoder weights across GPUs.
```

Passing `--use_seq_device_map` explicitly still works and is unaffected.

### Testing

- Reproduced on 4x GB200 with `diffusiongemma-26B-A4B-it` and the `nvfp4_experts_only` recipe; `--use_seq_device_map` resolves the crash, confirming the device-mapping cause.
- Validated on oci-hsg (4x GB200): with this patch and no CLI flag, `diffusiongemma-26B-A4B-it` loads correctly and the meta-tensor crash no longer reproduces.
- `is_diffusion_gemma` checked against both config spellings, `architectures=None`, `architectures=[]`, and a `gemma3` negative to confirm no over-match — `get_model_type` already orders `DiffusionGemma` before `Gemma` for exactly this substring-collision reason.
- `pre-commit run --files examples/hf_ptq/example_utils.py` passes (ruff, ruff-format, mypy, bandit).

- Did you write any new necessary tests?: ❌ — no existing unit coverage for `get_model` device-map selection; happy to add a config-level test for `is_diffusion_gemma` if wanted.
- Did you get Claude approval on this PR?: ❌ — pending

### Additional Information

NVBug 6524370. Same class of failure as the existing `t5` workaround in `get_model`; a general "any tied encoder/decoder model" rule was considered but rejected, since `tie_word_embeddings=True` holds for most decoder-only LLMs where `auto` is fine and forcing sequential would regress large-model runs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DiffusionGemma model loading on multi-GPU systems by keeping related model weights together.
  * Added more reliable DiffusionGemma model recognition across supported configurations.
  * Preserved existing automatic device allocation for single-GPU systems and other supported models.
  * Improved loading reliability by applying appropriate memory limits during multi-GPU setup.

### DIFF
--- a/examples/hf_ptq/example_utils.py
+++ b/examples/hf_ptq/example_utils.py
@@ -304,6 +304,20 @@ def is_speculative(hf_config):
     )
 
 
+def is_diffusion_gemma(hf_config) -> bool:
+    """Check if the model architecture is DiffusionGemma.
+
+    Underscores are ignored: the family is spelled ``diffusion_gemma`` in configs
+    and ``DiffusionGemma`` in class names. The nested ``text_config`` is checked too,
+    since multi-modal wrappers keep the family name there.
+    """
+    names = []
+    for cfg in (hf_config, getattr(hf_config, "text_config", None)):
+        names.append(getattr(cfg, "model_type", None) or "")
+        names.extend(getattr(cfg, "architectures", None) or [])
+    return any("diffusiongemma" in name.lower().replace("_", "") for name in names)
+
+
 def get_tokenizer(ckpt_path, trust_remote_code=False, **kwargs) -> PreTrainedTokenizerBase:
     print(f"Initializing tokenizer from {ckpt_path}")
 
@@ -695,6 +709,19 @@ def get_model(
     # Note: Forcibly converting the model precision between bf16 and fp16 may introduce accuracy drop
     model_kwargs = config_kwargs.copy()
     model_kwargs.setdefault("dtype", "auto")
+
+    # DiffusionGemma ties encoder/decoder weights. device_map "auto" (balanced) can split
+    # a tied pair across GPUs, leaving one side on the meta device and breaking generation.
+    # Sequential packs the model onto GPU 0 first (up to gpu_mem_percentage), keeping tied
+    # modules together for checkpoints that fit; larger ones can still spill and split a
+    # tied pair, and need an explicit single-device map. Multi-GPU only: a single-GPU split
+    # cannot separate a tied pair, and sequential would needlessly cap max_memory there.
+    if device != "cpu" and torch.cuda.device_count() > 1 and is_diffusion_gemma(hf_config):
+        print(
+            "Detected DiffusionGemma model. Using device_map='sequential'; the balanced "
+            "'auto' mapping can split its tied encoder/decoder weights across GPUs."
+        )
+        use_seq_device_map = True
 
     if use_seq_device_map:
         device_map = "sequential"
--- a/tests/examples/hf_ptq/test_example_utils.py
+++ b/tests/examples/hf_ptq/test_example_utils.py
@@ -316,3 +316,106 @@ def test_get_model_uses_expected_dtype_kwarg(
     else:
         assert "trust_remote_code" not in calls["from_config"]
     assert calls["from_pretrained"]["trust_remote_code"] is True
+
+
+@pytest.mark.parametrize(
+    ("model_type", "architecture", "device_count", "expected_device_map"),
+    [
+        # DiffusionGemma ties encoder/decoder weights; "auto" can split a tied pair
+        # across GPUs, so multi-GPU loads must fall back to "sequential".
+        ("diffusion_gemma", "DiffusionGemmaForConditionalGeneration", 2, "sequential"),
+        # Detection must also work off ``architectures`` alone, without ``model_type``.
+        (None, "DiffusionGemmaForConditionalGeneration", 2, "sequential"),
+        # Single GPU cannot split a tied pair, so it keeps the unrestricted "auto" map.
+        ("diffusion_gemma", "DiffusionGemmaForConditionalGeneration", 1, "auto"),
+        # "gemma" is a substring of "diffusiongemma"; other Gemmas must not match.
+        ("gemma3", "Gemma3ForCausalLM", 2, "auto"),
+    ],
+)
+def test_get_model_device_map_for_diffusion_gemma(
+    monkeypatch, model_type, architecture, device_count, expected_device_map
+):
+    calls = {}
+    hf_config = SimpleNamespace(
+        architectures=[architecture],
+        dtype=torch.float16,
+        model_type=model_type,
+        torch_dtype=torch.bfloat16,
+    )
+
+    class FakeModel:
+        def eval(self):
+            calls["eval"] = True
+
+        def parameters(self):
+            return iter(())
+
+    class FakeArchitecture:
+        @staticmethod
+        def _from_config(config, **kwargs):
+            return FakeModel()
+
+        @staticmethod
+        def from_pretrained(*args, **kwargs):
+            calls["from_pretrained"] = kwargs
+            return FakeModel()
+
+    monkeypatch.setattr(
+        example_utils.AutoConfig, "from_pretrained", lambda *args, **kwargs: hf_config
+    )
+    # Set rather than delete: ``transformers`` lazy-imports, so a deleted real class
+    # (e.g. Gemma3ForCausalLM) reappears on the next ``hasattr`` and the real one loads.
+    # raising=False: DiffusionGemma may not exist in the installed transformers.
+    monkeypatch.setattr(example_utils.transformers, architecture, FakeArchitecture, raising=False)
+    monkeypatch.setattr(example_utils, "is_nemotron_vl", lambda config: False)
+    monkeypatch.setattr(example_utils, "is_speculative", lambda config: False)
+    monkeypatch.setattr(example_utils, "init_empty_weights", lambda include_buffers: nullcontext())
+    monkeypatch.setattr(example_utils, "get_max_memory", lambda: {0: 1024})
+    monkeypatch.setattr(example_utils, "infer_auto_device_map", lambda model, max_memory: {"": 0})
+    monkeypatch.setattr(torch.cuda, "device_count", lambda: device_count)
+
+    example_utils.get_model("checkpoint", device="cuda", trust_remote_code=True)
+
+    assert calls["from_pretrained"]["device_map"] == expected_device_map
+    # Sequential caps per-GPU memory; "auto" must stay unrestricted.
+    if expected_device_map == "sequential":
+        assert calls["from_pretrained"]["max_memory"] == {0: 1024 * 0.8}
+    else:
+        assert "max_memory" not in calls["from_pretrained"]
+
+
+@pytest.mark.parametrize(
+    ("hf_config", "expected"),
+    [
+        (SimpleNamespace(model_type="diffusion_gemma", architectures=None), True),
+        (SimpleNamespace(model_type=None, architectures=["DiffusionGemmaForCausalLM"]), True),
+        (SimpleNamespace(model_type="gemma3", architectures=["Gemma3ForCausalLM"]), False),
+        # Multi-modal wrappers keep the family name on the nested ``text_config``.
+        (
+            SimpleNamespace(
+                model_type="multimodal",
+                architectures=["SomeWrapperForConditionalGeneration"],
+                text_config=SimpleNamespace(model_type="diffusion_gemma"),
+            ),
+            True,
+        ),
+        (
+            SimpleNamespace(
+                model_type="multimodal",
+                text_config=SimpleNamespace(architectures=["DiffusionGemmaForCausalLM"]),
+            ),
+            True,
+        ),
+        # A non-DiffusionGemma nested config must not match.
+        (
+            SimpleNamespace(
+                model_type="multimodal", text_config=SimpleNamespace(model_type="gemma3")
+            ),
+            False,
+        ),
+        # Stub configs may omit either attribute entirely.
+        (SimpleNamespace(), False),
+    ],
+)
+def test_is_diffusion_gemma(hf_config, expected):
+    assert example_utils.is_diffusion_gemma(hf_config) is expected


examples/hf_ptq/example_utils.py
```diff
@@ -696,6 +707,18 @@ def get_model(
...
+    # (balanced) can place the two sides of a tied pair on different GPUs; the tie then
+    # cannot be honored and one side is left on the meta device, so generation dies with
+    # "Tensor.item() cannot be called on meta tensors". Sequential mapping keeps tied
+    # modules together. Same class of failure as the T5 case handled below.
+    if device != "cpu" and is_diffusion_gemma(hf_config):
+        print(
+            "Detected DiffusionGemma model. Using device_map='sequential'; the balanced "
+            "'auto' mapping can split its tied encoder/decoder weights across GPUs."
+        )
+        use_seq_device_map = True
```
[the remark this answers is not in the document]
juhi10071998: Fixed in `fafe1dbe` — added `torch.cuda.device_count() > 1` to the condition, so automatic sequential mapping is now multi-GPU only. Single-GPU runs keep `device_map="auto"` with no `max_memory` cap, as before. Good catch: the tie can only be split when there is more than one device, so the single-GPU path never needed this and was losing 20% of usable memory for nothing.

examples/hf_ptq/example_utils.py
```diff
@@ -304,6 +304,16 @@ def is_speculative(hf_config):
     )
 
 
+def is_diffusion_gemma(hf_config) -> bool:
+    """Check if the model architecture is DiffusionGemma.
+
+    Underscores are ignored: the family is spelled ``diffusion_gemma`` in configs
+    and ``DiffusionGemma`` in class names.
+    """
+    names = [getattr(hf_config, "model_type", None) or "", *(hf_config.architectures or [])]
```

Two small things:

- `hf_config.architectures` is accessed directly here, while the rest of this file consistently uses `getattr(config, "architectures", [])` (see `_is_multimodal_config`, `is_nemotron_vl`). Configs built in tests / non-`PretrainedConfig` stubs won't have the attribute and will raise `AttributeError`. Suggest `*(getattr(hf_config, "architectures", None) or [])`.
- DiffusionGemma detection now exists in three places: `MODEL_NAME_TO_TYPE` (`modelopt/torch/export/model_utils.py`), `_reorder_canonical_first` in the same file (which does the same `"diffusiongemma"` / `"diffusion_gemma"` normalization on the class name), and this helper. Consider putting the config-level predicate next to the others in `model_utils.py` so future spellings only need updating once.

examples/hf_ptq/example_utils.py
```diff
@@ -696,6 +706,17 @@ def get_model(
     model_kwargs = config_kwargs.copy()
     model_kwargs.setdefault("dtype", "auto")
 
+    # DiffusionGemma ties encoder/decoder weights. device_map "auto" (balanced) can split
+    # a tied pair across GPUs, leaving one side on the meta device and breaking generation.
+    # Sequential mapping keeps tied modules together. Multi-GPU only: a single-GPU split
```

"Sequential mapping keeps tied modules together" holds only when the model fits on the first GPU. Because setting `use_seq_device_map = True` also caps `max_memory` to `gpu_mem_percentage` (0.8 by default), a checkpoint larger than 0.8 x one GPU will still spill to the next device and can still land one side of a tied pair on `meta` — i.e. the same crash, just harder to hit. The `t5`/`bart` precedent referenced in the PR body actually sets `device_map = None` (whole model on one device), which is the stronger guarantee. Please either soften this comment to state the limitation, or explain why sequential is always sufficient for this family (e.g. size bound of the released checkpoints).

examples/hf_ptq/example_utils.py
```diff
@@ -696,6 +706,17 @@ def get_model(
...
+    # cannot separate a tied pair, and sequential would needlessly cap max_memory there.
+    if device != "cpu" and torch.cuda.device_count() > 1 and is_diffusion_gemma(hf_config):
+        print(
+            "Detected DiffusionGemma model. Using device_map='sequential'; the balanced "
```
cjluo-nv, line 721: > _Bot comment._

This device-map selection is testable with the existing harness: `tests/examples/hf_ptq/test_example_utils.py::test_get_model_uses_expected_dtype_kwarg` already stubs `AutoConfig.from_pretrained`, `init_empty_weights`, `get_max_memory`, and `infer_auto_device_map` and asserts on the kwargs passed to `from_pretrained`. Adding a case that (a) patches `torch.cuda.device_count` to 2 with a `model_type="diffusion_gemma"` config and asserts `device_map == "sequential"` + `max_memory` present, and (b) a `gemma3` negative asserting `device_map == "auto"`, would pin both the fix and the substring-collision concern described in the PR body (currently only verified manually).

cjluo-nv (commented):

Small, well-motivated fix: auto-select `device_map="sequential"` for DiffusionGemma in `examples/hf_ptq/example_utils.get_model`, mirroring the existing per-model `bart`/`t5` handling. Placement (before the `use_seq_device_map` block, so `max_memory` capping still applies) is correct, and the multi-GPU + non-CPU guard is right. Three non-blocking points:

1. **No test, although the harness already exists.** The PR body says "no existing unit coverage for `get_model` device-map selection", but `tests/examples/hf_ptq/test_example_utils.py::test_get_model_uses_expected_dtype_kwarg` already monkeypatches `AutoConfig.from_pretrained`, `init_empty_weights`, `get_max_memory`, and `infer_auto_device_map` and asserts on the `from_pretrained` kwargs. A case with a `diffusion_gemma` config + patched `torch.cuda.device_count()` asserting `device_map == "sequential"` (plus a `gemma3` negative for the substring-collision concern the PR body describes as manually verified) is a handful of lines there. Worth adding since the whole fix is a detection heuristic.

2. **The comment overstates the guarantee.** `sequential` only keeps a tied pair together when the model fits on the first GPU; because `use_seq_device_map` also caps `max_memory` to `gpu_mem_percentage` (0.8), a model that doesn't fit under that cap will still spill and can still split the tied pair onto `meta`. The `t5`/`bart` precedent it cites actually uses `device_map=None` (single device), which is the stronger guarantee. Either soften the comment or explain why sequential is sufficient for this family.

3. **Minor robustness/dedup**: `hf_config.architectures` is accessed directly while the rest of this file uses `getattr(config, "architectures", [])`; and DiffusionGemma detection now lives in three places (`MODEL_NAME_TO_TYPE`, `_reorder_canonical_first`, and this new helper).

No licensing changes, no prompt-injection content in the PR text.

examples/hf_ptq/example_utils.py
juhi10071998: Done in `b45f9c36` — switched to `*(getattr(hf_config, "architectures", None) or [])`.

Worth noting the file is split rather than consistent here: `_is_multimodal_config` (179) and `is_nemotron_vl` (204) use `getattr`, but `is_speculative` (302) and `get_model` (792-794) access `.architectures` directly. Taking the defensive form anyway.

On consolidating into `model_utils.py`: agreed there is drift risk, though the two existing spots key off the *class name* (`type(model).__name__`) while this one keys off the *config*, so it is not a pure duplicate. Leaving it here to keep this an examples-only bugfix — happy to do the move as a follow-up if you would rather have it now.

examples/hf_ptq/example_utils.py
juhi10071998: Fair — the comment claimed more than the code delivers. Softened in `b45f9c36`:

```
# Sequential packs the model onto GPU 0 first (up to gpu_mem_percentage), keeping tied
# modules together for checkpoints that fit; larger ones can still spill and split a
# tied pair, and need an explicit single-device map.
```

You are right that this is a mitigation, not a guarantee, and that `t5`/`bart` use the stronger `device_map = None`. I did not follow them there because `None` puts the whole model on one device with no headroom cap, which would regress large checkpoints that legitimately need the split. For the reported case (26B-A4B, ~52 GB vs 0.8 x 186 GB on GB200) sequential is comfortably sufficient; beyond that bound the limitation is now stated rather than implied.

examples/hf_ptq/example_utils.py
juhi10071998: Added in `b45f9c36`, using that harness:

- `test_get_model_device_map_for_diffusion_gemma` — `diffusion_gemma` @ 2 GPUs asserts `device_map == "sequential"` and `max_memory == {0: 819.2}`; @ 1 GPU asserts `"auto"` with no `max_memory` (the single-GPU regression); `gemma3` @ 2 GPUs asserts `"auto"` for the substring collision.
- `test_is_diffusion_gemma` — both spellings, plus `SimpleNamespace()` with neither attribute.

Verified against the pre-fix file: the multi-GPU `sequential` case fails, the two `auto` cases pass (they are guards for unchanged behavior). 22 passed on the file.

One gotcha for future cases here: `monkeypatch.delattr(transformers, arch)` does not stick for classes that exist, since `transformers` lazy-imports and the real class reappears on the next `hasattr`. Used `setattr(..., raising=False)` instead.

cjluo-nv (commented):

Re-review of #2041 (auto `device_map="sequential"` for DiffusionGemma in `examples/hf_ptq/example_utils.get_model`). The head commit resolves the substantive items from the previous round; one flagged-Major item is labeled "addressed" but is not present in the code, so a human should confirm scope.

Addressed since last round (verified against the file at head):
- Missing test → `tests/examples/hf_ptq/test_example_utils.py` now has `test_get_model_device_map_for_diffusion_gemma` (multi-GPU → `sequential` + `max_memory == {0: 819.2}`; single-GPU → `auto` with no `max_memory`; `gemma3` negative → `auto`) and `test_is_diffusion_gemma` (both spellings, missing attrs). I traced `get_model` with these stubs — the patched `AutoConfig`/`init_empty_weights`/`get_max_memory`/`infer_auto_device_map`/`torch.cuda.device_count` path reaches `from_pretrained` and `is_model_on_gpu` cleanly (empty `parameters()`), so the tests exercise the real selection logic rather than a stub.
- Overstated comment → softened to state the "checkpoints that fit" limitation and the need for an explicit single-device map beyond it.
- `hf_config.architectures` → now `*(getattr(hf_config, "architectures", None) or [])`.
- Single-GPU regression → gated on `torch.cuda.device_count() > 1`, so single-GPU keeps unrestricted `auto` (pinned by a test case).

Remaining for owner judgment:
- :speech_balloon: CodeRabbit's Major finding on the `has_pack_quantized_config(hf_config)` branch is annotated "✅ Addressed in commit 6ef39d9", but the code at head still hardcodes `device_map="auto"` (and drops `max_memory`) in that branch. Published DiffusionGemma compressed-tensors checkpoints exist, and such a load would bypass this fix and can still split the tied pair onto `meta` — i.e. the same crash on a different entry path. It's pre-existing code and arguably out of scope for this bugfix, but the resolution label doesn't match the diff, so please confirm intent.
- :speech_balloon: Author replied on consolidation: *the two existing spots key off `type(model).__name__` while this keys off the config, so it's not a pure duplicate; happy to move to `model_utils.py` as a follow-up* — still worth an owner call, since DiffusionGemma spelling normalization now lives in three places (`MODEL_NAME_TO_TYPE`, `_reorder_canonical_first`, `is_diffusion_gemma`) and future spellings must be updated in all of them.
- Minor, optional: CodeRabbit's last suggestion (a multi-GPU `model_type=None` + `architectures=["DiffusionGemma..."]` device-map case) is not added; `test_is_diffusion_gemma` covers architecture-only detection at the predicate level, which I consider adequate.

No licensing changes and no prompt-injection content in the PR text or comments.

tests/examples/hf_ptq/test_example_utils.py
```diff
@@ -316,3 +316,81 @@ def from_pretrained(*args, **kwargs):
...
+def test_get_model_device_map_for_diffusion_gemma(
+    monkeypatch, model_type, architecture, device_count, expected_device_map
+):
+    calls = {}
+    hf_config = SimpleNamespace(
+        architectures=[architecture],
+        dtype=torch.float16,
+        model_type=model_type,
+        torch_dtype=torch.bfloat16,
+    )
```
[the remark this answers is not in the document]
juhi10071998: Added in `6acad706` — a fourth parametrize case with `model_type=None` and `architectures=["DiffusionGemmaForConditionalGeneration"]` @ 2 GPUs, asserting `device_map == "sequential"` through `get_model`. Previously that path was only covered at the helper level by `test_is_diffusion_gemma`.

Regression check: `tests/examples/hf_ptq/` is 27 failed / 41 passed on this branch vs 27 failed / 33 passed on `origin/main` — identical failure set (pre-existing CUDA-extension and GPU-dependent example tests), +8 from the new cases, none introduced.

examples/hf_ptq/example_utils.py
```diff
@@ -696,6 +707,18 @@ def get_model(
...
+    # cannot be honored and one side is left on the meta device, so generation dies with
+    # "Tensor.item() cannot be called on meta tensors". Sequential mapping keeps tied
+    # modules together. Same class of failure as the T5 case handled below.
+    if device != "cpu" and is_diffusion_gemma(hf_config):
+        print(
+            "Detected DiffusionGemma model. Using device_map='sequential'; the balanced "
+            "'auto' mapping can split its tied encoder/decoder weights across GPUs."
+        )
+        use_seq_device_map = True
+
```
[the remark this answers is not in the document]
juhi10071998: Leaving as-is for now — out of scope for this fix.

To be clear about what remains: the `has_pack_quantized_config` branch hardcodes `device_map="auto"` and never passes `max_memory`, so it bypasses the computed `device_map` entirely. That is pre-existing and not specific to DiffusionGemma — it already ignores `--use_seq_device_map` for every pack-quantized model. The consequence here is that a pack-quantized DiffusionGemma checkpoint would still hit the meta-tensor crash, since this fix cannot reach that branch.

Worth fixing, but it changes loading behavior for all pack-quantized models, so it belongs in its own PR rather than riding along with an examples-level bugfix.

Note the "✅ Addressed in commit 6ef39d9" badge on this thread is incorrect — `6ef39d9` only tightened comments and did not touch this branch.

examples/hf_ptq/example_utils.py
juhi10071998: Confirming the owner call: consolidating the DiffusionGemma predicate into `model_utils.py` will be a follow-up PR, not this one. Keeping this change scoped to the examples-level bugfix so it can land for NVBug 6524370.

The drift risk is real and noted — `MODEL_NAME_TO_TYPE`, `_reorder_canonical_first`, and `is_diffusion_gemma` would all need updating if a new spelling appears.

examples/hf_ptq/example_utils.py
```diff
@@ -304,6 +304,19 @@ def is_speculative(hf_config):
...
+    """Check if the model architecture is DiffusionGemma.
+
+    Underscores are ignored: the family is spelled ``diffusion_gemma`` in configs
+    and ``DiffusionGemma`` in class names.
+    """
+    names = [
+        getattr(hf_config, "model_type", None) or "",
+        *(getattr(hf_config, "architectures", None) or []),
+    ]
+    return any("diffusiongemma" in name.lower().replace("_", "") for name in names)
```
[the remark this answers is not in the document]
juhi10071998: Fixed in `6c28e5b6` — `is_diffusion_gemma` now iterates `(hf_config, getattr(hf_config, "text_config", None))`, matching `get_original_hf_quant_method` (625) and `has_pack_quantized_config` (746).

Added three `test_is_diffusion_gemma` cases: nested `text_config.model_type`, nested `text_config.architectures`, and a nested `gemma3` negative. Verified against the prior helper — both nested positives fail there, the negative passes (it guards unchanged behavior).

examples/hf_ptq/example_utils.py
```diff
@@ -696,6 +709,19 @@ def get_model(
...
+    # Sequential packs the model onto GPU 0 first (up to gpu_mem_percentage), keeping tied
+    # modules together for checkpoints that fit; larger ones can still spill and split a
+    # tied pair, and need an explicit single-device map. Multi-GPU only: a single-GPU split
+    # cannot separate a tied pair, and sequential would needlessly cap max_memory there.
+    if device != "cpu" and torch.cuda.device_count() > 1 and is_diffusion_gemma(hf_config):
+        print(
+            "Detected DiffusionGemma model. Using device_map='sequential'; the balanced "
+            "'auto' mapping can split its tied encoder/decoder weights across GPUs."
+        )
+        use_seq_device_map = True
```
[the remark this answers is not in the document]
juhi10071998: Agreed on the analysis, deferring to a follow-up. The distinction you draw is the right one: `bart`/`t5`/mxfp4 special-case loads that are *broken* under the alternative, whereas `auto` here is risky rather than always fatal, so a hard override is a stronger stance than the surrounding code takes.

Keeping this PR scoped to the NVBug 6524370 fix. A tri-state flag is new CLI surface and deserves its own review; the README note belongs with it so the documented workaround matches whatever the flag ends up being. Tracking alongside the `model_utils.py` consolidation follow-up.

Noting the escape hatch that exists today: `CUDA_VISIBLE_DEVICES=0` bypasses the override entirely, since it is gated on `torch.cuda.device_count() > 1`.

cjluo-nv (approved):

Re-review of #2041 (auto `device_map="sequential"` for DiffusionGemma in `examples/hf_ptq/example_utils.get_model`). Everything substantive from the prior rounds is now resolved in-tree, and the two items left open are explicitly deferred with reasoning the author has stated on-thread.

Verified at head:
- **Nested `text_config` (previous SUGGESTION)** → `is_diffusion_gemma` now iterates `(hf_config, getattr(hf_config, "text_config", None))`, matching `get_original_hf_quant_method` (625) and `has_pack_quantized_config` (746). Three new `test_is_diffusion_gemma` cases cover nested `model_type`, nested `architectures`, and a nested `gemma3` negative. `getattr(None, ...)` on a missing `text_config` degrades to `""`/`[]` correctly, and the bare `SimpleNamespace()` case pins that.
- **Missing test (critical, prior round)** → `test_get_model_device_map_for_diffusion_gemma` now has four parametrized cases: `diffusion_gemma` @2 GPUs → `sequential` + `max_memory == {0: 819.2}`; `model_type=None` + architecture-only @2 GPUs → `sequential` (CodeRabbit's last request, addressed in `6acad706`); @1 GPU → `auto` with no `max_memory`; `gemma3` @2 GPUs → `auto`. I traced `get_model` under the stubs — `AutoConfig`/`init_empty_weights`/`get_max_memory`/`infer_auto_device_map`/`torch.cuda.device_count` all patch the real path, `infer_auto_device_map → {"": 0}` keeps `on_cpu` False so the `auto` cases legitimately assert `max_memory` absent, and the empty `parameters()` keeps `is_model_on_gpu` from blowing up. Real selection logic, not a stub.
- **Single-GPU regression + overstated comment** → still fixed (`torch.cuda.device_count() > 1` gate; comment states the "checkpoints that fit" limitation).
- **Placement** → the block sits before `if use_seq_device_map:` so the `max_memory` cap applies, and after it the `bart`/`t5` `device_map = None` overrides can't collide.

Deferred with stated rationale (accepting as reasonable, not blocking):
- `has_pack_quantized_config` still hardcodes `device_map="auto"` and drops `max_memory`, so a pack-quantized DiffusionGemma checkpoint can still hit the meta-tensor crash. Author confirmed this is pre-existing (that branch already ignores `--use_seq_device_map` for *every* pack-quantized model) and that fixing it changes loading behavior for all such models, so it belongs in its own PR. The author also correctly flagged that CodeRabbit's "✅ Addressed in commit 6ef39d9" badge on that thread is wrong — good catch, and it removes the label/diff mismatch I raised last round.
- Consolidating the DiffusionGemma predicate into `model_utils.py` and a tri-state escape-hatch flag (+ README note) are both tracked as follow-ups; `CUDA_VISIBLE_DEVICES=0` is a workable stopgap given the multi-GPU gate.

Nit, non-blocking: the test module docstring (lines 15-19) still claims the file covers only `load_mtp_weights`, though it now also covers `get_original_hf_quant_method`, `_resolve_init_config`, and two `get_model` selection paths — one-line fix if you touch the file again.

No licensing changes (existing headers only). No prompt-injection content in the PR text or comments.

PR #2042, opened 2026-07-31
[skill] evaluation: correct the stale vLLM CUDA-13 image tag rule

cjluo-nv:
### What does this PR do?

Type of change: Bug fix (agent skill documentation)

The `evaluation` skill instructed the agent to **"append `-cu130` to the image tag"** for NVFP4 checkpoints on Blackwell B300/GB300 (sm_103). That was correct for v0.19.x, but **vLLM inverted its tag convention at v0.20.0**: the *unsuffixed* tag is now the CUDA-13 build, and `-cu129` is the CUDA-12 opt-out.

Consequences of the stale rule:

- `v0.20.1-cu130` / `v0.24.0-cu130` / `v0.26.0-cu130` **do not exist** — following the rule literally asks for a missing tag.
- The documented fallback (`cu130-nightly-<arch>`) points at ~v0.20-era builds that are *older* than several models' documented minimum vLLM version, so it can't serve as an escape hatch either.

This replaces the "append a suffix" instruction with a version-keyed table plus the durable check: **select a tag whose config blob reports `CUDA_VERSION` ≥ 13**, resolving the child manifest for the platform you actually deploy on.

While the PR was open the default image was also bumped, and review surfaced two follow-on corrections. Full contents:

1. **Tag-convention fix** — version-keyed table, `-cu130` fallback removed.
2. **Default image `v0.19.1` → `v0.26.0`** (latest vLLM release) everywhere it was pinned: `SKILL.md` Step 3 and the Step 7.5 table, `example_eval.yaml`, `example_eval_next.yaml`. Version specifics that the bump made stale or self-contradictory were dropped (the `e.g. v0.20.0` bump example and the MiniMax-M2.7 `≥0.20.0` anecdote, both now *below* the default; the failure-mode lesson is kept).
3. **Convention boundary corrected to v0.20.0** — the first draft said `≤ v0.20.x` suffixed / `≥ ~v0.21` unsuffixed. Off by a minor release in both rows; see Testing.
4. **Config blob resolved per deployment platform** — the check said "arm64 child". GB300/Grace is arm64, but plenty of B300 deployments are `linux/amd64`.
5. **Same corrections applied to the `deployment` skill**, which carried the original append rule untouched: its NVFP4 note, `references/support-matrix.md`, `references/benchmarking.md`, and the `:latest` pins in `references/setup.md` (now `v0.26.0`, matching the evaluation skill's never-`:latest` stance).

An earlier revision of this branch also carried a `.claude/skills/benchmark-model-kernels` symlink, added automatically by `tools/precommit/sync_claude_skills.sh` — it repairs missing symlinks repo-wide on any touch of `.agents/skills/`, and #1980 landed that skill without its link. It has been dropped from this branch to keep the scope on the vLLM image guidance. Worth its own one-line PR: without the symlink, Claude Code doesn't load that skill at all.

### Usage

```bash
# Durable check — resolve the child manifest for YOUR platform (arm64 for
# Grace/GB300, amd64 for x86) and read CUDA_VERSION from its config blob:
#   v0.19.1            -> CUDA_VERSION=12.9.1   (unsuffixed = CUDA 12, old convention)
#   v0.19.1-cu130      -> CUDA_VERSION=13.0.1   (suffixed = CUDA 13, old convention)
#   v0.20.0            -> CUDA_VERSION=13.0.2   (transition release: ships both suffixes)
#   v0.26.0            -> CUDA_VERSION=13.0.2   (unsuffixed = CUDA 13, new convention)
#   v0.26.0-cu129      -> CUDA_VERSION=12.9.1   (suffixed = CUDA 12, new convention)
```

### Testing

Verified empirically against the Docker registry API for `vllm/vllm-openai` — resolved each tag's child manifests and read `CUDA_VERSION` / `TORCH_CUDA_ARCH_LIST` from the config blob.

**Where the convention flips (arm64):**

| release | unsuffixed | `-cu130` | `-cu129` |
|---|---|---|---|
| v0.18.0 | 12.9.1 | 13.0.1 | absent |
| v0.19.0 / v0.19.1 | 12.9.1 | 13.0.1 | absent |
| **v0.20.0** | **13.0.2** | 13.0.2 | 12.9.1 |
| v0.20.1 | 13.0.2 | **absent** | 12.9.1 |
| v0.20.2 | 13.0.2 | **absent** | 12.9.1 |
| v0.21.0 … v0.26.0 | 13.0.2 | absent | 12.9.1 |

v0.20.0 is the transition release — it publishes both suffixes *and* its unsuffixed tag is already CUDA 13. That duplication is what hid the boundary: confirming `v0.20.0-cu130` exists reads as "old convention still applies at 0.20", while `v0.20.1-cu130` and `v0.20.2-cu130` don't exist at all.

**Why the platform matters.** `CUDA_VERSION` is identical across children on every tag checked (v0.26.0, v0.26.0-cu129, v0.20.0, v0.19.1, v0.19.1-cu130, kimi-k3), but `TORCH_CUDA_ARCH_LIST` is not:

```
v0.26.0  amd64  7.5 8.0 8.6 8.9 9.0 10.0 12.0
v0.26.0  arm64  8.0 8.7 8.9 9.0 10.0 11.0 12.0
v0.19.1  amd64  7.0 7.5 8.0 8.9 9.0 10.0 12.0
v0.19.1  arm64  8.7 8.9 9.0 10.0+PTX 12.0
```

`11.0` appears only on arm64, `7.5` / `8.6` only on amd64 — so the arch check has to read the child you'll actually run.

**Default bump.** The `0.26.0` family is `{,-aarch64,-x86_64} × {,-cu129} × {,-ubuntu2404}` — 12 tags, `cu129` the only CUDA axis, no `-cu130`. The new default is therefore already a CUDA-13 build, and NVFP4 on B300/GB300 needs no suffix at all.

Docs-only change; no runtime code touched. `pre-commit` clean.

- Did you write any new necessary tests?: N/A (agent skill documentation)

### Additional Information

Split out of the GDPVal skill work (#2039) because it is independent of GDPVal and applies to every NVFP4-on-Blackwell deployment the skill generates. Now spans both the `evaluation` and `deployment` skills, all under `.agents/`.

## Summary by CodeRabbit

* **Documentation**
  * Updated vLLM image guidance to version 0.26.0.
  * Clarified CUDA image tag conventions across supported versions.
  * Added guidance for validating platform-specific CUDA versions and GPU architecture settings.
  * Updated NVFP4 deployment guidance for B300/GB300 hardware, including CUDA 13 requirements and known kernel limitations.
  * Refreshed evaluation recipes and serving-image requirements.

### DIFF
--- a/.agents/skills/deployment/SKILL.md
+++ b/.agents/skills/deployment/SKILL.md
@@ -125,12 +125,16 @@ python -m vllm.entrypoints.openai.api_server \
 
 For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 
-> **NVFP4 on Blackwell B300/GB300 (sm_103): append `-cu130` to the image tag**
-> (e.g. `vllm/vllm-openai:v0.19.1-cu130` — release tags are multi-arch). The
-> default cu12 build has **no sm_103 FP4 kernel**, so vLLM loads the checkpoint
-> then dies at engine init with `CUDA error: no kernel image is available for
-> execution on the device` (affects the `flashinfer` and `cutlass` NVFP4
-> backends; `marlin` separately fails on non-64-divisible layer dims).
+> **NVFP4 on Blackwell B300/GB300 (sm_103) needs a CUDA-13 image.** Current vLLM
+> release tags are CUDA-13 unsuffixed (e.g. `vllm/vllm-openai:v0.26.0` — release
+> tags are multi-arch), with `-cu129` as the CUDA-12 opt-out; older releases
+> spelled it the other way round (`-cu130` suffix = CUDA 13), so `-cu130` does
+> not exist on current tags. Don't trust the tag name — read `CUDA_VERSION` from
+> the tag's arm64 config blob. A cu12 build has **no sm_103 FP4 kernel**, so vLLM
+> loads the checkpoint then dies at engine init with `CUDA error: no kernel image
+> is available for execution on the device` (affects the `flashinfer` and
+> `cutlass` NVFP4 backends; `marlin` separately fails on non-64-divisible layer
+> dims).
 > Cross-check via
 > `recipes.vllm.ai/<org>/<model>?hardware=b300` (JS-rendered — fetch the raw
 > markdown at `github.com/vllm-project/recipes/blob/main/<org>/<model>.md`). For
--- a/.agents/skills/deployment/SKILL.md
+++ b/.agents/skills/deployment/SKILL.md
@@ -125,12 +125,13 @@ python -m vllm.entrypoints.openai.api_server \
 
 For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 
-> **NVFP4 on Blackwell B300/GB300 (sm_103) needs a CUDA-13 image.** Current vLLM
-> release tags are CUDA-13 unsuffixed (e.g. `vllm/vllm-openai:v0.26.0` — release
-> tags are multi-arch), with `-cu129` as the CUDA-12 opt-out; older releases
-> spelled it the other way round (`-cu130` suffix = CUDA 13), so `-cu130` does
-> not exist on current tags. Don't trust the tag name — read `CUDA_VERSION` from
-> the tag's arm64 config blob. A cu12 build has **no sm_103 FP4 kernel**, so vLLM
+> **NVFP4 on Blackwell B300/GB300 (sm_103) needs a CUDA-13 image.** From v0.20.0
+> on, vLLM release tags are CUDA-13 unsuffixed (e.g. `vllm/vllm-openai:v0.26.0` —
+> release tags are multi-arch), with `-cu129` as the CUDA-12 opt-out; v0.19.x and
+> earlier spelled it the other way round (`-cu130` suffix = CUDA 13). v0.20.0
+> ships both suffixes; after it, `-cu130` does not exist. Don't trust the tag
+> name — select a tag whose arm64 config blob reports `CUDA_VERSION` >= 13. A
+> cu12 build has **no sm_103 FP4 kernel**, so vLLM
 > loads the checkpoint then dies at engine init with `CUDA error: no kernel image
 > is available for execution on the device` (affects the `flashinfer` and
 > `cutlass` NVFP4 backends; `marlin` separately fails on non-64-divisible layer
--- a/.agents/skills/deployment/SKILL.md
+++ b/.agents/skills/deployment/SKILL.md
@@ -130,7 +130,9 @@ For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 > release tags are multi-arch), with `-cu129` as the CUDA-12 opt-out; v0.19.x and
 > earlier spelled it the other way round (`-cu130` suffix = CUDA 13). v0.20.0
 > ships both suffixes; after it, `-cu130` does not exist. Don't trust the tag
-> name — select a tag whose arm64 config blob reports `CUDA_VERSION` >= 13. A
+> name — select a tag whose config blob reports `CUDA_VERSION` >= 13, resolving
+> the child manifest for the platform you deploy on (arm64 for Grace/GB300,
+> amd64 for x86 hosts); `TORCH_CUDA_ARCH_LIST` differs between the two. A
 > cu12 build has **no sm_103 FP4 kernel**, so vLLM
 > loads the checkpoint then dies at engine init with `CUDA error: no kernel image
 > is available for execution on the device` (affects the `flashinfer` and
--- a/.agents/skills/deployment/SKILL.md
+++ b/.agents/skills/deployment/SKILL.md
@@ -126,18 +126,16 @@ python -m vllm.entrypoints.openai.api_server \
 For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 
 > **NVFP4 on Blackwell B300/GB300 (sm_103) needs a CUDA-13 image.** From v0.20.0
-> on, vLLM release tags are CUDA-13 unsuffixed (e.g. `vllm/vllm-openai:v0.26.0` —
-> release tags are multi-arch), with `-cu129` as the CUDA-12 opt-out; v0.19.x and
-> earlier spelled it the other way round (`-cu130` suffix = CUDA 13). v0.20.0
-> ships both suffixes; after it, `-cu130` does not exist. Don't trust the tag
-> name — select a tag whose config blob reports `CUDA_VERSION` >= 13, resolving
-> the child manifest for the platform you deploy on (arm64 for Grace/GB300,
-> amd64 for x86 hosts); `TORCH_CUDA_ARCH_LIST` differs between the two. A
-> cu12 build has **no sm_103 FP4 kernel**, so vLLM
-> loads the checkpoint then dies at engine init with `CUDA error: no kernel image
-> is available for execution on the device` (affects the `flashinfer` and
-> `cutlass` NVFP4 backends; `marlin` separately fails on non-64-divisible layer
-> dims).
+> on, release tags are CUDA-13 unsuffixed (e.g. `vllm/vllm-openai:v0.26.0`) with
+> `-cu129` the CUDA-12 opt-out; v0.19.x and earlier were the other way round
+> (`-cu130` = CUDA 13), and `-cu130` no longer exists after v0.20.0. Don't trust
+> the tag name — select a tag reporting `CUDA_VERSION` >= 13 in the config blob
+> of your platform's child manifest (arm64 Grace/GB300, amd64 x86);
+> `TORCH_CUDA_ARCH_LIST` differs between the two. A cu12 build has **no sm_103
+> FP4 kernel**, so vLLM loads the checkpoint then dies at engine init with `CUDA
+> error: no kernel image is available for execution on the device` (affects the
+> `flashinfer` and `cutlass` NVFP4 backends; `marlin` separately fails on
+> non-64-divisible layer dims).
 > Cross-check via
 > `recipes.vllm.ai/<org>/<model>?hardware=b300` (JS-rendered — fetch the raw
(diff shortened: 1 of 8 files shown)


.agents/skills/evaluation/SKILL.md
```diff
@@ -166,7 +166,14 @@ For how to choose `--tensor-parallel-size` / `--data-parallel-size` / `--pipelin
...
+> **NVFP4 on Blackwell B300/GB300 (sm_103) needs a CUDA-13 build** — the cu12 build has no sm_103 FP4 kernel, so engine init dies with `CUDA error: no kernel image is available`. **Verify CUDA ≥13 in the tag itself; do not blindly append a suffix — vLLM inverted its tag convention:**
+>
+> | vLLM version | CUDA-13 tag | CUDA-12 tag |
+> | --- | --- | --- |
+> | ≤ v0.20.x | **suffixed** `-cu130` (e.g. `v0.20.0-cu130`) | unsuffixed |
+> | ≥ ~v0.21 | **unsuffixed** (e.g. `v0.24.0-ubuntu2404`, `v0.26.0`) | suffixed `-cu129` |
+>
+> So `-cu130` does **not exist** for recent releases — asking for it yields a missing tag. Confirm by reading `CUDA_VERSION` from the tag's **arm64** config blob (registry API) rather than trusting the name, and check the arch you need is in `TORCH_CUDA_ARCH_LIST`. Multimodal on sm_103 may also need `--mm-encoder-attn-backend TRITON_ATTN`. Full note in `recipes/examples/example_eval.yaml`.
```
meenchen, line 177: > _Bot comment._

The linked “Full note” in `.agents/skills/evaluation/recipes/examples/example_eval.yaml` still says to append `-cu130` and fall back to `cu130-nightly-x86_64`, so an agent copying the canonical template will still follow exactly the stale rule this PR is fixing. Please update that note in the same change. A repository search also finds the unconditional append rule in `.agents/skills/deployment/SKILL.md`; since this convention applies to deployment generally, that guidance should be corrected too (or the PR should explain why it intentionally remains different).

meenchen, line 177: > _Bot comment._

Please resolve the config blob for the deployment platform rather than always using the arm64 child. GB300/Grace deployments may be arm64, but B300 deployments can be `linux/amd64`, and a multi-arch tag does not guarantee identical `CUDA_VERSION` across child images. The durable instruction should say to inspect the target platform’s child manifest (for example, arm64 for Grace and amd64 for x86 hosts).

meenchen (commented):

The new version table is useful, but the repository still directs agents to use the obsolete `-cu130` rule in the linked evaluation template and deployment skill. The proposed registry check also hard-codes arm64 even though the guidance covers both arm64 and amd64 deployments.

chadvoegele (approved):

.agents/skills/evaluation/SKILL.md
```diff
@@ -166,7 +166,14 @@ For how to choose `--tensor-parallel-size` / `--data-parallel-size` / `--pipelin
...
```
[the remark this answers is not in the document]
cjluo-nv: Confirmed and fixed in `4d28ccbd2` — you were right, and the boundary was off by a minor release in **both** rows.

Re-verified against the registry (`CUDA_VERSION` from the config blob):

| release | unsuffixed | `-cu130` | `-cu129` |
|---|---|---|---|
| v0.18.0 | 12.9.1 | 13.0.1 | absent |
| v0.19.0 / v0.19.1 | 12.9.1 | 13.0.1 | absent |
| **v0.20.0** | **13.0.2** | 13.0.2 | 12.9.1 |
| v0.20.1 | 13.0.2 | **absent** | 12.9.1 |
| v0.20.2 | 13.0.2 | **absent** | 12.9.1 |
| v0.21.0 … v0.26.0 | 13.0.2 | absent | 12.9.1 |

The table now reads **≤ v0.19.x suffixed / ≥ v0.20.0 unsuffixed**, with v0.20.0 called out as the transition release. v0.20.0 is what hid the boundary: it publishes both suffixes *and* its unsuffixed tag is already CUDA 13, so verifying only that `v0.20.0-cu130` exists reads as "the old convention still applies at 0.20" — while `v0.20.1-cu130` and `v0.20.2-cu130` don't exist at all. Following the old row on either would have produced exactly the missing-tag failure this PR set out to fix.

On the second half of your suggestion — "accepts only a release tag **or nightly** whose registry config reports `CUDA_VERSION >= 13`" — I've taken the framing but not the nightly.

Adopted: the blob check is now the *selection criterion* rather than a confirmation step ("select any tag whose config blob reports `CUDA_VERSION` ≥ 13"), and it resolves the child manifest for the deployment platform rather than always arm64.

Not adopted: re-admitting `nightly` as a fallback. This PR removed that path deliberately. The `cu130-nightly-<arch>` tags it pointed at are ~v0.20-era builds — older than several models' documented minimum vLLM version — so the escape hatch drops you below the floor the same section tells you to respect one paragraph earlier. `nightly` is also a moving tag, which contradicts the never-`:latest` reproducibility rule in that same section. When a pinned release genuinely predates a model, the right answer is the image that model's own recipe names, not a nightly.

.agents/skills/evaluation/SKILL.md
cjluo-nv: Agreed on both counts — done in `1f04f3d30`.

`recipes/examples/example_eval.yaml` was the worse of the two: `SKILL.md` points at it as the "full note", so the canonical template an agent copies still carried the verbatim "append `-cu130`" rule plus the `cu130-nightly-x86_64` fallback. Its comment block now matches the corrected `SKILL.md` guidance, and the nightly fallback is gone.

The deployment skill was not intentionally different — it just wasn't in the original single-file scope. It's corrected now, and while I was there, the same stale wording turned up in two more files under it:

- `.agents/skills/deployment/SKILL.md` — the NVFP4 note (the unconditional append rule you found)
- `.agents/skills/deployment/references/support-matrix.md` — "CUDA-13 (`cu130`) serving image … see the `-cu130` note"
- `.agents/skills/deployment/references/benchmarking.md` — "NVFP4 on Blackwell sm_103 needs a cu130 image"
- `.agents/skills/deployment/references/setup.md` — the two `vllm/vllm-openai:latest` pins, now `v0.26.0`, matching the evaluation skill's never-`:latest` stance

The convention now reads the same way in both skills.

cjluo-nv: Adopted in `0764efef9` — the instruction now resolves the child manifest **for the platform you deploy on** (arm64 for Grace/GB300, amd64 for x86 hosts) instead of always reading arm64.

One correction to the premise, for the record. I checked whether `CUDA_VERSION` actually varies across children, and on every tag I pulled it's identical:

| tag | amd64 | arm64 |
|---|---|---|
| v0.26.0 | 13.0.2 | 13.0.2 |
| v0.26.0-cu129 | 12.9.1 | 12.9.1 |
| v0.20.0 | 13.0.2 | 13.0.2 |
| v0.19.1 | 12.9.1 | 12.9.1 |
| v0.19.1-cu130 | 13.0.1 | 13.0.1 |
| kimi-k3 | 13.0.1 | 13.0.1 |

So the arm64 shortcut wasn't producing wrong CUDA answers — the multi-arch tags are built from one CUDA base per tag. `TORCH_CUDA_ARCH_LIST` is the field that genuinely differs:

```
v0.26.0  amd64  7.5 8.0 8.6 8.9 9.0 10.0 12.0
v0.26.0  arm64  8.0 8.7 8.9 9.0 10.0 11.0 12.0
v0.19.1  amd64  7.0 7.5 8.0 8.9 9.0 10.0 12.0
v0.19.1  arm64  8.7 8.9 9.0 10.0+PTX 12.0
```

`11.0` appears only on arm64; `7.5` / `8.6` only on amd64. Since the same sentence tells the reader to check their arch against that list, reading the arm64 child on an x86 host answers the arch question from an image they'll never run. So the change is worth making — for the arch check rather than the CUDA check — and the note now says the list is platform-specific.

.agents/skills/deployment/SKILL.md
```diff
@@ -125,12 +125,17 @@ python -m vllm.entrypoints.openai.api_server \
 
 For NVFP4 checkpoints, use `--quantization modelopt_fp4`.
 
-> **NVFP4 on Blackwell B300/GB300 (sm_103): append `-cu130` to the image tag**
-> (e.g. `vllm/vllm-openai:v0.19.1-cu130` — release tags are multi-arch). The
-> default cu12 build has **no sm_103 FP4 kernel**, so vLLM loads the checkpoint
-> then dies at engine init with `CUDA error: no kernel image is available for
-> execution on the device` (affects the `flashinfer` and `cutlass` NVFP4
-> backends; `marlin` separately fails on non-64-divisible layer dims).
+> **NVFP4 on Blackwell B300/GB300 (sm_103) needs a CUDA-13 image.** From v0.20.0
```
meenchen, line 128: > _Bot comment._

The linked `references/setup.md` still recommends `vllm/vllm-openai:latest` in both its official-image table and Docker example. Step 3 sends users to that file when vLLM is absent, so an agent can still select a moving image instead of the `v0.26.0` default established here. This also contradicts the PR body and author reply, which say those two pins were updated. Please restore that update (both occurrences).

meenchen (commented):

The earlier correctness issues are addressed: the v0.20.0 convention boundary is fixed, the deployment/template copies now use the same rule, and manifest inspection is platform-specific. The design choice is also adequately justified for this documentation fix: the PR prefers model-specific `recipes.vllm.ai` guidance plus registry config inspection over tag-name inference or moving nightly tags; a shared common reference would reduce duplication, but is not necessary for this small change. One stale linked setup path remains, however, despite the author reply and PR body saying its two `:latest` examples were pinned.

.agents/skills/deployment/SKILL.md
cjluo-nv: that's ok. I decided not to change the deployment skill behavior.

PR #2046, opened 2026-08-03
fix(quantization): Qwen3-VL MoE PTQ on transformers>=5.12 (NVBug 6518551)

Edwardf0t1:
### What does this PR do?

Type of change: Bug fix

Fixes [NVBug 6518551](https://nvbugspro.nvidia.com/bug/6518551) / [OMNIML-5581](https://jirasw.nvidia.com/browse/OMNIML-5581): PTQ of `Qwen3-VL-30B-A3B-Instruct` on `transformers` 5.12.1 crashes during `mtq.quantize`:

```
File "modelopt/torch/quantization/plugins/huggingface.py", line 899, in _setup
    nn.Linear(self.hidden_size, expert_dim, bias=False)
AttributeError: 'QuantQwen3VLMoeTextExperts' object has no attribute 'hidden_size'. Did you mean: 'hidden_dim'?
```

**Root cause.** transformers 5.12 moved `Qwen3VLMoeTextExperts` onto the standard `@use_experts_implementation` fused layout:

| | `transformers<5.12` | `transformers>=5.12` |
|---|---|---|
| dims | `hidden_size`, `expert_dim` | `hidden_dim`, `intermediate_dim` |
| `gate_up_proj` | `(E, hidden, 2*expert_dim)` | `(E, 2*intermediate_dim, hidden)` |
| forward | `torch.bmm` / `@` | `F.linear` twice per expert |

The legacy `_QuantQwen3VLMoeTextExperts` wrapper only understands the older layout, yet stayed statically registered at import time. Because `register_fused_experts_on_the_fly` skips any type already in `QuantModuleRegistry`, that stale registration shadowed the generic `_QuantFusedExperts` — which already handles the new layout correctly (Mixtral, Qwen3-MoE, Qwen3.5-MoE, MiniMax-M2/M3 all go through it).

**Fix.** Detect the new layout via the `_apply_gate` class attribute that `@use_experts_implementation` sets, and skip the static registration so on-the-fly detection claims it with `_QuantFusedExperts`.

The legacy wrapper must stay registered on `transformers<5.12`: that layout is structurally indistinguishable from a generic fused-experts module (`num_experts` + 3-D `gate_up_proj`/`down_proj`), but its `torch.bmm`-based forward is not intercepted by the generic wrapper, so simply deleting the registration would silently quantize nothing. `pyproject.toml` supports `transformers>=4.56,<5.13`, so both layouts must keep working.

### Usage

```bash
python hf_ptq.py --model Qwen3-VL-30B-A3B-Instruct \
  --recipe general/ptq/fp8_default-kv_fp8 \
  --dataset cnn_dailymail --calib_size 512 \
  --export_path Qwen3-VL-30B-A3B-Instruct-fp8_default-kv_fp8
```

### Testing

New `TestQwen3VLMoeTextExperts` in `tests/unit/torch/quantization/plugins/test_fused_experts.py` exercises the *real* transformers module (not a synthetic stand-in) and adapts to the installed layout:

- `test_registration_matches_installed_layout` — new layout is left to on-the-fly detection and resolves to `_QuantFusedExperts`; old layout keeps the explicit registration.
- `test_convert_and_forward_matches_reference` — conversion succeeds and is numerically transparent before calibration.
- `test_quantize_collects_amax` — calibration actually reaches the experts rather than silently no-op'ing.

Verified in two environments (there was no prior test coverage for quantized Qwen3-VL MoE experts):

| | `transformers==4.57.6` (old) | `transformers==5.12.1` (new) |
|---|---|---|
| new tests, without the fix | 3 passed | **3 failed** — reproduces the reported `AttributeError` |
| new tests, with the fix | 3 passed | 3 passed |
| full `test_fused_experts.py` | 48 passed | 48 passed |
| full `tests/unit/torch/quantization/plugins/` | 99 passed, 6 skipped | 99 passed, 6 skipped |

`pre-commit run --files ...` passes on all changed files.

- Is this change backward compatible?: ✅ — `transformers<5.12` keeps the legacy wrapper unchanged.

### Additional Information

Reported by Kenny Kang (SW-GPU) on GB200×4, `nvcr.io/nvidia/pytorch:26.05-py3`, modelopt 0.46.0rc0. Committed for ModelOpt 0.46.0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Qwen3-VL MoE post-training quantization compatibility with Transformers 5.12 and later.
  * Improved handling of updated fused-expert layouts while preserving support for older layouts.
  * Ensured FP8 calibration produces valid activation ranges across supported layouts.

* **Tests**
  * Added regression coverage for Qwen3-VL MoE expert conversion, numerical accuracy, and FP8 calibration across Transformers versions.


cjluo-nv (commented):

Small, well-scoped fix: the static registration of the legacy `_QuantQwen3VLMoeTextExperts` wrapper is now skipped when the class carries `_apply_gate` (transformers>=5.12's `@use_experts_implementation` fused layout), so `register_fused_experts_on_the_fly` claims the module with the generic `_QuantFusedExperts`. That matches how Mixtral/Qwen3-MoE/MiniMax already flow through `_fused_experts_wrapper_class`, i.e. no new abstraction is introduced, and CHANGELOG + tests are included. No licensing surface touched. A few things I'd like the owner to eyeball before this lands:

- **Detection hinges on an undocumented transformers internal.** The whole fix is `not hasattr(Qwen3VLMoeTextExperts, "_apply_gate")`. If that private attribute is renamed/removed in a later 5.x while the fused layout stays, the legacy wrapper silently re-registers and users get the same confusing `AttributeError: ... has no attribute 'hidden_size'`. The file already has a version constant precedent (`TRANSFORMERS_VERSION_GE_5_0`); consider OR-ing in a version check, and/or adding a guard in `_QuantQwen3VLMoeTextExperts._setup` (e.g. assert the legacy `gate_up_proj` orientation / absence of `_apply_gate`) so a future rename fails with an actionable message instead of the original bug. I can't verify from the repo that `_apply_gate` is set on the class in 5.12 — that rests entirely on the author's two-environment run.
- **Export path for the new layout is untested here.** Coverage stops at calibration (`test_quantize_collects_amax`). Because the wrapper changes, the exported per-expert naming for Qwen3-VL MoE now goes through `_export_fused_experts` (per-expert `<idx>.gate_proj/...`) on >=5.12 vs. the legacy wrapper's unrolled `nn.ModuleList` naming on <5.12 — i.e. the exported checkpoint layout differs by installed transformers version for the same model. That is probably the intended/mainline behavior (same as Mixtral), but someone with unified-export context should confirm no downstream naming regression, since there is no Qwen3-VL-specific handling in `modelopt/torch/export`.
- **Tests hard-import `transformers.models.qwen3_vl_moe`.** `pyproject.toml` allows `transformers>=4.56`, where that module doesn't exist yet; the three new tests would error (not skip) on the supported floor. A module/class-level `pytest.importorskip("transformers.models.qwen3_vl_moe")` would keep them hermetic.
- Minor: `test_registration_matches_installed_layout` asserts global registry state (`registered is None`) for the real transformers class, so any other test in the session that quantizes a Qwen3-VL-MoE model on the fly without unregistering would make it fail spuriously. Low risk today, just noting the coupling.

cjluo-nv (approved):

PR #2054, opened 2026-08-03
Fix vLLM 0.24+ compatibility: registry TypeError and MoE RoutedExperts port

kevalmorabia97:
### What does this PR do?

Type of change: Bug fix

vLLM 0.24 (shipped in `nemo:26.08`) reworked the fused-MoE layer and broke ModelOpt twice:

1. **`FusedMoE` became a factory function.** Registering it put a plain function into `QuantModuleRegistry`, so *every* later lookup raised `TypeError: issubclass() arg 2 must be a class, ...` — taking down ~120 `tests/gpu_megatron` tests (TE, Megatron chaining, MSE calibrator) that never touch vLLM.
2. **Expert weights moved onto a `RoutedExperts` submodule** of `MoERunner`, leaving the MoE fakequant path with no registration target.

Changes:

- `_DMRegistryCls.register` validates that keys are `nn.Module` subclasses before mutating the registry, so upstream changes fail at the registration site instead of at a much later lookup.
- Register whichever fused-MoE module class the installed release provides — `RoutedExperts` (0.24+) or `FusedMoE`/`SharedFusedMoE` (older) — both via `_QuantFusedMoEBase`. `MoERunner` calls `forward_modular`/`forward_monolithic` directly, so those are hooked instead of `forward`.
- The kernel patch also covers `experts.triton_moe`: 0.24 binds the kernel names at import, so patching only `fused_moe` would leave fakequant **silently inactive**.
- `examples/vllm_serve/vllm_reload_utils.py`: expert quantizer keys gain the `.routed_experts` hop, and **shared** experts are no longer excluded from gate/up merging (the guard skipped any key containing `"experts"`, so shared-expert state matched no module and reloaded uncalibrated).
- CI `gpu_vllm` runs one container per layout: **`v0.26.0`** (RoutedExperts) and **`v0.20.0`** (legacy). 0.24 — the version in `nemo:26.08` — takes an identical code path to 0.26, and is covered by the local runs below rather than by CI.

Test-fixture updates for newer vLLM (not product bugs): tiny Llama `head_dim=2` → `64` (rejected by `FLEX_ATTENTION`), tiny DeepSeek MLA dims → `128/64/128` (0.26 accepts only fixed triples), FlashInfer builder stub `causal=True` plus `q_data_type_prefill`/`q_data_type_decode` and `flashinfer_trtllm_api_decode_kernel`. All version-tolerant.

### Usage

No API change — existing `mtq.quantize` / `examples/vllm_serve` flows work unmodified on old and new vLLM.

### Testing

Run in **both** containers, with identical results: `vllm/vllm-openai:v0.26.0` (what CI pins) and `nemo:26.08.rc3` (vLLM `0.24.1.dev0`, the environment that motivated the PR).

- `tests/gpu_vllm`: **73 passed, 1 skipped** on each (was 70 passed / 3 failed).
- `tests/gpu_megatron`: all pass (previously ~120 failures from the registry `TypeError`).
- `tests/unit/torch/opt/test_dynamic.py`: 2 passed, incl. the new `test_register_rejects_non_module_classes`.

MoE fakequant was also verified by enumerating every weight-owning module after `mtq.quantize`, on tiny DeepSeek-V3 (MLA + routed MoE + shared experts) and tiny Qwen3-MoE:

```
model.layers.0.mlp.experts.routed_experts [QuantRoutedExperts]
  w13_input_quantizer=3.484  w13_weight_quantizer=0.0840
  w2_input_quantizer=0.1060  w2_weight_quantizer=0.0845
model.layers.0.mlp.shared_experts.gate_up_proj [QuantMergedColumnParallelLinear]  OK
```

Populated *weight* amax (not just input amax) proves the patched kernel really ran — i.e. `forward_modular`/`forward_monolithic` are the live entry points and `experts.triton_moe` is the binding site. Unquantized modules were only the expected ones (embeddings, RMSNorms, router `gate`, `lm_head`).

**Known gaps (pre-existing, not regressions)**

- `DeepSeekV2FusedQkvAProjLinear` is unquantized: it subclasses `MergedColumnParallelLinear` but overrides `forward`, so the registry's shared-forward rule declines it. Pre-0.24 the equivalent layers were `ReplicatedLinear`, which was never quantized either.
- MoE fakequant hooks only the Triton expert kernels; FlashInfer/CUTLASS/DeepGEMM MoE backends bypass them (hence `moe_backend="triton"` in the fixtures).
- `_setup` still requires a plain `UnquantizedFusedMoEMethod`; a `FusedMoEModularMethod` swap still asserts.
- The registry key moves `vllm_FusedMoE` → `vllm_RoutedExperts` and quantizer paths gain `.routed_experts`, so a `modelopt_state` saved under an older vLLM will not restore onto 0.24+ as-is. Documented in the CHANGELOG.

- Is this change backward compatible?: ✅ — all new paths are feature-detected; older vLLM keeps the `FusedMoE` registration, covered by the `v0.20.0` CI job.
- Did you write any new necessary tests?: ✅ — registry-validation unit test, plus reload key-mapping assertions for routed and shared experts.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — under 0.46 Bug Fixes.

### Additional Information

Found while bumping the Megatron test environment from `nemo:26.06` to `nemo:26.08.rc3`.

## Summary by CodeRabbit

* **New Features**
  * Added compatibility for newer vLLM MoE implementations, module layouts, and routed-expert configurations.
  * Improved model reload support for models using routed-expert submodules.
* **Bug Fixes**
  * Improved detection and patching of vLLM MoE execution paths across supported configurations.
  * Registry validation now rejects invalid module registrations without partially applying changes.
* **Tests**
  * Expanded GPU coverage for vLLM 0.24.0, dynamic module validation, and causal attention metadata paths.


modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -112,6 +112,20 @@ def _import_attention_module():
...
         continue
 
+
+def _is_module_cls(obj) -> bool:
+    """Whether obj is an nn.Module class we can register a dynamic module for."""
+    return isinstance(obj, type) and issubclass(obj, torch.nn.Module)
+
+
+# vLLM >= 0.24 turned ``FusedMoE`` into a factory function returning a ``MoERunner`` pipeline and
+# dropped ``UnquantizedFusedMoEMethod``, so the fakequant MoE module below no longer applies there.
```
cjluo-nv: > _Bot comment._

Two suggestions here:

1. `_QuantFusedMoEBase._setup()` also dereferences `vllm_fused_moe_layer.UnquantizedFusedMoEMethod`, which the PR body says was dropped in 0.24. The current gate only checks that `FusedMoE` is an `nn.Module` subclass, so if a version ever keeps a class-shaped `FusedMoE`/`SharedFusedMoE` while dropping `UnquantizedFusedMoEMethod`, conversion will blow up with an `AttributeError` at `_setup` instead of being skipped. Consider folding `hasattr(vllm_fused_moe_layer, "UnquantizedFusedMoEMethod")` into the gate (and applying it to the `SharedFusedMoE` path too).
2. When the gate is False, MoE quantization is silently a no-op — `mtq.quantize()` will happily return a model whose experts are untouched. Please emit a one-time `warnings.warn(...)` here so users on nemo:26.08 know fakequant MoE is unsupported rather than discovering it via unexpected accuracy numbers.

Nit: `_has_fused_moe_module` reads as "the python module exists"; `_has_fused_moe_cls` would be clearer.

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -604,9 +618,11 @@ def fold_weight(self, keep_attrs: bool = False):
             torch.cuda.empty_cache()
 
 
-@QuantModuleRegistry.register({vllm_fused_moe_layer.FusedMoE: "vllm_FusedMoE"})
-class _QuantVLLMFusedMoE(_QuantFusedMoEBase):
-    pass
+if _has_fused_moe_module:
```
cjluo-nv: > _Bot comment._

`tests/gpu_vllm/torch/quantization/test_vllm_dynamic_modules.py` asserts `summary["moe_count"] >= 2` for both the Qwen3-MoE and DeepSeek-V3 fixtures. With this gate those tests will now fail on vLLM 0.24+ (0 converted MoE modules) instead of erroring at import — so the tests still don't pass in the target container. Please add a `pytest.mark.skipif(not vllm_plugin._has_fused_moe_module, ...)` (or equivalent branch) to those cases so the new behavior is asserted rather than left failing.

modelopt/torch/opt/dynamic.py
```diff
@@ -1044,6 +1044,10 @@ class DynamicLinear(DynamicModule):
         def decorator(dm_class: type[_DMRegistryCls.T]) -> type[_DMRegistryCls.T]:
             """Register dnn_class with appropriate nn_class."""
             for nn_cls_, key in cls_to_key.items():
+                # Fail here instead of during a much later issubclass() lookup
```
cjluo-nv: > _Bot comment._

Good fail-fast. Two small points: (a) elsewhere in this file the same check is spelled `isinstance(x, type)` (see `_DMRegistryCls.get`, and `_is_module_cls` added in the vllm plugin in this same PR) — worth picking one spelling; (b) this new validation path isn't covered by any test. A two-line unit test asserting `DMRegistry.register({some_function: "key"})` raises would lock in the intended behavior.

cjluo-nv (commented):

Small, focused compat shim for vLLM 0.24+ (`FusedMoE` became a factory function). The mechanism is right and the `dynamic.py` fail-fast assert is a reasonable improvement. Two things I'd like addressed before merge:

1. **Silent degradation**: on vLLM 0.24+ MoE layers are now simply never converted, so `mtq.quantize()` on an MoE model returns a model whose experts are in full precision with no diagnostic. A one-time `warnings.warn` (or logger warning) when `_has_fused_moe_module` is False would keep users from silently mis-measuring accuracy/perf in the nemo:26.08 container.
2. **Test coverage / claim of "fixed and tested"**: `tests/gpu_vllm/torch/quantization/test_vllm_dynamic_modules.py` asserts `summary["moe_count"] >= 2` for the Qwen3-MoE and DeepSeek fixtures. With this patch those two tests will fail (0 MoE modules converted) rather than error at import, so "most modelopt tests fail" isn't fully fixed for the MoE cases. Please gate/skip those assertions on `vllm_plugin._has_fused_moe_module` so the intended state is explicit and CI on 0.24+ is green. A tiny unit test for the new `register()` assert in `dynamic.py` would also be cheap.

Also a minor robustness note on the gating condition (inline). No licensing or design concerns.

kevalmorabia97:
Addressed review feedback in `cd17d37a3b`:

- **Gate all MoE registrations on `UnquantizedFusedMoEMethod`** — `FusedMoE`, `SharedFusedMoE` and `RoutedExperts` are now skipped when it is not importable, so `_setup()` cannot fail mid-conversion. Added a one-time `warnings.warn` at import when no fused-MoE module class is available, since MoE quantization is otherwise a silent no-op.
- **Renamed** `_has_fused_moe_module` → `_has_fused_moe_cls` (it gates a class, not a module).
- **`register()` validation** — validate every entry before mutating `_registry`/`_key_registry` so a bad entry cannot leave a partial registration behind, and spell the check as `isinstance(x, type)` to match the rest of the file. Covered by a new unit test in `tests/unit/torch/opt/test_dynamic.py`.
- Trimmed added comments/docstrings to 1-2 lines.

Two notes on comments I did not apply:

- *Skip the MoE tests on vLLM 0.24+*: obsolete — `e3e96826ea` ports the fakequant MoE module to the `RoutedExperts` layout, so `test_tiny_qwen3_moe_quantize` and `test_tiny_deepseek_mla_quantize` pass on 0.24 rather than converting 0 modules.
- *Replace `assert` with explicit exceptions in `register()`*: left as-is for consistency — `_DMRegistryCls`/`DynamicModule` use asserts throughout for these invariants, and the atomicity half of that comment is fixed.

Verified in `nemo:26.08.rc3` (vLLM 0.24.1): `tests/gpu_vllm` 73 passed / 1 skipped, `tests/gpu_megatron` passing, pre-commit clean.

cjluo-nv (commented):

Re-review of the vLLM 0.24+ compat fix (now +187/-43 across plugin, registry, example, tests, CI).

**Design-review protocol (gate fired on 6 directories): no new subsystem.** The problem is narrow — vLLM 0.24 turned `FusedMoE` into a factory and moved expert weights onto `RoutedExperts`, breaking `QuantModuleRegistry` lookups and the MoE fakequant hook. The change reuses the existing in-repo pattern for third-party integrations (capability-probed `plugins/*.py` + `QuantModuleRegistry.register({cls: "key"})` gated by `_is_module_cls`/`importlib` probes, exactly as `CrossAttention`/`EncoderOnlyAttention`/`MLAAttention` and `SharedFusedMoE` already do in this same file). The only new abstraction is extracting the kernel monkey-patching into a `_fakequant_moe_kernels()` contextmanager, which is then shared by `forward`, `forward_modular` and `forward_monolithic` — that's dedup, not a second system. No alternative library (omegaconf, importlib-based registry, etc.) applies here. Design is fine; no justification needed in the PR body.

**Previously raised, now addressed:**
- :speech_balloon: Author replied in `cd17d37a3b`: all MoE registrations (`FusedMoE`, `SharedFusedMoE`, `RoutedExperts`) are now gated on `UnquantizedFusedMoEMethod` being importable, so `_QuantFusedMoEBase._setup()` can no longer assert mid-conversion, and a one-time import-time `warnings.warn` fires when no fused-MoE class is available — the silent-degradation concern is resolved. `_has_fused_moe_module` → `_has_fused_moe_cls` renamed as suggested.
- :speech_balloon: Author replied that the "skip the MoE tests on 0.24+" request is obsolete: `e3e96826ea` ports fakequant to the `RoutedExperts` layout, so `moe_count >= 2` is satisfied rather than skipped. I confirmed statically that `_QuantVLLMRoutedExperts` subclasses `_QuantFusedMoEBase` (which the test's `moe_count` counts) and that the kernel patch now also covers `experts.triton_moe`, matching the `moe_backend="triton"` fixture — still worth an owner eye because correctness here depends on vLLM 0.24 internals I can't verify from this repo (that `forward_modular`/`forward_monolithic` are the real entry points, and that `SharedFusedMoE` is itself a factory in 0.24 so the gate nulls it rather than converting a weight-less wrapper alongside its inner `RoutedExperts`).
- Registry validation is now atomic (validate-then-mutate), spelled `isinstance(x, type)` to match the rest of `dynamic.py`, and covered by `tests/unit/torch/opt/test_dynamic.py::test_register_rejects_non_module_classes` (including the partial-registration case). The `assert` → explicit-exception request was declined for consistency with the rest of `_DMRegistryCls`; that's reasonable and consistent with project convention.

**Remaining, minor — for the owner to weigh:**
1. `examples/vllm_serve/vllm_reload_utils.py::_EXPERTS_INFIX` re-derives the layout with `importlib.util.find_spec("...fused_moe.routed_experts")`, while the plugin already computes the authoritative flag (`_has_routed_experts_cls`, which also requires `UnquantizedFusedMoEMethod`). On a release where the module exists but registration is skipped, the example would insert `.routed_experts` for module paths that don't exist (degrades to the "missing quantizer key" warning, not a crash). Reusing the plugin flag would keep one source of truth. The PR body notes this path is verified by inspection only.
2. With the CI container bumped `v0.20.0` → `v0.24.1`, the legacy `FusedMoE`/`SharedFusedMoE` registration branches now have no CI coverage at all — the backward-compat claim rests on the pre-bump green run.
3. Nit: `_QuantVLLMRoutedExperts` still inherits `_QuantFusedMoEBase.forward`, which wraps a `super().forward(hidden_states, router_logits)` call into `RoutedExperts.forward` that raises by design — harmless, but an explicit override/comment would make the intent clearer.

No licensing concerns (no license/SPDX/vendoring changes; touched files keep the existing NVIDIA header). No prompt-injection attempts in the PR content.

kevalmorabia97:
Addressed the Claude review in `6ac54ef00b` (that review targeted `e3e96826ea`/`cd17d37a3b`; some of it had already landed).

**Finding 1 — `_EXPERTS_INFIX` diverging from the plugin (IMPORTANT): fixed.** The example now imports `_has_routed_experts_cls` from the plugin instead of re-deriving the layout with `find_spec`. That is slightly stronger than the suggested `RoutedExperts is not None`, since it is the exact flag the registration is gated on (it also requires `UnquantizedFusedMoEMethod` and the `forward_*` entry points). Verified against the real module tree: the mapping now emits `model.layers.0.mlp.experts.routed_experts.w13_input_quantizer`, matching the probed module path on vLLM 0.24.1.

**Finding 2 — `UnquantizedFusedMoEMethod is None` producing a misleading `_setup` assert: already impossible, and the suggested fix would be worse.** Since `cd17d37a3b`, all three registrations (`FusedMoE`, `SharedFusedMoE`, `RoutedExperts`) are gated on that symbol, so a `None` means nothing is registered and `_setup` never runs — we emit the one-time warning instead. Raising `ImportError` at plugin import would be actively harmful here: `modelopt/torch/utils/import_utils.py::import_plugin` catches `Exception` and only warns, so it would silently disable the *entire* vLLM plugin, losing linear and attention quantization as well.

**Finding 3 — `_QuantVLLMRoutedExperts` robustness (SUGGESTION): part 2 taken, part 1 declined.** Registration now also requires `RoutedExperts` to expose `forward_modular`/`forward_monolithic`, so a future rename surfaces at import (skip + warn) rather than as an `AttributeError` at the first expert forward. I did not override `forward` to raise: `RoutedExperts.forward` already raises vLLM's own clear error, and the inherited wrapper restores the kernel patches in its `finally` before propagating it — an override would replace a precise upstream message with our own. Documented in the class docstring instead.

**Finding 4 — `_fakequant_moe_kernels` re-entrancy (SUGGESTION): documented, guard declined.** The stated failure mode does not occur for properly nested context managers: the inner `finally` restores the outer's `partial` and the outer then restores the true originals, so LIFO nesting unwinds correctly. The suggested `if _moe_fakequant_active.get(): yield; return` guard would introduce a bug — a nested *different* MoE module would then quantize its activations with the enclosing module's quantizers. The real residual is thread/stream concurrency, which is pre-existing; added a one-line note that patching is process-wide and not thread-safe.

**Coverage note (also raised by @cjluo-nv):** added a second `gpu_vllm` CI matrix entry on `vllm-openai:v0.20.0` so the pre-0.24 `FusedMoE`/`SharedFusedMoE` branches keep running. I could not verify that job locally (only the 0.24.1 container is available here), so its first run on this PR is the real check.

examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -39,6 +40,14 @@
 from modelopt.torch.quantization.nn import SequentialQuantizer, TensorQuantizer
 from modelopt.torch.quantization.utils import is_quantized
 
+# vLLM >= 0.24 moved the fused expert weights (and their quantizers) onto a ``routed_experts``
+# submodule of the MoE layer, so merged expert keys need that extra hop.
+_EXPERTS_INFIX = (
+    ".routed_experts"
+    if importlib.util.find_spec("vllm.model_executor.layers.fused_moe.routed_experts") is not None
+    else ""
+)
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in `6ac54ef00b`: the example now imports `_has_routed_experts_cls` from the plugin instead of re-probing with `find_spec`, so the keys always match the modules registration actually converted. Verified against the real module tree on vLLM 0.24.1 — the mapping emits `model.layers.0.mlp.experts.routed_experts.w13_input_quantizer`.

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -112,6 +114,54 @@ def _import_attention_module():
...
+    RoutedExperts = None
+
+# vLLM >= 0.24 turned ``FusedMoE`` into a factory function and moved the expert weights onto a
+# ``RoutedExperts`` submodule; register whichever module class this release provides.
+_has_fused_moe_cls = UnquantizedFusedMoEMethod is not None and _is_module_cls(
+    getattr(vllm_fused_moe_layer, "FusedMoE", None)
+)
+_has_routed_experts_cls = UnquantizedFusedMoEMethod is not None and _is_module_cls(RoutedExperts)
+if vllm_shared_fused_moe_layer is not None and not (
+    UnquantizedFusedMoEMethod is not None
```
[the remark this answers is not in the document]
kevalmorabia97: Already impossible since `cd17d37a3b`: all three registrations are gated on `UnquantizedFusedMoEMethod`, so a `None` means nothing registers and `_setup` never runs (we emit the one-time warning instead). Not taking the `raise ImportError` variant — `modelopt/torch/utils/import_utils.py::import_plugin` catches `Exception` and only warns, so it would silently disable the whole vLLM plugin, losing linear and attention quantization too.

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -618,6 +686,24 @@ class _QuantVLLMSharedFusedMoE(_QuantFusedMoEBase):
...
 
+if _has_routed_experts_cls:
+
+    @QuantModuleRegistry.register({RoutedExperts: "vllm_RoutedExperts"})
+    class _QuantVLLMRoutedExperts(_QuantFusedMoEBase):
+        """Expert-weight owner of the vLLM >= 0.24 ``MoERunner`` pipeline.
+
+        ``MoERunner`` calls ``forward_modular``/``forward_monolithic`` instead of ``__call__``.
+        """
+
```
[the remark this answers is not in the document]
kevalmorabia97: Part 2 taken in `6ac54ef00b`: registration now also requires `RoutedExperts` to expose `forward_modular`/`forward_monolithic`, so a rename surfaces at import (skip + warn) rather than as an `AttributeError` at serving time. Part 1 declined: `RoutedExperts.forward` already raises vLLM's own clear error and the inherited wrapper restores the kernel patches in its `finally` before propagating, so an override would only replace a precise upstream message with ours. Documented in the class docstring.

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -559,35 +628,32 @@ def _invoke_fused_moe_quantized_function(
         else:
```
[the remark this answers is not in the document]
kevalmorabia97: Documented rather than guarded, in `6ac54ef00b`. The described failure does not occur for properly nested context managers: the inner `finally` restores the outer's `partial` and the outer then restores the true originals, so LIFO nesting unwinds correctly. The suggested `if _moe_fakequant_active.get(): yield; return` guard would introduce a bug — a nested *different* MoE module would quantize its activations with the enclosing module's quantizers. The real residual is thread/stream concurrency (pre-existing), now noted in the docstring.

cjluo-nv (commented):

Re-review at `6ac54ef00b` (vLLM 0.24+ compat: registry validation + `RoutedExperts` MoE port). +195/-42 across 7 files.

**Design gate (fired on 6 directories): still settled, no new subsystem.** The change reuses the existing in-repo third-party-integration pattern — capability-probed `plugins/vllm.py` + `QuantModuleRegistry.register({cls: "key"})` gated on `_is_module_cls`/`importlib` probes, exactly as `CrossAttention`/`EncoderOnlyAttention`/`MLAAttention`/`SharedFusedMoE` already do in the same file — and the only new abstraction (`_fakequant_moe_kernels()` contextmanager) is dedup shared by `forward`/`forward_modular`/`forward_monolithic`. No already-imported library applies. Proceeded to correctness.

**Previously raised, now resolved (dropped from the ask list):** all three MoE registrations gated on an importable `UnquantizedFusedMoEMethod` (+ one-time import-time `warnings.warn`), so `_setup()`'s `assert type(self.quant_method) is UnquantizedFusedMoEMethod` is unreachable with a `None` comparand; `register()` is validate-then-mutate with `isinstance(x, type)` and covered by `test_register_rejects_non_module_classes` including the partial-registration case; `_EXPERTS_INFIX` in `examples/vllm_serve/vllm_reload_utils.py` now imports the plugin's authoritative `_has_routed_experts_cls` instead of re-probing with `find_spec`; the registration gate additionally requires `forward_modular`/`forward_monolithic` to exist, so a future rename surfaces at import. I verified statically that `_QuantVLLMFusedMoE` and the `vllm_FusedMoE` key have no consumers outside the plugin, that `_QuantVLLMRoutedExperts` subclasses `_QuantFusedMoEBase` (what the tests' `moe_count` counts), and that `create_tiny_llama_dir` forwards `head_dim`/`hidden_size` to `LlamaConfig`.

**Why still a nudge rather than approve:**
- :speech_balloon: Author replied: *added a second `gpu_vllm` matrix entry on `vllm-openai:v0.20.0` so the legacy `FusedMoE`/`SharedFusedMoE` branches keep running, but could not run that job locally (only the 0.24.1 container is available)* — this PR also changes two shared test inputs that the legacy job must tolerate: the `tiny_llama` fixture (`hidden_size=128`, `num_attention_heads=2`, `head_dim=64`) and `causal=False`→`True` in `test_flashinfer_metadata_builder_patch_stashes_common_metadata`. The `causal` flip is justified by 0.24's `all_uses_trtllm = causal and ...`; whether 0.20's `build` stays on the same all-TRTLLM path is unverified. Please confirm the `v0.20.0` matrix job is green before merge.
- :speech_balloon: Author replied: *`RoutedExperts.forward` already raises vLLM's own clear error and the inherited wrapper restores the kernel patches in `finally`, so overriding it would replace a precise upstream message; re-entrancy is LIFO-safe and the suggested `if _moe_fakequant_active.get(): yield; return` guard would mis-attribute a nested module's quantizers* — both rationales check out against the code; documented in the class docstring / contextmanager docstring. Not blocking, just noting they were declined deliberately.
- `examples/vllm_serve/vllm_reload_utils.py` remains untested ("verified by inspection"). It is now consistent with the plugin flag, but the failure mode of a wrong expert key is still silent (dropped by `filter_modelopt_state_quantizer_state_for_model`, re-added uncalibrated, weight-quantizer invariant satisfied by `_disabled: True`). Also note the plugin is now imported directly rather than through `import_plugin`, so a plugin-import failure surfaces as a hard example-import error.
- Correctness of the port rests on vLLM 0.24 internals that can't be checked from this repo (that `forward_modular`/`forward_monolithic` are the live entry points, that `experts.triton_moe` is the binding site, that `SharedFusedMoE` is now a factory so the gate nulls it rather than converting a weight-less wrapper). The author's module-tree probe showing populated `w13_weight_quantizer`/`w2_weight_quantizer` amax is good evidence; an owner familiar with vLLM should still sign off.
- Nits: PR body's Changes section still says the CI container was "bumped `v0.20.0` → `v0.24.1`" though both are now kept; CodeRabbit's suggestion to also assert `"nn.Linear" not in registry` (string view of `_key_registry`) in the rollback test is unaddressed and cheap.

No licensing concerns (no LICENSE/SPDX/vendoring changes; touched files keep the existing NVIDIA header). No prompt-injection attempts in the PR content; the bot comments' "Prompt for AI Agents" blocks were treated as data, not instructions.

kevalmorabia97:
Addressed the re-review in `6b4b192db9`.

**CI image — the `v0.24.1` pin was broken.** The pull failed with `docker.io/vllm/vllm-openai:v0.24.1: not found`: vLLM never released 0.24.1 (releases go 0.24.0 → 0.25.0), and `nemo:26.08` ships a `0.24.1.dev0+gee0da84ab` build. Repinned to **`v0.24.0`**, which I verified upstream is the first release carrying the new layout and exposes everything the plugin gates on:

| symbol at `v0.24.0` | |
|---|---|
| `def FusedMoE(...)` in `layer.py` | factory, so the `RoutedExperts` path is the one exercised |
| `class RoutedExperts` + `forward_modular`:1063 / `forward_monolithic`:1100 | same line numbers as the dev build tested locally |
| `class UnquantizedFusedMoEMethod` in `unquantized_fused_moe_method.py` | gate satisfied |
| `experts/triton_moe.py` with 3 `invoke_fused_moe_triton_kernel` refs | patch site present |

For contrast, `v0.20.0` has `class FusedMoE(PluggableLayer)` — so the two matrix entries genuinely cover the new and legacy branches. **The `v0.20.0` job is already green on this PR**, which answers the "confirm before merge" ask for the legacy half, including the two shared test inputs (`tiny_llama` `head_dim=64` and the `causal=True` flip).

**Reload path is no longer untested.** `test_tiny_qwen3_moe_quantize` now maps HF expert keys through `_convert_key_for_vllm` and asserts the result resolves against the quantizer module paths of the booted model. A stale infix now fails loudly in CI instead of silently serving uncalibrated experts — which was the failure mode both reviews flagged as most impactful.

**Nits:** PR body no longer says the container was "bumped" (both are kept), and the rollback test now also asserts `"nn.Linear" not in registry`, covering the string-key view alongside the class view.

Local runs on `nemo:26.08.rc3`: `tests/gpu_vllm/torch/quantization/test_vllm_dynamic_modules.py` 10 passed, full `tests/gpu_vllm` 73 passed / 1 skipped, unit `test_dynamic.py` 2 passed.

cjluo-nv (commented):

Re-review of the vLLM 0.24+ compat fix (registry validation + `RoutedExperts` MoE port), now +225/-42 across 8 files at the CI-repin commit.

**Design gate (fired on 6 directories): still settled — no new subsystem.** The change reuses the established in-repo third-party-integration pattern: capability-probed `plugins/vllm.py` + `QuantModuleRegistry.register({cls: "key"})` gated on `_is_module_cls`/`importlib` probes, exactly as `CrossAttention`/`EncoderOnlyAttention`/`MLAAttention`/`SharedFusedMoE` already do in the same file. The only new abstraction, the `_fakequant_moe_kernels()` contextmanager, is dedup shared by `forward`/`forward_modular`/`forward_monolithic`, replacing the previous inline try/finally. No already-installed library (OmegaConf, importlib-based registry, etc.) applies. Proceeded to correctness.

**Previously raised, now resolved (verified in the tree, dropped from the ask list):**
- All three MoE registrations gated on an importable `UnquantizedFusedMoEMethod` (+ `forward_modular`/`forward_monolithic` presence for `RoutedExperts`), with a one-time import-time `warnings.warn` when no fused-MoE class is available — the silent-degradation and misleading-`_setup`-assert concerns are both structurally impossible now.
- `register()` is validate-then-mutate with `isinstance(x, type)`, covered by `test_register_rejects_non_module_classes`; the rollback test now asserts both the class view (`nn.Linear not in registry`) and the string view (`"nn.Linear" not in registry`), which matches `_DMRegistryCls.__contains__`'s two branches.
- :speech_balloon: Author replied: *the reload path is no longer inspection-only* — `test_tiny_qwen3_moe_quantize` now runs two HF expert keys through `_convert_key_for_vllm` and asserts the resulting module paths exist in the booted model's `quantizer_names`. I confirmed the assertion is meaningful: the infix is derived from the plugin's authoritative `_has_routed_experts_cls` (not a `find_spec` re-probe), so a stale infix fails loudly in both container variants instead of silently serving uncalibrated experts.
- :speech_balloon: Author replied: *the `v0.24.1` pin never existed upstream* — repinned to `v0.24.0` and the legacy `v0.20.0` entry is kept as a second matrix job, so both the new `RoutedExperts` and legacy `FusedMoE`/`SharedFusedMoE` branches stay covered. PR body no longer claims a bump.

**Why nudge rather than approve:**
- :speech_balloon: Author replied: *the `v0.20.0` job is already green on this PR* — I can't verify CI state from the repo, and this PR still changes inputs that job must tolerate (`tiny_llama` fixture `hidden_size=128`/`num_attention_heads=2`/`head_dim=64`, `causal=False`→`True` in the FlashInfer metadata-builder test, and the newly added example-module import inside `test_tiny_qwen3_moe_quantize`). Please confirm the `v0.20.0` matrix job is green on the current head before merge.
- Correctness of the port rests on vLLM 0.24 internals that can't be checked from this repo: that `forward_modular`/`forward_monolithic` are the live `MoERunner` entry points, that `experts.triton_moe` is the kernel binding site, and that `SharedFusedMoE` is now a factory so the gate nulls it rather than converting a weight-less wrapper. The author's module-tree probe showing populated `w13_weight_quantizer`/`w2_weight_quantizer` amax is strong evidence; an owner familiar with vLLM should still sign off, especially as this is slated for cherry-pick into 0.46.0.
- Note for the cherry-pick: the registry key moves `vllm_FusedMoE` → `vllm_RoutedExperts` and quantizer paths gain `.routed_experts`, so a `modelopt_state` saved under an older vLLM will not restore onto 0.24 as-is. That's documented in the PR body but not in CHANGELOG.
- Minor/out of scope: `tests/examples/megatron_bridge/test_distill.py` gets an unrelated `@pytest.mark.timeout(360)` bump; harmless but unconnected to the vLLM fix.

Declined suggestions (`RoutedExperts.forward` override, `_fakequant_moe_kernels` re-entrancy guard) check out against the code and are documented in docstrings — not blocking. No licensing concerns (no LICENSE/SPDX/vendoring changes; touched files keep the existing NVIDIA header). No prompt-injection attempts in the PR content; the bot comments' "Prompt for AI Agents" blocks were treated as data, not instructions.

.github/workflows/gpu_tests.yml
```diff
@@ -50,6 +50,10 @@ jobs:
           - example: gpu_trtllm
             timeout: 15
             container_image: nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc20
+          - example: gpu_vllm
+            timeout: 15
+            container_image: docker.io/vllm/vllm-openai:v0.24.0
```
kinjalpatel27: can we upgrade it to 0.26.0?

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -559,35 +634,33 @@ def _invoke_fused_moe_quantized_function(
...
-        originals = {n: getattr(vllm_fused_moe_package, n) for n in _FUSED_MOE_KERNEL_FUNCS}
+    @contextmanager
+    def _fakequant_moe_kernels(self):
+        """Route vLLM's fused-MoE kernels through this module's quantizers.
+
+        They are module-level functions, so fakequant is installed by name for this forward.
+        Patching is process-wide: safe for the LIFO nesting vLLM does, but not thread-safe.
+        """
+        assert _FUSED_MOE_KERNEL_TARGETS, "No vLLM fused-MoE kernel entry point found to patch"
+        originals = [
```
kinjalpatel27, line 652: can we keep the comment for why not using `replace_function` here?

kinjalpatel27 (approved):
Generally LGTM. minor comments

.github/workflows/gpu_tests.yml
kevalmorabia97: Bumped to `v0.26.0` in `aa0c04ec4c`. Checked upstream first — v0.26.0 carries the same layout the plugin gates on (`def FusedMoE(...)` factory, `RoutedExperts` with `forward_modular`/`forward_monolithic`, `unquantized_fused_moe_method.py`, `experts/triton_moe.py`), so it takes the same code path as the `0.24.1.dev0` build in `nemo:26.08` that I validated locally.

Heads-up: I can only run the nemo:26.08 container here, so the 0.26.0 job is verified by this PR's CI rather than locally. If it goes red on drift unrelated to this fix, the safe fallback is `v0.24.0` (first release with this layout, closest to what nemo:26.08 ships).

modelopt/torch/quantization/plugins/vllm.py
kevalmorabia97: Restored in `aa0c04ec4c`, inside `_fakequant_moe_kernels`:

```python
# Patch by hand rather than with ``replace_function``: that context manager conflicts
# with torch.compile (same reason ``_VLLMParallelLinear.forward`` swaps quant_method).
```

The refactor into the context manager had dropped the commented-out `replace_function` block that carried this rationale; it is now stated directly next to the `setattr` loop.

kevalmorabia97:
Addressed the latest round in `aa0c04ec4c`.

**@kinjalpatel27** — both taken:
- CI image bumped to **`v0.26.0`**. Verified upstream first that it carries the same layout the plugin gates on (`def FusedMoE(...)` factory, `RoutedExperts` + `forward_modular`/`forward_monolithic`, `unquantized_fused_moe_method.py`, `experts/triton_moe.py`), so it exercises the same code path as the `0.24.1.dev0` build in `nemo:26.08`. I can only run the nemo container locally, so this job is verified by CI here; fallback if it drifts is `v0.24.0`.
- Restored the `replace_function` rationale next to the `setattr` loop in `_fakequant_moe_kernels` — the refactor into the context manager had dropped the commented-out block that carried it.

**@cjluo-nv** — the two open items from the re-review:
- The **`v0.20.0` job is green** on this PR, so the legacy branch plus the shared test-input changes (`tiny_llama` `head_dim=64`, the `causal=True` flip, and the new example-module import in `test_tiny_qwen3_moe_quantize`) are confirmed on the old layout. The new-layout job now runs 0.26.0 instead of 0.24.0.
- **CHANGELOG**: added under **0.46 Bug Fixes** (this fix ships in 0.46), including the `vllm_FusedMoE` → `vllm_RoutedExperts` key move and the `.routed_experts` path change that prevent restoring a `modelopt_state` saved under an older vLLM.

On the "owner should sign off on the vLLM internals" point: the evidence is a module-tree probe after `mtq.quantize` on tiny DeepSeek-V3 and Qwen3-MoE, where `w13_weight_quantizer`/`w2_weight_quantizer` come back with calibrated amax. Weight amax only appears if the patched kernel actually ran with `B is self.w13_weight` matching, which pins down both that `forward_modular`/`forward_monolithic` are the live entry points and that `experts.triton_moe` is the binding site. `SharedFusedMoE` needs no gate on 0.24+: the `shared_fused_moe` module does not exist there, so the existing import loop already leaves it `None`; shared experts are a plain MLP whose linears quantize through the parallel-linear registrations (also confirmed in the probe).

.github/workflows/gpu_tests.yml
```diff
@@ -50,6 +50,10 @@ jobs:
           - example: gpu_trtllm
             timeout: 15
             container_image: nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc20
+          - example: gpu_vllm
+            timeout: 15
+            container_image: docker.io/vllm/vllm-openai:v0.26.0
```
Edwardf0t1, line 56: The PR description is stale relative to this line, and I think the drift hides a decision worth making explicitly.

The description says CI pins `v0.24.0` (with a paragraph on why `v0.24.1` was never released, and a Testing section reported against `nemo:26.08.rc3` / vLLM `0.24.1.dev0`), but this pins **`v0.26.0`** — and the diff carries 0.26-specific test fixes (the MLA 128/64/128 dimension triple, `q_data_type_prefill`/`q_data_type_decode`, `FlashInferDecodeKernel`) that only make sense on 0.26.

Net effect: the matrix is now **0.20 + 0.26**, so **0.24 — the version in `nemo:26.08` that motivated this PR — has no CI coverage at all.** I think that's defensible: I checked and 0.24.0 and 0.26.0 share the same `RoutedExperts` layout (`routed_experts.py` first appears at 0.24; 404 on 0.20–0.23), so 0.26 is a reasonable proxy. But right now it reads as an accident of an un-refreshed description rather than a deliberate choice.

Could you refresh the description (and double-check the CHANGELOG wording, which is also framed entirely around 0.24) so the pinned version, the container the results came from, and the coverage story all agree?

Also a nit on line 56: the comment says "pre-0.24" while the entry above is now 0.26 — still accurate, but easy to misread as "the entry above is 0.24".

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -618,6 +695,25 @@ class _QuantVLLMSharedFusedMoE(_QuantFusedMoEBase):
         pass
 
 
+if _has_routed_experts_cls:
```
Edwardf0t1, line 703: `_has_fused_moe_cls` and `_has_routed_experts_cls` are treated as independent, but they describe two layouts that should be mutually exclusive.

They are, on every released version — I checked that `routed_experts.py` is 404 for 0.20 through 0.23 and present from 0.24, so this is not reachable today. But nothing in the code enforces it. If a future release (or a nightly caught mid-transition) ships a `FusedMoE` *class* alongside `RoutedExperts`, both registrations fire, and since `RoutedExperts` is a submodule of the `MoERunner` that `FusedMoE` returns, `mtq.quantize` converts parent *and* child. The parent then owns no `w13_weight`/`w2_weight`, so its `_invoke_fused_moe_quantized_function` falls through to `raise ValueError("Cannot determine first or second layer of expert")` — at serving time, which is exactly the class of late, confusing failure this PR is trying to eliminate.

Suggest making the branch structure encode the invariant, e.g. `elif _has_routed_experts_cls:` on the earlier block, or an `assert not (_has_fused_moe_cls and _has_routed_experts_cls)` next to the flag definitions. One line, and it matches the "fail at the registration site" philosophy the `dynamic.py` change establishes.

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -618,6 +695,25 @@ class _QuantVLLMSharedFusedMoE(_QuantFusedMoEBase):
...
+
+    @QuantModuleRegistry.register({RoutedExperts: "vllm_RoutedExperts"})
+    class _QuantVLLMRoutedExperts(_QuantFusedMoEBase):
+        """Expert-weight owner of the vLLM >= 0.24 ``MoERunner`` pipeline.
+
+        ``MoERunner`` calls ``forward_modular``/``forward_monolithic`` instead of ``__call__``,
+        so the inherited ``forward`` is never used (``RoutedExperts.forward`` raises by design).
```
Edwardf0t1: > so the inherited ``forward`` is never used (``RoutedExperts.forward`` raises by design)

The first half is true in practice, but the inherited `forward` isn't inert — it *shadows* the deliberate error. `_QuantFusedMoEBase.forward` has signature `(self, hidden_states, router_logits)`, whereas upstream is:

```python
def forward(self, *args, **kwargs) -> torch.Tensor:
    raise AssertionError("Call forward_modular or forward_monolithic instead.")
```

So after quantization, a stray `routed_experts(x)` raises `TypeError: forward() missing 1 required positional argument: 'router_logits'` instead of vLLM's explicit "Call forward_modular or forward_monolithic instead." — and if someone happens to call it with two positional args, they instead reach `super().forward(...)` and get the `AssertionError` from a confusing depth.

Cheap fix, and it makes the docstring literally true:

```python
def forward(self, *args, **kwargs):
    return RoutedExperts.forward(self, *args, **kwargs)
```

Minor, since nothing in vLLM takes this path — but it costs one line to keep the upstream diagnostic intact.

Edwardf0t1 (commented):
A few inline notes from a read-through. I verified the core claims against upstream vLLM sources (v0.20.0 / v0.23.0 / v0.24.0 / v0.26.0) and the diagnosis and fix placement both hold up — `FusedMoE` is a `def` in 0.24/0.26 and a `class` through 0.23, `triton_moe.py` really does bind `invoke_fused_moe_triton_kernel` at import time (so the extra patch target is necessary, not defensive), and `MoERunner` calls `routed_experts.forward_modular`/`forward_monolithic` as call-time attribute lookups, so the `__class__` swap does intercept. Comments below are the three items I'd want looked at before merge; none are blocking correctness on the versions in CI.

examples/vllm_serve/vllm_reload_utils.py
```diff
@@ -37,8 +37,14 @@
     restore_quantizer_state,
 )
 from modelopt.torch.quantization.nn import SequentialQuantizer, TensorQuantizer
+from modelopt.torch.quantization.plugins.vllm import _has_routed_experts_cls
 from modelopt.torch.quantization.utils import is_quantized
 
+# vLLM >= 0.24 moved the fused expert weights (and their quantizers) onto a ``routed_experts``
+# submodule of the MoE layer, so merged expert keys need that extra hop. Follow the plugin's
+# registration flag rather than re-probing, so keys always match the modules it converted.
+_EXPERTS_INFIX = ".routed_experts" if _has_routed_experts_cls else ""
```
mxinO, line 46: Maybe not related to this PR, but I can't find where is the `shared_experts` handled?

mxinO (commented):
LGTM, thanks.

.github/workflows/gpu_tests.yml
kevalmorabia97: Refreshed the description and CHANGELOG, and made the coverage story explicit rather than accidental.

CI is now **0.20.0 (legacy layout) + 0.26.0 (RoutedExperts layout)**, one container per layout. Your reading is right: 0.24 and 0.26 take identical code paths in the plugin, so 0.26 covers 0.24 while also catching newer upstream drift. 0.24 itself is covered by local runs — I ran the full suite in both `vllm/vllm-openai:v0.26.0` and `nemo:26.08.rc3` (0.24.1.dev0), 73 passed / 1 skipped on each. The description now says exactly that, and is quite a bit shorter.

CHANGELOG now reads "verified on 0.24 (as shipped in `nemo:26.08`) and 0.26". Matrix comments reworded by layout, so neither says "pre-0.24" next to a 0.26 entry.

modelopt/torch/quantization/plugins/vllm.py
kevalmorabia97: Fixed in `964f58c1bf`. `_has_fused_moe_cls` now requires `not _has_routed_experts_cls`, so if a release ever ships both, `RoutedExperts` wins.

I went with precedence rather than an assert because plugin import runs under `import_plugin`, which catches `Exception` and only warns — an assert there would silently disable the *entire* vLLM plugin (linear and attention quantization included) instead of just the ambiguous MoE case. Precedence keeps the invariant enforced at the registration site while degrading to the correct layout: `RoutedExperts` owns the expert weights, so registering its parent too would convert both and leave the parent hitting your `ValueError("Cannot determine first or second layer of expert")`.

modelopt/torch/quantization/plugins/vllm.py
kevalmorabia97: Taken in `964f58c1bf`:

```python
def forward(self, *args, **kwargs):
    # Keep vLLM's own "call forward_modular/forward_monolithic instead" error rather
    # than the inherited (hidden_states, router_logits) signature.
    return RoutedExperts.forward(self, *args, **kwargs)
```

You are right that the inherited `forward` was shadowing rather than inert. This is also why I declined an earlier suggestion to raise our own `RuntimeError` here — delegating keeps upstream's diagnostic verbatim, which is the part worth preserving.

examples/vllm_serve/vllm_reload_utils.py
kevalmorabia97: Good catch — it was not handled, and that was a real (pre-existing) bug. Fixed in `964f58c1bf`.

vLLM merges the shared expert's `gate_proj`/`up_proj` into `shared_experts.gate_up_proj` (confirmed on a booted tiny DeepSeek-V3: `model.layers.0.mlp.shared_experts.gate_up_proj [QuantMergedColumnParallelLinear]`), but the gate/up merge branch was guarded by `"experts" not in key` — and `"shared_experts"` contains `"experts"`. The per-expert regex needs `experts.<i>.`, so shared-expert keys matched neither branch and were copied through as `shared_experts.gate_proj.*`, which exists in HF but not in vLLM. Result: those keys were dropped at load and the shared experts served uncalibrated, silently.

The guard is now `not re.search(r"\\.experts\\.\\d+\\.", key)` — only *routed* experts are excluded — and `test_tiny_deepseek_mla_quantize` (`n_shared_experts=1`) asserts the mapped key resolves against a real quantizer module path on the booted model.

mxinO (approved):

shengliangxu (approved):

PR #2055, opened 2026-08-03
feat(megatron-bridge): Nemotron-Nano-3 W4A16 NVFP4 four_over_six PTQ/QAD support

yueshen2016:
### What does this PR do?

**Type of change:** Bug fix + new feature

Adds Megatron-Bridge support for W4A16 NVFP4 (`four_over_six`) PTQ and QAD on Nemotron-style
hybrid Mamba/attention MoE models, and fixes a ModelOpt bug found while doing so.

> **The bug is not Nemotron-specific.** Any model with an **untied `lm_head`** quantized through
> **Megatron-Bridge** silently gets a **BF16 output layer** instead of the quantized one the recipe
> asked for — no error, no warning.

---

## The bug

A recipe enabling `*output_layer*weight_quantizer` produced `lm_head.weight` as BF16
`[vocab, hidden]` instead of packed NVFP4. The *same recipe file* under Megatron-LM quantized it
correctly, so this was never a configuration problem. The result is a checkpoint that does not match
its own recipe, is larger than intended, and keeps the output projection — on the hot path for every
generated token — at full precision.

### Root cause

`_MegatronParallelLinear.sharded_state_dict()` special-cases `output_layer` and asks
**Megatron-LM's global args** whether embeddings are untied:

```python
from megatron.training import get_args as _mlm_get_args
_untied = bool(getattr(_mlm_get_args(), "untie_embeddings_and_output_weights", False))
except Exception:
    _untied = False          # falls back to "tied"
if not _untied:
    return super().sharded_state_dict(...)   # drops ALL quantizer state
```

**Megatron-Bridge has no global args store**, so the call raises and the handler concludes "tied" for
a model that is untied. Observed directly:

```
megatron.training.get_args() -> AssertionError: args is not initialized
model.share_embeddings_and_output_weights = False      # i.e. genuinely UNTIED
```

Fixing only that is **not sufficient**: the dist-checkpoint loader silently skips any checkpoint key
the model does not advertise, and `sharded_state_dict()` can only advertise a buffer that already
exists. On restore the quantizer exists and is enabled but its scale buffers do not, so the
calibrated values in the checkpoint have nowhere to land.

## The fix

1. **`_resolve_output_layer_untied(model)`** reads `share_embeddings_and_output_weights` off the
   model — Megatron-Core carries it under **both** frameworks (Bridge sets it from the HF config,
   Megatron-LM from `--untie-embeddings-and-output-weights`) — and records it on the config so
   `sharded_state_dict()` can consult it. `get_args()` remains the fallback, so **Megatron-LM
   behavior is unchanged**, and an unknown result still means "tied" (the conservative branch).
2. **Materialize missing weight-quantizer scale buffers before the load plan is built**, so the
   loader has somewhere to write. Two details that are easy to get wrong:
   - `_amax` must be allocated **flat** as `[numel // block, 1]`. `_process_quantizer_amax` exposes
     it to the checkpoint as an `[out_features, blocks]` **view over the same storage**, so the
     loader writes straight through. Allocating the *viewed* shape loads successfully but leaves the
     wrong in-memory rank, which then breaks the exporter's scale math.
   - `_global_amax` is registered with `register_buffer` directly: its property lives on
     `StaticBlockScaleQuantizer`, and on restore the module is still a plain `TensorQuantizer`.

### Export guard

`examples/megatron_bridge/export_quantized_megatron_to_hf.py` now **raises** instead of quietly
emitting BF16 when an enabled NVFP4 weight quantizer is missing either scale, naming the module and
the attribute. Static-block NVFP4 needs **both**; an `_amax`-only check let a half-restored quantizer
through and failed much later inside `NVFP4QTensor.quantize`, where `scale * scale_2` broadcasts
`[N, 1]` against `[N]` into an `N x N` allocation. `MODELOPT_ALLOW_UNCALIBRATED_NVFP4=1` restores the
previous behavior, reported rather than silent.

*A silent fallback is what let this ship unnoticed; making it loud is arguably the more important
half of the change.*

## Megatron-Bridge PTQ/QAD enablement

- **Teacher-build fix**: gate the non-grouped MoE spec to the **quantized student only**. It was
  previously applied to the BF16 teacher too, misplacing its MoE expert weights so the teacher
  predicted ground truth far worse than its own student, and QAD distilled toward a broken teacher.
- `--student_nongrouped_experts` (default **off**, so existing grouped behavior is preserved)
- SFT-masked distillation (`--sft`, `--sft_dataset_root`) — loss masked to completion tokens
- Student initialization from a Megatron checkpoint (Bridge exposes no hook, so
  `DistillationProvider.provide` is wrapped; marked TODO for upstream replacement)
- `--calib_random_offset`, matching Megatron-LM's `--calib-use-random-offset`

## Testing

Verified end to end on a hybrid Mamba/attention MoE model, against a **Megatron-LM-produced
reference checkpoint**:

| | before | after | Megatron-LM reference |
|---|---|---|---|
| `lm_head.weight` | BF16 `[131072, 2688]` | **`U8 [131072, 1344]`** | `U8 [131072, 1344]` |
| `lm_head.weight_scale` / `_scale_2` | absent | **present** | present |
| total keys | 18485 | **18487** | 18487 |
| `hf_quant_config` exclusions | 73 (incl. `lm_head`) | **72 (no `lm_head`)** | 72 (no `lm_head`) |

Full pipeline exercised: PTQ → QAD (200 iters, 0 NaN) → HF export → compressed-tensors conversion →
vLLM serving (coherent generation) → 4 accuracy benchmarks, with the quantized `lm_head` surviving
every stage. The fix also propagates through QAD: the QAD checkpoint now contains
`output_layer.weight_quantizer._amax` / `._global_amax`, which it previously lacked entirely.

**No new unit tests.** Reproducing this needs a Megatron-Bridge model, a quantized checkpoint and a
dist-checkpoint round-trip, which does not fit the existing unit-test harness. Guidance on where a
regression test belongs would be welcome — the highest-value assertion is that `sharded_state_dict()`
advertises the `output_layer` weight-quantizer keys for an untied model under Bridge.

- Is this change backward compatible?: ✅ — Megatron-LM keeps the `get_args()` path; the new flag
  defaults off; unknown tiedness still resolves to the previous conservative branch. The export guard
  is the one behavior change (silent BF16 → error), reversible via
  `MODELOPT_ALLOW_UNCALIBRATED_NVFP4=1`.
- If you copied code from any other sources or added a new PIP dependency: ✅ N/A — no copied code,
  no new dependencies.
- Did you write any new necessary tests?: ❌ — see **Testing** above.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ❌ —
  happy to add an entry; this is arguably a critical bug fix and probably warrants one.

### Additional Information

Two Megatron-Bridge compatibility shims were removed after being shown unnecessary rather than
assumed so: a `DistillationProvider.to_cfg_dict` monkeypatch (a 5-iteration distillation run trains
and checkpoints cleanly without it) and an `InferenceCudaGraphScope` enum stub for an unused
Megatron-LM-PTQ import path (zero occurrences across a full PTQ/QAD/export run).

## Summary by CodeRabbit

- **New Features**
  - Added support for distilling models initialized from Megatron checkpoints, including state restoration and Hugging Face export.
  - Added grouped MoE expert support and configurable random offsets for text calibration.
  - Added expanded QAD, SFT, and per-token loss configuration.
  - Added NVFP4 calibration-scale validation with an optional BF16 export fallback.

- **Bug Fixes**
  - Improved quantization-state and output-layer checkpoint restoration.
  - Prevented repeated quantizer conversion and improved handling of incomplete calibration data.
  - Improved checkpoint compatibility for models with tied or untied output weights.
  - Added stricter configuration validation and improved MTP checkpoint handling.

### DIFF
--- a/examples/megatron_bridge/distill.py
+++ b/examples/megatron_bridge/distill.py
@@ -25,15 +25,18 @@ import contextlib
 import os
 
 import torch
-from _distillation_provider import convert_to_distillation_provider
-from export_distilled_megatron_to_hf import export_llm_to_hf, save_vlm_to_hf
 from megatron.bridge import AutoBridge
+from megatron.bridge.models.distillation_provider import (
+    DistillationProvider,
+    convert_to_distillation_provider,
+)
 from megatron.bridge.recipes.utils.optimizer_utils import (
     distributed_fused_adam_with_cosine_annealing,
 )
 from megatron.bridge.training.config import (
     CheckpointConfig,
     ConfigContainer,
+    FinetuningDatasetConfig,
     GPTDatasetConfig,
     LoggerConfig,
     MockGPTDatasetConfig,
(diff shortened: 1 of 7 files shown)


yueshen2016:
Closing — this PR's content has been split into three focused PRs, and the parts of it that were needed have all landed elsewhere.

| what was here | where it went |
|---|---|
| `output_layer` tiedness fix (`plugins/megatron.py`) | #2112 |
| SFT-masked data support (`distill.py`) | #2113 |
| `--calib_random_offset`, MTP handling, uncalibrated-export guard | #2141 |
| `distill/plugins/megatron.py` NVFP4 promotion | **dropped** — MCore's two-pass restore already promotes `output_layer` (`quant_module_set_extra_state` → `maybe_promote_nvfp4_static_quantizer`); the block also overwrote the checkpoint's replicated `_global_amax` with a rank-local `reduce_amax` |
| `--grouped_experts` | **dropped** — obsolete since #1550 merged; `load_mbridge_model_from_hf` now defaults to `moe_grouped_gemm=True` on both the quantize and export sides |

The three replacements were verified together end to end on Nemotron-Nano-3.5: PTQ + QAD + export on main + #2112 + #2113 + #2141 reproduce the reference run (6382/862 quantizer counts, byte-identical `lm_head` export, iteration-10 KD loss within 0.085% at identical LR).

PR #2056, opened 2026-08-03
[NVBug: 6538278] Pass trust_remote_code to the TRT-LLM model load in deploy/eval examples

cjluo-nv:
### What does this PR do?

Type of change: Bug fix

Fixes [nvbug 6538278](https://nvbugspro.nvidia.com/bug/6538278).

`examples/hf_ptq/run_tensorrt_llm.py` loaded the **tokenizer** with `args.trust_remote_code` but constructed `LLM()` without it, so it defaulted to `False`. Deploying any checkpoint that ships custom modeling code (`auto_map`) — e.g. `Llama-3.3-Nemotron-Super-49B-v1` (DeciLM) — failed at executor init:

```
Failed to initialize executor on rank 0: The repository ... contains custom code which
must be executed to correctly load the model ... Please pass the argument trust_remote_code=True.
-> ValueError -> Executor worker returned error -> run_tensorrt_llm.py subprocess exit 1
```

The `modelopt.deploy.llm.LLM` wrapper already accepts and forwards `trust_remote_code` (`modelopt/deploy/llm/generate.py:153`), and `scripts/huggingface_example.sh` already forwards `--trust_remote_code` to the script — only the call site dropped it.

The same omission exists in the sibling TRT-LLM deploy paths driven by the same launcher and the same checkpoints, so they are fixed together:

| File | Fix |
| --- | --- |
| `examples/hf_ptq/run_tensorrt_llm.py` | the reported bug |
| `examples/llm_eval/lm_eval_tensorrt_llm.py` | `LLM()` ignored the `trust_remote_code` that lm-eval injects into `model_args` |
| `examples/llm_eval/mmlu.py` | `LLM()` ignored it (the tokenizer already used it) |
| `examples/hf_ptq/scripts/huggingface_example.sh` | the `mmlu` stage never forwarded the flag at all |

All four propagate the **user-provided** flag; none hardcode `trust_remote_code=True` (per SECURITY.md).

### Usage

No API change. Existing flag now takes effect on the model load:

```bash
scripts/huggingface_example.sh --model <Llama-3.3-Nemotron-Super-49B-v1> \
    --quant fp8 --tasks quant --trust_remote_code
```

### Testing

- New CPU regression test `tests/examples/hf_ptq/test_run_tensorrt_llm.py` stubs the `tensorrt_llm`-dependent import and asserts both the tokenizer **and** the model load receive the flag, parametrized over `True`/`False`. Confirmed it fails without the fix (`KeyError: 'trust_remote_code'`) and passes with it.
- Verified against the real `fire` package that a bare `--trust_remote_code` maps to `True` in `mmlu.py`'s `**kwargs`, and is absent (defaulting to `False`) when not passed.
- Simulated the launcher's argument assembly both ways: with `TRUST_REMOTE_CODE=false` the emitted commands are byte-identical to before this change.
- `bash -n` on the launcher; full pre-commit clean.
- GPU deploy of the `fp8` DeciLM checkpoint was verified by the bug reporter on GB10 with this one-line change (model loaded, TRT engine built, generation succeeded, deploy exit 0). The `mmlu` / `lm_eval` stages are not GPU-verified here.

### Additional Information

nvbug 6538278 / OMNIML-5659. Reported against modelopt 0.46.0rc0 on GB10 / DGX Spark (`tensorrt-llm/release:1.3.0rc22`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added consistent handling of the `trust_remote_code` setting across TensorRT-LLM inference and MMLU evaluation workflows.
  * Ensured tokenizer and model loading receive the configured remote-code behavior.
* **Tests**
  * Added coverage validating remote-code settings and preserving KV-cache behavior when context logits are requested.

### DIFF
--- a/examples/hf_ptq/run_tensorrt_llm.py
+++ b/examples/hf_ptq/run_tensorrt_llm.py
@@ -73,6 +73,7 @@ def run(args):
         tokenizer=tokenizer,
         max_batch_size=len(input_texts),
         enable_kv_cache_reuse=False,
+        trust_remote_code=args.trust_remote_code,
     )
     torch.cuda.cudart().cudaProfilerStart()
     outputs = llm.generate_text(input_texts, args.max_output_len)
--- a/examples/hf_ptq/scripts/huggingface_example.sh
+++ b/examples/hf_ptq/scripts/huggingface_example.sh
@@ -314,6 +314,10 @@ if [[ $TASKS =~ "mmlu" ]]; then
         mmlu_flags+=" --limit $MMLU_LIMIT "
     fi
 
+    if $TRUST_REMOTE_CODE; then
+        mmlu_flags+=" --trust_remote_code "
+    fi
+
     python mmlu.py \
         --model_name causal \
         --model_path $MODEL_ABS_PATH \
--- a/examples/llm_eval/lm_eval_tensorrt_llm.py
+++ b/examples/llm_eval/lm_eval_tensorrt_llm.py
@@ -68,6 +68,7 @@ class TRTLLM(TemplateAPI):
             # logits only for the recomputed suffix on shared-prefix requests (e.g. hellaswag),
             # truncating context_logits and breaking parse_logprobs. Disable it.
             enable_kv_cache_reuse=False,
+            trust_remote_code=bool(kwargs.get("trust_remote_code", False)),
         )
         self.max_length = max_length - 1
         logger.info("Loaded TRT-LLM")
--- a/examples/llm_eval/mmlu.py
+++ b/examples/llm_eval/mmlu.py
@@ -284,6 +284,7 @@ def main(
             medusa_choices=medusa_choices,
             max_seq_len=MAX_SEQ_LEN,
             max_batch_size=1,
+            trust_remote_code=kwargs.get("trust_remote_code", False),
         )
     else:
         model = select_model(
--- a/tests/examples/hf_ptq/test_run_tensorrt_llm.py
+++ b/tests/examples/hf_ptq/test_run_tensorrt_llm.py
@@ -1,0 +1,93 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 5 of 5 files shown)


meenchen (approved):

Small, focused bug fix: `--trust_remote_code` now reaches `modelopt.deploy.llm.LLM` (not just the tokenizer) in `run_tensorrt_llm.py`, `mmlu.py`, and `lm_eval_tensorrt_llm.py`, plus the launcher now forwards the flag to the `mmlu` stage.

Verification I did:
- `modelopt/deploy/llm/generate.py::LLM.__init__` does accept `trust_remote_code: bool = False` and forwards it to the TRT-LLM pytorch backend, so all three call sites pass a valid kwarg (and `mmlu.py`'s call keeps every other existing kwarg valid).
- `parser.sh` defines `TRUST_REMOTE_CODE=false` and `--trust_remote_code ) TRUST_REMOTE_CODE=true`, so the new `if $TRUST_REMOTE_CODE; then mmlu_flags+=...` block matches the existing idiom used for `PTQ_ARGS`, `RUN_ARGS`, and `lm_eval_flags` — with the default (`false`) the emitted command is unchanged, so this is backward compatible.
- `fire` maps a bare `--trust_remote_code` to `True` in `main(**kwargs)`; the shell path always passes `--checkpoint_dir`, so the `select_model` branch is unaffected.
- lm-eval's `--trust_remote_code` appends `trust_remote_code=True` to `model_args` and `handle_arg_string` converts it to a real bool, so the new `kwargs.get(...)` read works. None of the changes hardcode `True`.
- New test file's license header matches the canonical `LICENSE_HEADER` exactly (2026, "& AFFILIATES"), so no licensing concern.
- Test is CPU-only and consistent with existing precedent in the same directory (`test_hf_ptq_args.py`, `test_example_utils.py`), asserts both tokenizer and model receive the flag, and pins `enable_kv_cache_reuse=False` (needed for `generate_context_logits`).

Minor nits, none blocking:
- `bool(kwargs.get("trust_remote_code", False))` in `lm_eval_tensorrt_llm.py` is inconsistent with the plain `kwargs.get(...)` in `mmlu.py`, and `bool("false") is True` — harmless given lm-eval's arg parsing already coerces to bool, but the wrapper buys nothing and could mislead for a security-sensitive flag.
- The test could reuse the existing `_test_utils/examples/hf_ptq_example_utils.py` sys.path shim pattern; also `_RecordingLLM.last` is never reset and `run_tensorrt_llm` (with the stubbed `LLM`) stays in `sys.modules` after the fixture tears down, which could leak into a later test in the same session.
- The `livecodebench`/`simple_eval` stage deploys via `trtllm-serve $SAVE_PATH` and still doesn't forward `--trust_remote_code`, so the same DeciLM-style failure remains there; worth a follow-up if that path is in scope.
- Only the `run_tensorrt_llm.py` change is covered by a test; the `mmlu.py`/`lm_eval` one-liners are not, which is reasonable for their size.

PR #2059, opened 2026-08-03
Fix FSDP2 handling for tied embeddings

realAsma:
### What does this PR do?

Type of change: Bug fix

Groups distinct input and output embedding modules that share a weight into one FSDP2 parameter group before Hugging Face Trainer and Accelerate apply their normal auto-wrapping. This prevents PyTorch from rejecting the shared parameter when distillation prepares both the student and teacher models.

The path is limited to FSDP2 models with genuinely tied input/output weights; untied models and other distributed backends are unchanged.

### Usage

No user-facing API change is required. Existing FSDP2 distillation commands work without additional configuration.

### Testing

- `pytest_pwd tests/unit/torch/opt/plugins/test_transformers_fsdp.py tests/unit/torch/distill/plugins/test_huggingface_kd.py -q` (6 passed)
- Changed-file pre-commit suite (passed)
- Two-rank, one-step offline distillation in `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc22` with PyTorch 2.12.0a0, Accelerate 1.14.0, and Transformers 5.5.4:
  - Baseline reproduced the duplicate tied-parameter error.
  - The production change completed successfully with the standalone workaround disabled.

### Additional Information

The failure occurs before the first training step when FSDP2 separately wraps the two modules that own a tied embedding weight.

### DIFF
--- a/modelopt/torch/opt/plugins/transformers.py
+++ b/modelopt/torch/opt/plugins/transformers.py
@@ -64,6 +64,44 @@ def is_liger_available():
     return True
 
 
+def _fully_shard_tied_embeddings(model, accelerator):
+    """Put tied input/output embeddings in the same FSDP2 parameter group.
+
+    Accelerate otherwise visits the two owner modules independently. PyTorch rejects
+    the second ``fully_shard`` call because their shared weight is already managed by
+    the first group.
+    """
+    if not getattr(accelerator, "is_fsdp2", False):
+        return
+
+    input_embedding = getattr(model, "get_input_embeddings", lambda: None)()
+    output_embedding = getattr(model, "get_output_embeddings", lambda: None)()
+    if (
+        input_embedding is None
+        or output_embedding is None
+        or input_embedding is output_embedding
+        or getattr(input_embedding, "weight", None) is not getattr(output_embedding, "weight", None)
+    ):
+        return
+
+    from torch.distributed.fsdp import FSDPModule, MixedPrecisionPolicy, fully_shard
+
+    if isinstance(input_embedding, FSDPModule) or isinstance(output_embedding, FSDPModule):
+        return
+
+    fsdp_plugin = accelerator.state.fsdp_plugin
+    mesh = getattr(accelerator, "torch_device_mesh", None)
+    fully_shard(
+        [input_embedding, output_embedding],
+        reshard_after_forward=fsdp_plugin.reshard_after_forward,
+        offload_policy=fsdp_plugin.cpu_offload,
+        mp_policy=fsdp_plugin.mixed_precision_policy or MixedPrecisionPolicy(),
+        mesh=(
+            mesh[tuple(accelerator.parallelism_config.fsdp_dim_names)] if mesh is not None else None
+        ),
+    )
+
+
 @contextmanager
 def _undo_torch_init_override_by_transformers():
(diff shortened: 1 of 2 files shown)


realAsma:
BB: this does not look like a ModelOpt side error to me. This is related to Huggingface and PyTorch. Can you check or search how Huggingface handles tied weights with FSDP2 and report here?

realAsma:
> _:robot: Bot comment._

🐝 I’m checking how Hugging Face handles tied weights with FSDP2 and will report back here.

realAsma:
> _:robot: Bot comment._

Checked upstream handling. This is not a ModelOpt-originated error:

- PyTorch FSDP2 deliberately requires shared/tied parameters to be assigned to the same `fully_shard` instance; its RFC states that shared parameters must be in the same group, and the runtime error prescribes `fully_shard([module_a, module_b])`: https://github.com/pytorch/pytorch/issues/114299#issue-2018305003
- Accelerate's current FSDP2 preparation walks auto-wrap matches bottom-up and then calls `fully_shard(model)`; it does not detect and group the distinct input/output embedding modules that own one shared parameter: https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/fsdp_utils.py
- Hugging Face's recent tied-embedding fix addresses a different case: the config says tied but the checkpoint contains separate weights. Their guidance there is `tie_word_embeddings=False`: https://github.com/huggingface/accelerate/issues/3870 and https://github.com/huggingface/accelerate/pull/3878
- Our reproducer has genuinely tied weights (the two attributes are the same `Parameter`), so disabling tying would change the model and is not the right remedy.

Conclusion: the grouping in this PR matches PyTorch's required workaround, but it is compensating for an upstream Accelerate/FSDP2 integration gap. I recommend filing/fixing this in Accelerate; we should keep the ModelOpt workaround only if we explicitly want a compatibility bridge until an upstream release contains the fix.

realAsma:
> _:robot: Bot comment._

Confirmed: this is an upstream Hugging Face/PyTorch integration issue, not ModelOpt-owned behavior.

Accelerate's FSDP2 preparation independently calls `fully_shard(module, ...)` for each auto-wrap match before sharding the root ([source](https://github.com/huggingface/accelerate/blob/f1e6505b9025739917df26ffb47456ae947b35d4/src/accelerate/utils/fsdp_utils.py#L711-L718)). PyTorch explicitly rejects a tied/shared parameter split across FSDP groups and directs callers to `fully_shard([module_a, module_b])` ([source](https://github.com/pytorch/pytorch/blob/b4a129a422c02f3a9df68c50dd04728a5475e531/torch/distributed/fsdp/_fully_shard/_fsdp_state.py#L213-L222)). The workaround in this PR follows that PyTorch prescription.

Hugging Face also recently merged native FSDP2 handling that rewrites the input-embedding/output-head plan when they are tied before applying `fully_shard` ([transformers#46707](https://github.com/huggingface/transformers/pull/46707)); see `_resolve_tied_embed_lm_head_plan`). That is further evidence the ownership belongs in the HF distributed integration layer.

Recommendation: report/fix the legacy Trainer→Accelerate FSDP2 path upstream. I would not merge a permanent ModelOpt-specific policy copy unless we deliberately want a version-gated compatibility workaround for affected HF releases. No code changes prepared in this pass.

PR #2060, opened 2026-08-04
Preserve HF PTQ checkpoint sidecar files [NV BUG 6491822]

jenchen13:
### What does this PR do?

Type of change: Bug fix

In `hf_ptq.py` when exporting a PTQ checkpoint, it would drop some files from the original BF16 checkpoint because it uses a whitelist pattern to allow certain files. However that is brittle and can drop files such as reasoning parsers. 

Now we make hf_ptq.py match Megatron-Core export behavior by copying all non-safe tensor files, but filter only allowed non-safetensor files for more safety.

### Usage

```python
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Hugging Face checkpoint handling to preserve eligible sidecar files while excluding weights, indexes, stale quantization metadata, and unsupported artifacts.
  * Preserved existing export files and applied consistent file filtering.
  * Improved snapshot resolution when remote code is disabled.
  * Ensured unified exports handle generation configuration files correctly.

* **Tests**
  * Added coverage for sidecar copying, exclusions, existing-file preservation, supported file patterns, and snapshot downloads.


examples/hf_ptq/example_utils.py
```diff
@@ -56,6 +57,25 @@
 
 SPECULATIVE_MODEL_LIST = ["Eagle", "Medusa"]
 
+_EXPORT_OWNED_CHECKPOINT_FILES = {
+    "config.json",
+    "hf_quant_config.json",
+}
+_CHECKPOINT_WEIGHT_FILE_PATTERNS = (
+    "*.safetensors",
```
cjluo-nv, line 80: > _Bot comment._

`*.gguf` is missing from the weight-file pattern list (as are `*.npz`, `*.pkl`, `*.mlmodel`, `*.tar*`/`*.zip` archives). Many HF repos ship GGUF conversions alongside safetensors; with the blanket copy these get copied into the export dir, and via `_resolve_model_path` they'd also be *downloaded* (see comment on the `snapshot_download` call). Please extend the list.

examples/hf_ptq/example_utils.py
```diff
@@ -56,6 +57,25 @@
 
 SPECULATIVE_MODEL_LIST = ["Eagle", "Medusa"]
 
+_EXPORT_OWNED_CHECKPOINT_FILES = {
+    "config.json",
```
cjluo-nv, line 105: > _Bot comment._

`hf_ptq.py` explicitly supports already-quantized sources (pack-quantized/compressed-tensors and MXFP4 gpt-oss). Those checkpoints often carry stale quant metadata sidecars — `recipe.yaml` (llm-compressor), `quantize_config.json` (AutoAWQ/GPTQ), `quant_config.json` — which the new blanket copy will drop next to the freshly written `hf_quant_config.json`, potentially confusing vLLM/SGLang scheme detection. Consider adding those to `_EXPORT_OWNED_CHECKPOINT_FILES` (or a separate "stale quant metadata" skip set).

examples/hf_ptq/example_utils.py
```diff
@@ -895,11 +915,13 @@ def _resolve_model_path(model_name_or_path: str, trust_remote_code: bool = False
             try:
                 local_path = snapshot_download(
                     repo_id=model_name_or_path,
-                    allow_patterns=["*.py", "*.json"],  # Only download Python files and config
+                    ignore_patterns=list(_CHECKPOINT_WEIGHT_FILE_PATTERNS),
                 )
```
cjluo-nv, line 981: > _Bot comment._

This changes an intentionally tiny fetch (`allow_patterns=["*.py", "*.json"]`) into "download every file in the repo that isn't a recognized weight file". For HF-ID sources that ship GGUF conversions, images/videos, or archives, this can be many GB downloaded at export time (and `*.gguf` isn't even in the ignore list). Suggest keeping an allow-list here (or adding a size/pattern guard) and broadening only the local copy step, which is what the bug is actually about.

examples/hf_ptq/example_utils.py
```diff
@@ -997,29 +1006,22 @@ def copy_custom_model_files(source_path: str, export_path: str, trust_remote_cod
...
-                if file_path.name in copied_files:  # e.g. matched by both pattern lists
-                    continue
-                dest_path = export_dir / file_path.name
-                try:
-                    shutil.copy2(file_path, dest_path)
-                    copied_files.append(file_path.name)
-                    print(f"Copied custom model file: {file_path.name}")
-                except Exception as e:
-                    print(f"Warning: Failed to copy {file_path.name}: {e}")
+    for file_path in sorted(source_dir.iterdir()):
```

This reimplements an existing repo helper: `modelopt/torch/export/plugins/hf_checkpoint_utils.py::copy_non_safetensor_files_from_ckpt` already does "copy every non-safetensors file from a local HF checkpoint dir verbatim" and is the Megatron-Core path this PR says it wants to match (its docstring even documents which files modelopt owns). Since `examples/hf_ptq` already imports from `modelopt.torch.export`, consider calling it (extended with an `exclude`/extra-pattern argument for `hf_quant_config.json` + the broader weight patterns) instead of maintaining a second, subtly different copy routine.

examples/hf_ptq/example_utils.py
```diff
@@ -934,49 +956,36 @@ def _resolve_model_path(model_name_or_path: str, trust_remote_code: bool = False
...
-    # Executable custom model/config code + other custom JSON: only used with trust_remote_code.
-    code_patterns = [
-        "configuration_*.py",
-        "modeling*.py",
-        "tokenization_*.py",
-        "*.json",
-    ]
-
     # Resolve the source path (handles both local paths and HF model IDs)
     resolved_source_path = _resolve_model_path(source_path, trust_remote_code)
```
cjluo-nv, line 1047: > _Bot comment._

The old docstring deliberately gated executable code (`modeling*.py`, `configuration_*.py`, `tokenization_*.py`) behind `trust_remote_code`; after this change those files are copied into the export unconditionally and `trust_remote_code` only affects HF-ID resolution. That's likely intentional given the goal, but please call it out in the PR body (it's a behavior change relevant to SECURITY.md guidance) and add a test pinning the new parity between `trust_remote_code=True/False`.

tests/examples/hf_ptq/test_example_utils.py
```diff
@@ -49,6 +49,54 @@ def _write_safetensors(path, tensors):
...
+        "generation_config.json",
+    ]:
+        assert (export_dir / file_name).read_text() == source_files[file_name]
+
+    assert (export_dir / "config.json").read_text() == '{"export": "config"}\n'
+    assert (export_dir / "hf_quant_config.json").read_text() == '{"export": "quant"}\n'
+    assert not (export_dir / "model.safetensors.index.json").exists()
+    assert not (export_dir / "model-00001-of-00001.safetensors").exists()
+    assert not (export_dir / "pytorch_model.bin").exists()
+    assert not (export_dir / "pytorch_model.bin.index.json").exists()
```
cjluo-nv: > _Bot comment._

Good happy-path test. Missing cases worth adding: (a) subdirectories in the source (e.g. `original/`) are silently skipped — pin that so the loss of nested sidecars is a deliberate decision; (b) `trust_remote_code=False` now copies `modeling_*.py`, which is the actual behavior change; (c) the `_resolve_model_path` ignore-pattern change (mock `snapshot_download` and assert what is/ isn't requested), since that's the riskiest part of the diff.

cjluo-nv (commented):

The direction (blanket-copy non-weight sidecars instead of a brittle whitelist) is right and the new test pins the local-dir happy path. A few things to address before merge:

1. **Duplicated logic** — the repo already has exactly this helper: `modelopt/torch/export/plugins/hf_checkpoint_utils.py::copy_non_safetensor_files_from_ckpt` (that *is* the Megatron-Core behavior the PR body cites, and its docstring already documents the "modelopt owns config.json / generation_config.json / hf_quant_config.json / preprocessor_config.json" convention). `examples/hf_ptq` already imports freely from `modelopt.torch.export`, so consider reusing/extending it (e.g. add an `exclude`/extra-patterns arg) rather than adding a second, slightly different implementation that will drift from it.

2. **`_resolve_model_path` now downloads far more than before** — swapping `allow_patterns=["*.py", "*.json"]` for `ignore_patterns=<weight patterns>` turns a few-KB metadata fetch into "download everything that isn't a recognized weight file". `*.gguf` (and `*.npz`, `*.pkl`, `*.tar*`, `*.zip`, media assets) are not in the ignore list, so an HF-ID source can now pull multi-GB artifacts at export time. Please either extend the pattern list or keep an allow-list for the *download* path while broadening only the local *copy* path.

3. **`trust_remote_code` no longer gates copying executable `*.py`** — the removed docstring explicitly justified that gating; now `modeling*.py`/`configuration_*.py`/`tokenization_*.py` are copied into the export even for native (non-remote-code) loads, and the argument only affects path resolution. That may well be intentional, but it's a behavior/security-posture change that isn't mentioned in the PR body or covered by a test.

4. **Stale quantization metadata can now leak** — `hf_ptq.py` explicitly supports already-quantized sources (pack-quantized/compressed-tensors, MXFP4). Those checkpoints ship files like `recipe.yaml` / `quantize_config.json` / `quant_config.json`, which will now be copied verbatim next to a freshly written NVFP4/FP8 `hf_quant_config.json` and can confuse deployment stacks. Worth adding to `_EXPORT_OWNED_CHECKPOINT_FILES` (and deciding explicitly whether `generation_config.json`/`preprocessor_config.json` should be source-wins or export-wins, since the Megatron helper treats them as modelopt-owned).

5. **Test coverage** is limited to the local-directory happy path: no coverage for the `snapshot_download` ignore-pattern change, for subdirectory skipping, or for the `trust_remote_code=True/False` parity that just changed.

examples/hf_ptq/example_utils.py
```diff
@@ -56,6 +59,25 @@
...
+    "*.json",
+    "*.md",
+    "*.model",
+    "*.py",
+    "*.tiktoken",
+    "*.txt",
+]
+_HF_PTQ_EXPORT_OWNED_FILES = {
+    "config.json",
+    "generation_config.json",
```

Unresolved from the previous round: `generation_config.json` is excluded for **both** callers, but only the unified-HF path writes one (`save_pretrained`). On the TRT-LLM path, `export_tensorrt_llm_checkpoint` writes only `config.json` (and `quant_cfg.json` for auto-quant) — there is no `generation_config` write anywhere in `model_config_export.py`. Before this PR the file was copied here via the old `"*.json"` pattern under `--trust_remote_code`; now it is neither copied nor written, so the export silently loses the source sampling defaults.

Compare `unified_export_megatron.py`, which calls the same helper with no `exclude_files` and then explicitly re-saves `GenerationConfig.from_pretrained(...)` — i.e. it guarantees the file exists before superseding the source copy.

Suggest either passing the exclusion set in from the caller (so the TRT-LLM branch keeps `generation_config.json`) or writing a fresh `generation_config.json` on that path. Excluding `config.json` on the TRT-LLM path *is* correct (different schema), so this needs to be per-caller rather than one shared set.

examples/hf_ptq/example_utils.py
```diff
@@ -56,6 +59,25 @@
 
 SPECULATIVE_MODEL_LIST = ["Eagle", "Medusa"]
 
+_HF_SIDECAR_DOWNLOAD_ALLOW_PATTERNS = [
```
cjluo-nv, line 68: > _Bot comment._

The local-dir path is now blanket-copy-minus-exclusions, but the HF-model-ID path is still bounded by this extension allow-list, so the two paths disagree about what counts as a sidecar. Notably missing: `chat_template.jinja` (transformers ≥ 4.51 saves chat templates as a standalone file, and many current hub repos ship one) and extensionless `LICENSE`/`NOTICE` — the shared helper's own docstring names `LICENSE` as something it exists to preserve. Note the new test's `accuracy_chart.png` assertion can only ever hold for a local source dir, so it reads as broader coverage than it gives.

At minimum add `"*.jinja"`, `"LICENSE*"`, `"NOTICE*"`; ideally derive both paths from one source of truth (e.g. `ignore_patterns=HF_CHECKPOINT_WEIGHT_FILE_PATTERNS` for the download too, which already covers the bulk-artifact over-download concern) and pin download patterns vs. copy exclusions in a test so they can't drift.

tests/examples/hf_ptq/test_example_utils.py
```diff
@@ -49,6 +49,114 @@ def _write_safetensors(path, tensors):
...
+def test_copy_custom_model_files_preserves_python_sidecars_with_trust_remote_code(tmp_path):
+    source_dir = tmp_path / "source"
+    export_dir = tmp_path / "export"
+    source_dir.mkdir()
+    export_dir.mkdir()
+    (source_dir / "super_v3_reasoning_parser.py").write_text("parser\n")
+
+    example_utils.copy_custom_model_files(str(source_dir), str(export_dir), trust_remote_code=True)
+
+    assert (export_dir / "super_v3_reasoning_parser.py").read_text() == "parser\n"
```
cjluo-nv: > _Bot comment._

This test can't observe the flag it's named after: `copy_custom_model_files` only forwards `trust_remote_code` to `_resolve_model_path`, which returns immediately for a local directory. So it's an exact duplicate of the `.py`-copy assertion in the preceding test.

That matters because the `trust_remote_code` gate on executable `*.py` is precisely what this PR changes — `modeling*.py`/`configuration_*.py`/`tokenization_*.py` are now copied for native (non-remote-code) loads, whereas the removed docstring explicitly justified gating them. Either drop this test, or parametrize `trust_remote_code=[True, False]` over a `modeling_custom.py` fixture and assert the (now intentional) equality so the change is pinned as deliberate. Please also call the behavior change out in the PR body — it's a security-posture change relative to SECURITY.md guidance.

modelopt/torch/export/plugins/hf_checkpoint_utils.py
```diff
@@ -253,25 +280,48 @@ def load_multimodal_components(
...
+    copied_files = []
     os.makedirs(dst, exist_ok=True)
-    for entry in os.listdir(src):
+    for entry in sorted(os.listdir(src)):
         sp = os.path.join(src, entry)
         if not os.path.isfile(sp):
             continue
         if entry.endswith(".safetensors") or entry == "model.safetensors.index.json":
             continue
+        if entry in exclude_files or _matches_any_pattern(entry, exclude_patterns):
```
cjluo-nv: > _Bot comment._

The old `copy_custom_model_files` wrapped each `shutil.copy2` in `try/except` and warned per file. The new loop doesn't, so a single unreadable entry (bad permissions, dangling HF-cache blob symlink) aborts the whole sidecar copy — and since this runs *after* the quantized weights are written, the user gets a traceback on an otherwise-complete export. A per-file `try/except` + warning would preserve the previous resilience.

Also, the hardcoded `entry.endswith(".safetensors") or entry == "model.safetensors.index.json"` check is now a second source of truth: for the Megatron caller (no `exclude_patterns`) it still misses index files not named exactly `model.safetensors.index.json`. Consider making `HF_CHECKPOINT_WEIGHT_FILE_PATTERNS` the always-applied baseline and keeping `exclude_patterns` purely additive.

modelopt/torch/export/plugins/hf_checkpoint_utils.py
```diff
@@ -29,6 +31,31 @@
...
+    "*.pkl",
+    "*.pt",
+    "*.pth",
+    "*.tar",
+    "*.tar.bz2",
+    "*.tar.gz",
+    "*.tar.xz",
+    "*.tflite",
+    "*.tgz",
+    "*.zip",
```
cjluo-nv: > _Bot comment._

`HF_CHECKPOINT_WEIGHT_FILE_PATTERNS` is a new public name in a module star-exported via `plugins/__init__.py`, and this module has no `__all__` (so the star-import also re-exports `fnmatch`/`json`/`os`/`shutil`/`torch`/...). Per CONTRIBUTING.md, declare the public API with `__all__`, or make the constant private (`_HF_...`) if `examples/hf_ptq` is the only intended consumer — otherwise it becomes an API surface with a compat obligation.

Minor naming nit: `*.npy`/`*.npz`/`*.pkl`/`*.zip`/`*.tar*` aren't HF *weight* formats; some legitimate sidecars are `.npy` (precomputed stats) and will now be silently dropped. Something like `..._LARGE_ARTIFACT_PATTERNS` would describe the intent more honestly.

cjluo-nv (commented):

Re-review of the sidecar-preservation change. Good progress since the last round — most of the critical feedback is now addressed:

**Resolved**
- *Duplicated logic*: `copy_custom_model_files` now delegates to `modelopt/torch/export/plugins/hf_checkpoint_utils.py::copy_non_safetensor_files_from_ckpt` with new keyword-only `exclude_files`/`exclude_patterns` args. The Megatron caller (`unified_export_megatron.py`) passes neither, so its behavior is unchanged — signature extension is backward compatible.
- *Over-download on the HF-ID path*: `snapshot_download` is back on an allow-list (`_HF_SIDECAR_DOWNLOAD_ALLOW_PATTERNS`) and is pinned by `test_resolve_model_path_snapshot_download_stays_allowlisted`.
- *Stale quant metadata*: `recipe.yaml/yml`, `quantize_config.json`, `quant_config.json`, `quantization_config.json` are in `_HF_PTQ_EXPORT_OWNED_FILES` and asserted absent in the new test.
- *GGUF / other bulk artifacts*: covered by `HF_CHECKPOINT_WEIGHT_FILE_PATTERNS` and asserted in the test, which also pins subdirectory skipping.

**Still unresolved from the previous round**
1. `generation_config.json` is excluded unconditionally, but only the unified-HF caller writes one. On the TRT-LLM path `export_tensorrt_llm_checkpoint` writes only `config.json` (+ `quant_cfg.json`) — I confirmed there is no `generation_config` write in `model_config_export.py` — so that file used to be copied (via the old `*.json` pattern under `--trust_remote_code`) and is now neither copied nor written. Silent loss of the source sampling defaults.
2. The `trust_remote_code` behavior change (executable `modeling*.py`/`configuration_*.py` now copied for native loads too) is still not mentioned in the PR body, and `test_copy_custom_model_files_preserves_python_sidecars_with_trust_remote_code` cannot observe the flag (local dir → `_resolve_model_path` returns at its first branch), so it duplicates the previous test under a misleading name. Two prior reviewers flagged this.
3. The HF-ID allow-list omits `*.jinja` (standalone `chat_template.jinja` on transformers ≥ 4.51) and extensionless `LICENSE`/`NOTICE`, so the two code paths still disagree on what a sidecar is — the exact class of bug this PR fixes survives for hub-ID sources.

Plus two smaller items: dropping the per-file `try/except` makes one unreadable file abort the whole sidecar copy after weights are written, and the new public `HF_CHECKPOINT_WEIGHT_FILE_PATTERNS` lands in a star-exported module with no `__all__`.

Direction and test quality are much improved; please resolve (1) and (2) before merge.

examples/hf_ptq/example_utils.py
jenchen13: fixed

cjluo-nv (commented):

Re-review #3 (head of `hf_ptq_sidecars`, 3 files / +177-68). The critical items from the previous rounds are all resolved in the current diff:

- **Duplicated logic** — `copy_custom_model_files` now delegates to `hf_checkpoint_utils.copy_non_safetensor_files_from_ckpt` with a keyword-only `exclude_files`; the Megatron caller passes none, so the signature change is backward compatible.
- **Over-download on the HF-ID path** — back on an allow-list (`_HF_SIDECAR_DOWNLOAD_ALLOW_PATTERNS`), now including `*.jinja` / `LICENSE*` / `NOTICE*`, and pinned by `test_resolve_model_path_snapshot_download_stays_allowlisted`.
- **Stale quant metadata / GGUF & bulk artifacts** — `recipe.yaml|yml`, `quantize_config.json`, `quant_config.json`, `quantization_config.json` in `_HF_PTQ_EXPORT_OWNED_FILES`; weights/GGUF/archives in `_HF_CHECKPOINT_WEIGHT_FILE_PATTERNS`; all asserted absent in the new test.
- **New public symbol without `__all__`** — constant renamed `_HF_CHECKPOINT_WEIGHT_FILE_PATTERNS` (private).
- **Second source of truth for weight skipping** — the hardcoded `.safetensors` / `model.safetensors.index.json` checks are gone; the pattern list is now the always-applied baseline, which also closes the Megatron index-file gap.
- **Misnamed `trust_remote_code` test** — dropped (commit d53d4fc), one of the two options both prior reviewers offered.

Two things I'd like the owner to sign off on rather than approve blind, plus small nits. No prompt-injection attempts observed in the PR content.

1. :speech_balloon: **`generation_config.json`** — author resolved the previously-flagged silent loss by removing it from `_HF_PTQ_EXPORT_OWNED_FILES` and documenting "source generation files intentionally still win" in the docstring, and the new test pins it. Still worth a human look because this also *flips* behavior on the unified-HF path: `save_pretrained` writes a validated `generation_config.json` and the source copy now overwrites it (previously it was export-wins). That's the deliberate choice the earlier review asked for, but it applies to both callers, so please confirm source-wins is intended for the unified-HF export too — and note the Megatron helper's own path (`unified_export_megatron.py`) does the opposite (re-saves `GenerationConfig.from_pretrained` *after* the copy).

2. **Shared-library behavior change is untested/unmentioned.** `copy_non_safetensor_files_from_ckpt` previously skipped only `*.safetensors` + `model.safetensors.index.json`; it now also skips `*.bin`, `*.pt`, `*.pth`, `*.ckpt`, `*.h5`, `*.msgpack`, `*.npy`, `*.npz`, `*.pkl/.pickle`, `*.onnx`, `*.pb`, `*.tflite`, and archives. The Megatron export caller passes no `exclude_patterns`, so its sidecar set shrank — mostly a fix (source weights no longer land in the export), but `*.npy`/`*.npz`/`*.pkl` are not always weights (precomputed stats/filters shipped next to a checkpoint) and would now be dropped silently. `tests/unit/torch/export/test_hf_checkpoint_utils.py` gained no coverage for this function (only the example-level test exercises it indirectly), and neither the PR body nor CHANGELOG mentions the library-side change.

Minor / owner's call:
- Per-file `try/except` + warning around `shutil.copy2` was dropped (flagged in two prior rounds, still open): one unreadable file or dangling HF-cache blob symlink now aborts the whole sidecar copy *after* the quantized weights are written.
- The download allow-list and the copy exclusion list remain two independent sources of truth for "what is a sidecar"; nothing ties them together in a test, so they can drift again.
- The new test no longer covers subdirectory skipping (an earlier round's review noted it as pinned); worth re-adding one `source/original/…` fixture assertion since nested sidecars are silently skipped by design.
- The `trust_remote_code` copy-parity change (executable `modeling*.py` / `configuration_*.py` now copied for native loads too) is documented in the new docstring but still not called out in the PR body, and the checkbox list in the description is left unfilled (Changelog / tests / BC).
- CodeRabbit's symlink-dereference findings look inapplicable here: HF cache snapshots are symlink-to-blob trees, so rejecting symlink sources would break the cache path, and `shutil.copy2` was already used before this PR. Fine to skip, but say so explicitly if you want the bot comments resolved.

cjluo-nv (commented):

Re-review #4 (head of `hf_ptq_sidecars`, 5 files / +234-72). Most of the previously-flagged items are now resolved — but this revision introduces a **new critical regression in the shared library helper**, and the accompanying new unit test appears to assert behavior the implementation no longer has (i.e. it should fail in CI).

**Blocking (new in this revision)**
- `copy_non_safetensor_files_from_ckpt` no longer skips anything by default: this diff deletes the `entry.endswith(".safetensors") or entry == "model.safetensors.index.json"` guard and replaces it *only* with caller-supplied `exclude_files`/`exclude_patterns`. The weight-pattern list moved out of the library into `examples/hf_ptq` (`_HF_PTQ_WEIGHT_FILE_PATTERNS`), so the other caller — `unified_export_megatron.py:GPTModelExporter.save_pretrained`, which passes no exclusions — will now copy the **source BF16 `*.safetensors` shards and `model.safetensors.index.json`** into the Megatron export directory, right before `save_safetensors_by_layer_index` writes its own shards/index. That both blows up export size and can leave stale source shards + a stale index next to the quantized ones (differing shard counts → broken/incorrect checkpoint on load). An earlier review round suggested making the pattern list the *always-applied baseline* with `exclude_patterns` purely additive; that is the fix here (or restore the hardcoded skip). The function name and docstring ("Copy every non-safetensors file…") are also now inaccurate.
- `tests/unit/torch/export/test_hf_checkpoint_utils.py::test_copy_non_safetensor_files_from_ckpt_supports_additional_exclusions` asserts `not (default_dst / "model.safetensors").exists()` for a default (no-exclusion) call — with the current implementation `model.safetensors` *is* copied, so this test should fail. Please confirm CI ran; the assertion encodes the behavior the library ought to have, which reinforces the point above.

**Resolved since re-review #3**
- `generation_config.json` is now per-caller: `hf_ptq.py` passes `exclude_files={"generation_config.json"}` only on the unified-HF path (where `save_pretrained` writes a validated one) and `None` on the TRT-LLM path (source-wins, restoring the previously-dropped file). That closes the long-standing item 1.
- Public constant → example-local private `_HF_PTQ_WEIGHT_FILE_PATTERNS` (no `__all__` obligation), and `*.npy/npz/pkl` are no longer imposed on the Megatron caller.
- Download allow-list includes `*.jinja` / `LICENSE*` / `NOTICE*` and is pinned by a test; stale quant metadata and GGUF/bulk artifacts are excluded and asserted.
- The misnamed `trust_remote_code` test was dropped.

**Still open / minor (owner's call)**
- Per-file `try/except` + warning around `shutil.copy2` is still absent (flagged in three rounds): one unreadable file or dangling HF-cache blob symlink aborts the whole sidecar copy *after* the quantized weights are written.
- Download allow-list vs. copy exclusion list remain two independent sources of truth, untied by any test.
- No coverage for subdirectory skipping (nested sidecars are silently dropped by design).
- PR body still doesn't mention the `trust_remote_code` copy-parity change (executable `modeling*.py`/`configuration_*.py` now copied for native loads) and the checkbox list (tests / changelog / BC) is unfilled.
- CodeRabbit's symlink-dereference findings look inapplicable (HF cache snapshots are symlink-to-blob trees); worth replying to close them out.

No prompt-injection attempts observed in the PR content.

modelopt/torch/export/plugins/hf_checkpoint_utils.py
```diff
@@ -253,25 +255,44 @@ def load_multimodal_components(
...
-    for entry in os.listdir(src):
+    for entry in sorted(os.listdir(src)):
+        if entry in exclude_files or _matches_any_pattern(entry, exclude_patterns):
+            continue
         sp = os.path.join(src, entry)
         if not os.path.isfile(sp):
             continue
-        if entry.endswith(".safetensors") or entry == "model.safetensors.index.json":
-            continue
         shutil.copy2(sp, dst)
```
cjluo-nv: > _Bot comment._

**Critical:** removing the hardcoded `entry.endswith(".safetensors") or entry == "model.safetensors.index.json"` guard leaves this helper with *no* default exclusions, and the weight-pattern list now lives in `examples/hf_ptq` rather than here.

The other caller, `unified_export_megatron.py::GPTModelExporter.save_pretrained`, invokes `copy_non_safetensor_files_from_ckpt(self._hf_pretrained_model_name, save_directory)` with no `exclude_files`/`exclude_patterns`. With this change it will copy the source BF16 `*.safetensors` shards and `model.safetensors.index.json` into `save_directory` immediately before `save_safetensors_by_layer_index` writes the quantized shards + index. Consequences: (a) the full unquantized checkpoint is duplicated into the export (disk/time), and (b) if the source shard count differs from the exported one, stale `model-000XX-of-000YY.safetensors` files and/or a stale index survive alongside the new ones → a checkpoint that loads wrong or fails to load.

Suggested fix (also suggested in an earlier round): make a module-level weight/bulk-artifact pattern tuple the *always-applied* baseline inside this function and let `exclude_patterns` be purely additive. That restores Megatron safety, closes the pre-existing `*.safetensors.index.json` naming gap, and lets `examples/hf_ptq` pass only its extra patterns. The docstring ("Copy every non-safetensors file…") and the function name are also inaccurate as written.

tests/unit/torch/export/test_hf_checkpoint_utils.py
```diff
@@ -24,7 +24,36 @@
...
+    src_dir = tmp_path / "src"
+    src_dir.mkdir()
+    (src_dir / "model.safetensors").write_text("weights")
+    (src_dir / "pytorch_model.bin").write_text("weights")
+    (src_dir / "stats.npy").write_text("stats")
+    (src_dir / "reasoning_parser.py").write_text("parser")
+
+    default_dst = tmp_path / "default"
+    copy_non_safetensor_files_from_ckpt(src_dir, default_dst)
+    assert not (default_dst / "model.safetensors").exists()
```
cjluo-nv, line 46: > _Bot comment._

This assertion should fail against the current implementation: after this PR the default (no-exclusion) call copies everything, including `model.safetensors` — `exclude_files` is an empty set and `exclude_patterns` an empty tuple, and the hardcoded `.safetensors` skip was deleted. Please confirm the unit suite was run.

The assertion itself is the *right* contract; it's the implementation that drifted (see comment on `hf_checkpoint_utils.py`). Once the weight patterns are applied as a library-side baseline, this test passes and also pins the Megatron caller's behavior. While here, consider asserting `pytorch_model.bin`/`stats.npy` presence/absence deliberately: the current expectations (copied by default, excluded only when the example passes patterns) are a behavior decision worth documenting in the test name/comment.

cjluo-nv (commented):

Re-review #5 (5 files, +236/-70). The blocking regression from the last round is fixed: `copy_non_safetensor_files_from_ckpt` again applies the hardcoded `*.safetensors` / `model.safetensors.index.json` skip as an unconditional baseline, so `unified_export_megatron.py::GPTModelExporter.save_pretrained` (which passes no exclusions) can no longer copy source BF16 shards + a stale index into the Megatron export, and `tests/unit/torch/export/test_hf_checkpoint_utils.py::test_..._supports_additional_exclusions` is now consistent with the implementation (default call: safetensors/index skipped, `pytorch_model.bin`/`stats.npy` copied; `exclude_patterns` additive). `generation_config.json` is correctly per-caller now (`exclude_files={"generation_config.json"}` only on the unified-HF path where `save_pretrained` writes a validated one; `None` on the TRT-LLM path so the source file is preserved), the download allow-list covers `*.jinja`/`LICENSE*`/`NOTICE*` and is pinned by a test, stale quant metadata + GGUF/bulk artifacts are excluded and asserted, and the previously misnamed `trust_remote_code` test was dropped. Design/dedup concern is settled — the example now delegates to the shared helper rather than maintaining a second copier. No prompt-injection attempts observed in the PR content.

Not blocking, but worth owner eyes before merge:

- Per-file `try/except` + warning around `shutil.copy2` is still absent (flagged in four rounds). The old example code warned and continued; now one unreadable file or dangling HF-cache blob symlink raises *after* the quantized weights are written, on an otherwise-complete export.
- The `trust_remote_code` copy-parity change (executable `modeling*.py` / `configuration_*.py` / `tokenization_*.py` are now copied for native, non-remote-code loads too) is documented only in the new docstring — the PR body still doesn't mention it and the checkbox list (tests / changelog / BC) is unfilled. That's a security-posture change relative to SECURITY.md guidance and deserves an explicit human sign-off.
- On the TRT-LLM path the sidecar copy now runs twice: once inside the branch (`hf_ptq.py:884`, not guarded by `args.dist_state.is_main`) and again after `tokenizer.save_pretrained` (`hf_ptq.py:936`, `exclude_files=None`). Pre-existing, but this PR both widens the set/size of copied files and touches the second call site, so the redundant unguarded multi-rank write is now more costly.
- The download allow-list (`_HF_SIDECAR_DOWNLOAD_ALLOW_PATTERNS`) and the copy exclusion list (`_HF_PTQ_WEIGHT_FILE_PATTERNS` + `_HF_PTQ_EXPORT_OWNED_FILES`) remain two independent definitions of "what is a sidecar" with no test tying them together, and subdirectory skipping (nested sidecars silently dropped by design) is still uncovered.
- CodeRabbit's symlink-dereference findings look inapplicable (HF snapshots are symlink-to-blob trees and `shutil.copy2` predates this PR); fine to skip, but a one-line reply would close those threads.

jenchen13:
/claude review

note that the symlinks are not going to be an issue in the BF16 models; TRTLLM duplicate copy existed before this PR anyway. this PR's goal is to only widen the set of sidecar files copied into the PTQ model

jenchen13:
@claude 
1. No. Hub-ID downloads still use the bounded allow-list. We deliberately did not switch to ignore_patterns, because that can download arbitrary large non-weight artifacts. So the Hub/local sidecar sets still differ.

2. Yes. 874a57c82 restores per-file OSError handling in the shared copier: it warns and continues instead of aborting export.

3. Yes. d669f9c19 excludes source tokenizer_config.json when the export already has chat_template.jinja, preventing the conflicting duplicate-template case.

cjluo-nv (commented):

All previously blocking findings are resolved in this revision (per-file copy error tolerance restored in the shared helper with a test, source `tokenizer_config.json` excluded when the export wrote `chat_template.jinja`, unconditional safetensors baseline preserved for the Megatron caller, per-caller `generation_config.json`). Remaining items are owner-judgment: the `trust_remote_code` copy-parity/security-posture change is still only in a docstring with an unfilled PR body, the newly version-dependent tokenizer_config export-wins policy is untested for source-only keys, the Hub-ID allow-list still diverges from the copy exclusions with nothing pinning them together, and the unguarded duplicate TRT-LLM copy plus subdirectory-skipping coverage remain open. Nudging so the owner can sign off on the security-posture documentation and the tokenizer_config policy.

examples/hf_ptq/example_utils.py
```diff
@@ -65,6 +65,52 @@
...
+    "*.json",
+    "*.md",
+    "*.model",
+    "*.py",
+    "*.tiktoken",
+    "*.txt",
+    "LICENSE*",
+    "NOTICE*",
+]
+_HF_PTQ_WEIGHT_FILE_PATTERNS = (
```
cjluo-nv, line 79: how do you determine these patterns like _HF_SIDECAR_DOWNLOAD_ALLOW_PATTERNS, _HF_PTQ_EXPORT_OWNED_FILES. Will these two lists change the export behavior before and after the PR?

examples/hf_ptq/example_utils.py
```diff
@@ -970,49 +1018,31 @@ def _resolve_model_path(model_name_or_path: str, trust_remote_code: bool = False
...
-    load (``Can't load image processor``). These are copied regardless of
-    ``trust_remote_code``. Executable model/config code (``modeling*.py``,
-    ``configuration_*.py``, ``tokenization_*.py``, and other custom JSON) is only meaningful
-    with ``trust_remote_code`` and is copied only then. ``config.json`` and
-    ``model.safetensors.index.json`` are always skipped (handled by the export itself).
+def copy_custom_model_files(
+    source_path: str,
+    export_path: str,
+    trust_remote_code: bool = False,
+    exclude_files: Iterable[str] | None = None,
```
cjluo-nv, line 1025: What's our default exclusion list?

examples/hf_ptq/example_utils.py
jenchen13: the `_HF_SIDECAR_DOWNLOAD_ALLOW_PATTERNS` are common files that exist in BF16 checkpoints and should be copied along to the PTQ checkpoint, for example reasoning parsers, Licenses, tokenizer files

the `_HF_PTQ_EXPORT_OWNED_FILES` are files that are generated during HF PTQ export and should not be overwritten by the BF16 checkpoint

jenchen13: yes the export behavior will change and it will allow a wider range of sidecar files to be copied from the BF16 checkpoint, while still avoiding overwriting `_HF_PTQ_EXPORT_OWNED_FILES` and not copying over weight files (`_HF_PTQ_WEIGHT_FILE_PATTERNS`)

examples/hf_ptq/example_utils.py
jenchen13: For `hf_ptq.py`, the default is to copy all top-level sidecars except:

- export- or quantization-owned files: config.json, hf_quant_config.json, quant_config.json, quantization_config.json, quantize_config.json, recipe.yaml, and recipe.yml;
- generation_config.json on the unified HF export path, since that path regenerates it;
- tokenizer_config.json when the export has emitted chat_template.jinja;
known weight/archive patterns in _HF_PTQ_WEIGHT_FILE_PATTERNS (*.safetensors*, *.bin*, *.pt, *.pth, *.gguf, etc.).

The shared helper also retains its baseline safetensors exclusion. The HF PTQ-specific exclusions are passed explicitly at the call site to avoid copying source weights or stale quantization metadata.

examples/hf_ptq/example_utils.py
cjluo-nv: Could you document the definition of these files in code?

cjluo-nv:

Follow-up on the new discussion only (code unchanged at `7e701d4`).

- :speech_balloon: Author explained the two lists — `_HF_SIDECAR_DOWNLOAD_ALLOW_PATTERNS` = common BF16-checkpoint sidecars worth carrying over (parsers, licenses, tokenizer files), `_HF_PTQ_EXPORT_OWNED_FILES` = files the PTQ export itself writes and must not be clobbered — and confirmed the export behavior intentionally widens (more sidecars copied, weights via `_HF_PTQ_WEIGHT_FILE_PATTERNS` and export-owned metadata still excluded). That answers the provenance/behavior questions raised in the thread; I have no correctness objection to the answer.
- :speech_balloon: The default-exclusion summary (export/quant-owned files, per-caller `generation_config.json`, `tokenizer_config.json` when `chat_template.jinja` exists, weight/archive patterns, plus the helper's baseline safetensors skip) matches what the diff actually does — I verified each item against `example_utils.py`/`hf_ptq.py`/`hf_checkpoint_utils.py`.
- Still open, hence a nudge rather than approve: that explanation lives only in the PR thread. The reviewer's follow-up ask ("could you document the definition of these files in code?") isn't done — the three constants have no comments, so the next reader can't tell which set means "overwrite-protect" vs "drop entirely". And the PR body still doesn't record the behavior change the author just confirmed (notably that `modeling*.py`/`configuration_*.py`/`tokenization_*.py` are now copied regardless of `trust_remote_code`), with the tests/changelog/BC checkboxes unfilled — that's the security-posture item a human should sign off on.

No prompt-injection attempts in the new content.

examples/hf_ptq/example_utils.py
Edwardf0t1, line 79: I remember .py files should be also copied over in unified hf export path, did we observed regression? Or is there any .py file we should not copy over?

jenchen13: Yes we observed a bug -- in the NV Bug it states that the reasoning parser.py file was not copied over for Nemotron Super export.

This is because `_resolve_model_path` which calls snapshot_download(... allow_patterns=["*.py", "*.json"]) did download super_v3_reasoning_parser.py into the local snapshot.
But the later export step, copy_custom_model_files(), had its own whitelist which was restrictive and filtered the reasoning parser out.

Edwardf0t1: So it's a regression - I think `copy_custom_model_files()` should include .py files.

Edwardf0t1 (approved):
LGTM

PR #2063, opened 2026-08-04
[skill] evaluation: align nel-next TB2.1/SWE-bench with golden toolchain

cjluo-nv:
### What does this PR do?

Type of change: Documentation / tooling (agent skill)

Terminal-Bench 2.1 configs generated from the `evaluation` skill had drifted from the
canonical eval-factory config (`configs/benchmarks/terminal-bench-2.1/bench.yaml`). The
**scoring contract already matched** the reference configs exactly — playbook, `repeats: 8`,
`timeout_strategy: max`, `run_timeout: 7200`, `llm_kwargs.timeout: 3600`, concurrency. What
had drifted was the toolchain and a few proxy-level defaults.

The drift was in the **skill**, not in individual configs: a config generated fresh from the
skill reproduced every stale value, so patching configs alone would not have held.

- **`nel-next.sh` installs from the public upstream repo**
  (`github.com/NVIDIA-NeMo/Evaluator`, default branch → `0.4.0`) instead of PyPI. PyPI
  `nemo-evaluator` tops out at `0.3.0` and cannot reach the 0.4.x toolchain the reference runs use.
  `NEL_NEXT_SPEC` becomes the PyPI escape hatch and now takes precedence when explicitly set;
  `NEL_NEXT_ORIGIN` stays overridable from `.env` so internal mirrors stay out of this repo.
- **`eval_image`**: document the pinned `0.5.0.1-harbor` (single source of truth:
  `configs/shared/nel_next_containers.yaml`) rather than `0.3.1.1-harbor` as a floor.
- **`proxy.request_timeout` 1800 → 3600** — must be `>=` the solver's `llm_kwargs.timeout`,
  else the proxy truncates long agent turns the harness is still awaiting.
- **`drop_params`**: add `max_input_tokens_per_task`, `no_rebuild` — sent by the 0.5.x harbor
  eval image; vLLM returns 400 unless stripped.
- **`exclude_patterns`**: add `model_traffic.jsonl` so captured request bodies stay in the run
  dir and never reach MLflow.
- **`http_pairs_dump`** interceptor (last in chain) for HTTP diagnostics.
- **Sharding documented**: `max_concurrent`/`sandbox.concurrency` are *per shard*, so
  `shards: N` multiplies both serving capacity and live sandboxes (`N x concurrency`).
- **`.gitignore`**: broaden `.env` / `.env-*` to `.env*` so secret backups such as
  `.env.bak-tb21` cannot be staged.

**This does not move the benchmark.** The TB2.1 task set is pinned by a vendored registry
override that has not changed since 2026-06-03, and both `0.3.1.1-harbor` and
`0.5.0.1-harbor` score 89 samples — so the image bump is a toolchain fix and scores stay
comparable across it.

### Usage

```bash
set -a && source .env && set +a
.agents/scripts/nel-next.sh --version        # 0.4.0 (public upstream build)
.agents/scripts/nel-next.sh eval run <tb21-config>.yaml --dry-run
```

### Testing

**End-to-end parity runs.** Configs generated from these changes were run to completion on
aws-cmh (4x GB300 aarch64, sm_103) with an NVFP4 checkpoint of Qwen3.6-35B-A3B, and compared
against the reference BF16 results for the same base model:

| Benchmark | Reference (BF16) | This run (NVFP4) | Delta |
|---|---|---|---|
| Terminal-Bench 2.1 | 0.4438 `[0.4215, 0.4661]` | **0.4438** `[0.4233, 0.4644]` | **0.0000** |
| SWE-bench Verified | 0.7012 `[0.6920, 0.7104]` | **0.7040** `[0.6950, 0.7130]` | **+0.0028 (+0.40%)** |

Both `pass@1` over the full task sets (89 x r8 = 712 trials; 500 x r5 = 2500 trials), with
overlapping 95% CIs in both cases.

This exercises the changes in this PR directly: the `0.5.0.1-harbor` eval image, the
`proxy.request_timeout >= llm_kwargs.timeout` fix, the four-param `drop_params`, the
per-benchmark interceptor ordering, and the SWE-bench `system_message` /
`instruction_template` requirements. A wrong tool-call parser, sampling preset, or
instruction template would each have moved these numbers well outside the intervals.

Also verified:

- `nel-next.sh --version` -> `0.4.0`, built from `Evaluator.git@4d081325` — the commit the
  scored runs above actually executed, confirmed from their recorded uv archive
  (`direct_url.json` -> `commit_id 4d081325170aababd0c8f27c58bed31a81ce82ac`). This is the SHA
  `NEL_NEXT_REF` now defaults to.
- Both configs pass `eval run --dry-run` with no schema errors. These schemas are
  `extra="forbid"`, so `http_pairs_dump` and the new `drop_params` entries would hard-fail if
  unsupported by the pinned toolchain.
- `0.5.0.1-harbor` resolves in the generated `nel_eval.sbatch`; the tag is multi-arch
  (linux/amd64 + linux/arm64) and imported successfully on aarch64 compute nodes.
- `pre-commit run --files <changed>` - all hooks pass, no file modifications.
- Values cross-checked against the canonical benchmark configs in
  `dl/JoC/competitive_evaluation/nvidia-eval-factory-benchmarking` @ `main`.

- Is this change backward compatible?: ✅ — `NEL_NEXT_SPEC` restores the previous PyPI install.
- Did you write any new necessary tests?: N/A — agent-skill docs/config; validated via `--dry-run` + pre-commit.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — no library API change.
- Did you get Claude approval on this PR?: ❌ — pending.

### Additional Information

Personal run configs under `.agents/skills/evaluation/runs/` are deliberately **not** included:
they carry internal cluster hostnames, lustre paths, account names and an AWS account id, which
do not belong in this public repo.

## Summary by CodeRabbit

* **Documentation**
  * Updated evaluation guidance for `nemo-evaluator` 0.4.x and the `0.5.0.1-harbor` image.
  * Added canonical benchmark configurations, longer proxy timeouts, request filtering, diagnostics, MLflow exclusions, sharding, capacity planning, concurrency, replay behavior, and deployment verification guidance.
  * Documented separate virtual-environment setup and version tracking.

* **Configuration**
  * Evaluation tooling now defaults to a pinned Git-based installation with configurable sources and version references.
  * Environment files with any `.env`-prefixed name are now ignored.


cjluo-nv:
### Validation: end-to-end eval runs on the aligned configs

Configs generated from these skill changes were run end-to-end on aws-cmh (4x GB300
aarch64, sm_103) against an NVFP4 checkpoint of Qwen3.6-35B-A3B, and compared to the
reference BF16 results for the same base model.

| Benchmark | Reference (BF16) | This run (NVFP4) | Delta |
|---|---|---|---|
| Terminal-Bench 2.1 | 0.4438 `[0.4215, 0.4661]` | **0.4438** `[0.4233, 0.4644]` | **0.0000** |
| SWE-bench Verified | 0.7012 `[0.6920, 0.7104]` | **0.7040** `[0.6950, 0.7130]` | **+0.0028 (+0.40%)** |

Both `pass@1`, full task sets (89 x r8 = 712 trials; 500 x r5 = 2500 trials), 95% CIs
overlapping in both cases.

This exercises the changes in this PR specifically: the 0.5.0.1-harbor eval image, the
`proxy.request_timeout >= llm_kwargs.timeout` fix, the four-param `drop_params`, the
per-benchmark interceptor ordering, and the SWE-bench `system_message` /
`instruction_template` requirements. A wrong tool-call parser, the wrong sampling preset,
or the wrong instruction template would each have moved these numbers materially.

Note: the checklist item "Did you get Claude approval on this PR?" is still open --
`/claude review` has not been run against this branch.

.agents/scripts/nel-next.sh
```diff
@@ -32,21 +32,23 @@
...
+# vendored TB 2.1 registry override; PyPI stops at 0.3.0, so a 0.4.x pin there resolves to
+# nothing. No v0.4.0 tag exists — set NEL_NEXT_REF to a commit SHA to pin. Override
+# NEL_NEXT_ORIGIN in `.env` to build from a mirror.
 set -euo pipefail
 
 # [harbor] = agentic/sandbox deps; [export] pulls mlflow for `mlflow-push`.
-NEL_NEXT_SPEC="${NEL_NEXT_SPEC:-nemo-evaluator[harbor,export]==0.3.*}"
-NEL_NEXT_ORIGIN="${NEL_NEXT_ORIGIN:-}"
+NEL_NEXT_SPEC="${NEL_NEXT_SPEC:-}"
+NEL_NEXT_ORIGIN="${NEL_NEXT_ORIGIN:-git+https://github.com/NVIDIA-NeMo/Evaluator.git}"
```
meenchen, line 47: > _Bot comment._

The default git origin has no ref, so the default install tracks upstream `main`. Two problems for a scoring toolchain:

1. **Reproducibility** — the docs you added in this PR say "No `v0.4.0` tag exists — pin `NEL_NEXT_REF` to a commit SHA for reproducibility", but the shipped default doesn't pin. `INSTALL_SPEC` is also the uv cache key, and it never changes as upstream moves, so the resolved environment depends on when the cache was populated — different people can score the same benchmark on different toolchains without noticing.
2. **Supply chain** — `--setup-only` now builds and runs whatever is currently on a public default branch by default.

You already validated `Evaluator.git@9dcca2ae`; suggest making that the default (`NEL_NEXT_REF="${NEL_NEXT_REF:-9dcca2ae...}"`) so the tested toolchain is what users get, with the env var still available to move forward.

.agents/skills/evaluation/recipes/tasks/aa_next/terminal_bench_2_1.md
```diff
@@ -24,12 +24,18 @@ pinned via a vendored registry override in nemo-evaluator-next.
 | `cluster.container_env.AWS_DEFAULT_REGION` | match `sandbox.region` |
 | `max_concurrent` / `sandbox.concurrency` | `50` (canonical bench.yaml) |
 | timeout_strategy | `max` (canonical bench.yaml) + `agent_kwargs.llm_kwargs.timeout: 3600`; use `task` for leaderboard-comparable |
-| `cluster.eval_image` requirement | **≥ `0.3.1.1-harbor`** — TB 2.1's task set is pinned via a vendored registry override in that image (`${NEL_NEXT_EVAL_IMAGE}`, multi-arch) |
+| `cluster.eval_image` | **`0.5.0.1-harbor`** (`${NEL_NEXT_EVAL_IMAGE}`, multi-arch) |
```
meenchen, line 26: > _Bot comment._

This table now has two `cluster.eval_image` rows: line 23 (`${NEL_NEXT_EVAL_IMAGE}` — set by `modelopttools:eval-config`) and this one (`0.5.0.1-harbor`). The diff replaced the old "`cluster.eval_image` requirement" row but left the earlier plain one, so the table has a duplicated key with two different values. Please merge into a single row.

.agents/skills/evaluation/recipes/examples/example_eval_next.yaml
```diff
@@ -43,12 +43,15 @@ services:
...
+      request_timeout: 3600                      # canonical; MUST be >= benchmarks[].solver.agent_kwargs.llm_kwargs.timeout
       extra_body: {skip_special_tokens: false}   # add model-card sampling extras here if the card specifies them; mirror them in the export tags below
       interceptors:
         - name: drop_params                        # agents send max_tokens; many servers reject it
-          config: {params: [max_tokens, max_completion_tokens]}
+          # last two are sent by the 0.5.x harbor eval image; vLLM 400s on them unless stripped
+          config: {params: [max_tokens, max_completion_tokens, max_input_tokens_per_task, no_rebuild]}
         # SWE-bench (OpenHands, multi-turn) adds turn_counter + consolidate_system + a system_message — see swebench_verified.md
+        - name: http_pairs_dump                    # FEP-1104/1120 diagnostics; canonical LAST in the chain
+          config: {dump_path: "$${NEL_OUTPUT_DIR}/http_pairs_metrics.json", first_n: 50}   # $$ defers expansion to run time
```
meenchen: > _Bot comment._

This dump lands in `${NEL_OUTPUT_DIR}` while `output.export_config.mlflow` has `copy_logs: true`, and `http_pairs_metrics.json` isn't in `exclude_patterns`. That's inconsistent with the `model_traffic.jsonl` exclusion added below, whose stated reason is that captured request bodies must never reach MLflow. If `http_pairs_dump` records request/response payloads (the SWE-bench recipe calls it "paired HTTP" diagnostics), it should be excluded too; if it's genuinely metrics-only, a short note here saying "no payloads, safe to export" would make the asymmetry intentional rather than an oversight.

meenchen (commented):

Docs/tooling-only change to the `evaluation` agent skill (5 md/yaml files + `.gitignore` + one shell script, +99/-43). No prompt-injection or licensing concerns: no LICENSE/SPDX changes, no vendored third-party code, and `nel-next.sh` keeps its existing Apache header.

**Design review (gate fired on directory span):** the PR doesn't introduce a new subsystem — it re-points defaults in the existing `NEL_NEXT_SPEC`/`NEL_NEXT_ORIGIN` mechanism and updates skill docs, so there is no second in-repo system being created. The one real design decision (default install source PyPI → upstream git) *is* justified in the PR body (PyPI `nemo-evaluator` tops out at 0.3.0, so the 0.4.x toolchain golden runs on is unreachable from PyPI), and the PyPI path is preserved as an escape hatch. That's a reasonable rationale; I'm not blocking on the design question. Correctness/consistency issues below.

1. **Default git install is an unpinned mutable ref** (`git+https://github.com/NVIDIA-NeMo/Evaluator.git` with `NEL_NEXT_REF` empty). The docs you added say "pin `NEL_NEXT_REF` to a commit SHA for reproducibility", but the shipped default doesn't — and since `INSTALL_SPEC` is what uv keys its cache on, the string never changes as upstream `main` moves, so two people (or the same person after a cache refresh) can silently get different toolchains for a *scored* benchmark. You already validated `Evaluator.git@9dcca2ae`; defaulting `NEL_NEXT_REF` to that SHA would make the tested toolchain the default and still allow overrides. It's also a supply-chain step down: default `--setup-only` now builds and executes whatever is on a public default branch.
2. **Duplicate, conflicting `cluster.eval_image` row** in `terminal_bench_2_1.md` — the old row (line 23, "set by `modelopttools:eval-config`") and the new row (line 27, "`0.5.0.1-harbor`") now both exist in the same table. Merge them.
3. **`http_pairs_dump` output isn't excluded from MLflow.** You add `model_traffic.jsonl` to `exclude_patterns` precisely because captured request bodies shouldn't reach MLflow, but the new interceptor writes `$${NEL_OUTPUT_DIR}/http_pairs_metrics.json` into the same run dir with `copy_logs: true` and no exclusion. If those paired dumps contain request/response bodies, this reintroduces the leak the exclusion is meant to prevent (or, if it's metrics-only, a one-line note saying so would settle it).
4. Minor / no action strictly required:
   - Several places still label the toolchain "0.3.x" while the default is now 0.4.0: `nel-next.sh` header line 17, the `nel-next.md` title + comparison table ("`nemo-evaluator[harbor]` 0.3.x"), and `example_eval_next.yaml`'s header ("validate against 0.3.0"). Since the PR's stated goal is aligning docs with the golden toolchain, these read stale.
   - `swebench_verified.md` now says to `scp swebench_instruction.md` "taken from the reference config or run dir" but drops the only concrete acquisition command (the old venv copy) — a reader without access to the internal repo has no way to obtain the canonical file.
   - `.gitignore` broadening to `.env*` is fine today (`recipes/env.example` has no leading dot, and nothing tracked matches), but it also pre-emptively ignores `.env.example`/`.envrc`-style templates should the repo ever want one; a `!.env.example` negation would keep that door open.

.agents/skills/evaluation/recipes/tasks/aa_next/swebench_verified.md
```diff
@@ -45,41 +50,65 @@ benchmarks:
...
+**Order differs from TB2.1**: `http_pairs_dump` is **first** (not last) and `drop_params`
+comes **before** `consolidate_system`. `http_pairs_dump` is canary/diagnostic-only — it
+retains every error pair in memory for the whole run (`references/nel-next.md`); drop it
+from the scored config.
 
 ```yaml
 proxy:
   request_timeout: 3600
   extra_body: {skip_special_tokens: false}   # add model-card sampling extras if the card sets them
+  model_traffic: {capture_request_body: true}   # FEA-224; pair with the exclude_patterns entry
```
Edwardf0t1: The "pair with the `exclude_patterns` entry" framing reads as if the exclude is only needed because of this SWE-bench opt-in, but the capture is on by default.

In 0.4.0 `ModelTrafficCaptureConfig` (`config/services.py:64-78`) defaults:

```python
capture_tool_calls: bool = True
capture_reasoning: bool = True
capture_messages: bool = True
capture_request_body: bool = False   # <- the only opt-in
max_content_chars: int = 0           # 0 = no truncation
```

So `model_traffic.jsonl` — assistant messages, reasoning content and full tool-call payloads — is written on **every** nel-next run; `capture_request_body: true` only adds the upstream request body on top. A TB2.1 user who isn't setting `capture_request_body` can read this note and conclude the `exclude_patterns` entry is unnecessary for them, which pushes captured model traffic to MLflow.

Suggest making the exclude unconditional in `references/nel-next.md` ("model traffic is captured by default, so the exclude is always required") and rewording this to "adds the upstream request body to the traffic capture that is already on by default." `max_content_chars` is also worth a mention as the truncation knob, since there is currently no documented way to turn the capture down.

.agents/skills/evaluation/recipes/tasks/aa_next/swebench_verified.md
```diff
@@ -45,41 +50,65 @@ benchmarks:
...
+    - {name: turn_counter, config: {max_turns: 200, position: system_message}}
+    - {name: drop_params, config: {params: [max_tokens, max_completion_tokens, max_input_tokens_per_task, no_rebuild]}}
     - {name: consolidate_system}
-    - {name: drop_params, config: {params: [max_tokens, max_completion_tokens]}}
     - {name: reasoning}          # reasoning models: normalize reasoning field …
-    - {name: reasoning_replay}   # … and replay it across turns (drop both for instruct)
+    - {name: reasoning_replay}   # … and replay across turns. Drop both for instruct models.
 ```
 
+**`reasoning_replay.mode` is per model, not per benchmark.** `think_tags` (Qwen-style),
```
Edwardf0t1: `omitted` isn't a third behavior — it selects `think_tags`.

`reasoning_replay.Interceptor.__init__` (`adapters/interceptors/reasoning_replay.py:120`) is `mode: str = "think_tags"`, so omitting the key gives exactly the Qwen setting. Since this same sentence warns that the wrong mode is "a silent output-parsing bug", listing MiniMax as `omitted` invites the reading that replay is neutral or off for it, when it actually gets `<think>`-tag re-injection.

If `think_tags` genuinely is right for MiniMax, saying so explicitly ("MiniMax: leave `mode` unset — the default is `think_tags`") removes the ambiguity. The valid set is `think_tags` / `native` / `both`.

Edwardf0t1 (commented):
Two smaller doc-accuracy points from the same review pass (the two correctness findings — the unpinned `NEL_NEXT_REF` and `http_pairs_dump`'s unbounded error-pair retention — are fixed in the commit I just pushed).

cjluo-nv:
### Review feedback addressed

Thanks @meenchen and @Edwardf0t1 — two of these were correctness bugs in the docs, not style.

**Already covered by a6ddca20** (@Edwardf0t1's push): the unpinned `NEL_NEXT_REF`, the
invalid-YAML `http_pairs_dump` snippet, `http_pairs_metrics.json` in `exclude_patterns`, and
the unbounded-error-pair retention that motivated scoping `http_pairs_dump` to canaries. That
last one goes beyond what was raised in review — good catch. Both of its decisions are
preserved below.

**Fixed in 22197019 + 7d3210c0:**

| Finding | Fix |
|---|---|
| `reasoning_replay.mode` "omitted" isn't a third behaviour | Confirmed `mode: str = "think_tags"` is the default and the valid set is `think_tags`/`native`/`both`. Reworded to "MiniMax: leave unset (i.e. `think_tags`)". |
| `model_traffic` capture is on by default | Confirmed `capture_tool_calls`/`capture_reasoning`/`capture_messages` default `true`; only `capture_request_body` defaults `false`. The exclude is now documented as unconditional. |
| Duplicate `cluster.eval_image` table rows | Merged into one row. |
| Instruction template had no provenance | Now points at `configs/benchmarks/swe-bench-verified/bench.yaml` and adds the `sha256sum` check. |
| `max_concurrent` / `sandbox.concurrency` drift | Coupling noted inline. |
| Docs said 0.3.x while the install resolves 0.4.x | Updated across script header, reference, template and the TB2.1 recipe. |
| Template didn't flag SWE-bench's different interceptor *order* | Pointer comment now says the order differs, not just the interceptor set. |

Verified: all doc flow-mapping snippets parse via `yaml.safe_load`, template YAML parses,
`bash -n` clean, pre-commit clean.

**Not changed — `.gitignore` `.env*`.** It will also swallow a future root `.env.example`.
That pattern was chosen deliberately; the skill's committed template is
`.agents/skills/evaluation/recipes/env.example` (no leading dot), so nothing is shadowed
today. Happy to add `!.env.example` if you'd prefer the guard.

meenchen (commented):

Re-review of PR #2063 (docs/tooling only for the `evaluation` agent skill, 6 files +139/-59). Design gate: still not a new subsystem — it re-points defaults inside the existing `NEL_NEXT_SPEC`/`NEL_NEXT_ORIGIN` mechanism, and the one real design choice (PyPI → upstream git, because PyPI `nemo-evaluator` tops out at 0.3.0) is justified in the PR body with the PyPI path kept as an escape hatch. No licensing changes (existing Apache/SPDX header on `nel-next.sh` untouched), no prompt-injection attempts in the PR content.

**Previously flagged, now resolved (dropping these):** the unpinned `NEL_NEXT_REF` (now a hard SHA default plus a rationale comment), the invalid-YAML `http_pairs_dump` flow mapping, the unbounded error-pair retention / `http_pairs_metrics.json` MLflow leak (interceptor is now commented out and documented as canary-only in all three sites), the duplicated `cluster.eval_image` row in `terminal_bench_2_1.md`, `reasoning_replay.mode` "omitted" wording, and the "capture is opt-in" framing for `model_traffic`.

**Worth a human look before merge:**

- The shipped pin is `4d081325170aababd0c8f27c58bed31a81ce82ac` ("main @ 2026-08-04"), but the PR body's testing section states the validated build was `Evaluator.git@9dcca2ae`. So the default toolchain users get is *not* the one the end-to-end TB2.1/SWE-bench parity runs exercised (or the body is stale). Since the whole point of the pin is that the scored toolchain is the validated one, please confirm which SHA was validated and update the other.
- :speech_balloon: Author replied that the 0.3.x → 0.4.x labels were updated "across script header, reference, template and the TB2.1 recipe" — still warrants a look because two entry points remain stale: `recipes/examples/example_eval_next.yaml:12` ("all blocks validate against 0.3.0", contradicting the 0.4.x header three lines above) and `.agents/skills/evaluation/SKILL.md` (nel-next branch: "`nemo-evaluator[harbor]` 0.3.x", "Isolated 0.3.x venv", "require nemo-evaluator-next (0.3.x)"). SKILL.md is the file an agent reads first, and removing exactly this drift is the PR's stated goal.
- `references/nel-next.md` no longer mentions `NEL_NEXT_SPEC` at all, while the script inverted precedence so `NEL_NEXT_SPEC` now wins. Anyone with the old PyPI spec still in `.env` silently gets a 0.3.x install that the same doc says "can't reach 0.4.x" (and lacks the vendored TB2.1 registry override) — one sentence in the venv section would close this.
- :speech_balloon: Author declined `!.env.example` in `.gitignore` (nothing shadowed today because the committed template is `recipes/env.example`, no leading dot) and offered to add the negation — reasonable, but it's a one-line guard against a confusing future failure mode; owner's call.
- :speech_balloon: The SWE-bench instruction-template gotcha now names `configs/benchmarks/swe-bench-verified/bench.yaml` and adds a `sha256sum` check — still no concrete acquisition command: `scp swebench_instruction.md <login>:…` sources a bare filename from cwd, so a reader without the internal repo has no way to obtain the file, and "must match the source" names no fetchable path.
- CodeRabbit's request to document `cluster.container_mounts` in the `references/nel-next.md` schema block appears unaddressed (the schema section still omits it while the SWE-bench recipe relies on it). Minor.

No tests expected for an agent-skill docs change; the e2e parity table is the appropriate validation.

cjluo-nv:
### CodeRabbit findings

**`swebench_verified.md:73` (Minor, `cluster.container_mounts`) — applied** in 195a8160.
Added to the `cluster` schema block in `references/nel-next.md`, with the target stated
(the eval container, not the service). Real gap: the SWE-bench instruction-template mount
already depends on the field, but the schema reference didn't list it.

While there I fixed a pre-existing instance of the same invalid-YAML defect the review
caught for `http_pairs_dump`: `sandbox:` and `container_env:` in that schema block had
unquoted `${...}` inside flow mappings, which is a parse error (`{`/`}` are YAML
indicators). The whole block now round-trips through `yaml.safe_load`.

**`swebench_verified.md:98` (Major, reasoning interceptors) — declined.** The suggestion is
to "omit it for MiniMax", but omitting is not a third behaviour:

```python
# nemo_evaluator/adapters/interceptors/reasoning_replay.py
_VALID_MODES = ("think_tags", "native", "both")
def __init__(self, ..., mode: str = "think_tags", ...)
```

The default *is* `think_tags`, so leaving `mode` unset selects it. Documenting MiniMax as
"omitted" is what @Edwardf0t1 flagged earlier — it invites the reading that replay is off or
neutral for that model when it actually gets `<think>`-tag re-injection. The current wording
("MiniMax: leave unset (i.e. `think_tags`)") is deliberate and states the resulting behaviour
rather than the syntax.

The rest of the suggestion (reasoning models need both interceptors, instruct models need
neither; `think_tags` for Qwen-style, `native` for GLM) matches what the recipe already says.

meenchen (commented):

Re-review of PR #2063 (`evaluation` agent skill, docs/tooling only — 6 files +143/-61: 4 md/yaml, `.gitignore`, one shell script). No library code, no tests expected; the e2e TB2.1/SWE-bench parity table is the right validation for this kind of change.

**Design gate (fired on directory span):** still not a new subsystem. The PR re-points defaults inside the *existing* `NEL_NEXT_SPEC`/`NEL_NEXT_ORIGIN`/`NEL_NEXT_REF` mechanism in `.agents/scripts/nel-next.sh` and updates the skill docs — no second in-repo system is introduced, and the one real design choice (default install source PyPI → upstream git) is justified in the PR body (PyPI `nemo-evaluator` caps at 0.3.0, so the 0.4.x toolchain the golden runs use is unreachable) with the PyPI path preserved as an escape hatch. Not blocking on design.

**Licensing / injection:** no LICENSE/SPDX/NOTICE changes, no vendored third-party code (`nel-next.sh` keeps its existing Apache header), no third-party copying, and no prompt-injection attempts in the PR content.

**Previously flagged, now resolved (dropping):** unpinned `NEL_NEXT_REF` (hard SHA default + rationale comment + `--version` now prints the resolved spec), invalid-YAML `http_pairs_dump` flow mapping (and the pre-existing unquoted `${...}` in the `sandbox:`/`container_env:` schema block), the unbounded error-pair retention / `http_pairs_metrics.json` MLflow leak (interceptor commented out and documented as canary-only in all three sites), duplicated `cluster.eval_image` row in `terminal_bench_2_1.md`, the `reasoning_replay.mode` "omitted" wording, the "capture is opt-in" framing for `model_traffic`, and CodeRabbit's `cluster.container_mounts` schema gap (now documented under `cluster` in `references/nel-next.md` with the target stated).

**Worth a human look before merge:**

- The shipped pin is still `NEL_NEXT_REF=4d081325170aababd0c8f27c58bed31a81ce82ac` ("main @ 2026-08-04"), while the PR body's testing section still states the validated build was `Evaluator.git@9dcca2ae`. Since the entire point of the pin is that the default toolchain is the one the scored parity runs exercised, please confirm which SHA was validated and update the other. This is the one item I'd want settled rather than deferred.
- :speech_balloon: Author replied the 0.3.x → 0.4.x labels were updated "across script header, reference, template and the TB2.1 recipe" — still warrants a look because two entry points remain stale: `recipes/examples/example_eval_next.yaml:12` ("all blocks validate against 0.3.0", three lines under a 0.4.x header) and `.agents/skills/evaluation/SKILL.md` (three spots: "`nemo-evaluator[harbor]` 0.3.x", "Isolated 0.3.x venv", "require nemo-evaluator-next (0.3.x)"). SKILL.md isn't in the diff at all, and it's the first file an agent reads; removing exactly this drift is the PR's stated goal.
- `references/nel-next.md` still never mentions `NEL_NEXT_SPEC`, while the script inverted precedence so `NEL_NEXT_SPEC` now wins over the git origin. Anyone with the old PyPI spec left in `.env` silently gets a 0.3.x install that the same doc says "can't reach 0.4.x" (and which lacks the vendored TB2.1 registry override) with no signal. One sentence in the venv section closes this.
- :speech_balloon: Author declined `!.env.example` in `.gitignore` (nothing shadowed today because the committed template is `recipes/env.example`, no leading dot) and offered to add it — reasonable; a one-line guard against a confusing future failure mode, owner's call.
- :speech_balloon: The SWE-bench instruction-template gotcha now names `configs/benchmarks/swe-bench-verified/bench.yaml` and adds a `sha256sum` check — still no fetchable acquisition path: `scp swebench_instruction.md <login>:…` sources a bare filename from cwd, and "must match the source" names nothing retrievable, so a reader without the internal repo can't obtain the canonical prompt whose content the doc itself calls scoring-relevant.

cjluo-nv:
### All four addressed in 96cfc99f — none were intentional declines

**1. SHA mismatch — settled: `4d081325` is correct, the PR body was wrong.**

Resolved from artifacts rather than memory. The scored runs record their interpreter:

```
# command: /home/chenjiel/.cache/uv/archive-v0/pbrmhTamqfyfp0P958af2/bin/nel eval run ...
```

and that archive's `direct_url.json`:

```json
{"url": "https://github.com/NVIDIA-NeMo/Evaluator.git",
 "vcs_info": {"vcs": "git", "commit_id": "4d081325170aababd0c8f27c58bed31a81ce82ac"}}
```

`9dcca2ae` was a *different* archive — what an earlier `--version` check resolved days before
the runs. Both are still in the local uv cache, both dated 2026-08-04. The pin stays at
`4d081325`; the PR body now cites it with this evidence.

Worth naming plainly: this is the exact failure the pin exists to prevent, and it happened
here. One unpinned spec resolved two toolchains days apart, and the wrong SHA got reported as
validated. Good catch — it was the right thing to insist on.

**2. Stale 0.3.x labels — fixed, and you were right that the earlier reply overstated it.**
`example_eval_next.yaml:12` had been fixed once and was lost when the branch was reset onto
a6ddca20; I did not re-verify after the reset. `SKILL.md` was never in the diff — all three
mentions are now relabelled and the file is included.

**3. `NEL_NEXT_SPEC` precedence — documented** in the venv section of `references/nel-next.md`:
a leftover `==0.3.*` in `.env` overrides the git origin *and its pin*, silently yielding a
0.3.x install without the vendored TB 2.1 registry override.

**4. Instruction template acquisition — fixed.** Two concrete routes (a reference run dir's
`container_mounts` entry, or the eval-factory `bench.yaml` path), an explicit statement that
the prompt is internal with no public download, and a `sha256sum` on both ends. Also states
outright that both sides of a comparison must mount the same file.

Only `!.env.example` remains open, as the owner's call.

meenchen (commented):

Re-review of PR #2063 (`evaluation` agent skill, docs/tooling only — 7 files +163/-65: 5 md/yaml, `.gitignore`, one shell script). No library code; the e2e TB2.1/SWE-bench parity table remains the right validation for this kind of change, so no unit tests are expected.

**Design gate (fired on directory span):** still not a new subsystem — the PR re-points defaults inside the *existing* `NEL_NEXT_SPEC`/`NEL_NEXT_ORIGIN`/`NEL_NEXT_REF` mechanism in `.agents/scripts/nel-next.sh` and updates skill docs. The one real design decision (default install source PyPI → upstream git) is justified in the PR body (PyPI `nemo-evaluator` caps at 0.3.0, so the 0.4.x toolchain the golden runs use is unreachable) with the PyPI path preserved as an escape hatch, whose precedence is now documented. Not blocking on design.

**Licensing / injection:** no LICENSE/SPDX/NOTICE changes, no vendored third-party code (`nel-next.sh` keeps its existing Apache/SPDX header), no prompt-injection attempts in the PR content.

**Previously flagged, now resolved:**
- SHA mismatch (the one item I'd held for): the shipped pin `4d081325…` and the PR body's testing section now agree, and the body cites the recorded uv archive's `direct_url.json` (`commit_id 4d081325…`) as the build the scored runs executed; `9dcca2ae` is explained as a different, earlier archive. Settled with artifact evidence rather than deferred.
- Stale 0.3.x labels: `SKILL.md` is now in the diff (all three mentions relabelled), and `example_eval_next.yaml:12` no longer says "validate against 0.3.0" — it now points at "the pinned nel-next build". Script header, reference, template and both recipes are consistent at 0.4.x.
- `NEL_NEXT_SPEC` precedence: documented in the venv section of `references/nel-next.md`, including that a leftover `==0.3.*` in `.env` silently overrides the git origin *and* its pin and loses the vendored TB 2.1 registry override.
- Instruction-template acquisition: two concrete routes (reference run dir's `container_mounts` entry, or the `bench.yaml` `instruction_template` path), an explicit "internal, no public download" statement, and `sha256sum` on both ends plus the "same file on both sides of a comparison" requirement.
- Earlier items confirmed still fixed: hard-SHA pin + rationale comment, `--version` printing the resolved spec, quoted flow-mapping YAML throughout, `http_pairs_dump` commented out and documented as canary-only in all three sites, merged `cluster.eval_image` row in `terminal_bench_2_1.md`, `reasoning_replay.mode` default wording, unconditional `model_traffic` exclude framing, and `cluster.container_mounts` in the schema block.

**Minor, left open deliberately (no action needed for approval):** `!.env.example` in `.gitignore` (nothing shadowed today — the committed template is `recipes/env.example`, and there is no tracked root `.env*`); the Fargate quota is still not named at the two "check `N × concurrency`" sites; a user-supplied `NEL_NEXT_SPEC` that omits `[export]` would break `mlflow-push`, though the documented example spells the extras out; CodeRabbit's `nemo-evaluator-next` → `nemo-evaluator` naming nit is cosmetic since the skill uses it as a label, not an install spec.

Complex PR: spans 6 directories (≥ 5). Looping in a human for approval.

.agents/skills/evaluation/references/nel-next.md
```diff
@@ -131,9 +145,25 @@ with its own `run_id`, copying the shared `services:` block.
 
 ## Rules & gotchas
 
-- **`eval_image`** = `${NEL_NEXT_EVAL_IMAGE}`. `0.3.1.1-harbor` is multi-arch and is
-  the minimum for **TB 2.1**; older `0.17.x/0.18.x-harbor-<arch>` are arch-suffixed.
+- **`eval_image`** = `${NEL_NEXT_EVAL_IMAGE}` → `0.5.0.1-harbor` (multi-arch). Re-check
```
Edwardf0t1, line 157: Worth one sentence recording *why* this pair is safe, rather than a change.

The login-side CLI is now pinned to nemo-evaluator **0.4.0** (`nel-next.sh` `NEL_NEXT_REF=4d081325`), and it renders `config_<bench>.yaml` from the benchmark model dump; the compute node then parses that file inside `eval_image` = **0.5.0.1-harbor**. All the config models are `extra="forbid"` and the dump carries non-`None` defaults, so a field renamed or dropped between 0.4.x and 0.5.x surfaces as a pydantic `ValidationError` *after* the nodes are allocated and the checkpoint is in vLLM — and `--dry-run` can't catch it, since it validates against 0.4.0 on the login node. Before this PR the pair was matched (0.3.x CLI + `0.3.1.1-harbor`).

The strong counter-argument is that the TB2.1/SWE-bench parity runs passed on exactly this pair, so the combination is empirically validated even though the version numbers look two minors apart. If that's the reasoning, stating it here ("0.4.0 client + 0.5.0.1 image is the validated combination — the image is on its own release train, not the client's") turns a number mismatch that a future reader will trip over into a documented invariant, and tells them what to re-check when either side moves.

.agents/skills/evaluation/recipes/tasks/aa_next/terminal_bench_2_1.md
```diff
@@ -20,16 +20,21 @@ pinned via a vendored registry override in nemo-evaluator-next.
 | sandbox | `ecs_fargate`, `stateful: true` (agent + verifier share one container) |
 | `sandbox.region` | match the region in `${HARBOR_ECR_REPOSITORY}` (us-east-1 with the eval-config default) |
 | `sandbox.ecr_repository` | `${HARBOR_ECR_REPOSITORY}` — set by the `modelopttools:eval-config` skill (internal harbor ECR) |
-| `cluster.eval_image` | `${NEL_NEXT_EVAL_IMAGE}` — set by `modelopttools:eval-config` |
 | `cluster.container_env.AWS_DEFAULT_REGION` | match `sandbox.region` |
 | `max_concurrent` / `sandbox.concurrency` | `50` (canonical bench.yaml) |
 | timeout_strategy | `max` (canonical bench.yaml) + `agent_kwargs.llm_kwargs.timeout: 3600`; use `task` for leaderboard-comparable |
-| `cluster.eval_image` requirement | **≥ `0.3.1.1-harbor`** — TB 2.1's task set is pinned via a vendored registry override in that image (`${NEL_NEXT_EVAL_IMAGE}`, multi-arch) |
+| `cluster.eval_image` | **`0.5.0.1-harbor`** (`${NEL_NEXT_EVAL_IMAGE}`, multi-arch) |
```
Edwardf0t1: Optional, and squarely the owner's call — but this PR exists partly because the previous shared value (`≥ 0.3.1.1-harbor`) had to be chased through three files, and the change re-creates that shape one level up.

After this PR `0.5.0.1-harbor` appears in 5 places (`references/nel-next.md:148`, `example_eval_next.yaml:82`, this row, `terminal_bench_2_1.md:58`, `swebench_verified.md:23`) and `exclude_patterns: ["shard*", "model_traffic.jsonl"]` in 4 (`nel-next.md:157`, `example_eval_next.yaml:110`, `terminal_bench_2_1.md:29`, `swebench_verified.md:28`). Same for `request_timeout: 3600` and the four-entry `drop_params` list. None of these are per-benchmark deltas — they're shared defaults, and both task files already point at `references/nel-next.md`.

The failure mode is quiet: the next image bump or `drop_params` change updates some copies, and a stale one becomes the value someone uses for a scored run without any signal. If the per-benchmark tables are meant as a standalone checklist, keeping them but marking the shared rows ("shared — see `references/nel-next.md`") would get most of the protection without restructuring.

Edwardf0t1 (commented):
Two last items from the review pass, both non-blocking — the substantive findings are all addressed now, and the SHA resolution via the run's uv `direct_url.json` in 96cfc99f is a better answer than the pin I pushed (I picked `main`'s HEAD, you picked the build that actually ran).

cjluo-nv:
### All three addressed in a5cdf278

**@Edwardf0t1 — client 0.4.x + image 0.5.0.1-harbor.** Documented as an invariant in
`references/nel-next.md`, with your reasoning: the image is on its own release train, the
client renders `config_<bench>.yaml` and the *image* parses it, every model is
`extra="forbid"`, so a renamed/dropped field between the two would surface as a pydantic
`ValidationError` only after allocation and checkpoint load — and client-side `--dry-run`
can't catch it. Stated plainly that the pair is safe **because the parity runs executed it,
not by construction**, with "re-canary whenever either side moves" as the trigger. You were
right that this was a number mismatch a future reader would trip over; before this PR the pair
was matched.

**@Edwardf0t1 — duplicated shared values.** Verified the counts: `0.5.0.1-harbor` and
`no_rebuild` in 4 files each, `model_traffic.jsonl` in 4, `request_timeout: 3600` in 3. Took
the cheap option you offered — the per-benchmark tables stay as standalone checklists, but the
four shared rows in each are now marked *(shared — see `references/nel-next.md`)*, so a stale
copy is visibly a copy. Fair hit that this PR re-created the scattered-value shape it set out
to remove.

While marking them I found that the `exclude_patterns` correction (capture is on by default →
the exclude is unconditional) had landed in `swebench_verified.md` but was **lost from
`references/nel-next.md`** when the branch was reset onto a6ddca20. Restored. That is the
second fix lost to that reset, which is itself an argument for your point about shared values
living in one place.

**@coderabbitai — package naming.** Partially applied. The pip package is
`nemo-evaluator[harbor,export]`, so the two spots that meant *the package* are corrected. But
`nemo-evaluator-next` is also the eval **image** repo
(`gitlab-master.nvidia.com/…/nemo-evaluator-next:0.5.0.1-harbor`) and the upstream git repo —
a blanket replace would corrupt the image path, so image/repo references are unchanged.

chadvoegele (approved):

PR #2066, opened 2026-08-04
Use lm-eval 0.4.12's built-in trtllm backend, deprecate lm_eval_tensorrt_llm.py

cjluo-nv:
### What does this PR do?

Type of change: documentation / example update (with a behaviour fix)

lm-evaluation-harness **0.4.12** is the first release that ships a TensorRT-LLM backend
(`lm_eval.models.trtllm_causallms`, registered as `trtllm`) — it is absent in 0.4.10 and
0.4.11. This example no longer maintains its own, so:

- Pin `lm_eval[api,ifeval]>=0.4.12,<0.5` (the 0.5.0.dev line drops the file) and bump
  `lm_eval_hf.py`'s version guard to match.
- **Delete** `examples/llm_eval/lm_eval_tensorrt_llm.py` (the `trt-llm` model). Replace
  `python lm_eval_tensorrt_llm.py --model trt-llm --model_args tokenizer=<tok>,checkpoint_dir=<ckpt>`
  with `python lm_eval_trtllm.py --model trtllm --model_args model=<ckpt>,tokenizer=<tok>`.
- Add `examples/llm_eval/lm_eval_trtllm.py`, whose entire content is one corrected
  `_parse_logprobs` plus `cli_evaluate()` (see below). `lm_eval_hf.py` stays HF-only.
- `examples/hf_ptq/scripts/huggingface_example.sh` and the docs use the upstream backend.
  `parser.sh` gains `--input` (`BUILD_MAX_INPUT_LEN`, default 4096) — it already *echoed*
  that variable but never parsed or defaulted it, so it printed empty on every run.

#### Why `lm_eval_trtllm.py` exists: an upstream off-by-one

TensorRT-LLM aligns `prompt_logprobs` to the *next* token. `executor/base_worker.py`:

```python
# Pass prompt_token_ids with an offset of 1 for correct mapping to the context logits
prompt_token_ids = generation_result._generation_request.prompt_token_ids[1:] + first_generation_token
```

So entry `i` is the distribution that predicted `tokens[i + 1]`, and `_topk_logprobs`
appends that token's id when it is not in the top-k. lm-eval's `_parse_logprobs` instead
reads `prompt_logprobs[i][tokens[i]]` and applies its own shift on top, which raises
`KeyError` on the **first request of every loglikelihood task** (hellaswag, mmlu, arc, ...):

```
File ".../lm_eval/models/trtllm_causallms.py", line 324, in _parse_logprobs
    current_token_logprob = prompt_logprob[tokens[i]]
KeyError: 6503
```

Probed against TRT-LLM 1.3.0rc23 with a 14-token prompt for `prompt_logprobs` 0, 1 and 2:
`tokens[i]` is missing at **every** position, `tokens[i+1]` is present at every position.
Only `generate_until` tasks work unpatched. **This wants an upstream issue against
EleutherAI/lm-evaluation-harness.**

The override also fails loudly rather than quietly: it checks `prompt_logprobs` covers
every prompt token and raises on a missing token, instead of skipping the term and
silently inflating the reported accuracy.

#### Defaults that must be set explicitly

`TRTLLM.__init__` accepts `**kwargs` but forwards only a fixed set to the **TensorRT-LLM
`LLM` API**, so extra `--model_args` aimed at the engine are silently dropped. (lm-eval's
own named parameters — `max_gen_toks`, `batch_size`, `truncation_side`, ... — are honored
normally.) Two engine defaults are unsafe for few-shot eval:

- `tensor_parallel_size` defaults to **1** (the deleted wrapper used every visible GPU).
- `max_input_len` defaults to **2048**, and longer prompts are silently left-truncated —
  5-shot MMLU/gsm8k prompts exceed that.

### Usage

```bash
python lm_eval_trtllm.py --model trtllm \
    --model_args model=<quantized checkpoint dir>,tokenizer=<HF model folder>,tensor_parallel_size=<tp>,max_batch_size=<bs>,max_input_len=4096,max_output_len=512 \
    --tasks hellaswag,gsm8k \
    --batch_size <bs>
```

Flat arguments (no `run` subcommand) are what 0.4.12's `HarnessCLI.parse_args` inserts
`run` for automatically (`_cli/harness.py:48-51`); this is the exact command form used for
the results below.

### Testing

**Unit** — `tests/examples/llm_eval/test_lm_eval_trtllm.py`, no GPU and no `tensorrt_llm`
install: stubs the response object and pins the `i-1` alignment, the `rank != 1` →
`is_greedy` rule, the `ctxlen=0` edge, and both `RuntimeError` paths. Mutation-checked —
dropping the `-1` shift is caught by 5/5 cases, ignoring `ctxlen` by 4/5. A sixth test is a
**tripwire**: it asserts lm-eval's own implementation is still misaligned, so a future
0.4.x that fixes the bug fails the test and says to delete this file rather than being
silently re-broken by the override.

**End to end** — `nvidia/Qwen3.5-122B-A10B-NVFP4` (NVFP4 MoE, 256 experts) on **4x B300**,
TRT-LLM 1.3.0rc23, lm-eval 0.4.12, `--limit 32`:

| run | hellaswag acc | hellaswag acc_norm | gsm8k flexible | gsm8k strict |
|---|---|---|---|---|
| deleted impl (`trt-llm`), tp=4 | 0.7188 | 0.7812 | 0.8438 | 0.7812 |
| `lm_eval_trtllm.py`, tp=1 | 0.7188 | 0.7812 | 0.8438 | 0.8125 |
| `lm_eval_trtllm.py`, tp=2 | 0.7188 | 0.7812 | 0.9062 | 0.8125 |
| `lm_eval_trtllm.py`, tp=4 | 0.7188 | 0.7812 | 0.8750 | 0.8438 |

- hellaswag (the loglikelihood path this PR fixes) is **identical at every tp and identical
  to the deleted implementation** — the alignment fix is exact, not approximate.
- gsm8k varies by 1–2 samples out of 32 (generation path: upstream uses native `stop=`
  sequences and per-request `SamplingParams`; the old wrapper used beam-search-of-1 with
  post-hoc string truncation).
- Without the override, every hellaswag run above dies with the `KeyError`.
- Re-verified at tp=4 after the code moved out of `lm_eval_hf.py` into `lm_eval_trtllm.py`.

Note: NVFP4 fused-MoE has no CUTLASS tactic on Hopper (`No supported MoE GEMM tactic
remains after replacing unsupported NO_SMEM epilogues.`), so this had to be validated on
Blackwell.

### Feature parity notes

Gained from upstream: `loglikelihood_rolling` (was `NotImplementedError`), pipeline
parallelism, `add_bos_token` auto-detection, prompt truncation, per-request sampling params,
`prompt_logprobs` instead of full-vocab context logits (much lower memory), thinking-tag
handling, `batch_size=auto`.

Not reachable through the upstream backend (were set by `modelopt.deploy.llm.LLM`):
`enable_attention_dp` for MoE, `CudaGraphConfig`, `enable_chunked_prefill`,
`moe_expert_parallel_size=1`, and `free_gpu_memory_fraction=0.7` with a capped
`kv_cache.max_tokens` — upstream uses the TRT-LLM default 0.9 (observed allocating 218 GiB
of paged KV cache on B300), so OOM risk is higher on smaller GPUs. This is documented in
`examples/llm_eval/README.md`, and `huggingface_example.sh` honours a preset `LM_EVAL_TP`
so users can lower the tensor-parallel size without editing the script.

- Is this change backward compatible?: ❌ — `lm_eval_tensorrt_llm.py` is removed and the CLI changes (`--model trt-llm` → `trtllm`, `checkpoint_dir=` → `model=`). Migration command is in the README and CHANGELOG.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ — no new dependency; existing `lm_eval` pin tightened.
- Did you write any new necessary tests?: ✅ — `tests/examples/llm_eval/test_lm_eval_trtllm.py` (6 cases, no GPU).
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — under 0.47 *Deprecations*.
- Did you get Claude approval on this PR?: ✅ — reviewed, feedback addressed in `dcedd37b4` and `622b97c26`.

## Summary by CodeRabbit

* **New Features**
  * Added TensorRT-LLM evaluation through lm-evaluation-harness’s `trtllm` backend.
  * Added configurable input/output lengths, batching, tensor parallelism, and build input length.
  * Improved prompt log-probability alignment for more accurate evaluation results.

* **Documentation**
  * Updated evaluation instructions, truncation guidance, backend limitations, and configuration examples.

* **Deprecations**
  * Removed the legacy TensorRT-LLM evaluation script and entry point.

* **Updates**
  * lm-evaluation-harness now requires versions 0.4.12 through 0.4.x.


kevalmorabia97 (approved):
LGTM. Please review unresolved coderabbit / claude contents

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -297,11 +297,16 @@ if [[ $TASKS =~ "lm_eval" ]]; then
...
-    python lm_eval_tensorrt_llm.py \
-        --model trt-llm \
-        --model_args tokenizer=$MODEL_PATH,checkpoint_dir=$SAVE_PATH,max_gen_toks=$BUILD_MAX_OUTPUT_LEN \
+    echo "Using the following config: max output $BUILD_MAX_OUTPUT_LEN max batch $BUILD_MAX_BATCH_SIZE tp $LM_EVAL_TP"
+
+    # max_input_len defaults to 2048, which silently truncates 5-shot prompts; size the
+    # engine for a 4096-token context plus the requested generation length.
+    python lm_eval_hf.py \
+        --model trtllm \
+        --model_args model=$SAVE_PATH,tokenizer=$MODEL_PATH,tensor_parallel_size=$LM_EVAL_TP,max_batch_size=$BUILD_MAX_BATCH_SIZE,max_gen_toks=$BUILD_MAX_OUTPUT_LEN,max_input_len=4096,max_output_len=$BUILD_MAX_OUTPUT_LEN \
```
cjluo-nv: do we have BUILD_MAX_INPUT_LEN here as well?

examples/llm_eval/lm_eval_hf.py
```diff
@@ -292,6 +297,39 @@ def _inject_modelopt_args_into_model_args(args):
     args.model_args = model_args
 
 
+def _patch_trtllm_parse_logprobs():
```
cjluo-nv: is this a bug with the official TRTLLM backend?

examples/llm_eval/lm_eval_hf.py
```diff
@@ -331,6 +369,9 @@ def _enforce_accuracy_gate(output_path, task, lower_bound):
     _add_modelopt_args(run_parser)
     args = cli.parse_args()
     _inject_modelopt_args_into_model_args(args)
+    if getattr(args, "model", None) == "trtllm":
```
cjluo-nv: why we handle trtllm here? This should be hf only.

examples/llm_eval/lm_eval_tensorrt_llm.py
```diff
@@ -13,201 +13,114 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
-import gc
-import logging
+"""[Deprecated] ModelOpt's TensorRT-LLM backend for lm-evaluation-harness.
```
cjluo-nv: I think it is ok to deprecate this file entirely

examples/hf_ptq/scripts/huggingface_example.sh
cjluo-nv: Good catch — and it was more broken than it looked. `parser.sh:163` already **echoed** `BUILD_MAX_INPUT_LEN`, but it was never in the getopt list, never parsed and never defaulted, so it printed empty on every run.

`dcedd37b4` wires it up: `--input` -> `BUILD_MAX_INPUT_LEN`, `DEFAULT_BUILD_MAX_INPUT_LEN=4096`, and the lm_eval command now passes `max_input_len=$BUILD_MAX_INPUT_LEN` instead of the hardcoded 4096. The config echo reports it too.

examples/llm_eval/lm_eval_hf.py
cjluo-nv: Yes — a genuine bug in lm-eval 0.4.12's backend, confirmed both in source and at runtime.

TensorRT-LLM aligns `prompt_logprobs` to the **next** token. `tensorrt_llm/executor/base_worker.py::_compute_pytorch_prompt_logprobs` does:

```python
# Pass prompt_token_ids with an offset of 1 for correct mapping to the context logits
prompt_token_ids = generation_result._generation_request.prompt_token_ids[1:] + first_generation_token
```

so entry `i` is the distribution that predicted `tokens[i + 1]`, and `_topk_logprobs` appends that token's id when it is not in the top-k. lm-eval instead reads `prompt_logprobs[i][tokens[i]]` and applies its own shift on top of that.

Probed directly against TRT-LLM 1.3.0rc23 with a 14-token prompt, for `prompt_logprobs` 0, 1 and 2 — `tokens[i]` is missing at **every** position, `tokens[i+1]` is present at every position. So it is not an edge case: it raises `KeyError` on the first request of every loglikelihood task (hellaswag, mmlu, arc, winogrande, ...). Only `generate_until` tasks work unpatched.

The corrected indexing now lives in `lm_eval_trtllm.py` and reproduces the removed implementation's hellaswag numbers exactly (0.7188 / 0.7812 at tp=1/2/4).

examples/llm_eval/lm_eval_hf.py
cjluo-nv: Agreed — moved in `dcedd37b4`. `lm_eval_hf.py` is back to HF-only: the `_patch_trtllm_parse_logprobs` helper, its call site, and the `logging` import are all gone.

The override now lives in the new `examples/llm_eval/lm_eval_trtllm.py`, whose entire content is that one corrected `_parse_logprobs` plus `cli_evaluate()`. `huggingface_example.sh` and the docs call it instead.

examples/llm_eval/lm_eval_tensorrt_llm.py
cjluo-nv: Done — deleted outright in `dcedd37b4`, no translating shim.

`examples/llm_eval/README.md` and the CHANGELOG entry carry the replacement command:
`python lm_eval_tensorrt_llm.py --model trt-llm --model_args tokenizer=<tok>,checkpoint_dir=<ckpt>`
-> `python lm_eval_trtllm.py --model trtllm --model_args model=<ckpt>,tokenizer=<tok>,tensor_parallel_size=<tp>,max_input_len=4096`.

examples/hf_ptq/scripts/huggingface_example.sh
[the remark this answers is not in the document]
cjluo-nv: Both valid — fixed in `dcedd37b4`. The block runs after `pushd ../llm_eval/`, so a relative `--model` would not resolve; switched the tokenizer to `MODEL_ABS_PATH` (matching the mmlu block) and quoted `--model_args`.

The entry point also changed to `lm_eval_trtllm.py`, and `max_input_len` now comes from the new `--input` / `BUILD_MAX_INPUT_LEN` option rather than being hardcoded.

examples/llm_eval/lm_eval_hf.py
```diff
@@ -292,6 +297,39 @@ def _inject_modelopt_args_into_model_args(args):
...
+        prompt_logprobs = outputs.outputs[0].prompt_logprobs
+        continuation_logprobs = 0.0
+        is_greedy = True
+        # Token 0 has no preceding distribution, so it can never be scored.
+        for i in range(max(ctxlen, 1), len(tokens)):
+            logprob = prompt_logprobs[i - 1].get(tokens[i])
+            if logprob is None:
+                # TRT-LLM always appends the actual token to the top-k dict.
+                logger.warning(f"Token {tokens[i]} missing from prompt_logprobs[{i - 1}]")
+                continue
```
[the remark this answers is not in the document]
cjluo-nv: The message no longer contains the token id — it reports the position only. Note the code also moved: `lm_eval_hf.py` is HF-only again and this now lives in `examples/llm_eval/lm_eval_trtllm.py`, where it raises rather than logs:

```
f"tokens[{i}] is missing from prompt_logprobs[{i - 1}]; the returned "
"logprobs are misaligned with the requested tokens."
```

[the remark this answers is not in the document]
cjluo-nv: Took the hardening, pushing back on the two premises.

**Taken** (`dcedd37b4`): the override now checks `len(prompt_logprobs) >= len(tokens) - 1` and **raises** on a missing token instead of `continue`. You are right that silently dropping a term inflates the score, which is the worst failure mode for an eval script.

**Premise 1 — `maybe_truncate` shifting the indices — does not apply to this path.** `maybe_truncate` is only called in `generate_until`. `_loglikelihood_tokens` truncates the request itself:

```python
inp = (context_enc + continuation_enc)[-(self.max_length) :]
```

and passes that same `inp` as `tokens` to `_parse_logprobs`. The engine and the caller therefore see the same token list by construction; they cannot disagree about truncation.

**Premise 2 — `enable_block_reuse` — is factually wrong.** The upstream backend does have the equivalent knob and already sets it, at `trtllm_causallms.py:139-142`:

```python
# If reuse kv cache, no logprobs can be computed.
kv_cache_config = KvCacheConfig(
    enable_block_reuse=False,
)
```

So the invariant the deleted wrapper enforced via `enable_kv_cache_reuse=False` is still enforced, by upstream, for the same stated reason.

The length check goes in anyway as cheap insurance, but neither named mechanism can produce the misalignment.

examples/hf_ptq/scripts/huggingface_example.sh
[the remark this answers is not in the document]
cjluo-nv: Did both, in `dcedd37b4`.

Escape hatch: `LM_EVAL_TP=${LM_EVAL_TP:-$(python -c "import torch; print(max(torch.cuda.device_count(), 1))")}`, with a comment naming the DeepEP/SM 12.0 failure.

Docs: `examples/llm_eval/README.md` gains a fourth NOTE covering the tuning the upstream backend does not expose — expert parallelism left at the TRT-LLM default, and the KV cache at 90% of free memory rather than the wrapper's 70% — telling users to lower `tensor_parallel_size` if they hit either. Agreed the caveat should not have lived only in the PR body.

examples/llm_eval/lm_eval_tensorrt_llm.py
```diff
@@ -13,201 +13,114 @@
...
+    found = _find_option(out, "--batch_size")
+    batch_size = found[1] if found else None
+
+    found = _find_option(out, "--model")
+    if found and found[1] in ("trt-llm", "trt_llm"):
+        _set_option(out, "--model", found[0], "trtllm")
+
+    found = _find_option(out, "--model_args")
+    if found:
+        _set_option(out, "--model_args", found[0], _translate_model_args(found[1], batch_size))
```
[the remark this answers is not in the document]
cjluo-nv: Moot — `lm_eval_tensorrt_llm.py` was deleted outright in `dcedd37b4` rather than kept as a translating shim, so there is no `_translate` left to gate.

examples/llm_eval/lm_eval_tensorrt_llm.py
```diff
@@ -13,201 +13,114 @@
...
+
+    max_gen_toks = int(args.get("max_gen_toks", _DEFAULT_MAX_GEN_TOKS))
+    # Upstream defaults to max_input_len=2048, short enough to silently left-truncate
+    # 5-shot prompts, so always derive it from the legacy sizing instead.
+    args.setdefault("max_output_len", str(max_gen_toks))
+    max_length = args.pop("max_length", None)
+    args.setdefault(
+        "max_input_len",
+        str(int(max_length) - max_gen_toks) if max_length else str(_DEFAULT_MAX_INPUT_LEN),
+    )
```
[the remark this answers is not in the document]
cjluo-nv: Moot — `lm_eval_tensorrt_llm.py` was deleted outright in `dcedd37b4`, so the `max_gen_toks` passthrough and the `int()` / negative-`max_input_len` edge cases go with it. `max_input_len` is now supplied directly by the caller (`--input` / `BUILD_MAX_INPUT_LEN`).

examples/llm_eval/lm_eval_hf.py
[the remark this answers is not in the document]
cjluo-nv: Moot — the patch is no longer in `lm_eval_hf.py`, so there is no `args.model` string guard to normalize. `lm_eval_trtllm.py` installs the override unconditionally at import, which also removes the failure mode you describe (a guard that silently does not fire).

examples/hf_ptq/scripts/huggingface_example.sh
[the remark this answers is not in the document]
cjluo-nv: Both fixed in `dcedd37b4` (same change as the CodeRabbit thread on this line): `tokenizer=$MODEL_ABS_PATH` and the whole `--model_args` value quoted.

Correct that it was inherited — the deleted `lm_eval_tensorrt_llm.py` invocation had the same `$MODEL_PATH` bug — but the line was being rewritten anyway, so it is fixed rather than carried forward.

`max_input_len` is no longer hardcoded either; it comes from the new `--input` / `BUILD_MAX_INPUT_LEN` option, which `parser.sh` echoed but never actually parsed or defaulted.

examples/llm_eval/lm_eval_trtllm.py
```diff
@@ -0,0 +1,84 @@
...
+
+    python lm_eval_trtllm.py --model trtllm \
+        --model_args model=<quantized checkpoint dir>,tokenizer=<HF model folder>,\
+tensor_parallel_size=<tp>,max_batch_size=<max batch size>,max_input_len=4096 \
+        --tasks <comma separated tasks> --batch_size <max batch size>
+"""
+
+from importlib.metadata import version
+
+from lm_eval.__main__ import cli_evaluate
```
meenchen, line 32: > _Bot comment._

`cli_evaluate` on lm-eval 0.4.12: `lm_eval_hf.py` (same directory, same version floor) has to hook `HarnessCLI._subparsers.choices["run"]`, which implies 0.4.12's CLI is subcommand-based. Has `python lm_eval_trtllm.py --model trtllm --tasks ... --batch_size ...` (flat args, no `run` subcommand) actually been run against 0.4.12? The PR body's usage/testing sections both reference `lm_eval_hf.py --model trtllm` plus a deprecation shim, neither of which exists in this diff, so the shipped entry point looks unvalidated — and `test_qwen3_eval_fp8` is the only thing covering it.

examples/llm_eval/lm_eval_trtllm.py
```diff
@@ -0,0 +1,84 @@
...
+                f"tokens[{i}] is missing from prompt_logprobs[{i - 1}]; the returned "
+                "logprobs are misaligned with the requested tokens."
+            )
+        continuation_logprobs += logprob.logprob
+        if logprob.rank != 1:
+            is_greedy = False
+
+    return continuation_logprobs, is_greedy
+
+
```
meenchen, line 113: > _Bot comment._

Two robustness points on the monkeypatch:

1. It silently no-ops semantics if upstream changes: with only a `>=0.4.12` floor and a `<0.5` pin, a 0.4.13 that *fixes* `_parse_logprobs` would be silently overridden by this patch (double-shifting again), and a rename/signature change would surface as an obscure failure. Consider asserting the attribute exists (and ideally a version ceiling / a comment tying removal to the upstream fix).
2. There is no test for `_parse_logprobs` itself. This is the only new logic in the PR and needs no GPU — a stub `outputs` (`prompt_logprobs` list of `{token_id: Logprob(logprob, rank)}`) can pin the `i-1` alignment, the `is_greedy`/`rank != 1` rule, the `max(ctxlen, 1)` edge case, and both `RuntimeError` branches. Please add one; the GPU example test's reference numbers were produced with a different implementation.

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -297,11 +297,18 @@ if [[ $TASKS =~ "lm_eval" ]]; then
...
+    # which fails in DeepEP kernels for MoE checkpoints on some GPUs (e.g. SM 12.0).
+    LM_EVAL_TP=${LM_EVAL_TP:-$(python -c "import torch; print(max(torch.cuda.device_count(), 1))")}
+
+    echo "Using the following config: max input $BUILD_MAX_INPUT_LEN max output $BUILD_MAX_OUTPUT_LEN max batch $BUILD_MAX_BATCH_SIZE tp $LM_EVAL_TP"
+
+    # max_input_len defaults to 2048, which silently truncates 5-shot prompts, so pass it
+    # explicitly; the engine's max_seq_len is max_input_len + max_output_len.
+    python lm_eval_trtllm.py \
+        --model trtllm \
+        --model_args "model=$SAVE_PATH,tokenizer=$MODEL_ABS_PATH,tensor_parallel_size=$LM_EVAL_TP,max_batch_size=$BUILD_MAX_BATCH_SIZE,max_gen_toks=$BUILD_MAX_OUTPUT_LEN,max_input_len=$BUILD_MAX_INPUT_LEN,max_output_len=$BUILD_MAX_OUTPUT_LEN" \
```
meenchen, line 311: > _Bot comment._

`max_gen_toks=$BUILD_MAX_OUTPUT_LEN` is carried over from the old backend, but the PR body states the upstream `TRTLLM.__init__` only forwards a fixed set of kwargs and silently drops the rest. Is `max_gen_toks` in that set? If it is dropped, generation length is no longer capped by `--output`, which is exactly what `test_qwen3_eval_fp8` relies on (`output=128` "Cap generation length: gsm8k/humaneval otherwise generate up to 1024 tokens/sample") — the test would get slower and the updated comment would be misleading. Worth confirming, and dropping the arg if it's a no-op.

meenchen (commented):

**Design review (protocol applied).** Problem: the repo maintained its own `trt-llm` lm-eval backend (`lm_eval_tensorrt_llm.py`, 213 lines on top of `modelopt.deploy.llm.LLM`); lm-eval 0.4.12 now ships `lm_eval.models.trtllm_causallms`, so the local copy is redundant. Deleting it in favour of upstream is the right direction — this PR *removes* a subsystem rather than adding one, and the net is -231/+152. Alternatives considered: (a) keep the local backend (rejected in the body with a feature-parity table — reasonable), (b) fix upstream and pin (the body says an upstream issue/PR is still needed and marks the shim as temporary — reasonable), (c) put the `_parse_logprobs` patch inside the existing `lm_eval_hf.py` entry point instead of adding a second entry script — the PR body actually *describes* option (c) as what was implemented, but the diff implements a separate `lm_eval_trtllm.py`. That gap is my main concern.

Findings:

1. **The PR body no longer describes the diff, and the testing evidence is therefore unattributable.** The body says `lm_eval_tensorrt_llm.py` "becomes a **deprecation shim**", that "`lm_eval_hf.py` patches `_parse_logprobs` when `--model trtllm` is selected", that the usage is `python lm_eval_hf.py --model trtllm ...`, that "the deprecation shim was run end to end", and answers "Is this change backward compatible?: ✅". The diff instead *deletes* the script (correctly reflected in `CHANGELOG.rst`, so the CHANGELOG and the body disagree) and adds a new `lm_eval_trtllm.py` that calls `lm_eval.__main__.cli_evaluate`. So the shipped code path — a flat `python lm_eval_trtllm.py --model trtllm --tasks ...` invocation through `cli_evaluate` on 0.4.12 — is not the one the accuracy tables were produced with. Please either update the body or re-state which command was actually run.

2. **`cli_evaluate` vs 0.4.12's subcommand CLI is unverified (CI-breaking if wrong).** `lm_eval_hf.py` has to reach into `HarnessCLI._subparsers.choices["run"]`, i.e. 0.4.12's CLI is subcommand-based. It's not obvious that `lm_eval.__main__.cli_evaluate` still accepts flat `--model/--tasks/--batch_size` args in 0.4.12; if it doesn't, `test_qwen3_eval_fp8` (the only coverage) fails. Related: `huggingface_example.sh` forwards `$lm_eval_flags`, which can contain `--trust_remote_code` — `lm_eval_hf.py` goes out of its way to move that flag into `model_args` and null it out in the namespace, so please confirm the plain CLI accepts it and that it actually reaches the trtllm backend/tokenizer.

3. **No unit test for the off-by-one fix**, which is the only new logic in the PR and is pure Python. A stub `outputs` object (`SimpleNamespace(outputs=[SimpleNamespace(prompt_logprobs=[{tok: Logprob(logprob, rank)}...])])`) would pin the alignment, the `is_greedy` computation, and both `RuntimeError` paths without a GPU. Right now the fix is only exercised by a GPU/TRT-LLM example test whose reference numbers came from a different implementation (see 1).

4. Minor: duplicated version guard between `lm_eval_hf.py` and `lm_eval_trtllm.py`, and the trtllm path loses `lm_eval_hf.py`'s extras (`--accuracy_lower_bound`, trust-remote-code plumbing). Folding the patch into `lm_eval_hf.py` under `args.model == "trtllm"` (as the body describes) would avoid both.

Licensing: new file carries the standard NVIDIA/Apache-2.0 header (year 2024 vs. `LICENSE_HEADER`'s 2026 — cosmetic, matches existing files); no licensing surface. Doc/CHANGELOG references to the removed script are consistent — no dangling `lm_eval_tensorrt_llm.py` invocations remain in the repo.

examples/llm_eval/lm_eval_trtllm.py
cjluo-nv: Yes — that exact flat-arg form is what produced the tp=1/2/4 numbers in the PR body. It works because 0.4.12's CLI keeps a legacy path in `HarnessCLI.parse_args` (`lm_eval/_cli/harness.py:48-51`):

```python
if len(sys.argv) > 2 and sys.argv[1] not in self._subparsers.choices:
    # Backward compatibility: arguments provided but no valid subcommand - insert 'run'
    sys.argv.insert(1, "run")
```

`lm_eval_hf.py` reaches for `_subparsers.choices["run"]` for a different reason — it *adds* the ModelOpt arguments (`--quant_cfg`, `--calib_size`, ...) to that subparser. `lm_eval_trtllm.py` adds no arguments, so plain `cli_evaluate()` is enough.

You were right that it looked unvalidated, though: the PR description was stale — it still documented `lm_eval_hf.py --model trtllm` and a deprecation shim, neither of which survived `dcedd37b4`. Rewritten now to match the shipped entry point, and `622b97c26` adds unit coverage so it is not resting on the GPU example test alone.

examples/llm_eval/lm_eval_trtllm.py
cjluo-nv: Both taken, in `622b97c26`.

**Test** — `tests/examples/llm_eval/test_lm_eval_trtllm.py`, no GPU and no `tensorrt_llm` install (the module import is guarded upstream, and `_parse_logprobs` is pure Python over the response object). Stubs `prompt_logprobs` as `{token_id: Logprob(logprob, rank)}` and covers the `i-1` alignment, the `rank != 1` -> `is_greedy` rule, the `max(ctxlen, 1)` edge, and both `RuntimeError` branches. Mutation-checked so the cases are load-bearing: dropping the `-1` shift is caught by 5/5, ignoring `ctxlen` by 4/5.

**Drift** — agreed a silent double-shift on a fixed 0.4.13 is the dangerous case. I did not add a version ceiling, since that would also block unrelated 0.4.x fixes. Instead the module keeps the original in `_UPSTREAM_PARSE_LOGPROBS` and a test asserts it is *still* misaligned:

```python
def test_upstream_is_still_misaligned():
    """Tripwire: when this fails, upstream fixed the bug and this file can be deleted."""
    with pytest.raises(KeyError):
        lm_eval_trtllm._UPSTREAM_PARSE_LOGPROBS(...)
```

So the day upstream fixes it, CI fails and points at the removal rather than silently re-breaking the alignment. Plus a `hasattr` check that errors clearly if `_parse_logprobs` is renamed or removed.

One correction on the aside: the reference numbers were **not** produced by a different implementation. The tp=4 run was re-done against `lm_eval_trtllm.py` exactly as shipped after the code moved out of `lm_eval_hf.py`, and reproduces hellaswag 0.7188 / 0.7812. The unit test is still the right ask — it pins the alignment in a way an end-to-end accuracy number cannot.

examples/hf_ptq/scripts/huggingface_example.sh
cjluo-nv: Checked — `max_gen_toks` is honored, not dropped, so `--output` still caps generation and `test_qwen3_eval_fp8`'s comment stays accurate.

It is an explicit named parameter of `TRTLLM.__init__`, not part of the `**kwargs` that get discarded (`lm_eval/models/trtllm_causallms.py`):

```
:60   max_gen_toks: int = 256,           # named __init__ parameter
:86   self._max_gen_toks = max_gen_toks
:575  def max_gen_toks(self) -> int: return self._max_gen_toks
:411  kwargs, until, max_gen_toks = self.modify_gen_kwargs(
:412      gen_kwargs, eos=eos, default_max_gen_toks=self.max_gen_toks)
:427  SamplingParams(max_tokens=max_gen_toks, stop=until, **kwargs)
```

The PR body caused the confusion and I have fixed it: the "only forwards a fixed set" claim is about the kwargs handed to the **TensorRT-LLM `LLM` API** (`tensor_parallel_size`, `max_input_len`, `kv_cache_config`, ...), where engine-level extras really are silently dropped. lm-eval's own named parameters are consumed by the backend normally. Keeping the argument.

meenchen (commented):

Re-review of PR #2066 (head `b68114bb`). Design question is settled: this *removes* the 213-line local `trt-llm` backend in favour of lm-eval 0.4.12's built-in `trtllm` backend, and the maintainer explicitly asked for the override to live outside `lm_eval_hf.py` ("This should be hf only"), which the current diff does. Net size is small and confined to `examples/` + `tests/`. Licensing: new files carry only the standard NVIDIA/Apache header — no licensing surface. No prompt-injection attempts in the PR content.

**Previous critical findings — resolved**
- :speech_balloon: Body/diff mismatch (shim vs. deleted script, `lm_eval_hf.py --model trtllm` vs. new entry point): body rewritten to describe the shipped `lm_eval_trtllm.py`, CHANGELOG and README agree, no dangling references to the deleted script.
- :speech_balloon: Flat-arg `cli_evaluate` on 0.4.12's subcommand CLI: author cited `lm_eval/_cli/harness.py:48-51` (auto-inserts `run` when argv[1] isn't a subcommand) and states the accuracy tables were produced with exactly that command form. Reasonable.
- :speech_balloon: Missing unit test for the off-by-one: `tests/examples/llm_eval/test_lm_eval_trtllm.py` adds 6 cases pinning the `i-1` alignment, `rank != 1` → `is_greedy`, the `ctxlen=0` edge, both `RuntimeError` paths, and an upstream-still-broken tripwire; follows the existing `test_cast_mxfp4_to_nvfp4.py` sys.path convention. Silent-inflation fallback (`continue`) replaced by a raise plus a length guard, addressing the earlier eval-correctness concern.
- Shell fixes landed: `tokenizer=$MODEL_ABS_PATH`, quoted `--model_args`, `LM_EVAL_TP` override with the DeepEP/SM 12.0 rationale, and `--input`/`BUILD_MAX_INPUT_LEN` wired into `parser.sh` (previously echoed but never parsed — the fix is correct and defaulted to 4096, matching the `test_qwen3_eval_fp8` comment update).

**Still worth an owner look**
- **CI reach of the new tests.** `example_tests.yml` runs `tests/examples/llm_eval` only in the nightly `trtllm-non-pr` matrix; the PR-gated trtllm job is `hf_ptq` only. So these deliberately GPU-free tests — including the tripwire meant to catch a future 0.4.x that fixes upstream, and the assumption that `TRTLLM._parse_logprobs` is a staticmethod with `(tokens, outputs, ctxlen)` — won't run on PRs. Placing them somewhere the PR gate collects would make the protection real.
- :speech_balloon: **`--trust_remote_code`** — the author answered the flat-arg half of the earlier question but not this half. `huggingface_example.sh` still appends `--trust_remote_code` to `$lm_eval_flags`, and it now flows into the plain lm-eval CLI, whereas `lm_eval_hf.py` goes out of its way to rewrite it into `model_args` and null it in the namespace. Behaviour matches the deleted script (also plain `cli_evaluate`), so it's not a regression, but nothing confirms it reaches the trtllm tokenizer, and CI doesn't exercise it.
- **Minor / optional:** the `>=0.4.12` version guard is now duplicated in `lm_eval_hf.py` and `lm_eval_trtllm.py`; the trtllm path has no `--accuracy_lower_bound` equivalent; the guard `len(prompt_logprobs) < len(tokens) - 1` still passes if exactly one prompt token were dropped (the author argues truncation can't occur on the loglikelihood path, which looks right); and the test only does `importorskip("lm_eval")` — `importorskip("lm_eval.models.trtllm_causallms")` would skip rather than error if upstream ever imports `tensorrt_llm` eagerly.

cjluo-nv:
### `--trust_remote_code` — verified, and now pinned by tests (`196c32606`)

Fair hit: nothing confirmed it. It does reach the tokenizer, and on 0.4.12 lm-eval does the rewriting itself, so this entry point does not need `lm_eval_hf.py`'s manual handling:

```
_cli/run.py:355              EvaluatorConfig.from_cli(args)
  -> evaluate_config.py:225    instance._configure()
  -> evaluate_config.py:276    ..._set_trust_remote_code()
  -> evaluate_config.py:420-435
        datasets.config.HF_DATASETS_TRUST_REMOTE_CODE = True
        self.model_args["trust_remote_code"] = True
  -> trtllm_causallms.py:48     named __init__ parameter (not part of the dropped **kwargs)
       :110  AutoTokenizer.from_pretrained(..., trust_remote_code=trust_remote_code)
       :148  llm_kwargs["trust_remote_code"]
```

So it reaches **both** the tokenizer and the TensorRT-LLM engine.

Worth recording the consequence: **`lm_eval_hf.py`'s manual rewriting is now redundant** — it predates lm-eval doing this natively. It is harmless (it populates `model_args`, then nulls the namespace flag so the built-in path no-ops to the same result), so I have left it alone; removing it is a separate cleanup, not this PR's business.

`196c32606` adds three GPU-free tests so this is not resting on my reading of the source: flag set -> injected into `model_args` **and** the datasets flag flips; flag unset -> not injected; and `trust_remote_code` is still a parameter of `TRTLLM.__init__`. That third one matters — without it the first two only prove lm-eval injects *a key*, and an upstream rename would leave it landing in `**kwargs` and being silently dropped.

Not run end to end: that needs a `trust_remote_code` model on Blackwell (fresh allocation, ~10 min load) for a flag whose path is now pinned statically at every hop. Happy to if you want it.

cjluo-nv:
### Remaining items from the re-review

**CI reach of the new tests** — confirmed: `trtllm-pr` is `[hf_ptq]`, `llm_eval` only appears in `trtllm-non-pr`, so these run nightly. Owner call, and the decision is to **leave them nightly**. The reasoning: the tripwire fires on a new lm-eval release rather than on a diff, so a nightly cadence matches its trigger, and the alignment logic is a temporary shim scheduled for deletion once upstream is fixed. Adding `llm_eval` to `trtllm-pr` would also drag `test_qwen3_eval_fp8` (GPU, 900s) onto every PR, which is the split the `pr/non-pr` comment exists to maintain. Filtering per-job would mean editing `_example_tests_runner.yml`, shared by every example — out of blast radius for this PR. Noting the residual gap honestly: a broken edit to `lm_eval_trtllm.py` is not caught by a PR gate beyond ruff.

**Taken** (`d5f868cf1`): `importorskip("lm_eval.models.trtllm_causallms")` instead of `importorskip("lm_eval")`, so collection skips rather than errors if that module ever imports `tensorrt_llm` eagerly.

**Answered, no change:**

- *`len(prompt_logprobs) < len(tokens) - 1` still passes if exactly one prompt token were dropped.* Correct, but the per-token `RuntimeError` covers it: a one-position shift means `tokens[i]` is essentially never present in the shifted dict, and across a full task it is a certainty rather than a probability. Tightening to `== len(tokens)` would make it deterministic, but I have only observed that exact length on TRT-LLM 1.3.0rc23, so it would bet the whole loglikelihood path on one version's response shape. The tolerant length guard plus the loud per-token check fails in the safe direction without that bet.
- *Duplicated `>=0.4.12` guard across the two entry points.* Deduping needs a third shared module to save three lines in each; not worth the indirection for two independent scripts.
- *No `--accuracy_lower_bound` equivalent on the trtllm path.* True, and equally true of the deleted `lm_eval_tensorrt_llm.py` — so not a regression. Reasonable follow-up if you want the gate available for TensorRT-LLM runs.

meenchen (commented):

Re-review of PR #2066 (examples/tests only, +331/−231). Design question stays settled: this **removes** the 213-line in-repo `trt-llm` lm-eval backend in favour of lm-eval 0.4.12's built-in `trtllm` backend, and the only new logic is a narrowly scoped `_parse_logprobs` override in a thin entry point — which is where the maintainer asked it to live ("this should be hf only"). No new subsystem, no licensing surface beyond the standard NVIDIA/Apache header on the new files, and no prompt-injection attempts in the PR content.

**Previously-raised items now resolved**
- `importorskip("lm_eval.models.trtllm_causallms")` replaces the bare `lm_eval` skip, so collection skips rather than errors if the backend ever imports `tensorrt_llm` eagerly (`d5f868c`).
- Off-by-one fix is unit-tested (`tests/examples/llm_eval/test_lm_eval_trtllm.py`): `i-1` alignment, `rank != 1` → `is_greedy`, `ctxlen=0` edge, both `RuntimeError` paths, plus an upstream-still-broken tripwire; the silent-`continue` fallback is gone in favour of a hard raise.
- Shell side is correct: `--input`/`BUILD_MAX_INPUT_LEN` now actually parsed and defaulted in `parser.sh` (it was previously echoed only), `tokenizer=$MODEL_ABS_PATH`, quoted `--model_args`, `LM_EVAL_TP` escape hatch with the DeepEP/SM 12.0 rationale, and the lost `free_gpu_memory_fraction`/expert-parallel tuning documented in `examples/llm_eval/README.md`.

**Still worth an owner look before sign-off**
- :speech_balloon: **CI reach — author decided to keep the new tests nightly-only** (`llm_eval` appears only in the `trtllm-non-pr` matrix; `trtllm-pr` is `hf_ptq`), reasoning that the tripwire fires on new lm-eval releases rather than diffs and that adding `llm_eval` to `trtllm-pr` would drag the 900s GPU `test_qwen3_eval_fp8` onto every PR. Reasonable, but the consequence is that the deliberately GPU-free tests protecting the *only* new logic in this PR never gate a PR (ruff aside), and the tripwire also encodes a version-dependent assumption — that upstream `_parse_logprobs` is a staticmethod callable as `(tokens=, outputs=, ctxlen=)`; if that changes it fails with `TypeError`, not the asserted `KeyError`, and only overnight. Worth a maintainer decision on whether these belong in a PR-run suite.
- :speech_balloon: **`--trust_remote_code`** — author traced the full path (`EvaluatorConfig._set_trust_remote_code` → `model_args["trust_remote_code"]` → named `TRTLLM.__init__` param → tokenizer and `llm_kwargs`) and added three GPU-free tests. Flagging only because the coverage is static (config-level injection + `inspect.signature`), never exercising `TRTLLM` construction, and no e2e run with a `trust_remote_code` model was done — acceptable to me, but it's a security-adjacent flag so an owner should accept the level of assurance.
- **Minor / convention:** `tests/examples/llm_eval/test_lm_eval_trtllm.py` still has function-local imports (`import datasets` + `from lm_eval.config.evaluate_config import EvaluatorConfig` at ~120 and ~135, `import inspect` at ~148) with no stated justification; the module already has a top-level `importorskip` guard, so these can move to module scope after it. Not blocking. Also unaddressed-by-choice and reasonable: the duplicated `>=0.4.12` guard across the two entry points, no `--accuracy_lower_bound` on the trtllm path (parity with the deleted script), and `len(prompt_logprobs) < len(tokens) - 1` tolerating a hypothetical one-token drop (the per-token raise covers it in practice).

examples/llm_eval/lm_eval_trtllm.py
```diff
@@ -0,0 +1,95 @@
...
+        raise RuntimeError(
+            f"prompt_logprobs has {len(prompt_logprobs)} entries for {len(tokens)} tokens; "
+            "the engine scored a different prompt than was requested."
+        )
+
+    continuation_logprobs = 0.0
+    is_greedy = True
+    # Token 0 has no preceding distribution, so it can never be scored.
+    for i in range(max(ctxlen, 1), len(tokens)):
+        logprob = prompt_logprobs[i - 1].get(tokens[i])
```
meenchen, line 100: [P1] Require the TensorRT-LLM version that provides this layout. This `i - 1` mapping only became valid in TensorRT-LLM 1.3.0rc11. The hf_ptq README still recommends `release:1.2.0`; in that version `_compute_pytorch_prompt_logprobs` does not pass prompt token IDs into `compute_logprobs`, so `prompt_logprobs=1` retains only top-1. A normal non-greedy continuation token is therefore absent and this branch raises, breaking likelihood tasks that worked through the deleted wrapper. See [v1.2.0](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.2.0/tensorrt_llm/executor/base_worker.py#L830-L845) versus [v1.3.0rc11](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.3.0rc11/tensorrt_llm/executor/base_worker.py#L846-L871). Please add an explicit `>=1.3.0rc11` runtime guard and update the documented container requirement, or keep a compatible path for the supported 1.1/1.2 releases.

tests/examples/llm_eval/test_lm_eval_trtllm.py
```diff
@@ -0,0 +1,168 @@
...
+
+    `huggingface_example.sh` passes the flag, and unlike `lm_eval_hf.py` this entry point
+    does no rewriting of its own -- it relies on lm-eval doing it. On the live path that is
+    `_cli/run.py:355` -> `EvaluatorConfig.from_cli` -> `_configure()` ->
+    `_set_trust_remote_code()`.
+    """
+    import datasets
+    from lm_eval.config.evaluate_config import EvaluatorConfig
+
+    monkeypatch.setattr(datasets.config, "HF_DATASETS_TRUST_REMOTE_CODE", False)
```
meenchen: [P1] These tests fail with supported `datasets` 4.x because `HF_DATASETS_TRUST_REMOTE_CODE` is not defined until lm-eval assigns it. `monkeypatch.setattr(..., raising=True)` therefore raises `AttributeError` before testing the production behavior. I reproduced this with exact `lm_eval==0.4.12` and `datasets==4.8.4`: 7 tests passed and these two trust-remote-code tests failed. Please use `raising=False` or defensively establish/remove the attribute. The relevant `trtllm-non-pr` suite was skipped by the current PR run, which is why the checks remain green.

examples/llm_eval/lm_eval_trtllm.py
cjluo-nv: Confirmed and fixed in `6a54747c3`. I diffed the two tags rather than taking it on trust, and you are right:

```python
# v1.2.0 -- no prompt token ids
logprobs_result = compute_logprobs(logprob_params.prompt_logprobs, None,
                                   context_logits, None, None)

# v1.3.0rc11 -- prompt token ids passed in
prompt_token_ids = generation_result._generation_request.prompt_token_ids[1:] + first_generation_token
logprobs_result = compute_logprobs(logprob_params.prompt_logprobs, None,
                                   context_logits, None, None,
                                   prompt_token_ids)
```

With `tokens=None`, `_topk_logprobs` never reaches its "append the requested token if it is not in top-k" branch, so `prompt_logprobs=1` keeps only the argmax. A non-greedy continuation token is genuinely absent -- there is no correct value to recover, for this override or for lm-eval's own `_parse_logprobs`. And since `hf_ptq/README.md` recommends `release:1.2.0`, a user following the docs would hit exactly this.

**Guard placement.** I put it in `_parse_logprobs`, not at startup. A blanket startup abort would break generative tasks (gsm8k, ifeval) on 1.1/1.2, which work fine on the old layout and worked through the deleted wrapper — that would be a fresh regression rather than a fix. So:

- `_check_trtllm_version()` raises from `_parse_logprobs` on the first loglikelihood request, checked once.
- `__main__` calls the same function and prints a **warning**, so an unusable container is visible immediately rather than after a multi-minute model load.

Net: loglikelihood on <1.3.0rc11 fails with an explicit "use a newer container" message instead of a confusing missing-token error; generative tasks are untouched.

**Docs**, as you asked: `examples/llm_eval/README.md` and `examples/hf_ptq/README.md` (right under the `release:1.2.0` recommendation) both state the requirement, and the CHANGELOG entry does too.

**Tests**: seven cases — reject on `1.1.0rc2` / `1.2.0` / `1.3.0rc10`, accept on `1.3.0rc11` / `1.3.0rc23` / `1.3.0` / `1.4.0`, plus one asserting the guard fires from `_parse_logprobs` and not only from `__main__`. A missing `tensorrt_llm` is treated as "nothing to check", which keeps the suite install-free; that branch is unreachable in a real run because the backend refuses to build a model without it.

tests/examples/llm_eval/test_lm_eval_trtllm.py
cjluo-nv: Reproduced and fixed in `6a54747c3`. I removed `HF_DATASETS_TRUST_REMOTE_CODE` from `datasets.config` to simulate 4.x and got exactly your result — **7 passed, 2 failed** with `AttributeError` from `monkeypatch.setattr`, before either test exercised anything.

My venv has `datasets` 3.3.2, which still defines the attribute. That is the whole reason it was green locally and red for you; thanks for pinning the version, it made this a two-minute repro instead of a guess.

Fixed with `raising=False` on both calls, which is the right form here: lm-eval creates the attribute on assignment, and monkeypatch still removes it again on teardown, so the isolation the tests wanted is preserved either way.

Now verified under both: 17 passed on `datasets` 3.3.2, and 17 passed with the attribute deleted. Also re-ran through the shared `tests/examples/conftest.py` rather than `--noconftest`, so the real collection path is covered.

And your closing point is well taken — the `trtllm-non-pr` suite being skipped is exactly why this reached you instead of CI. That is the nightly-only gap discussed above; both of these findings would have been caught by it.

examples/hf_ptq/README.md
```diff
@@ -31,6 +31,11 @@ This section focuses on Post-training quantization, a technique that reduces mod
 For Hugging Face models, please use the TensorRT-LLM docker image (e.g., `nvcr.io/nvidia/tensorrt-llm/release:1.2.0`).
 Visit our [installation docs](https://nvidia.github.io/Model-Optimizer/getting_started/2_installation.html) for more information.
 
+> **NOTE:** `--tasks lm_eval` needs **TensorRT-LLM >= 1.3.0rc11** for loglikelihood
```
cjluo-nv: Maybe we don't need this change

meenchen (approved):

PR #2078, opened 2026-08-05
fix(autotune): pre-check remote board connectivity before benchmark

willg-nv:
# fix(autotune): pre-check remote board connectivity before benchmark

## Summary

When using `--remoteAutoTuningConfig`, a lost connection to the remote board
causes trtexec to fail. The failure gets recorded as `error: true` in
`autotune_states.yaml`, permanently skipping those schemes on restart. This is
incorrect — the error is transient (network issue), not a property of the
scheme itself.

This PR adds a TCP connectivity pre-check before each trtexec invocation. If
the board is unreachable after configurable retries, the autotuner saves state
and exits cleanly instead of poisoning the state file.

## Changes

- **`common.py`**: Add `RemoteConnectionError` exception (subclass of `AutotunerError`)
- **`benchmark.py`**: Add `_check_remote_connectivity()` and `_try_connect()` helpers;
  call pre-check in `TrtExecBenchmark.run()` before launching trtexec
- **`workflows.py`**: Catch `RemoteConnectionError` in per-scheme loop, save state
  before re-raising
- **`__main__.py`**: Add `--remote_connection_retries` CLI argument (default: 3)
- **`CHANGELOG.rst`**: Add bug fix entry

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| No remote autotune | unchanged | unchanged |
| Board unreachable before trtexec | trtexec fails → scheme marked `error: true` | pre-check fails → state saved → clean exit |
| Board reachable, trtexec succeeds | latency recorded | latency recorded |
| Board drops mid-trtexec | trtexec fails → scheme marked `error: true` | same (pre-check passed, trtexec failure recorded) |

## New CLI option

```
--remote_connection_retries N
    Number of TCP connection attempts to the remote board before aborting.
    Only relevant when --remoteAutoTuningConfig is present in trtexec args.
    Default: 3
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote ONNX autotuning reliability by checking board connectivity before benchmarks run.
  * Added configurable connection retries, with three retries by default.
  * Autotuning now saves its progress and exits cleanly when the remote board is unavailable, preventing schemes from being incorrectly marked as failed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Changelog
 **Bug Fixes**
 
 - Fix EAGLE-3 training with context parallelism (``--cp_size > 1`` in ``examples/speculative_decoding``), which failed to start on ``accelerate >= 1.13`` and then raised ``got mixed torch.Tensor and DTensor``.
+- Fix ONNX Autotune remote autotuning to pre-check board connectivity (configurable retries via ``--remote_connection_retries``) before each trtexec invocation. If unreachable, the autotuner saves state and exits cleanly instead of running trtexec and permanently marking schemes as errored.
 
 0.46 (2026-08-17)
 ^^^^^^^^^^^^^^^^^
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Changelog
 **Bug Fixes**
 
 - Fix ``megatron_generate`` dropping the VLM vision inputs (``pixel_values`` / ``image_grid_thw`` / ``image_sizes``) after the first generated token when KV-cache decoding is off, including the automatic fallback under sequence parallelism, which made generation silently ignore the image. No other ModelOpt feature is affected.
+- Fix EAGLE-3 training with context parallelism (``--cp_size > 1`` in ``examples/speculative_decoding``), which failed to start on ``accelerate >= 1.13`` and then raised ``got mixed torch.Tensor and DTensor``.
+- Fix ONNX Autotune remote autotuning to pre-check board connectivity (configurable retries via ``--remote_connection_retries``) before each trtexec invocation. If unreachable, the autotuner saves state and exits cleanly instead of running trtexec and permanently marking schemes as errored.
 
 0.47.0 (2026-09-xx)
 ^^^^^^^^^^^^^^^^^^^
--- a/docs/source/guides/9_autotune.rst
+++ b/docs/source/guides/9_autotune.rst
@@ -255,6 +255,23 @@ To use remote autotuning during Q/DQ placement optimization, run with ``trtexec`
 
 Replace ``<remote autotuning config>`` with an actual remote autotuning configuration string (see ``trtexec --help`` for more details). Other TensorRT benchmark options (e.g. ``--timing_cache``, ``--warmup_runs``, ``--timing_runs``, ``--plugin_libraries``) are also available; run ``--help`` for details.
 
+**Connectivity pre-check:**
+
+When ``--remoteAutoTuningConfig`` is detected, the autotuner tests TCP connectivity to the remote board before each trtexec invocation. If the board is unreachable after retries, the autotuner saves state and exits cleanly — preventing transient network failures from permanently marking schemes as errored.
+
+Configure the retry count with ``--remote_connection_retries`` (default: 3):
+
+.. code-block:: bash
+
+   python -m modelopt.onnx.quantization.autotune \
+       --onnx_path model.onnx \
+       --output_dir ./model_remote_autotuned \
+       --use_trtexec \
+       --trtexec_benchmark_args "--remoteAutoTuningConfig=\"ssh://admin@192.168.1.100\" --safe --skipInference" \
+       --remote_connection_retries 5
+
+Each failed attempt is logged as a warning. If all retries fail, the process exits with an error message and preserved state. On restart (same ``--output_dir``), autotuning resumes from where it left off without re-testing already-profiled schemes.
+
 Low-Level API Usage
 ===================
 
--- a/docs/source/guides/9_autotune.rst
+++ b/docs/source/guides/9_autotune.rst
@@ -257,6 +257,23 @@ To use remote autotuning during Q/DQ placement optimization, run with ``trtexec`
 
 Replace ``<remote autotuning config>`` with an actual remote autotuning configuration string (see ``trtexec --help`` for more details). Other TensorRT benchmark options (e.g. ``--timing_cache``, ``--warmup_runs``, ``--timing_runs``, ``--plugin_libraries``) are also available; run ``--help`` for details.
 
+**Connectivity pre-check:**
+
+When ``--remoteAutoTuningConfig`` is detected, the autotuner tests TCP connectivity to the remote board before each trtexec invocation. If the board is unreachable after retries, the autotuner saves state and exits cleanly — preventing transient network failures from permanently marking schemes as errored.
+
+Configure the retry count with ``--remote_connection_retries`` (default: 3):
+
+.. code-block:: bash
+
+   python -m modelopt.onnx.quantization.autotune \
+       --onnx_path model.onnx \
+       --output_dir ./model_remote_autotuned \
+       --use_trtexec \
+       --trtexec_benchmark_args "--remoteAutoTuningConfig=\"ssh://admin@192.168.1.100\" --safe --skipInference" \
+       --remote_connection_retries 5
+
+Each failed attempt is logged as a warning. If all retries fail, the process exits with an error message and preserved state. On restart (same ``--output_dir``), autotuning resumes from where it left off without re-testing already-profiled schemes.
+
 Low-Level API Usage
 ===================
 
--- a/modelopt/onnx/quantization/autotune/benchmark.py
+++ b/modelopt/onnx/quantization/autotune/benchmark.py
@@ -82,12 +82,19 @@ def _check_remote_connectivity(trtexec_args: list[str], retries: int = 3) -> Non
     if config_value is None:
         return
 
+    config_value = config_value.strip("'\"")
+
     parsed = urllib.parse.urlparse(config_value)
     hostname = parsed.hostname
     if not hostname:
-        return
+        raise RemoteConnectionError(f"Missing hostname in remote autotuning URI: {config_value!r}")
 
-    port = parsed.port
+    try:
+        port = parsed.port
+    except ValueError as e:
+        raise RemoteConnectionError(
+            f"Invalid port in remote autotuning URI: {config_value!r} - {e}"
+        ) from e
     if port is None:
         port = _DEFAULT_PORTS.get(parsed.scheme, 22)
 
@@ -361,15 +368,17 @@ class TrtExecBenchmark(Benchmark):
             if result.returncode != 0:
                 self.logger.error(f"trtexec failed with return code {result.returncode}")
                 self.logger.error(f"stderr: {result.stderr}")
-                return float("inf")
+                return self._benchmark_failed()
 
             if not (match := re.search(self.latency_pattern, result.stdout, re.IGNORECASE)):
                 self.logger.warning("Could not parse median latency from trtexec output")
                 self.logger.debug(f"trtexec stdout:\n{result.stdout}")
-                return float("inf")
+                return self._benchmark_failed()
             latency = float(match.group(1))
             self.logger.info(f"TrtExec benchmark (median): {latency:.2f} ms")
             return latency
+        except RemoteConnectionError:
+            raise
         except FileNotFoundError:
             self.logger.error(
                 "'trtexec' binary not found. Please ensure TensorRT is installed and 'trtexec' is in PATH."
@@ -377,7 +386,24 @@ class TrtExecBenchmark(Benchmark):
             return float("inf")
         except Exception as e:
             self.logger.error(f"Benchmark failed: {e}")
-            return float("inf")
+            return self._benchmark_failed()
+
+    def _benchmark_failed(self) -> float:
+        """Classify a trtexec failure as transient (board lost) or genuine.
+
+        The pre-check in run() passed, but the board can drop while trtexec is
+        running. Re-test connectivity: if the board is gone the failure is a
+        network event, not a property of the scheme, and must not be recorded
+        as an error in the state file.
+
+        Returns:
+            float('inf') if the board is still reachable (genuine scheme failure).
+
+        Raises:
+            RemoteConnectionError: If the remote board is no longer reachable.
+        """
+        _check_remote_connectivity(self._base_cmd, retries=self._remote_connection_retries)
+        return float("inf")
 
 
 class TensorRTPyBenchmark(Benchmark):
--- a/modelopt/onnx/quantization/autotune/benchmark.py
+++ b/modelopt/onnx/quantization/autotune/benchmark.py
@@ -87,13 +87,15 @@ def _check_remote_connectivity(trtexec_args: list[str], retries: int = 3) -> Non
     parsed = urllib.parse.urlparse(config_value)
     hostname = parsed.hostname
     if not hostname:
-        raise RemoteConnectionError(f"Missing hostname in remote autotuning URI: {config_value!r}")
+        raise RemoteConnectionError(
+            f"Missing hostname in remote autotuning URI (scheme={parsed.scheme!r})"
+        )
 
     try:
         port = parsed.port
     except ValueError as e:
         raise RemoteConnectionError(
-            f"Invalid port in remote autotuning URI: {config_value!r} - {e}"
+            f"Invalid port in remote autotuning URI: {parsed.scheme}://{hostname} - {e}"
         ) from e
     if port is None:
         port = _DEFAULT_PORTS.get(parsed.scheme, 22)
--- a/modelopt/onnx/quantization/autotune/benchmark.py
+++ b/modelopt/onnx/quantization/autotune/benchmark.py
@@ -31,8 +31,10 @@ import importlib.util
 import os
 import re
 import shutil
+import socket
 import tempfile
 import time
+import urllib.parse
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any
@@ -41,6 +43,7 @@ import numpy as np
 import torch
 
 from modelopt.onnx.logging_config import logger
+from modelopt.onnx.quantization.autotune.common import RemoteConnectionError
 from modelopt.onnx.quantization.ort_utils import _check_for_trtexec, _run_trtexec
 
 TRT_AVAILABLE = importlib.util.find_spec("tensorrt") is not None
@@ -48,6 +51,71 @@ if TRT_AVAILABLE:
     import tensorrt as trt
 
 TORCH_CUDA_AVAILABLE = torch.cuda.is_available()
+
+
+_DEFAULT_PORTS = {"ssh": 22, "http": 80, "https": 443}
+
+
+def _check_remote_connectivity(trtexec_args: list[str], retries: int = 3) -> None:
+    """Test TCP connectivity to the remote autotuning board before running trtexec.
+
+    Scans trtexec_args for --remoteAutoTuningConfig, parses the URI to extract
+    hostname and port, then attempts a TCP connection with a 5-second timeout.
+    Retries up to `retries` times before raising an error.
+
+    Args:
+        trtexec_args: List of trtexec command-line arguments.
+        retries: Number of connection attempts before giving up (default: 3).
+
+    Raises:
+        RemoteConnectionError: If the remote board is unreachable after all retries.
+    """
+    config_value = None
+    for i, arg in enumerate(trtexec_args):
+        if arg.startswith("--remoteAutoTuningConfig="):
+            config_value = arg.split("=", 1)[1]
+            break
+        elif arg == "--remoteAutoTuningConfig" and i + 1 < len(trtexec_args):
+            config_value = trtexec_args[i + 1]
+            break
+
+    if config_value is None:
+        return
+
+    parsed = urllib.parse.urlparse(config_value)
+    hostname = parsed.hostname
+    if not hostname:
+        return
+
+    port = parsed.port
+    if port is None:
+        port = _DEFAULT_PORTS.get(parsed.scheme, 22)
+
+    last_error = _try_connect(hostname, port, retries)
+    if last_error is not None:
+        raise RemoteConnectionError(
+            f"Cannot reach remote autotuning board at {hostname}:{port} after {retries} attempts - "
+            f"{last_error}. Exiting to avoid marking schemes as errors in state file."
+        ) from last_error
+
+
+def _try_connect(hostname: str, port: int, retries: int) -> Exception | None:
+    """Attempt TCP connection with retries. Returns None on success, last error on failure."""
+    last_error = None
+    for attempt in range(1, retries + 1):
+        try:
+            conn = socket.create_connection((hostname, port), timeout=5)
+            conn.close()
+            return None
+        except (TimeoutError, OSError) as e:  # noqa: PERF203
+            last_error = e
+            if attempt < retries:
+                logger.warning(
+                    f"Remote board connection attempt {attempt}/{retries} failed "
+                    f"({hostname}:{port}): {e}. Retrying..."
+                )
+                time.sleep(2)
+    return last_error
 
 
 def _validate_shape_range(min_shape: list, opt_shape: list, max_shape: list) -> None:
@@ -159,6 +227,7 @@ class TrtExecBenchmark(Benchmark):
         timing_runs: int = 10,
         plugin_libraries: list[str] | None = None,
         trtexec_args: list[str] | None = None,
+        remote_connection_retries: int = 3,
     ):
         """Initialize the trtexec benchmark.
 
@@ -170,8 +239,12 @@ class TrtExecBenchmark(Benchmark):
             trtexec_args: Additional command-line arguments to pass to trtexec.
                          These are appended after the standard arguments.
                          Example: ['--fp16', '--workspace=4096', '--verbose']
+            remote_connection_retries: Number of TCP connection attempts to the remote
+                         board before giving up (default: 3). Only used when
+                         --remoteAutoTuningConfig is present in trtexec_args.
         """
         super().__init__(timing_cache_file, warmup_runs, timing_runs, plugin_libraries)
+        self._remote_connection_retries = remote_connection_retries
         self.trtexec_args = trtexec_args if trtexec_args is not None else []
         self.temp_dir = tempfile.mkdtemp(prefix="trtexec_benchmark_")
         self.engine_path = os.path.join(self.temp_dir, "engine.trt")
@@ -252,6 +325,8 @@ class TrtExecBenchmark(Benchmark):
         """
         if not os.path.exists(self.timing_cache_file):
             self.logger.debug(f"Will create timing cache: {self.timing_cache_file}")
+
+        _check_remote_connectivity(self._base_cmd, retries=self._remote_connection_retries)
 
         try:
             model_path = path_or_bytes
--- a/modelopt/onnx/quantization/autotune/workflows.py
+++ b/modelopt/onnx/quantization/autotune/workflows.py
@@ -295,7 +295,12 @@ def region_pattern_autotuning_workflow(
         baseline_path = output_dir / "baseline.onnx"
         autotuner.export_onnx(str(baseline_path), insert_qdq=False)
         baseline_log = logs_dir / "baseline.log"
-        baseline_latency = benchmark_onnx_model(str(baseline_path), str(baseline_log))
+        try:
+            baseline_latency = benchmark_onnx_model(str(baseline_path), str(baseline_log))
+        except RemoteConnectionError:
+            logger.error("Remote board connection lost during baseline, saving state before exit")
+            autotuner.save_state(str(state_path))
+            raise
         autotuner.submit(baseline_latency)
         logger.info(f"Baseline: {baseline_latency:.2f} ms")
     else:
@@ -377,7 +382,14 @@ def region_pattern_autotuning_workflow(
     final_model_path = output_dir / "optimized_final.onnx"
     autotuner.export_onnx(str(final_model_path), insert_qdq=True)
     final_log = logs_dir / "final.log"
-    final_latency = benchmark_onnx_model(str(final_model_path), str(final_log))
+    try:
+        final_latency = benchmark_onnx_model(str(final_model_path), str(final_log))
+    except RemoteConnectionError:
+        logger.error(
+            "Remote board connection lost during final measurement, saving state before exit"
+        )
+        autotuner.save_state(str(state_path))
+        raise
 
     if final_latency > 0 and final_latency != float("inf"):
         speedup = baseline_latency / final_latency
--- a/modelopt/onnx/quantization/autotune/workflows.py
+++ b/modelopt/onnx/quantization/autotune/workflows.py
@@ -29,7 +29,7 @@ import onnx
 from modelopt.onnx.logging_config import logger
 from modelopt.onnx.quantization.autotune.autotuner import QDQAutotuner
 from modelopt.onnx.quantization.autotune.benchmark import TensorRTPyBenchmark, TrtExecBenchmark
-from modelopt.onnx.quantization.autotune.common import Config, PatternCache
+from modelopt.onnx.quantization.autotune.common import Config, PatternCache, RemoteConnectionError
 from modelopt.onnx.quantization.qdq_utils import get_quantized_tensors
 
 _benchmark_instance = None
@@ -75,6 +75,8 @@ def benchmark_onnx_model(
         logger.debug(f"Benchmark result: {latency:.2f} ms")
         return latency
 
+    except RemoteConnectionError:
+        raise
     except Exception as e:
         logger.error(f"Benchmark error: {e}", exc_info=True)
         return float("inf")
@@ -87,6 +89,7 @@ def init_benchmark_instance(
     warmup_runs: int = 5,
     timing_runs: int = 20,
     trtexec_args: list[str] | None = None,
+    remote_connection_retries: int = 3,
 ):
     """Initialize global TensorRT benchmark instance for model performance measurement.
 
@@ -103,6 +106,9 @@ def init_benchmark_instance(
                     Higher values give more stable median (default: 20)
         trtexec_args: Additional command-line arguments to pass to trtexec as a string (only used if use_trtexec=True).
                      Example: '--fp16 --workspace=4096 --verbose'
+        remote_connection_retries: Number of TCP connection attempts to the remote board
+                     before aborting (default: 3). Only relevant when --remoteAutoTuningConfig
+                     is present in trtexec_args.
     """
     global _benchmark_instance
     try:
@@ -113,6 +119,7 @@ def init_benchmark_instance(
                 timing_runs=timing_runs,
                 plugin_libraries=plugin_libraries,
                 trtexec_args=trtexec_args,
+                remote_connection_retries=remote_connection_retries,
             )
             logger.info("Trtexec benchmark initialized")
         else:
@@ -330,9 +337,14 @@ def region_pattern_autotuning_workflow(
             model_bytes = autotuner.export_onnx(None, insert_qdq=True)
             test_log = logs_dir / f"region_{region.id}_scheme_{scheme_idx}.log"
             flush_timing_cache = (iteration_count % 10) == 0
-            latency = benchmark_onnx_model(
-                model_bytes, str(test_log), flush_timing_cache=flush_timing_cache
-            )
+            try:
+                latency = benchmark_onnx_model(
+                    model_bytes, str(test_log), flush_timing_cache=flush_timing_cache
+                )
+            except RemoteConnectionError:
+                logger.error("Remote board connection lost, saving state before exit")
+                autotuner.save_state(str(state_path))
+                raise
 
             autotuner.submit(latency, success=(latency != float("inf")))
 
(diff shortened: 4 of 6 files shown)


modelopt/onnx/quantization/autotune/benchmark.py
arata-miyamoto, line 351: A suggestion for handling "mid-trtexec drops". Without this fix, autotune reports a scheme that failed due to board loss during `trtexec` execution as "inf" latency, and it never retries the scheme on resume.

Note that this code cannot detect brief network loss that starts during trtexec execution and recovers immediately before the connectivity check in `_benchmark_failed()`. However, it still can handle persistent network losses during autotune.

```Python
    ....
            if result.returncode != 0:
                  self.logger.error(f"trtexec failed with return code {result.returncode}")
                  self.logger.error(f"stderr: {result.stderr}")
                  return self._benchmark_failed()

              if not (match := re.search(self.latency_pattern, result.stdout, re.IGNORECASE)):
                  self.logger.warning("Could not parse median latency from trtexec output")
                  self.logger.debug(f"trtexec stdout:\n{result.stdout}")
                  return self._benchmark_failed()
              ...
          except RemoteConnectionError:      # Caught by the `except Exception` below without this.
              raise
          except FileNotFoundError:
              ...
              return float("inf")
          except Exception as e:
              self.logger.error(f"Benchmark failed: {e}")
              return self._benchmark_failed()

      def _benchmark_failed(self) -> float:
          """Classify a trtexec failure as transient (board lost) or genuine.

          The pre-check in run() passed, but the board can drop while trtexec is
          running. Re-test connectivity: if the board is gone the failure is a
          network event, not a property of the scheme, and must not be recorded
          as an error in the state file.

          Returns:
              float('inf') if the board is still reachable (genuine scheme failure).

          Raises:
              RemoteConnectionError: If the remote board is no longer reachable.
          """
          _check_remote_connectivity(self._base_cmd, retries=self._remote_connection_retries)
          return float("inf")
    ....
```

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -50,6 +53,71 @@
...
+        return
+
+    parsed = urllib.parse.urlparse(config_value)
+    hostname = parsed.hostname
+    if not hostname:
+        return
+
+    port = parsed.port
+    if port is None:
+        port = _DEFAULT_PORTS.get(parsed.scheme, 22)
```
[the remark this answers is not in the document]
arata-miyamoto: This is correct. When autotuning runs with a command as described in [the document](https://github.com/willg-nv/Model-Optimizer/blob/3eee98128f92b2e61a53301d0bd5c4a209ffc2b8/docs/source/guides/9_autotune.rst#remote-autotuning), the URL in `config_value` includes quotation marks.

```
   ...
   --trtexec_benchmark_args "--remoteAutoTuningConfig=\"ssh://admin@192.168.1.100\" --safe --skipInference"
```

As coderabbit suggests , matching quotes in `config_value` needs to be removed before `urlparse()`.

modelopt/onnx/quantization/autotune/benchmark.py
willg-nv: thanks for catching this , applied the fixe.

modelopt/onnx/quantization/autotune/workflows.py
```diff
@@ -288,7 +295,13 @@ def region_pattern_autotuning_workflow(
         baseline_path = output_dir / "baseline.onnx"
         autotuner.export_onnx(str(baseline_path), insert_qdq=False)
         baseline_log = logs_dir / "baseline.log"
-        baseline_latency = benchmark_onnx_model(str(baseline_path), str(baseline_log))
+        try:
+            baseline_latency = benchmark_onnx_model(str(baseline_path), str(baseline_log))
+        except RemoteConnectionError:
+            logger.error("Remote board connection lost during baseline, saving state before exit")
+            autotuner.save_state(str(state_path))
+            logger.info(f"State saved to {state_path}, use --resume to continue")
```
arata-miyamoto: The parser doesn't have `--resume` option. The correct message should be something like "re-run the same command to continue".

arata-miyamoto (changes requested):

modelopt/onnx/quantization/autotune/workflows.py
willg-nv: fixed, this was imagined by AI. 😅

arata-miyamoto (approved):

cjluo-nv (commented):

Comment: the pre-check logic is new, easily unit-testable pure code but ships with no tests, and an unparseable `--remoteAutoTuningConfig` now hard-aborts a run that previously worked.

Needs action:
- [ ] Add unit tests in `tests/unit/onnx/quantization/autotune/` for `_check_remote_connectivity` (both `--flag=value` and `--flag value` forms, default ports, missing/invalid port, no-remote-config no-op) and `_try_connect` retry/backoff with a mocked `socket.create_connection`.
- [ ] Don't raise `RemoteConnectionError` when the config string has no parseable hostname (e.g. `192.168.1.100:22`, non-URI trtexec syntax) — log a warning and skip the pre-check instead of aborting the whole autotune. See inline comment in `benchmark.py`.
- [ ] Document the `_benchmark_failed()` post-failure re-check: the PR table and `docs/source/guides/9_autotune.rst` both say a mid-trtexec board drop behaves as before, but the code now aborts. Fix one or the other.
- [ ] Factor the three identical `except RemoteConnectionError: save_state + log + raise` blocks in `workflows.py` into one helper.
- [ ] Remove the unrelated EAGLE-3 context-parallelism entry from `CHANGELOG.rst` (looks like a stray rebase artifact).

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -50,6 +53,80 @@
...
+
+    if config_value is None:
+        return
+
+    config_value = config_value.strip("'\"")
+
+    parsed = urllib.parse.urlparse(config_value)
+    hostname = parsed.hostname
+    if not hostname:
+        raise RemoteConnectionError(
```
cjluo-nv, line 90: > _Bot comment._

Raising here turns an unrecognized config string into a hard abort of the entire autotune run. `urllib.parse.urlparse` only yields a hostname for scheme-prefixed URIs — `--remoteAutoTuningConfig=192.168.1.100:22` or any non-URI trtexec syntax parses to `hostname=None` and kills a run that works today. Safer: `logger.warning(...)` and `return` (skip the pre-check) when the value can't be parsed; the same applies to the invalid-port branch just below. Also, falling back to port 22 for an unknown scheme can produce false "unreachable" aborts — consider skipping instead of guessing.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -286,23 +370,42 @@ def run(
...
+
+        The pre-check in run() passed, but the board can drop while trtexec is
+        running. Re-test connectivity: if the board is gone the failure is a
+        network event, not a property of the scheme, and must not be recorded
+        as an error in the state file.
+
+        Returns:
+            float('inf') if the board is still reachable (genuine scheme failure).
+
+        Raises:
```
cjluo-nv, line 404: > _Bot comment._

`_benchmark_failed()` changes behavior the PR description explicitly says is unchanged ("Board drops mid-trtexec → same"), and the docs only mention a pre-check "before each trtexec invocation". Please align the PR body and `9_autotune.rst` with what the code does. Note this also adds up to `retries * (5s timeout + 2s sleep)` to every genuine trtexec failure when a remote config is present — worth mentioning in the docs.

modelopt/onnx/quantization/autotune/workflows.py
```diff
@@ -303,7 +310,15 @@ def region_pattern_autotuning_workflow(
         baseline_path = output_dir / "baseline.onnx"
         autotuner.export_onnx(str(baseline_path), insert_qdq=False, model_transform=model_transform)
         baseline_log = logs_dir / "baseline.log"
-        baseline_latency = benchmark_onnx_model(str(baseline_path), str(baseline_log))
+        try:
```
cjluo-nv, line 313: > _Bot comment._

This `except RemoteConnectionError: log → save_state → log → raise` block is repeated verbatim three times (baseline, per-scheme, final). Wrap it once, e.g. `_benchmark_or_checkpoint(autotuner, state_path, ...)`, so the message/state-saving logic lives in one place.

CHANGELOG.rst
```diff
@@ -13,6 +13,8 @@ Changelog
 **Bug Fixes**
 
 - Fix ``megatron_generate`` dropping the VLM vision inputs (``pixel_values`` / ``image_grid_thw`` / ``image_sizes``) after the first generated token when KV-cache decoding is off, including the automatic fallback under sequence parallelism, which made generation silently ignore the image. No other ModelOpt feature is affected.
+- Fix EAGLE-3 training with context parallelism (``--cp_size > 1`` in ``examples/speculative_decoding``), which failed to start on ``accelerate >= 1.13`` and then raised ``got mixed torch.Tensor and DTensor``.
```
cjluo-nv, line 16: > _Bot comment._

The EAGLE-3 context-parallelism entry is unrelated to this PR — likely picked up from another branch during a rebase. Please drop it and keep only the autotune line.

modelopt/onnx/quantization/autotune/benchmark.py
willg-nv: fixed, thanks!

modelopt/onnx/quantization/autotune/workflows.py
willg-nv: fixed, thanks!

CHANGELOG.rst
willg-nv: removed.

modelopt/onnx/quantization/autotune/benchmark.py
willg-nv: updated

modelopt/onnx/quantization/autotune/workflows.py
```diff
@@ -134,6 +141,35 @@ def init_benchmark_instance(
...
+    save-state-then-raise behaviour when a ``RemoteConnectionError`` propagates.
+
+    Returns:
+        Measured latency in ms, or ``float('inf')`` on non-connection failure.
+
+    Raises:
+        RemoteConnectionError: After saving autotuner state to *state_path*.
+    """
+    try:
+        return benchmark_onnx_model(model_path, log_file, flush_timing_cache=flush_timing_cache)
```
cjluo-nv, line 163: > _Bot comment._

The previously requested tests are still absent. Please add mocked tests for flag parsing, defaults/fallbacks, connection success/retry exhaustion and backoff, plus workflow tests covering baseline, per-scheme and final failures. Verify RemoteConnectionError saves state and propagates without submitting an errored scheme, and reload the checkpoint to prove the interrupted scheme remains eligible. Also exercise the post-failure re-check and preserve float('inf') for genuine benchmark failures. These are the critical paths this fix promises to protect; they do not require a real board.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -50,6 +53,92 @@
...
+        logger.warning(
+            "Could not parse hostname from --remoteAutoTuningConfig=%r; "
+            "skipping connectivity pre-check.",
+            config_value,
+        )
+        return
+
+    try:
+        port = parsed.port
+    except ValueError:
```
cjluo-nv, line 113: > _Bot comment._

The fallback change reintroduces the previously flagged credential leak, now at warning level. For example, `ssh://admin:secret@board:bad` reaches this branch and logs the entire URI, including the password. The missing-host warning does the same. Omit the raw config value from both messages (or reliably redact userinfo), and add a log-capture test asserting credentials never appear.

cjluo-nv (commented):

Comment: checkpoint handling is improved, but critical regression tests and input-validation fixes remain outstanding, and URI warnings reintroduce credential exposure.

Needs action:
- [ ] Add mocked connectivity and checkpoint/resume tests under `tests/unit/onnx/quantization/autotune/`, including persistent mid-trtexec disconnects and ordinary benchmark failures.
- [ ] Redact credentials from the new warnings in `benchmark.py`; see inline comment.
- [ ] Validate API retry counts and handle missing values and URI-parser exceptions in `benchmark.py`; see inline comments.
- [ ] :speech_balloon: Author replied “updated”: the guide now describes post-failure checks, but update the PR body’s mid-trtexec row, which still says behavior is unchanged.

No action needed:
- :heavy_check_mark: Resolved since the last review: exception propagation/checkpoint coverage, duplicated save-state handlers, and unrelated changelog changes.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -170,8 +260,12 @@ def __init__(
             trtexec_args: Additional command-line arguments to pass to trtexec.
                          These are appended after the standard arguments.
                          Example: ['--fp16', '--workspace=4096', '--verbose']
+            remote_connection_retries: Number of TCP connection attempts to the remote
+                         board before giving up (default: 3). Only used when
+                         --remoteAutoTuningConfig is present in trtexec_args.
         """
         super().__init__(timing_cache_file, warmup_runs, timing_runs, plugin_libraries)
+        self._remote_connection_retries = remote_connection_retries
```
cjluo-nv, line 294: > _Bot comment._

The direct-API retry validation concern is still unresolved: only run_autotune validates the range. TrtExecBenchmark(remote_connection_retries=0) or a negative value makes _try_connect return None without connecting, silently bypassing both checks. Require an integer in the documented 1–10 range before storing it, document that range here, and test invalid direct API inputs.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -50,6 +53,92 @@
...
+        retries: Number of connection attempts before giving up (default: 3).
+
+    Raises:
+        RemoteConnectionError: If the remote board is unreachable after all retries.
+    """
+    config_value = None
+    for i, arg in enumerate(trtexec_args):
+        if arg.startswith("--remoteAutoTuningConfig="):
+            config_value = arg.split("=", 1)[1]
+            break
```
cjluo-nv, line 77: > _Bot comment._

The prior missing-value finding remains open. A trailing `--remoteAutoTuningConfig` falls through to the no-config return, allowing trtexec to fail and the scheme to be recorded as errored. Distinguish an absent flag from a present flag without a value and raise RemoteConnectionError for the latter (also handle an empty inline value). This does not require rejecting nonempty, unrecognized trtexec configuration syntax.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -50,6 +53,92 @@
...
+        elif arg == "--remoteAutoTuningConfig" and i + 1 < len(trtexec_args):
+            config_value = trtexec_args[i + 1]
+            break
+
+    if config_value is None:
+        return
+
+    config_value = config_value.strip("'\"")
+
```
cjluo-nv, line 89: > _Bot comment._

The new warning-and-skip fallback does not cover exceptions from urlparse itself. For example, `ssh://[board` raises ValueError before the hostname/port checks. benchmark_onnx_model then converts this pre-check exception into float('inf'), potentially recording a scheme failure. Catch parser/hostname ValueError as well, applying the same credential-safe warning-and-skip policy used for unrecognized configurations, and test this case.

modelopt/onnx/quantization/autotune/workflows.py
```diff
@@ -134,6 +141,35 @@ def init_benchmark_instance(
...
+    except RemoteConnectionError:
+        logger.error("Remote board connection lost, saving state before exit")
+        autotuner.save_state(str(state_path))
+        logger.info(
```
cjluo-nv, line 167: > _Bot comment._

The new mid-region recovery path calls `save_state()`, but that serializer writes only `profiled_patterns`, not `current_profile_pattern_schemes` (only its signature is saved). Newly generated, interrupted schemes are not added to the pattern cache either. On reload the active pattern is seeded from a filtered cache, and `_seed_from_cache()` resets its measurements. Consequently the interrupted candidate can disappear and completed active-region benchmarks run again, contrary to the new resume guarantee. Preserve/restore the active pattern's full scheme state without marking it complete. Test an interruption after at least one successful non-baseline scheme, then verify the interrupted hash is retained with `error=False` and completed measurements are not repeated.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,577 @@
...
+
+# ============================================================================
+# Workflow integration — interrupted scheme remains eligible after reload
+# ============================================================================
+
+
+def _create_simple_conv_onnx_model_local():
+    """Import-safe model builder (avoids torch at import time for onnx-only helper)."""
+    from _test_utils.onnx.quantization.autotune.models import _create_simple_conv_onnx_model
+
```
cjluo-nv: > _Bot comment._

The critical integration request is only partially covered. This test manually saves state without raising a connection error or calling the workflow, and its final assertion merely checks that an empty pattern collection can be created. No test invokes `region_pattern_autotuning_workflow`, exercises a final-measurement interruption, or makes `TrtExecBenchmark.run()` pass its pre-check and fail its post-failure check. Add those mocked paths, exercise `benchmark_onnx_model` exception propagation, assert no failed-scheme submission on connection loss, and actually generate/compare the resumed candidate. Also assert two connectivity calls in the existing run-failure tests: their `inf` assertion would still pass if the re-check were removed.

cjluo-nv (commented):

Comment: input-validation fixes are addressed, but the new recovery path does not preserve active-region progress, and critical integration coverage remains incomplete.

Needs action:
- [ ] Preserve the interrupted scheme and completed active-region measurements across checkpoint reload; see `workflows.py` inline comment.
- [ ] Complete `test_remote_connectivity.py` coverage for persistent mid-trtexec loss, actual workflow baseline/scheme/final failures, and resumed scheme identity.
- [ ] Remove collection-time global dependency stubs from `test_remote_connectivity.py`; they contaminate other tests.
- [ ] Move unjustified function-local imports in `test_remote_connectivity.py` to the module import section.
- [ ] :speech_balloon: Author replied “updated”: the guide describes post-failure checks, but update the PR body's mid-trtexec row, which still claims unchanged behavior.

No action needed:
- :heavy_check_mark: Resolved since the last review: API/CLI retry bounds, empty values and parser exceptions, credential-safe warnings, exception propagation, shared checkpoint handling, and unrelated changelog edits. Parsing/retry tests and baseline-save coverage are now present.
- The new test header matches `LICENSE_HEADER`; no existing tests were modified or deleted.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,577 @@
...
+    for mod_name in ("tensorrt", "tensorrt.tensorrt"):
+        if mod_name not in sys.modules:
+            stub = MagicMock()
+            # importlib.util.find_spec() introspects __spec__; give it a
+            # real ModuleSpec so the guard `find_spec("tensorrt") is not None`
+            # evaluates to True without actually loading the native library.
+            stub.__spec__ = importlib.machinery.ModuleSpec(mod_name, None)
+            sys.modules[mod_name] = stub
+            _ORIG_MODULES[mod_name] = None  # mark for cleanup
+        else:
```
cjluo-nv: > _Bot comment._

These stubs are installed during pytest collection and never restored (`_ORIG_MODULES` is never consumed). Checking `sys.modules` does not check whether TensorRT is installed: an installed but not-yet-imported package is replaced by a MagicMock, making subsequent imports and availability checks across the suite order-dependent. `benchmark.py` already handles TensorRT absence, so remove the unnecessary TensorRT stubs and follow neighboring autotuner tests' normal imports. If testing an unavailable-dependency environment is necessary, isolate it in a subprocess rather than modifying the shared interpreter at collection time.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,577 @@
...
+
+# ============================================================================
+# _benchmark_or_save — workflow-level state saving
+# ============================================================================
+
+
+class TestBenchmarkOrSave:
+    """_benchmark_or_save saves autotuner state on RemoteConnectionError."""
+
+    @staticmethod
```
cjluo-nv, line 452: > _Bot comment._

Move `QDQAutotuner` and `_benchmark_or_save` imports from these test methods to the top-level import section (also the repetitions below). No circular-import, optional-dependency, or startup-cost reason is given for keeping these imports local. The shared model builder is already used directly by `test_autotuner.py`; reuse that import rather than the extra local wrapper. The guarded torch import has a stated optional-dependency rationale, but that does not justify these unrelated local imports.

modelopt/onnx/quantization/autotune/workflows.py
willg-nv: The connectivity pre-check is a fast-fail guard that fires before trtexec launches — the typical interruption point is between regions or at most a few schemes into a new region. Re-profiling the interrupted region from scratch on resume is acceptable: the cost is bounded by num_schemes_per_region trtexec invocations (seconds each), while the complexity of serializing/restoring partial in-progress pattern state, distinguishing it from committed patterns on reload, and ensuring generate() skips already-measured hashes adds fragile bookkeeping for a marginal time saving. Completed regions are already checkpointed after each set_profile_region commit, so only the active region's partial progress is lost — not the entire session.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,648 @@
...
+        """Board still up after trtexec error -> genuine failure -> float('inf')."""
+        bench = self._make_bench()
+        with patch.object(bm, "_try_connect", return_value=None):
+            assert bench._benchmark_failed() == float("inf")
+
+    def test_board_unreachable_raises(self):
+        """Board dropped mid-trtexec -> RemoteConnectionError propagates."""
+        bench = self._make_bench()
+        with (
+            patch.object(bm, "_try_connect", return_value=OSError("gone")),
```
cjluo-nv, line 359: > _Bot comment._

The critical persistent mid-trtexec loss case is still absent. The direct `_benchmark_failed()` test proves the helper raises, while both `run()` tests keep the board reachable throughout. Add a `run()` test with `_try_connect` returning `None` for the pre-check and `OSError("gone")` for the post-failure check, with a mocked failing trtexec result. Assert two connectivity calls, one trtexec invocation, and propagation of `RemoteConnectionError` rather than `inf`. This specifically guards the exception-handling boundary that previously swallowed remote failures.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,648 @@
...
+            pytest.raises(RemoteConnectionError, match="board lost during final"),
+        ):
+            region_pattern_autotuning_workflow(
+                model,
+                output_dir=tmp_path / "output",
+                num_schemes_per_region=2,
+                state_file=str(tmp_path / "state.yaml"),
+            )
+
+        assert (tmp_path / "state.yaml").exists()
```
cjluo-nv: > _Bot comment._

The new workflow calls are useful, but checkpoint existence alone does not verify recovery. In this final-interruption test the region loop already creates the file, so removing the final save-on-error would still pass. Reload it and assert the final committed pattern and its measurements survived. Likewise, spy on `submit()` in the mid-scheme workflow test to prove no failed result is submitted, and add baseline interruption through the actual workflow. Exercise restart after a committed region plus successful measurements in the active region; assert committed regions are skipped and the active region remains eligible. The current fresh-candidate helper test neither verifies interrupted-scheme identity nor tests that more limited region-restart contract.

docs/source/guides/9_autotune.rst
```diff
@@ -257,6 +257,34 @@ To use remote autotuning during Q/DQ placement optimization, run with ``trtexec`
...
+
+.. code-block:: bash
+
+   python -m modelopt.onnx.quantization.autotune \
+       --onnx_path model.onnx \
+       --output_dir ./model_remote_autotuned \
+       --use_trtexec \
+       --trtexec_benchmark_args "--remoteAutoTuningConfig=\"ssh://admin@192.168.1.100\" --safe --skipInference" \
+       --remote_connection_retries 5
+
```
cjluo-nv, line 285: > _Bot comment._

The author's reply explicitly accepts re-profiling the active region instead of serializing its partial progress, but this sentence still promises restart without re-testing already-profiled schemes. `save_state()` stores committed patterns; active-region candidates/measurements are not fully restored and cache seeding resets latencies. Document the chosen region-level recovery boundary here: committed regions survive, the active region may be re-profiled, and the interrupted candidate is not guaranteed to be regenerated. Align the test/module docstrings claiming exact interrupted-scheme recovery too. The PR body's mid-trtexec row separately still says behavior is unchanged despite the post-failure re-check.

cjluo-nv (commented):

Comment: test isolation and propagation coverage improved, but critical disconnect/recovery tests remain incomplete and the documented resume guarantee is still inaccurate.

Needs action:
- [ ] Add a persistent mid-trtexec disconnect test through `TrtExecBenchmark.run()` in `test_remote_connectivity.py`; see inline comment.
- [ ] Strengthen workflow tests to verify submission history and reloaded checkpoint contents, including baseline and final interruptions; see inline comment.
- [ ] :speech_balloon: Author accepts restarting the active region — document that limitation in `9_autotune.rst` and test committed-region preservation rather than claiming exact scheme recovery.
- [ ] :speech_balloon: Author previously replied “updated” — update the PR body's mid-trtexec row; it still contradicts the implemented post-failure re-check.

No action needed:
- :heavy_check_mark: Resolved since the last review: global dependency stubs and local imports removed; workflow scheme/final calls, exception propagation, and two-check assertions now have coverage.
- Earlier validation, credential-warning, and checkpoint-helper fixes remain intact. The new header matches `LICENSE_HEADER`; no existing tests were modified or deleted.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
willg-nv: I think current Mock test is sufficent.

cjluo-nv (commented):

The problem is that transient board failures are being persisted as permanent scheme failures. I’ll first check existing benchmark and checkpoint mechanisms, then verify the previously requested disconnect and recovery tests against the current code.

cjluo-nv (commented):

The problem is that transient board loss can be recorded as a permanent scheme failure. I’ll check existing connectivity and checkpoint mechanisms first, then verify the remaining disconnect and restart tests.

cjluo-nv (commented):

Comment: recovery documentation and workflow assertions improved, but two critical disconnect/restart coverage gaps remain unresolved.

Needs action:
- [ ] :speech_balloon: Author considers existing mocks sufficient — add the persistent-disconnect test through `TrtExecBenchmark.run()` in `test_remote_connectivity.py`; see inline comment.
- [ ] Fix `test_restart_skips_committed_regions_resumes_active` to interrupt an active region after a committed region, then exercise restart; see inline comment.
- [ ] Explain in the PR body why benchmark-local TCP probing is preferred over classifying `_run_trtexec` failures or extending that shared runner with an optional guard.
- [ ] :speech_balloon: Author previously replied “updated” — correct the PR body's mid-trtexec row to describe persistent-loss detection and its brief-disconnect limitation.

No action needed:
- :heavy_check_mark: Resolved since the last review: region-level recovery documentation, baseline checkpoint reload coverage, and workflow submission-history assertions. Earlier validation, credential-warning, and test-isolation fixes remain intact.
- The 1,013-line change is test-heavy and cohesive; size alone is not blocking. No existing tests were modified or deleted, and the new header matches `LICENSE_HEADER`.
- Review was static; tests were not executed.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,764 @@
...
+    def test_board_reachable_returns_inf(self):
+        """Board still up after trtexec error -> genuine failure -> float('inf')."""
+        bench = self._make_bench()
+        with patch.object(bm, "_try_connect", return_value=None):
+            assert bench._benchmark_failed() == float("inf")
+
+    def test_board_unreachable_raises(self):
+        """Board dropped mid-trtexec -> RemoteConnectionError propagates."""
+        bench = self._make_bench()
+        with (
```
cjluo-nv, line 358: > _Bot comment._

The reply that the current mocks are sufficient does not resolve this critical coverage gap: both `run()` tests keep connectivity successful, while the unreachable-board test calls `_benchmark_failed()` directly. Add a `run()` test with `_try_connect` side effects `[None, OSError("gone")]` and a failing mocked trtexec result. Assert `RemoteConnectionError`, exactly two connectivity calls, and exactly one trtexec invocation. This verifies the actual pre-check → subprocess failure → post-check → exception-handler path, rather than its pieces independently.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,764 @@
...
+        ):
+            region_pattern_autotuning_workflow(
+                model,
+                output_dir=tmp_path / "output",
+                num_schemes_per_region=2,
+                state_file=str(tmp_path / "state.yaml"),
+            )
+
+        state_path = tmp_path / "state.yaml"
+        assert state_path.exists()
```
cjluo-nv, line 723: > _Bot comment._

This still does not exercise the recovery scenario in its name/docstring: the first run completes normally, so the resumed run has no active interrupted region. Use a fixture with distinct region patterns, commit one region, successfully measure a scheme in the next, then raise `RemoteConnectionError`. Reload/resume that checkpoint and assert the committed pattern and measurements are retained without re-benchmarking, while the active pattern remains eligible and is measured again. Also record byte-valued benchmark calls before the `call_count == 1` return in `_mock_benchmark_resume`; currently the first unexpected scheme measurement would escape `regions_entered`.

modelopt/onnx/quantization/autotune/benchmark.py
```diff
@@ -50,6 +53,106 @@
 TORCH_CUDA_AVAILABLE = torch.cuda.is_available()
 
 
+_DEFAULT_PORTS = {"ssh": 22, "http": 80, "https": 443}
+
```
cjluo-nv, line 57: > _Bot comment._

Design review: the problem is transient board failures becoming permanent scheme failures. Alternatives are (1) classify failures from the existing `ort_utils._run_trtexec` result, (2) add an optional connectivity guard to that shared runner, or (3) keep this benchmark-local stdlib `socket`/`urllib.parse` guard. The dependency manifest does not suggest a dedicated networking/retry dependency is necessary, and the implementation reuses the existing checkpoint serializer. Please add a brief PR-body rationale for choosing the third approach, including that TCP reachability cannot diagnose authentication failures or a disconnect that recovers before the post-check. No new framework is being requested.

willg-nv: trtexec stderr doesn't reliably distinguish board-lost from engine failure, and the shared runner shouldn't own connectivity concerns. Acknowledged TCP can't catch auth failures or mid-run disconnects that recover before the post-check.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,819 @@
...
+            side_effect=_mock_benchmark_resume,
+        ):
+            autotuner2 = region_pattern_autotuning_workflow(
+                model,
+                output_dir=tmp_path / "output2",
+                num_schemes_per_region=2,
+                state_file=str(state_path),
+            )
+
+        # Committed patterns from the first run are preserved intact.
```
cjluo-nv: > _Bot comment._

The revised test now interrupts and resumes a real workflow, resolving that part of the previous finding. However, these final assertions still pass if committed regions are re-benchmarked: `len(profiled_patterns) >= num_committed + 1` permits duplicate/replaced patterns, and `resume_scheme_calls` records only the string "scheme", not which pattern was measured. Snapshot committed pattern signatures and their scheme hashes/latencies before restart, record the active pattern or region for each benchmark (via a workflow/autotuner spy or the region-specific log path), and assert no resumed measurement belongs to a committed pattern and its saved measurements remain unchanged. Also identify the interrupted active pattern and assert that specific pattern is measured again. This completes the previously requested recovery-contract coverage rather than merely proving some work occurred after restart.

cjluo-nv (commented):

Not all comments are resolved: persistent-disconnect coverage is fixed, but the restart test still does not prove committed regions are skipped or preserved.

Needs action:
- [ ] Strengthen `test_restart_skips_committed_regions_resumes_active` with pattern-specific benchmark history and before/after measurement comparisons; see inline comment.
- [ ] :speech_balloon: Author previously replied “updated” — correct the PR body's mid-trtexec row to describe persistent-loss detection and the limitation when connectivity recovers before the post-check.
- [ ] :speech_balloon: Author explained unreliable stderr classification and keeping connectivity outside the shared runner — copy that rationale into the PR body, including TCP authentication limitations.

No action needed:
- :heavy_check_mark: Resolved since the last review: persistent disconnect through `run()`, actual interruption/restart setup, and recording every resumed scheme call. Earlier validation, credential-safe warnings, checkpoint propagation, test isolation, and region-level recovery documentation remain addressed.
- Design review: transient board loss must not become permanent scheme failure. Alternatives checked were `_run_trtexec` result classification, a shared-runner guard, and benchmark-local stdlib probing; the author's rationale supports the latter without another dependency.
- The 1,140-line change is cohesive and test-heavy. No existing tests are removed or weakened; the new header matches `LICENSE_HEADER`. Review was static; tests were not executed.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
willg-nv: Snapshotted committed pattern signatures/scheme-hashes/latencies before restart, spy on set_profile_region to record which pattern each resumed benchmark targets, and assert committed measurements are unchanged, no resumed call hits a committed pattern, and the interrupted pattern (from current_profile_pattern_schemes_signature in the state file) is specifically re-measured.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,889 @@
...
+        # 1. Committed patterns from the first run are preserved intact:
+        #    exact same count, same scheme hashes, same latencies.
+        profiled = autotuner2.profiled_patterns
+        assert len(profiled) >= num_committed + 1, (
+            f"Expected at least {num_committed + 1} profiled patterns after resume, "
+            f"got {len(profiled)}"
+        )
+        for ps in profiled:
+            sig = ps.pattern_signature
+            if sig in committed_snapshot:
```
cjluo-nv: > _Bot comment._

The snapshots and pattern-specific benchmark history address much of the previous finding, but this loop only validates committed signatures that happen to remain in `profiled`. If a committed pattern disappears entirely, the `if sig in committed_snapshot` body never checks it, and the aggregate count can be satisfied by other patterns. Build a signature-to-measurements mapping after restart, assert every snapshotted signature exists, and compare its measurements directly. Also require `interrupted_sig is not None` before restart and assert its resumed measurement unconditionally; the current optional assertion silently passes if saving the active signature regresses. These are the remaining gaps in the previously requested recovery-contract test.

tests/unit/onnx/quantization/autotune/test_remote_connectivity.py
```diff
@@ -0,0 +1,889 @@
...
+  baseline, per-scheme, and final benchmark failures.
+- Recovery boundary: committed regions survive restart; the active region
+  may be re-profiled from scratch (candidates/measurements are not serialized).
+"""
+
+import logging
+from unittest.mock import MagicMock, call, patch
+
+import pytest
+
```
cjluo-nv, line 35: > _Bot comment._

Move the `import yaml` inside `test_restart_skips_committed_regions_resumes_active` into this module-level import section. PyYAML is already a declared dependency, and no circular-import or optional-dependency reason is given for keeping it local. This is a minor convention issue, not independently blocking.

cjluo-nv (commented):

Comment: recovery coverage improved, but the committed-pattern preservation assertion remains incomplete and the PR description still contradicts the implementation.

Needs action:
- [ ] :speech_balloon: Author added snapshots and pattern tracking — make preservation assertions unconditional in `test_remote_connectivity.py`; see inline comment.
- [ ] :speech_balloon: Author previously replied “updated” — correct the PR body's mid-trtexec row to describe persistent-loss detection and missed disconnects that recover before the post-check.
- [ ] :speech_balloon: Author explained unreliable stderr classification and shared-runner separation — copy that rationale into the PR body, including TCP authentication limitations.
- [ ] Move the new function-local `yaml` import in `test_remote_connectivity.py` to module scope.

No action needed:
- :heavy_check_mark: Resolved: persistent disconnect propagation through `run()`, interrupted-workflow setup, and pattern-specific resumed benchmark tracking. Earlier validation, credential-safe warnings, checkpoint handling, and region-level recovery documentation remain addressed.
- Design alternatives checked: `_run_trtexec` result classification, a shared-runner guard, and benchmark-local stdlib probing. Existing checkpoint serialization is reused; no additional dependency is needed.
- Existing tests are not removed or weakened; the new header matches `LICENSE_HEADER`. Review was static; tests were not executed.

cjluo-nv (commented):

Nudge: the remaining recovery-test concerns are resolved, but the PR description still needs the behavior correction and design rationale requested previously.

Needs action:
- [ ] :speech_balloon: Author previously replied “updated” — correct the PR body's mid-trtexec row: persistent loss triggers checkpoint-and-exit; disconnects that recover before the post-check can still become scheme errors.
- [ ] :speech_balloon: Author explained unreliable stderr classification and shared-runner separation — copy that rationale into the PR body, comparing benchmark-local TCP probing with `_run_trtexec` result classification or an optional runner guard, and noting TCP cannot detect authentication failures.

No action needed:
- :heavy_check_mark: Resolved since the last review: unconditional committed-pattern preservation and interrupted-pattern assertions (critical coverage), and moving `yaml` to module scope (minor convention). Earlier disconnect propagation, validation, credential-safe warnings, and checkpoint fixes remain intact.
- Design alternatives were rechecked against `ort_utils.py` and `pyproject.toml`; existing checkpoint serialization and stdlib networking suffice without another dependency.
- No existing tests were removed or weakened; the new header matches `LICENSE_HEADER`. Review was static; tests were not executed.

ajrasane (approved):

PR #2079, opened 2026-08-05
Fix 2-GPU test_model_load_utils hang; test import and fixture cleanup

kevalmorabia97:
### What does this PR do?

Type of change: Bug fix + new tests (test-suite only; changes are confined to `tests/`)

- **Fix `test_parallel_load_and_export` hang on 2 GPUs.** The temp paths were built from `os.getpid()` *inside* the workers, so each rank got a different `ckpt_dir`; rank 1 then took `_resolve_checkpoint_dir`'s Hub branch and blocked on an extra barrier while rank 0 ran the loader's broadcasts. The checkpoint is now built once in the parent under `tmp_path` and passed in, so all ranks agree on the path.
- **Imports moved to module top** across `tests/gpu*` and `tests/examples`; function-local imports kept only where guarded (`importorskip`/`try`), where the import *is* the test (JIT compile), or where it must follow `sys.path` setup.
- **Reuse `_test_utils` instead of local copies:** added `get_tiny_mixtral`; deduped `assert_nodes_are_quantized` (5 copies), the accelerate-offload/layerwise config helpers, `make_quant_attention`, `get_dflash_config`, the NVFP4 amax assertions, and 3 copies of the `tiny_wan22_path` fixture.
- **Shared model-dir fixtures assert they were not modified** (`assert_unmodified_tree`): a file manifest is compared on teardown, so a test that writes into a session-scoped fixture directory fails instead of silently changing what later tests see.
- **Dropped dependency guards the CI env already guarantees** (diffusers, tensorrt_llm in `gpu_trtllm`, transformer_engine in `gpu_megatron`, transformers in examples) so a missing dep fails loudly instead of skipping.
- **`test_heterogenous_sharded_state_dict` is skipped on Blackwell** (sm_120), matching the existing marker and its TE/CUDA-13 rationale — same tracking issue as #1901.

### Testing

Local, 2x RTX 6000 Ada: `tests/gpu/torch/utils/test_model_load_utils.py` passes on 2 GPUs and on 1 GPU (previously hung on 2). Also ran the touched files in `tests/unit` (608 passed) and `tests/gpu` (~340 passed).

- Did you write any new necessary tests?: N/A — test-only PR
- Did you get Claude approval on this PR?: ❌ — not yet run

## Summary by CodeRabbit

* **Tests**
  * Added reusable validation utilities for generated files, quantization behavior, attention modules, model fixtures, offloading, and speculative decoding.
  * Expanded coverage for tiny Wan, Mixtral, Llama, and related model scenarios.
  * Consolidated duplicated setup and assertions across ONNX, GPU, quantization, export, and sparsity tests.
  * Improved fixture integrity checks, NVFP4 validation, and handling of identity inputs.
  * Reduced unnecessary dependency-based skips and isolated known platform-specific flakiness.

### DIFF
--- a/tests/_test_utils/fs_utils.py
+++ b/tests/_test_utils/fs_utils.py
@@ -19,8 +19,6 @@ from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
 
-__all__ = ["assert_unmodified_tree"]
-
 
 def _manifest(root: Path) -> dict[str, tuple[int, int]]:
     """``relative path -> (size, mtime_ns)`` for every file below ``root``."""
--- a/tests/_test_utils/fs_utils.py
+++ b/tests/_test_utils/fs_utils.py
@@ -15,39 +15,45 @@
 
 """Filesystem helpers for tests."""
 
-import stat
 from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
 
-# Read + traverse for everyone, write for no one.
-_RO_DIR = 0o555
-_RO_FILE = 0o444
+__all__ = ["assert_unmodified_tree"]
 
 
-def _set_read_only(root: Path, read_only: bool) -> None:
-    for path in [root, *root.rglob("*")]:
-        if path.is_symlink():
-            continue
-        if read_only:
-            path.chmod(_RO_DIR if path.is_dir() else _RO_FILE)
-        else:
-            path.chmod(path.stat().st_mode | stat.S_IWUSR)
+def _manifest(root: Path) -> dict[str, tuple[int, int]]:
+    """``relative path -> (size, mtime_ns)`` for every file below ``root``."""
+    return {
+        str(p.relative_to(root)): (p.stat().st_size, p.stat().st_mtime_ns)
+        for p in root.rglob("*")
+        if p.is_file() and not p.is_symlink()
+    }
 
 
 @contextmanager
-def read_only_tree(path: Path | str) -> Iterator[Path]:
-    """Make ``path`` read-only for the body, restoring write permission on exit.
+def assert_unmodified_tree(path: Path | str) -> Iterator[Path]:
+    """Fail if anything under ``path`` is added, removed, or rewritten inside the ``with``.
 
     For session/module-scoped model-directory fixtures: a test that writes into a shared
-    directory silently changes what every later test sees, so make that fail loudly at the
-    write instead. Write permission is restored on teardown so pytest's ``tmp_path``
-    retention cleanup can still remove the tree.
+    directory silently changes what every later test sees. Comparing a file manifest on
+    teardown catches that. ``chmod``-ing the tree read-only would report at the write rather
+    than at teardown, but it only works for an unprivileged user -- root has
+    ``CAP_DAC_OVERRIDE`` and writes straight through the permission bits, and the CI
+    containers run as root.
     """
     path = Path(path)
-    _set_read_only(path, read_only=True)
-    try:
-        yield path
-    finally:
-        if path.exists():
-            _set_read_only(path, read_only=False)
+    before = _manifest(path)
+    yield path
+    if not path.exists():
+        return
+    after = _manifest(path)
+    added = sorted(after.keys() - before.keys())
+    removed = sorted(before.keys() - after.keys())
+    changed = sorted(k for k in before.keys() & after.keys() if before[k] != after[k])
+    if added or removed or changed:
+        raise AssertionError(
+            f"shared fixture directory {path} was modified by a test "
+            f"(added={added}, removed={removed}, changed={changed}); "
+            "copy it into the test's own tmp_path instead of writing into the shared tree"
+        )
--- a/tests/_test_utils/fs_utils.py
+++ b/tests/_test_utils/fs_utils.py
@@ -1,0 +1,53 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Filesystem helpers for tests."""
+
+import stat
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+
+# Read + traverse for everyone, write for no one.
+_RO_DIR = 0o555
+_RO_FILE = 0o444
+
+
+def _set_read_only(root: Path, read_only: bool) -> None:
+    for path in [root, *root.rglob("*")]:
+        if path.is_symlink():
+            continue
+        if read_only:
+            path.chmod(_RO_DIR if path.is_dir() else _RO_FILE)
+        else:
+            path.chmod(path.stat().st_mode | stat.S_IWUSR)
+
+
+@contextmanager
+def read_only_tree(path: Path | str) -> Iterator[Path]:
+    """Make ``path`` read-only for the body, restoring write permission on exit.
+
+    For session/module-scoped model-directory fixtures: a test that writes into a shared
+    directory silently changes what every later test sees, so make that fail loudly at the
+    write instead. Write permission is restored on teardown so pytest's ``tmp_path``
+    retention cleanup can still remove the tree.
+    """
+    path = Path(path)
+    _set_read_only(path, read_only=True)
+    try:
+        yield path
+    finally:
+        if path.exists():
+            _set_read_only(path, read_only=False)
--- a/tests/_test_utils/fs_utils.py
+++ b/tests/_test_utils/fs_utils.py
@@ -46,7 +46,7 @@ def assert_unmodified_tree(path: Path | str) -> Iterator[Path]:
     before = _manifest(path)
     yield path
     if not path.exists():
-        return
+        raise AssertionError(f"shared fixture directory {path} was deleted by a test")
     after = _manifest(path)
     added = sorted(after.keys() - before.keys())
     removed = sorted(before.keys() - after.keys())
--- a/tests/_test_utils/torch/quantization/attention.py
+++ b/tests/_test_utils/torch/quantization/attention.py
@@ -15,6 +15,10 @@
 
 """Shared attention-quantization fixtures for the unit and gpu attention tests."""
 
+import pytest
+
+pytest.importorskip("transformers")
+
 from transformers import LlamaConfig
 from transformers.models.llama.modeling_llama import LlamaAttention
 
--- a/tests/_test_utils/torch/quantization/attention.py
+++ b/tests/_test_utils/torch/quantization/attention.py
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Shared attention-quantization fixtures for the unit and gpu attention tests."""
+
+from transformers import LlamaConfig
+from transformers.models.llama.modeling_llama import LlamaAttention
+
+from modelopt.torch.quantization.plugins.huggingface import _QuantAttention
+
+
+def make_quant_attention(hidden_size=128, num_q_heads=4, num_kv_heads=2):
+    """A single ``_QuantAttention``-converted Llama attention layer, pinned to the sdpa impl."""
+    config = LlamaConfig(
+        hidden_size=hidden_size,
+        num_attention_heads=num_q_heads,
+        num_key_value_heads=num_kv_heads,
+    )
+    quant_attention = _QuantAttention.convert(LlamaAttention(config, layer_idx=0))
+    quant_attention.config._attn_implementation = "sdpa"
+    return quant_attention
--- a/tests/_test_utils/torch/transformers_models.py
+++ b/tests/_test_utils/torch/transformers_models.py
@@ -35,6 +35,7 @@ from transformers import (
     GptOssConfig,
     LlamaConfig,
     LlamaForSequenceClassification,
+    MixtralConfig,
     NemotronConfig,
     PreTrainedModel,
     Qwen3Config,
@@ -561,6 +562,36 @@ def create_tiny_gpt_oss_dir(
     return _create_tiny_llm_dir(
         Path(tmp_path) / "tiny_gpt_oss",
         get_tiny_gpt_oss,
+        with_tokenizer=with_tokenizer,
+        **config_kwargs,
+    )
+
+
+##### MIXTRAL #####
+def get_tiny_mixtral(**config_kwargs) -> PreTrainedModel:
+    set_seed(SEED)
+    kwargs = {
+        "dtype": torch.bfloat16,
+        "hidden_size": 32,
+        "intermediate_size": 32,
+        "num_hidden_layers": 2,
+        "num_attention_heads": 4,
+        "num_key_value_heads": 2,
+        "num_local_experts": 4,
+        "num_experts_per_tok": 2,
+        "max_position_embeddings": 32,
+        "vocab_size": 32,
+    }
+    kwargs.update(config_kwargs)
+    return AutoModelForCausalLM.from_config(MixtralConfig(**kwargs))
+
+
+def create_tiny_mixtral_dir(
+    tmp_path: Path | str, with_tokenizer: bool = False, **config_kwargs
+) -> Path:
+    return _create_tiny_llm_dir(
+        Path(tmp_path) / "tiny_mixtral",
+        get_tiny_mixtral,
         with_tokenizer=with_tokenizer,
         **config_kwargs,
     )
--- a/tests/_test_utils/torch/transformers_models.py
+++ b/tests/_test_utils/torch/transformers_models.py
@@ -586,17 +586,6 @@ def get_tiny_mixtral(**config_kwargs) -> PreTrainedModel:
     return AutoModelForCausalLM.from_config(MixtralConfig(**kwargs))
 
 
-def create_tiny_mixtral_dir(
-    tmp_path: Path | str, with_tokenizer: bool = False, **config_kwargs
-) -> Path:
-    return _create_tiny_llm_dir(
-        Path(tmp_path) / "tiny_mixtral",
-        get_tiny_mixtral,
-        with_tokenizer=with_tokenizer,
-        **config_kwargs,
-    )
-
-
 ##### LLAMA #####
 def get_tiny_llama(**config_kwargs) -> PreTrainedModel:
     set_seed(SEED)
@@ -700,10 +689,6 @@ def get_tiny_bert(**config_kwargs) -> PreTrainedModel:
     }
     kwargs.update(config_kwargs)
     return AutoModelForQuestionAnswering.from_config(BertConfig(**kwargs))
-
-
-def create_tiny_bert_dir(tmp_path: Path | str, **config_kwargs) -> Path:
-    return _create_tiny_llm_dir(Path(tmp_path) / "tiny_bert", get_tiny_bert, **config_kwargs)
 
 
 ##### ViT (vision) #####
--- a/tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
+++ b/tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
@@ -561,6 +561,7 @@ def test_homogeneous_sharded_state_dict_hybrid(dist_workers, tmp_path, config):
         mixed_block_size_config,
     ],
 )
+@skip_flaky_on_blackwell
 def test_heterogenous_sharded_state_dict(dist_workers, tmp_path, config):
     dist_workers.run(
         partial(_test_sharded_state_dict, tmp_path, config, 256, None, False, False, {}),
--- a/tests/unit/onnx/quantization/test_qdq_rules_int8.py
+++ b/tests/unit/onnx/quantization/test_qdq_rules_int8.py
@@ -32,19 +32,10 @@ from _test_utils.onnx.lib_test_models import (
     build_small_grouped_conv_model,
     export_as_onnx,
 )
+from _test_utils.onnx.quantization.utils import assert_nodes_are_quantized
 
 from modelopt.onnx.quantization.quantize import quantize
 from modelopt.onnx.utils import get_opset_version, save_onnx
-
-
-def assert_nodes_are_quantized(nodes):
-    for node in nodes:
-        for inp_idx, inp in enumerate(node.inputs):
-            if isinstance(inp, gs.Variable) and node.i(inp_idx).op != "Identity":
-                assert node.i(inp_idx).op == "DequantizeLinear", (
-                    f"Input '{inp.name}' of node '{node.name}' is not quantized but should be!"
-                )
-    return True
 
 
 def assert_nodes_are_not_quantized(nodes):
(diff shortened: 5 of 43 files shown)


tests/_test_utils/torch/quantization/attention.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Shared attention-quantization fixtures for the unit and gpu attention tests."""
+
+from transformers import LlamaConfig
```
cjluo-nv, line 22: > _Bot comment._

This module imports `transformers` (and, below, `modelopt.torch.quantization.plugins.huggingface`, which needs transformers) unguarded at module top, and `tests/unit/torch/quantization/plugins/test_attention_quant.py` imports it at module scope.

Ruff/isort sorts `_test_utils.torch.quantization.attention` *before* `_test_utils.torch.transformers_models`, so the `pytest.importorskip("transformers")` that `transformers_models.py` runs at import time (which is what currently makes that test module skip cleanly) no longer executes first. In the `partial_unit(subset='torch')` lane — `session.install("megatron-core", ".[dev-test]")`, i.e. no `hf` extra, no transformers — this becomes a collection error instead of a skip. The test file's own `transformers = pytest.importorskip("transformers")` now runs too late to help.

Suggest adding `pytest.importorskip("transformers")` at the top of this helper module (mirroring `transformers_models.py`) before the `transformers` / plugin imports. Note the PR body says `tests/unit` passed locally, but that was presumably the full-install lane, which wouldn't catch this.

cjluo-nv (commented):

The core hang fix is correct and well-reasoned (parent-built `ckpt_dir` under `tmp_path` removes the per-rank `os.getpid()` divergence that sent rank 1 down `_resolve_checkpoint_dir`'s Hub branch and its extra barrier; `export_hf_checkpoint` already mkdirs the export dir, so dropping the rank-0 makedirs+barrier is safe). Design-wise the new helpers reuse the existing `tests/_test_utils` pattern rather than introducing a competing one, so no architectural objection.

However, several things in the surrounding cleanup look likely to break lanes that weren't exercised locally, and one change touches third-party licensing:

1. `tests/_test_utils/torch/quantization/attention.py` imports `transformers` (and `modelopt...plugins.huggingface`) unguarded at module top and is imported at module scope by `tests/unit/torch/quantization/plugins/test_attention_quant.py`. Because `_test_utils.torch.quantization.attention` sorts *before* `_test_utils.torch.transformers_models`, the `pytest.importorskip("transformers")` that used to fire first no longer does — in `nox -s "partial_unit(subset='torch')"` (installs only `megatron-core` + `.[dev-test]`, i.e. no `hf` extra) that turns a clean skip into a collection error. The PR itself keeps function-local imports elsewhere with the comment "unit tests must import without transformers installed", so this looks like an oversight.
2. `examples/alpamayo/requirements.txt` adds only the git dependency, but `tests/examples/alpamayo/test_quantize.py` imports `pandas` and calls `to_parquet` (needs pyarrow), and the torch example lane installs `[hf,dev-test]` which has neither. Since the tests were previously auto-skipped this was invisible. Separately, adding a new `git+https://github.com/nvlabs/alpamayo@…` dependency is a third-party licensing/dependency decision (PR body asserts Apache-2.0) that needs human/OSRB sign-off — that alone blocks an approve on licensing grounds.
3. `read_only_tree` is a no-op for root, which is how these tests run inside the CI containers (root bypasses DAC write bits). The "fail loudly at the write" property therefore only holds for local non-root runs, so it can produce CI-vs-local divergence rather than a guarantee. Worth documenting, or enforcing another way.
4. The drive-by `@skip_flaky_on_blackwell` on `test_heterogenous_sharded_state_dict` reduces coverage on the Blackwell runners and isn't mentioned anywhere in the PR description.

Also consider splitting: the CI lane restructuring + new alpamayo dependency carry materially different risk from the mechanical import/dedup cleanup and would be easier to review (and revert) on their own. No prompt-injection content was present in the PR text.

examples/alpamayo/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Alpamayo 1 (formerly Alpamayo-R1) provides the ``alpamayo_r1`` package that quantize.py
+# imports at module load. Pinned to the commit this example was tested against; see README.md.
+# The package itself is public -- only the model weights and the AV dataset are gated.
+git+https://github.com/nvlabs/alpamayo@4cda35d
```
cjluo-nv: > _Bot comment._

Two things here:

1. `tests/examples/alpamayo/test_quantize.py` imports `pandas` and calls `DataFrame.to_parquet` (so pyarrow too), and `quantize.read_clip_ids_from_parquet` needs the same at runtime. The torch example lane installs `-e ".[hf,dev-test]"`, which provides neither (pandas only comes via the `puzzletron` extra). Unless `alpamayo_r1` happens to pull both in transitively, enabling this matrix entry will fail at collection. Please pin `pandas` (and `pyarrow`) explicitly — the example itself needs them regardless of the tests.
2. Adding a `git+https://github.com/nvlabs/alpamayo@<sha>` dependency is a new third-party dependency sourced from an external repo. The PR body asserts Apache-2.0, but per `CONTRIBUTING.md` this needs an explicit human/legal sign-off rather than a reviewer taking the assertion at face value — flagging so an owner signs off.

tests/_test_utils/fs_utils.py
```diff
@@ -0,0 +1,53 @@
...
+        if read_only:
+            path.chmod(_RO_DIR if path.is_dir() else _RO_FILE)
+        else:
+            path.chmod(path.stat().st_mode | stat.S_IWUSR)
+
+
+@contextmanager
+def read_only_tree(path: Path | str) -> Iterator[Path]:
+    """Make ``path`` read-only for the body, restoring write permission on exit.
+
```
cjluo-nv, line 35: > _Bot comment._

Worth noting in the docstring (or reconsidering): the CI example/GPU containers run as root, and root bypasses DAC write permission bits, so `chmod 0o444/0o555` does not actually prevent writes there. The stated property ("a test that writes into a shared directory ... fail loudly at the write") therefore only holds for local non-root runs. Net effect is asymmetric: CI keeps silently mutating shared fixtures, while a developer running locally as a normal user may hit a `PermissionError` CI never shows. Either document that limitation or enforce it a way that works under root (e.g. snapshot + compare a manifest of the tree at fixture teardown).

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -561,6 +561,7 @@ def test_homogeneous_sharded_state_dict_hybrid(dist_workers, tmp_path, config):
         mixed_block_size_config,
     ],
 )
+@skip_flaky_on_blackwell
```
cjluo-nv, line 564: > _Bot comment._

This adds a Blackwell skip to `test_heterogenous_sharded_state_dict`, which isn't mentioned anywhere in the PR description ("CI lane fixes" doesn't cover silently dropping a test on the sm_120 runners that most of this CI now uses). It's consistent with the existing marker and its documented TE/CUDA-13 rationale, so I'm not objecting to the change itself — please call it out in the PR body and link it to the same tracking issue as #1901 so it doesn't become a permanent silent gap.

tests/_test_utils/torch/quantization/attention.py
kevalmorabia97: Fixed in 2ed5db0. `pytest.importorskip("transformers")` now runs at the top of this helper, before the `transformers` / huggingface-plugin imports, mirroring what `transformers_models.py` does. Your diagnosis was exactly right, including the isort-ordering detail — this was the `partial_unit(subset='torch')` collection failure on this PR, and the local run that I reported as green was indeed the full install.

tests/_test_utils/fs_utils.py
kevalmorabia97: Fixed in 2ed5db0, by dropping the `chmod` layer rather than documenting around it — you're right that root has `CAP_DAC_OVERRIDE` and the CI containers run as root, so the permission bits were checking nothing exactly where it matters.

The helper is now `assert_unmodified_tree`: it snapshots a manifest (relative path -> size + mtime_ns) at fixture setup and compares it on teardown, which holds under root too. It reports at teardown rather than at the offending write, which is the trade-off for working everywhere. Renamed since it no longer touches permissions.

tests/_test_utils/fs_utils.py
```diff
@@ -0,0 +1,53 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Filesystem helpers for tests."""
+
+import stat
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
```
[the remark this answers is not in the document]
kevalmorabia97: Done in 2ed5db0 — `__all__ = ["assert_unmodified_tree"]` added, with `_manifest` left private. (The helper was also rewritten in that commit: the `chmod` layer is gone, so `read_only_tree` is now `assert_unmodified_tree`.)

tests/unit/torch/export/test_quant_aware_conversion.py
```diff
@@ -355,21 +356,9 @@ def test_revert_quant_config_names_mapper():
...
-            num_hidden_layers=1,
-            num_attention_heads=4,
-            num_key_value_heads=2,
-            num_local_experts=2,
-            num_experts_per_tok=2,
-            vocab_size=64,
-            max_position_embeddings=64,
-        )
-    )
+    from _test_utils.torch.transformers_models import get_tiny_mixtral
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 2ed5db0 — added `# Import stays function-local: the helper needs transformers, which unit tests run without.` above the `importorskip`.

examples/alpamayo/requirements.txt
kevalmorabia97: Moot now — the alpamayo lane is out of this PR (which is `tests/`-only again) and `examples/alpamayo/requirements.txt` is gone with it. It failed at *Install dependencies* for an unrelated reason: `alpamayo_r1` requires `flash-attn>=2.8.3`, which has no PyPI wheels, so pip builds from source and the build backend can't see torch under build isolation. That needs `--no-build-isolation` (not expressible in a requirements.txt, which is all the shared example runner consumes) plus a much longer timeout.

On the dependency points, for whenever we revisit: `pandas`/`pyarrow` are actually reachable in that lane already (`datasets`, from the `hf` extra, hard-requires both), and the repo is Apache-2.0 — but agreed that adding a git-sourced third-party dependency wants an explicit owner sign-off rather than my say-so.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
kevalmorabia97: Called out in the PR description now, linked to the same tracking issue as #1901. Thanks for catching that it was going in unmentioned.

tests/_test_utils/torch/transformers_models.py
```diff
@@ -566,6 +567,36 @@ def create_tiny_gpt_oss_dir(
...
+
+def create_tiny_mixtral_dir(
+    tmp_path: Path | str, with_tokenizer: bool = False, **config_kwargs
+) -> Path:
+    return _create_tiny_llm_dir(
+        Path(tmp_path) / "tiny_mixtral",
+        get_tiny_mixtral,
+        with_tokenizer=with_tokenizer,
+        **config_kwargs,
+    )
```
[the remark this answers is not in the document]
kevalmorabia97: Skipping this one for consistency: 30 of the 36 functions in this module have no docstring, including every public `get_tiny_*` / `create_tiny_*` factory (`get_tiny_llama`, `get_tiny_qwen3`, `get_tiny_gpt_oss`, ...). Documenting only the new Mixtral pair would make it the odd one out. Happy to do the whole file as a separate cleanup if that's wanted.

tests/unit/onnx/quantization/test_qdq_rules_int8.py
```diff
@@ -77,7 +68,7 @@ def test_bias_add_rule(tmp_path):
 
     #   Check that all Conv nodes are quantized
     conv_nodes = [n for n in graph.nodes if n.op == "Conv"]
-    assert assert_nodes_are_quantized(conv_nodes)
+    assert assert_nodes_are_quantized(conv_nodes, ignore_identity_inputs=True)
```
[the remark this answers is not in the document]
kevalmorabia97: Pushing back on this one — I removed `ignore_identity_inputs=True` and re-ran the file:

```
FAILED test_resnet_residual_connections
FAILED test_resnet_residual_connection_with_downsample
E  AssertionError: Input 'onnx::Conv_25' of node '/0/conv2/Conv' is not quantized but should be!
E  AssertionError: Input 'onnx::Conv_36' of node '/0/downsample/downsample.0/Conv' is not quantized but should be!
```

The ResNet residual builders do feed conv inputs through `Identity`, so the flag is load-bearing. It also preserves this file's pre-existing behaviour exactly: the local helper it replaced skipped `Identity` producers for every call site in the file. Left as-is.

kevalmorabia97:
### Update — force-pushed `2ed5db0`

**Scope narrowed:** this PR is now `tests/`-only. The CI workflow changes (single merged TRT-LLM lane, `gpt-oss` moved to the TRT-LLM image) are out and will come as a separate PR, along with the `tests/examples/gpt-oss` guard removal that only makes sense once that lane moves. The `alpamayo` lane is dropped: `alpamayo_r1` requires `flash-attn>=2.8.3`, which has no PyPI wheels, so pip builds it from source and the build backend can't see torch under build isolation — that needs `--no-build-isolation` (not expressible in a `requirements.txt`, which is all the shared example runner consumes) plus a much longer timeout, for three CPU-only helper tests.

**Both red jobs fixed:**
- `partial-install (torch)` — `_test_utils/torch/quantization/attention.py` imported `transformers` unguarded, and isort sorts it before `transformers_models`, so the `importorskip` that used to make `test_attention_quant.py` skip cleanly no longer ran first. It now guards at the top of the helper itself.
- `torch (alpamayo)` — lane removed, per above.

**Review feedback:**
- The shared-fixture guard no longer uses `chmod`. Root has `CAP_DAC_OVERRIDE` and the CI containers run as root, so the permission bits checked nothing where it mattered. `read_only_tree` is now `assert_unmodified_tree`, which compares a file manifest (path -> size + mtime_ns) on teardown and therefore holds under root too.
- Added the missing `__all__` and the local-import rationale comment.
- Two findings I did not apply, with replies in-thread: the `ignore_identity_inputs=True` bypass is load-bearing (removing it fails `test_resnet_residual_connections` and `test_resnet_residual_connection_with_downsample` — those ResNet builders do feed conv inputs through `Identity`), and the Mixtral docstrings are skipped for consistency with the 30 other undocumented factories in that file.
- The Blackwell skip is now called out in the PR description, linked to #1901.

cjluo-nv (commented):

Re-review of the now `tests/`-only PR (41 files, +542/-588). Design protocol applied: the new helpers (`fs_utils.py`, `torch/quantization/{attention,offload}.py`, `torch/speculative/dflash.py`, additions to `onnx/quantization/utils.py` and `torch/quantization/quant_utils.py`) extend the existing `tests/_test_utils` package rather than introducing a competing helper layer, and each one replaces 2–5 verbatim copies, so there is no second system being introduced and no in-repo/dependency alternative that would do this better.

Verification of the previous round's blockers — all resolved:
- `tests/_test_utils/torch/quantization/attention.py` now runs `pytest.importorskip("transformers")` before the `transformers` / HF-plugin imports, so `partial_unit(subset='torch')` skips instead of erroring at collection.
- The alpamayo lane, `examples/alpamayo/requirements.txt` and all CI-workflow changes are out; no licensing/OSRB surface remains (no LICENSE/header/vendoring changes, new files carry the standard NVIDIA Apache-2.0 header).
- The root-bypassable `chmod` guard is gone; `read_only_tree` → `assert_unmodified_tree` (size+mtime_ns manifest compared on teardown), which works under root.
- Dropped dependency guards check out against the lanes: `gpu` installs `.[all,dev-test]` (diffusers comes via `hf`), `gpu_trtllm`/`gpu_megatron` get tensorrt_llm / transformer_engine from their containers.

Spot-checks of the mechanical parts look right: `_MIXTRAL_KWARGS` / `_LLAMA_KWARGS` preserve the old shapes and (where it matters for tolerances) the old fp32 dtype; `ignore_identity_inputs` defaults to `False` so the three files whose local helper did *not* skip Identity keep their stricter assertion, and only the ResNet/concat call sites opt in; removed module-top imports (`LlamaConfig`, `gs`, `diffusers`) are still used or genuinely unused. The core hang fix (parent-built `ckpt_dir`/`export_dir` under `tmp_path`, one path for all ranks) is unchanged and correct.

tests/_test_utils/fs_utils.py
```diff
@@ -0,0 +1,59 @@
...
+    than at teardown, but it only works for an unprivileged user -- root has
+    ``CAP_DAC_OVERRIDE`` and writes straight through the permission bits, and the CI
+    containers run as root.
+    """
+    path = Path(path)
+    before = _manifest(path)
+    yield path
+    if not path.exists():
+        return
+    after = _manifest(path)
```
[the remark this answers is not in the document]
kevalmorabia97: Not applying this one — the premise doesn't hold for the only way this helper is used (inside a generator fixture).

pytest does not throw test failures into a fixture's generator; it catches them and finalizes the fixture normally, so the post-`yield` comparison already runs. Reproduced with the exact scenario described (a test that mutates the shared dir *and* fails), using the current plain-`yield` code:

```
E   AssertionError: the real test failure          <- the test
E   AssertionError: MUTATED: contents changed      <- the guard, at teardown
1 failed, 1 error
```

A `finally` would only change behaviour when something is thrown *into* the generator — `GeneratorExit` from `gen.close()`, or `KeyboardInterrupt`. In those paths the session is already being torn down, so there are no later tests left to protect and the report has no value.

I did prototype the `try/except/else` version (comparison on both paths, warning instead of raise while unwinding so it can't mask the failure being propagated) and reverted it as unnecessary complexity. Happy to reinstate if you can point at a path where a test failure does reach the generator.

kevalmorabia97:
### Update — `c7e696e` + `3022924`

- **Windows crash fixed.** `tests/unit/torch/opt/plugins/test_hf_patching.py::test_nested_model_save_restore` takes the `skip_on_windows` fixture, matching `test_quantize_cpu.py::test_save_restore`. The pytest process was dying there with `STATUS_ILLEGAL_INSTRUCTION` (0xC000001D) and taking the whole unit run with it, which is why the failure looked anonymous — the last line printed was the preceding file's test.
- **`assert_unmodified_tree` now fails when the fixture root is deleted outright**, instead of returning silently. That was the loudest mutation it could see and the one case it ignored.

On the re-review's other two points:

- **Example/megatron lane coverage:** covered. The lanes that exercise the guarded fixtures (`torch (llm_qat)`, `(llm_distill)`, `(llm_sparsity)`, `(specdec_bench)`, `onnx (diffusers)`) all ran green on `2ed5db0`, which already had the guard active on `tiny_llama_path` / `tiny_qwen3_path` / `tiny_gpt_oss_path` — so it fired on nothing. @kevalmorabia97 additionally ran `examples/megatron_bridge` and `tests/gpu_megatron` locally, both passing.
- **Blackwell skip:** signed off by @kevalmorabia97 as repo owner/codeowner.

The `finally`-on-generator-close suggestion is still not applied, for the reason in that thread: pytest doesn't throw test failures into a fixture's generator, so the comparison already runs after ordinary failures; the only paths it would add are `GeneratorExit` / `KeyboardInterrupt`, where the session is already aborting and no later tests remain to protect.

cjluo-nv (commented):

Re-review #3 of this `tests/`-only PR (42 files, +544/-589). Design protocol re-applied: the new helpers (`fs_utils.py`, `torch/quantization/{attention,offload}.py`, `torch/speculative/dflash.py`, additions to `onnx/quantization/utils.py` and `torch/quantization/quant_utils.py`) extend the existing `tests/_test_utils` package rather than introducing a competing layer, and each replaces 2–5 verbatim copies — no second system, no in-repo/dependency alternative that fits better. Core hang fix unchanged and correct (parent-built `ckpt_dir`/`export_dir` under `tmp_path` gives every rank one path, removing the `os.getpid()` divergence into `_resolve_checkpoint_dir`'s Hub branch and the mismatched barrier).

Verified closed since the last pass:
- `assert_unmodified_tree` now raises on a deleted fixture root (`3022924`); `__all__` present, manifest is size+`mtime_ns` so it works under root in the CI containers.
- `tests/_test_utils/torch/quantization/attention.py` runs `pytest.importorskip("transformers")` before the `transformers`/HF-plugin imports, and it still sorts ahead of the `modelopt...plugins.huggingface` import in `tests/unit/torch/quantization/plugins/test_attention_quant.py`, so the torch-only unit lane skips instead of erroring at collection.
- Lane coverage for the guarded fixtures and the Blackwell skip are signed off by the owner.
- Spot-checks hold: `_MIXTRAL_KWARGS`/`_LLAMA_KWARGS` preserve the old shapes and the fp32 dtype where tolerances depend on it; `ignore_identity_inputs` defaults to `False` so only the ResNet/concat call sites relax; `LlamaConfig`/`LlamaAttention`/`_QuantAttention` are still used by `test_kitchen_fa` after the local-helper removal; `test_wan22_calibrated_export` only reads the now-session-scoped `tiny_wan22_path` and exports into `tmp_path`.
- No licensing surface: no LICENSE/NOTICE/vendoring changes; new files carry the project's standard Apache-2.0 NVIDIA header (year reads 2024 vs. the canonical `LICENSE_HEADER`'s 2026 — cosmetic only).

Nudging rather than approving only because of one drive-by that no reviewer has looked at yet (see reason) plus the overall size of the mechanical cleanup.

tests/_test_utils/fs_utils.py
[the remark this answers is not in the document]
kevalmorabia97: Reverting this one at maintainer request: `__all__` isn't used in the test-utils package (3 of ~38 modules had it, all pre-existing), so it's been dropped from `fs_utils.py` again rather than added to the other new helper modules. Flagging here so it doesn't look silently undone.

kevalmorabia97:
### Update — `687f66c`

Re-review #3 points:

- **Windows skip.** Fair catch that "Flaky on Windows" didn't match a `STATUS_ILLEGAL_INSTRUCTION` crash. Comment corrected. Deliberately **not** tracked: Windows is a supported platform for `modelopt.onnx` only, so a torch KD save/restore crash there is out of scope — confirmed by @kevalmorabia97, who also confirmed masking it is intended.
- **`create_tiny_mixtral_dir` unused.** Correct — removed. Also removed the pre-existing unused `create_tiny_bert_dir`; `get_tiny_bert` stays (used by `test_attention_quant.py:120`). No `create_tiny_*_dir` helper is unused now.
- **`__all__` only on `fs_utils.py`.** Resolved in the other direction: `__all__` isn't a convention in `tests/_test_utils` (3 of ~38 modules, all pre-existing), so it was dropped from `fs_utils.py` rather than added to the other new modules.

cjluo-nv (commented):

Re-review #4 of this `tests/`-only PR (42 files, +531/-593). Design protocol re-applied: the helpers (`fs_utils.py`, `torch/quantization/{attention,offload}.py`, `torch/speculative/dflash.py`, additions to `onnx/quantization/utils.py` and `torch/quantization/quant_utils.py`) extend the existing `tests/_test_utils` package rather than standing up a competing helper layer, and each replaces 2–5 verbatim copies — no second system, no better in-repo/dependency alternative. Core hang fix unchanged and correct: the checkpoint/export dirs are built once in the parent under `tmp_path` and passed into the workers, so all ranks agree on the path and rank 1 no longer takes `_resolve_checkpoint_dir`'s Hub branch and its extra barrier.

Verified closed in this pass:
- Windows skip on `test_nested_model_save_restore` — owner-signed-off; Windows support is scoped to `modelopt.onnx`, so masking a torch KD save/restore crash there is intentional and acceptable.
- `__all__` intentionally dropped from the new test-util modules, matching the package convention (only 3 of ~38 pre-existing modules define it); confirmed absent from `fs_utils.py`.
- `create_tiny_bert_dir` and `create_tiny_mixtral_dir` are gone from `transformers_models.py`, `get_tiny_bert` retained (still used by `test_attention_quant.py`); no dangling references anywhere in the tree.
- `assert_unmodified_tree` raises on a deleted fixture root; manifest is size + `mtime_ns`, so it works under root in the CI containers.
- `tests/_test_utils/torch/quantization/attention.py` runs `pytest.importorskip("transformers")` before the `transformers`/HF-plugin imports and still sorts ahead of the `modelopt...plugins.huggingface` import in the unit test, so the torch-only unit lane skips instead of erroring at collection.
- Spot-checks hold: no now-unused module-top imports (`gs` still used in the ONNX tests, `LlamaConfig`/`LlamaAttention`/`_QuantAttention` still used by `test_kitchen_fa`); `_MIXTRAL_KWARGS`/`_LLAMA_KWARGS` preserve the old shapes and the fp32 dtype where tolerances depend on it; `ignore_identity_inputs` defaults to `False` so only the ResNet/concat call sites relax (load-bearing there, per the author's reproduction).
- No licensing surface: no LICENSE/NOTICE/vendoring changes; new files carry the project's standard Apache-2.0 NVIDIA header (year reads 2024 vs. the canonical header's later year — cosmetic only).

The remaining size concern is a soft guideline on a cohesive, mechanical test-only cleanup, and the last open drive-by is now explicitly signed off by the codeowner. Approving.

Complex PR: 36 files changed (≥ 20); 1124 lines changed (≥ 1000); spans 27 directories (≥ 5); 36 existing test files modified or removed. Looping in a human for approval.

PR #2084, opened 2026-08-05
[NVBUG: 6562021] Fix vLLM FlashAttention KV cache layout handling

sychen52:
### What does this PR do?

Type of change: Bug fix

The ModelOptSparseAttentionImpl.forward (FlashAttention path) assumes the paged KV cache has shape [2, num_blocks, page_size, num_kv_heads, head_dim], where dimension 0 is the K/V split. unbind(0) on this shape returns exactly 2 tensors.
Newer vLLM versions changed the FlashAttention kv_cache layout to [num_blocks, 2, page_size, num_kv_heads, head_dim] (same as FlashInfer). unbind(0) now returns num_blocks tensors, causing the unpack error.

Now, we detect the layout first and then unbind.

### Usage

save as before.

### Testing
unittest

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FlashAttention compatibility with vLLM key/value cache layouts.
  * Automatically detects supported cache dimension ordering for accurate sparse-attention processing across legacy and newer layouts.

* **Tests**
  * Expanded coverage for supported cache layouts, cache validation, and paged-cache scenarios.

### DIFF
--- a/modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
+++ b/modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
@@ -47,6 +47,19 @@ from modelopt.torch.kernels.common.attention.triton_fa import attention as trito
 from modelopt.torch.kernels.quantization.attention.bmm2_qdq import fake_quant_v_onwrite
 
 
+@functools.cache
+def _flash_attention_kv_cache_layout() -> str:
+    """Return the installed vLLM backend's K/V packing contract."""
+    cache_shape = FlashAttentionBackend.get_kv_cache_shape(3, 16, 1, 16)
+    if cache_shape == (2, 3, 16, 1, 16):
+        return "kv-first"
+    if cache_shape == (3, 2, 16, 1, 16):
+        return "blocks-first"
+    if cache_shape == (3, 1, 16, 32):
+        return "packed"
+    raise RuntimeError(f"Unsupported vLLM FlashAttention KV cache shape {cache_shape}")
+
+
 def _target_sparse_ratio_for_phase(target_sparse_ratio, phase: str) -> float:
     """Return target sparsity for a phase, defaulting old checkpoint metadata."""
     if isinstance(target_sparse_ratio, float | int):
@@ -514,7 +527,13 @@ class ModelOptSparseAttentionImpl(FlashAttentionImpl):
         if resolved is None:
             return native_forward()
 
-        key_cache, value_cache = kv_cache.unbind(0)
+        cache_layout = _flash_attention_kv_cache_layout()
+        if cache_layout == "kv-first":
+            key_cache, value_cache = kv_cache.unbind(0)
+        elif cache_layout == "blocks-first":
+            key_cache, value_cache = kv_cache.unbind(1)
+        else:
+            key_cache, value_cache = kv_cache.transpose(1, 2).split(self.head_size, dim=-1)
         is_decode_only = attn_metadata.max_query_len <= 1
         common_kw = {
             "layer": layer,
--- a/tests/gpu_vllm/torch/sparsity/attention_sparsity/test_sparse_attn_worker.py
+++ b/tests/gpu_vllm/torch/sparsity/attention_sparsity/test_sparse_attn_worker.py
@@ -28,7 +28,7 @@ import torch
 import vllm
 from vllm.v1.attention.backend import CommonAttentionMetadata
 from vllm.v1.attention.backends import flashinfer as flashinfer_backend
-from vllm.v1.attention.backends.flash_attn import FlashAttentionImpl
+from vllm.v1.attention.backends.flash_attn import FlashAttentionBackend, FlashAttentionImpl
 from vllm.v1.attention.backends.flashinfer import (
     FlashInferBackend,
     FlashInferImpl,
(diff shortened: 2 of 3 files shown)


tests/gpu_vllm/torch/sparsity/attention_sparsity/test_sparse_attn_worker.py
```diff
@@ -541,6 +541,37 @@ def _make_flash_attention_impl(*, sparse=False, quantized=False):
     return impl
 
 
+def _flash_attention_kv_cache(num_blocks, page_size, num_kv_heads, head_size):
```
cjluo-nv, line 544: > _Bot comment._

The sibling GPU test module `tests/gpu_vllm/torch/sparsity/attention_sparsity/test_vllm_plugin.py` still hardcodes the legacy layout and was not updated: `_make_paged_cache` does `torch.stack([k_cache, v_cache], dim=0)`, `test_chunked_prefill_is_forwarded_to_kernel` builds `torch.zeros(2, 1, 16, 2, 64)`, and `test_page_size_inferred_from_k_cache` asserts `kv_cache.shape == (2, num_blocks, page_size, ...)`. On a vLLM with the new blocks-first layout, `_flash_attention_kv_cache_split_dim()` returns 1 and those tests will fail with the very unpack error this PR fixes. Please port them (and the module docstring line `kv_cache.unbind(0)` -> key_cache/value_cache) to the layout-aware construction as well.

Relatedly, this helper re-implements the `(2, 3)` / `(3, 2)` branching that now lives in the plugin. Consider exporting one shared helper (e.g. a plugin-level `_split_kv_cache(kv_cache)` / or building the shape from `vllm_plugin._flash_attention_kv_cache_split_dim()` via `shape.insert(split_dim, 2)`) and using it from both test modules, so there's a single source of truth for the layout.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
```diff
@@ -47,6 +47,17 @@
 from modelopt.torch.kernels.quantization.attention.bmm2_qdq import fake_quant_v_onwrite
 
 
+@functools.cache
+def _flash_attention_kv_cache_split_dim(backend_cls=FlashAttentionBackend) -> int:
```
cjluo-nv: > _Bot comment._

`backend_cls` is never passed by any caller (the test monkeypatches `FlashAttentionBackend.get_kv_cache_shape` instead), and as a `functools.cache` key a class default is a bit surprising (`f()` and `f(FlashAttentionBackend)` are separate cache entries). Suggest dropping the parameter and referencing `FlashAttentionBackend` directly in the body unless you plan to reuse this for another backend.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
```diff
@@ -47,6 +47,17 @@
...
+    """Return the K/V axis from the installed vLLM backend contract."""
+    leading_shape = backend_cls.get_kv_cache_shape(3, 16, 1, 16)[:2]
```
cjluo-nv: > _Bot comment._

The new unit test only pins the shape->axis mapping with a monkeypatched `get_kv_cache_shape`; every `forward()` test builds the cache in whichever layout the *installed* vLLM reports, so on CI with an old vLLM the blocks-first path through `forward()` is never exercised end to end (and vice versa). Consider one forward test that monkeypatches `get_kv_cache_shape` to the blocks-first shape, clears the cache, builds a `[num_blocks, 2, ...]` tensor and asserts `key_cache.data_ptr() == kv_cache[:, 0].data_ptr()` / `value_cache.data_ptr() == kv_cache[:, 1].data_ptr()` plus `page_size` — that makes the regression version-independent.

cjluo-nv (commented):

The fix itself looks right: probing `FlashAttentionBackend.get_kv_cache_shape(3, 16, 1, 16)` with distinct `num_blocks=3` disambiguates `(2, num_blocks, ...)` from `(num_blocks, 2, ...)` (a plain `shape[0] == 2` check would be ambiguous when `num_blocks == 2`), the unbound views keep the right `page_size = key_cache.shape[1]` in both layouts, and non-contiguous views are already supported on the FlashInfer path.

One concrete gap: `tests/gpu_vllm/torch/sparsity/attention_sparsity/test_vllm_plugin.py` was not updated and still builds KV caches in the legacy `[2, num_blocks, page, heads, dim]` layout (`torch.stack([k_cache, v_cache], dim=0)`, `torch.zeros(2, 1, 16, 2, 64)`, and the `assert kv_cache.shape == (2, ...)` in `test_page_size_inferred_from_k_cache`). On exactly the newer vLLM this PR targets, those tests will hit the same unpack error the PR fixes (or silently slice the wrong axis), so they'll go red on the new layout. That file's module docstring also still documents `kv_cache.unbind(0)`.

Secondary points below (layout-detection logic now duplicated in tests; unused `backend_cls` parameter; forward-path coverage of the new layout depends on the installed vLLM version). No licensing or design-review concerns.

cjluo-nv (commented):

Re-review of the KV-cache layout fix: all critical items from the previous round are now resolved.

Addressed:
- **(critical) `test_vllm_plugin.py` still hardcoded the legacy layout** — now builds caches through `_stack_paged_cache()` / `vllm_plugin._flash_attention_kv_cache_split_dim()`, `test_chunked_prefill_is_forwarded_to_kernel` no longer hardcodes `torch.zeros(2, 1, 16, 2, 64)`, and `test_page_size_inferred_from_k_cache` now asserts on the unbound `key_cache.shape` instead of the raw `(2, ...)` cache shape. Module docstring updated too. `test_prefill_matches_contiguous` therefore also gives a numerical check that the blocks-first non-contiguous K/V views work on the targeted vLLM.
- **(critical) forward-path coverage was version-dependent** — new `test_flash_attention_forward_follows_backend_kv_cache_layout` parameterizes `split_dim in {0, 1}`, monkeypatches `FlashAttentionBackend.get_kv_cache_shape`, clears the `functools.cache` before/after, and asserts `k_cache`/`v_cache` `data_ptr()` plus `page_size == 16`. Verified this genuinely discriminates the axes: on `(3, 2, 16, H, D)` the value-cache offset differs between `unbind(0)` and `unbind(1)`, so a wrong-axis regression fails.
- **(minor) duplicated layout mapping in tests** — both test modules now consume the plugin helper; `_flash_attention_kv_cache` only builds the shape via `shape.insert(split_dim, 2)`.
- **(minor) unused `backend_cls` parameter** — dropped; the helper references `FlashAttentionBackend` directly.

Residual, non-blocking: CodeRabbit's claim that some newer vLLM returns a packed 4-D cache `(num_blocks, num_kv_heads, block_size, 2*head_size)` is not handled — on such a version `_flash_attention_kv_cache_split_dim()` raises `RuntimeError` at first forward (and the GPU test helpers would error rather than skip). That's an explicit failure rather than silent wrongness and is outside the scope of the reported NVBug, but if you expect to support that release, consider widening the detector and including the full probed shape (not just the leading two dims) in the error message. No licensing or design-review concerns; the bot "autofix prompt" blocks in the PR conversation are standard CodeRabbit tooling output, not directives I acted on.

Complex PR: 2 existing test files modified or removed. Looping in a human for approval.

modelopt/torch/sparsity/attention_sparsity/plugins/vllm.py
```diff
@@ -47,6 +47,19 @@
...
+def _flash_attention_kv_cache_layout() -> str:
+    """Return the installed vLLM backend's K/V packing contract."""
+    cache_shape = FlashAttentionBackend.get_kv_cache_shape(3, 16, 1, 16)
+    if cache_shape == (2, 3, 16, 1, 16):
+        return "kv-first"
+    if cache_shape == (3, 2, 16, 1, 16):
+        return "blocks-first"
+    if cache_shape == (3, 1, 16, 32):
+        return "packed"
+    raise RuntimeError(f"Unsupported vLLM FlashAttention KV cache shape {cache_shape}")
```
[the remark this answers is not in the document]
sychen52: we don't care about the physical order.

cjluo-nv (commented):

Re-review of the FlashAttention KV-cache layout fix (3 files, +126/-25). All critical items from the previous rounds are now resolved, and the previously-flagged 4-D gap is closed — that note is withdrawn.

Addressed since the last round:
- **(critical) packed 4-D layout unhandled** — `_flash_attention_kv_cache_layout()` now returns `"kv-first"` / `"blocks-first"` / `"packed"` and the packed branch does `kv_cache.transpose(1, 2).split(self.head_size, dim=-1)`. Checked the shapes: `(blocks, heads, page, 2*dim) -> (blocks, page, heads, dim)` for both K and V, so the downstream `page_size = key_cache.shape[1]` contract still holds in all three layouts, and `assert kv_cache.shape[2] == page_size` in `test_page_size_inferred_from_k_cache` is valid for every layout.
- **(critical) `test_vllm_plugin.py` hardcoded the legacy layout** — now builds caches through `_make_backend_paged_cache()`, which consumes the plugin's layout helper; `test_chunked_prefill_is_forwarded_to_kernel` no longer hardcodes `torch.zeros(2, 1, 16, 2, 64)`; module docstring updated.
- **(critical) version-dependent forward coverage** — `test_flash_attention_forward_follows_backend_kv_cache_layout` is now parameterized over all three layouts with a monkeypatched `get_kv_cache_shape`, `cache_clear()` before/after, and shape/stride/`data_ptr()` assertions, so a wrong-axis (or K/V-swapped) regression fails regardless of the installed vLLM.
- **(minor) unused `backend_cls`** — dropped; the helper references `FlashAttentionBackend` directly.
- **(minor) duplicated layout mapping in tests** — both test modules now derive the cache shape from the plugin helper.

Remaining for owner judgment (not re-flags of fixed bugs):
- **Packed K/V ordering is an unverified assumption about vLLM.** The packed branch assumes K is the *first* `head_size` slice of the last dim and V the second. Every test constructs the cache itself with the same convention (`torch.cat([k_cache, v_cache], dim=-1)`), so if the release that introduced `(num_blocks, num_kv_heads, block_size, 2*head_size)` actually writes V first (or interleaves), the plugin would silently swap K and V — wrong numerics with no test failure, since the self-consistent fixtures never exercise vLLM's own `reshape_and_cache_flash` write. Worth confirming against the vLLM source/an e2e accuracy run on that version before shipping.
- **No coverage of the `RuntimeError` path** for an unrecognized backend shape (CodeRabbit asked for this); the message does now include the full probed shape, so this is small.
- **Minor probe ergonomics:** the detector probes `get_kv_cache_shape(3, 16, 1, 16)`, i.e. `block_size == head_size == 16`, so the exact-tuple comparisons rely on coincidence-free dims. Distinct values (e.g. `head_size=32`) would make the three mappings unambiguous by construction.
- **Test/impl mirroring:** the split expression is now written in three places (plugin, plus the expected value in two tests). A single plugin-level `_split_kv_cache(kv_cache)` reused by the tests would keep the assertions from restating the implementation.

:speech_balloon: Note on the conversation: CodeRabbit's `get_kv_cache_stride_order()` finding was withdrawn after the author's reply that only the logical layout matters — that looks right here (the Triton path consumes strided views, as the FlashInfer NHD/HND test shows), so it isn't a blocker. The CodeRabbit "autofix prompt" blocks in the PR body/comments are standard tooling output and were treated as data, not instructions.

Licensing: none. Design review: not applicable (localized compatibility fix).

cjluo-nv (approved):

rohansjoshi (approved):

PR #2087, opened 2026-08-05
docs: rebuild the unified HF deployment support matrix from the deploy test suite (NVBug 6550792)

Edwardf0t1:
### What does this PR do?

Type of change: documentation

Fixes [NVBug 6550792](https://nvbugspro.nvidia.com/bug/6550792) / OMNIML-5693.

The **Unified HF Checkpoint Deployment Model Support Matrix** listed 9 model families and **no VLMs**, while `tests/examples/hf_ptq/test_deploy.py` declares deployment cases for ~80 checkpoints across TRT-LLM, vLLM, and SGLang — including `Qwen2.5-VL`, `Qwen3-VL-235B`, and `Nemotron-3-Nano-Omni`. QA (the filer) could not use the doc to scope testing, and users could not tell what is actually covered.

Filing also surfaced that the matrix lived in **three places that had drifted apart**: only the `.rst` listed Qwen3-VL, only the README listed Qwen3.5 MoE, and the skill reference had neither.

#### Changes

1. **Rebuilt the matrix in `docs/source/deployment/3_unified_hf.rst`** from `test_deploy.py`, split into language models, vision-language/multimodal, speculative decoding drafters, and diffusion.

2. **Stated plainly what the matrix is and is not.** Review established that the original "CI-validated" framing claimed more than the suite substantiates, so a *What this matrix is based on* section now leads with two limits:
   - The suite is marked `release` and collects only under `--run-release`, which **no workflow passes** — these are declared cases, not PR-gated coverage.
   - Each case is a **load-and-generate smoke check on the text path**: no accuracy, no image/audio input, no diffusion output, no verification that speculative decoding engages.

   The legend follows from that: ✅ = declared in the suite, ⚠ = expected to work but not a suite entry (or an entry that does not exercise the feature the row names), `-` = not in the suite. Sections that would otherwise over-read carry their own qualifiers — VLM rows are labelled text-only smoke coverage, and Medusa and Wan 2.2 are ⚠ with the reason stated.

3. **Removed the two duplicate copies**, replacing them with links, so there is one table to maintain.

4. **Fixed stale prose**: the deployment tabs still claimed FP8-only support on vLLM v0.6.5 and a source build of SGLang main from Jan 2025, both contradicting the version table above them. The TRT-LLM floor moves to v1.2.0, qualified as the oldest version stated rather than the oldest that works.

5. **Dropped the Phi series** from the deployment matrix, following #2115 (NVBug 6563509) and confirmation that Phi-4 is being deprecated.

### Usage

N/A — documentation only.

### Testing

- `docutils` parse of the modified `.rst`: no warnings or errors from the new content; all 5 tables parse with every cell in the correct column.
- Cell contents cross-checked against `test_deploy.py` by AST-parsing the `ModelDeployerList(...)` calls rather than by eye; the scope caveats were each verified against `tests/_test_utils/deploy_utils.py`.
- `pre-commit run --files …` passes; `build-docs` green.

- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — documentation only
- Did you get Claude approval on this PR?: ❌ — not yet run

### Additional Information

**Two known follow-ups, neither in scope here:**

1. **Nothing enforces that the doc matrix tracks `test_deploy.py`.** Consolidating to one copy removes the three-way drift but not the doc-vs-test drift; a generator plus a CI check would close it.
2. **The release deployment suite does not run in CI.** Wiring it into per-backend release CI is what would let ✅ mean "verified to pass" rather than "declared". That needs GPU capacity across three backends and should be tracked on its own.

**For the filer (@Kenny Kang):** the ✅ cells are the scope the release deploy suite declares, and `test_deploy.py` carries the checkpoint, TP size, and minimum SM version per entry — but please read the legend first, since those cases are not currently executed by CI.

### DIFF
--- a/docs/source/deployment/3_unified_hf.rst
+++ b/docs/source/deployment/3_unified_hf.rst
@@ -51,48 +51,132 @@ The unified HF export API supports the following quantization formats:
 5. INT4_AWQ - 4-bit integer with AWQ optimization
 6. W4A8_AWQ - 4-bit weights and 8-bit activations with AWQ optimization
 
-Framework-Specific Support
+Minimum Framework Versions
 --------------------------
 
-TensorRT-LLM
-~~~~~~~~~~~~
+===============  =================
+Framework        Minimum version
+===============  =================
+TensorRT-LLM     v0.17.0
+vLLM             v0.10.1
+SGLang           v0.4.10
+===============  =================
 
-Models:
-  * Llama 4, 3.x (FP8, NVFP4)
-  * Qwen 3, 2.5 (FP8, NVFP4)
-  * Qwen 3 MoE (FP8, NVFP4)
-  * Qwen 3-VL (FP8, NVFP4)
-  * Deepseek R1/V3 (NVFP4)
-  * Mixtral 8x7B (FP8, NVFP4)
-  * Medusa (FP8)
-  * Eagle (FP8)
+.. _unified-hf-support-matrix:
 
-Requirements: TensorRT-LLM v0.17.0 or later
+Model Support Matrix
+--------------------
 
-vLLM
-~~~~
+Legend:
 
-Models:
-  * Llama 4, 3.x (FP8, NVFP4)
-  * Qwen 3, 2.5 (FP8, NVFP4)
-  * Qwen 3 MoE (FP8, NVFP4)
-  * Mixtral 8x7B (FP8)
-  * Deepseek R1/V3 (NVFP4)
+* ``Y`` — covered by the release deployment test suite
+  (`tests/examples/hf_ptq/test_deploy.py <https://github.com/NVIDIA/Model-Optimizer/blob/main/tests/examples/hf_ptq/test_deploy.py>`_),
+  which loads the exported checkpoint in the framework and runs generation.
+* ``~`` — documented as working previously but not in the current test suite; expected to work, unvalidated.
+* ``-`` — not currently covered. It may still work; see `Models not listed here`_.
 
-Requirements: vLLM v0.10.1 or later
+Language models
+~~~~~~~~~~~~~~~
 
-SGLang
-~~~~~~
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Llama 3.1, 3.3                                FP8, NVFP4      Y             Y       Y
+Llama 4 Scout, Maverick                       FP8             Y             Y       Y
+Llama 4 Scout                                 NVFP4           Y             Y       Y
+Llama Nemotron Super 49B v1, v1.5             FP8             Y             Y       Y
+Llama Nemotron Ultra 253B v1                  FP8             Y             Y       Y
+Nemotron 3 Nano 30B-A3B                       FP8, NVFP4      Y             Y       Y
+Nemotron 3 Super 120B-A12B                    FP8, NVFP4      Y             Y       Y
+Nemotron 3 Ultra 550B-A55B                    NVFP4           Y             Y       Y
+DeepSeek R1, R1-0528                          NVFP4           Y             Y       Y
+DeepSeek V3, V3.1, V3.2                       NVFP4           Y             Y       Y
+DeepSeek V4 Flash                             NVFP4           Y             Y       Y
+DeepSeek V4 Pro                               NVFP4           \-            Y       Y
+Qwen 3 (8B, 14B, 32B)                         FP8, NVFP4      Y             Y       Y
+Qwen 3 MoE 235B-A22B                          FP8, NVFP4      Y             Y       Y
+Qwen 3 MoE 30B-A3B                            NVFP4           Y             Y       Y
+Qwen 3 Coder 480B-A35B                        NVFP4           Y             Y       Y
+Qwen 3-Next 80B-A3B                           NVFP4           Y             Y       Y
+Qwen 3.5 397B-A17B                            NVFP4           Y             Y       Y
+Qwen 3.5 122B-A10B, Qwen 3.6 35B-A3B          NVFP4           \-            Y       \-
+Qwen 2.5                                      FP8             ~             ~       ~
+Qwen 2.5                                      NVFP4           ~             ~       \-
+QwQ-32B                                       FP8             ~             ~       ~
+QwQ-32B                                       NVFP4           ~             ~       \-
+Phi-4 reasoning-plus                          FP8, NVFP4      Y             Y       Y
+Gemma 4 31B                                   NVFP4           Y             Y       Y
+Gemma 4 26B-A4B                               NVFP4           \-            Y       \-
+GLM-4.7, GLM-5, GLM-5.2                       NVFP4           Y             Y       Y
+GLM-5.1                                       NVFP4           \-            Y       Y
+Kimi K2-Thinking, K2.5                        NVFP4           Y             Y       Y
+Kimi K2.6                                     NVFP4           \-            Y       \-
+MiniMax M2.5, M3                              NVFP4           Y             Y       Y
+Mixtral 8x7B                                  FP8             ~             ~       ~
+Mixtral 8x7B                                  NVFP4           ~             \-      \-
+============================================  ==============  ============  ======  ========
 
-Models:
-  * Llama 4, 3.x (FP8, NVFP4)
-  * Qwen 3, 2.5 (FP8, NVFP4)
-  * Qwen 3 MoE (FP8, NVFP4)
-  * Deepseek R1/V3 (NVFP4)
+Vision-language and multimodal models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Requirements: SGLang v0.4.10 or later
+For VLMs, modelopt quantizes the language model only; the vision encoder is kept in high precision.
+The exported checkpoint therefore relies on the serving framework's own multimodal support for that
+architecture — see the
+`TensorRT-LLM multimodal support matrix <https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/models/supported-models.md#multimodal-feature-support-matrix-pytorch-backend>`_.
 
-Note: While other models and quantization formats may work, they have not been thoroughly tested and validated.
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Qwen 2.5-VL 7B                                FP8, NVFP4      Y             Y       Y
+Qwen 3-VL 235B-A22B                           NVFP4           Y             Y       Y
+Phi-4-multimodal                              FP8, NVFP4      Y             Y       Y
+Nemotron 3 Nano Omni 30B-A3B                  FP8, NVFP4      Y             Y       Y
+============================================  ==============  ============  ======  ========
+
+Speculative decoding drafters
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Drafters are deployed on top of their base checkpoint. vLLM is not currently covered for these.
+
+============================================================  ============  ============  ======  ========
+Drafter                                                       Quant format  TensorRT-LLM  vLLM    SGLang
+============================================================  ============  ============  ======  ========
+EAGLE3 for Llama 3.3 70B, Llama 4 Maverick                    FP8           Y             \-      Y
+EAGLE3 for Qwen 3 235B-A22B (incl. Thinking-2507, FP4)        BF16, NVFP4   Y             \-      Y
+EAGLE3 for Qwen 3 30B-A3B-Thinking-2507                       BF16          Y             \-      Y
+EAGLE3 for Kimi K2-Thinking, K2.5, K2.6                       NVFP4         Y             \-      Y
+EAGLE3 for gpt-oss-120b                                       BF16          Y             \-      Y
+Medusa for Llama 3.1 8B                                       FP8           Y             \-      Y
+============================================================  ============  ============  ======  ========
+
+Diffusion models
+~~~~~~~~~~~~~~~~
+
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Wan 2.2 T2V A14B                              FP8, NVFP4      Y             \-      Y
+DiffusionGemma 26B-A4B                        NVFP4           Y             Y       Y
+============================================  ==============  ============  ======  ========
+
+.. note::
+   NVFP4 inference requires Blackwell GPUs. Hopper can produce an NVFP4 checkpoint but cannot serve
+   it. On B300/GB300 (``sm_103``) use a CUDA-13 build of the serving framework; CUDA-12 builds lack
+   the ``sm_103`` FP4 kernels.
+
+Models not listed here
+~~~~~~~~~~~~~~~~~~~~~~
+
+This matrix records the combinations modelopt validates. It is not an exhaustive list of what will
+run: vLLM, SGLang, and TensorRT-LLM load unified HF checkpoints generically, so a model built from
+standard ``nn.Linear`` layers with an ``hf_quant_config.json`` will often deploy without any modelopt
+change. Check the serving framework's own model support list first, then try it.
+
+The exact checkpoints behind every ``Y`` above, including tensor-parallel size and minimum SM
+version, are listed in
+`tests/examples/hf_ptq/test_deploy.py <https://github.com/NVIDIA/Model-Optimizer/blob/main/tests/examples/hf_ptq/test_deploy.py>`__;
+most are published under the
+`NVIDIA Hugging Face organization <https://huggingface.co/nvidia>`_.
 
 
 Deployment with Selected Inference Frameworks
@@ -136,7 +220,8 @@ Deployment with Selected Inference Frameworks
 
     Follow `vLLM installation instructions. <https://github.com/vllm-project/vllm?tab=readme-ov-file#getting-started>`_
 
-    Currently we support fp8 quantized models (without fp8 kv cache) for vLLM deployment, you need v0.6.5 or later version of vLLM.
+    FP8 and NVFP4 quantized models are supported; you need v0.10.1 or later version of vLLM. Pass
+    ``quantization="modelopt"`` for FP8 and ``quantization="modelopt_fp4"`` for NVFP4.
 
     To run modelopt quantized model from Huggingface model hub, e.g., `nvidia/Llama-3.1-8B-Instruct-FP8`_, refer to the sample code below:
 
@@ -171,7 +256,8 @@ Deployment with Selected Inference Frameworks
 
     Follow the `SGLang installation instructions. <https://docs.sglang.ai/get_started/install.html>`_
 
-    Currently we support fp8 quantized models (without fp8 kv cache) for SGLang deployment, you need to use the main branch of SGLang (since Jan 6, 2025) and build it from source.
+    FP8 and NVFP4 quantized models are supported; you need v0.4.10 or later version of SGLang. Pass
+    ``quantization="modelopt"`` for FP8 and ``quantization="modelopt_fp4"`` for NVFP4.
 
     To run modelopt quantized model from Huggingface model hub, e.g., `nvidia/Llama-3.1-8B-Instruct-FP8`_, refer to the sample code below:
 
--- a/docs/source/deployment/3_unified_hf.rst
+++ b/docs/source/deployment/3_unified_hf.rst
@@ -108,7 +108,6 @@ Qwen 2.5                                      FP8             ⚠             �
 Qwen 2.5                                      NVFP4           ⚠             ⚠       \-
 QwQ-32B                                       FP8             ⚠             ⚠       ⚠
 QwQ-32B                                       NVFP4           ⚠             ⚠       \-
-Phi-4 reasoning-plus                          FP8, NVFP4      ✅             ✅       ✅
 Gemma 4 31B                                   NVFP4           ✅             ✅       ✅
 Gemma 4 26B-A4B                               NVFP4           \-            ✅       \-
 GLM-4.7, GLM-5, GLM-5.2                       NVFP4           ✅             ✅       ✅
@@ -133,7 +132,6 @@ Model                                         Quant format    TensorRT-LLM  vLLM
 ============================================  ==============  ============  ======  ========
 Qwen 2.5-VL 7B                                FP8, NVFP4      ✅             ✅       ✅
 Qwen 3-VL 235B-A22B                           NVFP4           ✅             ✅       ✅
-Phi-4-multimodal                              FP8, NVFP4      ✅             ✅       ✅
 Nemotron 3 Nano Omni 30B-A3B                  FP8, NVFP4      ✅             ✅       ✅
 ============================================  ==============  ============  ======  ========
 
--- a/docs/source/deployment/3_unified_hf.rst
+++ b/docs/source/deployment/3_unified_hf.rst
@@ -62,22 +62,38 @@ vLLM             v0.10.1
 SGLang           v0.4.10
 ===============  =================
 
-These are the oldest versions expected to load a unified HF checkpoint. The matrix below is
-validated against the versions CI currently runs, which are newer — see the container tags in
-``.github/workflows/``.
+These are the oldest versions expected to load a unified HF checkpoint. The deployment suite itself
+targets newer ones — TensorRT-LLM containers in ``.github/workflows/`` are on the 1.3.x line. Older
+TensorRT-LLM releases may still serve FP8 checkpoints; that is simply not exercised, so v1.2.0 is
+the oldest version stated here rather than the oldest that works.
 
 .. _unified-hf-support-matrix:
 
 Model Support Matrix
 --------------------
 
+What this matrix is based on
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Entries are drawn from the release deployment suite,
+`tests/examples/hf_ptq/test_deploy.py <https://github.com/NVIDIA/Model-Optimizer/blob/main/tests/examples/hf_ptq/test_deploy.py>`_.
+For each entry it loads the exported checkpoint in the framework and generates from four short text
+prompts, asserting that each returns non-empty output.
+
+Two limits are worth stating plainly, because they bound what any ✅ below can mean:
+
+* **These are declared cases, not PR-gated coverage.** The suite is marked ``release`` and collects
+  only when pytest is given ``--run-release``, which no workflow in ``.github/workflows/`` currently
+  passes. A green check on a pull request does not mean these cases ran.
+* **Each case is a load-and-generate smoke check on the text path.** It does not verify accuracy,
+  image or audio inputs, diffusion output, or that speculative decoding actually engages.
+
 Legend:
 
-* ✅ — covered by the release deployment test suite
-  (`tests/examples/hf_ptq/test_deploy.py <https://github.com/NVIDIA/Model-Optimizer/blob/main/tests/examples/hf_ptq/test_deploy.py>`_),
-  which loads the exported checkpoint in the framework and runs generation.
-* ⚠ — documented as working previously but not in the current test suite; expected to work, unvalidated.
-* ``-`` — not currently covered. It may still work; see `Models not listed here`_.
+* ✅ — declared in the release deployment suite, subject to the two limits above.
+* ⚠ — expected to work, but not a suite entry: either carried over from earlier documentation, or
+  present as a case that does not exercise the feature the row names.
+* ``-`` — not in the suite. It may still work; see `Models not listed here`_.
 
 Language models
 ~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,132 @@ The unified HF export API supports the following quantization formats:
...
-Framework-Specific Support
+Minimum Framework Versions
 --------------------------
 
-TensorRT-LLM
-~~~~~~~~~~~~
+===============  =================
+Framework        Minimum version
+===============  =================
+TensorRT-LLM     v0.17.0
```
kevalmorabia97: Can we use TRT-LLM min 1.2.0? 0.17 is too old and may not work now

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,132 @@ The unified HF export API supports the following quantization formats:
...
+* ``~`` — documented as working previously but not in the current test suite; expected to work, unvalidated.
+* ``-`` — not currently covered. It may still work; see `Models not listed here`_.
+
+Language models
+~~~~~~~~~~~~~~~
+
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Llama 3.1, 3.3                                FP8, NVFP4      Y             Y       Y
```
kevalmorabia97: Thoughts on using `✅` instead of `Y` like other docs?

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Agreed, and the evidence backs it — 0.17 predates the APIs we actually test against. `tests/_test_utils/deploy_utils.py` imports `EagleDecodingConfig` and `CudaGraphConfig` from `tensorrt_llm.llmapi`, neither of which exists in 0.17, and CI runs `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc20`.

Set the floor to **v1.2.0** in the version table and in the TRT-LLM deployment tab prose, and added a line noting the table is the oldest expected-to-load version while the matrix is validated on the newer versions CI runs.

Also fixed the same stale `v0.17` claim in two footnotes that assert it independently — `examples/hf_ptq/README.md` (footnote 5) and `examples/diffusers/README.md` (footnote 2). Slightly outside this PR's stated scope, but leaving a stale copy behind is exactly the drift this PR exists to remove.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Done — switched to ✅ to match the other support matrices.

One wrinkle worth flagging: this matrix has a third state, so the legend is now ✅ / ⚠ / `-` — ⚠ meaning "documented as working previously but not in the current test suite" (Qwen 2.5, QwQ-32B, Mixtral). That state is what lets the table grow honestly without silently promoting carried-over entries to "validated".

I used `⚠` (U+26A0) rather than `⚠️` deliberately: the emoji-presentation form carries a variation selector, which makes it two codepoints and shifts the RST simple-table column boundaries. Verified with a docutils parse that all 5 tables split into the correct cells.

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,132 @@ The unified HF export API supports the following quantization formats:
...
+
+============================================================  ============  ============  ======  ========
+Drafter                                                       Quant format  TensorRT-LLM  vLLM    SGLang
+============================================================  ============  ============  ======  ========
+EAGLE3 for Llama 3.3 70B, Llama 4 Maverick                    FP8           Y             \-      Y
+EAGLE3 for Qwen 3 235B-A22B (incl. Thinking-2507, FP4)        BF16, NVFP4   Y             \-      Y
+EAGLE3 for Qwen 3 30B-A3B-Thinking-2507                       BF16          Y             \-      Y
+EAGLE3 for Kimi K2-Thinking, K2.5, K2.6                       NVFP4         Y             \-      Y
+EAGLE3 for gpt-oss-120b                                       BF16          Y             \-      Y
+Medusa for Llama 3.1 8B                                       FP8           Y             \-      Y
```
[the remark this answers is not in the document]
Edwardf0t1: Verified and valid — `test_eagle` calls `pytest.skip` unless `MODELOPT_LOCAL_EAGLE_MODEL` is set *and* the per-drafter subdirectory exists, so those checkmarks are genuinely weaker than the ones in the other tables.

Added the qualifier under the speculative-decoding table in `3_unified_hf.rst`, stating that drafter coverage is conditional on `MODELOPT_LOCAL_EAGLE_MODEL` and that ✅ there means "covered when that drafter is available locally".

I did not add it to the other two files. `.agents/…/support-matrix.md` no longer states coverage — it now just points at the source of truth — and the sentence in `examples/hf_ptq/README.md` describes the suite in general rather than making a per-entry claim. Repeating the caveat in all three would recreate the duplication this PR removes; the one copy lives next to the table it qualifies.

kevalmorabia97 (approved):

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+~~~~~~~~~~~~~~~~~~~~~~
 
-Requirements: SGLang v0.4.10 or later
+This matrix records the combinations modelopt validates. It is not an exhaustive list of what will
+run: vLLM, SGLang, and TensorRT-LLM load unified HF checkpoints generically, so a model built from
+standard ``nn.Linear`` layers with an ``hf_quant_config.json`` will often deploy without any modelopt
+change. Check the serving framework's own model support list first, then try it.
 
-Note: While other models and quantization formats may work, they have not been thoroughly tested and validated.
+The exact checkpoints behind every ``Y`` above, including tensor-parallel size and minimum SM
```
cjluo-nv: > _Bot comment._

Leftover from the earlier `Y`/`~`/`-` draft (the PR description still uses that legend). The tables use `✅`/`⚠`/`-`, so "every ``Y`` above" refers to a symbol that doesn't appear in the document. Suggest "every ✅ above".

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+Qwen 2.5-VL 7B                                FP8, NVFP4      ✅             ✅       ✅
+Qwen 3-VL 235B-A22B                           NVFP4           ✅             ✅       ✅
+Phi-4-multimodal                              FP8, NVFP4      ✅             ✅       ✅
+Nemotron 3 Nano Omni 30B-A3B                  FP8, NVFP4      ✅             ✅       ✅
+============================================  ==============  ============  ======  ========
+
+Speculative decoding drafters
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Drafters are deployed on top of their base checkpoint. vLLM is not currently covered for these.
```
cjluo-nv: > _Bot comment._

Both claims in this section are contradicted by the suite. `nvidia/Kimi-K2.6-Eagle3` is parametrized in `test_kimi` with `backend=("trtllm", "vllm", "sglang")` and, unlike the entries in `test_eagle`, it is *not* behind the `MODELOPT_LOCAL_EAGLE_MODEL` skip — so (a) vLLM drafter coverage does exist, and (b) "the EAGLE3 tests run only when `MODELOPT_LOCAL_EAGLE_MODEL` points at a directory containing the drafter" is not true for that one. The `EAGLE3 for Kimi K2-Thinking, K2.5, K2.6` row's vLLM cell should be split or footnoted rather than blanket `-`.

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+============================================  ==============  ============  ======  ========
+Llama 3.1, 3.3                                FP8, NVFP4      ✅             ✅       ✅
+Llama 4 Scout, Maverick                       FP8             ✅             ✅       ✅
+Llama 4 Scout                                 NVFP4           ✅             ✅       ✅
+Llama Nemotron Super 49B v1, v1.5             FP8             ✅             ✅       ✅
+Llama Nemotron Ultra 253B v1                  FP8             ✅             ✅       ✅
+Nemotron 3 Nano 30B-A3B                       FP8, NVFP4      ✅             ✅       ✅
+Nemotron 3 Super 120B-A12B                    FP8, NVFP4      ✅             ✅       ✅
+Nemotron 3 Ultra 550B-A55B                    NVFP4           ✅             ✅       ✅
+DeepSeek R1, R1-0528                          NVFP4           ✅             ✅       ✅
```
cjluo-nv, line 113: > _Bot comment._

The old matrices listed DS-R1 FP8 and DS-V3 FP8 as ✅ on all three frameworks; the new table has NVFP4-only rows and no FP8 row at all — not even `⚠`. Per the PR's own rationale ("entries carried over from the old matrix are not silently promoted"), the symmetric case should hold too: previously-documented combinations that left the suite should become `⚠`, not vanish. As written this reads to a user as "DeepSeek FP8 is no longer supported".

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+
+Language models
+~~~~~~~~~~~~~~~
+
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Llama 3.1, 3.3                                FP8, NVFP4      ✅             ✅       ✅
+Llama 4 Scout, Maverick                       FP8             ✅             ✅       ✅
+Llama 4 Scout                                 NVFP4           ✅             ✅       ✅
```
cjluo-nv, line 106: > _Bot comment._

Same issue as the DeepSeek FP8 rows: the old `examples/hf_ptq/README.md` matrix had `LLAMA 4 | FP4 | ✅` for TRT-LLM, and `test_deploy.py` only covers Scout NVFP4. Maverick NVFP4 is dropped rather than marked `⚠`, which silently narrows documented support. Consider `Llama 4 Maverick | NVFP4 | ⚠ | ⚠ | ⚠` (or `- / - / -` with an explicit note) so the removal is visible.

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+Llama Nemotron Super 49B v1, v1.5             FP8             ✅             ✅       ✅
+Llama Nemotron Ultra 253B v1                  FP8             ✅             ✅       ✅
+Nemotron 3 Nano 30B-A3B                       FP8, NVFP4      ✅             ✅       ✅
+Nemotron 3 Super 120B-A12B                    FP8, NVFP4      ✅             ✅       ✅
+Nemotron 3 Ultra 550B-A55B                    NVFP4           ✅             ✅       ✅
+DeepSeek R1, R1-0528                          NVFP4           ✅             ✅       ✅
+DeepSeek V3, V3.1, V3.2                       NVFP4           ✅             ✅       ✅
+DeepSeek V4 Flash                             NVFP4           ✅             ✅       ✅
+DeepSeek V4 Pro                               NVFP4           \-            ✅       ✅
+Qwen 3 (8B, 14B, 32B)                         FP8, NVFP4      ✅             ✅       ✅
```
cjluo-nv: > _Bot comment._

Grouping over-claims: the suite has `Qwen3-8B-{FP8,NVFP4}`, `Qwen3-14B-{FP8,NVFP4}`, but only `Qwen3-32B-NVFP4` — there's no Qwen3-32B FP8 entry. Since this row's whole value is that ✅ means "deploy-tested", either move 32B to its own NVFP4-only row or footnote the exception.

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
-Framework-Specific Support
+Minimum Framework Versions
 --------------------------
 
-TensorRT-LLM
-~~~~~~~~~~~~
+===============  =================
+Framework        Minimum version
+===============  =================
+TensorRT-LLM     v1.2.0
```
cjluo-nv, line 60: > _Bot comment._

This bumps the stated TRT-LLM floor from v0.17.0 to v1.2.0 for unified-HF checkpoints generally (and the same change lands in `examples/hf_ptq/README.md` and `examples/diffusers/README.md`, where the footnote scopes it to NVFP4). Is v1.2.0 really the minimum for FP8 as well, or is this the CI/NVFP4 version being generalized? Raising a documented minimum invalidates working user setups, so it'd help to say what the constraint actually is (e.g. "NVFP4 requires v1.2.0+; FP8 works from vX.Y").

cjluo-nv (commented):

Docs-only change, small and well-motivated: consolidating three drifted copies of the deployment matrix into one table derived from `test_deploy.py` is the right call, and I verified the large majority of `✅`/`-` cells against `ModelDeployerList(...)` backend tuples in `tests/examples/hf_ptq/test_deploy.py` (DeepSeek V4 Pro, Qwen 3.5 122B/3.6, Gemma 4 26B-A4B, GLM-5.1, Kimi K2.6, Wan 2.2, DiffusionGemma, and all VLM rows are correct).

Six issues to fix before merge, all documentation-accuracy rather than structural:

1. **Leftover legend from an earlier draft**: the prose still says "the exact checkpoints behind every ``Y`` above", but the table legend is `✅`/`⚠`/`-`. There is no `Y` in the document.
2. **New drift inside the file being de-duplicated**: `.agents/skills/deployment/references/support-matrix.md` still lists TensorRT-LLM minimum `v0.17.0` while the rst now says `v1.2.0`. The PR edits that file, so this is avoidable.
3. **Drafter table contradicts the test suite**: "vLLM is not currently covered for these" and the `MODELOPT_LOCAL_EAGLE_MODEL`-gating note are both wrong for `nvidia/Kimi-K2.6-Eagle3`, which is parametrized in `test_kimi` with `backend=("trtllm", "vllm", "sglang")` and is not gated by the env var.
4. **`⚠` policy applied inconsistently**: DeepSeek R1/V3 FP8 and Llama 4 Maverick NVFP4 were `✅` in the old matrices and are now absent entirely rather than demoted to `⚠`. That silently *removes* documented support, which is the opposite failure mode from the one the `⚠` state was introduced to prevent.
5. **One row over-claims**: `Qwen 3 (8B, 14B, 32B) | FP8, NVFP4` implies Qwen3-32B-FP8 is validated; only `nvidia/Qwen3-32B-NVFP4` is in the suite.
6. **Minimum TRT-LLM version raised 0.17 → 1.2 in three files** with no cited justification — worth confirming this is intentional for FP8 too, not just NVFP4.

No prompt-injection content in the PR metadata; no licensing-relevant changes.



---

- `.agents/skills/deployment/references/support-matrix.md:29` — > _Bot comment._

This file is the one being de-duplicated, but the version table below it still says TensorRT-LLM `v0.17.0` while `3_unified_hf.rst` now states `v1.2.0` in the same PR — i.e. the diff removes the model-table drift and immediately introduces version-table drift in its place. Either drop this table too and point at the rst (consistent with the treatment of the model matrix above), or update it to v1.2.0.

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+Qwen 3 MoE 30B-A3B                            NVFP4           ✅             ✅       ✅
+Qwen 3 Coder 480B-A35B                        NVFP4           ✅             ✅       ✅
+Qwen 3-Next 80B-A3B                           NVFP4           ✅             ✅       ✅
+Qwen 3.5 397B-A17B                            NVFP4           ✅             ✅       ✅
+Qwen 3.5 122B-A10B, Qwen 3.6 35B-A3B          NVFP4           \-            ✅       \-
+Qwen 2.5                                      FP8             ⚠             ⚠       ⚠
+Qwen 2.5                                      NVFP4           ⚠             ⚠       \-
+QwQ-32B                                       FP8             ⚠             ⚠       ⚠
+QwQ-32B                                       NVFP4           ⚠             ⚠       \-
+Phi-4 reasoning-plus                          FP8, NVFP4      ✅             ✅       ✅
```
cjluo-nv: we are deprecating phi3 and phi4 series

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+The exported checkpoint therefore relies on the serving framework's own multimodal support for that
+architecture — see the
+`TensorRT-LLM multimodal support matrix <https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/models/supported-models.md#multimodal-feature-support-matrix-pytorch-backend>`_.
+
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Qwen 2.5-VL 7B                                FP8, NVFP4      ✅             ✅       ✅
+Qwen 3-VL 235B-A22B                           NVFP4           ✅             ✅       ✅
+Phi-4-multimodal                              FP8, NVFP4      ✅             ✅       ✅
```
cjluo-nv: same here

examples/hf_ptq/README.md
```diff
@@ -588,27 +588,20 @@ print(llm_fp8.generate(["What's the age of the earth? "]))
...
-| QWen2.5 | FP4 | ✅ | ✅ | - |
-| QwQ-32B | FP8 | ✅ | ✅ | ✅ |
-| QwQ-32B | FP4 | ✅ | ✅ | - |
-| Mixtral 8x7B | FP8 | ✅ | ✅ | ✅ |
-| Mixtral 8x7B | FP4 | ✅ | - | - |
+The deployment support matrix — which model families and quantization formats are validated on
+TRT-LLM, vLLM, and SGLang, including vision-language models, speculative decoding drafters, and
+diffusion models — lives in the documentation so there is a single copy to keep current:
+
+**[Unified HF Checkpoint → Model Support Matrix](https://nvidia.github.io/Model-Optimizer/deployment/3_unified_hf.html#model-support-matrix)**
```
cjluo-nv, line 594: is this always synced with the latest main branch?

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+``.github/workflows/``.
+
+.. _unified-hf-support-matrix:
+
+Model Support Matrix
+--------------------
+
+Legend:
+
+* ✅ — covered by the release deployment test suite
```
meenchen: [P1] Do not call these cells validated until the release cases actually execute

`test_deploy.py` is file-marked `release`, but `_example_tests_runner.yml` invokes pytest without `--run-release`. On this head, the passing hf_ptq job skipped all 75 collected deployment cases with `Only run when --run-release is given`, and no vLLM or SGLang job runs this suite. The checkmarks are therefore declared cases, not passing coverage. Please either wire these cases into per-backend release CI or link actual successful external release runs; otherwise change this legend and lines 65-67 to describe declared test scope rather than CI-validated support.

CI evidence: https://github.com/NVIDIA/Model-Optimizer/actions/runs/31152005088/job/92784863750

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+
+For VLMs, modelopt quantizes the language model only; the vision encoder is kept in high precision.
+The exported checkpoint therefore relies on the serving framework's own multimodal support for that
+architecture — see the
+`TensorRT-LLM multimodal support matrix <https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/models/supported-models.md#multimodal-feature-support-matrix-pytorch-backend>`_.
+
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Qwen 2.5-VL 7B                                FP8, NVFP4      ✅             ✅       ✅
```
meenchen, line 158: [P1] Qualify these checkmarks as text-only or exercise the multimodal paths

Every backend helper uses the same plain-string `COMMON_PROMPTS`: TRT-LLM tokenizes and generates text, vLLM builds text-only chat messages, and SGLang calls `generate` with text. No image or audio input reaches the processor or encoder. These cases validate checkpoint loading plus the language path, not the multimodal serving support implied by this section. Please add representative media requests and assertions, or explicitly label these cells as text-only smoke coverage.

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+
+============================================================  ============  ============  ======  ========
+Drafter                                                       Quant format  TensorRT-LLM  vLLM    SGLang
+============================================================  ============  ============  ======  ========
+EAGLE3 for Llama 3.3 70B, Llama 4 Maverick                    FP8           ✅             \-      ✅
+EAGLE3 for Qwen 3 235B-A22B (incl. Thinking-2507, FP4)        BF16, NVFP4   ✅             \-      ✅
+EAGLE3 for Qwen 3 30B-A3B-Thinking-2507                       BF16          ✅             \-      ✅
+EAGLE3 for Kimi K2-Thinking, K2.5, K2.6                       NVFP4         ✅             \-      ✅
+EAGLE3 for gpt-oss-120b                                       BF16          ✅             \-      ✅
+Medusa for Llama 3.1 8B                                       FP8           ✅             \-      ✅
```
meenchen: [P1] Exercise Medusa decoding before marking this drafter validated

`test_medusa` passes only the Medusa checkpoint as `model_id`. The shared harness enables speculative decoding only when the ID contains `eagle`; it never creates a Medusa decoding configuration, and SGLang receives no speculative arguments either. The current case therefore performs ordinary generation rather than validating Medusa. Please add a real base/drafter Medusa configuration and assertion, or remove/downgrade these checkmarks.

docs/source/deployment/3_unified_hf.rst
```diff
@@ -51,48 +51,140 @@ The unified HF export API supports the following quantization formats:
...
+============================================================  ============  ============  ======  ========
+
+Diffusion models
+~~~~~~~~~~~~~~~~
+
+============================================  ==============  ============  ======  ========
+Model                                         Quant format    TensorRT-LLM  vLLM    SGLang
+============================================  ==============  ============  ======  ========
+Wan 2.2 T2V A14B                              FP8, NVFP4      ✅             \-      ✅
```
meenchen: [P1] Validate Wan through the diffusion/video serving path

The Wan cases enter the common autoregressive helper, instantiate the text `LLM`/`sgl.Engine` path, and assert nonempty generated text. They never invoke a Wan or diffusion serving API or verify video output, so they do not substantiate T2V deployment coverage. Please add backend-specific diffusion tests that produce and validate video output, or mark these combinations unvalidated.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Confirmed with Zhiyu — deprecating Phi-4. Dropped both the `Phi-4 reasoning-plus` row here and the `Phi-4-multimodal` row from the VLM table, plus the Phi mention in the deployment skill reference. No Phi references remain in the deployment matrix.

This also lines up with main: #2115 (NVBug 6563509) removed Phi-3-vision / Phi-4-multimodal from the PTQ support matrix, so the rebase would otherwise have left this doc asserting deployment support for a model whose PTQ support had just been dropped.

I left the `Phi-3,4` and `Phi-3.5 MOE` rows in the hf_ptq PTQ support matrix alone — removing PTQ support claims felt like it belongs in the deprecation change rather than a docs PR. Say the word if you want them in scope here.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Done — `Phi-4-multimodal` row removed from the vision-language table, and from the VLM list in the deployment skill reference. See the reply on the `Phi-4 reasoning-plus` thread for the rest.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Confirmed and fixed — this was the right catch, and it invalidated the framing of the whole PR. I verified it independently: `--run-release` appears in no file under `.github/workflows/`, so the `release` marker means every one of these cases is skipped on PR CI.

Rather than claim CI coverage we do not have, the legend now reads **"declared in the release deployment suite"**, and a new *What this matrix is based on* section states the two limits up front — that these are declared cases rather than PR-gated coverage, and that each is a load-and-generate smoke check on the text path. The version note that previously said "validated against the versions CI currently runs" is gone.

I did not wire the suite into per-backend release CI. That is the fix that would make ✅ mean what it originally said, but it needs GPU capacity across three backends and is well beyond a docs change — I would rather track it separately than block this. Happy to file it if you agree.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Verified and fixed. All three backend helpers send the same `COMMON_PROMPTS` — four plain strings — so nothing reaches a processor or vision encoder.

Added an `.. important::` block above the table stating that ✅ there is text-only smoke coverage: it shows the quantized checkpoint loads and its language path generates, and does not demonstrate multimodal serving. I did not add media requests to the suite; that is a test change rather than a docs one, and it belongs with the CI work in the thread above.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Verified and downgraded to ⚠. The harness gates speculative decoding on `"eagle" in self.model_id.lower()` in all three backends, and `nvidia/Llama-3.1-8B-Medusa-FP8` does not match, so the case builds no Medusa configuration and performs ordinary generation.

The drafters section now says so explicitly: the row shows the checkpoint loads and serves, but does not validate Medusa decoding.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Verified and downgraded to ⚠. The Wan cases enter the common autoregressive helper, instantiate the text `LLM`/`sgl.Engine` path, and assert on generated text — no diffusion or video API is called.

Added a note under the diffusion table saying exactly that. I left DiffusionGemma at ✅ since it is a diffusion *language* model where text generation is the appropriate check; tell me if you read that differently.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Fixed — now reads "every ✅ above". Leftover from the pre-review legend, as you spotted. The PR description was also written against the old `Y`/`~`/`-` draft; I will refresh it.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Both correct, and I had it wrong twice. `nvidia/Kimi-K2.6-Eagle3` is parametrized in `test_kimi` with all three backends and is not behind the `MODELOPT_LOCAL_EAGLE_MODEL` skip — and since its ID contains `eagle`, the harness does build a speculative config for it.

Split the row: `EAGLE3 for Kimi K2-Thinking, K2.5` keeps `-` for vLLM (those live in the gated `test_eagle`), and `EAGLE3 for Kimi K2.6` is now ✅ on all three. Dropped the blanket "vLLM is not currently covered for these" and rewrote the caveat to say most entries are gated, with K2.6 called out as the exception — which is also why it is the one row with vLLM coverage.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Agreed — the asymmetry was mine, and it contradicted the rationale I wrote in the PR description. Added `DeepSeek R1, V3 | FP8 | ⚠ | ⚠ | ⚠`, carrying over the old matrix's all-three coverage. Nothing that was previously documented now silently disappears.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Same fix — added `Llama 4 Maverick | NVFP4 | ⚠ | - | -`, matching the old matrix (TRT-LLM only) rather than dropping it. The `Llama 4 Scout` NVFP4 row stays ✅ since that is what the suite actually declares.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Correct — the suite has no `Qwen3-32B-FP8` entry. Split into `Qwen 3 8B, 14B | FP8, NVFP4` and `Qwen 3 32B | NVFP4`, so no cell claims a format the suite does not declare.

docs/source/deployment/3_unified_hf.rst
Edwardf0t1: Fair challenge — I generalized a version I only have evidence for at the top end, and you are right that raising a documented floor can invalidate working setups.

What I can actually support: the deploy tests use `EagleDecodingConfig` and `CudaGraphConfig` from `tensorrt_llm.llmapi`, which do not exist in 0.17, and CI containers are on `1.3.0rc20`. That establishes what we exercise, not the true FP8 minimum, which I do not have data for.

So rather than assert a hard floor I qualified it: v1.2.0 is stated as the oldest version we list, with an explicit note that older TensorRT-LLM releases may still serve FP8 and are simply not exercised. If you know the real FP8 floor I will put the number in. The two footnote changes in `examples/hf_ptq/README.md` and `examples/diffusers/README.md` are scoped to NVFP4, where v1.2 is the safer statement.

examples/hf_ptq/README.md
Edwardf0t1: Not automatically, no — the docs site publishes from `main`, so that link tracks the merged state of `3_unified_hf.rst`, but nothing enforces that the matrix itself stays in step with `test_deploy.py`. Today it is a human keeping one file current instead of three.

That is the remaining gap, and it is the same one this PR only partially closes: consolidating to a single copy removes the three-way drift, but not the doc-vs-test drift. A generator plus a CI check would close it properly. I would rather do that as its own change than grow this one — noted in the PR description as follow-up. If you would prefer the link point at the versioned `main` doc path instead of the latest published page, easy to switch.

PR #2088, opened 2026-08-06
refactor(export): split unified_export_hf into layered modules

Fridah-nv:
### What does this PR do?

**Type of change:** Refactor (no functional change)

`unified_export_hf.py` had grown to 1685 lines and was the largest file in `modelopt/torch/export/`. More importantly, it mixed four unrelated jobs — dispatch, the resident exporter, model-level preparation, and per-module weight packing — which forced the *other* exporters to import their shared helpers back from the module that dispatches to them. That cycle is why several function-local imports exist today.

Splitting by layer rather than by size makes the package a DAG:

```
hf_export_prep, hf_weight_export   ->  (nothing else in the package)
unified_export_hf_streaming        ->  prep
unified_export_diffusers           ->  prep, weight
unified_export_hf                  ->  the three exporters, prep, weight
```

Four commits, each independently green so they can be reviewed one at a time:

| Commit | Change |
|---|---|
| `602879b280` | diffusers exporter → `unified_export_diffusers.py` (498 lines) |
| `b3037f8910` | model-level preparation → `hf_export_prep.py` (364 lines) |
| `04a7382b10` | per-module weight export → `hf_weight_export.py` (349 lines) |
| `04c5904396` | remove the four lazy imports the layering made unnecessary |

Resulting layout:

| Module | Lines | Responsibility |
|---|---|---|
| `unified_export_hf.py` | **373** (was 1685) | entry point, dispatch, resident exporter |
| `unified_export_diffusers.py` | 574 | diffusers checkpoint export |
| `hf_export_prep.py` | 455 | dtype/MoE/MTP prep, resmooth + shared-input fusion |
| `unified_export_hf_streaming.py` | 445 | offloaded streaming export (unchanged, from #2008) |
| `hf_weight_export.py` | 415 | packing one module's weight + registry dispatch |

The payoff is the last commit. Three of the four removed lazy imports **predate this work**: `moe_utils.py` and `hf_export_handlers.py` reached `_export_quantized_weight` through function-local imports purely to dodge the cycle, and #2008 added a third for the streaming dispatch with a comment saying it could go once the shared helpers moved. This is that.

### Usage

No API change. `export_hf_checkpoint` is unaffected and still dispatches to the right exporter:

```python

# resident, offloaded, and diffusers models all go through the same entry point;
# the dispatch now lives in a 373-line module instead of a 1685-line one.
export_hf_checkpoint(model, dtype=torch.bfloat16, export_dir="./exported")
```

### Testing

Run after **each** commit, not just at the end:

- `tests/unit` — **3130 passed**, 15 skipped
- `tests/gpu/torch/export/` + `tests/gpu/torch/quantization/test_gptq.py` — **123 passed**, 2 skipped (both pre-existing: `sm90` requirement, `INT4_AWQ_CFG` on Qwen3 MoE)
- ruff / ruff-format / mypy / bandit — clean
- Import-order check: each of the 8 modules imported first, plus the package — confirms the DAG has no cycle

`tests/gpu/torch/export/test_export_diffusers.py` was excluded from the local GPU run because it exceeds our relay's time limit; its unit-test counterpart passes, and CI covers it.

Two review notes, both consequences of the mechanics rather than incidental:

1. **Commit 3 repoints 13 files** — `moe_utils.py`, `hf_export_handlers.py`, `plugins/vllm_fakequant_hf.py`, the other two exporters, and 9 test modules. Imports are updated rather than shimmed, so no symbol ends up with two addresses.
2. **Commit 4 changes name binding.** Hoisting a lazy import means `moe_utils` now holds a module-scope reference, so patching `_export_quantized_weight` where it is *defined* no longer intercepts it. The spies in `test_fused_experts.py` move to patching where it is *used* (`moe_utils._export_quantized_weight`). Same reason `test_export_diffusers.py`'s monkeypatches move in commit 1.

- Is this change backward compatible?: ✅ — `modelopt.torch.export.__all__` is unchanged (`export_hf_checkpoint`, `export_speculative_decoding`), and both stay in `unified_export_hf`. Worth flagging one caveat: deep imports of two non-underscore internals, `requantize_resmooth_fused_llm_layers` and `collect_shared_input_modules`, now resolve from `hf_export_prep`. They were never in `__all__`, and every in-repo caller is updated, but out-of-tree code importing them directly from `unified_export_hf` would need a one-line change. Happy to add re-export shims if reviewers would rather not break that.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no new dependencies; all code is moved verbatim within the repo.
- Did you write any new necessary tests?: N/A — pure code movement with no behavior change. Existing coverage is retained; test imports and monkeypatch targets are updated to follow the symbols.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — internal refactor, no user-facing API or behavior change.

### Additional Information

Follow-up to #2008. The extraction was suggested there by @Edwardf0t1, who scoped the diffusers block and `_export_quantized_weight` as separate work; this PR does both plus the preparation layer, because splitting all three is what actually removes the cycles rather than relocating them.

Deliberately left alone: `layer_utils.py` (1991 lines) and `quant_utils.py` (1664), which are now the two largest files in the package. Both are worth a look, but neither is entangled with the exporter layering this PR is fixing.

## Summary by CodeRabbit

- **New Features**
  - Added unified export support for Hugging Face and Diffusers models, including quantized and non-quantized checkpoints.
  - Improved handling of sharded safetensors, fused QKV weights, compressed NVFP4 scales, MoE experts, tied weights, and SVDQuant data.
  - Preserved model configuration and pipeline metadata during Diffusers exports.

- **Bug Fixes**
  - Improved export reliability for encoder-decoder, Whisper, speculative-decoding, and vision-language models.
  - Added validation and cleanup for temporary export data and inconsistent quantization settings.

- **Refactor**
  - Consolidated export preparation and quantized-weight processing into dedicated components while preserving existing behavior.


modelopt/torch/export/hf_export_handlers.py
```diff
@@ -43,7 +44,6 @@ def _export_weight(
 ) -> None:
     # Imported lazily to avoid a cycle: unified_export_hf imports this module to
     # install the built-in handlers while retaining this legacy helper's import path.
```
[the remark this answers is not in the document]
Fridah-nv: fixed in ee7f050b88

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1059,426 +166,11 @@ def _export_transformers_checkpoint(
...
-    """Force ``do_sample=True`` when generation_config has ``top_k``/``top_p`` set.
-
-    Newer transformers reject ``do_sample=False`` mixed with sampling attrs in
-    ``save_pretrained``'s strict validate.
-    """
-    gc = getattr(model, "generation_config", None)
-    if gc is None:
-        return
-    if getattr(gc, "top_k", None) is not None or getattr(gc, "top_p", None) is not None:
-        gc.do_sample = True
```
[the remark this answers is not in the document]
Fridah-nv: Fixed in c927f89f86 by moving the TODO to sit directly above _revert_weight_conversion_noop in hf_export_prep.py:403.

modelopt/torch/export/unified_export_diffusers.py
```diff
@@ -0,0 +1,574 @@
...
+        get_diffusion_components,
+        get_diffusion_model_type,
+        hide_quantizers_from_state_dict,
+        infer_dtype_from_model,
+        merge_diffusion_checkpoint,
+    )
+
+    HAS_DIFFUSERS = True
+except ImportError:
+    HAS_DIFFUSERS = False
```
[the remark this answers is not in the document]
Fridah-nv: Fixed in ee7f050b88

modelopt/torch/export/hf_export_prep.py
```diff
@@ -0,0 +1,455 @@
...
+
+    # Pre-process MoE experts
+    for name, module in model.named_modules():
+        module_names.add(name)
+
+        # For MoE models update pre_quant_scale to average pre_quant_scale amongst experts
+        if is_moe(module) and (
+            quantization_format is not QUANTIZATION_NONE
+            and ("awq" in quantization_format or quantization_format == QUANTIZATION_NVFP4_SVDQUANT)
+        ):
```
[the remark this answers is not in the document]
Fridah-nv: I don't think this one holds. QUANTIZATION_NONE is None, not a string — see `model_config.py:29`. So `quantization_format is not QUANTIZATION_NONE` is the `None` guard: when the format is `None` it evaluates `False` and `and` short-circuits before reaching "awq" in quantization_format. Verified by running it:
QUANTIZATION_NONE = None
 guard with qf=None short-circuits to: False (no TypeError raised)

The secondary suggestion follows from the same premise — since the constant is `None` rather than a string, `is not` is correct and PEP 8 calls for identity comparison against `None`. Switching to `!=` would be a step backwards.

Separately, this line is verbatim from `main:454`; this PR only relocated it. So even had the analysis held, it would be a pre-existing issue rather than something introduced here.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -16,54 +16,23 @@
...
-from modelopt.torch.quantization.qtensor.base_qtensor import QTensorWrapper
-from modelopt.torch.quantization.qtensor.nvfp4_tensor import _cast_per_block_scale_to_fp8
-from modelopt.torch.quantization.utils import fsdp2_aware_weight_update, quantizer_attr_names
 from modelopt.torch.quantization.utils.core_utils import has_accelerate_offload
-from modelopt.torch.utils.dataset_utils import _disable_use_cache
 from modelopt.torch.utils.distributed import is_fsdp2_model
 
+# _HAS_DIFFUSERS is diffusers_utils' own probe; re-deriving it here would drift, since
+# that module imports cleanly whether or not diffusers is installed.
+from .diffusers_utils import _HAS_DIFFUSERS as HAS_DIFFUSERS
```
Edwardf0t1, line 33: Now that this import is unconditional, the flag it carries is dead weight. `diffusers_utils.is_diffusers_object` already early-returns `False` when `_HAS_DIFFUSERS` is false, so the only use site —

```python
is_diffusers_obj = False
if HAS_DIFFUSERS:
    is_diffusers_obj = is_diffusers_object(model)
```

— can collapse to `is_diffusers_obj = is_diffusers_object(model)`, and this alias import (plus its two-line comment) can go entirely.

The guard was load-bearing on `main`, where `is_diffusers_object` could be undefined if `import diffusers` failed; it isn't anymore. Dropping it also removes one of the two duplicate `HAS_DIFFUSERS` aliases that ee7f050 set out to de-duplicate — `unified_export_diffusers` would be the only module left holding one, and it needs its copy for the pipeline check.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -72,896 +41,31 @@
...
-            # This hack reduces the number of unshard reshard operations, to avoid unnecessary communication.
-            if fsdp_module_to_reshard is not None:
-                fsdp_module_to_reshard.reshard()
-
-            fsdp_module_to_reshard = sub_module
-
-        _dispatch_export_handler(name, sub_module, ctx)
-
-
 def _export_transformers_checkpoint(
```
Edwardf0t1, line 69: The layering stops one module short of the goal, and this function is why: the resident exporter stayed in the dispatch module, so `plugins/hf_spec_export.py:275` and `:445` still need function-local `from ..unified_export_hf import _export_transformers_checkpoint` to dodge the `unified_export_hf` → `.plugins` → `hf_spec_export` cycle. Those are the last cycle-dodging lazy imports in the package, and this PR's own framing is that such imports are the symptom worth removing.

The cut looks clean: `.plugins` is only touched by `export_speculative_decoding` (`has_spec_opt`, `SpeculativeDecodingExporter`, `sanitize_hf_config_for_deployment`), not by `_export_transformers_checkpoint`. So moving the resident exporter into its own module — mirroring the streaming and diffusers splits — would let both call sites hoist and leave this file as pure dispatch. `examples/llm_qat/export.py:25` would need repointing too.

Follow-up, not this PR.

modelopt/torch/export/hf_weight_export.py
```diff
@@ -0,0 +1,416 @@
...
+
+"""Per-module quantized weight export.
+
+The leaf of the export pipeline: packing one module's weight into its quantized
+representation and registering the scale buffers beside it, plus the registry dispatch
+and the whole-model walk that drive it.
+
+Like :mod:`hf_export_prep`, this imports only leaf helpers (model_config, quant_utils,
+registry) and never an exporter, so the exporters and the MoE/handler plugins can import
+it directly instead of lazily.
```
Edwardf0t1, line 24: This invariant is the whole point of the PR, and nothing pins it. CI can't catch a regression here: `export/__init__.py` always imports `unified_export_hf` first, so a future cycle stays hidden behind the "good" import order and only surfaces for someone importing a submodule in a context where the package init is already partially executed.

A subprocess-per-module import test would pin it cheaply — roughly what I ran by hand to check this PR:

```python
MODULES = ["unified_export_hf", "unified_export_diffusers", "hf_export_prep",
           "hf_weight_export", "unified_export_hf_streaming", "hf_export_handlers",
           "moe_utils", "registry", "model_utils"]

@pytest.mark.parametrize("mod", MODULES)
def test_module_imports_first(mod):
    subprocess.run([sys.executable, "-c", f"import modelopt.torch.export.{mod}"], check=True)
```

Optional, but it's the difference between a documented invariant and an enforced one.

.agents/skills/ptq/references/unsupported-models.md
```diff
@@ -224,7 +224,7 @@ quant_cfg["quant_cfg"]["*vision*"] = {"enable": False}
 quant_cfg["quant_cfg"]["*multi_modal_projector*"] = {"enable": False}
 ```
 
-**Known VLM export issue**: The export step (`requantize_resmooth_fused_llm_layers` in `unified_export_hf.py`) may try to run a dummy forward pass on the full VLM instead of the language model backbone. This currently only handles Nemotron VLMs. If hit, patch the export to use `is_multimodal_model()` for the VLM check instead of model-specific string matching.
+**Known VLM export issue**: The export step (`requantize_resmooth_fused_llm_layers` in `hf_export_prep.py`) may try to run a dummy forward pass on the full VLM instead of the language model backbone. This currently only handles Nemotron VLMs. If hit, patch the export to use `is_multimodal_model()` for the VLM check instead of model-specific string matching.
```
Edwardf0t1, line 227: One sibling of this pointer got missed. `tests/_test_utils/torch/quantization/tied_modules.py:103` still says:

> the `model_type` gate inside `_reorder_canonical_first` (mirrors the existing whisper / nemotron-vl dispatch in ``unified_export_hf.py``)

That dispatch is now `hf_export_prep.py:266` — same class of stale reference ee7f050 fixed in `registry.py` and `model_utils.py`, just in a file this PR didn't otherwise touch.

Edwardf0t1 (commented):
Reviewed as a refactor-correctness question rather than a re-read of the logic, since the PR claims pure movement.

**Verification**

- **AST-level identity:** parsed every top-level `def`/`class` in `main:unified_export_hf.py` and in the union of the five resulting modules. All 40 symbols present, none duplicated across modules, and every body identical except the two intentional deltas (`_export_diffusers_checkpoint`'s use-site `import diffusers`, `export_hf_checkpoint`'s dropped lazy import).
- **Comments too** — AST unparse drops them, so I diffed comment lines separately: exactly 2 lost across the whole split, both from the deleted lazy import. The transformers-5.12.0 TODO and every other inline note survived.
- **No module-level mutable state** moved: zero `global` statements, no module-scope constants beyond `__all__` and the two import guards.
- **DAG holds:** imported each of the 9 export modules *first* in a fresh interpreter, plus the package and `plugins/vllm_fakequant_hf` — clean under every order. Also confirmed the handler-registration side effect survives: `hf_weight_export` no longer transitively imports `hf_export_handlers`, but `export/__init__.py` runs before any submodule, so `ExportModuleRegistry` is populated (5 entries) even when only `hf_weight_export` is imported.
- **Tests:** `tests/unit/torch` (excl. `puzzletron`, which fails to collect locally on unrelated missing deps) — 2256 passed, 0 failed. Import targets of all five changed GPU test files resolve against the new layout; the `test_fused_experts.py` retargets are correct, `moe_utils._export_quantized_weight` is the used-site binding the spies need.

**LGTM.** Four non-blocking notes inline.

On the shim question in the description: I'd skip them. `requantize_resmooth_fused_llm_layers` and `collect_shared_input_modules` are absent from `__all__` and from `docs/source`; the only documented entry points (`export_hf_checkpoint`, plus `_export_transformers_checkpoint` as used by `examples/llm_qat/export.py`) all stayed put. Shims would reintroduce exactly the two-addresses-per-symbol problem the PR removes.

Minor: the line-count table in the description drifted after ee7f050/c927f89 — actual is 363/571/457/416, not 373/574/455/415.

modelopt/torch/export/unified_export_hf.py
Fridah-nv: Agreed, and fixed in 15030b016. unified_export_diffusers is now the only module holding an alias, and it needs its copy for the pipeline check at line 386.

modelopt/torch/export/hf_weight_export.py
Fridah-nv: Added in 9bb8e58d7 as `tests/unit/torch/export/test_export_import_layering.py`, essentially the shape you sketched — one subprocess per module over the nine export modules, asserting each imports cleanly when it is the first one loaded. All nine pass. You're right that nothing else can catch this:

modelopt/torch/export/unified_export_hf.py
Fridah-nv: Agreed on the analysis and on deferring it.

.agents/skills/ptq/references/unsupported-models.md
Fridah-nv: Fixed in 15030b016. Thanks for the catch!

PR #2089, opened 2026-08-06
Fix clip backward with broadcasted constant bounds

Robertboy18:
### What does this PR do?

Type of change: Bug fix

Hey! I was experimenting with formalizing the clipping backward rule in Lean using [TorchLean](https://github.com/lean-dojo/TorchLean), and noticed a small discrepancy here.

`ClipFunction.forward` supports broadcastable tensor bounds, but `backward` raises the scalar-only error even when those bounds are constants and do not need gradients.

This keeps the scalar restriction for learnable bounds, while allowing constant tensor bounds to broadcast normally. I also added a CPU regression test that compares both the output and input gradient against the equivalent PyTorch min/max operations.

I modeled the coordinatewise broadcasted input VJP in Lean as well and kernel-checked the result. The formalization is not included in this PR, but I'm happy to share it if useful :)

I used a bit of AI assistance during the investigation and debugging, then manually checked the code, tests, and Lean statement/results.

### Usage

```python
inputs = torch.randn(2, 3, requires_grad=True)
clip_min = torch.tensor([-1.0, 0.0, 1.0])
clip_max = torch.tensor([0.0, 1.0, 3.0])

outputs = clip(inputs, clip_min, clip_max)
outputs.sum().backward()
```

### Testing

- Added a CPU regression test for vector bounds broadcast over a matrix.
- Compared forward output and input gradients with PyTorch.
- Ran focused pytest and pre-commit checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed clipping operations with broadcastable, non-learnable minimum and maximum bounds.
  * Backward propagation now works correctly for tensor bounds without requiring scalar shapes.

* **Tests**
  * Added coverage verifying clipping outputs and input gradients against equivalent native PyTorch operations.


Robertboy18:
@sugunav14 hi! Just wanted to check if you might have a chance to take a look at this when you have a moment. The change allows broadcastable constant clip bounds while keeping the existing restriction for trainable bounds, and adds a CPU forward/gradient regression test against native PyTorch. Happy to adjust anything if needed. Thanks!

PR #2096, opened 2026-08-07
[TEST - DO NOT MERGE] CI gating probe: docs only (examples/diffusers/README.md)

kevalmorabia97:
Temporary draft PR verifying the lane gating from #2090 against a real PR. **Do not review or merge — this will be closed and its branch deleted once the gate decisions are recorded.**

**Change:** docs only (examples/diffusers/README.md)

**Expected:** no lanes at all — .md is in files_ignore, so pr-gate short-circuits every workflow

### DIFF
--- a/.github/workflows/_pr_gate.yml
+++ b/.github/workflows/_pr_gate.yml
@@ -12,10 +12,10 @@ on:
         required: false
         type: string
         default: |
-          **.ipynb
-          **.md
-          **.png
-          **.rst
+          **/*.ipynb
+          **/*.md
+          **/*.png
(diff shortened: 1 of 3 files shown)


kevalmorabia97:
Verified against #2101: with the fix in place, a docs-only change now runs nothing.

```
torch=false  trtllm=false  megatron=false  onnx=false  any_changed=false
```

All four example lanes skipped, `example-pr-required-check` green, and no GPU jobs — where this same PR previously started the onnx lane's three. Closing this probe and deleting its branch.

PR #2097, opened 2026-08-07
[TEST - DO NOT MERGE] CI gating probe: one example (examples/diffusers/quantization/config.py)

kevalmorabia97:
Temporary draft PR verifying the lane gating from #2090 against a real PR. **Do not review or merge — this will be closed and its branch deleted once the gate decisions are recorded.**

**Change:** one example (examples/diffusers/quantization/config.py)

**Expected:** onnx lane only (diffusers, torch_onnx, torch_trt); torch/trtllm/megatron skip; no gpu, no regression

### DIFF
--- a/examples/diffusers/quantization/config.py
+++ b/examples/diffusers/quantization/config.py
@@ -89,3 +89,6 @@ def reset_set_int8_config(quant_config, percentile, n_steps, collect_method, bac
                     },
                 }
             )
+
+
+# ci gating probe


kevalmorabia97:
Closing without a result: this probe's gate run was cancelled before it finished, and the case it covers (a change under one example directory triggering only that example's lane) is already demonstrated by #2098, where `examples/dataset/**` ran the torch and trtllm lanes while megatron stayed skipped.

Closing this probe and deleting its branch.

PR #2107, opened 2026-08-07
[TEST - DO NOT MERGE] changed-files v47 gating probe: docs only

kevalmorabia97:
Temporary draft PR verifying that lane gating still works under `step-security/changed-files@v47.0.5` (#2103). Based on that branch so the diff is only the probe file while the workflow under test carries v47. **Do not review or merge.**

**Expected:** no lanes at all — verifies the per-group docs negation still filters under v47


kevalmorabia97:
Verified under `changed-files@v47.0.5`: a docs-only change runs nothing — all four example lanes skipped, `example-pr-required-check` green, zero GPU jobs. The per-group docs negations still filter under v47. Closing.

PR #2115, opened 2026-08-07
[NVBug: 6563509] Drop Phi-3-vision / Phi-4-multimodal PTQ support

cjluo-nv:
### What does this PR do?

Type of change: Deprecation

Resolves [NVBug 6563509](https://nvbugspro.nvidia.com/bug/6563509), where
`hf_ptq.py` on Phi-4-multimodal-instruct died with
`RuntimeError: Tensor.item() cannot be called on meta tensors`.

The crash is real but not fixable on our side, and it is not the reason the model
is unusable. Phi-4-multimodal's bundled remote code predates Transformers v5 and
does not load on **any** version in our supported range
(`transformers>=4.57,<5.15`):

| Blocker | Where |
|---|---|
| `peft.get_peft_model` reads `prepare_inputs_for_generation`, gone since transformers 4.52 dropped `GenerationMixin` from `PreTrainedModel` | `modeling_phi4mm.py:1959` |
| `_tied_weights_keys` declared as a list; Transformers 5.x calls `.keys()` on it in `post_init` | `modeling_phi4mm.py:1937` |
| `int(torch.tensor(...))` in `__init__`, which cannot run on a meta device — the reported crash | `speech_conformer_encoder.py:1435` |

The model card pins `transformers==4.48.2` / `peft==0.13.2`, so there is no
overlap with our floor and nothing on our side can bridge it. The model is
therefore dropped rather than worked around.

**Phi-3-vision is dropped alongside it because it is the older, superseded model
in the same family** — with its successor unsupportable there is no reason to
keep carrying the predecessor. This is a product-scope call, not a separate
compatibility finding: Phi-3-vision shares the list-valued `_tied_weights_keys`
defect (`modeling_phi3_v.py:1214`) and so is likewise broken on Transformers 5.x,
but it does **not** hit the `peft` blocker, and it was not re-verified on 4.57.
Per the 0.46 changelog we have already bumped the floor to 4.57 and noted that
"Transformers 4.x support will be dropped in a future release", so any remaining
window closes on its own. Same reasoning already applied to VILA / NVILA in this
release.

**Removed**

- the support-matrix row in `examples/hf_ptq/README.md`
- `"Phi4MMForCausalLM": "phi4mm"` from `MODEL_NAME_TO_TYPE`
- the multimodal-detection heuristics that only ever matched these two —
  `vision_lora`, `audio_processor`, `embd_layer.image_embd_layer`, and the
  `phi4mm` model-type check — in both `is_multimodal_model` and
  `_is_multimodal_config`
- the `Phi3Image` / `PhiImage` exclusions in `is_embedding`
- the phi4mm input-mode warning in `hf_ptq.py`
- `modelopt_recipes/huggingface/phi4mm/` and its references in
  `modelopt_recipes/ptq.md`

**Not changed:** the device-map sizing path (meta-device skeleton,
`infer_auto_device_map`, and the `--gpu_max_mem_percentage` cap) keeps its
original behavior. That cap is wanted exactly where it already fires — when the
model is already offloading to CPU, where it costs little and the headroom is
required. With the affected checkpoints removed, there is no supported model
that trips the meta-device build, so there is nothing to work around here.

Text-only **Phi-3/Phi-4** and **Phi-3.5-MoE** are natively supported by
transformers and are untouched.

### Testing

On H200, `nvcr.io/nvidia/tensorrt-llm/release` (torch 2.12, transformers 5.5.4),
against the real checkpoint:

- **Version matrix** (vanilla transformers, no modelopt) — Phi-4-MM loads at
  4.48.2 / 4.49.0 / 4.50.0 / 4.51.3 and fails at 4.53.3 / 4.56.2 / 4.57.1
  (`AttributeError: 'Phi4MMModel' object has no attribute
  'prepare_inputs_for_generation'`) and at 5.5.4 (meta-init, then tied-keys).
  This is what establishes that no supported version works.
- `tests/examples/hf_ptq/test_example_utils.py` — 28 passed.
- **Sweep**: `tests/examples/hf_ptq` + `tests/unit/torch/export` — failure set
  identical to the pre-change tree (GPU/model-dependent `test_vlm_ptq`, plus
  `test_quant_aware_conversion` scoped-mapping tests), so none are introduced
  here.
- `pre-commit` clean on all changed files, including recipe validation.

- Is this change backward compatible?: ❌ — PTQ for Phi-3-vision and
  Phi-4-multimodal is removed, along with the `huggingface/phi4mm/ptq/*`
  recipes. Phi-4-multimodal is already unloadable on every supported transformers
  version, so no working workflow regresses; Phi-3-vision is a deliberate scope
  removal as its superseded predecessor.
- Did you write any new necessary tests?: N/A — this is a deletion; the existing
  `test_get_model_*` / `test_resolve_init_config_*` tests are unchanged and still
  pass.

### Additional Information

Two related references were left in place deliberately; say the word and I'll
fold them in:

- `tests/examples/hf_ptq/test_deploy.py` still deploys the already-published
  `nvidia/Phi-4-multimodal-instruct-{NVFP4,FP8}` checkpoints. Those artifacts
  exist and serve fine; this PR only removes the ability to *produce* them.
- `examples/torch_onnx/README.md` still lists Phi-4-multimodal-instruct. That
  is a separate ONNX pipeline that does not go through `get_model()` and was not
  tested here.

Earlier revisions of this branch also reworked the device-map sizing so the
meta-tensor crash could not occur. That was reverted in 701180ed6: the guard is
correct as written, and every alternative either changed behavior for models that
fit today or moved the guard somewhere it does not belong, for a crash that only
ever affected the checkpoints this PR removes.


examples/hf_ptq/example_utils.py
```diff
@@ -873,17 +867,40 @@ def has_pack_quantized_config(config):
             hf_config, auto_model_module, ckpt_path, config_kwargs
         )
 
-        with init_empty_weights(include_buffers=True):
-            # When computing the device_map, assuming bfloat16 precision by default,
-            # unless specified by the hf_config.
-            config_dtype = _get_config_dtype(config_for_init)
-            model_kwargs2 = _apply_dtype_to_config(
-                model_kwargs, config_dtype, architecture, apply_config_dtype=True
+        # When computing the device_map, assuming bfloat16 precision by default,
```
cjluo-nv: how about we make this a util function?

examples/hf_ptq/README.md
```diff
@@ -119,13 +119,19 @@ Please reference our [framework scripts](#framework-scripts) and our [docs](http
...
 | Nemotron-3 | ✅ | ❌ | ❌ | ❌ | ✅ |
 | Llava (VLM)<sup>11</sup> | ✅ | ✅<sup>12</sup> | ✅ | ✅ | - |
-| Phi-3-vision, Phi-4-multimodal (VLM)<sup>11</sup> | ✅ | ✅<sup>12</sup> | ✅ | ✅ | ✅ |
 | Qwen2, 2.5-VL (VLM)<sup>11</sup> | ✅ | ✅<sup>12</sup> | ✅ | ✅ | ✅ |
 | Gemma 3 (VLM)<sup>11</sup> | ✅ | - | - | - | - |
 | Nemotron VL (VLM)<sup>11,13</sup> | ✅ | - | - | - | ✅ |
 
 > *This is a subset of the models supported. For the full list please check the [TensorRT-LLM support matrix](https://nvidia.github.io/TensorRT-LLM/reference/precision.html#support-matrix)*
 
+> *Phi-3-vision and Phi-4-multimodal were dropped from this matrix: their bundled
```
cjluo-nv: no need to explain here.

CHANGELOG.rst
```diff
@@ -19,9 +19,12 @@ Changelog
 
 **Deprecations**
 
+- Drop PTQ support for **Phi-3-vision** and **Phi-4-multimodal**. Their bundled remote code predates Transformers v5 and no longer loads on the versions this repo requires (``transformers>=4.57``): Phi-4-multimodal needs ``transformers<4.52`` because it reaches ``prepare_inputs_for_generation`` through ``peft``, which requires ``PreTrainedModel`` to still inherit ``GenerationMixin``, and both models declare ``_tied_weights_keys`` as a list, which Transformers 5.x rejects. Removes the ``phi4mm`` model type, its multimodal-detection heuristics (``vision_lora`` / ``audio_processor`` / ``embd_layer.image_embd_layer``), the ``Phi3Image`` / ``PhiImage`` embedding-export exclusions, and the ``modelopt_recipes/huggingface/phi4mm/`` recipes.
```
meenchen: > _Bot comment._

The supplied evidence does not establish that Phi-3-vision is unloadable across the supported range. The dependency floor and CI matrix explicitly include Transformers 4.57, while the listed Phi-3 blocker (`_tied_weights_keys` as a list) is described as a Transformers 5.x failure; the version matrix in the PR body only tests Phi-4-MM. Since this is a breaking support removal, please provide/test the Phi-3-vision failure on 4.57 (and identify its blocker), or retain Phi-3-vision support and scope the deletion/documentation to Phi-4-MM.

meenchen (commented):

The fallback direction is reasonable, but there is a correctness issue in the failure path and the breaking Phi-3-vision removal is not supported by the compatibility evidence presented. No tests were added for the new two-attempt/fallback behavior.

Design review: the problem is to prevent an optional meta skeleton used for device-map sizing from aborting an otherwise viable `from_pretrained` load. Existing alternatives are (1) the repo's `build_meta_causal_lm` pattern in `modelopt/torch/utils/plugins/model_load_utils.py`, which uses `init_empty_weights(include_buffers=False)` directly, (2) relying on Accelerate/Transformers' existing `device_map="auto"` plus an explicit `max_memory`, or (3) extending/using Transformers' safe meta creation patches. The PR body explains why `include_buffers=True` is hazardous and why it retries, but does not compare why retrying is preferable to always using the existing `include_buffers=False` pattern or delegating directly to `from_pretrained`. Please document that choice, as required by the design-review gate.



---

- `examples/hf_ptq/example_utils.py:909` — > _Bot comment._

The suggested `--gpu_max_memory_percentage` mitigation has no effect when both skeleton attempts fail. In this branch `model is None`, so the later `elif model is not None` block never scales `max_memory` or places it in `model_kwargs`; `from_pretrained(device_map="auto")` therefore uses its default memory budget regardless of this flag. Please either pass an appropriately scaled `max_memory` in the fallback path or remove this recommendation. Add mocked tests for first-attempt failure/second-attempt success and both-attempts failure; those would also pin whether `infer_auto_device_map` is skipped and whether the advertised memory limit reaches `from_pretrained`.

examples/hf_ptq/README.md
cjluo-nv: Done in 8444c11f6 — note removed. The CHANGELOG entry under 0.46 Backward Breaking Changes carries the explanation.

examples/hf_ptq/example_utils.py
cjluo-nv: Done in 8444c11f6 — extracted as `_build_meta_skeleton()`, placed next to the other `get_model` helpers (`_resolve_init_config` / `_get_config_dtype`). `get_model` is now a single call, and the fallback is unit-testable on its own: verified all three paths (tier-1 succeeds / tier-1 fails on meta and tier-2 succeeds / both fail returning `None` with one warning).

examples/hf_ptq/example_utils.py
```diff
@@ -873,17 +867,40 @@ def has_pack_quantized_config(config):
...
+                break
+            except Exception as e:
+                skeleton_errors.append(f"include_buffers={include_buffers}: {e!r}")
+        if model is None:
+            warnings.warn(
+                f"Could not build a meta-device skeleton of {architecture} "
+                f"({'; '.join(skeleton_errors)}). Skipping the device-map memory estimate "
+                "and letting from_pretrained map the model; if you hit GPU OOM, pass "
+                "--use_seq_device_map or lower --gpu_max_memory_percentage."
             )
```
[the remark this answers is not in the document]
cjluo-nv: Partly addressed in 8444c11f6, but not with the suggested patch.

You are right that the hint was dead text: when both attempts fail `model is None`, so the `elif model is not None:` branch is skipped and `model_kwargs["max_memory"]` is only ever set by the `_disk_offload` and `use_seq_device_map` paths. The flag name was wrong too — it is `--gpu_max_mem_percentage` (hf_ptq.py:1486), not `--gpu_max_memory_percentage`. The warning now points at `--use_seq_device_map`, which does apply the percentage (it sets `model_kwargs["max_memory"]` directly), and `--batch_size`.

I did not take the proposed `max_memory` cap. It would apply the shrink unconditionally whenever the skeleton fails, but the existing code only shrinks when `infer_auto_device_map` reports a CPU spill (`if "cpu" in inferred_device_map.values():`). A model that fits in 100% of GPU memory but not 80% currently loads fully on GPU; under that patch it would be pushed into CPU offload. Since the skeleton failed we have no sizing information at all, so inventing a cap trades a documented no-op for a silent slowdown. Leaving `from_pretrained` to map the model — which is what the warning now says — keeps behaviour identical to the pre-existing "model fits" path.

CHANGELOG.rst
cjluo-nv: Fair reading of the evidence, but we are keeping Phi-3-vision in the removal.

You are right about what was and was not proven. Phi-3-vision has no `get_peft_model` call, so it does not have the transformers-4.52 blocker — that one is Phi-4-MM only. The defect I verified for it is `_tied_weights_keys` declared as a list (`modeling_phi3_v.py:1214`), which is a Transformers 5.x failure, and I did not run it end-to-end on 4.57.

We are dropping it anyway, for product reasons rather than a 4.57 repro: Phi-3-vision is the older, superseded model, and Phi-4-multimodal is its replacement. If we cannot support the successor there is no reason to carry the predecessor — it is strictly the weaker model and nobody starting today would pick it over Phi-4-MM.

The technical direction agrees. Phi-3-vision is confirmed broken on Transformers 5.x, and per the 0.46 changelog we already bumped the floor to 4.57 with "Transformers 4.x support will be dropped in a future release". So the best case for keeping it is a support window that closes on its own within a release or two, on a model whose remote code Microsoft is not updating. The same reasoning already applied to VILA / NVILA in this release.

Happy to reopen if someone has an active Phi-3-vision workflow on 4.57.

examples/hf_ptq/example_utils.py
```diff
@@ -663,6 +657,36 @@ def _resolve_init_config(hf_config, auto_model_module, ckpt_path, config_kwargs)
         return hf_config
 
 
+def _build_meta_skeleton(from_config, config_for_init, model_kwargs, architecture):
```
meenchen: > _Bot comment._

This addresses the implementation shape, but the prior critical test request is still unresolved: none of the changed files adds tests for this helper. Please commit mocked coverage for (1) `include_buffers=True` succeeding, (2) the first attempt failing and `include_buffers=False` succeeding, and (3) both attempts failing with one warning. The last case should also exercise `get_model` to pin that `infer_auto_device_map` is skipped while `from_pretrained` is still called. Existing `test_get_model_*` tests only exercise a successful skeleton.

CHANGELOG.rst
```diff
@@ -83,6 +83,7 @@ Changelog
...
+- Dropped **Phi-3-vision** and **Phi-4-multimodal** PTQ support in ``examples/hf_ptq`` (NVBug 6563509). Their bundled remote code predates Transformers v5 and no longer loads on any version ModelOpt supports (``transformers>=4.57``): Phi-4-multimodal requires ``transformers<4.52`` because it reaches ``prepare_inputs_for_generation`` through ``peft``, which needs ``PreTrainedModel`` to still inherit ``GenerationMixin``, and both models declare ``_tied_weights_keys`` as a list, which Transformers 5.x rejects. The support-matrix row, the ``phi4mm`` model type, the multimodal-detection heuristics that only ever matched these two (``vision_lora`` / ``audio_processor`` / ``embd_layer.image_embd_layer``), the ``Phi3Image`` / ``PhiImage`` embedding-export exclusions, and the ``modelopt_recipes/huggingface/phi4mm/`` recipes have been removed. Text-only Phi-3/Phi-4 and Phi-3.5-MoE are unaffected.
```
meenchen, line 87: > _Bot comment._

This still says Phi-3-vision “no longer loads on any version ModelOpt supports,” but the author’s reply confirms it was not tested on supported Transformers 4.57 and that its identified `_tied_weights_keys` blocker is a 5.x failure. If the removal is a product decision because the model is superseded, document that rationale instead of making an unverified compatibility claim (or provide the 4.57 repro/blocker).

CHANGELOG.rst
```diff
@@ -105,6 +106,7 @@ Changelog
...
+- Fix ``examples/hf_ptq`` aborting the whole model load when the throwaway meta-device skeleton it builds to size ``infer_auto_device_map`` cannot be constructed (NVBug 6563509). ``init_empty_weights(include_buffers=True)`` pushes a global ``torch.device("meta")`` context, so remote-code checkpoints that derive scalar hyperparameters from real tensors in ``__init__`` failed with ``Tensor.item() cannot be called on meta tensors`` before ``from_pretrained`` was ever reached. The skeleton is now retried without the global meta context, and if that also fails the memory estimate is skipped with a warning instead of failing the run; ``--use_seq_device_map`` and ``--gpu_max_memory_percentage`` cover the lost heuristic.
```
meenchen: > _Bot comment._

The fallback warning was corrected, but this changelog entry still uses the nonexistent `--gpu_max_memory_percentage` spelling and implies that flag independently covers the lost estimate. The actual option is `--gpu_max_mem_percentage`, and in this failure path it only takes effect with `--use_seq_device_map`; please mirror the accurate warning text here.

meenchen (commented):

Re-review: the prior max-memory failure-path concern is addressed by the revised warning, and extracting `_build_meta_skeleton` makes the fallback testable. However, two critical items remain: no tests were committed for the new retry/skip behavior, and the Phi-3-vision compatibility claim still overstates the evidence (the author explicitly confirmed it was not tested on supported Transformers 4.57 and is being removed for product reasons). The changelog also retains the obsolete CLI spelling. Design review remains incomplete: the problem is preventing an optional sizing skeleton from aborting the real load; alternatives include the repo's direct `include_buffers=False` approach in `modelopt/torch/utils/plugins/model_load_utils.py`, delegating to Transformers/Accelerate `device_map="auto"` with `max_memory`, and Transformers' safe meta-creation patches. The PR discusses the last two and why global-meta creation is hazardous, but still does not explain why the two-tier retry is preferable to always using the existing `include_buffers=False` pattern. Please document that tradeoff before approval.

examples/hf_ptq/example_utils.py
cjluo-nv: Both points addressed in 1b062da36 — and you were right that the retry needed justifying. It did not survive the justification, so it is gone.

**Design.** The probe answers exactly one boolean: will the model spill to CPU, so should `max_memory` get the `gpu_max_mem_percentage` haircut. (`inferred_device_map` has two references in the file — it is built, tested with `"cpu" in .values()`, and discarded.) `infer_auto_device_map` reaches that via `compute_module_sizes`, which is `tensor.numel() * dtype_byte_size(tensor.dtype)` — shapes and dtypes, never storage.

So the probe only has to reproduce the module tree, and the binding constraint is that it must be **no stricter than the loader it predicts**:

| | how `from_pretrained` builds the model |
|---|---|
| transformers 4.57.6 | `init_contexts = [no_init_weights(), init_empty_weights()]` — `include_buffers` defaults to False (modeling_utils.py:4378) |
| transformers 5.5.4 | `torch.device("meta")` + `meta_device_safe_creation_ops()` (redirects `torch.linspace` to CPU) |
| old probe | bare `torch.device("meta")`, no patch |

`include_buffers=True` was stricter than both. accelerate implements it as a bare global device context, so it captures scratch arithmetic in `__init__` that has nothing to do with weights. Transformers added `meta_device_safe_creation_ops` precisely because pre-v5 remote code derives scalars that way. A model doing `int(torch.tensor(...))` in `__init__` loads fine via `from_pretrained` on 4.x and 5.x, but died in our probe — an optimization killing runs it exists only to speed up.

So the answer to "why not just use the existing `include_buffers=False` pattern" is: we should, and now do. Single build, no retry loop, no `noqa: PERF203`, no error accumulation. Measured cost on real checkpoints is nil — Qwen3-8B and DeepSeek-R1-Distill-Llama-70B both retain 0.0 MiB and produce identical sized totals (15.256 / 131.417 GiB) either way, for +0.09s and +0.19s.

One note on `build_meta_causal_lm`: it uses `include_buffers=False` for a different reason, not as a general preference. Its skeleton is kept and has real weights loaded into it under FSDP2, and `inv_freq` / `original_inv_freq` are `persistent=False` (verified absent from `state_dict()`), so they are computed at init and never loaded — on meta they would stay meta. It needs real buffers; this probe needs none. Same call, opposite motivation.

**Tests.** Added in `tests/examples/hf_ptq/test_example_utils.py`: the probe uses permissive patching, it survives a meta-hostile `__init__` (`int(torch.tensor(...))`, the Phi-4-MM shape), it returns `None` with a warning on failure, and `get_model` skips `infer_auto_device_map` while still calling `from_pretrained` — asserting no `max_memory` cap is invented, per the CodeRabbit thread. 32 passed in the file.

examples/hf_ptq/example_utils.py
cjluo-nv: Follow-up: `_build_meta_skeleton()` simplified further in 1b062da36. The two-tier retry is gone — it is now a single `include_buffers=False` build, because that is what Transformers' own `from_pretrained` uses and the old `include_buffers=True` probe was stricter than the loader it was predicting. Details in the design thread.

CHANGELOG.rst
```diff
@@ -83,6 +83,7 @@ Changelog
...
+- Dropped **Phi-3-vision** and **Phi-4-multimodal** PTQ support in ``examples/hf_ptq`` (NVBug 6563509). Their bundled remote code predates Transformers v5 and no longer loads on any version ModelOpt supports (``transformers>=4.57``): Phi-4-multimodal requires ``transformers<4.52`` because it reaches ``prepare_inputs_for_generation`` through ``peft``, which needs ``PreTrainedModel`` to still inherit ``GenerationMixin``, and both models declare ``_tied_weights_keys`` as a list, which Transformers 5.x rejects. The support-matrix row, the ``phi4mm`` model type, the multimodal-detection heuristics that only ever matched these two (``vision_lora`` / ``audio_processor`` / ``embd_layer.image_embd_layer``), the ``Phi3Image`` / ``PhiImage`` embedding-export exclusions, and the ``modelopt_recipes/huggingface/phi4mm/`` recipes have been removed. Text-only Phi-3/Phi-4 and Phi-3.5-MoE are unaffected.
```
meenchen: > _Bot comment._

This prior compatibility comment is still unresolved. The author confirmed Phi-3-vision was not tested on supported Transformers 4.57 and that its identified `_tied_weights_keys` failure applies to 5.x, so “no longer loads on any version ModelOpt supports” remains unsupported for that model. Please describe Phi-3-vision as a product/supersession removal (using the rationale from the author's reply), or provide the missing 4.57 repro/blocker. Keep the technical incompatibility claim scoped to Phi-4-MM where the version matrix supports it.

meenchen (commented):

Re-review: the implementation concern and most prior comments are addressed. The max-memory finding was correctly resolved by fixing the warning rather than inventing a fallback cap; the CLI spelling is now correct; and the new tests cover permissive skeleton construction, a meta-hostile initializer, helper failure/warning, and the `get_model` skip-and-continue path.

One important prior comment remains unresolved: the changelog still says Phi-3-vision cannot load on any supported Transformers version, although the author confirmed it was not tested on supported 4.57 and is being removed for product/supersession reasons. The PR body also remains stale: it describes and claims testing for a two-attempt retry that commit `1b062da36` removed. Please update both to describe the current single `include_buffers=False` probe and accurately characterize the Phi-3-vision removal.

Design review: the problem is preventing an optional model-sizing skeleton from aborting a potentially viable real load. The relevant alternatives are the repo's existing `build_meta_causal_lm` pattern (`modelopt/torch/utils/plugins/model_load_utils.py`) using `include_buffers=False`, delegating sizing/mapping to Accelerate/Transformers via `device_map="auto"` and `max_memory`, and Transformers' patched safe-meta construction. The author's design-thread reply now gives a convincing reason to adopt the first option, and the code/docstring do so, but the required PR-body rationale is not yet updated and currently documents the superseded retry design.

This PR also deletes recipe files carrying SPDX/license headers, so the licensing-related file removal should receive human sign-off before merge.

examples/hf_ptq/example_utils.py
```diff
@@ -873,17 +902,19 @@ def has_pack_quantized_config(config):
...
+        model_kwargs2 = _apply_dtype_to_config(
+            model_kwargs, config_dtype, architecture, apply_config_dtype=True
+        )
+        if auto_model_module not in [AutoModelForCausalLM, AutoModel]:
+            model_kwargs2.pop("trust_remote_code", None)
+        model_kwargs2.pop("max_memory", None)
+
+        # Only a sizing aid for ``infer_auto_device_map`` below; ``None`` when the model
+        # cannot be built on meta, in which case the estimate is skipped.
+        model = _build_meta_skeleton(from_config, config_for_init, model_kwargs2, architecture)
```
meenchen: > _Bot comment._

[P1] Please release this parameter-only skeleton before loading the checkpoint. With include_buffers=False, parameters are meta tensors, but registered buffers and other tensors retained by the constructor are real CPU allocations. This local remains referenced while the right-hand side of the later model = auto_model_module.from_pretrained(...) materializes the actual checkpoint, so the two allocations overlap; disk-offload also builds the skeleton even though that branch never calls infer_auto_device_map. For models with large constructor buffers or caches, this can bypass max_cpu_memory_gb and OOM the memory-constrained path. Please skip skeleton construction in disk-offload mode and release model after device-map inference, before from_pretrained().

cjluo-nv: Both points were correct, and they pushed the design somewhere better — addressed in 4e0157d0f, though not the way you suggested.

You were right on both counts against the code as it stood: Python evaluates the right-hand side before rebinding, so `model = auto_model_module.from_pretrained(...)` kept the probe alive for the whole checkpoint load, and with `include_buffers=False` its buffers are real CPU allocations. And the `_disk_offload` branch never touched the probe, so it was paying for one unused in exactly the memory-constrained path.

Rather than release the probe and skip it in disk-offload, the probe is gone. It only ever produced one boolean — will the weights fit on the GPUs, so should `max_memory` get the haircut — and that number is already on disk. `metadata.total_size` in the safetensors index equals what `compute_module_sizes` computed on the constructed model, exactly: Qwen3-8B 16381470720 B = 15.256 GiB, DeepSeek-R1-Distill-Llama-70B 141107412992 B = 131.417 GiB, both matching the meta model.

So `_checkpoint_size_bytes()` reads the index (falling back to summing shards, `None` if unreadable), and no model is constructed during device-map selection. That removes the overlap you flagged by construction, removes the disk-offload waste, and removes the original NVBug crash site entirely rather than tolerating it — Phi-4-multimodal now sizes at 10.383 GiB and reaches `from_pretrained` with no meta-tensor error. `_build_meta_skeleton`, its fallback warning, `_resolve_init_config`, the `from_config` resolution, and the accelerate `init_empty_weights` / `infer_auto_device_map` imports all went with it. Net 74 lines lighter.

Known approximation, worth flagging: total-bytes-vs-summed-GPU-budget ignores the per-device packing and no-split-module modelling `infer_auto_device_map` did. For a heuristic that only chooses whether to apply an 80% haircut, total size is the dominant term.

Tests cover the index path, the shard-sum fallback, a malformed index, capping only when the weights exceed the budget, and no cap when the size is unknown; the `get_model` tests also assert nothing is constructed during sizing.

CHANGELOG.rst
cjluo-nv: Addressed in 17dfd3b7a — you were right that the entry was making an unverified claim.

The changelog now splits the two rationales. Phi-4-multimodal is described as unloadable on any supported version, which the matrix does support. Phi-3-vision is described as a supersession removal: "dropped alongside it: it is the older, superseded model in the same family, so with its successor unsupportable there is no reason to keep carrying the predecessor." A parenthetical records what was actually verified — it shares the list-valued `_tied_weights_keys` defect so is broken on Transformers 5.x, but does not hit the `peft` blocker. No 4.57 claim is made for it. The PR description carries the same split.

CHANGELOG.rst
cjluo-nv: Addressed in 17dfd3b7a — you were right that the entry was making an unverified claim.

The changelog now splits the two rationales. Phi-4-multimodal is described as unloadable on any supported version, which the matrix does support. Phi-3-vision is described as a supersession removal: "dropped alongside it: it is the older, superseded model in the same family, so with its successor unsupportable there is no reason to keep carrying the predecessor." A parenthetical records what was actually verified — it shares the list-valued `_tied_weights_keys` defect so is broken on Transformers 5.x, but does not hit the `peft` blocker. No 4.57 claim is made for it. The PR description carries the same split.

CHANGELOG.rst
cjluo-nv: Addressed — first in 1b062da36, and the entry has since been rewritten again in 4e0157d0f.

You were right on both parts: the changelog still carried the nonexistent `--gpu_max_memory_percentage` spelling, and it implied that flag alone covered the lost estimate when in this path it only takes effect together with `--use_seq_device_map`.

That text is now gone entirely rather than corrected, because the fallback it described no longer exists: the device map is sized from the checkpoint on disk instead of from a constructed model, so there is no skeleton to fail and no estimate to lose. The entry now describes that mechanism, and the only remaining degradation is "size unreadable → no cap applied", the same as before.

examples/hf_ptq/example_utils.py
cjluo-nv: Follow-up: the helper this thread was about no longer exists as of 4e0157d0f. `_build_meta_skeleton` and its four tests are gone — the device map is now sized from the checkpoint on disk, so no model is constructed and there is no fallback path to cover. The replacement `_checkpoint_size_bytes` has its own tests (index metadata, shard-sum fallback, malformed index) plus `get_model` tests for capping only when the weights exceed the GPU budget, no cap when the size is unknown, and an assertion that nothing is constructed during sizing.

examples/hf_ptq/example_utils.py
cjluo-nv: Final state on this one: the helper you asked me to extract has now been deleted outright in 4e0157d0f. Sizing the device map from the checkpoint index removes the need to construct a model at all, so `_build_meta_skeleton` and everything that fed it (`_resolve_init_config`, the `from_config` resolution, the accelerate imports) are gone. Net 74 lines lighter than before the extraction.

examples/hf_ptq/example_utils.py
```diff
@@ -644,23 +637,38 @@ def get_original_hf_quant_method(config) -> str | None:
...
+    On-disk size is the checkpoint's own dtype, which is what gets loaded under the
+    default ``dtype="auto"``.
     """
-    if auto_model_module in [AutoModelForCausalLM, AutoModel]:
-        return hf_config
-    if not type(hf_config).__module__.startswith("transformers_modules"):
-        return hf_config
-    builtin_config_kwargs = {k: v for k, v in config_kwargs.items() if k != "trust_remote_code"}
+    path = Path(ckpt_path)
+    if not path.is_dir():
```
meenchen: > _Bot comment._

[P1] Preserve sizing for Hugging Face model IDs. This returns None whenever ckpt_path is not a local directory, but --pyt_ckpt_path explicitly accepts a Hugging Face model name and the README recommends passing the model card directly. Every org/model invocation therefore bypasses the headroom logic below; a large Hub model that device_map=auto spills to CPU can again fill GPU memory and OOM during calibration. Please resolve or download the small checkpoint index, or query Hub file metadata/cache, before returning unknown, and add a repository-ID test. Reserve the None fallback for genuinely unreadable layouts.

examples/hf_ptq/example_utils.py
```diff
@@ -904,8 +897,13 @@ def has_pack_quantized_config(config):
...
             )
         else:
-            inferred_device_map = infer_auto_device_map(model, max_memory=max_memory)
-            if "cpu" in inferred_device_map.values():
+            # Weights that do not fit on the GPUs leave no room for calibration
+            # activations, because device_map="auto" packs them to ~100% of free memory.
+            # Cap the budget in that case. Sized from the checkpoint on disk rather than
+            # from a constructed model, so no checkpoint ``__init__`` runs here.
+            checkpoint_bytes = _checkpoint_size_bytes(ckpt_path)
+            gpu_budget = sum(size for dev, size in max_memory.items() if isinstance(dev, int))
```
meenchen: > _Bot comment._

[P1] Preserve per-device placement when deciding whether to reserve calibration headroom. Total checkpoint bytes fitting within the sum of GPU memory does not mean Accelerate can keep the model off CPU: parameters are indivisible and placement is per device. For example, two 9.5 GiB leaf parameters total 19 GiB across two 10 GiB GPUs, but infer_auto_device_map places one on GPU 1 and one on CPU, leaving the used GPU 95% full. This aggregate check reports that case as fitting and omits max_memory precisely when calibration may OOM. Please use a placement-aware or tensor-aware check, or a conservative condition that cannot produce this false fit, and add a regression test.

examples/hf_ptq/example_utils.py
cjluo-nv: Correct, and it is one of the reasons the sizing logic is now gone entirely in 1b6a79d99.

You were right about the gap: `--pyt_ckpt_path` accepts a model card (its help says so, and the README uses `openai/gpt-oss-20b` at line 245), `hf_ptq.py:578` passes it straight into `get_model` unresolved, and `_checkpoint_size_bytes` bailed on anything that was not a local directory. Every hub-ID run silently lost the guard.

Rather than resolve hub metadata, the cap itself has been removed. It was only ever wanted in the offload case, where it is close to free — the run is already streaming weights, and the headroom is required because at 100% packed calibration OOMs immediately. Applied to a model that *does* fit it would convert a fully resident load into an offloaded one, which is a large certain slowdown introduced where there was none. That is why the original condition fired only on spill, and knowing which side you are on is exactly what needs a constructed model — the thing NVBug 6563509 is about.

So `device_map="auto"` now sizes itself; `from_pretrained` runs its own `infer_auto_device_map` on the real model regardless (`transformers/modeling_utils.py:1362`). `--use_seq_device_map` is the supported answer to GPU OOM, and `--gpu_max_mem_percentage` applies there and to `--offload_folder` — which is what its help text already said: "the percentage of available GPU memory to use for loading the model when device_map is set to sequential."

examples/hf_ptq/example_utils.py
cjluo-nv: The principle is right and the code is gone as of 1b6a79d99, but the suggested direction would have been wrong, and it is worth recording why.

On the principle: yes, an aggregate byte count cannot reproduce per-device placement, since parameters are indivisible. (The worked example does not hold — two 9.5 GiB parameters across two 10 GiB GPUs place one each and both fit. Three 6 GiB blocks on two 10 GiB GPUs is the real shape: 18 GiB aggregate against 20 GiB, yet the third spills.)

On the fix: a "conservative condition that cannot produce a false fit" would have been a regression. The cap is only wanted when the model is already offloading, where it is close to free and the headroom is required — at 100% packed, calibration OOMs immediately. Applied to a model that *does* fit, it converts a fully resident load into an offloaded one: a large, certain slowdown introduced where there was none. So a false "does not fit" is the more expensive error here, and biasing the check toward capping trades a rare loud OOM the user can fix with a flag for a guaranteed slowdown across a whole band of models.

Since the check must not be biased in either direction, and since answering it accurately requires constructing the model — which is the whole of NVBug 6563509 — the cap has been removed instead. `device_map="auto"` sizes itself; `from_pretrained` runs its own `infer_auto_device_map` on the real model anyway. `--use_seq_device_map` is the supported answer to GPU OOM.

No regression test, because there is no longer a condition to regress.

examples/hf_ptq/example_utils.py
cjluo-nv: Superseded by 701180ed6 — this code is back to its original form on main.

`example_utils.py` and its tests are restored to main except for the Phi-specific multimodal-detection heuristics. The meta-device skeleton, `infer_auto_device_map`, and the `--gpu_max_mem_percentage` cap all keep their original behavior, so there is nothing left in this thread to act on.

Reasoning: the cap is wanted exactly where it already fires — when the model is already offloading to CPU, where it costs little and the headroom is required. Every alternative explored on this branch (a more permissive skeleton, sizing from the checkpoint index, an unconditional budget) either changed behavior for models that fit today or moved the guard somewhere it does not belong, and all of it was for a crash that only ever affected the checkpoints this PR removes. NVBug 6563509 is resolved by dropping Phi-4-multimodal, not by changing the loader.

examples/hf_ptq/example_utils.py
cjluo-nv: Superseded by 701180ed6 — this code is back to its original form on main.

`example_utils.py` and its tests are restored to main except for the Phi-specific multimodal-detection heuristics. The meta-device skeleton, `infer_auto_device_map`, and the `--gpu_max_mem_percentage` cap all keep their original behavior, so there is nothing left in this thread to act on.

Reasoning: the cap is wanted exactly where it already fires — when the model is already offloading to CPU, where it costs little and the headroom is required. Every alternative explored on this branch (a more permissive skeleton, sizing from the checkpoint index, an unconditional budget) either changed behavior for models that fit today or moved the guard somewhere it does not belong, and all of it was for a crash that only ever affected the checkpoints this PR removes. NVBug 6563509 is resolved by dropping Phi-4-multimodal, not by changing the loader.

examples/hf_ptq/example_utils.py
cjluo-nv: Superseded by 701180ed6 — this code is back to its original form on main.

`example_utils.py` and its tests are restored to main except for the Phi-specific multimodal-detection heuristics. The meta-device skeleton, `infer_auto_device_map`, and the `--gpu_max_mem_percentage` cap all keep their original behavior, so there is nothing left in this thread to act on.

Reasoning: the cap is wanted exactly where it already fires — when the model is already offloading to CPU, where it costs little and the headroom is required. Every alternative explored on this branch (a more permissive skeleton, sizing from the checkpoint index, an unconditional budget) either changed behavior for models that fit today or moved the guard somewhere it does not belong, and all of it was for a crash that only ever affected the checkpoints this PR removes. NVBug 6563509 is resolved by dropping Phi-4-multimodal, not by changing the loader.

examples/hf_ptq/example_utils.py
cjluo-nv: Superseded by 701180ed6 — this code is back to its original form on main.

`example_utils.py` and its tests are restored to main except for the Phi-specific multimodal-detection heuristics. The meta-device skeleton, `infer_auto_device_map`, and the `--gpu_max_mem_percentage` cap all keep their original behavior, so there is nothing left in this thread to act on.

Reasoning: the cap is wanted exactly where it already fires — when the model is already offloading to CPU, where it costs little and the headroom is required. Every alternative explored on this branch (a more permissive skeleton, sizing from the checkpoint index, an unconditional budget) either changed behavior for models that fit today or moved the guard somewhere it does not belong, and all of it was for a crash that only ever affected the checkpoints this PR removes. NVBug 6563509 is resolved by dropping Phi-4-multimodal, not by changing the loader.

examples/hf_ptq/example_utils.py
cjluo-nv: Superseded by 701180ed6 — this code is back to its original form on main.

`example_utils.py` and its tests are restored to main except for the Phi-specific multimodal-detection heuristics. The meta-device skeleton, `infer_auto_device_map`, and the `--gpu_max_mem_percentage` cap all keep their original behavior, so there is nothing left in this thread to act on.

Reasoning: the cap is wanted exactly where it already fires — when the model is already offloading to CPU, where it costs little and the headroom is required. Every alternative explored on this branch (a more permissive skeleton, sizing from the checkpoint index, an unconditional budget) either changed behavior for models that fit today or moved the guard somewhere it does not belong, and all of it was for a crash that only ever affected the checkpoints this PR removes. NVBug 6563509 is resolved by dropping Phi-4-multimodal, not by changing the loader.

CHANGELOG.rst
cjluo-nv: Superseded by 701180ed6 — this code is back to its original form on main.

`example_utils.py` and its tests are restored to main except for the Phi-specific multimodal-detection heuristics. The meta-device skeleton, `infer_auto_device_map`, and the `--gpu_max_mem_percentage` cap all keep their original behavior, so there is nothing left in this thread to act on.

Reasoning: the cap is wanted exactly where it already fires — when the model is already offloading to CPU, where it costs little and the headroom is required. Every alternative explored on this branch (a more permissive skeleton, sizing from the checkpoint index, an unconditional budget) either changed behavior for models that fit today or moved the guard somewhere it does not belong, and all of it was for a crash that only ever affected the checkpoints this PR removes. NVBug 6563509 is resolved by dropping Phi-4-multimodal, not by changing the loader.

CHANGELOG.rst
meenchen: Can you please update this part to drop ph4mm?

meenchen (approved):

PR #2116, opened 2026-08-08
Add Qwen 3.5 0.8B Puzzletron configs

j-rausch:
### What does this PR do?

Type of change: New example and tests.

Qwen 3.5 0.8B does not have a version-pinned Puzzletron model example. It also needs a clear separation between a conservative default and broader targets that have not yet been validated end to end.

This PR pins the public checkpoint identity and geometry, then provides two search choices:

- `model.yaml` is the default and searches only FFN intermediate sizes `[3072, 2048]`.
- `advanced.yaml` is an explicit opt-in. It inherits the pinned model config and exposes six pruning axes. The axis structure is adapted from the existing Qwen 3.5 9B config, while the concrete targets are derived from the pinned 0.8B geometry. These targets were not selected from completed 0.8B search results and have not been fully runtime-validated.

Hugging Face conversion now passes the configured model revision to checkpoint resolution, so a repository ID resolves to the pinned teacher snapshot instead of the repository's current revision. Converted teachers record that source and revision locally and are reused only when the configured identity matches. Replacements are built and validated in a clean sibling directory before publication, preventing stale shards and preserving the previous teacher if conversion or the directory swap is interrupted. In distributed runs, rank-zero failures are broadcast before later barriers so peers exit promptly instead of waiting for a timeout. A pinned legacy teacher without this record is converted once to establish the identity; unpinned legacy checkpoints retain the existing resume behavior.

The outcome is a narrow default for normal example reuse plus a clearly labeled advanced surface for follow-up validation, without presenting derived targets as measured results. The proposed advanced `gdn_key_head_dim` change from `128` to `96` remains documented but disabled with no executable candidate values until physical runtime-equivalence evidence is available.

### Usage

Select the default model config in a run:

```yaml
- /families/qwen3_5/qwen3p5_0p8b/model@_global_
```

Replace that line with the advanced choice when evaluating the broader validated search surface:

```yaml
- /families/qwen3_5/qwen3p5_0p8b/advanced@_global_
```

### Testing

The dedicated Puzzletron v2 CPU suite covers the pinned public model identity, exact default and advanced domains, portable configuration discovery, Hydra inheritance and revision resolution, blocked-axis behavior, offline propagation of optional Hugging Face revisions through conversion, and rejection of a converted teacher from a different revision. No GPU smoke has been run, so the advanced configuration remains experimental.

## Summary by CodeRabbit

- **New Features**
  - Added support for the Qwen3.5 0.8B model, including standard and advanced pruning/search configurations.
  - Added pinned model revision support for more reproducible model conversion.

- **Bug Fixes**
  - Improved conversion recovery after interruptions or failures.
  - Conversion results are now validated before publishing, with safer reuse of compatible checkpoints.
  - Changed model revisions or incompatible layouts now trigger reconversion automatically.

- **Tests**
  - Added coverage for Qwen3.5 configuration validation and reliable conversion workflows.

### DIFF
--- a/examples/puzzletron/configs/families/qwen3_5/qwen3p5_0p8b/advanced.yaml
+++ b/examples/puzzletron/configs/families/qwen3_5/qwen3p5_0p8b/advanced.yaml
@@ -42,12 +42,12 @@ search_space:
       enabled: false
       teacher_value: 1
       values: []
-    # Physical realization of the 128 -> 96 target remains blocked on runtime
-    # equivalence evidence; selecting this overlay does not clear that blocker.
+    # The proposed 128 -> 96 target remains blocked on runtime-equivalence
+    # evidence. Keep it documented but non-executable until that gate passes.
     gdn_key_head_dim:
-      enabled: true
+      enabled: false
       teacher_value: 128
-      values: [96]
+      values: []
     gdn_value_head_dim:
       enabled: true
       teacher_value: 128
--- a/examples/puzzletron/configs/families/qwen3_5/qwen3p5_0p8b/advanced.yaml
+++ b/examples/puzzletron/configs/families/qwen3_5/qwen3p5_0p8b/advanced.yaml
@@ -1,0 +1,54 @@
+# @package _global_
+
+defaults:
+  - /families/qwen3_5/qwen3p5_0p8b/model@_global_
+  - _self_
+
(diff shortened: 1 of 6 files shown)


j-rausch:
@coderabbitai review

j-rausch:
@coderabbitai review

j-rausch:
@coderabbitai review

PR #805, opened 2026-01-21
vllm fakequant reload with modelopt state for HF

kinjalpatel27:
## What does this PR do?

**Type of change:**  new feature

**Overview:** 
- Added support to reload HF exported checkpoint with modelopt state in vLLM fakequant
- Added support to reload all quantizer parameters using `QUANT_FILE_PATH` instead of only amax

## Usage

```python
cd $PWD/examples/vllm_serve
python hf_ptq_export.py --pyt_ckpt_path meta-llama/Llama-3.2-3B-Instruct --quant_cfg NVFP4_DEFAULT_CFG --dataset cnn_dailymail --export_path llama3.2-3b --trust_remote_code --batch_size 1 --calib_size 512 --kv_cache_quant_cfg NVFP4_KV_CFG

cd $PWD/examples/vllm_serve
MODELOPT_STATE_PATH=llama3.2-3b/vllm_fq_modelopt_state.pth python vllm_serve_fakequant.py ../llm_ptq/llama3.2-3b/ -tp 1 --served-model-name llama3.2-3b--host 0.0.0.0 --port 8001 --trust-remote-code --enforce-eager  --disable-custom-all-reduce --gpu-memory-utilization 0.8
```

## Testing
- Exported the checkpoint using hf_ptq, reloaded to vllm fakequant example, manually checked the quantizer values
- Repeated above with mixed quantization by disabling quantization for few layers
- exported checkpoint with megatron-lm with nvfp4, reloaded the checkpoint with vllm

## Before your PR is "*Ready for review*"

- **Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md)** and your commits are signed.
- **Is this change backward compatible?**: Yes
- **Did you write any new necessary tests?**: No
- **Did you add or update any necessary documentation?**: Yes
- **Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?**: Yes

## Summary by CodeRabbit

* **New Features**
  * vLLM can reload fakequant from ModelOpt state, load external quantizer files, perform TP-aware conversion/sharding, and includes a HF PTQ export utility. New env vars: QUANT_FILE_PATH, MODELOPT_STATE_PATH, CALIB_BATCH_SIZE.

* **Documentation**
  * Serve README updated with export/load workflows, example commands, env var usage, and known limitations.

* **Bug Fixes**
[...]

PR #2120, opened 2026-08-10
Add optional MLflow tracking to the vLLM fake-quant server

cjluo-nv:
### What does this PR do?

Type of change: new feature

Wires `examples/vllm_serve/vllm_serve_fakequant.py` up to `modelopt.torch.utils.mlflow` via `--mlflow <tracking-uri>`, the same way #2023 did for `hf_ptq.py`, so a fake-quant serve records **what it actually quantized** and an evaluation of that endpoint can be traced back to a recipe. Without the flag, behavior is unchanged — every hook is gated on it.

Three design points worth review:

1. **The run is recorded in the vLLM worker, not the launcher.** `vllm_serve_fakequant.py` is the API-server frontend; the engine and its workers are separate processes whose stdout it never sees, so a run opened there would capture none of the calibration. The launcher instead only settles the tracking configuration — validating the URI, naming the experiment, recording the command the user actually typed — and publishes it through the environment, which is how every other setting in this example (`QUANT_CFG`, `RECIPE_PATH`, …) already reaches the workers. Global rank 0 opens the run, so a TP-8 serve produces one run.

2. **The run covers load-through-warm-up, not the server's lifetime.** It opens *before the weights load*, so an unreachable server or a missing token fails in seconds rather than after a load and a full calibration, and it closes `FINISHED` once the model is quantized and warmed up. A run that stayed open for the serving lifetime would never close cleanly on SIGTERM.

3. **`recipe/quant_cfg.yaml` is only written on the preset path.** With `RECIPE_PATH`, `get_quant_config` returns the recipe's `quantize` section unchanged and `resolved_recipe.yaml` already carries it. With `QUANT_CFG`/`KV_QUANT_CFG` it is the *only* record of what ran: the params carry the preset names, while the config reaching `mtq.quantize` is those two deep-copied, merged, and — for an MLA model — extended at runtime with `*kv_c_bmm_quantizer` / `*k_pe_bmm_quantizer` by inspecting the loaded model.

Uploaded artifacts:

| Artifact | Contents |
| --- | --- |
| `command.txt` | The launcher's invocation, copy-pasteable, credentials masked |
| `version.txt` | The ModelOpt version that ran |
| `recipe/resolved_recipe.yaml` | `RECIPE_PATH` with its `$import`s expanded |
| `recipe/quant_cfg.yaml` | Merged `QUANT_CFG`/`KV_QUANT_CFG` + MLA fixup (preset path only) |
| `logs/<script>.log` | The rank-0 worker's stdout/stderr, including a crash traceback |
| `summary/quant_summary.txt` | The per-quantizer summary |

Plus the quantization *and* serving settings as searchable params, and `user` / `hostname` / `modelopt_version` / `git_sha` / `vllm_version` tags. The `checkpoint_path` tag matches the one `hf_ptq.py` sets, so a checkpoint's PTQ run and every serve of it join up.

Two small library additions, both consumed by the new example module:

- `command_text(argv=None)` — records another process's invocation, since a spawned worker's own `sys.argv` is vLLM plumbing rather than anything a user typed.
- `MlflowRunLogger.log_text()` — uploads a value settled midway through a run, so a crash during calibration still keeps the config that caused it.

The example `Dockerfile` installs the `mlflow` extra; the client remains optional and is imported only once tracking is enabled.

### Usage

```bash
RECIPE_PATH=<recipe.yaml> python vllm_serve_fakequant.py <model_path> -tp 8 \
  --host 0.0.0.0 --port 8000 \
  --mlflow https://<your-mlflow-server>/
```

```
[mlflow] tracking to https://<your-mlflow-server>, experiment $USER/vllm_serve_fakequant/<model>-<recipe>
(Worker_TP0) [mlflow] run: https://<your-mlflow-server>/#/experiments/19/runs/1c6679448f25...
```

`--mlflow-experiment` / `--mlflow-run-name` override the defaults. `$MLFLOW_TRACKING_URI` enables tracking on its own and is best-effort; an explicit `--mlflow` overrides it and fails loudly.

> This is the **quantization** tracking server. It is unrelated to any server an evaluation harness exports its scores to — NeMo Evaluator Launcher has its own `export.mlflow.tracking_uri`. The README calls this out.

### Testing

**Unit — 87 passing** (`tests/examples/vllm_serve/test_vllm_mlflow_utils.py`, 33 new; `tests/unit/torch/utils/test_mlflow.py`, +5). `vllm_mlflow_utils` deliberately imports no vLLM, so the whole launcher→worker handover is covered without a GPU, a server, or the mlflow client.

**End to end on aws-cmh** (4× GB300, `simple_evals.gpqa_diamond`, Nemotron-3.5-Lightning-30B-A3B-BF16 fake-quantized with `general/ptq/nvfp4_mlp_only-kv_fp8_cast`): run `FINISHED` in 261.5 s, opened by `Worker_TP0` only, all artifacts present and verified by content — `command.txt` held the launcher's invocation rather than the worker's spawn argv, and `resolved_recipe.yaml` was 6797 B against 1845 B of source. 104 quantizers enabled (92 NVFP4 dynamic block-16 expert weight/input with calibrated amax, 12 FP8 KV bmm). The eval then ran to completion against the served endpoint, 22/22 requests HTTP 200.

Two bugs the hardware run caught, both fixed here with regression tests:

- `--mlflow_run_name` was rejected. vLLM's `FlexibleArgumentParser.parse_args` rewrites **every** `--foo_bar` to `--foo-bar` before matching, so a flag registered only under the underscored spelling is unreachable from its CLI. Both spellings are now registered. A unit test on a plain `ArgumentParser` could not have caught this.
- `recipe/quant_cfg.yaml` uploaded a Python `repr` blob under a `.yaml` name: a recipe's `quantize` is a `QuantizeConfig`, `yaml.safe_dump` raises `RepresenterError` on it, and the old JSON fallback stringified the object. `_dump_yaml` now unwraps pydantic via `model_dump(mode="json")` and raises otherwise, with the caller downgrading that to a warning so a bad config cannot take down a serve.

**Known coverage gap:** the preset (`QUANT_CFG`/`KV_QUANT_CFG`) path — the only one that now writes `recipe/quant_cfg.yaml` — is covered by unit test but has not been exercised on hardware; the canary used `RECIPE_PATH`. Likewise the case where `$MLFLOW_TRACKING_URI` is present *inside* the deployment container and `--mlflow` overrides it is unit-tested only: NeMo Evaluator Launcher forwards only declared env vars, so the eval server's URI never entered the container in the canary.

- Is this change backward compatible?: ✅ — new optional flags only; no `--mlflow` means no behavior change.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ — no new dependency. Uses the existing optional `nvidia-modelopt[mlflow]` extra (`mlflow-skinny`, Apache-2.0) added in #2023; the example `Dockerfile` now installs it. No code copied from other sources.
- Did you write any new necessary tests?: ✅ — 38 new tests.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — 0.47 Misc.
- Did you get Claude approval on this PR?: ❌ — `/claude review` not yet run.

### Additional Information

Follows #2023, which added `MlflowRunLogger` and the `hf_ptq.py` integration.

Note for anyone tracking from an OCI cluster: `mlflow-modelopt.nvidia.com` is unreachable from oci-nrt and oci-hsg. TCP 443 completes and the connection is then reset on the first application byte, regardless of SNI or protocol, one RTT away — the PDX PaaS ingress appears to apply a source-IP policy, and the OCI clusters egress from Oracle-owned addresses (`155.248.190.0`, `168.110.199.1`) rather than NVIDIA's. gcp-nrt, aws-cmh and cw-dfw all reach it. This is an infrastructure matter, not a property of this change, but it determines where the feature is usable today.

## Summary by CodeRabbit

* **New Features**
  * Added optional MLflow tracking for vLLM fake-quantization serving runs.
  * Records serving, quantization, worker, and invocation metadata, including configuration and summary artifacts.
  * Supports tracking URI, credentials, environment, and command-line configuration.
  * Added command and text artifact logging for active MLflow runs.
* **Documentation**
  * Documented setup, configuration, recorded artifacts, lifecycle, and fallback behavior.
  * Updated the example container to include MLflow support.
* **Tests**
  * Added comprehensive coverage for tracking configuration, logging, failures, and disabled tracking.

### DIFF
--- a/examples/vllm_serve/vllm_mlflow_utils.py
+++ b/examples/vllm_serve/vllm_mlflow_utils.py
@@ -1,0 +1,401 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""MLflow tracking for the vLLM fake-quant server, mirroring ``examples/hf_ptq``.
+
+The quantization this example performs happens inside the vLLM **worker** process, not in
+``vllm_serve_fakequant.py``: the launcher is the API-server frontend, and the engine and its
+workers are separate processes whose output it never sees. So the launcher only settles the
+tracking configuration -- validating the URI, naming the experiment, recording the command
+the user actually typed -- and publishes it through the environment, the same way every
+other setting in this example reaches the workers. Rank 0 opens the run and uploads the
+recipe, the effective quantization config, the calibration log and the quantizer summary.
+
+The run covers weight load through warm-up and closes ``FINISHED`` when the server is ready
+to serve, rather than staying open for the server's whole lifetime.
+
+Nothing here imports vLLM, so the tracking can be exercised without it.
+"""
+
+import argparse
+import contextlib
+import os
+import shutil
+import tempfile
+import warnings
+from collections.abc import Iterator
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+from modelopt.torch.utils.mlflow import (
+    MlflowRunLogger,
+    command_text,
+    default_experiment_name,
+    validate_tracking_uri,
+)
+
+TOOL_NAME = "vllm_serve_fakequant"
+
+# Written by the launcher, read by the workers. The two MLflow-owned names are MLflow's own,
+# so a shell that already exports them opts in without touching the command line.
+TRACKING_URI_ENV = "MLFLOW_TRACKING_URI"
+EXPERIMENT_ENV = "MLFLOW_EXPERIMENT_NAME"
+RUN_NAME_ENV = "MODELOPT_MLFLOW_RUN_NAME"
+REQUIRED_ENV = "MODELOPT_MLFLOW_REQUIRED"
+COMMAND_ENV = "MODELOPT_MLFLOW_COMMAND"
+
+# Everything the rank-0 worker needs in its environment to reach the tracking server. The
+# credentials are never set here, only forwarded when the launching shell exported them --
+# without that, a Ray worker authenticates as nobody and the run fails to open.
+MLFLOW_ENV_VARS = frozenset(
+    {
+        TRACKING_URI_ENV,
+        EXPERIMENT_ENV,
+        RUN_NAME_ENV,
+        REQUIRED_ENV,
+        COMMAND_ENV,
+        "MLFLOW_TRACKING_TOKEN",
+        "MLFLOW_TRACKING_USERNAME",
+        "MLFLOW_TRACKING_PASSWORD",
+        "MLFLOW_TRACKING_INSECURE_TLS",
+        "MLFLOW_HTTP_REQUEST_MAX_RETRIES",
+    }
+)
+
+
+def add_mlflow_args(parser: argparse.ArgumentParser) -> None:
+    """Add the MLflow tracking flags to the launcher's parser.
+
+    The multi-word flags are registered under both spellings. vLLM's
+    ``FlexibleArgumentParser`` rewrites every ``--foo_bar`` on the command line to
+    ``--foo-bar`` before matching, so the dashed spelling is the one that has to exist for
+    the flag to be reachable at all; the underscored spelling is what ``hf_ptq`` uses and
+    keeps these usable with a plain ``argparse.ArgumentParser``.
+    """
+    parser.add_argument(
+        "--mlflow",
+        default=None,
+        help=(
+            "Track this server's calibration on an MLflow server "
+            "(e.g. https://<your-mlflow-server>/), uploading the command, the resolved "
+            "recipe, the quantization config actually applied, the worker log and the "
+            "quantizer summary. This is the quantization tracking server, which is "
+            "unrelated to any tracking server an evaluation harness exports its scores to. "
+            "MLflow's own $MLFLOW_TRACKING_URI enables tracking without this flag, which "
+            "overrides it. A URI taken from the environment is best-effort: if it is "
+            "unusable the server warns and serves untracked."
+        ),
+    )
+    parser.add_argument(
+        "--mlflow-experiment",
+        "--mlflow_experiment",
+        default=None,
+        help=(
+            "MLflow experiment name. Default: "
+            f"$USER/{TOOL_NAME}/<model basename>-<recipe name, or the quantization config>."
+        ),
+    )
+    parser.add_argument(
+        "--mlflow-run-name",
+        "--mlflow_run_name",
+        default=None,
+        help="MLflow run name. Default: the UTC start time as YYYYmmdd-HHMMSS.",
+    )
+
+
+def resolve_mlflow_args(args: argparse.Namespace, parser: argparse.ArgumentParser) -> None:
+    """Settle the tracking configuration and publish it to the worker processes.
+
+    Validating here rather than in the worker is what makes a typo in the URI fail at launch
+    instead of after the weights are on the GPUs. As in ``hf_ptq``, only ``--mlflow`` is a
+    deliberate request and therefore fatal when unusable; ``$MLFLOW_TRACKING_URI`` is
+    commonly exported for unrelated tooling and must not take a serve down with it.
+    """
+    required = args.mlflow is not None
+    uri = args.mlflow or os.environ.get(TRACKING_URI_ENV) or None
+    if uri:
+        try:
+            uri = validate_tracking_uri(uri)
+        except ValueError as e:
+            if required:
+                parser.error(f"--mlflow: {e}")
+            warnings.warn(f"Ignoring ${TRACKING_URI_ENV}, continuing untracked: {e}")
+            uri = None
+    if not uri:
+        # A rejected URI must not reach the workers, which would try it again and fail there.
+        os.environ.pop(TRACKING_URI_ENV, None)
+        return
+
+    os.environ[TRACKING_URI_ENV] = uri
+    os.environ[REQUIRED_ENV] = "1" if required else "0"
+    # The workers' own sys.argv is vLLM's spawn plumbing; this is the command a user ran.
+    os.environ[COMMAND_ENV] = command_text()
+    os.environ[EXPERIMENT_ENV] = (
+        args.mlflow_experiment
+        or os.environ.get(EXPERIMENT_ENV)
+        or default_experiment_name(TOOL_NAME, args.model, quant_variant())
+    )
+    if args.mlflow_run_name:
+        os.environ[RUN_NAME_ENV] = args.mlflow_run_name
+    print(f"[mlflow] tracking to {uri}, experiment {os.environ[EXPERIMENT_ENV]}")
+
+
+def quant_variant() -> str:
+    """What distinguishes this serve of the model, for the default experiment name.
+
+    Read from the environment rather than taken as an argument because that is where this
+    example's quantization settings live, and both the launcher and a worker that was
+    started directly (``vllm serve --worker-cls fakequant_worker.FakeQuantWorker``) need it.
+    """
+    if recipe := os.environ.get("RECIPE_PATH"):
+        return Path(recipe.rstrip("/")).stem
+    quant_cfg = os.environ.get("QUANT_CFG")
+    kv_quant_cfg = os.environ.get("KV_QUANT_CFG")
+    if quant_cfg or kv_quant_cfg:
+        return "-".join(cfg for cfg in (quant_cfg, kv_quant_cfg) if cfg)
+    if os.environ.get("MODELOPT_STATE_PATH"):
+        return "modelopt_state"
+    if os.environ.get("QUANT_FILE_PATH"):
+        return "quantizer_state"
+    return "unquantized"
+
+
+class FakeQuantMlflowTracker:
+    """Records one vLLM fake-quant worker's calibrate-and-serve as an MLflow run.
+
+    Inert unless the launcher published a tracking URI *and* this is the global rank-0
+    worker, so the worker needs no branching: every method is a no-op otherwise, and the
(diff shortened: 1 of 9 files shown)


ChenhanYu (approved):

examples/vllm_serve/vllm_mlflow_utils.py
```diff
@@ -0,0 +1,401 @@
...
+                texts=self._start_texts(),
+                files=self._files,
+            )
+        except BaseException:
+            # An explicit --mlflow is fatal here by design; leave no staging directory behind.
+            self._discard_staging()
+            raise
+
+    def log_quant_config(self, quant_cfg: Any) -> None:
+        """Upload the merged ``QUANT_CFG``/``KV_QUANT_CFG`` config, when that is what ran.
```
meenchen, line 249: > _Bot comment._

When tracking comes from `$MLFLOW_TRACKING_URI`, `MlflowRunLogger.start()` handles an open failure by setting `enabled = False` and returning rather than raising. In that path this tracker has already created `_staging`, but every later method (including `finish`) returns immediately because the logger is disabled, so the temp directory is never removed. Please detect that post-`start()` state and call `_discard_staging()` (and add a regression test for a best-effort start failure).

examples/vllm_serve/vllm_mlflow_utils.py
```diff
@@ -0,0 +1,401 @@
...
+        # already printed is not repeated.
+        mtq.print_quant_summary(model, output_dir=str(self._staging))
+
+    def finish(self, status: str) -> None:
+        """Upload the log and the summary, and close the run with *status*.
+
+        Only the first call has an effect: vLLM drives the worker through several guarded
+        steps, and a run already closed as ``FINISHED`` must not be reopened or downgraded
+        by a failure in whatever the server does next.
+        """
```
meenchen, line 286: > _Bot comment._

Please move this ModelOpt import to the top of the file. The same applies to `from modelopt.recipe import load_recipe` in `_start_texts`; neither has a stated circular-import, optional-dependency, platform, or startup-cost reason for being local. (The local `vllm` import is justified by this module's explicit no-vLLM design.)

meenchen (commented):

Design review: the PR solves the launcher/worker split needed to associate vLLM fake-quantization with an MLflow run. The main alternatives are reusing the existing `MlflowRunLogger`/HF PTQ integration (which this does), calling the already-installed optional MLflow client directly, or keeping only local logs/artifacts. The PR body explicitly explains why tracking must be opened in the worker and why the existing logger is extended, so the architectural choice is reasonably justified. The large diff is cohesive and well tested, and the new-file license headers match `LICENSE_HEADER`. Two small implementation issues remain: a best-effort connection failure leaks its staging directory, and two new ModelOpt imports are unnecessarily function-local.

examples/vllm_serve/vllm_mlflow_utils.py
cjluo-nv: Confirmed and fixed in d30483c14. `start()` now calls `_discard_staging()` when `MlflowRunLogger.start()` returns with `enabled` false, which is the best-effort path you described — the existing `except BaseException` only covered the explicit `--mlflow` case that raises.

Two regression tests in `tests/examples/vllm_serve/test_vllm_mlflow_utils.py`, one per path. I checked the best-effort one fails without the fix rather than passing vacuously.

examples/vllm_serve/vllm_mlflow_utils.py
cjluo-nv: Fixed in d30483c14 — both `modelopt.torch.quantization` and `from modelopt.recipe import load_recipe` are now at module scope.

You were right that there was no stated reason, and there was no real one either: the module already imports `modelopt.torch.utils.mlflow` at the top, which pulls both `torch` and `modelopt.torch.quantization` into `sys.modules`, so deferring them saved nothing. `vllm_ptq_utils.py` and `fakequant_worker.py` already import them at the top too. The local `vllm` import stays, as you noted.

examples/vllm_serve/vllm_mlflow_utils.py
```diff
@@ -0,0 +1,401 @@
...
+            # Never let an unserializable config take down a serve that would have worked.
+            print(f"[mlflow] WARNING: could not serialize the quantization config: {e}")
+            return
+        self._logger.log_text("recipe/quant_cfg.yaml", text)
+
+    def log_quant_summary(self, model: Any) -> None:
+        """Stage the per-quantizer summary for upload; a no-op when untracked."""
+        if not self._logger.enabled or self._staging is None:
+            return
+        import modelopt.torch.quantization as mtq
```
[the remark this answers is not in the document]
cjluo-nv: Fixed in d30483c14: both ModelOpt imports moved to module scope. The deferral saved nothing — the top-level `modelopt.torch.utils.mlflow` import already brings `torch` and `modelopt.torch.quantization` into `sys.modules`. The remaining local import (`vllm`, in `_vllm_version`) is deliberate and carries the reason in a comment: this module is designed to import no vLLM so the launcher/worker handover can be unit-tested without it.

examples/vllm_serve/vllm_mlflow_utils.py
```diff
@@ -0,0 +1,401 @@
...
+    # The workers' own sys.argv is vLLM's spawn plumbing; this is the command a user ran.
+    os.environ[COMMAND_ENV] = command_text()
+    os.environ[EXPERIMENT_ENV] = (
+        args.mlflow_experiment
+        or os.environ.get(EXPERIMENT_ENV)
+        or default_experiment_name(TOOL_NAME, args.model, quant_variant())
+    )
+    if args.mlflow_run_name:
+        os.environ[RUN_NAME_ENV] = args.mlflow_run_name
+    print(f"[mlflow] tracking to {uri}, experiment {os.environ[EXPERIMENT_ENV]}")
```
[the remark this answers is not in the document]
cjluo-nv: The first point is fixed in d30483c14. `resolve_mlflow_args` now prints the URI through a `_without_credentials()` helper that strips any `user:token@`, matching what `MlflowRunLogger` already does for every URI it prints or uploads. It mattered more than it looks: that line is captured into the worker log which the run then uploads as an artifact. Regression test asserts the credentials are absent from stdout while the unmasked URI still reaches the workers, which need it to authenticate.

The second point (sanitizing captured worker stdout, including `MODELOPT_STATE_PATH`) is not changed by this commit.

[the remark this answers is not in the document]
cjluo-nv: On the second point — sanitizing captured worker stdout, including `MODELOPT_STATE_PATH` — I am not planning to change this, for three reasons:

1. It is already documented behaviour, not an oversight. `MlflowRunLogger`'s class docstring carries a `Note:` saying the captured log "is whatever the script printed, so a secret echoed to stdout still reaches the server. Prefer passing credentials via the environment." That is the contract this example relies on, and it points users at the right mitigation.
2. `MODELOPT_STATE_PATH` is a filesystem path the user passes on their own command line, to a tracking server they name themselves with `--mlflow`. It is not a credential, and the artifact inherits that server's access control. The `print` at `fakequant_worker.py:68` also predates this PR — it came in with #805 (48da2812c, 2026-03-30).
3. A scrubbing pass over all worker stdout would strip exactly what the log exists for. The reason to upload it is to debug a calibration that went wrong, and calibration diagnostics are mostly module names and paths.

Happy to reconsider if there is a concrete value here that is secret rather than merely path-shaped.

examples/vllm_serve/vllm_mlflow_utils.py
```diff
@@ -0,0 +1,401 @@
...
+import yaml
+
+from modelopt.torch.utils.mlflow import (
+    MlflowRunLogger,
+    command_text,
+    default_experiment_name,
+    validate_tracking_uri,
+)
+
+TOOL_NAME = "vllm_serve_fakequant"
```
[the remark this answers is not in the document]
cjluo-nv: I would rather not add this one, though I want to be accurate about the convention: `sparse_attn_worker.py` in this same directory does define `__all__`, so it is not unheard of here. Across `examples/` it is 10 files out of 237, and the other five modules in `examples/vllm_serve/` do not.

My reason is the guideline's own stated rationale: it is about packages — "`__init__.py` files re-export submodules with `from .module import *`... makes star-imports safe by limiting them to the curated `__all__` names." `vllm_mlflow_utils` is a standalone module imported by path from `vllm_serve_fakequant.py` and `fakequant_worker.py`, sits in no package, is re-exported from no `__init__.py`, and is never star-imported, so none of that applies. Its public surface is already pinned by the two explicit `from vllm_mlflow_utils import ...` statements that use it.

That said this is a preference, not a principle — say the word and I will add the four names.

examples/vllm_serve/vllm_mlflow_utils.py
```diff
@@ -0,0 +1,417 @@
...
+    ``MlflowRunLogger`` masks the same thing in every URI it prints or uploads, and this
+    line ends up in the worker log that the run itself uploads, so it has to match.
+    """
+    parsed = urlparse(uri)
+    return parsed._replace(netloc=parsed.netloc.rpartition("@")[2]).geturl()
+
+
+def quant_variant() -> str:
+    """What distinguishes this serve of the model, for the default experiment name.
+
```
meenchen, line 174: > _Bot comment._

The remaining URI-redaction concern is still present: this removes only `user:token@`, while preserving query parameters and fragments. For example, `--mlflow 'https://host/path?access_token=secret'` prints the secret here; because `command_text()` relies on the same user-info-only redaction in `modelopt/torch/utils/mlflow.py`, it also stores that URI in `command.txt`, and `MlflowRunLogger.run_url` can print it again. Please redact/drop query and fragment data (preferably centrally in the MLflow utility and reuse it here), and add a regression test showing a query token is absent from launcher output and the uploaded command/run URL while the unmodified URI still reaches MLflow.

meenchen (commented):

Re-review: the design question is adequately addressed. The problem is carrying fake-quant tracking across the vLLM launcher/worker boundary; alternatives include the existing HF-PTQ `MlflowRunLogger` pattern, direct use of the already-optional `mlflow-skinny` client, or local-only artifacts. This PR reuses and minimally extends the in-repo logger, and the PR body explains why the worker owns the run and why environment handoff is needed. The 1,133-line diff is large but cohesive, with substantial unit coverage and a reported hardware run. The new source/test headers match `LICENSE_HEADER`.

Previous review status: the staging-directory leak (correctness) is fixed with both best-effort and required-path regression tests; the function-local ModelOpt imports (minor) are moved to module scope; the `__all__` and broad worker-log-scrubbing suggestions were withdrawn. One security issue remains from the URI-redaction thread: query/fragment secrets are still emitted and can also enter `command.txt`.

realAsma (approved):

PR #1913, opened 2026-07-05
test: direct unit suites for opt Hparam and BaseSearcher (78 tests)

arham766:
### What does this PR do?

Type of change: new tests

Part of the unit-coverage initiative in #1902. Direct suites for the two search-framework primitives (no direct coverage existed; the duplication audit against test_dynamic.py, tests/unit/torch/nas, and test_autoquant.py is in the review notes):

- **test_hparam.py (44 tests)** — construction/dedup/original folding, configurability incl. the _force_configurable context manager, active-value validation with exact error messages, active_slice/enforce_order LongTensor semantics, choices-setter subset rules, importance registration (exact max-normalization values, multi-estimator summation, unsortable rejection), __iand__ merging, and pickle round-trip (required for Megatron distributed broadcast per the source NOTE).
- **test_searcher.py (34 tests)** — default-config contract, sanitize merge rules, the search() hook order and training-mode restoration, deepcopied default state, checkpoint path-resolution table (file/dir/extensionless/pathlib), strict vs non-strict state loading, end-to-end resume across two search() calls, scoring, and forward-loop construction with exact iteration counts.

### Usage

N/A — tests only.

### Testing

78 tests, <1s, hermetic (CPU-only, tmp_path, deterministic). Adversarially reviewed: 12 expectations independently re-derived from source, 6/6 seeded mutations killed, no vacuous tests, Windows/Linux path-agnostic. Full tests/unit/torch/opt dir passes alongside (excluding the pre-existing test_lr_config.py env failures).

### Before your PR is "*Ready for review*"

- Is this change backward compatible?: ✅
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in CONTRIBUTING.md: N/A
- Did you write any new necessary tests?: ✅
- Did you update Changelog?: N/A
- Did you get Claude approval on this PR?: N/A (external contributor)

### Additional Information

Issue: #1902

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit coverage for `Hparam`, including defaults, configurability, active/choices updates, ordering, importance handling, merging, and pickle roundtrips.
  * Added extensive `BaseSearcher` tests covering search flow, checkpoint save/load, resume behavior, scoring, and forward-loop/data-loader execution rules.

PR #2128, opened 2026-08-10
Fix Hparam configurability restoration after exceptions

taking-lying-flat:
### What does this PR do?

Type of change: Bug fix

`Hparam._force_configurable()` restored `_is_configurable` only after a normal context-manager exit. If code inside the context raised, the restoration statement was skipped and a previously non-configurable `Hparam` remained permanently configurable.

Wrap the `yield` in `try`/`finally` so the original state is restored on both normal and exceptional exits. Add a regression test that raises inside the context and verifies the temporary configurable state does not leak.

### Usage

No call-site changes are required. Existing users of the context manager retain the same behavior, including when an operation inside the context fails:

```python
with hp._force_configurable():
    hp.choices = new_choices
```

### Testing

- Verified the new regression test fails against the pre-fix implementation because `hp.is_configurable` remains `True`.
- `.venv/bin/python -m pytest tests/unit/torch/opt/test_dynamic.py -q` (`3 passed`)
- `.venv/bin/python -m pytest tests/unit/torch/opt -q` (`66 passed, 2 skipped`)
- Relevant pre-commit hooks passed: file/merge/line-ending checks, Ruff check and format, mypy, license insertion check, Bandit.
- `git diff --check`

- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — this is a small, backward-compatible correctness fix rather than a critical bug fix.
- Did you get Claude approval on this PR?: N/A — not available to an external contributor before submission.

### Additional Information

Related prior work: #1913 was closed without merging and only tested the normal-exit path. This PR adds exception-path coverage together with the production fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration handling so temporary settings are reliably restored even when an error occurs.
  * Exceptions raised during configuration operations continue to be reported correctly.

* **Tests**
  * Added coverage for configuration restoration after a failed operation.

### DIFF
--- a/modelopt/torch/opt/hparam.py
+++ b/modelopt/torch/opt/hparam.py
@@ -96,8 +96,10 @@ class Hparam:
         """Context manager to temporarily set hparam to be configurable."""
         original_value = self._is_configurable
         self._is_configurable = True
-        yield
-        self._is_configurable = original_value
+        try:
+            yield
+        finally:
+            self._is_configurable = original_value
 
     @property
     def is_sortable(self):
--- a/tests/unit/torch/opt/test_dynamic.py
+++ b/tests/unit/torch/opt/test_dynamic.py
@@ -26,6 +26,17 @@ from modelopt.torch.opt.dynamic import (
 )
 
 
+def test_force_configurable_restores_state_after_exception():
+    hp = Hparam([1, 2])
+    hp._is_configurable = False
+
+    with pytest.raises(RuntimeError, match="forced failure"), hp._force_configurable():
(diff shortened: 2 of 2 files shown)


kevalmorabia97 (approved):

kevalmorabia97:
/ok to test [5496251](https://github.com/NVIDIA/Model-Optimizer/pull/2128/commits/5496251b3952c3cfa3c290ebb6f4c6273b92f52d)

PR #2132, opened 2026-08-10
Skip weight quantizers with no stored weight in fold_weight

### LINKED ISSUE #2131: fold_weight crashes with AttributeError: 'NoneType' object has no attribute 'data' on Megatron models with tied word embeddings
## Describe the bug

On Megatron-Core models with tied word embeddings
(`share_embeddings_and_output_weights=True`), the `output_layer` stores
`weight = None` and borrows the embedding weight at forward time
(`skip_weight_param_allocation=True`). ModelOpt still attaches an
`output_layer.weight_quantizer` to it, so `mtq.fold_weight` matches the pair and
dereferences the `None` weight:

```
modelopt/torch/quantization/nn/modules/quant_module.py, in fold_weight
    weight.data.copy_(attr(weight.float()).to(weight.dtype))
AttributeError: 'NoneType' object has no attribute 'data'
```

`QuantModule.fold_weight` selects pairs on the `*_weight_quantizer` name and
`fake_quant` alone, and `_fold_weight_quantizer` accesses `weight.data` with no
`None` guard (still the case on current `main`):

https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/6b02f528bbc46b5ed474fade47add23f6810f0f3/modelopt/torch/quantization/nn/modules/quant_module.py#L144-L145

Normal QAT forwards never hit this because the Megatron plugin quantizes the weight
*argument* passed to the linear functional, not the stored `module.weight`;
`fold_weight` is the first path that reads the stored attribute, which is genuinely
`None` on the tied layer. (HF models tie via a shared tensor instead of `None`, so
only Megatron is affected.) There is no workaround from outside: `mtq.fold_weight`
walks every `QuantModule`, and per-module `fold_weight()` can't skip one pair.

## Steps/Code to reproduce bug

```python
from megatron.core.models.gpt import GPTModel

# Any GPTModel with tied embeddings, e.g. Qwen3-0.6B:
# output_layer.weight is None (borrowed from the embedding at forward time).
model = GPTModel(..., share_embeddings_and_output_weights=True)
model = mtq.quantize(model, mtq.NVFP4_DEFAULT_CFG, forward_loop)

mtq.fold_weight(model)  # AttributeError: 'NoneType' object has no attribute 'data'
```

Hit in practice in NeMo-RL quantization-aware GRPO (Megatron backend, Qwen3-0.6B);
https://github.com/NVIDIA-NeMo/RL/pull/3441 currently reimplements the per-pair fold
with a `None` guard as a workaround.

## Expected behavior

`fold_weight` skips pairs whose weight is not a tensor:

```python
weight = getattr(self, weight_name)
if not isinstance(weight, torch.Tensor):
    continue  # e.g. Megatron tied-embedding output_layer: weight = None
self._fold_weight_quantizer(attr, (weight,), keep_attrs)
```

Skipping is correct: the tied `output_layer` has no stored weight, and the tensor it
borrows belongs to the embedding module, whose own pair is folded on that module.

## System information

- ModelOpt: 0.46.0.dev86 (also present on `main` @ 6b02f52)
- Model: Megatron-Core `GPTModel` with `share_embeddings_and_output_weights=True` (Qwen3-0.6B)
- Quant config: NVFP4 defaults; crash is independent of format

babyplutokurt:
# Skip weight quantizers with no stored weight in `fold_weight`

### What does this PR do?

Type of change: Bug fix

Fixes #2131.

On Megatron-Core models with tied word embeddings
(`share_embeddings_and_output_weights=True`), the `output_layer` is built with
`skip_weight_param_allocation`: it stores `weight = None` and borrows the embedding
weight at forward time, while ModelOpt still attaches a `weight_quantizer` to it.
`QuantModule.fold_weight` matches the pair on the `*_weight_quantizer` attribute name
and `fake_quant` alone, then dereferences `weight.data`, so `mtq.fold_weight` crashes:

```
modelopt/torch/quantization/nn/modules/quant_module.py:145, in _fold_weight_quantizer
    weight.data.copy_(quantizer(weight.float().contiguous()).to(weight.dtype))
AttributeError: 'NoneType' object has no attribute 'data'
```

This PR skips pairs whose weight attribute is not a tensor and leaves their quantizer
untouched: there is nothing stored to fold, and the shared weight keeps being
quantized at forward time through the still-enabled quantizer (the embedding module
folds its own stored weight as before). HF models are unaffected — they express tying
as a shared tensor rather than `None`, which is why the crash only surfaced on
Megatron.

### Usage

```python
from megatron.core.models.gpt import GPTModel

# Any GPTModel with tied embeddings, e.g. Qwen3-0.6B:
# output_layer.weight is None (borrowed from the embedding at forward time).
model = GPTModel(..., share_embeddings_and_output_weights=True)
model = mtq.quantize(model, mtq.NVFP4_DEFAULT_CFG, forward_loop)

mtq.fold_weight(model)  # previously: AttributeError; now folds all stored weights
```

### Testing

- New unit test `test_fold_weight_skips_none_weight` in
  `tests/unit/torch/quantization/test_tensor_quant_cpu.py` builds the tied-layer shape
  (a converted `QuantLinear` with `weight = None`) — it reproduces the exact
  `AttributeError` at `quant_module.py:145` without the fix and passes with it,
  asserting the skipped quantizer stays enabled with its `_amax` intact.

- Is this change backward compatible?: ✅ (previously-crashing calls now succeed; folding behavior for stored weights is unchanged)
- Did you get Claude approval on this PR?: ❌ (external contribution — cannot self-trigger `/claude review`)

### Additional Information

Related issue: #2131. Downstream workaround that this fix would let us drop:
NVIDIA-NeMo/RL#3441 (`nemo_rl/modelopt/models/policy/workers/weight_folding.py`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed weight folding crashes for quantized models with missing or non-tensor weights.
  * Such weights are now safely skipped without changing their associated quantizer settings.
  * Normal tensor weights continue to fold as expected.

* **Tests**
  * Added regression coverage to verify quantizers and calibration values remain unchanged when weights are unavailable.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,7 @@ Changelog
 
 **Bug Fixes**
 
+- Fix ``mtq.fold_weight`` crashing with ``AttributeError: 'NoneType' object has no attribute 'data'`` on Megatron-Core models with tied word embeddings: the tied ``output_layer`` (built with ``skip_weight_param_allocation``) stores ``weight = None`` and borrows the embedding weight at forward time, yet still carries a ``weight_quantizer``. Weight-quantizer pairs whose weight is not a stored tensor are now skipped and left untouched.
 - Fix EAGLE-3 training with context parallelism (``--cp_size > 1`` in ``examples/speculative_decoding``), which failed to start on ``accelerate >= 1.13`` and then raised ``got mixed torch.Tensor and DTensor``.
 
 0.46 (2026-08-17)
--- a/modelopt/torch/quantization/nn/modules/quant_module.py
+++ b/modelopt/torch/quantization/nn/modules/quant_module.py
@@ -157,6 +157,14 @@ class QuantModule(DynamicModule):
         transform is baked into the stored weight and then disabled, so subsequent forwards use
         the stored weight directly. Calibration buffers (``_pre_quant_scale``, ``_amax``) are
         dropped unless ``keep_attrs``.
+
+        Quantizers whose weight attribute is not a tensor are skipped and left untouched:
+        modules such as Megatron tied-embedding output layers (built with
+        ``skip_weight_param_allocation``) store ``weight = None`` and receive the shared
+        weight as a forward-time argument, so there is nothing stored to fold. Such quantizers
+        stay enabled after folding, so callers that assert all weight quantizers are disabled
+        once folding completes (e.g. ``_check_all_weight_quantizers_disabled`` in the vLLM
+        fakequant export plugin) must account for them.
         """
         # Handle all attributes that end with _weight_quantizer
         for name in dir(self):
@@ -173,6 +181,10 @@ class QuantModule(DynamicModule):
                     f"{name} doesn't have a corresponding {weight_name} in {self.__class__.__name__}"
                 )
                 weight = getattr(self, weight_name)
+                if not isinstance(weight, torch.Tensor):
+                    # e.g. Megatron tied-embedding output_layer: weight is None and
+                    # borrowed at forward time, so there is nothing stored to fold.
+                    continue
                 self._fold_weight_quantizer(attr, (weight,), keep_attrs)
 
 
--- a/tests/unit/torch/quantization/test_tensor_quant_cpu.py
+++ b/tests/unit/torch/quantization/test_tensor_quant_cpu.py
@@ -332,6 +332,25 @@ def test_fold_weight_keep_attrs_keeps_amax(monkeypatch):
         unregister_quant_backend(backend_name)
 
 
+def test_fold_weight_skips_none_weight():
+    """A weight quantizer with no stored weight is skipped instead of crashing.
+
+    Megatron-Core tied-embedding output layers are built with
+    ``skip_weight_param_allocation``: the module stores ``weight = None`` and borrows the
+    embedding weight at forward time, while still carrying a ``weight_quantizer``.
+    ``fold_weight`` must skip the pair, leaving the quantizer intact for forward-time use.
+    """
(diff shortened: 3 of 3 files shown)


cjluo-nv (approved):

Straightforward, well-scoped fix. `fold_weight` now skips missing/non-tensor stored weights before dereferencing them, preserving the forward-time quantizer needed by Megatron tied-output layers. The regression test reproduces the `weight=None` shape and verifies the quantizer and calibration state remain intact, and the changelog entry is under the current development release.

kevalmorabia97:
/ok to test [44e0d34](https://github.com/NVIDIA/Model-Optimizer/commit/44e0d349452c7546b20cbb0d6fa5d8fe900677d8)

babyplutokurt:
Added the docstring suggestion from claude.

Also rebased on latest main already and re-push

@kevalmorabia97 ^

babyplutokurt:
Added the docstring suggestion from claude.

Also rebased on latest main already and re-push

@kevalmorabia97 ^

kevalmorabia97:
@babyplutokurt can you please resolve merge conflicts

babyplutokurt:
@kevalmorabia97 I think @mxinO [https://github.com/NVIDIA/Model-Optimizer/pull/2140](url) already covers this PR and we can close it.

right now [https://github.com/NVIDIA-NeMo/RL/pull/3441](url) is waitting for modelOpt Pin updated in nemo_RL. 

Can you correct me if I am wrong about the status? @mxinO

PR #2136, opened 2026-08-10
feat(export): export each decoder layer as layerwise calibration finishes it

Fridah-nv:
### What does this PR do?

Type of change: new feature

Layerwise calibration can already resume, but only through a full-precision scratch checkpoint, and a completed run still pays for a second whole-model export pass over it.

`layerwise.export_dir` writes each decoder layer to its own quantized shard as soon as calibration finishes with it, so the directory is a complete, loadable checkpoint when the last layer lands and `export_hf_checkpoint()` is skipped. The shards *are* the resume artifact: a restarted run reuses layers already on disk instead of recalibrating and re-exporting them, so no full-precision copy of the model accumulates. The resume directory beside it holds only the current boundary's cached activations and the per-layer output shapes.

Setting the config field is the whole switch — no CLI flag. `hf_ptq.py` rewrites its value to `--export_path`, and derives the resume directory (`<export_path>.layerwise_resume`) when you haven't chosen one.

One shard per layer is what makes resume safe: shards are written whole and named from the layer index, so a crash can lose the layer in flight but never corrupt an earlier one, and a re-run overwrites in place.

Because a resumed run never recalibrates the layers it skipped, **the in-memory model is not valid for inference afterwards**; the field implies `--skip_generate`.

Includes a pre-existing `main` fix this depends on: `_is_layerwise` used `getattr` on an algorithm that YAML parses as a **dict**, so it answered `False` for every layerwise recipe in the repo and the batch-size probe it gates was never skipped. **Behaviour change:** `--batch_size 0` now yields `batch_size=1` for layerwise recipes, as its comment intends. Detection also now scans every algorithm entry rather than the first, so a list-form recipe whose `layerwise` block is not first is recognised as layerwise — same batch-size consequence. Nothing else on the non-fused paths changes: `FUSION_FREE_FORMATS` is the exact set the inline list held, `save_non_weight_artifacts` is a lift of the streaming exporter's own block, and the calibration-loop changes are gated on an exporter being present.

**Refused before calibration starts**, since each would otherwise produce a silently different checkpoint rather than fail:

| Refused | Why |
|---|---|
| AWQ / SVDQuant | need pre-quant-scale steps that are still whole-model |
| Weight-tied quantized modules | `sync_tied_input_amax` merges amaxes across a partner that may be uncalibrated or already written |
| Multi-process (FSDP2) | every rank would write the same shards |
| Multimodal (VLM) | calibration runs on the extracted language model |
| MTP models | exclusions applied after calibration has written everything |
| AutoQuantize recipes | only the mono-quantize path retargets `export_dir` |
| Spec-dec, `--vllm_fakequant_export`, non-dense sparsity, `int8_smoothquant`, encoder-decoder `model_type` | each routes to a second exporter that would overwrite `--export_path` |
| `export_dir` on more than one algorithm entry, or on any but the last | export finalizes shards as calibration walks the layers, so a later pass would change the model after its checkpoint was written |

Shards are also bound to the run that produced them (`.layerwise_export.json`: model class, layer count, formats, KV-cache format, and a digest of the resolved quant config), so one run's manifest cannot finalize another's shards. Source weights are not digested — that would mean reading the whole model — so differently-trained weights at the same path compare equal.

### Why a separate exporter

Three reuse paths were considered before adding one:

- **Extend `_StreamingShardWriter`.** It buffers by `max_shard_size` into `__shard_part_*` temp names and renames to canonical names only in `finalize()`, once the shard count is known. The resume invariant needs the opposite: a stable `model-layer-00007.safetensors` committed when layer 7 finishes, so "shard exists" means "layer done" across a restart. Forcing a per-layer flush still leaves temp names, finalize-time renaming, and an in-memory `_key_to_part` — every method would change.
- **Keep the layerwise checkpoint and run the streaming exporter at the end.** This works, and it is why the pitch above is *not* durability: that already exists. What it leaves is a second whole-model pass owed *after* calibration finishes — itself needing a GPU session — where per-layer export makes the last calibrated layer also the last exported one. Scratch size only separates them for weight-mutating calibrators: `save_layer_state` is off under per-layer export, but with `calib_mutates_weights: false` (the shipped recipe) the checkpoint holds just amax buffers either way.
- **Factor a shared per-module writer around `ExportContext`.** The right long-term shape, but it touches all three existing export paths; doing it here makes this change larger, not smaller.

One deliberate divergence from `_StreamingShardWriter` worth knowing about: it clones tensors that share storage, this path lets `save_file` raise instead. No model was found where the clone fires, and copying unattributed aliases can hide a real bug rather than surface it. If a checkpoint ever trips it, that is information we want.

Happy to take a different call on this — flagging it for maintainer sign-off rather than assuming it.

### Usage

```bash
python examples/hf_ptq/hf_ptq.py --pyt_ckpt_path <model> --export_path <out> \
    --recipe modelopt_recipes/general/ptq/nvfp4_experts_only-kv_fp8_layerwise_export.yaml
```

Interrupt and rerun the same command: calibration resumes from the last committed layer, finished shards are reused, and a run that had already finished every layer only re-runs `finalize()`.

```yaml
quantize:
  algorithm:
    method: max
    layerwise:
      enable: true
      calib_mutates_weights: false
      export_dir: /tmp/modelopt_layerwise_export   # presence is the switch; value replaced with --export_path
      # checkpoint_dir omitted -> derived as <export_path>.layerwise_resume
```

### Testing

Each row exports the same calibration two ways — per-layer, and whole-model via `export_hf_checkpoint()` — and compares them **tensor for tensor and config for config**.

The 35B row was re-run on the current head, against a baseline built from a `main` worktree rather than from this branch, so it covers both "per-layer differs from whole-model" and "this branch broke the shared whole-model path". The other rows date from earlier heads; the code they exercise is unchanged, but they are not fresh runs.

| Model | Config | Result |
|---|---|---|
| Qwen3.6-35B-A3B (40 layers, 256 fused experts) | NVFP4 W4A4 experts-only + FP8 KV | 123,513 tensors, 0 mismatched; `config.json`, `hf_quant_config.json`, `generation_config.json` all identical |
| Qwen3-30B-A3B (48 layers, 128 per-expert linears) | NVFP4 experts (`nvfp4_static` weights) + `mse`, offload | 74,163 tensors, 0 mismatched |
| Qwen3-30B-A3B | same, `SIGKILL` after 25/48 layers, then resumed | 74,163 tensors, 0 mismatched **vs the uninterrupted run** |
| Llama-3.1-8B-Instruct | FP8 dense + FP8 KV, resident | 803 tensors, 0 mismatched |

**Refusals verified on real checkpoints**, each writing **zero shards** and never reaching calibration — the "refused before calibration starts" claim above, demonstrated rather than asserted: multimodal and MTP (Qwen3.6-35B, the MTP case on a text-only view since the multimodal gate fires first), tied embeddings (Qwen3-0.6B), and multi-process (2-rank `torchrun --use_fsdp2`, Llama-3.1-8B).

**Served, not just compared.** Under vLLM 0.27.1 (Marlin NVFP4 kernels, SM 8.9): the 30B checkpoint exported three ways — whole-model, per-layer, per-layer-resumed-after-a-kill — and the 8B exported both ways all load and produce **identical greedy generations, 4/4 prompts** within each model.

**Index integrity** on every checkpoint above: each `weight_map` key resolves to the shard actually holding it; 0 missing, 0 extra, 0 mis-routed. Tensor equality alone never exercises that, and it is the one artifact per-layer export builds differently.

Resume state stays bounded: **332 KB beside 22 GB** of shards on the 35B, **396 KB beside 19 GB** on the 30B — the committed boundary's activations only, not one set per layer.

Not covered: the `trust_remote_code` `*.py` copy path. Nemotron-Nano-12B-v2-Base fails with a CUDA illegal memory access on these cards, on the whole-model baseline too, so it is an environment limit rather than a result.

**Comparing the configs is new, and it caught a real bug.** `get_quant_config` reports on the quantizer modules, which `export_layer` replaces as it goes, so reading it in `finalize()` described a model with no quantizers left: the checkpoint advertised `quant_algo: null` while its weights were packed NVFP4, and under the shipped experts-only recipe `hf_quant_config.json` was not written at all. It is snapshotted in `__init__` now, beside the kv-cache format already captured there — which is why that one field was correct while the rest were not. Uniform FP8 and NVFP4 hid it because their configs survive the conversion; only a mixed model loses its algo, and mixed is what every shipped layerwise-export recipe is. Reverting the fix fails `test_moe_export_matches` and passes the ten uniform-format cases, matching what the 35B shows.

**24 GPU tests** in `tests/gpu/torch/export/test_layerwise_export.py`. The equivalence oracle is a cross-product: {FP8, NVFP4, NVFP4 + `get_qdq_activations_from_prev_layer`, mixed FP8/NVFP4, KV-cache} × {fresh, resumed-after-interruption}, each compared tensor-for-tensor against `export_hf_checkpoint`. Plus MoE export; resume fail-fast; resume artifacts replaced and pruned; complete-manifest finalize-only; shards-without-manifest refusal; shards-from-a-different-run refusal (format and module selection); identity-without-shards does not block a rerun; export-does-not-mutate-the-model; index routes every key to the shard holding it; AWQ refusal (from config, and after calibration); export-without-`checkpoint_dir`.

**Unit tests** in `tests/examples/hf_ptq/test_example_utils.py` cover the list-valued `algorithm` shapes: which entry owns export, whose `checkpoint_dir` is derived, per-entry resume bases, both ambiguity refusals, and the recipe shapes `recipe_layerwise_blocks` normalizes (dict, list order, config object, and the empty cases).

`tests/gpu/torch/export/` 150 passed / 2 skipped (pre-existing env skips) · `tests/unit/recipe` 284 · `tests/unit/torch/export` 186 · `test_layerwise_calibrate` 33 · `test_example_utils` 42 · pre-commit clean.

Also verified: the exported directory reloads through `AutoModelForCausalLM` and runs a forward.

Not a speed win: per-layer export was slower than the streaming export in one offload pairing (271s vs 208s, the per-layer fusion probe), though those runs shared GPUs so the magnitude is not cleanly measured.

- Is this change backward compatible?: ✅ — `export_dir` defaults to `None`; existing paths unchanged when unset, except the batch-size change noted above.
- Did you get Claude approval on this PR?: ❌ — not yet; draft.

### Additional Information

**Pre-existing bug found on the way, not fixed here.** Layerwise calibration leaves `self_attn.o_proj`'s input amax at `0.0` on every layer but the last, so a full-NVFP4 layerwise model cannot be exported by *any* path. `get_qdq_activations_from_prev_layer=True` avoids it, pinning the cause to the pre-`calib_func` capture pass — which also explains why only the last layer, the one that skips it, is correct. That combination now works with per-layer export (it asserted on layer 0 until review caught it). Hidden until now because the shipped NVFP4 layerwise recipes are experts-only; the NVFP4 tests here exclude `o_proj` for the same reason. Deserves its own issue.


modelopt/torch/quantization/model_calib.py
```diff
@@ -2090,13 +2099,32 @@ def layerwise_calibrate(
...
+        exporter = LayerwiseExporter(model, export_dir)
+
     ckpt = _CheckpointState.from_folder(
         checkpoint_dir,
         num_layers,
         save_every=save_every,
         calib_mutates_weights=calib_mutates_weights,
+        save_layer_state=exporter is None,
     )
     start_layer = ckpt.start_layer if ckpt else 0
```
cjluo-nv, line 2104: > _Bot comment._

`_CheckpointState.from_folder()` relies on `detect_resume_point()`, which deliberately returns `None` when `last_completed_layer + 1 >= num_layers`; this makes `start_layer` become 0 for a fully committed manifest. There is a real crash window here: `export_layer(last)` writes the last shard, then `ckpt.save(last, ...)` commits the complete manifest, and only afterwards `exporter.finalize()` writes the tail/index/config. If the process dies in between, rerunning recalibrates and rewrites every layer instead of reusing all shards/finalizing them. A normal rerun after completion does the same. Please preserve `start_layer == num_layers` for the export mode and support a finalize-only resume (without indexing `transformer_layers[start_layer]`), with a regression test simulating a complete manifest but absent tail/index.

cjluo-nv (commented):

The feature addresses a real durability problem: avoiding loss of already-calibrated decoder layers and the second whole-model export pass. Existing alternatives are (1) extending/reusing the existing streaming exporter and `_StreamingShardWriter` in `modelopt/torch/export/unified_export_hf_streaming.py`, (2) retaining the existing layerwise checkpoint artifacts and invoking that streaming exporter as the final/resume phase, or (3) factoring a shared per-module shard writer around the existing `ExportContext`/export-handler path. The PR body acknowledges the streaming exporter and its tail-pass duplication, but does not justify why a separate 549-line `LayerwiseExporter` is preferable or why the existing writer cannot be extended; this remains an architectural concern for a 1,286-line PR. More importantly, I found two correctness issues in the core path: the hf_ptq integration calls an undefined helper, and a completed manifest is treated as a fresh run, leaving a crash window after the last layer that defeats the durability claim. New-file license headers match `LICENSE_HEADER`, and the GPU equivalence/resume tests are useful, but they do not cover either failure below.



---

- `examples/hf_ptq/example_utils.py:1193` — > _Bot comment._

`_layerwise_checkpoint_dir_location` is not defined or imported anywhere in this file/repository. Consequently the documented `hf_ptq.py` path reaches `colocate_layerwise_checkpoint_dir()` and raises `NameError` whenever the layerwise config has a checkpoint directory (including the new shipped recipe). Please implement/reuse the intended lookup and add an example-utils or hf_ptq integration test, since the direct `mtq.quantize` GPU tests bypass this code.

modelopt/torch/quantization/model_calib.py
Fridah-nv: Confirmed, and thanks for catching this.  
`layerwise_calibrate` now detects a complete manifest in export mode before any layer patching, asserts the shards are all present, calls finalize() and returns — so no `transformer_layers[start_layer]` indexing, as you asked. Regression test added:

examples/hf_ptq/hf_ptq.py
```diff
@@ -1128,17 +1195,43 @@ def quantize_main(
...
+    if args.layerwise_export:
+        if isinstance(recipe, ModelOptAutoQuantizeRecipe):
+            # Only the mono-quantize path retargets export_dir and runs the refusals;
+            # auto_quantize would export to the placeholder and skip the real export.
+            raise NotImplementedError(
+                "layerwise.export_dir is not supported with an AutoQuantize recipe; "
+                "use a PTQ recipe, or drop export_dir and export afterwards."
+            )
+        # A resumed run leaves skipped layers without amax; only the checkpoint is whole.
+        args.skip_generate = True
```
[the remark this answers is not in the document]
Fridah-nv: This can't occur: `MaxCalibConfig(layerwise={"enable": False, "export_dir": "/tmp/x"})` raises `ValidationError`. `validate_layerwise_checkpoint_dir` in `config.py` already rejects `export_dir` without `enable=True`, so the recipe fails to load and the empty-export-directory path is unreachable.

examples/hf_ptq/example_utils.py
```diff
@@ -1142,6 +1142,36 @@ def _layerwise_checkpoint_dir(algorithm) -> str | None:
...
+    That is not hypothetical: the shipped recipes default ``checkpoint_dir`` to
+    ``/tmp/modelopt_layerwise_ckpt``, and ``/tmp`` is container-local. A run that outlasts
+    its GPU session -- the case this whole path exists for -- comes back to a wiped
+    manifest and redoes everything.
+
+    Co-locating the two makes the invariant structural rather than a thing the user has to
+    know. Only applied when per-layer export is on; without it ``checkpoint_dir`` is an
+    ordinary resume directory whose placement is the caller's business.
+    """
+    current = _layerwise_checkpoint_dir(quant_cfg.get("algorithm"))
```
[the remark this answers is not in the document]
Fridah-nv: Rewritten against the current API, which also drops the dead legacy-flat branch in  5b458976ab

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+@contextlib.contextmanager
+def transient_module_state(module: nn.Module):
+    """Undo everything export does to ``module``, so calibration can continue through it.
+
+    A resident model has no materialization window to discard the packed weights, grafted
+    expert submodules and new buffers export leaves behind. Restoring the dicts suffices,
+    and costs references not a deep copy, since export rebinds rather than mutates.
+    """
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 24ecfbb8f7.  Re-verified with NVFP4 test with 0 differences, exports still byte-identical, with an NVFP4 regression test.

examples/hf_ptq/example_utils.py
```diff
@@ -1173,6 +1203,22 @@ def resolve_checkpoint_dir(quant_cfg: dict, model_path: str) -> tuple[dict, str]
...
+    # Detection accepts one algorithm or a list, so the retarget must too.
+    for entry in algorithm if isinstance(algorithm, list) else [algorithm]:
+        if isinstance(entry, dict) and isinstance(entry.get("layerwise"), dict):
+            entry["layerwise"]["export_dir"] = export_path
+    return quant_cfg
+
+
 def add_mlflow_args(parser: argparse.ArgumentParser) -> None:
     """Add the MLflow tracking flags."""
     parser.add_argument(
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 39c80d7676. `set_layerwise_export_dir` now counts what it retargets and raises `ValueError` naming the algorithm it saw when the count is zero. Verified across four shapes: the valid case retargets; `layerwise-without-export_dir`,` no-layerwise-block`, and a list containing neither all raise.
The other half of your comment — not writing export_dir into entries that never opted in — was fixed earlier in 24ecfbb8f7.

CHANGELOG.rst
```diff
@@ -15,6 +15,7 @@ Changelog
...
+- Add ``layerwise.export_dir`` (``--layerwise_export`` in ``examples/hf_ptq/hf_ptq.py``), which writes each decoder layer to its own quantized HF checkpoint shard as soon as layerwise calibration finishes it, so the directory is already a complete checkpoint when the last layer lands and the separate ``export_hf_checkpoint()`` pass over a full-precision intermediate is skipped. Combined with ``layerwise.checkpoint_dir`` the shards double as the resume artifact, so a run interrupted mid-model restarts without recalibrating or re-exporting finished layers; the output is byte-identical to the whole-model export. Supports FP8 and NVFP4 on single-process models, resident or accelerate-offloaded; fusing formats work because each layer rediscovers its own q/k/v and gate/up scale-fusion groups. AWQ and SVDQuant (whose pre-quant-scale steps are still whole-model), multi-process jobs, weight-tied quantized modules, multimodal models, MTP models and speculative decoding are refused with ``NotImplementedError`` before calibration starts, and since a resumed run never recalibrates the layers it skipped the field implies ``--skip_generate``.
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 39c80d7676

examples/hf_ptq/example_utils.py
```diff
@@ -1142,6 +1142,39 @@ def _layerwise_checkpoint_dir(algorithm) -> str | None:
...
+    ``resume`` rather than ``checkpoint``: with per-layer export on the shards *are* the
+    weights, so this directory holds no checkpoint -- only the resume point, the next
+    layer's cached activations, and the shape metadata skip mode needs.
+    """
+    if _layerwise_checkpoint_dir(quant_cfg.get("algorithm")) is not None:
+        return quant_cfg, False
+
+    algorithm = quant_cfg.get("algorithm")
+    if not isinstance(algorithm, dict) or not isinstance(algorithm.get("layerwise"), dict):
+        return quant_cfg, False
```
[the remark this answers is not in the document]
Fridah-nv: Good catch, fixed in c5f9e384ba.

modelopt_recipes/general/ptq/nvfp4_experts_only-kv_fp8_layerwise_export.yaml
```diff
@@ -0,0 +1,63 @@
...
+    NVFP4 static weight and dynamic activation for expert layers only (W4A4), FP8 KV cache,
+    max layerwise calibration, exporting each decoder layer to a quantized checkpoint shard
+    as soon as it is calibrated. Setting layerwise.export_dir is what enables that; hf_ptq.py
+    rewrites its value to --export_path, since the destination is per-run.
+
+    An interrupted run then resumes without recalibrating or re-exporting finished layers,
+    which is the point for a PTQ run that outlasts its GPU session. The resume state lives
+    beside the checkpoint at <export_path>.layerwise_resume unless you set
+    layerwise.checkpoint_dir yourself. Export rediscovers the q/k/v and gate/up scale-fusion
+    groups per layer, so the checkpoint matches a whole-model export byte for byte.
```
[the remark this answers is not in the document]
Fridah-nv: the wording is fixed in c5f9e384ba.

examples/hf_ptq/example_utils.py
```diff
@@ -1142,6 +1142,36 @@ def _layerwise_checkpoint_dir(algorithm) -> str | None:
...
+
+    Co-locating the two makes the invariant structural rather than a thing the user has to
+    know. Only applied when per-layer export is on; without it ``checkpoint_dir`` is an
+    ordinary resume directory whose placement is the caller's business.
+    """
+    current = _layerwise_checkpoint_dir(quant_cfg.get("algorithm"))
+    target = os.path.join(export_path, ".layerwise_checkpoint")
+    if current is None or current == target:
+        return quant_cfg, False
+
```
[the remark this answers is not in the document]
Fridah-nv: Agreed, and this is fixed in 90aa931774 — I took your second option.
The resume state is now a sibling: `<export_path>.layerwise_resume`, never inside the published directory. It still shares storage and lifetime with the shards, which is what that docstring was actually arguing for.

examples/hf_ptq/hf_ptq.py
```diff
@@ -1128,17 +1200,43 @@ def quantize_main(
...
+        if isinstance(obj, dict):
+            return obj.get("layerwise")
+        return getattr(obj, "layerwise", None)
+
+    def _layerwise_get(cfg, key, default=None):
+        if cfg is None:
+            return default
+        return cfg.get(key, default) if isinstance(cfg, dict) else getattr(cfg, key, default)
+
+    layerwise_cfg = _layerwise_cfg(recipe)
```
cjluo-nv: > _Bot comment._

`_layerwise_cfg()` returns the first `layerwise` block in a list, but `QuantizeAlgoCfgType` explicitly supports sequential algorithm lists. This makes export order unsafe. For example, if the first algorithm has `export_dir` and a later algorithm calibrates the model without it, the first pass finalizes the shards, the later pass changes the model, and `args.layerwise_export=True` causes `export_quantized()` to skip the normal final export—silently leaving a checkpoint for the intermediate state. In the reverse order, this code misses the opt-in and leaves the exporting pass pointed at its placeholder. Please either reject `export_dir` with multi-algorithm lists, or inspect all entries and enforce that exactly the final calibration pass owns export; add an integration/unit regression for both orderings.

cjluo-nv (commented):

The durability problem is real: preserve completed layer calibration across GPU-session loss and avoid a second whole-model export pass. Existing alternatives remain (1) extending `_StreamingShardWriter` in `unified_export_hf_streaming.py`, (2) using the existing layerwise checkpoint as the durable calibration artifact and running the streaming exporter on completion/resume, or (3) factoring a shared per-module writer around `ExportContext`; the installed stack already includes `safetensors`/Transformers, so no new dependency is needed. The PR body now explicitly acknowledges the parallel-exporter question and tail-pass duplication, but still does not explain why a separate 601-line `LayerwiseExporter` is preferable to those reuse paths. This architectural choice needs maintainer sign-off, especially at 1,460 changed lines.

Re-review: the earlier critical `NameError` and completed-manifest crash window are fixed, and the latter now has a useful finalize-only regression test. The stale scratch-directory placement, config identity, retarget no-op, INT8 token, changelog, and mutation concerns were also addressed. However, list-valued calibration algorithms still have correctness holes: export detection observes only the first layerwise block, and default checkpoint derivation can bind the wrong blocks or fail to bind the exporting block. This can silently leave the checkpoint at an intermediate calibration state while `hf_ptq` skips the final export. There are still no committed unit tests for these pure `example_utils` list transformations. New source headers match `LICENSE_HEADER`.

examples/hf_ptq/example_utils.py
```diff
@@ -1134,12 +1134,38 @@ def copy_custom_model_files(
...
+    would ship in the checkpoint. An explicit path is left alone.
+    """
+    if _layerwise_checkpoint_dir(quant_cfg.get("algorithm")) is not None:
+        return quant_cfg, False
+
+    quant_cfg = copy.deepcopy(quant_cfg)
+    blocks = _layerwise_blocks(quant_cfg.get("algorithm"))
+    for block in blocks:
+        block["checkpoint_dir"] = export_path.rstrip("/") + ".layerwise_resume"
+    return quant_cfg, bool(blocks)
```

This list handling still associates the resume path incorrectly. If block A has an explicit `checkpoint_dir` but no export and block B has `export_dir` with no checkpoint, `_layerwise_checkpoint_dir()` finds A and this returns unchanged, so B gets no resume manifest. If none has an explicit path, this assigns the same checkpoint directory to *every* layerwise algorithm, including entries that did not opt into export; sequential calibration passes can then overwrite/read one another's manifest. Derive the sibling path only for block(s) whose `export_dir` is set, preserve unrelated blocks, and reject ambiguous multiple exporting blocks. Please commit unit tests for these list shapes—the current diff has no tests for these helpers.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2171,6 +2210,24 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+                        cache = kwargs_input.get("past_key_values")
+                        if cache is not None:
+                            kwargs_input = dict(kwargs_input)
+                            if hasattr(cache, "reset"):
+                                cache.reset()
+                            else:
+                                kwargs_input["past_key_values"] = None
+                        m(*args, **kwargs_input)
+
+                    exporter.export_layer(layer_idx, layer, _fusion_probe)
```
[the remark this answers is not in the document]
Fridah-nv: Confirmed and fixed in 4074508793

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,601 @@
...
+                differing = sorted(
+                    k for k in set(previous) | set(identity) if previous.get(k) != identity.get(k)
+                )
+                raise RuntimeError(
+                    f"{self._export_dir} holds shards from a different run (differing: "
+                    f"{differing}); resuming would finalize them against this run's "
+                    "manifest. Use a fresh export directory."
+                )
+        else:
+            path.write_text(json.dumps(identity, indent=2))
```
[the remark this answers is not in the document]
Fridah-nv: Agreed, fixed in 4074508793. Gated the refusal on `completed_layers() `as you proposed.

examples/hf_ptq/hf_ptq.py
Fridah-nv: Right on both orderings — fixed in 0d2e78b41c. I took your second option: `layerwise_export_block()` inspects every entry and enforces that exactly one sets export_dir and that it is the last, since export finalizes shards as calibration walks the layers and any later pass would change the model afterwards.

examples/hf_ptq/example_utils.py
Fridah-nv: Both hazards are real and both are fixed in 0d2e78b41c

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,602 @@
...
+
+    Each case would otherwise differ from a whole-model export without failing. The format
+    gate is the central one: this path never runs ``requantize_resmooth_fused_llm_layers``,
+    so it is limited to formats whose scales it can reproduce per layer.
+
+    Supporting AWQ and SVDQuant needs that pass's pre-quant-scale steps made per-layer.
+    """
+    from modelopt.torch.utils import distributed as dist
+
+    unsupported = sorted(str(f) for f in _module_formats(model) - SUPPORTED_FORMATS)
```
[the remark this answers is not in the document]
Fridah-nv: Fixed by re-running the format check on the first exported layer, which catches any format only visible post-calibration rather than blacklisting AWQ by name. Regression test added.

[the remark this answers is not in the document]
Fridah-nv: But its second failure mode is wrong, and I'm rejecting it. It claims _fuse_shared_inputs → _update_pre_quant_scale corrupts weights for the remaining layers because "transient_module_state deliberately does not snapshot parameters." It does snapshot them — dict(m._parameters); the docstring says parameters aren't clone-restored, which the bot read as not captured. And _update_pre_quant_scale rebinds (module.weight = nn.Parameter(...)), which the dict restore undoes. Measured directly: layerwise-AWQ with export vs without gives MODEL_DRIFT: 0 — every parameter identical. The gate bug is real; the corruption claim is not.

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,602 @@
...
+                continue
+            if _holds_meta_tensor(module):
+                # No window was offered yet the weights are absent. Packing would raise
+                # deep in the handler, skipping would drop it silently.
+                raise RuntimeError(
+                    f"{name!r} holds meta tensors but was not offered a materialization "
+                    "window, so its weights cannot be exported. Export without export_dir "
+                    "and use export_hf_checkpoint() for this model."
+                )
+            _dispatch_export_handler(name, module, self._ctx)
```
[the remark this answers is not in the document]
Fridah-nv: This is valid, took the second fix

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,602 @@
...
+        """Tie the shards to the run that produced them.
+
+        The manifest records no model or quantization identity and
+        :meth:`assert_shards_present` only checks existence, so one run's manifest could
+        otherwise finalize another's shards into a valid-looking, wrong checkpoint.
+        """
+        model = self._ctx.model
+        # The quant config carries the per-module contract, so two runs differing in which
+        # modules are quantized hash differently even when the format names match.
+        quant_contract = hashlib.sha256(
```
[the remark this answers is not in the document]
Fridah-nv: accurate but partially accepted. Shared the duplicate get_quant_config (clean, no contract change). Skipped sharing _module_formats: it spans a module-level validator and a method, so it'd mean making a validator return data — muddier contract for a gain the bot itself calls "seconds against a multi-hour calibration."

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,632 @@
...
+        if probe_forward is None:
+            raise RuntimeError(
+                f"layer format {layer_format!r} needs input-sharing groups to fuse its "
+                "scales, but no probe_forward was supplied to rediscover them."
+            )
+
+        input_to_linear, _ = collect_shared_input_modules(
+            layer_module, lambda: probe_forward(layer_module)
+        )
+        _fuse_shared_input_modules(
```
Edwardf0t1, line 289: **Per-layer fusion drops the whole-model path's expert-name propagation.**

`requantize_resmooth_fused_llm_layers` uses the `fused_linears` returned by `_fuse_shared_input_modules` and, for any group matching `experts?\.\d+`, replays `preprocess_linear_fusion` over *every* sibling expert (`unified_export_hf.py:518-541`) — explicitly because "the dummy forward may not be able to activate all the experts." Here the return value is discarded and there is no equivalent, so only the experts the probe batch routed tokens to get fused. Layerwise also forces `batch_size=1`, which makes the probe maximally narrow.

`sync_moe_gate_up_amax` is a partial safety net: it max-merges gate/up `weight_quantizer.amax`, but not `NVFP4StaticQuantizer.global_amax`, which `preprocess_linear_fusion` unifies separately (`quant_utils.py:1527-1536`).

So an NVFP4 experts-only layerwise-export run with a static weight quantizer (MSE / local hessian) leaves gate/up of any expert the probe missed with different `weight_scale_2`, and an engine that fuses them into `gate_up_proj` picks up a wrong scale for those experts. The equivalence testing in the PR description used `max` calibration, where the gate/up merge does cover this — which would explain the 0-mismatched-tensor result not catching it.

examples/hf_ptq/hf_ptq.py
```diff
@@ -768,6 +770,65 @@ def mono_quantize(
...
+    """
+    if is_multimodal_model(full_model):
+        raise NotImplementedError(
+            "layerwise.export_dir does not support multimodal models: calibration runs on the "
+            "extracted language model, so the shards and config.json would describe that "
+            "submodel rather than the full VLM, and the VLM export path would then "
+            "overwrite config.json with the unquantized source config."
+        )
+
+    if mtp_layer_prefixes:
```
Edwardf0t1, line 789: **This MTP refusal never fires on the normal (non-FSDP2) load path.**

`assert_layerwise_export_compatible` is called with `mtp_layer_prefixes = getattr(full_model, "_mtp_layer_prefixes", None)`, but that attribute is only set before quantization in the FSDP2 branch. On the normal load path it is set inside `export_quantized` via `load_mtp_weights` — i.e. after calibration has run.

For a GLM-4.7-style MTP checkpoint loaded normally, the "refused before calibration starts" contract therefore doesn't hold: the run pays for full calibration and writes every shard, the index and the configs to `--export_path`, and only then hits the later `if mtp_state_dict: raise`. It does fail loudly, but it leaves behind a directory that loads fine and silently lacks the MTP weights — which is the outcome this check exists to prevent, just deferred to the end of the run.

`mtp_layer_prefixes_from_checkpoint(args.pyt_ckpt_path)` is already imported for the FSDP2 branch and would make this pre-check real.

Edwardf0t1 (commented):
Two findings from a review of the layerwise export path. Both are about the per-layer path diverging from the whole-model path it replaces; the rest of the design (the `transient_module_state` revert, the resume/manifest interlocks, the per-layer MoE ordering) checks out.

examples/hf_ptq/example_utils.py
```diff
@@ -1134,12 +1134,70 @@ def copy_custom_model_files(
...
+    A sibling, not a child: nothing deletes the resume state, so inside ``export_path`` it
+    would ship in the checkpoint. An explicit path is left alone.
+    """
+    quant_cfg = copy.deepcopy(quant_cfg)
+    # The exporting block specifically: another pass's explicit checkpoint_dir says nothing
+    # about where this one resumes from.
+    block = layerwise_export_block(quant_cfg.get("algorithm"))
+    if block is None or block.get("checkpoint_dir") is not None:
+        return quant_cfg, False
+    block["checkpoint_dir"] = export_path.rstrip("/") + ".layerwise_resume"
```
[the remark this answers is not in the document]
Fridah-nv: Confirmed and fixed in 333d2dcdd7.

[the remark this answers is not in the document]
Fridah-nv: The fix prunes every boundary but the committed one, after _write_manifest commits it, so a crash mid-write still leaves the previous boundary resumable. A completed run now keeps none, which is correct since there is nothing left to resume from; there's an assertion on that.

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,632 @@
...
+        assert not self._finalized, "finalize() called twice"
+        # Unlike export_layer, the tail is converted in place: its tensors are read from
+        # model.state_dict() after dispatch, so restoring first would collect unpacked
+        # weights. The model is left in export form and must not be used for inference.
+        self._finalized = True
+
+        model = self._ctx.model
+        quant_config = get_quant_config(model, is_modelopt_qlora=self._ctx.is_modelopt_qlora)
+        _add_mtp_exclusions(model, quant_config)
+        _warn_on_unsynced_moe_gate_up(model)
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 333d2dcdd7. I removed the call rather than moving it: unlike the streaming exporter, where it sits before any layer is packed and reports a genuine miss, there is no point on this path where it could tell the truth. The reason is recorded at the call site so it doesn't get reinstated by symmetry with the other exporters.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2090,14 +2100,44 @@ def layerwise_calibrate(
...
+        save_layer_state=exporter is None,
     )
     start_layer = ckpt.start_layer if ckpt else 0
 
+    if exporter is not None and checkpoint_dir is not None:
+        # detect_resume_point returns None once the manifest is complete, which would put
+        # start_layer back at 0 and recalibrate everything. The shards are already on disk,
+        # so a completed manifest means finalize-only.
+        manifest = _read_manifest(checkpoint_dir)
+        if manifest is not None and manifest.get("last_completed_layer", -1) + 1 >= num_layers:
```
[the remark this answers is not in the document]
Fridah-nv: Both applied in 333d2dcdd7.

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,632 @@
...
+
+    def _write_index(self) -> None:
+        """Build ``model.safetensors.index.json`` from the shards on disk.
+
+        From disk because shards this run resumed past were never seen in memory; from the
+        layer count rather than a glob, so a longer previous run's leftovers cannot leak in.
+        """
+        from safetensors import safe_open
+
+        shards = [self._export_dir / layer_shard_name(i) for i in range(len(self._layers))]
```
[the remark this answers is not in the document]
Fridah-nv: Good catch, applied in 333d2dcdd7.

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,632 @@
...
+        """Re-run the format gate once real calibration state exists.
+
+        AWQ and SVDQuant are only distinguishable after calibration -- their discriminators
+        (``_pre_quant_scale``, ``svdquant_lora_a``) are registered by the calibrator -- so
+        the constructor's gate sees plain NVFP4 and passes. Checking again on the first
+        exported layer fails loudly instead of shipping unfused pre_quant_scale.
+        """
+        if self._calibrated_format_checked:
+            return
+        self._calibrated_format_checked = True
```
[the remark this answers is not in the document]
Fridah-nv: Dropped in 333d2dcdd7; it now checks unconditionally.

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,632 @@
...
+from pathlib import Path
+
+import torch
+import torch.nn as nn
+from safetensors.torch import save_file
+
+from .model_config import FUSION_FREE_FORMATS, QUANTIZATION_NVFP4
+from .quant_utils import get_quant_config, get_quantization_format
+
+__all__ = [
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 333d2dcdd7 by dropping __all__.

modelopt/torch/export/layerwise_export.py
Fridah-nv: Valid, fixed in 333d2dcdd7 — and your diagnosis of why the equivalence testing missed it is right. _fuse_shared_input_modules' return value was discarded, so there was no equivalent of the sibling replay. I added _fuse_unrouted_experts, mirroring the whole-model loop scoped to the layer, which is sound because experts never span one.

examples/hf_ptq/hf_ptq.py
Fridah-nv: Confirmed and fixed in 333d2dcdd7. Using mtp_layer_prefixes_from_checkpoint(args.pyt_ckpt_path) as you suggested, since it's already imported. I gated the fallback on args.layerwise_export so it only affects the path that needs a pre-calibration answer; a layerwise run refuses before the exclusion logic would run, so nothing else changes behaviour.

modelopt/torch/export/layerwise_export.py
Fridah-nv: I was trying to test this update with e2e tests, but cannot make `_fuse_unrouted_experts` actually matters with my setting:  Qwen3-30B-A3B with nvfp4_static expert weights. The reason is that `sync_moe_gate_up_amax` already iterates every expert routed or not, and gate/up is the only input-sharing group among expert linears, so it has already unified the amax that `weight_scale_2` and `global_amax` both derive from.
Do you have an architecture in mind where `sync_moe_gate_up_amax `doesn't reach? I think ` sync_moe_gate_up_amax` makes ` _fuse_unrouted_experts` redundant

Fridah-nv: I just removed the 333d2dc because it's unreachable and not able to be tested. Let me know if you have further concerns on this!

examples/hf_ptq/hf_ptq.py
```diff
@@ -1128,17 +1200,45 @@ def quantize_main(
...
+    if args.layerwise_export:
+        if isinstance(recipe, ModelOptAutoQuantizeRecipe):
+            # Only the mono-quantize path retargets export_dir and runs the refusals;
+            # auto_quantize would export to the placeholder and skip the real export.
+            raise NotImplementedError(
+                "layerwise.export_dir is not supported with an AutoQuantize recipe; "
+                "use a PTQ recipe, or drop export_dir and export afterwards."
+            )
+        # A resumed run leaves skipped layers without amax; only the checkpoint is whole.
+        args.skip_generate = True
```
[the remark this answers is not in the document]
Fridah-nv: Both applied in 30c91eb42b.

realAsma:
> so a crash can lose the layer in flight but never corrupt an earlier one, and a re-run overwrites in place.

I am reading through the code. 
Why do we need rerun and overwriting -> we should be enabling resume automatically with layerwise expert right?

realAsma:
RB:

> <!DOCTYPE html>
> AWQ / SVDQuant | need pre-quant-scale steps that are still whole-model
> -- | --
> 

need pre-quant-scale steps that are still whole-model -> Is this true? I dont think so. pre-quant-scale fusion is entirely limited to one layer (the fusion is to a previous linear layer or pre_norm layer , but in the same decor layer)

Can you check?

realAsma:
> _:robot: Bot comment._

🐝 Looking into this now — I’ll verify the actual fusion/resmoothing scope against the current head.

realAsma:
> _:robot: Bot comment._

Re: [this RB question](https://github.com/NVIDIA/Model-Optimizer/pull/2136#issuecomment-5443851888): you’re right. The *operations* are not inherently whole-model: `apply_pre_quant_scale_and_smooth` changes one linear, and the current layerwise exporter itself documents the relevant fusion/resmoothing groups as intra-layer.

The current refusal is therefore a limitation of this implementation, not a true cross-layer dependency. The whole-model helper (`requantize_resmooth_fused_llm_layers`) traverses the model, but its AWQ work is local fusion into the preceding linear or LayerNorm and its MoE resmoothing is per MoE block. A layerwise path could run those before writing each decoder shard, provided it includes the layer’s preceding norm/linear where needed. Today it only re-discovers shared-input groups per layer, so it does not implement those AWQ/SVDQuant steps and conservatively rejects them.

I’d change the wording from “steps ... are still whole-model” to “not yet implemented in the per-layer exporter”; the unsupported-format guard remains appropriate.

modelopt/torch/quantization/config.py
```diff
@@ -822,12 +839,13 @@ def _coerce_layerwise(cls, value):
...
-                "Set layerwise.enable=True or remove layerwise.checkpoint_dir."
-            )
+        """Raise if a layerwise directory is set but layerwise.enable is False."""
+        for field in ("checkpoint_dir", "export_dir"):
+            if getattr(self.layerwise, field) is not None and not self.layerwise.enable:
+                raise ValueError(
+                    f"layerwise.{field} requires layerwise.enable=True. "
+                    f"Set layerwise.enable=True or remove layerwise.{field}."
+                )
         return self
```
realAsma: Should we just get rid of this check? 
It does not make sense to me. The behavior should be layerwise will be done `if layerwise.enable=True`

modelopt/torch/quantization/model_calib.py
```diff
@@ -2064,12 +2065,22 @@ def layerwise_calibrate(
...
 
+    If ``export_dir`` is passed, each layer is additionally written to a quantized HF
+    checkpoint shard as soon as it is calibrated, leaving a complete checkpoint when the
+    last layer lands and removing the need for a separate ``export_hf_checkpoint()`` pass.
+    Those shards then serve as the resume artifact, so the per-layer weight and quantizer
+    files are not written and finished layers are skipped rather than restored. The model
+    is left unusable for inference either way: export converts each layer in place once its
+    shard is written, and a resumed run additionally never re-calibrates the layers it
+    skipped.
+
```
realAsma: nit -> can we make this short? We should just ask to refer to Layerwise Config doc string here and avoid repeating things.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2090,14 +2101,51 @@ def layerwise_calibrate(
...
+    if exporter is not None:
+        if start_layer > 0:
+            exporter.assert_shards_present(start_layer)
+        elif checkpoint_dir is not None:
+            # Shards but no manifest means the resume record was lost; calibration would
+            # restart at 0 and overwrite them. Without checkpoint_dir there is no resume to
+            # lose, so re-exporting is the documented behaviour.
+            exporter.assert_no_orphan_shards(
+                manifest_present=_read_manifest(checkpoint_dir) is not None
+            )
```
realAsma: nit: Is it possible to move this as a helper method or class method of `LayerwiseExporter`? That will make the layerwise implementation overall more readable.

realAsma: RB: Wdyt?

modelopt/torch/quantization/model_calib.py
```diff
@@ -2185,6 +2255,20 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+        resumed = (
+            f" This run resumed at layer {start_layer}, so layers 0..{start_layer - 1} were "
+            "never re-calibrated either."
+            if start_layer > 0
+            else ""
+        )
+        warn_rank_0(
+            "The exported checkpoint is complete, but per-layer export leaves the model in "
+            "export form: it must not be used for inference." + resumed
+        )
```
realAsma: nit: move this to `exporter.finalize()`

modelopt/torch/quantization/model_calib.py

🐝 I’m checking the ownership boundary between calibration resume state and the exporter.

I agree with extracting it, but not as a `LayerwiseExporter` method: the manifest is calibration/checkpoint state, and moving `_read_manifest` plus layer-count validation into export would blur that ownership. A small calibration-local helper (e.g. `_finalize_completed_layerwise_export(exporter, checkpoint_dir, num_layers)`) would remove the dense branch while keeping the exporter responsible only for shard validation/finalization.

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,491 @@
...
+            if idx is not None:
+                self._layer_names[idx] = name
+        # Descendants too: the tail pass must skip anything a layer shard already covered.
+        self._decoder_owned_ids = {id(m) for layer in layers for m in layer.modules()}
+
+        # Sole owner of model/dtype/qlora; a parallel copy on self would be a second source.
+        self._ctx = ExportContext(
+            model=model,
+            dtype=_resolve_export_dtype(model, dtype),
+            is_modelopt_qlora=is_modelopt_qlora,
```
realAsma: Do we need to support QLoRA? It is not a heavily used feature. Will removing QloRA support simplify any thing? Wdyt?

realAsma: RB: What are your thoughts?

modelopt/torch/quantization/model_calib.py
```diff
@@ -2168,9 +2216,31 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+                        # As _layer_forward_loop does: these tuples were replayed once, so
+                        # the cache would give the probe kv_len twice the mask width.
+                        cache = kwargs_input.get("past_key_values")
+                        if cache is not None:
+                            kwargs_input = dict(kwargs_input)
+                            if hasattr(cache, "reset"):
+                                cache.reset()
+                            else:
+                                kwargs_input["past_key_values"] = None
+                        m(*args, **kwargs_input)
```
realAsma: @Fridah-nv  Could you please remind me what this is for?

realAsma: RB: Can you figure out why we need this? What is this `_fusion_probe` ?

🐝 Looking into this now.

Fridah-nv: This is for getting input amax for uncalibrated experts. But I agree the closure does not belong here, moving it to inside the export handler

I traced it: this is a one-batch probe for `collect_shared_input_modules()`'s forward hooks. During layerwise export it rediscovers modules that receive the same activation, so `_fuse_shared_input_modules()` can give the group the single input scale required by fused export kernels. It runs only for formats that need that fusion; fusion-free formats return before calling it.

The uncalibrated-expert fallback is handled separately by `_prepare_moe_inputs()`, so the probe's direct responsibility is shared-input/scale discovery rather than expert amax collection. Agree that placing this replay closure with the exporter/that operation would make the ownership clearer.

realAsma: Okay I understand why this is needed?

 Can we give `layer_inputs` as inputs to `exporter.export_layer` ? This will make this file more easier to read and follow.

modelopt/torch/quantization/model_calib.py

🐝 Looking into the resume guard now.

I think this still fails open for a syntactically valid but incomplete/hand-edited manifest. `from_folder()` then returns `start_layer=0`, while `_read_manifest(...) is not None` makes `assert_no_orphan_shards()` accept existing shards, so this run overwrites them. The earlier `last is not None` branch only handles a complete manifest. I’d make the orphan check require a *validated resumable* manifest (or raise for any manifest missing valid `last_completed_layer`/`num_layers`) before accepting shards.

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,491 @@
...
+        tensors: dict[str, torch.Tensor] = {}
+
+        # Settle the scales, then change the representation. Order matters at both seams:
+        # scales derive from amax, so they must be final before packing, and the restack
+        # consumes packed per-expert tensors, so it must follow dispatch.
+        #   amax for experts calibration never routed to
+        _prepare_moe_inputs(layer_module, self._ctx.dtype, self._ctx.is_modelopt_qlora)
+        #   one input_scale per fused group, one weight_scale_2 per gate/up pair
+        self._unify_shared_input_scales(layer_module, probe_forward)
+        sync_moe_gate_up_amax(layer_module)
```
realAsma: So this is unifying the weight scales right? This behavior is not clear from the name.

Should we rename `self._unify_shared_input_scales` to `self._unify_shared_quantization_parameters` and then handle both input scale fusion and weight amax sync inside it?

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,498 @@
...
+        args, kwargs = layer_inputs[0]
+        # The tuples were replayed during calibration, so a live cache would give the probe
+        # twice the key length its mask expects.
+        cache = kwargs.get("past_key_values")
+        if cache is not None:
+            kwargs = dict(kwargs)
+            if hasattr(cache, "reset"):
+                cache.reset()
+            else:
+                kwargs["past_key_values"] = None
```
realAsma: Why do we need these handling etc for KV Cache? can we just disable KV Cache using PTQ and calibration and simplify the code please?

modelopt/torch/export/layerwise_export.py
```diff
@@ -0,0 +1,498 @@
...
+        # consumes packed per-expert tensors, so it must follow dispatch.
+        #   amax for experts calibration never routed to
+        _prepare_moe_inputs(layer_module, self._ctx.dtype, self._ctx.is_modelopt_qlora)
+        #   one input_scale per fused group, one weight_scale_2 per gate/up pair
+        self._unify_shared_input_scales(layer_module, layer_inputs)
+        sync_moe_gate_up_amax(layer_module)
+
+        for sub_name, sub_mod in layer_module.named_modules():
+            full_name = f"{layer_name}.{sub_name}" if sub_name else layer_name
+            _dispatch_export_handler(full_name, sub_mod, self._ctx)
```
realAsma, line 252: RB: What is _dispatch_export_handler doing? Can you explain to me?

modelopt/torch/export/layerwise_export.py
realAsma, line 1: Overall looks very nice!!

realAsma, line 1: RB: Can you review this file? Are there any redundancies or scope for any further simplifications? Review applying this repos coding guidelines.

modelopt/torch/export/layerwise_export.py
realAsma: > _:robot: Bot comment._

🐝 I’m checking the export path now.

realAsma: > _:robot: Bot comment._

`_dispatch_export_handler` is the per-module bridge from the export walk to the module-specific export logic.

For each module it:
1. skips PEFT wrapper modules in ModelOpt QLoRA (their `base_layer` is handled separately);
2. calls `unpack_weight()` when the weight is stored packed/FP8, so export sees the live weight and quantizer state;
3. finds the first matching handler in `ExportModuleRegistry` and invokes it.

Those handlers implement the actual conversion for cases such as quantized linears/embeddings and MoE expert layouts—replacing weights with the deployment representation and registering the required scale buffers. Modules with no registered handler are left unchanged. In this layerwise path, it mutates just the calibrated layer before its `state_dict` is collected into that layer’s shard.

examples/hf_ptq/hf_ptq.py
realAsma, line 1: There are some some low-level details of layerwise in this file. Can we move them to `example_utils`? That will improve the readability of this file.

realAsma: RB: Can you also take a look? Can you suggest how to refactor for this?

realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

I agree. I’d move the recipe-shape traversal out of `quantize_main`: make `example_utils` expose one narrow helper that accepts the recipe/algorithm and returns the layerwise blocks (plus dict/object-safe field access). Then `hf_ptq.py` reduces to `is_layerwise = any(...enable...)` and `args.layerwise_export = any(...export_dir...)`, while the existing `example_utils` helpers remain the single owner of export-entry validation, retargeting, and resume-dir resolution.

I would keep the script-level policy in `hf_ptq.py`—AutoQuant refusal, `skip_generate`, compatibility checks, and user logging—because those depend on CLI/runtime flow. Add focused `test_example_utils.py` coverage for dict/list and recipe-object shapes so the extraction contract does not drift.

modelopt/torch/export/layerwise_export.py
realAsma: RB: Thanks! we call `_reconstruct_fused_moe_linear(layer_module)` after this to packing back into 3D tensors. What does 'back' mean? What changed the weights from their original form so that we need this conversion back to 3D?

realAsma: > _:robot: Bot comment._

🐝 Acknowledged — I’m tracing the MoE weight representation change.

realAsma: > _:robot: Bot comment._

“Back” means back to the source `MoELinear` checkpoint layout. `_QuantMoELinear._setup()` first removes its original 3D `weight` (`[num_experts, out, in]`) and expands each slice into a separate 2D `nn.Linear` (`[out, in]`) so every expert can have its own quantizers and be calibrated only on tokens routed to it. Export dispatch quantizes those per-expert weights/scales; reconstruction then stacks them along the expert dimension and removes `experts`, restoring the original key/layout (including stacked scales) that vLLM expects.

modelopt/torch/export/layerwise_export.py
realAsma: > _:robot: Bot comment._

🐝 Acknowledged — reviewing this file against the repository coding standards.

realAsma: > _:robot: Bot comment._

I found one maintainability cleanup worth requesting; no independent correctness blocker in this file.

It has 17 function-local imports (for example lines 50, 73, 114, 155–161, 231–234, 320–331, 446, and 465). CONTRIBUTING asks for module-top imports unless a concrete circular-dependency, optional-dependency, or unusually-heavy-import reason is documented. Please consolidate the normal internal/safetensors imports at the top, retaining only genuinely necessary deferred imports with a short reason. This makes the module's dependencies visible and removes repeated local import plumbing.

I would not combine `completed_layers`, `assert_shards_present`, and `assert_no_orphan_shards`: they encode distinct resume invariants. A follow-up-only comment/docstring trim could help readability, but it is not worth broad refactoring in this PR.

modelopt/torch/export/layerwise_export.py
realAsma: > _:robot: Bot comment._

🐝 Acknowledged — I’m checking the QLoRA-specific export path and its real maintenance cost.

realAsma: > _:robot: Bot comment._

I’d remove layerwise QLoRA support for this PR. It is more than a flag: correct QLoRA export needs the PEFT `base_layer` skip plus the distinct base-model/adapter artifact layout. The layerwise caller never passed `is_modelopt_qlora` to `LayerwiseExporter`, so this path was not actually wired or covered as QLoRA support; the latest change simply removes that dangling parameter.

The simplification is worthwhile, but I’d ask for an explicit early `NotImplementedError` when the model is QLoRA (and a focused test). Otherwise a QLoRA model can enter layerwise export as ordinary export and produce an invalid/mislaid checkpoint rather than a clear unsupported-feature error. Normal whole-model QLoRA export can remain supported.

modelopt/torch/export/layerwise_export.py
Fridah-nv: Agreed and fixed in  589be3a697.

modelopt/torch/export/layerwise_export.py
Fridah-nv: I don't think `ExportContext` as a general structure should carry `is_modelopt_qlora` (a specific module check), and the flag is scattered in the export code. We can do a clean up with another PR.

modelopt/torch/quantization/model_calib.py
Fridah-nv: Done in 589be3a697

modelopt/torch/quantization/model_calib.py
Fridah-nv: Done in 851e613738. `export_layer(layer_idx, layer, layer_inputs)` now takes the cached activations directly and the closure is gone from `model_calib.py`

modelopt/torch/quantization/model_calib.py
Fridah-nv: Extracted in 589be3a697 as `_reconcile_export_with_resume(exporter, checkpoint_dir, start_layer, num_layers)`, just above `layerwise_calibrate`.  Also the fix suggested above

examples/hf_ptq/hf_ptq.py
Fridah-nv: Done in 589be3a697 with suggested fix

modelopt/torch/export/layerwise_export.py
realAsma: > _:robot: Bot comment._

🐝 Acknowledged — checking the QLoRA/export implications.

realAsma: > _:robot: Bot comment._

I’d scope this feature to non-QLoRA rather than carry QLoRA support. QLoRA export has special semantics—skip PEFT wrappers/adapters and strip `base_layer` from checkpoint keys—so it is not just a plumbing boolean. The current tip has already removed `LayerwiseExporter`’s QLoRA argument, meaning layerwise export would otherwise proceed with `False` and can emit incompatible keys/adapters. Please add an early, explicit `NotImplementedError` for ModelOpt QLoRA when `export_dir` is used (and a test). This simplifies the new path; it does not require removing existing whole-model QLoRA export support.

modelopt/torch/quantization/config.py
Fridah-nv: This is checking is to refuse combinations like `layerwise.enable=False` + checkpoint_dir/export_dir specified. I think this is the validators' job. Do you agree?

realAsma: > This is checking is to refuse combinations like layerwise.enable=False + checkpoint_dir/export_dir specified. 

 if layerwise.enable is False I would not bother to validate or check other layerwise settings. I am not particular about this. Upto you!

modelopt/torch/quantization/model_calib.py
```diff
@@ -2047,6 +2048,37 @@ def postprocess(module, name):
     max_calibrate(model, forward_loop)
 
 
+def _reconcile_export_with_resume(
```
realAsma: Can we move this to utils/layerwise_calib.py? This way model_calib.py stays leaner with more algorithm specific things.

realAsma (approved):
LGTM! looks great!

Edwardf0t1 (approved):
LGTM - do we see improvement in export time with this feature, or it's mainly memory saving?

Fridah-nv:
> LGTM - do we see improvement in export time with this feature, or it's mainly memory saving?

On this PR I only tested small model (Qwen3.6 35B) and there's no speedup in the export phase between export layerwise and export in the end. The PR is targeting the case when layewise exceed our 4-hour GPU allocation window and need to save the checkpoint and resume, after this PR we don't need to resume the finished layers any more.
In case of memory, peak memory usage can be less than the original hf path, peak memory for export is now one layer.

modelopt/torch/quantization/config.py
Fridah-nv: Removed in 8fbb734209.

modelopt/torch/export/layerwise_export.py
Fridah-nv: Done in 0cb804b48d on function-local imports

PR #2140, opened 2026-08-11
Add reversible weight folding and support TE grouped weights

mxinO:
### What does this PR do?

Adds `mtq.temporarily_fold_weights(model, snapshot_device=None)` for repeated inference over a frozen fake-quantized model. The context snapshots affected weights and quantizer runtime state, calls each module's native `fold_weight(keep_attrs=True)`, and restores state on exit, including after exceptions.

It also:
- keeps retained weight pre-quant scales inactive while folded;
- supports permanent folding for Transformer Engine `GroupedLinear` weights with grouped or shared `TensorQuantizer`s;
- discovers vLLM fused-MoE `w13`/`w2` weights for weight calibration and temporary snapshots; and
- keeps SVDQuant LoRA residuals separate when folding with `keep_attrs=True`.

`SequentialQuantizer` weights and weights or quantizers shared across `QuantModule` instances are not supported by the temporary context.

Related use case: NVIDIA-NeMo/RL#3441.

### Testing

Focused CPU folding and restoration tests pass. Transformer Engine grouped folding is covered by GPU CI.


modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -245,6 +245,19 @@ def iter_weights_for_calibration(self):
             if weight_i is not None:
                 yield weight_i, (self.weight_quantizer[i] if grouped else self.weight_quantizer)
 
+    def fold_weight(self, keep_attrs: bool = False):
```
cjluo-nv, line 248: > _Bot comment._

This override is fully generic — it only depends on `iter_weights_for_calibration()` — and is a near-duplicate of `_QuantFusedExperts.fold_weight` in `plugins/huggingface.py` (which additionally *lacks* the per-quantizer batching and the `SequentialQuantizer` branch you just added). Suggest hoisting this body into `QuantModule` as e.g. `_fold_weights_from_calibration_iter(keep_attrs)` and calling it from both places (and possibly from vLLM's `_QuantFusedMoEBase.fold_weight`). That removes the third copy of this pattern and, as a side effect, makes HF fused MoE experts fold W4A8-style `SequentialQuantizer` per-expert quantizers, which they currently skip: `set_quantizer_attributes_full` with a list `cfg` replaces `gate_up_proj_weight_quantizers.N` with a `SequentialQuantizer` (matched through `_normalize_fused_experts_quantizer_name`), and `isinstance(q, TensorQuantizer)` there is False, so `fold_weight` silently does nothing.

modelopt/torch/quantization/nn/modules/quant_module.py
```diff
@@ -145,10 +145,16 @@ def _fold_weight_quantizer(
...
-        if not keep_attrs:
-            for attr_name in ("_pre_quant_scale", "_amax"):
-                if hasattr(quantizer, attr_name):
-                    delattr(quantizer, attr_name)
+        tensor_quantizers = (
+            quantizer if isinstance(quantizer, SequentialQuantizer) else (quantizer,)
+        )
+        for tensor_quantizer in tensor_quantizers:
+            # The pre-quant scale is already baked into the folded weight.
+            tensor_quantizer._enable_pre_quant_scale = False
```
cjluo-nv: > _Bot comment._

Correct fix — `pre_quant_scale` is applied ahead of the `_disabled` early return in `TensorQuantizer.forward`, so a retained `_pre_quant_scale` would be re-applied on every subsequent forward (and on a second `fold_weight` call) after the transform was already baked in. Two follow-ups:

1. This changes the observable meaning of `keep_attrs=True`: `quantizer.pre_quant_scale` now returns `None` even though the buffer is retained, and `pre_quant_scale`'s setter asserts `self._enable_pre_quant_scale`, so re-setting it after a fold raises. Please note this in the `fold_weight` docstring below (which currently only talks about dropping buffers) and reconsider the "Changelog: N/A" — `keep_attrs` is public API.
2. Since `_enable_pre_quant_scale` is part of `get_modelopt_state()`, the flag is persisted; worth confirming a save/restore of a folded model is still what you want (i.e. no path re-enables pqs expecting the pre-fold value).

tests/gpu_megatron/torch/quantization/plugins/test_transformer_engine.py
```diff
@@ -118,6 +118,44 @@ def test_quantize(model_cls, config):
...
+            quantizer(weight.float().contiguous()).to(weight.dtype)
+            for weight, quantizer in zip(weights, quantizers)
+        ]
+
+    mtq.fold_weight(model)
+
+    for weight, expected_weight, quantizer in zip(weights, expected_weights, quantizers):
+        assert torch.allclose(weight, expected_weight)
+        assert not quantizer.is_enabled
+        tensor_quantizers = (
```
cjluo-nv: > _Bot comment._

This is the only coverage for the new TE path and per the PR description it was never executed (no TE locally), so please make sure it runs in CI before merge. Two cheap strengthenings while you're here: (a) assert post-fold forward equality (`model(calib_data[0])` before vs. after `fold_weight`, as the CPU test does) — that is what actually catches an off-by-one in the batching/disable ordering end-to-end; (b) assert `not tensor_quantizer._enable_pre_quant_scale`, so the new pre-quant-scale inertness is covered on the grouped/shared path too, not only in `test_fold_weight_supports_sequential_quantizer`.

cjluo-nv (commented):

The core change is correct and well-scoped: `_fold_weight_quantizer` now iterates the stages of a `SequentialQuantizer` (its `disable()`/`disable_rotate()` already broadcast via `_QuantizerContainerBase`), the base `fold_weight` accepts sequential weight quantizers, and the new `_QuantTEGroupedLinear.fold_weight` fixes a real bug — previously a `GroupedQuantizer` was silently skipped by the `dir()` scan, and a *shared* `TensorQuantizer` on TEGroupedLinear would even trip the `assert hasattr(self, weight_name)` since `_setup` deletes `self.weight`. Batching weights per quantizer before `disable()` is the right ordering, and the new CPU unit test verifies output invariance plus the keep_attrs matrix.

Three things worth addressing before merge:

1. **The same bug is left unfixed one file over.** `_QuantFusedExperts.fold_weight` (`plugins/huggingface.py`) still gates on `isinstance(q, TensorQuantizer)`, so per-expert quantizers that `set_quantizer_attributes_full` promoted to `SequentialQuantizer` (a list `cfg` matches `gate_up_proj_weight_quantizers.N` via `_normalize_fused_experts_quantizer_name`, e.g. W4A8) are still silently skipped — folding becomes a no-op there. Since the helper you just generalized handles sequential fine, this is a one-token fix.

2. **Duplicated logic.** The new TE `fold_weight` is entirely generic over `iter_weights_for_calibration()`; it is a near-copy of `_QuantFusedExperts.fold_weight`, and vLLM's `_QuantFusedMoEBase.fold_weight` is a third variant. Consider hoisting the quantizer→weights batching loop into `QuantModule` (e.g. `_fold_weights_from_calibration_iter`) and having TE and fused-experts both call it — that removes the duplication and fixes (1) in the same stroke.

3. **Behavior change under `keep_attrs=True` is undocumented.** Setting `_enable_pre_quant_scale = False` is correct (the weight-quantizer `pre_quant_scale` is applied *before* the `_disabled` early-return in `TensorQuantizer.forward`, so a retained buffer would double-apply — and `_apply_weight_pre_quant_scale` with `_ENABLE_FOLDING_PQS_TO_WEIGHTS=False` does put a live pqs on weight quantizers during auto_quantize). But it means `quantizer.pre_quant_scale` now returns `None` after folding while `_pre_quant_scale` is still present, and `pre_quant_scale`'s setter asserts on `_enable_pre_quant_scale`. Please state this in the `fold_weight` docstring, and reconsider "Changelog: N/A" given `keep_attrs=True` is a public flag whose observable semantics changed.

Also flagging (nudge-level, no action strictly required): the only coverage for the new TE path is a `tests/gpu_megatron` test the author states was never executed locally, and it asserts only weights/`_amax` — not post-fold forward equality or the new `_enable_pre_quant_scale` flag.

modelopt/torch/quantization/plugins/transformer_engine.py
mxinO: Updated after the final simplification: SequentialQuantizer weight folding remains explicitly unsupported. The temporary context now discovers enabled fake-quant pairs through each module's iter_weights_for_calibration(), snapshots only those weights, and still dispatches folding through every plugin's native fold_weight implementation. TE keeps its module-specific GroupedQuantizer-to-weight0...weightN batching; there is no centralized mutation recorder.

modelopt/torch/quantization/nn/modules/quant_module.py
mxinO: Addressed and documented in both the fold_weight docstring and changelog. For permanent fold_weight(..., keep_attrs=True), the retained _pre_quant_scale is intentionally inactive because it is already baked into the stored weight; that inactive state is the expected persisted folded state. temporarily_fold_weights snapshots the prior _enable_pre_quant_scale value and restores it on exit. CPU tests cover retained-but-inactive PQS, output equality, and restoration.

modelopt/torch/quantization/model_quant.py
```diff
@@ -733,6 +735,92 @@ def fold_weight(model: nn.Module, keep_attrs: bool = False):
...
+
+    Args:
+        model: Quantized model whose weights will be temporarily folded.
+        snapshot_device: Device used to store parameter snapshots. ``None`` keeps each snapshot
+            on the parameter's device; ``"cpu"`` avoids the additional accelerator memory.
+    """
+    fold_pairs = []
+    for module in model.modules():
+        if not isinstance(module, QuantModule):
+            continue
```
cjluo-nv, line 771: > _Bot comment._

**Snapshot coverage can silently miss weights that `fold_weight` mutates.** The snapshot set comes from `iter_weights_for_calibration()`, but the mutation comes from each plugin's `fold_weight()`. These disagree today: `plugins/vllm.py::_QuantFusedMoEBase.fold_weight` folds `w13_weight` / `w2_weight`, while the base `iter_weights_for_calibration()` (via `weight_attr_names` → `representative_weight_quantizer`) looks for `w13_weight_weight_quantizer` / `w2_weight_weight_quantizer` and therefore yields *nothing* for that module. Entering this context on a vLLM fake-quant model would fold `w13/w2` permanently and leave their quantizers disabled after exit — silent, unrecoverable weight corruption with no error.

Suggested fixes (either is fine): (a) derive the snapshot from what will actually be folded, e.g. snapshot `named_parameters(recurse=False)` of every `QuantModule` that will be folded; or (b) explicitly detect modules that override `fold_weight` but expose no matching pairs from `iter_weights_for_calibration()` and raise `NotImplementedError` before any mutation. Please also add a unit test with a fake `QuantModule` whose `fold_weight` mutates a weight not surfaced by the calibration iterator, asserting the context refuses rather than corrupting.

modelopt/torch/quantization/model_quant.py
```diff
@@ -733,6 +735,92 @@ def fold_weight(model: nn.Module, keep_attrs: bool = False):
...
+                    "temporarily_fold_weights does not support SequentialQuantizer"
+                )
+            if not isinstance(quantizer, TensorQuantizer) or not quantizer.fake_quant:
+                continue
+            local_weight = weight.to_local() if hasattr(weight, "to_local") else weight
+            fold_pairs.append((local_weight, quantizer))
+
+    weight_snapshots = {}
+    quantizer_states = {}
+    for weight, quantizer in fold_pairs:
```
cjluo-nv, line 785: > _Bot comment._

**The documented "shared weight / shared weight quantizer is unsupported" limitation isn't enforced.** With a weight tied across two quantized modules, `fold_weight` folds it twice (second module's quantizer is still enabled), so forwards *inside* the context are double-fake-quantized — wrong results, no error (state is restored on exit, so it fails silently). With a shared quantizer, the first fold disables it and the second module's weight is never folded, while the shared quantizer's transform is now off for it.

You already have `fold_pairs` built before any mutation, so the guard is cheap: raise if any `id(weight)` appears with two distinct quantizers, or if any quantizer id appears with two distinct weight ids across different modules. Note your earlier reply on this thread stated the context "raises before mutation and a transposed shared-storage test verifies rejection and rollback" — that guard/test isn't present in the final diff, so either restore it or drop the claim.

modelopt/torch/quantization/model_quant.py
```diff
@@ -733,6 +735,92 @@ def fold_weight(model: nn.Module, keep_attrs: bool = False):
...
+            weight_id = id(weight)
+            if weight_id not in weight_snapshots:
+                weight_snapshots[weight_id] = (
+                    weight,
+                    weight.detach().clone()
+                    if snapshot_device is None
+                    else weight.detach().to(snapshot_device, copy=True),
+                )
+        quantizer_states.setdefault(
+            quantizer,
```

Two smaller points on the restore path:

1. If any `weight.copy_(snapshot)` raises (e.g. one param was moved to a different device/dtype inside the region), the loop aborts and **no** quantizer state is restored, leaving the model half-folded with disabled quantizers. Collect exceptions, finish both loops, then re-raise the first.
2. The restored state is four hand-picked private attrs (`_disabled`, `_rotate`, `_enable_pre_quant_scale`, `_input_dtype`). That's correct for today's `_fold_weight_quantizer`, but it will silently rot if folding ever mutates anything else (e.g. `_block_sizes`/`_axis` via `_block_sizes_to_axis`, `_original_shape`/`_block_reshape_size` via `_setup_for_blockquant`, `_dequantize`). Consider reusing the existing `get_quantizer_state_dict`/`set_quantizer_state_dict` (`utils/core_utils.py`) or `TensorQuantizer.get_modelopt_state(properties_only=True)` so the snapshot is exhaustive by construction.

modelopt/torch/quantization/plugins/transformer_engine.py
```diff
@@ -245,6 +245,16 @@ def iter_weights_for_calibration(self):
...
```
cjluo-nv, line 248: > _Bot comment._

:speech_balloon: Author replied: *TE deliberately keeps its module-specific `GroupedQuantizer`→`weight0..weightN` batching; there is no centralized mutation recorder* — flagging for the owner only, since this body is still byte-for-byte the generic "group weights by quantizer, then call `_fold_weight_quantizer`" loop over `iter_weights_for_calibration()` and is now the third copy of that pattern (`_QuantFusedExperts.fold_weight`, `_QuantFusedMoEBase.fold_weight`). Hoisting it to `QuantModule` would also fix the fused-experts sequential gate in one place. Not blocking.

tests/gpu_megatron/torch/quantization/plugins/test_transformer_engine.py
```diff
@@ -118,6 +118,40 @@ def test_quantize(model_cls, config):
...
+        output_after = model(calib_data[0])
+    if isinstance(output_before, tuple):
+        output_before = output_before[0]
+        output_after = output_after[0]
+    assert torch.allclose(output_after, output_before)
+
+    for weight, expected_weight, quantizer in zip(weights, expected_weights, quantizers):
+        assert torch.allclose(weight, expected_weight)
+        assert not quantizer.is_enabled
+        assert not hasattr(quantizer, "_amax")
```
cjluo-nv, line 152: > _Bot comment._

The reply on this thread says the grouped-linear tests now cover "permanent and temporary folding for shared and per-expert quantizers, assert post-fold forward equality, verify retained PQS is inactive while folded, and verify weights/quantizer state are restored afterward". The final diff only has `test_fold_weight_grouped_linear` (permanent fold; forward equality + `_amax` removal) — no `temporarily_fold_weights` exercise and no `_enable_pre_quant_scale` assertion. Since TE grouped is the only non-CPU path in this PR and it can't be run locally, please re-add the temporary-fold grouped test (both `share_weight_quantizer` values) before merge.

Minor: `assert torch.allclose(output_after, output_before)` compares a fold done in `float()` then cast back to the weight dtype against a forward that quantizes in the native dtype; if `TEGroupedLinear` weights are bf16/fp16 here, default `allclose` tolerances may be tight enough to flake — consider explicit `rtol/atol`.

cjluo-nv (commented):

Re-review of PR #2140, now reshaped from "fix sequential folding" into a new public API: `mtq.temporarily_fold_weights` (+ TE grouped `fold_weight`). Size is fine (10 files, +325/-19) and the CPU tests are much better than the previous round (restore-after-exception, rollback when folding fails, SequentialQuantizer rejection, disabled-transform/`weight=None`, SVDQuant residual kept separate).

Design-review note (gate fired): the PR body doesn't discuss alternatives for the new context. The main design question is the *discovery/mutation split*: the context discovers what to snapshot via `QuantModule.iter_weights_for_calibration()` but delegates mutation to each plugin's `fold_weight()`. Those two sets are not guaranteed to agree, and where they don't the context silently fails to restore (see inline: vLLM `_QuantFusedMoEBase`). The repo also already has `get_quantizer_state_dict`/`set_quantizer_state_dict` and `TensorQuantizer.get_modelopt_state()` for capturing quantizer state, versus the four hand-picked private attrs here — please say in the PR body why those weren't reused, and why the snapshot is keyed off the calibration iterator rather than off what `fold_weight` actually touches.

Status of prior review points:
- ✅ `keep_attrs=True` pre-quant-scale semantics are now documented in the `fold_weight` docstring and CHANGELOG.
- ✅ New CPU coverage for the temporary context is meaningful.
- ❌ TE GPU test: the reply says the tests "cover permanent and temporary folding ... verify retained PQS is inactive while folded, and verify weights/quantizer state are restored afterward", but the final diff only contains `test_fold_weight_grouped_linear` (permanent fold; asserts forward equality and `_amax` removal). The temporary-fold grouped test and the `_enable_pre_quant_scale` assertion aren't there.
- :speech_balloon: Duplication (TE / HF fused-experts / vLLM fold loops) — author explained the deliberate choice to keep TE's batching module-specific; the third near-copy still stands, flagging for the owner rather than re-arguing.
- ❌ `_QuantFusedExperts.fold_weight`'s `isinstance(q, TensorQuantizer)` gate (silent no-op fold for W4A8 per-expert `SequentialQuantizer`s) is unchanged, and the base `_fold_weight_quantizer`'s sequential support from the earlier commits was reverted, so `mtq.fold_weight` now *silently* skips sequential weight quantizers while the new context raises for them.
- Also: the earlier reply stated the context "raises before mutation" for weights/quantizers shared across modules and that a shared-storage test verifies rejection + rollback — that guard and test are not in the final diff; the limitation is now only prose in the docstring.

Nit: CHANGELOG mentions the pre-quant-scale change but not the SVDQuant `fold_weight(keep_attrs=True)` behavior change (residual is no longer baked into the weight), which is also public-facing.

No licensing concerns (standard NVIDIA/Apache headers, no vendored code).



---

- `modelopt/torch/quantization/nn/modules/quant_module.py:175` — > _Bot comment._

With the sequential support from the earlier commits reverted, `fold_weight` now *silently* skips a `SequentialQuantizer` weight quantizer (the `isinstance(attr, TensorQuantizer)` gate), while `temporarily_fold_weights` raises `NotImplementedError` for the same case. That asymmetry means `mtq.fold_weight(model, ...)` on a W4A8 model reports success while folding nothing — the exact silent no-op flagged last round for `_QuantFusedExperts.fold_weight` (whose `isinstance(q, TensorQuantizer)` gate is also unchanged). If sequential folding is out of scope now, please make the permanent path warn or raise too, so the two entry points agree and the no-op isn't silent.

tests/gpu_megatron/torch/quantization/plugins/test_transformer_engine.py
cjluo-nv, line 152: > _Bot comment._

This is still only the **permanent** fold path. The previous review asked for the temporary-fold grouped test back (both `share_weight_quantizer` values) plus an `assert not tensor_quantizer._enable_pre_quant_scale`, and an intermediate commit on this branch (`10326e3`) did have `test_temporarily_fold_weight_grouped_linear` — it has since been dropped. Since TE grouped is the only non-CPU path in this PR and per the description it can't be exercised locally, please re-add it so CI covers `temporarily_fold_weights` over `GroupedQuantizer` and shared-quantizer layouts (the batching/`disable()` ordering in the new `fold_weight` is exactly what an end-to-end forward-equality check catches).

Minor, while you're here: `torch.allclose(output_after, output_before)` compares a fold done in `float()` then cast back against a forward that quantizes in the native dtype; if the GroupedLinear weights are bf16/fp16 the default tolerances may be tight enough to flake — consider explicit `rtol`/`atol`.

modelopt/torch/quantization/model_quant.py
cjluo-nv, line 771: > _Bot comment._

The docstring says sharing a weight or weight quantizer across `QuantModule`s is unsupported, but nothing enforces it, so both cases fail silently rather than loudly:

- **Shared weight** (e.g. a weight tied across two quantized modules): each module's `fold_weight` folds it, so the second fold fake-quantizes an already-folded weight. Forwards inside the context are double-quantized; on exit the snapshot restores the original, so nothing ever surfaces.
- **Shared quantizer across modules**: the first `_fold_weight_quantizer` disables it and clears `_enable_pre_quant_scale`, so the second module's weight is never folded while its transform is now off — forwards inside the context skip quantization for that weight entirely.

`fold_pairs` is already materialised before any mutation, so the guard is a few lines: raise if any `id(weight)` maps to two distinct quantizer identities, or if a quantizer identity maps to weights owned by two different modules (TE's grouped case, which legitimately shares one quantizer over `weight0..weightN`, is handled inside a single module's `fold_weight`, so key the check by owning module). Please also add a unit test for the rejection, since an earlier reply on this thread claimed such a guard + test existed.

modelopt/torch/quantization/model_quant.py
```diff
@@ -733,6 +735,92 @@ def fold_weight(model: nn.Module, keep_attrs: bool = False):
...
+                )
+            if not isinstance(quantizer, TensorQuantizer) or not quantizer.fake_quant:
+                continue
+            local_weight = weight.to_local() if hasattr(weight, "to_local") else weight
+            fold_pairs.append((local_weight, quantizer))
+
+    weight_snapshots = {}
+    quantizer_states = {}
+    for weight, quantizer in fold_pairs:
+        if (
```
cjluo-nv, line 786: > _Bot comment._

Two things on the restore path, both carried over from the previous round:

1. If any `weight.copy_(snapshot)` raises (e.g. a parameter was moved to another device/dtype inside the region), the loop aborts and **no** quantizer state is restored at all — the model is left half-folded with disabled quantizers. Collect the exceptions, finish both loops, then re-raise the first.
2. The snapshot is four hand-picked private attrs. That is correct for today's `_fold_weight_quantizer`, but it silently rots if folding ever touches anything else (`_axis`/`_block_sizes` via `_block_sizes_to_axis`, `_original_shape`/`_block_reshape_size` via `_setup_for_blockquant`, `_dequantize`). This repo already has the exhaustive-by-construction idiom: `set_quantizer_by_cfg_context` snapshots `TensorQuantizer.get_modelopt_state(properties_only=True)` and restores via `set_from_modelopt_state` (and `utils/core_utils.py` has `get_quantizer_state_dict`/`set_quantizer_state_dict`). Either reuse one of those, or say in the PR body why the four-attr snapshot is preferred — right now the PR introduces a second, narrower way to save/restore quantizer runtime state next to an existing one.

modelopt/torch/quantization/nn/modules/quant_linear.py
```diff
@@ -164,14 +164,18 @@ def forward(self, input, *args, **kwargs):
...
+        should_fold = (
             hasattr(self, "weight_quantizer")
             and hasattr(self, "weight")
+            and isinstance(self.weight, torch.Tensor)
+            and isinstance(self.weight_quantizer, TensorQuantizer)
             and self.weight_quantizer.fake_quant
-        ):
+        )
+        super().fold_weight(keep_attrs)
+        if should_fold:
```
cjluo-nv, line 175: > _Bot comment._

Adding `not keep_attrs` here changes **permanent** `mtq.fold_weight(model, keep_attrs=True)` on SVDQuant models, not just the temporary context: previously the LoRA residual was baked into the weight *and* the `_svdquant_lora_a/b` buffers were retained, so `SVDQuantLinear.forward` added the residual a second time; now the residual is not baked and the forward applies it once. That's the right fix, but it means the reply on this thread ("This PR no longer changes permanent SVDQuant keep_attrs behavior") doesn't match the diff — the observable output of a public API changed for `keep_attrs=True`.

Please (a) add a CHANGELOG line for it alongside the pre-quant-scale note, and (b) cover the permanent `fold_weight(keep_attrs=True)` SVDQuant case in a test — `tests/unit/torch/quantization/test_calib.py` only exercises it through `temporarily_fold_weights`.

modelopt/torch/quantization/plugins/vllm.py
```diff
@@ -562,6 +562,11 @@ def _setup(self):
         )
         self.parallel_state = create_parallel_state()
 
+    def iter_weights_for_calibration(self):
+        """Yield the fused MoE weights with their corresponding quantizers."""
```
cjluo-nv, line 566: > _Bot comment._

Worth calling out that this override does more than feed the new snapshot: `iter_weights_for_calibration()` previously yielded *nothing* for fused MoE (`weight_attr_names` looks for `w13_weight_weight_quantizer`), so adding it also changes weight-only `max_calibrate` and `promote_static_block_weight_quantizers` for every vLLM fake-quant MoE model. That's plausibly a fix, but it's a behavior change on a path with no test in this PR.

One thing to double-check: this yields the full 3-D `w13_weight`/`w2_weight`, whereas `fold_weight` below folds per-expert slices `weight[i]` with the *same* quantizer. For a per-channel (`axis=0`) or static-block weight quantizer, the amax/`_original_shape` locked in from the 3-D tensor won't match the 2-D slices, so the fold either raises "Input shape has changed" or broadcasts wrongly. Please confirm the configs you care about (per-tensor?) and, if the mismatch is real, either align the fold with the calibrated orientation or reject it explicitly rather than folding silently.

cjluo-nv (commented):

Re-review of PR #2140 (11 files, +330/-19). Good progress since the last round: the vLLM snapshot-coverage bug is genuinely fixed (`_QuantFusedMoEBase.iter_weights_for_calibration` now surfaces `w13`/`w2`, so the temporary context snapshots exactly the tensors vLLM's `fold_weight` mutates — that was the "silent, unrecoverable weight corruption" item), the `keep_attrs=True` pre-quant-scale semantics are documented in the `fold_weight` docstring + CHANGELOG, `_QuantHFParallelLinear.enable_weight_access_and_writeback` is now exception-safe (with a test), and the CPU coverage for the new context is meaningful (restore-after-exception, rollback when a fold fails, SequentialQuantizer rejection, disabled-transform/`weight=None`, SVDQuant residual kept separate).

Still open from the previous review:

- **Requested TE test not re-added (was present, then removed).** The only non-CPU path in this PR is TE grouped folding, and the author states it cannot be run locally. Commit `10326e3` contained `test_temporarily_fold_weight_grouped_linear` (per CodeRabbit's inline reference to lines 155-205); the final diff has only `test_fold_weight_grouped_linear` (permanent fold). So the explicitly requested temporary-fold + `_enable_pre_quant_scale` coverage on the grouped/shared path regressed out of the branch.
- **Documented "shared weight / shared weight quantizer unsupported" limitation is still unenforced.** `fold_pairs` is built before any mutation, so the guard is cheap; without it, a weight tied across two quantized modules is double-fake-quantized inside the context and a shared quantizer leaves the second module's weight unfolded with its transform disabled — both silent (state is restored on exit, so nothing surfaces).
- **Restore path** still aborts on the first `weight.copy_` failure (leaving quantizer state unrestored), and still snapshots four hand-picked private attrs.
- **Design gate is unaddressed in the PR body.** The repo already has the exact idiom for this: `mtq.set_quantizer_by_cfg_context` snapshots/restores quantizer state via `TensorQuantizer.get_modelopt_state(properties_only=True)` / `set_from_modelopt_state` (see `tests/unit/torch/quantization/test_tensor_quant_cpu.py::test_set_quantizer_cxt_restores_on_exception`), and `utils/core_utils.py` has `get_quantizer_state_dict`/`set_quantizer_state_dict`. Please state in the PR body why those weren't reused, and why the snapshot set is keyed off `iter_weights_for_calibration()` rather than off what each plugin's `fold_weight()` actually touches (the two are only guaranteed to agree because every plugin was patched by hand in this PR — a future plugin that overrides `fold_weight` without overriding the iterator silently corrupts weights again).
- **Public behavior change not in the CHANGELOG, and the inline reply doesn't match the diff.** `SVDQuantLinear.fold_weight` now gates the residual bake on `not keep_attrs`, which changes permanent `mtq.fold_weight(model, keep_attrs=True)` on SVDQuant models, not just the temporary context (see inline).
- :speech_balloon: Author replied on the TE `fold_weight` duplication: *TE deliberately keeps module-specific `GroupedQuantizer`→`weight0..weightN` batching; no centralized mutation recorder* — not re-arguing, but flagging for the owner that this is now the third near-copy of "group weights by quantizer, then `_fold_weight_quantizer`" (`_QuantFusedExperts.fold_weight`, `_QuantFusedMoEBase.fold_weight`), and hoisting it would also fix `_QuantFusedExperts.fold_weight`'s unchanged `isinstance(q, TensorQuantizer)` gate (W4A8 per-expert `SequentialQuantizer`s → silent no-op fold, while the new context raises for the same case).

No licensing concerns (standard NVIDIA/Apache headers, no vendored code). Nit: the new CHANGELOG entry is four sentences; the repo's new conciseness guideline asks for one to two.

tests/gpu_megatron/torch/quantization/plugins/test_transformer_engine.py
mxinO: We are intentionally not re-adding a separate TE temporary-fold test. The TE-specific test covers the new grouped/shared quantizer batching and post-fold forward equality, while the generic temporary-fold tests cover weight/runtime-state restoration and retained PQS becoming inactive (through the public `pre_quant_scale is None` behavior). The current TE comparison has also passed GPU CI without a tolerance failure.

modelopt/torch/quantization/model_quant.py
mxinO: This is an explicitly unsupported caller configuration, and we are intentionally keeping that limitation documented rather than adding partial identity-based enforcement. A complete guard would need to reason about parameter ownership and aliased storage/views; checking only object identities would still accept unsupported aliases while implying stronger validation than the API provides. The target use case has module-owned weights and quantizers, so no code change is planned here.

modelopt/torch/quantization/model_quant.py
mxinO: The context is limited to a frozen inference region; moving, replacing, or otherwise mutating parameters inside it is outside the contract, so we are not adding multi-error recovery for a failed restore operation. The four saved fields are the runtime fields the current folding implementation mutates. `get_quantizer_state_dict` only captures tensor state, while `get_modelopt_state(properties_only=True)` is broader checkpoint/config metadata and still excludes runtime block-shape caches, so neither is an equivalent transaction snapshot. We prefer the targeted state here.

modelopt/torch/quantization/nn/modules/quant_linear.py
mxinO: Agreed that `not keep_attrs` also corrects permanent `fold_weight(keep_attrs=True)`: the previous behavior retained the LoRA factors after baking their residual and therefore applied it twice. The earlier reply claiming no permanent behavior change was outdated and has been deleted. The existing SVDQuant temporary-fold test executes this exact `keep_attrs=True` folded state and verifies forward equality plus retained factors, so a second permanent-path test would duplicate the behavior under test. The PR body already calls out that retained SVDQuant residuals stay separate.

modelopt/torch/quantization/plugins/vllm.py
mxinO: We are intentionally keeping this iterator. vLLM's fake-quant forward applies the weight quantizer to the full 3-D fused weight, so weight calibration must use that same rank/orientation; the iterator also ensures the full tensors are snapshotted before any temporary fold attempt. The existing per-expert slicing in vLLM's permanent `fold_weight` predates this PR and is only compatible with slice-safe formats (for example dynamic/scalar-scale cases); broader per-channel/static-block folding support is a separate vLLM change. Temporary vLLM folding is not the training use case targeted by this API, so we are not expanding this PR to redesign its memory-bounded fold path.

modelopt/torch/quantization/nn/modules/quant_linear.py
```diff
@@ -164,14 +164,18 @@ def forward(self, input, *args, **kwargs):
 
     def fold_weight(self, keep_attrs: bool = False):
         """Fold the weight for faster eval."""
-        super().fold_weight(keep_attrs)
-        if (
+        should_fold = (
             hasattr(self, "weight_quantizer")
             and hasattr(self, "weight")
+            and isinstance(self.weight, torch.Tensor)
```
cjluo-nv, line 170: do we have cases that the self.weight is not a tensor and self.weight_quantizer is not a tensorquantizer?

modelopt/torch/quantization/nn/modules/quant_module.py
```diff
@@ -141,22 +141,31 @@ def _fold_weight_quantizer(
...
+
         for weight in weights:
             weight.data.copy_(quantizer(weight.float().contiguous()).to(weight.dtype))
         quantizer.disable()
         quantizer.disable_rotate()
-        if not keep_attrs:
-            for attr_name in ("_pre_quant_scale", "_amax"):
-                if hasattr(quantizer, attr_name):
-                    delattr(quantizer, attr_name)
+        if hasattr(quantizer, "_pre_quant_scale"):
```
cjluo-nv: could you explain a bit more where we have folded the prequant scale?

modelopt/torch/quantization/nn/modules/quant_module.py
```diff
@@ -141,22 +141,31 @@ def _fold_weight_quantizer(
...
+            if keep_attrs:
```
cjluo-nv: will this be confusing that If keep_attrs, you keep the _pre_quant_scale, but disabled it

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -524,8 +524,10 @@ def enable_weight_access_and_writeback(self):
             weight = self.weight
             # TODO: To support TP + FSDP, we need to redistribute the tensor with replicate instead of shard
             self.weight = nn.Parameter(weight.to_local())
-            yield
-            self.weight = weight
+            try:
```
cjluo-nv, line 527: why adding the try finally?

modelopt/torch/quantization/nn/modules/quant_linear.py
mxinO: Yes. `weight` can be `None` for tied/output modules, and `weight_quantizer` can be a `SequentialQuantizer`. These checks mirror what the base fold can actually process.

modelopt/torch/quantization/nn/modules/quant_module.py
mxinO: It is folded by `quantizer(weight)` above. `TensorQuantizer` applies `pre_quant_scale` before quantization, and that result is copied back into the weight.

modelopt/torch/quantization/nn/modules/quant_module.py
mxinO: `keep_attrs` means retaining the buffer for restoration, not keeping its effect active. Since the scale is already baked into the weight, it must be disabled to avoid applying it twice.

modelopt/torch/quantization/plugins/huggingface.py
mxinO: We temporarily replace the DTensor weight with its local parameter. If folding raises, `finally` restores the original DTensor; otherwise the module would remain in the temporary local state.

realAsma:
RB: Help me understand this PR. Do you know what is the motivations for this PR? What problem is this trying to solve?

realAsma:
> _:robot: Bot comment._

🐝 Acknowledged — I’m tracing the intended inference workflow and the folding changes now.

realAsma:
> _:robot: Bot comment._

The main motivation is a repeated frozen-model inference workflow—called out by the NeMo RL use case, where log-probabilities are recomputed across several microbatches after fake quantization.

Today `fold_weight()` is effectively permanent: it bakes fake quantization/pre-quant scaling/rotation into the stored weights and disables or drops quantizer state. Re-running the normal fake-quant path on every forward adds work, but permanently folding is unsuitable when the model must return to its original quantized state afterward.

This PR adds `mtq.temporarily_fold_weights(...)`: a scoped fast path that snapshots affected weights and runtime quantizer state, folds using each module’s native implementation for the context, then restores everything—even on an exception. It also makes that folding cover real model layouts that were previously missed: Transformer Engine `GroupedLinear` (including shared/grouped quantizers) and vLLM fused-MoE weights, while keeping SVDQuant LoRA residuals separate.

So the problem being solved is efficient repeated inference over a frozen fake-quantized model without making a destructive, irreversible fold or corrupting calibration/runtime state for later use.

modelopt/torch/quantization/nn/modules/quant_module.py
cjluo-nv: @realAsma could you confirm?

modelopt/torch/quantization/model_quant.py
```diff
@@ -733,6 +735,92 @@ def fold_weight(model: nn.Module, keep_attrs: bool = False):
...
+        with torch.no_grad():
+            for weight, snapshot in weight_snapshots.values():
+                weight.copy_(snapshot)
+        for quantizer, state in quantizer_states.items():
+            (
+                quantizer._disabled,
+                quantizer._rotate,
+                quantizer._enable_pre_quant_scale,
+                quantizer._input_dtype,
+            ) = state
```
realAsma: what is the point of caching `quantizer_states`? it is not used anywhere.

🐝 I’m looking into this.

`quantizer_states` is used in the `finally` block (lines 815–821) to restore the runtime flags that `fold_weight(..., keep_attrs=True)` mutates: disabled/enabled, rotation, pre-quant-scale enablement, and input dtype. That makes the context reversible even when its body or folding raises.

`setdefault` captures each quantizer only once; a quantizer can appear for multiple weights within one module (for example grouped/shared TE handling), so this avoids redundant snapshots/restores. The focused exception-restoration test covers this path. PTAL otherwise: no additional blocker found; required CI is green.

modelopt/torch/quantization/nn/modules/quant_module.py
```diff
@@ -141,22 +141,31 @@ def _fold_weight_quantizer(
...
+                # The scale is already baked into the folded weight.
+                quantizer._enable_pre_quant_scale = False
+            else:
+                delattr(quantizer, "_pre_quant_scale")
+        if not keep_attrs and hasattr(quantizer, "_amax"):
+            delattr(quantizer, "_amax")
```
realAsma: RB: This code change does not look clean to me. multiple if branching; Can you please propose a cleaner code: 

May be keep the for loop and do a special condition for <if keep_attrs: 
                # The scale is already baked into the folded weight.
                quantizer._enable_pre_quant_scale = False> ?

realAsma: > _:robot: Bot comment._

🐝 Acknowledged — I’m checking the cleanest shape for this cleanup.

realAsma: > _:robot: Bot comment._

I agree. The cleanup can keep the existing attribute loop and make the retained-scale case the only special case:

```python
if keep_attrs and hasattr(quantizer, "_pre_quant_scale"):
    # The scale is already baked into the folded weight.
    quantizer._enable_pre_quant_scale = False
elif not keep_attrs:
    for attr_name in ("_pre_quant_scale", "_amax"):
        if hasattr(quantizer, attr_name):
            delattr(quantizer, attr_name)
```

This makes the two modes explicit: retain all calibration attrs while disabling only the already-applied pre-scale, or remove both attrs. It also avoids the current nested handling of `_pre_quant_scale`.

modelopt/torch/quantization/model_quant.py
```diff
@@ -733,6 +735,92 @@ def fold_weight(model: nn.Module, keep_attrs: bool = False):
             module.fold_weight(keep_attrs)
 
 
+@contextmanager
+def temporarily_fold_weights(
```
realAsma: @mxinO 

I do not like the fact that we are introducing additional high level APIs for temporary fold weights.  Can we add a new argument `temporary` to our existing `mtq.fold_weights` API to accomplish this without introducing a new high level API?

realAsma: RB: Can you please propose a design on how to do this in a minimal and clean way?

🐝 Acknowledged — I’m checking the smallest compatible design.

modelopt/torch/quantization/nn/modules/quant_module.py
realAsma: RB:

nit:
Can we have a more helpful comment: 
```
# The scale is already baked into the folded weight. 
# set enable_pre_quant_scale to False so that it is not double applied.
```

modelopt/torch/quantization/model_quant.py

Suggested minimal shape: keep `mtq.fold_weight(model, keep_attrs=False)` as the eager, permanent API, and add keyword-only `temporary=False, snapshot_device=None`.

```python
with mtq.fold_weight(model, temporary=True, snapshot_device="cpu"):
    run_inference()
```

Internally, extract the current loop into private `_fold_weight(...)`. The public function calls it directly when `temporary=False`; when `temporary=True`, it returns a private context manager that contains the existing snapshot/restore logic and calls `_fold_weight(model, keep_attrs=True)`.

That keeps the temporary-only requirements (retained quantizer attributes and restoration on exceptions) inside the private manager, avoids another public API/export, and preserves all existing permanent callers unchanged. I would document that `keep_attrs` is only meaningful for the permanent path (the temporary path must retain attrs). Move the current tests to the `temporary=True` form and retain/add coverage for default permanent folding and exception restoration.

modelopt/torch/quantization/nn/modules/quant_module.py
realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

Suggested wording:

```python
# The pre-quantization scale has been folded into the weight, so disable it to avoid applying it twice.
```

This keeps the rationale and the double-application failure mode in one comment.

realAsma: @mxinO  I feel that we should just make `keep_attrs` to be True always and just avoid deleting the attributes. This could simplify our implementations and code. 

`fold_weights` API then updates weights in place and  disables quantization, rotation, pre_quant_scale etc.

modelopt/torch/quantization/model_quant.py
```diff
@@ -733,6 +735,92 @@ def fold_weight(model: nn.Module, keep_attrs: bool = False):
...
+                quantizer._disabled,
+                quantizer._rotate,
+                quantizer._enable_pre_quant_scale,
+                quantizer._input_dtype,
+            ),
+        )
+    try:
+        fold_weight(model, keep_attrs=True)
+        yield
+    finally:
```
realAsma: Instead of caching all these and restoring, can we create a new API say `preserve_quantizer_attributes_context` and call the `fold_weights` within that context?

See 
 https://github.com/NVIDIA/Model-Optimizer/blob/94915a1e5d8a0c6466471c3bd140dbd1662e95ff/modelopt/torch/quantization/conversion.py#L506 `

modelopt/torch/quantization/model_quant.py
realAsma: RB: I missed that temporarily_fold_weights is a context manager. The original `fold_weights` API is not a context manager. I want to keep backward compatibility with the original fold_weights API as well as support usage like `with mtq.fold_weight(model, temporary=True, snapshot_device="cpu"):` . 

Can you propose a design for this.

Also when we use `fold_weights` a context method, using temporary=False looks weird to me? General expectation is that a context method applies a function temporarily within the context. 

Wdyt?
Can you give a few design suggestions?

modelopt/torch/quantization/model_quant.py
realAsma: We can then simplify `set_quantizer_by_cfg_context` as well which should apply the `quant_cfg` and call `preserve_quantizer_attributes_context` underneath.

modelopt/torch/quantization/model_quant.py

🐝 I’m tracing the existing API and call sites now; I’ll suggest a backward-compatible split between permanent folding and a clearly temporary context API.

modelopt/torch/quantization/model_quant.py
realAsma: RB: Wdyt? If this is a good idea, please propose a design

modelopt/torch/quantization/model_quant.py

I agree: a context API with `temporary=False` is semantically awkward.

I’d keep one public entry point (the existing singular `mtq.fold_weight`) and split the implementations privately:

```python
def fold_weight(model, keep_attrs=False, *, temporary=False, snapshot_device=None):
    if temporary:
        if keep_attrs:
            raise ValueError("temporary folding owns state retention")
        return _temporarily_fold_weights(model, snapshot_device=snapshot_device)
    if snapshot_device is not None:
        raise ValueError("snapshot_device requires temporary=True")
    _fold_weight_permanently(model, keep_attrs=keep_attrs)
```

So existing `mtq.fold_weight(model)` remains an eager permanent operation returning `None`; only `with mtq.fold_weight(model, temporary=True, snapshot_device="cpu"):` yields a context manager. The temporary implementation can use the current snapshot/restore logic and always fold internally with retained attrs. Since `temporarily_fold_weights` is new in this PR, I would remove it rather than publish two high-level APIs.

The main trade-off is the union return type, but it preserves the old call contract and avoids a misleading permanent context. I’d add tests for the unchanged permanent call, restoration on normal/exceptional exit, and rejecting `snapshot_device` without `temporary=True`.

modelopt/torch/quantization/model_quant.py

🐝 I’m looking into the context-manager design and its interaction with `set_quantizer_by_cfg_context`.

modelopt/torch/quantization/model_quant.py
realAsma: @mxinO  PTAL

modelopt/torch/quantization/model_quant.py

Yes—this is a better shared abstraction. Extract the current snapshot/restore portion of `set_quantizer_by_cfg_context` into `preserve_quantizer_attributes_context(model)`: snapshot every `TensorQuantizer` (and each `SequentialQuantizer` child) with `get_modelopt_state(properties_only=True)`, then restore attributes and the original quantizer type in `finally`.

Then make `set_quantizer_by_cfg_context` a thin wrapper:
```python
with preserve_quantizer_attributes_context(model):
    set_quantizer_by_cfg(model, quant_cfg)
    yield
```
and make `temporarily_fold_weights` use the same context around `fold_weight(..., keep_attrs=True)`. It should still snapshot/restore the affected weight tensors in place: folding necessarily mutates weights, whereas the new context owns only quantizer state. That removes the hand-maintained `_disabled/_rotate/_enable_pre_quant_scale/_input_dtype` tuple and automatically preserves other quantizer attributes.

I’d retain the current sequential-config restriction only on `set_quantizer_by_cfg_context`; the preservation context itself should support sequential quantizers. Add direct preservation tests (including exception and type restoration), plus the existing temporary-fold tests to prove weights and all quantizer attributes round-trip.

modelopt/torch/quantization/nn/modules/quant_module.py
realAsma: > _:robot: Bot comment._

🐝 Acknowledged — checking the wording.

realAsma: > _:robot: Bot comment._

Suggested wording:

```python
# The pre-quantization scale has been folded into the weight, so disable it to avoid applying it again.
```

This makes both the retained state and the reason for disabling the flag explicit.

modelopt/torch/quantization/model_quant.py

🐝 Acknowledged — I’m reviewing the API design and compatibility tradeoffs.

I agree that `temporary=False` is awkward here. I’d keep the existing permanent API exactly as the one-shot command:

```python
mtq.fold_weight(model, keep_attrs=False)
```

and expose the reversible behavior as a separately named context manager:

```python
with mtq.temporarily_fold_weights(model, snapshot_device="cpu"):
    ...
```

That makes the lifetime obvious, keeps the current `fold_weight` call fully backward-compatible, and keeps `snapshot_device` scoped to the only mode that needs a snapshot.

I would avoid `fold_weight(..., temporary=True)`: it gives one function two unrelated return contracts (normal call vs. context manager), and an accidental call without `with` would silently do nothing. It also leaves `temporary=False` as a redundant and confusing spelling of the default permanent operation.

If we want a shorter context name, `fold_weight_context(model, ...)` is the clearest alternative, but my preference is the current `temporarily_fold_weights` wording. The tests should explicitly cover permanent folding, temporary restoration (including exceptions), and that the temporary path preserves parameter identity and quantizer state.

modelopt/torch/quantization/model_quant.py

🐝 Looking into the design.

modelopt/torch/quantization/nn/modules/quant_module.py
realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

Suggested wording:
```python
# The scale is already baked into the folded weight.
# Disable pre-quant scaling so it is not applied twice.
```
This explains both the state change and the double-application it prevents.

modelopt/torch/quantization/model_quant.py

🐝 Acknowledged — I’m checking the API and compatibility trade-offs now.

🐝

I agree that a `temporary=False` flag on a context-oriented API is surprising, and overloading `fold_weight` would make its return type conditional (`None` vs. context manager).

I would keep two explicit operations:
- `mtq.fold_weight(model, keep_attrs=False)`: the existing permanent, non-context API—unchanged for compatibility.
- `with mtq.temporarily_fold_weights(model, snapshot_device="cpu"):`: the reversible operation.

If we want a shorter public name, `folded_weights(...)` / `temporary_fold_weights(...)` is clearer than adding `temporary=True`; it signals the scoped lifetime and cannot be mistaken for permanent mutation. An alias can be added later without changing `fold_weight`.

The contract/tests should make the distinction explicit: permanent folding persists after the call; the context restores weights and quantizer state on normal and exceptional exit.

realAsma:
> _:robot: Bot comment._

🐝 Acknowledged — the supplied review-thread ID is no longer available through GitHub, so I’m reviewing the referenced file and responding at PR level.

realAsma:
> _:robot: Bot comment._

The referenced thread itself now returns 404, so this is a PR-level reply.

`temporarily_fold_weights` is a reversible inference-only wrapper: it discovers each `QuantModule`’s `(weight, quantizer)` pairs, snapshots changed weights plus the quantizer runtime flags, calls existing `fold_weight(..., keep_attrs=True)`, and restores both in `finally` (including when the body raises). `snapshot_device="cpu"` trades transfer time for avoiding an extra accelerator-resident copy.

The core intent is sound, but I would not treat the docstring’s unsupported cases as sufficient protection: shared weights/quantizers can be folded more than once or leave a second weight with its transform disabled, yet the code does not reject them. Also, a restore failure in `weight.copy_` stops later weight/state restoration. I’d ask for explicit pre-mutation alias checks and best-effort restoration before accepting this public context API.

realAsma (approved):
LGTM! Left some comments, please address them.

modelopt/torch/quantization/nn/modules/quant_module.py
mxinO: Added.

modelopt/torch/quantization/model_quant.py
mxinO: Good suggestion, addressed.

PR #2146, opened 2026-08-11
Add Nemotron Lightning 3.5 NVFP4 recipe and QAD example

jenchen13:
### What does this PR do?

Type of change: New example

Add Nemotron Lightning 3.5 NVFP4 recipe and QAD example
Also exclude MTP in default disabled quantizers
### Usage

```python
# uv run launch.py --yaml examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/megatron_lm_qad.yaml --yes
```

## Summary by CodeRabbit

- **New Features**
  - Added an end-to-end NVFP4 quantization, distillation, and export workflow for NVIDIA Nemotron 3.5 Lightning 30B-A3B.
  - Added a PTQ configuration supporting NVFP4 W4A16 quantization with optimized scaling and FP8 support for selected components.

- **Bug Fixes**
  - Preserved custom model output locations when provided, while retaining the existing default path.

- **Configuration**
  - Disabled quantization for MTP modules by default.
  - Updated an existing Nemotron workflow to use the aggressive NVFP4 quantization profile.


modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B/lightning_w4a16_nvfp4_4o6.yaml
```diff
@@ -0,0 +1,73 @@
...
+    # KV cache -> FP8.
+    - quantizer_name: '*[kv]_bmm_quantizer'
+      enable: true
+      cfg:
+        num_bits: e4m3
+    # MCore lm_head (output_layer) W4A16 4/6 weight-only.
+    - quantizer_name: '*output_layer*weight_quantizer'
+      enable: true
+      cfg: {$import: nvfp4_four_over_six}
+
```
kevalmorabia97: Can you also disable MTP quantization in the end in the recipe so we dont quantize MTP? While our export just copies BF16 MTP weights, it may be confusing to see MTP quantized in PTQ logs but not show up in exported ckpt.

```
    # Disable MTP quantizers.
    - quantizer_name: 'mtp.*'
      enable: false
```

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/megatron_lm_qad.yaml
```diff
@@ -0,0 +1,151 @@
...
+  # Quantize the teacher checkpoint into the NVFP4 student checkpoint.
+  task_1:
+    script: common/megatron_lm/quantize/quantize.sh
+    args:
+      - --seq-length 32768 --max-position-embeddings 32768
+      - --calib-size 768
+      - --skip-generate
+      - --export-default-te-spec
+    environment:
+      - MLM_MODEL_CFG: nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16
```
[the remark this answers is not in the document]
jenchen13: they share the same config this is fine

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/megatron_lm_qad.yaml
```diff
@@ -50,7 +50,7 @@ pipeline:
       - --export-default-te-spec
     environment:
       - MLM_MODEL_CFG: nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16
-      - QUANT_CFG: MAMBA_MOE_NVFP4_CONSERVATIVE_CFG
+      - QUANT_CFG: MAMBA_MOE_NVFP4_AGGRESSIVE_CFG
```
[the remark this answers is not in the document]
jenchen13: fixed

modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml
```diff
@@ -38,6 +38,8 @@
     enable: false
   - quantizer_name: '*router*'
     enable: false
+  - quantizer_name: 'mtp.*'
```
kevalmorabia97, line 41: @claude  will this rule be applied in the end after all rules from 406 recipe yaml is applied? For example, if we first disable `mtp.*` but then in our recipe, we enable `*output_layer*weight_quantizer`, will this re-enable `mtp.*output_layer*weight_quantizer`?

cjluo-nv, line 41: > _Bot comment._

Two concerns with this entry:

1. **Anchoring.** `mtp.*` has no leading wildcard, so it only matches quantizer names that begin literally with `mtp.`. The existing MTP exclusions in the repo use `*mtp*` (`modelopt_recipes/huggingface/qwen3_5/ptq/w4a16_nvfp4-fp8_attn-kv_fp8_cast.quant_cfg.yaml`). Under any wrapper/root prefix that Megatron or HF adds (`module.`, `model.`, `decoder.`) this pattern silently matches nothing and the MTP block stays quantized. Unless you have verified the exact module names for the MCore path this runs on, `*mtp.*` (or `*mtp*`) is the safer choice.

2. **Precedence.** This unit is only effective where it is `$import`ed, and only at that position — later entries override it. It works for `mamba_moe_nvfp4_{aggressive,conservative}` (imported after the W4A4 enables), but note that a recipe importing this unit and then enabling `*output_layer*weight_quantizer` / `*lm_head*weight_quantizer` afterwards will re-enable `mtp.…output_layer.weight_quantizer`, since those patterns are leading-wildcard. Worth a one-line comment here that this unit must be imported *after* all enable rules.

modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B-BF16/ptq/w4a16_nvfp4_4o6.yaml
```diff
@@ -0,0 +1,73 @@
...
+#   NVFP4 weights use the four_over_six (4/6) per-block scale selection.
+#   Weight-only; activations BF16. MSE selects per-block M=6 (mult 1.0) vs
+#   M=4 (mult 1.5); the choice is baked into the amax by MSE.
+#
+# HF -> MCore name mapping for Nemotron-3.5-Lightning-30B-A3B (Mamba/Attn/MoE hybrid):
+#   - mixer.experts.<N>.{up,down}_proj   -> mlp.experts.local_experts.<N>.linear_fc{1,2}
+#   - mixer.shared_experts.{up,down}_proj -> mlp.shared_experts.linear_fc{1,2}
+#   - mixer.in_proj / out_proj            -> mixer.in_proj / out_proj      (same name; W4A16)
+#   - lm_head                             -> output_layer                  (W4A16)
+imports:
```
cjluo-nv, line 14: > _Bot comment._

This recipe does not import `configs/ptq/units/default_disabled_quantizers` (compare the Ultra 4/6 mirror, which also skips it, and the Super mirror, which relies on explicit `*` disable). Consequence: the `mtp.*` exclusion added in this same PR has **no effect on this recipe**, even though the PR body ties the two changes together. If the Lightning MCore model has an MTP block containing a MoE transformer layer, `*mlp.experts*weight_quantizer` / `*mixer.experts*weight_quantizer` below will quantize it. Either import the unit (last, after the enables) or add an explicit `*mtp*` disable at the end of this list.

Separately, file placement/naming: the sibling checkpoint mirrors live at `huggingface/models/nvidia/<checkpoint>/ptq/<scheme>.yaml` (e.g. `Nemotron-3-Ultra-550B-A55B/ptq/nvfp4-4o6.yaml`). This one is `…/Nemotron-3.5-Lightning-30B-A3B/lightning_w4a16_nvfp4_4o6.yaml` — no `ptq/` segment, redundant `lightning_` prefix, underscores instead of hyphens. Please align (and update `QUANT_CFG` in the launcher yaml accordingly). Also: the license header is truncated to the two SPDX lines (missing the Apache block from `LICENSE_HEADER`) and the year is 2024 on a brand-new file.

modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B/lightning_w4a16_nvfp4_4o6.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - quantizer_name: '*'
+      enable: false
+    # W4A16 NVFP4 4/6 weight-only (block 16, static, four_over_six: true). HF + MCore names.
+    - quantizer_name: '*mixer.experts*weight_quantizer'
+      enable: true
+      cfg: {$import: nvfp4_four_over_six}
+    - quantizer_name: '*mixer.shared_experts*weight_quantizer'
+      enable: true
+      cfg: {$import: nvfp4_four_over_six}
+    - quantizer_name: '*lm_head*weight_quantizer'
```
cjluo-nv: > _Bot comment._

Enabling NVFP4 4/6 on `*lm_head*weight_quantizer` and (line 71) `*output_layer*weight_quantizer` is a deliberate deviation from the other Nemotron-3 mirrors, which keep `lm_head`/`output_layer` in BF16 (and from `default_disabled_quantizers`, which disables both). Two things to confirm since the PR's Testing section is empty:

- MCore→HF export/deployment of a quantized `output_layer` on this architecture (tied embeddings in particular — cf. the recent `fold_weight` fix for Megatron tied output layers).
- With an MTP block present, both patterns are leading-wildcard and therefore also match `mtp.…output_layer.weight_quantizer` / any MTP-side head. If the intent is "main head only, MTP BF16", these need to be scoped or followed by an `*mtp*` disable.

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/megatron_lm_qad.yaml
cjluo-nv, line 53: > _Bot comment._

`MLM_MODEL_CFG` is the **Nano** model config (`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16`) in tasks 1, 2 and 3 of a Lightning pipeline, while every checkpoint path points at Lightning. Every other launcher example keeps `MLM_MODEL_CFG` equal to the model directory (see `NVIDIA-Nemotron-3-Ultra-550B-A55B-BF16/megatron_lm_ptq.yaml`). Looks like a copy-paste leftover — building a Lightning MCore checkpoint under the Nano arch config will either fail to load or silently mis-shape. If it is intentional (Megatron-LM has no Lightning cfg yet), please add a comment saying so and fix the path derivation noted below.

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/megatron_lm_qad.yaml
```diff
@@ -0,0 +1,151 @@
...
+      - QUANT_CFG: huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B/lightning_w4a16_nvfp4_4o6
+      - MLM_MODEL_CKPT: /cicd/megatron-lm-bf16/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-MCore
+      - MLM_MODEL_SAVE: /cicd/megatron-lm/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16
```
cjluo-nv: > _Bot comment._

`MLM_MODEL_SAVE` set here is a no-op: `common/megatron_lm/quantize/quantize.sh` does `export MLM_MODEL_SAVE="/cicd/megatron-lm/${MLM_MODEL_CFG}"` unconditionally. Combined with `MLM_MODEL_CFG` being the Nano config (line 53), the quantized student is written to `/cicd/megatron-lm/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16`, while task_2 loads `MLM_MODEL_CKPT: /cicd/megatron-lm/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16` — a path nothing ever writes. As written this pipeline (with `allow_to_fail: false`) should fail at the QAD step; the Nano example avoids this by leaving the save path at the `MLM_MODEL_CFG` default. Please fix `MLM_MODEL_CFG` (and drop the redundant `MLM_MODEL_SAVE`, or make `quantize.sh` respect it).

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/megatron_lm_qad.yaml
```diff
@@ -0,0 +1,151 @@
...
+# quantizes a separate MCore student, distills the student for 400 iterations,
+# and exports the resulting checkpoint. The training task uses an explicit
+# Nemotron-Post-Training-Dataset-v2 chat shard so Hugging Face Datasets does not
+# prepare the repository's other large splits.
+#
+# PTQ topology: 1 B200 node x 4 GPUs, TP=1, PP=1, CP=1, EP=4, ETP=1.
+# QAD topology: 2 B200 nodes x 4 GPUs, TP=2, PP=1, CP=1, EP=4, ETP=1.
+# With micro-batch-size=1 and global-batch-size=16, train-samples=6400 produces
+# 400 iterations. To use another sequence length, change both --seq-length and
+# --max-position-embeddings. Our final QAD run used a sequence length of 524,288.
```
cjluo-nv, line 13: > _Bot comment._

"Our final QAD run used a sequence length of 524,288" doesn't match the pipeline (`--seq-length 32768`, job name `..._32k_400iter`) and is implausible for a 30B model on 8 GPUs. Did you mean 524,288 *tokens per global batch* (16 × 32768), or was the run at a different seq length? As written this line will send readers down the wrong path.

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/megatron_lm_qad.yaml
cjluo-nv, line 53: > _Bot comment._

`CONSERVATIVE → AGGRESSIVE` is a real numerics change to an existing `allow_to_fail: false` pipeline: `mamba_moe_nvfp4_conservative` is the aggressive preset *plus* `*mixer.in_proj*` / `*mixer.out_proj*` disabled, so the Mamba projections move from BF16 to NVFP4 W4A4 — whereas the published Super/Ultra checkpoint mirrors deliberately keep `mixer.{in,out}_proj` at FP8 because NVFP4 there hurts accuracy. Please add the rationale/accuracy data to the PR body (the Testing section is currently empty), and consider splitting this out of the Lightning PR since it is unrelated to the new example. Removing `QUANT_CFG` from the export task looks fine (it only feeds the log line and the `EXPORT_DIR` default, which is set explicitly), but `export.sh`'s header still documents it as required env — worth updating that comment.

cjluo-nv (commented):

**Answer to the ordering question first** (`default_disabled_quantizers` vs. the 4/6 recipe):

- `quant_cfg` is a flat ordered list and **the last matching entry wins** (`docs/source/guides/_quant_cfg.rst` → "Ordering and Precedence"; enable-only entries flip state without resetting attributes). There is no implicit "defaults applied at the end" step — `default_disabled_quantizers` only takes effect where a preset/recipe `$import`s it, at exactly that position in the list.
- In the presets it is imported *after* the enable rules (`configs/ptq/presets/model/mamba_moe_nvfp4_aggressive.yaml`: `base_disable_all` → `w4a4_nvfp4_nvfp4` → `default_disabled_quantizers` → `mamba_moe_disabled_quantizers`), so the new `mtp.*` disable does win there. That's the path the Nano QAD switch to `MAMBA_MOE_NVFP4_AGGRESSIVE_CFG` uses, so the new entry is effective for that pipeline.
- Yes to your specific worry: if a recipe re-enables *after* the import (the `qwen3_5` recipes do exactly this for `*lm_head*weight_quantizer`), the later enable wins, and because the enable patterns are leading-wildcard (`*output_layer*weight_quantizer`) they also match `mtp.…output_layer.weight_quantizer` — MTP would be re-enabled. Exclusions must come last, or the enable patterns must be scoped.
- Important for this PR: `lightning_w4a16_nvfp4_4o6.yaml` **never imports `default_disabled_quantizers`**, so the new `mtp.*` entry has no effect on the new Lightning recipe at all, even though the PR body presents the two changes as related. The recipe's own `*mlp.experts*weight_quantizer` / `*mixer.experts*weight_quantizer` / `*output_layer*weight_quantizer` rules will match MTP-internal modules if the Lightning MCore model carries an MTP block.
- Also note `mtp.*` is *anchored* (no leading `*`), unlike the `*mtp*` used in the qwen3_5 recipes: it only matches quantizer names that literally start with `mtp.`, so under any wrapper/root prefix (`module.`, `model.`, `decoder.`) it silently matches nothing.

Other findings (details inline): the new launcher pipeline points `MLM_MODEL_CFG` at the **Nano** model config and sets `MLM_MODEL_SAVE`, which `quantize.sh` unconditionally overrides — as written, task_2 loads a checkpoint path that task_1 never writes; the Nano `CONSERVATIVE → AGGRESSIVE` switch is an unexplained numerics change (Mamba `in_proj`/`out_proj` go from BF16 to NVFP4 W4A4, whereas the published Super/Ultra mirrors deliberately keep them at FP8) bundled into a Lightning PR; recipe file placement/naming deviates from the `<checkpoint>/ptq/<scheme>.yaml` convention; PR "Testing" section is empty and there is no CHANGELOG entry.

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/megatron_lm_qad.yaml
jenchen13: this is intentional to change it to aggressive config

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/megatron_lm_qad.yaml
jenchen13: I mean the final Lightning NVFP4 QAD run, this is fine

tools/launcher/examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/megatron_lm_qad.yaml
jenchen13: this is intentional, they are the same architecture

modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B/lightning_w4a16_nvfp4_4o6.yaml
jenchen13: intentional

modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml
jenchen13: intentional we want to skip MTP only

modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B-BF16/ptq/w4a16_nvfp4_4o6.yaml
jenchen13: all quantizers are disabled first with 
```
    - quantizer_name: '*'
      enable: false
 ```
 then select expert/mamba/lm head quantizers are enabled
 therefore MTP should be disabled since it was not explicitly enabled

File: modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml (hunk 4 above)
Comment (jenchen13) [reply modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml:41]: i think usually this `default_disabled` yaml is used in the beginning of a quant config .. to disable mtp.*output_layer*weight_quantizer I can explicitly add that in my lightning recipe

File: modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B-BF16/ptq/w4a16_nvfp4_4o6.yaml [hunk 9]
```diff
@@ -0,0 +1,72 @@
...
+    - quantizer_name: '*mixer.out_proj*input_quantizer'
+      enable: true
+      cfg: {$import: fp8_default}
+    # KV cache -> FP8.
+    - quantizer_name: '*[kv]_bmm_quantizer'
+      enable: true
+      cfg:
+        num_bits: e4m3
+    # MCore lm_head (output_layer) W4A16 4/6 weight-only.
+    - quantizer_name: 'output_layer.weight_quantizer'
```
jenchen13, line 67: @kevalmorabia97 removed the `*` wild cards from output_layer so it doesn't capture MTP layers

kevalmorabia97: What if we use `*language_model.*output_layer.weight_quantizer` like naming in all rules? So only language_model's submodules are picked up?

yueshen2016 (approved):

modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B-BF16/ptq/w4a16_nvfp4_4o6.yaml
kevalmorabia97: Its not just applicable to output_layer. Other layers in mtp also get quantized. For example
```
mtp.layers.0.mtp_model_layer.layers.0.self_attention.core_attention.k_bmm_quantizer TensorQuantizer((4, 3) bit fake per-tensor amax=1.52e+01 calibrator=MaxCalibrator quant)
mtp.layers.0.mtp_model_layer.layers.0.self_attention.core_attention.v_bmm_quantizer TensorQuantizer((4, 3) bit fake per-tensor amax=7.00e+00 calibrator=MaxCalibrator quant)
...
mtp.layers.0.mtp_model_layer.layers.1.mlp.experts.linear_fc1.weight_quantizer.0  StaticBlockScaleQuantizer((2, 1) bit fake block_sizes={-1: 16, 'type': 'static', 'scale_bits': (4, 3), 'four_over_six': 1}, amax=[0.0142, 0.1809](311808) calibrator=MseCalibrator quant)
mtp.layers.0.mtp_model_layer.layers.1.mlp.experts.linear_fc1.weight_quantizer.1  StaticBlockScaleQuantizer((2, 1) bit fake block_sizes={-1: 16, 'type': 'static', 'scale_bits': (4, 3), 'four_over_six': 1}, amax=[0.0129, 0.1934](311808) calibrator=MseCalibrator quant)
mtp.layers.0.mtp_model_layer.layers.1.mlp.experts.linear_fc1.weight_quantizer.2  StaticBlockScaleQuantizer((2, 1) bit fake block_sizes={-1: 16, 'type': 'static', 'scale_bits': (4, 3), 'four_over_six': 1}, amax=[0.0122, 0.1685](311808) calibrator=MseCalibrator quant)
```

kevalmorabia97: Actually `*language_model*.output_layer.weight_quantizer` wont work. From `.quant_summary.txt`, I see language model entries as `decoder.layers.*` so perhaps we need to use `*decoder.*output_layer.weight_quantizer` so nothing except LLM decoder is activated by the rules

jenchen13: output_layer.weight_quantizer works and excludes MTP layers

jenchen13: The LM head is not under decoder.* in the MCore quantizer namespace. The PTQ summary reports it exactly as output_layer.weight_quantizer; decoder.layers.* applies to the internal model layers. Therefore *decoder.*output_layer.weight_quantizer would not match the LM head at all. I changed the rule to the exact name output_layer.weight_quantizer, and verified in PTQ that it enables the root LM head without enabling any MTP-qualified output layer. 
-- codex

kevalmorabia97: I just ran MBridge PTQ with your recipe and still see same mtp layers quantized as I posted above. 
Also MTP doesnt actually have output_layer submodule

meenchen (approved):

kevalmorabia97 (approved):

modelopt_recipes/huggingface/models/nvidia/Nemotron-3.5-Lightning-30B-A3B-BF16/ptq/w4a16_nvfp4_4o6.yaml
jenchen13: fixed ty

PR #2149, opened 2026-08-11
Support fine-tuning released DFlash/DSpark drafters (causal SWA, attention sink, warm start)

h-guo18:
# Support fine-tuning released DFlash/DSpark drafters (causal SWA, attention sink, warm start)

### What does this PR do?

Type of change: New feature + bug fix

Adds what ModelOpt was missing to fine-tune an already-published DFlash/DSpark draft
model. The concrete target is
[`nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark`](https://huggingface.co/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark)
on its hybrid Mamba/attention/MoE base, but every change is generic.

Before this PR that checkpoint could not be trained faithfully — or even loaded: its
attention-sink tensors were dropped as unexpected keys, its block-causal attention had no
implementation, and its capture layers were silently overwritten with ModelOpt's defaults.

**New user-facing options** (all default to today's behavior, so existing runs are unchanged):

| Option | Values | Purpose |
| --- | --- | --- |
| `dflash_draft_attention` | `bidirectional` (default) / `causal` | Block-internal attention pattern. `causal` restricts a query at block position `i` to draft positions `<= i`. |
| `dflash_attention_sink` | `false` (default) / `true` | Learnable per-head `attention_sink_bias [num_heads]` on every draft layer — one extra logit appended before the softmax and dropped after, so a head can put probability mass nowhere instead of being forced to attend inside its window (the GPT-OSS formulation). |
| `dflash_init_checkpoint` | path | Warm-start the draft from an exported checkpoint instead of a random init. Any missing/unexpected/wrong-shaped tensor raises rather than warns. |
| `dflash_architecture_config.target_layer_ids` | list | Which base layers feed the draft's `fc`. Previously recomputed unconditionally with no override. |

**Bugs fixed along the way** (each one silently corrupts training rather than failing):

- The exporter hard-coded `dflash_config.causal: False` and only wrote it under SWA, so even
  a correctly-trained causal draft would be served non-causally. It now reflects the trained
  setting, and emits `attention_sink_bias` when enabled.
- `_build_generate_swa_mask` returned `None` whenever `swa_window_size` was unset, which
  would have dropped the causal structure at generation time while training used it.
- `target_layer_ids` was recomputed from the uniform default on every convert. The released
  drafter uses `[1,5,19,29,41,51]`; the default for a 52-layer base is `[1,11,20,30,39,49]`
  — *different layers*. Here it surfaced as a matmul shape error only because the plane
  counts disagreed; with a matching count it would have trained on the wrong features
  silently.
- The streaming dataset assumed the draft's aux layers all sit below the base's final layer
  (`aux = planes[:-1]`, `target = planes[-1]`). A draft whose top aux id *is* the final layer
  cannot get an extra plane — vLLM captures each layer once — so `final_aux_is_base_hidden`
  now lets the last plane serve both roles. It is derived from the model, not configured by
  hand.
- DSpark head weights load from either the flat layout ModelOpt exports (upstream DeepSpec
  convention) or the nested `markov_head.` layout the NVIDIA release uses. Without the remap
  the two `[131072, 512]` Markov tables — ~14% of the draft's parameters — stay randomly
  initialized while everything else warm-starts, with no error.
- `nemotron_h` is enabled in `_FINAL_NORM_TYPE_BY_MODEL_TYPE`: despite the hybrid stack,
  `NemotronHModel.norm_f` is a plain RMSNorm, and without the entry the offline/streaming
  fake base raises instead of reconstructing the distillation target.

### Usage

```yaml
dflash:
  dflash_init_checkpoint: /path/to/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark
  dflash_draft_attention: causal
  dflash_attention_sink: true
  dflash_swa_window_size: 1024
  dflash_block_size: 8
  dflash_mask_token_id: 990
  dflash_architecture_config:
    target_layer_ids: [1, 5, 19, 29, 41, 51]
```

A full worked example is at
`modelopt_recipes/general/speculative_decoding/dspark_nemotron35_warmstart.yaml`.

### Testing

**Unit tests** — 124 pass (`test_hf_dflash.py`, `test_hf_dspark.py`, `test_hf_domino.py`,
`test_hf_dflash_offline.py`, `test_modeling_final_norm.py`), 32 of them new: causal mask
structure (lower-triangular per block, no cross-block leakage, context visibility
unchanged), the sink math (degenerates to plain attention at `-inf`, absorbs mass
monotonically, receives gradient), warm-start load/reject paths, Markov key remapping, and
explicit `target_layer_ids`.

**Checkpoint compatibility** — the released drafter loads with zero missing/unexpected keys
and zero shape mismatches; all 77 tensors (6 attention sinks and both Markov tables
included) match bit-exactly, and a training step runs with gradients reaching the sink and
Markov parameters.

**End-to-end streaming training** — Nemotron-3.5 base served by vLLM (1 node, TP8) feeding
8 trainer GPUs over NIXL; the draft warm-starts from the released checkpoint and trains with
`causal` + sink + SWA 1024. 128 Daring-Anteater conversations, 20 epochs (the plot shows the
first 5, where the trend is clearest — the curves flatten after that):

warm-start training curves

Over the first 5 epochs loss falls **1.85 → 1.36** and train accuracy rises
**0.25 → 0.49**; across the full 20 epochs they reach **1.21** and **0.48** (peak 0.54)
before flattening. This validates the pipeline end-to-end — capture layers, plane split,
mask direction, sink loading and warm-start weights all have to be right for this curve to
appear. It is *not* a model-quality result: 128 samples over 20 epochs overfits by
construction, and the corpus is not generated by the base model, so the absolute numbers are
not meaningful.

### TODO (follow-up)

**A complete, robust checkpoint/config converter.** Both conversions are handled ad hoc here:

- *Draft config → training config.* The recipe transcribes ~15 fields by hand from the
  drafter's `config.json`. Only the shape-bearing ones (`num_hidden_layers`,
  `num_attention_heads`, `intermediate_size`, `markov_rank`) fail loudly when mistyped; the
  rest — `mask_token_id`, `causal`, `swa_window_size`, `block_size` — train "successfully" on
  a wrong value and only surface later as a mysteriously low acceptance length. A converter
  should derive the whole block from the checkpoint, including its aliases (`pard_token`,
  `dspark_markov_rank`, `dflash_query_causal`, top-level `sliding_window` /
  `attention_sink_bias`) and duplicated fields.
- *Weight layout.* The `markov_head.` remap is a load-time hook. A converter should normalize
  layouts explicitly, and decide whether export should also emit the release's aliases so a
  round-trip reproduces the original format (today it renames `architectures` to
  `DFlashDraftModel`).
- *Base config.* Serving this base on vLLM needs its `config.json` layer-type vocabulary
  updated for the transformers-5 path (`mamba` → `linear_attention`, `attention` →
  `full_attention`, plus a matching `hybrid_override_pattern`). That is done by hand today and
  is not covered by this PR.

- **Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/CONTRIBUTING.md)** and your commits are signed.
- **Is this change backward compatible?**: Yes
- **Did you write any new necessary tests?**: Yes
- **Did you add or update any necessary documentation?**: Yes
- **Did you update [Changelog](https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/CHANGELOG.rst)?**: No

## Summary by CodeRabbit

* **New Features**
  * Added configurable causal or bidirectional attention for DFlash models.
  * Added optional attention sinks, checkpoint warm starts, and explicit target-layer selection.
  * Improved streaming data handling for shared auxiliary and base hidden states.
  * Added Nemotron-3.5 Lightning DSpark warm-start training and serving recipes.

* **Bug Fixes**
  * Preserved configured attention behavior during model export.
  * Prevented warm-start checkpoints from being reapplied during restoration.
  * Improved checkpoint compatibility, validation, and attention-mask handling.

### DIFF
--- a/examples/speculative_decoding/main.py
+++ b/examples/speculative_decoding/main.py
@@ -291,6 +291,7 @@ def train():
         train_len=training_args.training_seq_len,
         answer_only_loss=training_args.answer_only_loss,
         shift_labels=not is_dflash,
+        final_aux_is_base_hidden=recipe.data.final_aux_is_base_hidden,
     )
 
     callbacks = [EagleTrainingPlot(training_args.ar_validate_steps, training_args.estimate_ar)]
--- a/modelopt/torch/export/plugins/hf_spec_export.py
+++ b/modelopt/torch/export/plugins/hf_spec_export.py
@@ -412,10 +412,10 @@ class DFlashExporter(SpeculativeDecodingExporter):
         else:
             config["layer_types"] = ["full_attention"] * draft_config.num_hidden_layers
 
-        # Sliding-window attention: all draft layers use non-causal SWA (MiMo-style). vLLM's
+        # Sliding-window attention: all draft layers use SWA. vLLM's
         # _resolve_layer_attention reads dflash_config.use_swa + swa_window_size; with
-        # layer_types left all "full_attention" it applies a non-causal sliding window to
-        # every draft layer (window from swa_window_size / top-level sliding_window).
+        # layer_types left all "full_attention" it applies a sliding window to every draft
+        # layer (window from swa_window_size / top-level sliding_window).
         swa_window = getattr(self.model, "dflash_swa_window_size", None)
         if swa_window is not None:
             config["sliding_window"] = swa_window
@@ -423,9 +423,22 @@ class DFlashExporter(SpeculativeDecodingExporter):
                 {
                     "use_swa": True,
                     "swa_window_size": swa_window,
-                    "causal": False,
                 }
             )
+
+        # Block-internal attention pattern. Emitted unconditionally (not just under SWA):
+        # vLLM's _dflash_layer_causal treats dflash_config.causal as an all-layer override,
+        # and its default differs per layer type, so writing it explicitly is what keeps
+        # inference consistent with how the draft was actually trained.
+        config["dflash_config"]["causal"] = (
+            getattr(self.model, "dflash_draft_attention", "bidirectional") == "causal"
+        )
+
+        # Learnable per-head attention sink. vLLM reads dflash_config.attention_sink_bias to
+        # decide whether to build the sink parameter and pass it to its attention kernel.
+        if getattr(self.model, "dflash_attention_sink", False):
+            config["dflash_config"]["attention_sink_bias"] = True
+            config["attention_sink_bias"] = True
 
         # Inject the export-time YaRN rope_scaling from the dflash_export_rope_scaling
         # config field (empty dict disables). Mirrors eagle's eagle_export_rope_scaling.
--- a/modelopt/torch/speculative/dflash/conversion.py
+++ b/modelopt/torch/speculative/dflash/conversion.py
@@ -80,4 +80,7 @@ def restore_dflash_model(
 ) -> nn.Module:
     """Function for restoring a previously converted model to a DFlash model."""
     assert not metadata, "No metadata expected!"
+    # Warm start is a one-time training-time init; the restored weights would overwrite it
+    # anyway, and replaying it on the meta device would raise.
+    config = config.model_copy(update={"dflash_init_checkpoint": None})
     return convert_to_dflash_model(model, config)[0]
--- a/modelopt/torch/speculative/plugins/hf_dflash.py
+++ b/modelopt/torch/speculative/plugins/hf_dflash.py
@@ -72,6 +72,7 @@ Draft model components:
 """
 
 import logging
+from pathlib import Path
 from typing import Any
 
 import torch
@@ -441,15 +442,45 @@ class HFDFlashModel(DFlashModel):
             self.dflash_config.hidden_size // self.dflash_config.num_attention_heads,
         )
         self.dflash_config.block_size = self.dflash_block_size
+        # On the draft config so _build_draft_module stays a pure function of it.
+        self.dflash_config.attention_sink_bias = self.dflash_attention_sink
 
-        # Target layer IDs
+        # Which base layers feed the draft's `fc`: explicit ids win, else the uniform default.
         num_target_layers = (
             base_config.num_orig_hidden_layers
             if self.dflash_offline
             else base_config.num_hidden_layers
         )
         num_draft_layers = self.dflash_config.num_hidden_layers
-        self.target_layer_ids = build_target_layer_ids(num_target_layers, num_draft_layers)
+        user_target_layer_ids = config.dflash_architecture_config.get("target_layer_ids")
+        if user_target_layer_ids:
+            if len(user_target_layer_ids) != num_draft_layers:
+                raise ValueError(
+                    f"dflash_architecture_config.target_layer_ids has "
+                    f"{len(user_target_layer_ids)} entries but the draft has "
+                    f"{num_draft_layers} layers; one target layer per draft layer is required."
+                )
+            if max(user_target_layer_ids) >= num_target_layers:
+                raise ValueError(
+                    f"dflash_architecture_config.target_layer_ids {user_target_layer_ids} "
+                    f"references a layer beyond the base model's {num_target_layers} layers."
+                )
+            if len(set(user_target_layer_ids)) != len(user_target_layer_ids):
+                raise ValueError(
+                    f"dflash_architecture_config.target_layer_ids {user_target_layer_ids} "
+                    "contains duplicates; each draft layer needs a distinct capture layer "
+                    "(the streaming producer captures each base layer at most once)."
+                )
+            # forward() indexes hidden_states[lid + 1], where a negative id silently wraps.
+            if min(user_target_layer_ids) < 0:
+                raise ValueError(
+                    f"dflash_architecture_config.target_layer_ids {user_target_layer_ids} "
+                    "must be non-negative base-layer indices."
+                )
+            self.target_layer_ids = list(user_target_layer_ids)
+            logger.info("DFlash: using explicit target_layer_ids %s", self.target_layer_ids)
+        else:
+            self.target_layer_ids = build_target_layer_ids(num_target_layers, num_draft_layers)
         self.dflash_config.target_layer_ids = self.target_layer_ids
 
         # mask_token_id: validated by DFlashConfig, auto-detected from tokenizer context
@@ -466,6 +497,10 @@ class HFDFlashModel(DFlashModel):
         # Factory hook: subclasses (e.g. Domino) override to build an augmented
         # draft module while reusing all of DFlash's modify() setup.
         self.dflash_module = self._build_draft_module(self.dflash_config)
+        # Warm start from an exported draft checkpoint, before the dtype/device move below
+        # so the loaded tensors get cast alongside the rest of the module.
+        if self.dflash_init_checkpoint:
+            self._load_init_checkpoint(self.dflash_init_checkpoint)
         # Match base model dtype/device. Skip if base is on meta (during from_pretrained
         # restore — the model will be moved to the correct device after weight loading).
         if self.dflash_offline:
@@ -485,6 +520,81 @@ class HFDFlashModel(DFlashModel):
     def _build_draft_module(self, dflash_config):
         """Build the draft module. Subclasses override to use an augmented module."""
         return DFlashModule(dflash_config)
+
+    # Draft-module entries that legitimately come from the base model rather than the
+    # exported draft checkpoint, so their absence (or presence) is not an error.
+    _INIT_CKPT_IGNORED_KEYS = ("embed_tokens.weight", "lm_head.weight")
+
+    def _load_init_checkpoint(self, path: str):
+        """Warm-start ``self.dflash_module`` from an exported draft checkpoint.
+
+        Accepts either the export directory (containing ``model.safetensors``) or the
+        safetensors file itself. The architecture is fixed by ``dflash_architecture_config``
+        at this point, so the checkpoint has to match it: any missing, unexpected, or
+        wrong-shaped tensor raises. Loading part of a draft and leaving the rest randomly
+        initialized looks like a warm start but trains from a corrupted starting point, so
+        it is rejected instead of warned about.
+        """
+        from safetensors.torch import load_file
+
+        ckpt = Path(path)
+        if ckpt.is_dir():
+            ckpt = ckpt / "model.safetensors"
+        if not ckpt.is_file():
+            raise FileNotFoundError(
+                f"dflash_init_checkpoint: no draft weights at {ckpt}. Expected an exported "
+                "draft directory containing model.safetensors, or the file itself."
+            )
+
+        state_dict = load_file(str(ckpt))
+        # Tolerate a `dflash_module.` prefix so a raw training checkpoint also works.
+        state_dict = {
+            (k.split("dflash_module.", 1)[1] if "dflash_module." in k else k): v
+            for k, v in state_dict.items()
+        }
+        state_dict = {
+            k: v
+            for k, v in state_dict.items()
+            if k not in self._INIT_CKPT_IGNORED_KEYS and "rotary_emb" not in k
+        }
+
+        # Shape-check against the module's own view of each key. Subclasses may remap keys
+        # on load (DSpark accepts a nested ``markov_head.`` layout), so resolve through the
+        # same hooks first — otherwise a wrong-shaped remapped tensor would skip this check
+        # and fail later with a much less obvious error.
+        module_sd = self.dflash_module.state_dict()
+        resolved = dict(state_dict)
+        for hook in self.dflash_module._load_state_dict_pre_hooks.values():
+            hook(resolved, "", None, True, [], [], [])
+        mismatched = [
+            f"{k}: checkpoint {tuple(v.shape)} vs module {tuple(module_sd[k].shape)}"
+            for k, v in resolved.items()
+            if k in module_sd and v.shape != module_sd[k].shape
+        ]
+        if mismatched:
+            raise ValueError(
+                "dflash_init_checkpoint: shape mismatch between "
+                f"{ckpt} and the configured draft architecture:\n  " + "\n  ".join(mismatched)
+            )
+
+        # strict=False, then check by hand: the module's own load hooks (e.g. DSpark's
+        # markov_head remap) run first, and buffers such as rotary_emb are excluded above.
+        incompatible = self.dflash_module.load_state_dict(state_dict, strict=False)
+        missing = [
+            k
+            for k in incompatible.missing_keys
+            if "rotary_emb" not in k and k not in self._INIT_CKPT_IGNORED_KEYS
+        ]
+        if missing or incompatible.unexpected_keys:
+            raise ValueError(
+                f"dflash_init_checkpoint: {ckpt} does not match the configured draft "
+                "architecture.\n"
+                f"  missing from checkpoint: {sorted(missing)}\n"
+                f"  unexpected in checkpoint: {sorted(incompatible.unexpected_keys)}"
+            )
+        logger.info(
+            "DFlash: warm-started draft module from %s (%d tensors).", ckpt, len(state_dict)
+        )
 
     def get_exporter(self):
         """Get the exporter for the DFlash draft model."""
@@ -567,13 +677,18 @@ class HFDFlashModel(DFlashModel):
     def _build_draft_attention_mask(
         self, seq_len, anchor_positions, block_keep_mask, n_blocks, dtype, device, window=None
     ):
-        """Build SDPA attention mask: context (causal) + draft (bidirectional within block).
+        """Build SDPA attention mask: context (causal) + draft (per ``dflash_draft_attention``).
 
-        When ``window`` is not None, all layers use non-causal sliding-window attention
-        (MiMo-style): each draft query only sees context positions within ``window`` tokens
-        before its own position. Block-internal attention stays bidirectional and is left
-        un-windowed (the config enforces ``window >= block_size``, so a full block always
-        fits inside the window and windowing it would be a no-op).
+        When ``window`` is not None, all layers use sliding-window attention: each draft
+        query only sees context positions within ``window`` tokens before its own position.
+        Block-internal attention is left un-windowed (the config enforces
+        ``window >= block_size``, so a full block always fits inside the window and windowing
+        it would be a no-op).
+
+        Block-internal visibility follows ``self.dflash_draft_attention``:
+        ``"bidirectional"`` (default, MiMo-style) lets every query see the whole block, while
+        ``"causal"`` restricts a query at block position ``i`` to draft positions ``<= i`` so
+        the block is modelled autoregressively.
         """
         bsz = anchor_positions.shape[0]
         block_size = self.dflash_block_size
@@ -598,6 +713,12 @@ class HFDFlashModel(DFlashModel):
         is_draft = kv_indices >= seq_len
         kv_block_ids = (kv_indices - seq_len) // block_size
         mask_draft = is_draft & (q_block_ids == kv_block_ids)
+        if self.dflash_draft_attention == "causal":
+            # Autoregressive within the block: query at block position i sees draft
+            # positions <= i only. Compare positions *within* the block so the term is
+            # independent of which block the query belongs to.
+            kv_pos_in_block = (kv_indices - seq_len) % block_size
+            mask_draft = mask_draft & (kv_pos_in_block <= (q_indices % block_size))
         # Valid block
         valid_block = block_keep_mask.view(bsz, 1, n_blocks, 1).repeat_interleave(block_size, dim=2)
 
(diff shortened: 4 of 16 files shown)


examples/speculative_decoding/main.py
```diff
@@ -285,12 +285,41 @@ def train():
...
+    )
+    _base_depth = getattr(_base_cfg, "num_orig_hidden_layers", None) or getattr(
+        _base_cfg, "num_hidden_layers", 0
+    )
+    final_aux_is_base_hidden = bool(
+        is_dflash
+        and getattr(model, "target_layer_ids", None)
+        and _base_depth
+        and max(model.target_layer_ids) + 1 >= _base_depth
+    )
```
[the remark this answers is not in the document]
h-guo18: Not applying this one — `target_layer_ids` are zero-based, so `>=` is correct here.

The validation a few lines up in `hf_dflash.py` is `max(user_target_layer_ids) >= num_target_layers`, i.e. the largest legal id on a 52-layer base is **51**, not 52. The recipe matches: `target_layer_ids: [1, 5, 19, 29, 41, 51]`. So a top aux id of 51 on a 52-layer base *is* the final layer, and `max + 1 >= depth` firing there is the intended behavior — switching to `==` would keep the same truth value for the in-range case while turning the out-of-range case into a silent `False` instead of the guard it is today.

The reason this looked one-based is a genuine error, just in the comment rather than the code: it cited the released drafter's ids as `[2,6,20,30,42,52]`. Fixed in 1218d1cf2d to `[1,5,19,29,41,51]`, explicitly marked zero-based.

modelopt/torch/export/plugins/hf_spec_export.py
```diff
@@ -412,21 +412,34 @@ def _export_config(self):
...
+        # inference consistent with how the draft was actually trained.
+        config["dflash_config"]["causal"] = (
+            getattr(self.model, "dflash_draft_attention", "bidirectional") == "causal"
+        )
+
+        # Learnable per-head attention sink. vLLM reads dflash_config.attention_sink_bias to
+        # decide whether to build the sink parameter and pass it to its attention kernel.
+        if getattr(self.model, "dflash_attention_sink", False):
+            config["dflash_config"]["attention_sink_bias"] = True
+            config["attention_sink_bias"] = True
```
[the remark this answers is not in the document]
h-guo18: Leaving the top-level key in place.

The nested `dflash_config["attention_sink_bias"]` — the one vLLM actually reads — is already written on the line above, so the top-level key is redundant rather than wrong: nothing consumes it, and removing it changes no behavior on the vLLM side.

It's there for consistency with how `sliding_window` is emitted ~25 lines up, which writes both the top-level key and the nested `dflash_config` entry for the same reason. Keeping the two sink/SWA blocks structurally identical is worth more than dropping one unused line.

If it should instead be `add_swa_attention_sink_bias` (the fallback name), that's a real question — but it depends on the target vLLM version's `_dflash_layer_causal` / sink-resolution path, which I'd rather pin against an actual vLLM commit than a search result. Not blocking this PR.

modelopt/torch/speculative/plugins/modeling_dflash.py
```diff
@@ -205,21 +217,70 @@ def forward(self, hidden_states, target_hidden, position_embeddings, attention_m
...
+                # The eager sink path applies only the caller-supplied mask; a per-layer
+                # window from config.layer_types would be silently dropped. DFlash windows
+                # the context through the attention mask instead (dflash_swa_window_size),
+                # so this combination is rejected rather than trained with the wrong mask.
+                raise NotImplementedError(
+                    "dflash_attention_sink is not supported together with a per-layer "
+                    "sliding window from dflash_architecture_config.layer_types. Use "
+                    "dflash_swa_window_size for the draft's sliding window instead."
+                )
+            attn_output = self._sink_attention(q, k, v, attention_mask)
```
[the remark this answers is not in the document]
h-guo18: Fair point, but leaving it in `forward` for now.

Both operands are indeed fixed at construction, so hoisting the check to `__init__` would turn it into a construction-time contract and stop re-evaluating a constant per layer per step. The cost of the current placement is small though — a `None` check on an attribute — and the failure it guards is a hard `NotImplementedError` either way, so this is readability rather than correctness.

Deferring because moving it means constructing the sink parameter and resolving `self.sliding_window` in a specific order in `__init__`, which touches the DFlash/DSpark/Domino subclass init path that this PR is otherwise not changing. Worth doing alongside the config→training converter already listed as follow-up in the PR description.

examples/speculative_decoding/main.py
```diff
@@ -285,12 +285,41 @@ def train():
...
+        or model.config
+    )
+    _base_depth = getattr(_base_cfg, "num_orig_hidden_layers", None) or getattr(
+        _base_cfg, "num_hidden_layers", 0
+    )
+    final_aux_is_base_hidden = bool(
+        is_dflash
+        and getattr(model, "target_layer_ids", None)
+        and _base_depth
+        and max(model.target_layer_ids) + 1 >= _base_depth
```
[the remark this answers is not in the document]
h-guo18: Agreed in principle, deferring to the follow-up.

The duplication is real: this block re-derives `target_layer_ids` and the base depth (including the `num_orig_hidden_layers` fallback) that `HFDFlashModel.modify` already resolved, and any entry point not going through this script silently gets `final_aux_is_base_hidden=False`.

Not doing it here because a `final_aux_is_base_hidden` property on `HFDFlashModel` has to resolve the base config the same way `modify()` does, and `modify()` currently does that inline against a local `base_config` — so the property either duplicates that resolution again or `modify()` gets refactored to stash it. That's the same cleanup as the "complete, robust checkpoint/config converter" item in the PR description, and I'd rather do it in one pass with a unit test for the property than bolt a third copy of the resolution on now.

The comment cited the wrong (one-based) layer ids, which I've fixed in 1218d1cf2d — that was the actually-misleading part.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -466,6 +492,10 @@ def modify(self, config):
         # Factory hook: subclasses (e.g. Domino) override to build an augmented
         # draft module while reusing all of DFlash's modify() setup.
         self.dflash_module = self._build_draft_module(self.dflash_config)
+        # Warm start from an exported draft checkpoint, before the dtype/device move below
+        # so the loaded tensors get cast alongside the rest of the module.
+        if self.dflash_init_checkpoint:
+            self._load_init_checkpoint(self.dflash_init_checkpoint)
```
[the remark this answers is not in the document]
h-guo18: Good catch — fixed in 94ba8c9ed7. This was the one real bug in the batch.

Confirmed the mechanism: `restore_dflash_model` → `convert_to_dflash_model` → `modify(config)`, with `dflash_init_checkpoint` serialized into `modelopt_state`, so `_load_init_checkpoint` re-ran on every restore against the training-time path. And it runs *before* the `base_device.type != "meta"` guard, so the meta case isn't covered either.

Took the suggested fix (`config.model_copy(update={"dflash_init_checkpoint": None})` in the restore entrypoint) — verified `model_copy` returns a `DFlashConfig` and leaves the caller's config untouched.

Also added `test_restore_does_not_replay_warm_start`, which does exactly the scenario you described: convert with a warm start, save, `rmtree` the warm-start directory, then restore and compare every draft parameter against the pre-save model. As you noted, `TestInitCheckpoint` only covered the convert path.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -441,15 +442,40 @@ def modify(self, config):
...
+                raise ValueError(
+                    f"dflash_architecture_config.target_layer_ids has "
+                    f"{len(user_target_layer_ids)} entries but the draft has "
+                    f"{num_draft_layers} layers; one target layer per draft layer is required."
+                )
+            if max(user_target_layer_ids) >= num_target_layers:
+                raise ValueError(
+                    f"dflash_architecture_config.target_layer_ids {user_target_layer_ids} "
+                    f"references a layer beyond the base model's {num_target_layers} layers."
+                )
```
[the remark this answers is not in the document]
h-guo18: Fixed in 5547b3594e — added the lower-bound check next to the existing upper-bound one, plus `test_negative_id_raises`.

Confirmed the failure mode: `forward()` indexes `raw_outputs.hidden_states[lid + 1]`, so e.g. `-2` wraps to a real layer and the draft silently trains on features from a layer it was never meant to see, rather than raising.

modelopt/torch/speculative/plugins/modeling_dspark.py
```diff
@@ -120,6 +120,30 @@ def __init__(self, config):
...
+        under a ``markov_head.`` parent module. Rewriting those keys in place here lets
+        either layout load without changing the export format (which would break drafters
+        already trained and deployed with the flat names).
+        """
+        nested = prefix + cls._NESTED_HEAD_PREFIX
+        for key in [k for k in state_dict if k.startswith(nested)]:
+            flat = prefix + key[len(nested) :]
+            # A flat key already present wins: never clobber an explicit match.
+            state_dict.setdefault(flat, state_dict.pop(key))
+            state_dict.pop(key, None)
```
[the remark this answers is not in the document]
h-guo18: Both points are right — fixed in d5a53ee0c5.

Verified the dead line by executing the loop body: `setdefault(flat, state_dict.pop(key))` evaluates its argument first, so `key` is already removed and the following `state_dict.pop(key, None)` returns `None` every time (`flat != key` always holds, since the nested prefix is non-empty).

Also took the logging suggestion for the collision case. The behavior was already correct (flat wins), but silently dropping one of two `[131072, 512]` Markov tables — ~14% of the draft's parameters, invisible at runtime if wrong — is worth a line. Added a module-level `logger = logging.getLogger(__name__)`, matching `hf_dflash.py` / `hf_dspark.py` / `hf_domino.py`.

Checked all three paths still behave: nested-only promotes to flat, collision keeps flat and warns, and a non-empty `prefix` still resolves correctly.

modelopt/torch/speculative/plugins/modeling_dflash.py
```diff
@@ -205,21 +217,70 @@ def forward(self, hidden_states, target_hidden, position_embeddings, attention_m
...
+            # [B, 1, Q, KV] additive mask, already sliced to the kv length by the caller.
+            attn_weights = attn_weights + attention_mask[..., : k.shape[-2]]
+
+        sinks = sink_bias.view(1, -1, 1, 1).expand(
+            attn_weights.shape[0], -1, attn_weights.shape[-2], -1
+        )
+        combined = torch.cat([attn_weights, sinks.to(attn_weights.dtype)], dim=-1)
+        combined = combined - combined.amax(dim=-1, keepdim=True)
+        probs = F.softmax(combined, dim=-1, dtype=torch.float32).to(q.dtype)
+        attn_weights = probs[..., :-1]  # drop the sink column
```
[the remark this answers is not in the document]
h-guo18: I checked this one carefully because the math is right, but the memory conclusion doesn't hold — measuring it, the proposed version retains **more** than the current one, so I'm not applying it.

**The math is equivalent.** Folding the sink into the denominator reproduces `softmax([logits, sink])[..., :-1]` to fp32 rounding: max-diff `1.19e-07` in fp32 and `3.9e-03` in bf16 (one bf16 ulp at these magnitudes), and a fully-masked `-inf` row normalizes to zeros without NaN in both versions. No disagreement there.

**But it costs memory rather than saving it.** Walking the autograd graph and summing the tensors actually retained for backward, at the recipe's shape `[1, 32, 512, 2560]` in bf16:

| | per layer | × 6 draft layers |
| --- | --- | --- |
| current (`cat` + `F.softmax(dtype=fp32)`) | 0.25 GB | 1.51 GB |
| proposed (manual fp32 exp/divide) | 0.50 GB | 3.02 GB |

The reason is that the four tensors in the estimate aren't all live. `F.softmax(..., dtype=torch.float32)` fuses the fp32 upcast, max-subtraction, exp and normalization into a single autograd node that saves only its **bf16 output** — the fp32 `probs` and the `combined - amax` intermediate are transient, never retained. Writing the same math as separate `.to(float32)` / `.exp()` / `.sum()` / `/` ops turns those intermediates into graph nodes that each save an fp32 operand, which is what doubles the footprint. The `cat` adds one column (`2560 → 2561`, +0.04%), and the `[..., :-1]` slice is a view.

So the current form is already the memory-lean one, and it's the formulation `transformers`' GPT-OSS `eager_attention_forward` uses — worth staying aligned with, since that's where the semantics come from.

Agreed on the underlying point that the sink forces this layer off SDPA and that eager `[B, H, Q, KV]` weights dominate draft-layer memory. `flex_attention` with a `score_mod` is the fix that actually addresses that, since it keeps the fused kernel and never materializes the matrix — noting it as follow-up rather than doing it here, as it needs a compile-path/backend-support check across the versions this has to run on.

modelopt/torch/speculative/plugins/hf_dflash.py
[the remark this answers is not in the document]
h-guo18: The duplicate case is a good catch — added in a7feba7187, along with `test_duplicate_ids_raise`.

You're right that it's the one failure mode here that stays silent. Traced it: `[1, 5, 5, 29, 41, 51]` passes the count check, so `fc`'s input width is correct, and passes the range check, so the `hidden_states[lid + 1]` gather in `hf_dspark.py:354` / `hf_domino.py:327` succeeds — layer 5 twice, one capture point gone, training proceeds normally. That's exactly the class the PR description calls out.

The negative case landed separately in 5547b3594e (CodeRabbit raised it on the same line), so both bounds are now covered.

modelopt/torch/speculative/plugins/hf_dflash.py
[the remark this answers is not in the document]
h-guo18: Fixed in 94ba8c9ed7 — this matches the finding Claude raised on the same line, and it was the one real bug in this review round.

Confirmed the path: `restore_dflash_model` → `convert_to_dflash_model` → `modify(config)`, with `dflash_init_checkpoint` persisted in `modelopt_state`, so the warm start re-ran on every restore against the training-time path — and it sits *before* the `base_device.type != "meta"` guard, so the `from_pretrained` meta case was unguarded too.

Rather than checking the base device, the fix clears the field on the restore entrypoint: warm start is a one-time initialization, not part of the converted model's definition, so restore should never replay it regardless of device. Added `test_restore_does_not_replay_warm_start`, which saves a warm-started model, deletes the warm-start directory, and restores.

ChenhanYu (approved):

examples/speculative_decoding/main.py
```diff
@@ -285,12 +285,42 @@ def train():
...
+        f"draft target_layer_ids={getattr(model, 'target_layer_ids', None)}, "
+        f"final_aux_is_base_hidden={final_aux_is_base_hidden}"
+    )
     data_module = make_speculative_data_module(
         tokenizer,
         recipe.data,
         train_len=training_args.training_seq_len,
         answer_only_loss=training_args.answer_only_loss,
         shift_labels=not is_dflash,
+        final_aux_is_base_hidden=final_aux_is_base_hidden,
```
h-guo18: resolve logic here is over complicated.  lets just exposed it as a manual set arg in yaml

modelopt/torch/speculative/dflash/conversion.py
```diff
@@ -80,4 +80,12 @@ def restore_dflash_model(
 ) -> nn.Module:
     """Function for restoring a previously converted model to a DFlash model."""
     assert not metadata, "No metadata expected!"
+    # Warm start is a one-time training-time initialization, not part of the converted
```
h-guo18: trim this comment to 1 line

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -441,15 +442,59 @@ def modify(self, config):
...
+            # Duplicates pass both checks above -- the count is right, so `fc` gets the
```
h-guo18: remove this comment. code is self explanatory.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -441,15 +442,59 @@ def modify(self, config):
...
+            # features twice and dropping a capture point entirely. Streaming makes it
+            # worse: the producer captures each base layer at most once, so a duplicated id
+            # also yields fewer planes than `fc` expects.
+            if len(set(user_target_layer_ids)) != len(user_target_layer_ids):
+                raise ValueError(
+                    f"dflash_architecture_config.target_layer_ids {user_target_layer_ids} "
+                    "contains duplicates; each draft layer needs a distinct capture layer "
+                    "(the streaming producer captures each base layer at most once)."
+                )
+            # Reject negatives explicitly: forward() indexes hidden_states[lid + 1], where a
```
h-guo18: trim to 1 line

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -441,15 +442,59 @@ def modify(self, config):
             self.dflash_config.hidden_size // self.dflash_config.num_attention_heads,
         )
         self.dflash_config.block_size = self.dflash_block_size
-
-        # Target layer IDs
+        # Attention-shape knobs the draft module needs at build time. Carried on the draft
```
h-guo18: trim to 1line

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -441,15 +442,59 @@ def modify(self, config):
...
+        # config (not read from DFlashConfig) so _build_draft_module stays a pure function
+        # of dflash_config and restored checkpoints rebuild an identical module.
+        self.dflash_config.attention_sink_bias = self.dflash_attention_sink
+
+        # Target layer IDs: which base layers feed the draft's feature-fusion `fc`.
```
h-guo18: 1line

modelopt/torch/speculative/plugins/modeling_dspark.py
```diff
@@ -120,6 +124,41 @@ def __init__(self, config):
         # existed, so initialize the new layers explicitly.
         self._init_head_weights(config)
 
+        self._register_load_state_dict_pre_hook(self._remap_nested_head_keys, with_module=False)
+
+    # Head submodules that some published checkpoints nest under a `markov_head.` parent.
+    _NESTED_HEAD_PREFIX = "markov_head."
+
+    @classmethod
+    def _remap_nested_head_keys(cls, state_dict, prefix, *args, **kwargs):
```
h-guo18: this shouldn't be here since it's for one module in one model. can we put it as a per-model special loader or hook under some per-model file?

modelopt/torch/speculative/plugins/modeling_final_norm.py
```diff
@@ -90,6 +90,9 @@ def extra_repr(self):
     "deepseek_v3": "rmsnorm",
     "kimi_k2": "rmsnorm",  # Kimi-K2 / K2-Thinking (DeepSeek-V3 arch) report model_type "kimi_k2"
     "kimi_k25": "rmsnorm",  # Kimi-K2.5 / K2.6 / K2.7 all report model_type "kimi_k25"
+    # Nemotron-H hybrid Mamba/attention/MoE (e.g. NVIDIA-Nemotron-3.5-Lightning-30B-A3B).
```
h-guo18: remove this comment

sychen52 (approved):

examples/speculative_decoding/main.py
h-guo18: Done in 297f543b3. `final_aux_is_base_hidden` is now a `data.*` recipe field (default `False`); the derivation in main.py is gone. The Nemotron-3.5 warm-start recipe sets it explicitly.

modelopt/torch/speculative/dflash/conversion.py
h-guo18: Done in 297f543b3 — trimmed to one sentence (restored weights overwrite it anyway; replaying on meta device would raise).

modelopt/torch/speculative/plugins/hf_dflash.py
h-guo18: Done in 297f543b3 — removed. The `raise` message below already states that each draft layer needs a distinct capture layer.

modelopt/torch/speculative/plugins/hf_dflash.py
h-guo18: Done in 297f543b3 — trimmed to one line.

modelopt/torch/speculative/plugins/hf_dflash.py
h-guo18: Done in 297f543b3 — trimmed to one line.

modelopt/torch/speculative/plugins/hf_dflash.py
h-guo18: Done in 297f543b3 — trimmed to one line.

modelopt/torch/speculative/plugins/modeling_dspark.py
h-guo18: Done in 297f543b3. Moved to a new `modeling_draft_checkpoint_compat.py`, alongside the other per-model tables (`modeling_final_norm.py`); `modeling_dspark.py` now only registers the hook. A future checkpoint quirk adds an entry there instead of touching the module that defines the head.

modelopt/torch/speculative/plugins/modeling_final_norm.py
h-guo18: Done in 297f543b3 — removed.

modelopt/torch/speculative/plugins/modeling_dflash.py
```diff
@@ -44,20 +44,75 @@
...
+
+    Falls back to the uncompiled function if ``torch.compile`` is unavailable in this
+    environment; the two are numerically interchangeable, so this only costs speed.
+    """
+    global _compiled_sink_attention
+    if _compiled_sink_attention is None:
+        try:
+            _compiled_sink_attention = torch.compile(_sink_attention_impl)
+        except Exception:
+            _compiled_sink_attention = _sink_attention_impl
```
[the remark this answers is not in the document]
h-guo18: Correct, and fixed in 8825052ae — thanks. Confirmed `torch.compile` is lazy, so the guard around `torch.compile()` could never fire.

I did not take the suggested catch-and-retry at the call site, though: it cannot distinguish a compiler failure from a genuine error raised by the attention math, so a real bug would be silently downgraded to a slow path. `torch._dynamo.config.suppress_errors` already does exactly this correctly (verified: real exceptions still propagate, compiler errors fall back to eager), so the dead guard is simply removed and the docstring points at that knob.

Issue #2160, opened 2026-08-12
[Bug] init_quantized_weights / --low_memory_mode exports numerically broken NVFP4 (quantizes meta tensors before weights load)


spped2000:
## Summary

`init_quantized_weights` (public API, exported in `modelopt/torch/quantization/plugins/accelerate.py:__all__`, and the mechanism behind `hf_ptq.py --low_memory_mode`) produces an NVFP4 checkpoint that is silently wrong: half-sized `weight_scale` tensors, dequantization cosine similarity **0.756** against the source BF16 (healthy NVFP4 is >0.99), and a checkpoint vLLM refuses to load.

There is no error at export time. The run completes and prints `Quantized model exported to: ...`.

## Root cause

In `patched_from_pretrained` (accelerate.py, current main a21173a4):

```python
with init_empty_weights():
    model = cls.from_config(config, dtype=torch_dtype)   # L230 — meta tensors

mtq.quantize(model, quant_cfg)                            # L232 — quantizes META
mtq.compress(model, config=mtq.CompressConfig(...))       # L233 — packs META
_device_map = get_model_device_map(model, gpu_mem_percentage)

return load_checkpoint_and_dispatch(                      # L236 — real weights load
    model, checkpoint=..., device_map=_device_map, *args, **kwargs,
)
```

Quantization and compression run against `init_empty_weights()` meta tensors, i.e. before any real weight values exist. Calibration/absmax has nothing to measure, and `mtq.compress` has already replaced each `Linear.weight` with a packed uint8 buffer — so `load_checkpoint_and_dispatch` then loads real BF16 values into buffers whose shape and scale metadata describe the *packed* view.

Observable consequence on a Qwen2 model: `k_proj.weight_scale` is exported as `[128, 28]` where NVFP4 block-16 over 896 in-features requires `[128, 56]` — the scale grid was computed over 448 packed **bytes** instead of 896 logical values. vLLM aborts at load:

```
RuntimeError: start (0) + length (304) exceeds dimension size (152).
```

Dequantizing the export by hand and comparing to the BF16 source gives cosine **0.756** (lo/hi nibble order; 0.015 with nibbles swapped, 0.758 assuming block-16-truncated) — the numbers are damaged at quantization time, not at load time, so no loader-side fix helps.

## Reproduction (one command, 0.5B model — no large hardware needed)

```bash
[...]

PR #2161, opened 2026-08-12
Fix init_quantized_weights: model kwargs leak into dispatch, missing tie_weights()

spped2000:
Two small defects on the `init_quantized_weights` path (public API; also what `hf_ptq.py --low_memory_mode` uses). Both fire before any weight is written, so the path cannot complete at all on common models.

### 1. Model-construction kwargs reach `load_checkpoint_and_dispatch()`

`patched_from_pretrained` forwards `**kwargs` verbatim, so `attn_implementation` (documented in `hf_ptq.py`'s own CLI) raises:

```
TypeError: load_checkpoint_and_dispatch() got an unexpected keyword argument 'attn_implementation'
```

It now goes to `cls.from_config()`, where a construction kwarg belongs.

### 2. `tie_weights()` is never called before quantization

Tied parameters such as `lm_head.weight` (`tie_word_embeddings: true` — e.g. Qwen2.5-0.5B) are absent from the checkpoint, so they stay on meta and `dispatch_model()` raises:

```
NotImplementedError: Cannot copy out of meta tensor; no data!
```

accelerate documents `tie_weights()` as a prerequisite of `load_checkpoint_and_dispatch()`.

### Reproduction

`Qwen/Qwen2.5-0.5B-Instruct` (local dir), `--qformat nvfp4 --low_memory_mode`, modelopt 0.43.0, NGC `nvcr.io/nvidia/vllm:26.05.post1-py3`, GB10/SM121 aarch64. Failure 1 fires immediately; with it patched, failure 2 fires at dispatch.

### Scope — please read alongside #2160

These two fixes let the path **run to completion**, but the resulting checkpoint is still numerically wrong: quantization and compression execute on `init_empty_weights()` meta tensors *before* real weights load, giving half-sized `weight_scale` and dequant cosine 0.756 vs the BF16 source. That root cause is filed separately as #2160 and is not addressed here — I kept this PR to the two mechanical bugs so it can be reviewed independently.

---

*Disclosure: prepared with assistance from Claude (Anthropic); all failures above were reproduced on real hardware before writing the patch.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quantized model loading with attention implementation and data type settings.
  * Added reliable data type resolution, including explicit, legacy, configured, and default values.
  * Ensured data type options are applied correctly during checkpoint loading.
  * Improved initialization of tied model parameters before checkpoint dispatch.
  * Prevented unsupported loading options from causing checkpoint-loading failures.
  * Improved reliability when constructing models from configuration before loading and dispatching checkpoints.

### DIFF
--- a/modelopt/torch/quantization/plugins/accelerate.py
+++ b/modelopt/torch/quantization/plugins/accelerate.py
@@ -224,12 +224,21 @@ def init_quantized_weights(
 
         # Model-construction kwargs must not reach load_checkpoint_and_dispatch(),
         # which does not accept them (e.g. attn_implementation -> TypeError).
+        # Both dtype aliases are popped: torch_dtype is not a parameter of
+        # load_checkpoint_and_dispatch() either, and the resolved value is
+        # passed back explicitly below.
         attn_implementation = kwargs.pop("attn_implementation", None)
+        dtype_kwarg = kwargs.pop("dtype", None)
+        legacy_dtype_kwarg = kwargs.pop("torch_dtype", None)
 
         with init_empty_weights():
             # Fix torch_dtype to match original model
-            torch_dtype = kwargs.get(
-                "dtype", kwargs.get("torch_dtype", getattr(config, "torch_dtype", torch.float16))
+            torch_dtype = (
+                dtype_kwarg
+                if dtype_kwarg is not None
+                else legacy_dtype_kwarg
+                if legacy_dtype_kwarg is not None
+                else getattr(config, "torch_dtype", torch.float16)
             )
             from_config_kwargs = {}
             if attn_implementation is not None:
@@ -251,6 +260,10 @@ def init_quantized_weights(
             model,
             checkpoint=pretrained_model_name_or_path,
             device_map=_device_map,
+            # Passed explicitly because both aliases were popped above;
+            # load_checkpoint_and_dispatch() accepts `dtype` but not
+            # `torch_dtype`, so forwarding kwargs verbatim was unsafe.
+            dtype=torch_dtype,
             *args,
             **kwargs,
         )
--- a/modelopt/torch/quantization/plugins/accelerate.py
+++ b/modelopt/torch/quantization/plugins/accelerate.py
@@ -222,12 +222,26 @@ def init_quantized_weights(
                 pretrained_model_name_or_path, trust_remote_code=trust_remote_code
             )
 
+        # Model-construction kwargs must not reach load_checkpoint_and_dispatch(),
+        # which does not accept them (e.g. attn_implementation -> TypeError).
+        attn_implementation = kwargs.pop("attn_implementation", None)
+
         with init_empty_weights():
             # Fix torch_dtype to match original model
             torch_dtype = kwargs.get(
                 "dtype", kwargs.get("torch_dtype", getattr(config, "torch_dtype", torch.float16))
             )
-            model = cls.from_config(config, dtype=torch_dtype)
+            from_config_kwargs = {}
+            if attn_implementation is not None:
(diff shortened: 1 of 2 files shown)


spped2000:
Good catch, fixed and pushed.

Confirmed the concern is real: `inspect.signature(accelerate.load_checkpoint_and_dispatch)` accepts `dtype` but **not** `torch_dtype`, so a caller using the legacy alias hit the same `TypeError` class this PR fixes for `attn_implementation` — `kwargs.get()` left it in place.

Both aliases are now popped, precedence (`dtype` > `torch_dtype` > `config.torch_dtype` > fp16) is unchanged, and I pass the resolved value explicitly as `dtype=` to `load_checkpoint_and_dispatch()` rather than dropping it — otherwise callers who previously passed `dtype` would silently lose the weight cast on load.

spped2000:
Third finding addressed (the out-of-diff one) — and it reproduces, so thank you for catching it.

Verified on transformers 5.13.1:

```python
>>> PretrainedConfig().torch_dtype
None
>>> getattr(PretrainedConfig(), "torch_dtype", torch.float16)
None          # not float16 — the getattr default never fires
```

Because `torch_dtype` is a deprecated *alias* that returns `None` rather than being absent, the existing `getattr(config, "torch_dtype", torch.float16)` yields `None` whenever the config carries no dtype. The model is then constructed as float32 and `dtype=None` reaches `load_checkpoint_and_dispatch()` — on the one code path whose purpose is loading under tight memory. Pre-existing, but squarely in the lines this PR touches.

Now reads `config.dtype` first, then the alias, then falls back to float16 explicitly. Added `test_init_quantized_weights_dtype_resolution` covering kwarg precedence, kwarg consumption (so neither alias leaks into dispatch), and the no-dtype-in-config fallback; assertions verified against transformers 5.13.1.

Note the PR now carries three distinct fixes. Happy to split the dtype one out if you would rather review them separately.

spped2000:
Both review points addressed — the first one was a fair hit and I have rewritten the test.

**The test was testing a copy of the logic, not the code.** As written it mirrored the resolution order in a local `resolve()` helper, so it would have passed even with `patched_from_pretrained` completely broken. It now drives the real `init_quantized_weights` context manager with the checkpoint I/O stubbed (`mtq.quantize`/`compress`, the accelerate device-map helpers, and `load_checkpoint_and_dispatch`) and asserts what each callee actually receives:

- `from_config` gets the resolved dtype **and** `attn_implementation`;
- dispatch receives neither `torch_dtype` nor `attn_implementation` (which it cannot accept) but does receive the resolved `dtype`;
- `tie_weights()` runs before quantization;
- with no dtype anywhere the float16 fallback fires — including the case where the deprecated `config.torch_dtype` alias returns `None`.

I checked it actually protects the code: reverting `accelerate.py` to the pre-PR implementation makes it fail on the `attn_implementation` assertion, and it passes with the fix. Full file: 3 passed.

`transformers` import moved to module scope behind `pytest.importorskip`, as suggested.

(One note on the suggested precedence `dtype > torch_dtype > config.dtype > config.torch_dtype > float16` — that is exactly what is implemented; `config.dtype` is read first and the alias only serves older transformers where `dtype` does not exist.)

spped2000:
Done — `accel_plugin` now imports at module scope, right after the `accelerate` skip guard and `pytest.importorskip("transformers")`, with an `# noqa: E402` since it necessarily follows those guards. Full test file still 3 passed.

That is all five review points from this round addressed. Summary of where the PR stands: three independent fixes on the `init_quantized_weights` path (kwargs routing, `tie_weights()`, dtype fallback) plus one regression test that drives the real loader. Root-cause defect on the same path — quantization running on meta tensors before weights load — remains #2160 and is deliberately out of scope here.

PR #2053, opened 2026-08-03
Add Cosmos3 Nano DFlash multimodal training recipe

skierat:
### What does this PR do?

  Type of change: new example

  Adds an end-to-end Cosmos3 Nano DFlash training recipe for multimodal speculative decoding.

  - Adds a notebook that prepares data, launches synthetic generation in Slurm, trains a DFlash draft model, exports it, and provides a vLLM smoke-test command.
  - Adds PAI-Understanding, VQA v2, and multilingual prompt sharding and distributed-generation helpers.
  - Adds an atomic, multimodal-safe merge and conservative deduplication flow.
  - Extends the VLM data collator to handle structured image/video messages, configurable visual bounds, and fixed DFlash sequence lengths.
  - Hardens generation launch scripts and preserves truncated generated responses.

  ### Usage

  ```bash
  cd examples/speculative_decoding/recipes

  export MODEL_PATH=/path/to/cosmos3-nano
  export PLAIN_TEXT_INPUT=/path/to/nemotron-chat-or-approved-user-data.jsonl

  jupyter lab train_dflash_cosmos3_nano.ipynb

  Run the notebook in order:

  1. Configure paths.
  2. Prepare prompts on a CPU-only node and generate target completions in a Slurm GPU allocation.
  3. Merge the four required sources and submit training.
  4. Export a saved checkpoint and run the vLLM deployment smoke test.

  ### Testing

  - jq empty examples/speculative_decoding/recipes/train_dflash_cosmos3_nano.ipynb
  - bash -n on the modified launch, worker, and recipe shell scripts.
  - Ran a two-step Cosmos3 Nano DFlash Slurm smoke job; it completed and wrote modelopt_state.pth.
  - Not run: pytest tests/unit/torch/speculative/plugins/test_hf_speculative_offline.py (pytest is unavailable in the current environment).

  ### Before your PR is "Ready for review"

  - Is this change backward compatible?: ✅
  - If you copied code from any other sources or added a new PIP dependency, did you follow guidance in CONTRIBUTING.md?: N/A
  - Did you write any new necessary tests?: ✅
  - Did you update CHANGELOG.rst?: N/A
  - Did you get Claude approval on this PR?: N/A

  ### Additional Information

  Security follow-up required before marking ready: the notebook hardcodes
  model.trust_remote_code=true and --trust_remote_code. Either parameterize
  this with a default of false, or obtain and document a security exception.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
[...]

PR #2173, opened 2026-08-12
Express the Cosmos3 Nano DFlash recipe as a launcher YAML example

h-guo18:
Targets `skierat/dflash_recipe` (#2053), not `main`. Merging this into that
branch before #2053 lands keeps the recipe as a single reviewable unit.

### What this changes

Ports the `train_dflash_cosmos3_nano.ipynb` workflow to `tools/launcher` as a
7-task pipeline, and replaces the notebook with a README that keeps its prose.

**None of the Python, shell, or tests from #2053 are touched.** The diff is
additive except for deleting the notebook:

```
recipes/README_dflash_cosmos3_nano.md                    +122
recipes/train_dflash_cosmos3_nano.ipynb                  -732
common/specdec/merge_dflash_datasets.sh                   +52
common/specdec/multimodal_prepare_shards.sh              +154
common/specdec/multimodal_synthetic_generation.sh        +128
examples/nvidia/Cosmos3-Nano/hf_online_dflash_multimodal.yaml +205
```

### Why

Every other speculative-decoding recipe in the repo is already a launcher YAML
(Qwen3-8B, Qwen3-30B-A3B, Kimi-K2.6, MiniMax-M3, gpt-oss-20b). This makes the
Cosmos3 Nano recipe consistent with them, and picks up three things the
notebook could not:

- **Training and export are already implemented.**
  `common/specdec/dflash_online_training.sh` runs the accelerate launch *and*
  exports every checkpoint, so the notebook's separate export step goes away.
  The vLLM smoke test reuses `common/specdec/vllm_smoke_test.sh`, which also
  asserts a minimum acceptance length rather than only printing a command.
- **CI coverage.** `tests/test_examples_resolve.py` validates the YAML
  structurally (60 passed locally, including this example). The notebook could
  only be checked with `jq empty`.
- **No committed cluster specifics.** The notebook hardcoded a personal
  `$HF_HOME`, a personal `.sqsh` image, and a specific Slurm account. Those
  become per-cluster overrides. The YAML's defaults are deliberately generic.

### Behavior parity

Every flag and default is carried over verbatim — the `training.*` / `dflash.*`
overrides, the six `VLM_*` caps, the dedup parameters, the per-source sample
counts and shuffle seeds.

The one real rewrite is shard-to-node assignment. The notebook cells computed a
node list by hand and passed `job_id` + comma-separated node names into
`launch_multimodal.sh`. The launcher owns the allocation, so
`multimodal_synthetic_generation.sh` derives both from `SLURM_JOB_ID` /
`SLURM_JOB_NODELIST` instead. I verified the rounding covers every shard with
no gaps and that the out-of-range guards fire.

### Prose

The notebook's explanatory content is preserved in
`recipes/README_dflash_cosmos3_nano.md`: why DFlash trains on the target's own
completions rather than human-written answers, what each data source
contributes, what the merge guarantees (absolute media paths → `vlm_img_dir=/`;
dedup removes redundancy, not temperature diversity), and the training
constraints — `training_seq_len % dflash_block_size == 0`, why the `VLM_*` caps
exist, and that setting `data.vlm_processor` is what selects the multimodal
collator. Per-step detail lives in YAML comments next to the settings.

### Not addressed here

Deliberately scoped to the example layout. The review comments on #2053 itself
(the removed `truncation=True` turning an over-length sample into a job-fatal
`ValueError`, the shape-based tensor padding, and `hash()` making parallel merge
non-reproducible across processes) are unchanged and still worth a look.


skierat:
Codex Findings:

  - [P1] Generate all sources before merging them — YAML lines 92–120
    (https://github.com/NVIDIA/Model-Optimizer/blob/ed68258e37240c3251bb4c79579f8869a17985de/tools/launcher/examples/nvidia/Cosmos3-Nano/hf_online_dflash_multimodal.yaml#L92-L120). task_3 generates only VQA
    outputs, but task_4 unconditionally merges PAI, VQA, and text output directories. The default pipeline therefore fails when the merge cannot find pai_outputs and text_outputs. Add generation tasks for PAI
    and text, or remove those sources from the default merge.

  - [P2] Restore padding before requesting PyTorch tensors — collator lines 599–610
    (https://github.com/NVIDIA/Model-Optimizer/blob/ed68258e37240c3251bb4c79579f8869a17985de/modelopt/torch/utils/plugins/transformers_dataset.py#L599-L610). The processor is called with return_tensors="pt" but
    no padding; _pad_sequence_tensors() runs only afterward. A batch containing different prompt/video token lengths therefore cannot be tensorized, so batch sizes above one can fail before manual padding runs.
    Pass padding=True/"longest" to apply_chat_template, then retain the fixed-length padding step. Hugging Face forwards the supplied processor kwargs to the tensor-producing call. Source
    (https://raw.githubusercontent.com/huggingface/transformers/v5.12.0/src/transformers/processing_utils.py)

PR #2174, opened 2026-08-12
Export quantized/co-trained MTP weights instead of copying BF16

yeyu-nvidia:
## What

`GPTModelExporter._get_mtp_state_dict` copied the MTP (multi-token prediction) head verbatim from the BF16 **pretrained** model instead of exporting the live model's MTP weights. `_get_state_dict` only walks `model.decoder.layers` and never `model.mtp`, so `self._state_dict` never contains any `mtp.*` keys and the `key not in self._state_dict` guard was always true — every MTP tensor came from the pretrained safetensors. There was a standing `# TODO Implement MTP export for quantized MTP`.

Consequence: any quantization or co-training applied to the MTP head during QAD was **silently discarded at export**; the exported draft head was always the original BF16 weights. This makes it impossible to evaluate MTP quantization or MTP co-training downstream.

## Fix

- Rewrite `_get_mtp_state_dict` to walk the live MCore `model.mtp` module and apply the same quantization rules used for the base decoder (mirroring `_get_eagle_module_state_dict`). The MTP inner attention/MoE layers are structurally identical to backbone hybrid layers, so the base layer walker (`_get_transformer_layer_state_dict` / `_get_mamba_layer_state_dict`) is reused with a **restricted** set of `mtp.*` naming rules aliased onto the standard rule keys — emitting `mtp.layers.{}.` HF keys. Restricted on purpose: any base rule key the walker references but that has no `mtp.` variant is simply absent (guarded), rather than silently emitting a wrong `backbone.` prefix.
- Keep the old copy behavior as `_copy_mtp_state_dict_from_pretrained`, used only when the live model has no `mtp` module (e.g. exporting a base-only checkpoint that grafts a pretrained head).
- Add the missing MTP inner-layer export rules to `nemotron_h_causal_lm_export` (attention `qkv`/`o_proj`/`norm`, MoE `router`/`experts`/`shared_experts`). The predictor projection rules (`mtp.enorm/hnorm/eh_proj/final_layernorm`) already existed.

## Validation

- For a `NemotronHForCausalLM` model (`num_nextn_predict_layers=1`, hybrid `*E`), the walker reproduces exactly the expected HF key layout — `mtp.layers.0` = attention (`enorm/hnorm/eh_proj/norm/mixer.{q,k,v,o}_proj`), `mtp.layers.1` = MoE (`norm/final_layernorm/mixer.gate/shared_experts/experts.{e}`) — matching what the existing `nemotron_h_causal_lm_import` (`is_mtp` keys) reads back.
- pre-commit clean (ruff / ruff-format / mypy / bandit).
- End-to-end re-export + downstream MTP spec-decode eval on a co-trained checkpoint is in progress; will post the before/after tensor diff here.

## Note for reviewer

@jenchen13 — this is the MTP-export bug you flagged (the L599 BF16 copy). Would appreciate your review, especially on the assumption that the MTP inner layers can be driven through the base layer walker with only a prefix swap, and whether any non-`*E` MTP configurations need additional inner-layer rules.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected repeated MTP imports to load the final layer normalization correctly.

* **Documentation**
  * Clarified MTP export mapping behavior and prefix-remapping guidance.
  * Improved documentation for MTP exports across supported architectures.

### DIFF
--- a/modelopt/torch/export/plugins/mcore_nemotron.py
+++ b/modelopt/torch/export/plugins/mcore_nemotron.py
@@ -155,26 +155,13 @@ nemotron_h_causal_lm_export: dict[str, CustomModuleMapping] = {
     # Grouped local experts (TEGroupedMLP: fused per-expert weights)
     "experts.linear_fc1": GroupedMLPSlicing("backbone.layers.{}.mixer.experts.{{}}.up_proj"),
     "experts.linear_fc2": GroupedMLPSlicing("backbone.layers.{}.mixer.experts.{{}}.down_proj"),
-    # MTP predictor projections (outer MultiTokenPredictionLayer)
+    # MTP predictor projections (outer MultiTokenPredictionLayer). The MTP inner
+    # attention/MoE layers are structurally identical to the backbone hybrid layers, so
+    # they reuse the base rules above via is_mtp=True (which retargets the backbone/model
+    # root to mtp, mirroring the importer) — only these predictor-specific keys are
+    # dedicated.
     "mtp.enorm": NameRemapping("mtp.layers.{}.enorm."),
     "mtp.hnorm": NameRemapping("mtp.layers.{}.hnorm."),
     "mtp.eh_proj": NameRemapping("mtp.layers.{}.eh_proj."),
     "mtp.final_layernorm": NameRemapping("mtp.layers.{}.final_layernorm."),
-    # MTP inner attention / MoE layers. Structurally identical to the backbone hybrid
-    # layers, so these mirror the base rules above with the `mtp.layers.{}` prefix; the
-    # `mtp.` namespace is aliased onto the standard rule keys by _get_mtp_state_dict.
-    "mtp.input_layernorm": NameRemapping("mtp.layers.{}.norm."),
-    "mtp.fused_norm": NameRemapping("mtp.layers.{}.norm.weight"),
-    "mtp.linear_qkv": QKVSlicing("mtp.layers.{}.mixer."),
-    "mtp.linear_proj": NameRemapping("mtp.layers.{}.mixer.o_proj."),
-    "mtp.pre_mlp_layernorm": NameRemapping("mtp.layers.{}.norm."),
-    "mtp.router": NameRemapping(
-        "mtp.layers.{}.mixer.gate.", {"mapping": {"expert_bias": "e_score_correction_bias"}}
-    ),
-    "mtp.shared_experts.linear_fc1": NameRemapping("mtp.layers.{}.mixer.shared_experts.up_proj."),
-    "mtp.shared_experts.linear_fc2": NameRemapping("mtp.layers.{}.mixer.shared_experts.down_proj."),
-    "mtp.local_experts.linear_fc1": NameRemapping("mtp.layers.{}.mixer.experts.{}.up_proj."),
-    "mtp.local_experts.linear_fc2": NameRemapping("mtp.layers.{}.mixer.experts.{}.down_proj."),
-    "mtp.experts.linear_fc1": GroupedMLPSlicing("mtp.layers.{}.mixer.experts.{{}}.up_proj"),
-    "mtp.experts.linear_fc2": GroupedMLPSlicing("mtp.layers.{}.mixer.experts.{{}}.down_proj"),
 }
--- a/modelopt/torch/export/plugins/mcore_nemotron.py
+++ b/modelopt/torch/export/plugins/mcore_nemotron.py
@@ -155,9 +155,26 @@ nemotron_h_causal_lm_export: dict[str, CustomModuleMapping] = {
     # Grouped local experts (TEGroupedMLP: fused per-expert weights)
     "experts.linear_fc1": GroupedMLPSlicing("backbone.layers.{}.mixer.experts.{{}}.up_proj"),
     "experts.linear_fc2": GroupedMLPSlicing("backbone.layers.{}.mixer.experts.{{}}.down_proj"),
-    # MTP
+    # MTP predictor projections (outer MultiTokenPredictionLayer)
     "mtp.enorm": NameRemapping("mtp.layers.{}.enorm."),
     "mtp.hnorm": NameRemapping("mtp.layers.{}.hnorm."),
     "mtp.eh_proj": NameRemapping("mtp.layers.{}.eh_proj."),
     "mtp.final_layernorm": NameRemapping("mtp.layers.{}.final_layernorm."),
+    # MTP inner attention / MoE layers. Structurally identical to the backbone hybrid
+    # layers, so these mirror the base rules above with the `mtp.layers.{}` prefix; the
+    # `mtp.` namespace is aliased onto the standard rule keys by _get_mtp_state_dict.
+    "mtp.input_layernorm": NameRemapping("mtp.layers.{}.norm."),
+    "mtp.fused_norm": NameRemapping("mtp.layers.{}.norm.weight"),
+    "mtp.linear_qkv": QKVSlicing("mtp.layers.{}.mixer."),
+    "mtp.linear_proj": NameRemapping("mtp.layers.{}.mixer.o_proj."),
+    "mtp.pre_mlp_layernorm": NameRemapping("mtp.layers.{}.norm."),
+    "mtp.router": NameRemapping(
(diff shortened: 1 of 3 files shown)


modelopt/torch/export/plugins/mcore_nemotron.py
```diff
@@ -155,9 +155,26 @@
...
+    "mtp.pre_mlp_layernorm": NameRemapping("mtp.layers.{}.norm."),
+    "mtp.router": NameRemapping(
+        "mtp.layers.{}.mixer.gate.", {"mapping": {"expert_bias": "e_score_correction_bias"}}
+    ),
+    "mtp.shared_experts.linear_fc1": NameRemapping("mtp.layers.{}.mixer.shared_experts.up_proj."),
+    "mtp.shared_experts.linear_fc2": NameRemapping("mtp.layers.{}.mixer.shared_experts.down_proj."),
+    "mtp.local_experts.linear_fc1": NameRemapping("mtp.layers.{}.mixer.experts.{}.up_proj."),
+    "mtp.local_experts.linear_fc2": NameRemapping("mtp.layers.{}.mixer.experts.{}.down_proj."),
+    "mtp.experts.linear_fc1": GroupedMLPSlicing("mtp.layers.{}.mixer.experts.{{}}.up_proj"),
+    "mtp.experts.linear_fc2": GroupedMLPSlicing("mtp.layers.{}.mixer.experts.{{}}.down_proj"),
```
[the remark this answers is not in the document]
jenchen13: if the MTP layers are structurally identical to the base model layers, then maybe we don't need to add `mtp.*` rules that are duplicates of the base model rules? It should be possible to just use the base model rules. Adding these mtp rules also makes the import & export rules diverge.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -597,12 +597,84 @@ def _get_transformer_layer_state_dict(self, layer, layer_id):
...
+            return self._copy_mtp_state_dict_from_pretrained()
+
+        # The MTP inner attention / MoE layers are structurally identical to the base
+        # decoder layers, so we reuse the base layer walker. We alias the ``mtp.*``
+        # naming rules onto the standard rule keys for the duration of the walk so the
+        # walker emits ``mtp.layers.{}.`` HF keys instead of ``backbone.layers.{}.``.
+        # A *restricted* alias set is used on purpose: any base rule key the walker
+        # references but that has no ``mtp.`` variant is simply absent (and its call is
+        # guarded), rather than silently emitting a wrong ``backbone.`` prefix.
+        mtp_rules = {
```
jenchen13: you can just reuse the base layer rules + add the mtp specific ones for enorm, hnorm, eh_proj, final_layernorm

yeyu-nvidia:
### End-to-end validation (NemotronH, `num_nextn_predict_layers=1`, hybrid `*E`)

Re-exported a **co-trained** checkpoint with this branch and compared the `mtp.*` tensors against the BF16 pretrained model that the old code copied from:

| tensor | fixed export vs BF16 pretrained source |
|---|---|
| `mtp.layers.0.eh_proj.weight` | **differs** |
| `mtp.layers.0.enorm / hnorm / norm` | **differs** |
| `mtp.layers.0.mixer.{q,k,v,o}_proj.weight` | **differs** |

Before this PR every `mtp.*` tensor was byte-identical to the pretrained model regardless of training (the copy path), so any QAD co-training of the MTP head was discarded at export. After the fix the exported head reflects the trained weights. Full layout (270 `mtp.*` tensors, `layers.0`=attention, `layers.1`=MoE) matches the `nemotron_h_causal_lm_import` round-trip.

**Note on precision:** in this particular checkpoint the MTP head was left unquantized (BF16, like the lm_head), so the exported MTP is BF16 with no `weight_scale`. The walker runs the same quantization rules as the base decoder, so it will emit NVFP4 weights + scales whenever the MTP module is quantized in the checkpoint — this validation just didn't exercise that path. @jenchen13 flagging in case the MTP head is expected to be quantized by the recipe.

modelopt/torch/export/plugins/mcore_nemotron.py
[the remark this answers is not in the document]
yeyu-nvidia: Good call — done in 89ef5962bd. I removed the duplicate `mtp.*` inner-layer rules and instead thread an `is_mtp` flag through the base layer walker + remapping helpers, which rewrite the target root `backbone`/`model` → `mtp` — mirroring exactly what the importer does. Only enorm/hnorm/eh_proj/final_layernorm stay as dedicated `mtp.*` rules now, so the import/export rule books stay symmetric.

modelopt/torch/export/unified_export_megatron.py
yeyu-nvidia: Good call — done in 89ef5962bd. I removed the duplicate `mtp.*` inner-layer rules and instead thread an `is_mtp` flag through the base layer walker + remapping helpers, which rewrite the target root `backbone`/`model` → `mtp` — mirroring exactly what the importer does. Only enorm/hnorm/eh_proj/final_layernorm stay as dedicated `mtp.*` rules now, so the import/export rule books stay symmetric.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -498,111 +498,201 @@ def _get_fused_norm_weight(self, module, primary_key: str = "fused_norm"):
...
+        Walks the live MCore ``model.mtp`` module and applies the same quantization
+        rules used for the base decoder, so the exported draft head reflects the
+        actual (quantized / co-trained) weights. Falls back to copying the BF16 MTP
+        weights from the pretrained model only when the live model has no ``mtp``
+        module (e.g. exporting a base-only checkpoint that grafts a pretrained head).
+        """
+        model = getattr(self, "model", None)
+        mtp = getattr(model, "mtp", None)
+        if mtp is None or not hasattr(mtp, "layers") or len(mtp.layers) == 0:
+            return self._copy_mtp_state_dict_from_pretrained()
```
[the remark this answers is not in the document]
jenchen13: MTP is usually always on the last pipeline stage though

modelopt/torch/export/plugins/megatron_importer.py
```diff
@@ -860,6 +860,11 @@ def _import_state_dict(self):
                         )
 
                     layer_id += 1
+
+                # Import the MTP block's own final_layernorm into the last inner layer
```
jenchen13: can you remove this comment?

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -498,111 +498,201 @@ def _get_fused_norm_weight(self, module, primary_key: str = "fused_norm"):
...
+
+        # The MTP inner attention / MoE layers are structurally identical to the base
+        # decoder layers, so the same layer walker + rules are reused with is_mtp=True,
```
jenchen13: AI comments are too long

jenchen13 (approved):

modelopt/torch/export/unified_export_megatron.py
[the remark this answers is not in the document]
yeyu-nvidia: Right — this runs only on the last-stage main rank (guarded by the `is_last_stage_main_rank` check in `save_pretrained`), and MTP lives on that stage, so `model.mtp` is present and the live-walker path is what actually executes for MTP exports. The `_copy_mtp_state_dict_from_pretrained` fallback only kicks in when the MCore model has no MTP at all — i.e. exporting a base-only checkpoint that grafts a pretrained MTP head (the prior behavior). The `getattr(self, "model", ...)` is just so the fallback's unit tests (which build the exporter with no live model via `object.__new__`) still pass. Happy to drop the fallback entirely if we don't want to support the base-only-graft case — let me know.

modelopt/torch/export/plugins/megatron_importer.py
yeyu-nvidia: Done in c8bb042ad9 — trimmed the verbose comments/docstrings across the export + import changes.

modelopt/torch/export/unified_export_megatron.py
yeyu-nvidia: Done in c8bb042ad9 — trimmed the verbose comments/docstrings across the export + import changes.

PR #2180, opened 2026-08-12
[OMNIML-5563] Add PETR ONNX PTQ and accuracy evaluation example

ajrasane:
### What does this PR do?

Type of change: new example, example simplification, and backward-breaking example migration

Adds end-to-end PETRv1/PETRv2 ONNX PTQ and reduces PETR/FAR3D to one shared workflow:

- quantizes the shared VoVNet image backbone/encoder to INT8 or FP8;
- runs both the selected historical and current PETRv2 six-camera sweeps through the same precision-matched TensorRT backbone engine using distinct execution contexts during accuracy evaluation;
- keeps the PETR head and FAR3D decoder in their exported mixed FP16/FP32 precision;
- reuses one NPZ calibration format, VoVNet exclusion helper, quantization entry point, and TensorRT runner;
- does not change generic Model Optimizer calibration behavior or its public CLI.

### Container boundary

Both examples use two targets from one Dockerfile, with no virtual environments:

- `evaluator`: a digest-pinned `nvcr.io/nvidia/pytorch:22.06-py3` base with the legacy PyTorch 1.13.1/OpenMMLab stack for source setup, metadata generation, ONNX export, direct PyTorch calibration capture, and final accuracy evaluation;
- `modelopt`: a digest-pinned `nvcr.io/nvidia/pytorch:26.07-py3` base for Model Optimizer, ONNX Runtime CUDA, AutoCast, INT8/FP8 quantization, and TensorRT engine builds.

Both targets use TensorRT `11.1.0.106`. Engines are built and evaluated on the same GPU architecture. Final metrics remain in the evaluator because they import the legacy model-framework postprocessing and dataset code; only artifacts cross the container boundary through the shared workspace.

PETR is used without patches. FAR3D applies only the official `patch/far3d.patch` from the pinned NVIDIA DL4AGX revision. This PR carries no patch files.

### Evaluator dependencies

The dependencies intentionally installed without transitive dependencies are listed in `requirements-evaluator-nodeps.txt`. Their pins rely on runtime packages supplied by the digest-pinned PyTorch 22.06 evaluator base.

`lyft-dataset-sdk` is required only by mmdet3d's eager dataset import; neither PETR nor FAR3D uses Lyft data. `flash-attn` remains in the main evaluator requirements because its compiled installation uses the evaluator build step rather than the intentionally dependency-free legacy package step.

Fresh setup and dependency approval is requested for the final reduced dependency set.

### Reproducible PETR metadata

The documented workflow mounts raw nuScenes read-only and creates a writable dataset view using symlinks. It then runs the pinned mmdetection3d converter and a temporary, untracked copy of PETR's pinned sweep generator configured only for the validation prefix and writable dataset root.

A clean run generated both metadata files with 6,019 validation records. The referenced camera, lidar, and sweep paths are absolute and resolvable through the writable dataset view.

### Example-local utilities

The per-batch NPZ streaming and TensorRT runtime utilities remain example-local because they execute in the legacy evaluator, where Model Optimizer is not installed. The core `CalibrationDataProvider` consumes one in-memory mapping of stacked arrays and does not provide this streamed per-file workflow.

### Validation

- Focused CPU tests: 10 passed.
- Broader ONNX quantization CPU tests: 326 passed.
- All applicable pre-commit and documentation checks, plus `git diff --check`, passed.
- Rebuilt both Docker targets and verified their exact dependency versions, imports, TensorRT `11.1.0.106`, GPU runtime initialization, and absence of virtual environments.
- Generated both PETR metadata files from a clean writable dataset view and verified 6,019 validation records plus resolvable data paths.
- PETRv1 passed a one-sample TensorRT regression smoke.
- PETRv2 passed FP16, INT8, and FP8 TensorRT smokes and full 6,019-sample validation. Both the selected historical and current sweeps are computed by the matching backbone engine; accuracy evaluation no longer extracts image features with PyTorch.
- FAR3D passed a recurrent two-frame TensorRT smoke covering plugin loading and recurrent state.

TensorRT `11.1.0.106` mAP follows. PETRv2 was remeasured after correcting its temporal feature path; the PETRv1 and FAR3D numerical paths are unchanged.

| Pipeline | FP16 | INT8 | FP8 |
| --- | ---: | ---: | ---: |
| PETRv1: 1 backbone pass + fixed typed mixed FP16/FP32 head | 0.3778 | 0.3707 | 0.3756 |
| PETRv2: 2 serial backbone passes + fixed typed mixed FP16/FP32 head | 0.4102 | 0.3982 | 0.4084 |
| FAR3D: 1 encoder pass + fixed mixed FP16/FP32 decoder | 0.241 | 0.235 | 0.239 |

Normalized engine-only performance improvement over each matching FP16 pipeline:

| Pipeline | INT8 speedup | FP8 speedup |
| --- | ---: | ---: |
| PETRv1 | 1.49x | 1.29x |
| PETRv2 | 1.51x | 1.30x |
| FAR3D | 1.69x | 1.40x |

Performance was measured with TensorRT `11.1.0.106` on an NVIDIA RTX 6000 Ada Generation GPU using five interleaved trials per engine component. Each component uses the median `trtexec`-reported GPU Compute Time with data transfers disabled and CUDA Graphs enabled. Component times are summed before normalization: PETRv1 uses one backbone pass plus its fixed head, PETRv2 uses two serial backbone passes plus its fixed head with no temporal cache assumed, and FAR3D uses one encoder pass plus its fixed decoder. Absolute latency values are intentionally not published.

Adapted files retain exact public-source references and upstream notices, and the top-level license attribution is updated.

- Is this change backward compatible?: ❌
- Did you update the changelog?: ✅


cjluo-nv (commented):

New self-contained PETRv1/v2 ONNX PTQ example (~990 lines, 12 files) modeled closely on the existing `examples/onnx_ptq/far3d/` example. The workflow looks plausible and the author reports full nuScenes validation numbers for all six configurations, so I'm mostly commenting on reuse, conventions, and a couple of silent-failure paths rather than on the quantization results.

Main points:

1. **Licensing needs human sign-off (no auto-approve).** `evaluate.py` is "Adapted from NVIDIA/DL4AGX .../petr-trt/export_eval" with an OpenMMLab copyright, and `prepare_sweep_metadata.py` is adapted from PETR `tools/generate_sweep_pkl.py` with a Megvii copyright. Third-party code import plus new third-party pip requirements (mmdet3d/mmcv-full/nuscenes-devkit/lyft-dataset-sdk/pycuda…) should get a maintainer/legal ack. Separately, `petr/Dockerfile` carries only the first two SPDX lines instead of the full canonical `LICENSE_HEADER` text that `far3d/Dockerfile` and every other new file in this PR use — please align it.

2. **`convert_to_fp16.py` duplicates a shipped CLI.** `python -m modelopt.onnx.autocast --onnx_path ... --output_path ... --calibration_data ... --keep_io_types --providers cuda:0 cpu` already does exactly what this 43-line script does (see `modelopt/onnx/autocast/__main__.py`). Recommend deleting the file and documenting the CLI invocation in the README instead.

3. **Substantial copy-paste from the FAR3D example.** `TRT_TO_TORCH` / `aligned_tensor` / `TensorRTRunner` / `import_plugin` / the `eval_kwargs` scrub in `evaluate.py`, `FileCalibrationReader` in `quantize.py`, and `find_backbone_nodes_to_exclude` (byte-identical VoVNet `OSA4_5` + `lateral_convs` heuristic) all already exist in `examples/onnx_ptq/far3d/`. Both examples run in the same Python 3.8 legacy venv layout, so a small shared helper module under `examples/onnx_ptq/` would avoid two copies drifting apart.

4. **Convention regression in `prepare_sweep_metadata.py`**: all work runs at module import time with no `main()`/`__main__` guard (the sibling `far3d/prepare_metadata.py` uses one), and it silently overwrites an existing output pkl where FAR3D explicitly refuses to.

5. Silent-failure risks: calibration input names are taken from ONNX `graph.input` order but zipped against engine-binding order; and the README tells the reader to run the container-only `/opt/petr/bin/python` before the container is built. Details inline.

Also worth noting: this README states TensorRT 11 no longer accepts `trtexec --fp16`, but `far3d/README.md` (same 26.07 base image) still instructs `--fp16`. If that's true, the sibling example's instructions are broken and should be fixed in a follow-up.

No tests, which is consistent with other examples here — fine.

examples/onnx_ptq/petr/convert_to_fp16.py
```diff
@@ -0,0 +1,43 @@
...
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Convert a PETR ONNX model to FP16")
+    parser.add_argument("input")
+    parser.add_argument("output")
+    parser.add_argument("calibration_data")
+    return parser.parse_args()
+
+
+def main():
```
cjluo-nv: > _Bot comment._

This script is a thin re-implementation of the AutoCast CLI that ModelOpt already ships (`modelopt/onnx/autocast/__main__.py`), which exposes `--onnx_path`, `--output_path`, `--calibration_data`, `--keep_io_types`, `--providers` and also calls `onnx_utils.save_onnx` for you. Suggest dropping this file and putting the equivalent `python -m modelopt.onnx.autocast --onnx_path ... --output_path ... --calibration_data ... --keep_io_types --providers cuda:0 cpu` command in the README so there's one code path to maintain.

examples/onnx_ptq/petr/quantize.py
```diff
@@ -0,0 +1,131 @@
...
+            missing = self.input_dtypes.keys() - batch.files
+            if missing:
+                raise ValueError(f"{batch_path} is missing inputs: {sorted(missing)}")
+            return {
+                name: batch[name].astype(dtype, copy=False)
+                for name, dtype in self.input_dtypes.items()
+            }
+
+
+def find_backbone_nodes_to_exclude(onnx_path):
```
cjluo-nv: > _Bot comment._

`find_backbone_nodes_to_exclude` is the same function as `find_encoder_nodes_to_exclude` in `examples/onnx_ptq/far3d/quantize.py` (identical `OSA4_5` / `lateral_convs` VoVNet heuristic); the only difference is that the regex anchoring was moved inside instead of at the call site. Since both examples target the same VoVNet backbone, please factor this into one shared helper rather than keeping two copies — otherwise a future fix to the exclusion heuristic will only land in one example.

examples/onnx_ptq/petr/quantize.py
```diff
@@ -0,0 +1,131 @@
...
+
+import numpy as np
+import onnx
+from onnxruntime.quantization.calibrate import CalibrationDataReader
+
+from modelopt.onnx.quantization import quantize
+from modelopt.onnx.utils import topologically_sort_graph_nodes
+
+
+class FileCalibrationReader(CalibrationDataReader):
```
cjluo-nv: > _Bot comment._

`FileCalibrationReader` here is essentially `FileCalibrationReader` + `DecoderCalibrationReader` from `far3d/quantize.py` merged. Same suggestion as for the node-exclusion helper: extract once under `examples/onnx_ptq/` and import from both examples.

examples/onnx_ptq/petr/evaluate.py
```diff
@@ -0,0 +1,250 @@
...
+
+def aligned_tensor(shape, dtype, device, alignment=256):
+    element_size = torch.empty((), dtype=dtype).element_size()
+    element_count = int(torch.tensor(shape).prod().item())
+    storage = torch.empty(element_count + alignment // element_size, dtype=dtype, device=device)
+    offset = ((-storage.data_ptr()) % alignment) // element_size
+    return storage[offset : offset + element_count].view(shape)
+
+
+class TensorRTRunner:
```
cjluo-nv: > _Bot comment._

`TRT_TO_TORCH`, `aligned_tensor`, `TensorRTRunner`, `import_plugin` and the `eval_kwargs` key-scrub loop in `main()` are near-verbatim copies of `far3d/evaluate.py`. Given both examples run in the same Python 3.8 legacy venv, a shared `examples/onnx_ptq/trt_runner.py` (or similar) would remove ~100 duplicated lines.

One behavioral difference worth calling out: this version binds engine inputs *positionally* (`zip(self.input_names, values)`) whereas the FAR3D runner binds by tensor name. Positional binding silently produces wrong results if the engine's IO-tensor order ever differs from the caller's list order, and `fit_shape` will happily squeeze/unsqueeze a mismatched tensor rather than fail. Consider keeping name-based binding, or at least validating each value's shape against `self.shapes[name]` without the squeeze/unsqueeze fallback.

examples/onnx_ptq/petr/prepare_calibration.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class CalibrationWriter:
+    def __init__(self, output_dir, onnx_path):
+        self.output_dir = Path(output_dir)
+        self.output_dir.mkdir(parents=True, exist_ok=True)
+        if any(self.output_dir.glob("*.npz")):
+            raise FileExistsError(f"{self.output_dir} already contains calibration batches")
+        graph = onnx.load(onnx_path, load_external_data=False).graph
+        self.names = [value.name for value in graph.input]
```
cjluo-nv: > _Bot comment._

`self.names` comes from the ONNX `graph.input` order, but `__call__` receives the buffers in the *engine* IO-tensor order produced by `TensorRTRunner`. TensorRT usually preserves the parser's input order, but if it doesn't, the only symptom is calibration `.npz` files whose keys are attached to the wrong tensors — and `quantize.py`'s reader looks up by name, so this would silently calibrate with swapped inputs (the current `len(values) != len(self.names)` check won't catch it).

Safer options: pass the engine's `input_names` (and shapes) into the writer and key the npz off those, or assert each value's shape matches the ONNX input's declared shape before saving.

examples/onnx_ptq/petr/prepare_sweep_metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Add PETRv2 camera sweeps to nuScenes metadata")
+    parser.add_argument("data_root")
+    parser.add_argument("--split", choices=("train", "val", "test"), default="val")
+    parser.add_argument("--output")
+    return parser.parse_args()
+
+
+args = parse_args()
```
cjluo-nv: > _Bot comment._

This module executes everything at import time — `args = parse_args()` plus the `NuScenes` load and the whole sweep loop live at module scope with no `main()` / `if __name__ == "__main__"` guard. The sibling `far3d/prepare_metadata.py` uses the standard `main()` pattern; please match it (it also makes the module importable/testable and keeps arg parsing out of import side effects).

Two related nits in the same block:
- `mmcv.dump(key_infos, info_path)` at the end will silently overwrite an existing pkl. `far3d/prepare_metadata.py` deliberately raises `FileExistsError` instead — worth doing the same here.
- `pickle.load(open(...))` leaks the file handle and reintroduces raw `pickle` even though `mmcv` is already imported for `dump`; `mmcv.load(path)` (or a `with open(...)` block) is cleaner.

examples/onnx_ptq/petr/README.md
```diff
@@ -0,0 +1,231 @@
...
+│   ├── sweeps/
+│   ├── v1.0-trainval/
+│   └── nuscenes_infos_val.pkl
+└── mmdetection3d/
+```
+
+PETRv2 also needs metadata for the previous camera sweeps:
+
+```bash
+/opt/petr/bin/python \
```
cjluo-nv: > _Bot comment._

Ordering problem: this step invokes `/opt/petr/bin/python` and `/opt/Model-Optimizer/examples/onnx_ptq/petr/prepare_sweep_metadata.py`, both of which only exist *inside* the container that's built in the next code block. Either move the sweep-metadata step after the `docker build` / `docker run` instructions, or note explicitly that it must be run from inside the container.

Also, the directory layout above lists `nuscenes_infos_val.pkl` alongside the downloaded nuScenes folders, but that file isn't part of the dataset release — the sweep script depends on it, so please document the mmdetection3d `create_data.py` (or PETR) step that generates it.

examples/onnx_ptq/petr/requirements.txt
kevalmorabia97, line 1: @claude summarize all new dependency licenses not included in pyproject.toml and report a table of dependency with their License name

examples/onnx_ptq/petr/convert_to_fp16.py
ajrasane: Removed the duplicate converter and switched the README to the shipped `python -m modelopt.onnx.autocast` CLI. A real PETRv1 head conversion and ONNX checker run passed.

examples/onnx_ptq/petr/quantize.py
ajrasane: Extracted `find_vovnet_nodes_to_exclude` into the shared ONNX PTQ utilities and updated both PETR and FAR3D to use it. It now raises when no sensitive nodes match; the actual PETRv1 and PETRv2 graphs matched 26 and 28 nodes, respectively.

examples/onnx_ptq/petr/quantize.py
ajrasane: Extracted the file/NPZ calibration readers into shared ONNX PTQ utilities and updated PETR and FAR3D to use them.

examples/onnx_ptq/petr/evaluate.py
ajrasane: Extracted the TensorRT execution core into a shared name-keyed runner. PETR now supplies and consumes tensors by exported names, validates names and shapes, and preserves the required one-leading-dimension adaptation. PETRv1, PETRv2, and FAR3D one-sample inference smokes passed.

examples/onnx_ptq/petr/prepare_calibration.py
ajrasane: Updated `CalibrationWriter` to validate the exact engine/ONNX input-name sets and write values by name instead of position. One-sample PETRv1/PETRv2 calibration produced NPZ keys, shapes, and dtypes that exactly match each ONNX graph.

examples/onnx_ptq/petr/prepare_sweep_metadata.py
ajrasane: Added a main guard, refusal to overwrite an existing output, and `mmcv.load` for the locally generated input metadata. The empty-sweep path is also guarded.

examples/onnx_ptq/petr/README.md
ajrasane: Moved the metadata steps into the container workflow, documented how to generate `nuscenes_infos_val.pkl`, pinned the source repositories, and aligned the dataset symlink and bind mount with the PETR configs.

examples/onnx_ptq/petr/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+FROM nvcr.io/nvidia/pytorch:26.07-py3
```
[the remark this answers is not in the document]
ajrasane: Documented and validated non-root execution with `docker run --user "$(id -u):$(id -g)" -e HOME=/tmp`. A fixed image user would not match arbitrary host bind-mount ownership; runtime UID/GID mapping preserves least privilege while keeping generated artifacts writable by the invoking user.

examples/onnx_ptq/petr/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        libgl1 \
+        libglib2.0-0 && \
+    rm -rf /var/lib/apt/lists/*
+
+ENV UV_PYTHON_INSTALL_DIR=/opt/python
+
+RUN python -m pip install --no-cache-dir uv && \
+    uv python install 3.8 && \
+    uv venv --seed --python 3.8 /opt/petr
```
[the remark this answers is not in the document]
ajrasane: Pinned uv, Python 3.8.20, every PETR requirements entry, and all three external source revisions. I did not add a cross-index transitive hash lock: this legacy build spans the NGC base image, apt, custom CUDA indexes, and source checkouts, so hashing only pip artifacts would not make the image immutable and is not the repository example convention. The rebuilt image passes `pip check` and the documented runtime smokes.

examples/onnx_ptq/petr/evaluate.py
```diff
@@ -0,0 +1,250 @@
...
+        stream.synchronize()
+        return outputs
+
+
+def import_plugin(cfg):
+    if cfg.get("custom_imports"):
+        import_modules_from_strings(**cfg.custom_imports)
+    if cfg.get("plugin"):
+        plugin_dir = os.path.dirname(cfg.plugin_dir).split("/")
+        importlib.import_module(".".join(plugin_dir))
```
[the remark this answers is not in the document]
ajrasane: Updated plugin loading to use `cfg.get("plugin_dir")`, so a truthy plugin flag without a directory no longer raises `AttributeError`.

examples/onnx_ptq/petr/prepare_calibration.py
```diff
@@ -0,0 +1,105 @@
...
+        ]
+        self.saved = 0
+
+    def __call__(self, values):
+        if len(values) != len(self.names):
+            raise ValueError(f"Received {len(values)} inputs, expected {len(self.names)}")
+        batch = {
+            name: value.detach().cpu().numpy().astype(dtype, copy=False)
+            for name, dtype, value in zip(self.names, self.dtypes, values)
+        }
```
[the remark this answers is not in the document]
ajrasane: Calibration capture is now name-keyed end to end. The writer rejects missing or unexpected engine inputs before dtype conversion, and real PETRv1/PETRv2 calibration batches were checked against ONNX names, shapes, and dtypes.

examples/onnx_ptq/petr/prepare_sweep_metadata.py
```diff
@@ -0,0 +1,160 @@
...
+info_prefix = args.split
+data_root = os.path.abspath(args.data_root) + os.sep
+num_prev = 5
+num_sweep = 5
+
+info_path = args.output or os.path.join(data_root, f"mmdet3d_nuscenes_30f_infos_{info_prefix}.pkl")
+key_infos = pickle.load(open(os.path.join(data_root, f"nuscenes_infos_{info_prefix}.pkl"), "rb"))
+nuscenes_version = "v1.0-test" if info_prefix == "test" else "v1.0-trainval"
+nuscenes = NuScenes(nuscenes_version, data_root)
```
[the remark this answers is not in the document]
ajrasane: Replaced the direct pickle call with `mmcv.load` and documented the trust boundary: the utility accepts metadata generated locally by the preceding documented MMDetection3D preparation step, not untrusted serialized input. No security suppression is used.

examples/onnx_ptq/petr/prepare_sweep_metadata.py
```diff
@@ -0,0 +1,160 @@
...
+            sweep_cams = {}
+            for cam in sensors:
+                if current_cams[cam]["prev"] == "":
+                    sweep_cams = sweep_lists[-1]
+                    break
+                sample_data = nuscenes.get("sample_data", current_cams[cam]["prev"])
+                sweep_cam = add_frame(sample_data, e2g_t, l2e_t, l2e_r_mat, e2g_r_mat)
+                current_cams[cam] = sample_data
+                sweep_cams[cam] = sweep_cam
+            sweep_lists.append(sweep_cams)
```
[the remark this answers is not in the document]
ajrasane: Guarded the empty-list case before reusing the last sweep, so a camera without a previous sample no longer indexes `sweep_lists[-1]` when no sweep has been collected.

examples/onnx_ptq/petr/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+```bash
+for version in v1 v2; do
+  model="PETR${version}"
+  python -m onnxsim \
+    "onnx_files/${model}.extract_feat.onnx" \
+    "onnx_files/sim_${model}.extract_feat.onnx"
+  python -m onnxsim \
+    "onnx_files/${model}.pts_bbox_head.forward.onnx" \
+    "onnx_files/sim_${model}.pts_bbox_head.forward.onnx"
```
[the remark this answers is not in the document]
ajrasane: Updated simplification to use `/opt/petr/bin/python -m onnxsim`. Model Optimizer AutoCast and quantization use the base environment through `env -u PYTHONPATH`; both the AutoCast CLI and a real graph conversion were validated.

examples/onnx_ptq/petr/requirements.txt
```diff
@@ -0,0 +1,25 @@
+--extra-index-url https://pypi.nvidia.com
+--find-links https://download.openmmlab.com/mmcv/dist/cu117/torch1.13.0/index.html
+
+einops
+ipython<9
+lyft-dataset-sdk
```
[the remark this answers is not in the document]
ajrasane: Pinned uv, Python 3.8.20, every PETR requirements entry, and the external repository commits. A pip-only hashed lock would still leave the NGC base, apt packages, custom CUDA indexes, and source installs outside the lock and is not used by neighboring examples. The rebuilt environment passes `pip check`, imports, quantization, and inference smokes.

examples/onnx_ptq/petr/requirements.txt
```diff
@@ -0,0 +1,25 @@
...
+mmsegmentation==0.30.0
+networkx>=2.2
+numba==0.57.1
+numpy<1.24
+nuscenes-devkit
+onnx
+onnx-graphsurgeon==0.6.1
+onnxruntime
+onnxsim
+opencv-python==4.5.5.64
```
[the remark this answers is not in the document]
ajrasane: Upgraded and pinned `opencv-python==4.8.1.78`. The rebuilt legacy stack imports successfully, and decoding plus resizing a real nuScenes image produced the same bytes as the previous version.

examples/onnx_ptq/petr/requirements.txt
```diff
@@ -0,0 +1,25 @@
...
+plyfile
+pycuda
+scikit-image
+setuptools<81
+tensorrt-cu13-bindings==11.1.0.106
```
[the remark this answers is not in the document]
ajrasane: Added the TensorRT binding justification to the PR body: the isolated Python 3.8 environment needs bindings matching the TensorRT 11.1 runtime because the base environment uses a newer Python. `@NVIDIA/modelopt-setup-codeowners` is now explicitly requested for approval.

examples/onnx_ptq/petr/requirements.txt
ajrasane: Validated the package licenses against their metadata. `plyfile==1.0.3` remains GPL-3.0-or-later (not BSD) and is the last Python 3.8-compatible release; `lyft-dataset-sdk==0.0.8` is CC BY-NC-SA 4.0 (not Apache) and is eagerly required by the legacy MMDetection3D dataset imports. Both are pinned and disclosed in the PR body, the nuScenes terms are linked, and setup-codeowner approval is requested.

ajrasane:
Implemented the review follow-ups:

- shared the calibration readers, VoVNet exclusions, and name-keyed TensorRT runner with FAR3D;
- removed the duplicate FP16 converter and documented the shipped AutoCast CLI;
- added name/shape-safe calibration capture, sweep overwrite/empty-history guards, and optional plugin handling;
- pinned the legacy environment and source revisions, upgraded OpenCV, corrected container/data setup, and added non-root runtime guidance;
- aligned copied-source headers, exact source links, license notices, and dependency/security disclosures; and
- requested `@NVIDIA/modelopt-setup-codeowners` review for the disclosed non-permissive/proprietary dependencies and adapted-source handling.

Validation includes all applicable pre-commit hooks, a rebuilt image with `pip check`, PETRv1/PETRv2 inference and calibration smokes, fresh INT8/FP8 quantization, AutoCast conversion, and a FAR3D regression smoke. Only FP16-head accuracy/latency results remain published.

examples/onnx_ptq/quantization_utils.py
```diff
@@ -0,0 +1,84 @@
...
+from pathlib import Path
+
+import numpy as np
+import onnx
+from onnxruntime.quantization.calibrate import CalibrationDataReader
+
+from modelopt.onnx.utils import topologically_sort_graph_nodes
+
+
+class FileCalibrationReader(CalibrationDataReader):
```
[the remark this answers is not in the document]
ajrasane: Added explicit `__all__` declarations for the shared calibration helpers, `TensorRTRunner`, and the PETR pipeline/runtime used across scripts. I did not add package-level star re-exports: `examples/onnx_ptq` and `examples/onnx_ptq/petr` are executable example directories without `__init__.py`, and turning them into packages would eagerly import optional TensorRT/MMDetection stacks and initialize TensorRT plugins. The scripts retain explicit imports at their use sites.

examples/onnx_ptq/quantization_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, calibration_dir, pattern):
+        self.batch_paths = sorted(Path(calibration_dir).glob(pattern))
+        if not self.batch_paths:
+            raise ValueError(f"No {pattern} calibration batches found in {calibration_dir}")
+        self.rewind()
```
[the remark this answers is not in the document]
ajrasane: Bounded calibration processing before materialization. Both quantization CLIs now expose a positive `--max-calibration-batches` limit (default 512); NPZ files are rejected above 128 MiB compressed or expanded, above 32 million tensor elements, for object arrays, or for unexpected archive members, and `allow_pickle=False` is explicit. Validation loaded the real 106 MB FAR3D decoder batch and all PETR layouts, while byte, element, and batch-count overflow cases were rejected.

examples/onnx_ptq/trt_runner.py
```diff
@@ -0,0 +1,148 @@
...
+
+        outputs = {}
+        for name, shape in self.output_shapes.items():
+            output = aligned_tensor(shape, self.tensor_dtypes[name], "cuda")
+            outputs[name] = output
+            self.context.set_tensor_address(name, output.data_ptr())
+
+        if not self.context.execute_async_v3(stream.cuda_stream):
+            raise RuntimeError("TensorRT execution failed")
+        stream.synchronize()
```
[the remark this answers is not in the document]
ajrasane: Removed the per-engine host synchronization. Dependent TensorRT and PyTorch work remains on the supplied stream; PETR and FAR3D now hand that stream to the current stream with `wait_stream()` before CPU reads. Exact synchronized-baseline comparisons passed for two PETRv1 INT8/FP16 samples, two PETRv2 FP8/FP16 samples, and two consecutive FAR3D INT8/FP16 samples, including decoder state carryover.

examples/onnx_ptq/far3d/quantize.py
```diff
@@ -14,80 +14,27 @@
...
-    downstream_tensors = set()
-    for node in graph.node:
-        is_osa = "OSA4_5" in node.name
-        is_downstream = any(name in downstream_tensors for name in node.input)
-        if is_osa or is_downstream:
-            excluded.add(node.name)
-        if "lateral_convs" in node.name or (is_downstream and not is_osa):
-            downstream_tensors.update(node.output)
-    return sorted(excluded)
+        return {"img": np.load(batch_path, allow_pickle=False)}
```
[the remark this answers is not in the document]
ajrasane: Added a shared `NpyCalibrationReader` and wired FAR3D encoder calibration through it. Before `np.load`, it opens and retains the same regular-file descriptor, enforces configurable 128 MiB and 32-million-element limits, rejects object arrays, and validates rank, static shape, and dtype against the named ONNX input. Validation loaded the real 51.6 MB `(1, 7, 640, 960, 3)` float32 batch, rejected byte, element, shape, dtype, and object cases before materialization, and completed a one-batch FAR3D INT8 encoder quantization.

examples/onnx_ptq/quantization_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        max_archive_bytes=MAX_NPZ_BYTES,
+        max_tensor_elements=MAX_TENSOR_ELEMENTS,
+    ):
+        if max_archive_bytes < 1 or max_tensor_elements < 1:
+            raise ValueError("NPZ byte and tensor-element limits must be positive")
+        graph = onnx.load(onnx_path, load_external_data=False).graph
+        self.input_dtypes = {
+            value.name: onnx.helper.tensor_dtype_to_np_dtype(value.type.tensor_type.elem_type)
+            for value in graph.input
+        }
```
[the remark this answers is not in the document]
ajrasane: All shared ONNX parsing now uses a same-descriptor regular-file loader with a configurable 512 MiB limit enforced before parsing; its bounded read also detects file growth after the initial size check. This covers the NPY and NPZ readers plus VoVNet exclusion discovery. The real FAR3D and PETR graphs fit the limit, non-regular and one-byte-limit cases were rejected, and one-batch FAR3D and PETR INT8 quantization completed successfully.

examples/onnx_ptq/quantization_utils.py
```diff
@@ -0,0 +1,231 @@
...
+            ):
+                raise ValueError(
+                    f"{batch_path} shape {shape} does not match ONNX input shape {self.input_shape}"
+                )
+            if dtype != self.input_dtype:
+                raise ValueError(
+                    f"{batch_path} dtype {dtype} does not match ONNX input dtype {self.input_dtype}"
+                )
+            array_file.seek(0)
+            return {self.input_name: np.load(array_file, allow_pickle=False)}
```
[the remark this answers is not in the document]
ajrasane: Added pre-materialization NPY/NPZ validation for exact declared-versus-remaining payload bytes, aggregate source and converted byte limits, tensor-element limits, CRC/truncation, exact member names, ONNX rank/static dimensions, and dtype. Wrong shape/dtype/name, object arrays, truncated standalone NPY, and truncated NPZ members are rejected before `np.load`; all four PETR layouts and the real FAR3D encoder/decoder layouts pass.

Dtype validation is exact by default. The pinned FAR3D exporter records `timestamp` as FP64 while the TensorRT calibration callback supplies FP32, so only that named input explicitly permits NumPy-safe FP32→FP64 widening; unsafe conversions remain rejected and converted bytes are bounded before `astype`. A full one-batch FAR3D INT8 encoder+decoder quantization completed successfully with this path.

examples/onnx_ptq/quantization_utils.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+class FileCalibrationReader(CalibrationDataReader):
+    def __init__(self, calibration_dir, pattern, max_batches=MAX_CALIBRATION_BATCHES):
+        if max_batches < 1:
+            raise ValueError("max_batches must be positive")
+        batch_paths = list(islice(Path(calibration_dir).glob(pattern), max_batches + 1))
+        if not batch_paths:
+            raise ValueError(f"No {pattern} calibration batches found in {calibration_dir}")
+        if len(batch_paths) > max_batches:
```
cjluo-nv: > _Bot comment._

This makes `max_batches` a hard failure rather than a cap, which contradicts the CLI help ("Maximum number of calibration batches to **load**") and is a behavior regression for FAR3D: `far3d/quantize.py` previously consumed every `*.npy`/`*.npz` in the directory, so anyone who ran `prepare_calibration.py --num-samples 1000` (or accumulated batches across runs) now gets a hard error from a command that used to work, with no hint that `--max-calibration-batches` is the knob.

Suggest either (a) truncating to the first `max_batches` after sorting, so the flag actually means "load at most N" and selection is deterministic, or (b) keeping the raise but renaming the flag/help to something like `--calibration-batch-limit` and mentioning it in the error message and both READMEs. Note also that as written the `islice` is taken over the unordered `glob()` before `sorted()`, so if you do switch to truncation the sort must come first.

examples/onnx_ptq/trt_runner.py
```diff
@@ -0,0 +1,149 @@
...
+                    f"Input {input_key} has shape {tuple(value.shape)}, expected {shape}"
+                )
+        return input_key, value
+
+    def __call__(self, stream, **inputs):
+        input_buffers = {}
+        callback_inputs = {}
+        for name, shape in self.input_shapes.items():
+            if name in self.state:
+                callback_name = name.rsplit(".1", maxsplit=1)[0]
```
cjluo-nv: > _Bot comment._

This strips `.1` unconditionally, unlike `resolve_name()`/`input_key()` which guard with `endswith(".1")`. For a state tensor whose engine name doesn't carry the suffix it happens to be a no-op today, but any name containing `.1` elsewhere (e.g. `memory_x.10`) is silently truncated and the calibration NPZ key would then not match any ONNX input — which the PETR writer would reject and the FAR3D writer would happily save under the wrong name. Please reuse the same guarded helper here (`base_name = name[:-2] if name.endswith(".1") else name`, or factor it out of `input_key`).

cjluo-nv (commented):

Re-review of the PETR ONNX PTQ example. Most of the previous round's critical items are genuinely resolved, but two moderate issues remain plus the licensing block, so this can't be approved.

**Previously flagged — resolved (verified in the diff):**
- `convert_to_fp16.py` deleted; README now uses the shipped `python -m modelopt.onnx.autocast` CLI. ✅
- FAR3D copy-paste extracted into `examples/onnx_ptq/trt_runner.py` + `quantization_utils.py`, and FAR3D was migrated to them (net −197 lines there). Engine inputs are now bound by name (`input_key`) rather than positionally. ✅
- `CalibrationWriter` validates the exact engine/ONNX input-name set and writes name-keyed NPZ. ✅
- `prepare_sweep_metadata.py` has `main()`/`__main__`, refuses to overwrite the output pkl, uses `mmcv.load`, and guards the empty `sweep_lists[-1]` case. ✅
- README ordering (container-first), `nuscenes_infos_val.pkl` generation step, pinned upstream commits, `/opt/petr/bin/python -m onnxsim`, `cfg.get("plugin_dir")`, `max_samples >=`, `math.prod`. ✅
- Stream-sync removal in the shared runner is compensated correctly: both pipelines call the engines inside `with torch.cuda.stream(stream)` and `wait_stream()` before host reads, and FAR3D's decoder state copies stay ordered on the same stream. ✅

**Still open / new:**
1. `FileCalibrationReader` turns `--max-calibration-batches` into a hard failure rather than a cap (inline comment). This is a behavior regression for FAR3D: a calibration dir with more than 512 batches used to quantize fine and now raises.
2. Cross-example contradiction is still unresolved: `petr/README.md` states "TensorRT 11 uses strongly typed networks and no longer accepts `--fp16`", while `far3d/README.md` — same `nvcr.io/nvidia/pytorch:26.07-py3` / TRT 11.1 base — still instructs `trtexec --onnx=far3d.encoder.onnx --fp16`. This PR already touches FAR3D's `evaluate.py`/`quantize.py`; if the PETR statement is right, the FAR3D instructions are broken and should be fixed here or in an immediate follow-up.
3. Design gate (PR crossed the complexity threshold): the new shared `examples/onnx_ptq/quantization_utils.py` is 295 lines, most of it a bespoke NPY/NPZ validation framework (byte/element budgets, NPY header parsing, zip-member allow-listing, per-input `safe_cast_inputs`). The repo already ships `CalibrationDataReader` implementations in `modelopt/onnx/quantization/calib_utils.py` (`CalibrationDataProvider`, `RandomDataProvider`), which is where a directory-streaming, shape/dtype-validating reader would naturally live — it runs in the base env where `modelopt` is importable (the module already imports `modelopt.onnx.utils`), it would be unit-testable, and both examples plus `--calibration_data` users would benefit. The PR body doesn't say why this went into `examples/` instead of extending `calib_utils`. (By contrast, `trt_runner.py` living under `examples/` is well justified — it must run in the legacy Py3.8 env without modelopt — worth stating that in the PR body too.) Related: none of this validation logic has automated tests; CodeRabbit explicitly asked for malformed/truncated-payload, wrong-shape and wrong-dtype coverage and the reply describes manual smoke runs only. A small pytest module under `tests/` for `quantization_utils` (no GPU/TRT needed) would be cheap and would pin the `safe_cast_inputs` FP32→FP64 exception.
4. Licensing — explicitly not approving on this axis, human/OSRB sign-off required: `LICENSE` gains a new copyright holder (Megvii), two files are adapted third-party sources (DL4AGX/OpenMMLab, PETR/Megvii) with new `.pre-commit-config.yaml` license-hook exclusions, and the new environment pulls `plyfile==1.0.3` (GPL-3.0-or-later per the author's own check), `lyft-dataset-sdk==0.0.8` (CC BY-NC-SA 4.0) and proprietary `tensorrt-cu13-bindings`. The disclosures in the PR body are good; `@NVIDIA/modelopt-setup-codeowners` still needs to sign off before merge.

Minor: `sys.path` bootstrapping is inconsistent across the new/changed scripts (`parents[3]` + `from examples.onnx_ptq.trt_runner import ...` in the evaluate scripts vs `parents[1]` + bare `from quantization_utils import ...` in the quantize scripts). The latter prepends `examples/onnx_ptq/` to `sys.path[0]`, which puts `evaluate.py`/`image_prep.py` ahead of same-named site-packages modules (e.g. HF `evaluate`); picking one pattern would be safer.

examples/onnx_ptq/quantization_utils.py
ajrasane: Updated `FileCalibrationReader` to sort all matching paths before selecting the first `max_batches`, so the option is now a deterministic cap and extra batches no longer cause a failure. Added a CPU regression test covering three files with a two-batch cap. Fixed in `2e69a6e9c`.

examples/onnx_ptq/trt_runner.py
ajrasane: Factored a guarded `_base_tensor_name` helper and reused it for both input lookup and state callback names. It strips only a terminal `.1`; names such as `memory_x.10` and `memory_x.11` remain unchanged. Fixed in `2e69a6e9c`.

ajrasane:
Addressed the remaining code and documentation feedback in `2e69a6e9c`:

- Corrected the FAR3D TensorRT 11 workflow to build accurately named calibration engines without removed precision flags. Both engines were built with TensorRT 11.1 and used to collect two encoder and decoder calibration batches.
- Kept the bounded archive readers example-scoped because their file layout, limits, and FAR3D dtype exception are workflow-specific; this rationale is now in the PR description.
- Added nine hermetic CPU tests covering valid NPY/NPZ batches, deterministic batch capping, truncated payloads, shape/dtype mismatches, and allowlisted safe casts.
- Standardized PETR and FAR3D quantizer imports on the repository-root namespace pattern.

The dependency/source licensing approval remains requested from the setup codeowners; no additional code change is requested in that thread.

ajrasane:
@coderabbitai review

ajrasane:
The new CPU coverage exposed a Windows portability issue: `os.O_NONBLOCK` is not available there. Fixed the file opener in `82f926d59` by using platform-safe flag fallbacks (including binary mode where available), and extended the valid-reader test to exercise the missing-flag path. The focused suite remains 9/9 passing and all applicable pre-commit hooks pass.

ajrasane:
@coderabbitai review

kevalmorabia97:
## Dependency license review

Diffed the PR's dependency surfaces against `pyproject.toml`, `uv.lock`, and every existing `examples/**/requirements*.txt`, then verified each license against **the pinned version's** upstream `LICENSE` file / wheel `dist-info` — not just the trove classifier, which is wrong or misleading in two cases below.

Scope note: a package is only counted as *new* if it appears in none of `pyproject.toml`, `uv.lock` (including as an indirect dependency), or any other example's requirements file. That rules out most of what the PR adds — `einops`, `ipython`, `mmcv-full`, `mmdet`, `mmsegmentation`, `networkx`, `onnxsim`, `opencv-python`, `tensorrt-cu13-bindings`, and `yapf` are all already present, nearly all via `examples/onnx_ptq/far3d/requirements.txt`.

### A. New direct dependencies

| Dependency | Pinned | License (SPDX) | Verified from |
|---|---|---|---|
| `numba` | 0.53.0 | BSD-2-Clause | upstream `LICENSE` @ `0.53.0` (no "Neither the name" clause) |
| `nuscenes-devkit` | 1.1.11 | Apache-2.0 :warning: | wheel `dist-info/LICENSE.txt` (1.1.11) |
| `pycuda` | 2026.1 | MIT/X11 | upstream `LICENSE` |
| `scikit-image` | 0.19.3 | BSD-3-Clause | upstream `LICENSE.txt` @ `v0.19.3` |
| `trimesh` | 2.35.39 | MIT | upstream `LICENSE.md` |

Four of these five (`numba`, `nuscenes-devkit`, `scikit-image`, `trimesh`) are mmdet3d `v1.0.0rc6`'s own `requirements/runtime.txt` pins hoisted into the PR; the pins match upstream exactly. `pycuda` is the only one that is not — see finding 3.

### B. New transitive dependencies

Pulled in by the above, unpinned, so exact terms follow resolution:

| Dependency | License (SPDX) | Verified from |
|---|---|---|
| `llvmlite` | BSD-2-Clause AND Apache-2.0 WITH LLVM-exception | upstream `LICENSE` @ `v0.36.0` |
| `pycocotools` | BSD-2-Clause | upstream `license.txt` (cocoapi) |
| `shapely` | BSD-3-Clause | upstream repo |
| `scikit-learn` | BSD-3-Clause | upstream repo |
| `imageio` | BSD-2-Clause | upstream repo |
| `tifffile` | BSD-3-Clause | upstream repo |
| `PyWavelets` | MIT | upstream repo |
| `pyquaternion` | MIT | upstream repo |
| `matplotlib` | matplotlib License (PSF-style, BSD-compatible) | PyPI metadata |
| `mmcls` | Apache-2.0 | PyPI metadata |
| `prettytable` | BSD-3-Clause | PyPI metadata |
| `terminaltables` | MIT | PyPI metadata |
| `pytools` | MIT | PyPI metadata |
| `descartes` | BSD | PyPI metadata |

All permissive. No GPL/AGPL and no copyleft anywhere in the new set.

Already covered elsewhere, so not listed: `pillow`, `fire`, `cachetools`, `six`, `mako`, `platformdirs` (all in `uv.lock`) and `tensorboard` (`examples/llm_qat/requirements.txt`).

### C. Non-pip components

| Component | Status | License | Verified from |
|---|---|---|---|
| MMDetection3D v1.0.0rc6 | package not new (`far3d/requirements-mmdet3d.txt` pins the same version), but **sourcing is** — git clone at commit `47285b3` + local patch instead of the PyPI wheel | Apache-2.0 | upstream `LICENSE` @ `v1.0.0rc6` |
| `nvcr.io/nvidia/pytorch:26.07-py3` | not new — identical base image to `examples/onnx_ptq/far3d/Dockerfile` | NVIDIA Deep Learning Container License | NGC container terms |
| `megvii-research/PETR` | new (cloned per README) | Apache-2.0 | upstream `LICENSE` |
| `NVIDIA/DL4AGX` (`AV-Solutions/petr-trt`) | new (cloned per README) | Apache-2.0 | subdir `LICENSE` |
| nuScenes **dataset** | new (user-supplied) | CC BY-NC-SA 4.0 (non-commercial) | nuscenes.org terms, linked in README |

### D. Findings

**1. The mmdet3d patch is doing real license work — worth saying so in the description.** `mmdet3d-nuscenes-only.patch` strips two entries from `requirements/runtime.txt`, both of which would otherwise have been *new* dependencies for this repo:

- `plyfile` — **GPL-3.0-or-later**
- `lyft_dataset_sdk` — **CC BY-NC-SA 4.0** (non-commercial)

The PR body describes this as removing "unused eager Lyft imports" and their runtime requirements. It is in fact the difference between a fully permissive dependency tree and one containing GPLv3 plus a non-commercial license. Recommend stating that explicitly for the `@NVIDIA/modelopt-setup-codeowners` review — it is the single strongest license argument in the PR.

**2. `nuscenes-devkit` metadata is self-contradictory, and it is genuinely new.** The 1.1.11 wheel ships an Apache-2.0 `LICENSE.txt` (Copyright 2021 Motional) but also carries `Classifier: License :: Free for non-commercial use`; GitHub resolves the repo to `NOASSERTION` for the same reason. The devkit *code* is Apache-2.0 — the non-commercial term belongs to the nuScenes *dataset* (CC BY-NC-SA 4.0). The README already links the terms of use; a one-line note distinguishing devkit-code from dataset would preempt the license-scanner flag this will trip, since this is the first time the package enters the repo.

**3. `pycuda==2026.1` appears unused.** It is the one new direct dependency not inherited from mmdet3d, and it has no corresponding import — `pycuda` occurs exactly once in the full PR diff, on the requirements line itself. `examples/onnx_ptq/trt_runner.py` does its device work through `torch`, not PyCUDA. If the DL4AGX exporters don't need it either, dropping it also removes the new `pytools` transitive.

### Not new, but flagged for completeness

These were already in the repo before this PR, so they are not blockers here, but they carry terms worth knowing about:

- `opencv-python` (already in `examples/diffusers`, `far3d`, `windows/onnx_ptq/sam2`; this PR bumps it to 4.8.1.78). Per `cv2/LICENSE-3RD-PARTY.txt` in the Linux wheel, the non-headless build bundles FFmpeg + libvpx under **LGPL-2.1** and Qt 5 under **LGPL-3.0**. Harmless for a locally built image; relevant if one is ever published. `opencv-python-headless` drops the Qt5/LGPL-3.0 payload.
- `mmcv-full` (already in `far3d`). `LICENSES.md` @ `v1.7.0` lists `upfirdn2d_kernel.cu` and `fused_bias_leakyrelu_cuda.cu` under the **NVIDIA Source Code License** (non-commercial, from StyleGAN2). PETR does not call these ops, but they are compiled into the installed binary.
- `tensorrt-cu13-bindings` (already in `far3d`). The bundled `LICENSE.txt` is the NVIDIA TensorRT SLA; §2.1 bars redistribution, sublicensing, and derivative works. Routing to `@NVIDIA/modelopt-setup-codeowners` is the right call.

### Bottom line

Five new direct dependencies and fourteen new transitives, all permissive (BSD-2, BSD-3, MIT, Apache-2.0, PSF-style). No copyleft and no non-commercial license enters the repo — specifically because the mmdet3d patch removes `plyfile` and `lyft_dataset_sdk`. Suggested pre-merge changes: (1) call out that removal as license hygiene in the PR body, (2) add a one-line devkit-vs-dataset note for `nuscenes-devkit`, (3) drop the unused `pycuda`.

.pre-commit-config.yaml
```diff
@@ -127,6 +127,9 @@ repos:
               examples/llm_eval/mmlu.py|
               examples/llm_eval/modeling.py|
               examples/onnx_ptq/far3d/evaluate.py|
+              examples/onnx_ptq/petr/evaluate.py|
```
kevalmorabia97, line 130: have you updated nvbug 6046893 with all 3?

kevalmorabia97 (approved):
approving as modelopt-setup-codeowners

examples/onnx_ptq/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+
+COPY examples/onnx_ptq/requirements-evaluator.txt /tmp/requirements-evaluator.txt
+# Avoid the training and development dependencies declared by the legacy projects.
+RUN env -u PIP_CONSTRAINT PIP_CONFIG_FILE=/dev/null FLASH_ATTENTION_FORCE_BUILD=TRUE MAX_JOBS=8 \
+        python -m pip install --no-cache-dir --no-build-isolation \
+        -r /tmp/requirements-evaluator.txt && \
+    env -u PIP_CONSTRAINT PIP_CONFIG_FILE=/dev/null python -m pip install \
+        --no-cache-dir --no-build-isolation --no-deps \
+        av2==0.2.1 \
+        lyft-dataset-sdk==0.0.8 \
```
cjluo-nv: > _Bot comment._

`lyft-dataset-sdk==0.0.8` is CC BY-NC-SA 4.0 (non-commercial). The setup-codeowner license review on this PR concluded the dependency tree was fully permissive *because* the mmdet3d patch removed `plyfile` and `lyft_dataset_sdk`; installing it explicitly here reverses that conclusion.

Two asks:
1. Re-confirm with `@NVIDIA/modelopt-setup-codeowners` — the repo rule routes non-permissive additions in `pyproject.toml`/`requirements*.txt` for approval, and `RUN pip install` lines in a Dockerfile bypass that path filter entirely. The same applies to `av2`, `mmdet3d`, `nuscenes-devkit`, `refile` here and `flash-attn==0.2.8` in `requirements-evaluator.txt`.
2. If `lyft-dataset-sdk` is only needed to satisfy an eager `mmdet3d` dataset import (it was described that way earlier), say so in a comment here, or keep it out and stub the import as the previous revision did.

Also worth a note in the Dockerfile: this whole block relies on `--no-deps` plus whatever the NGC 22.06 base already ships (numba, scikit-image, networkx, pandas, cachetools, fire…). That's an implicit contract that will break silently if the base image is ever bumped.

examples/onnx_ptq/petr/README.md
```diff
@@ -0,0 +1,175 @@
+# PETR ONNX PTQ and nuScenes evaluation
+
+This example quantizes the VoVNet image backbone in PETRv1 and PETRv2 to INT8 or FP8, keeps the detection head in mixed FP16/FP32, and evaluates TensorRT 11.1 engines on the nuScenes validation set. It follows the [NVIDIA DL4AGX PETR workflow](https://github.com/NVIDIA/DL4AGX/tree/9f7b29104c253d5bc68334e7b83b3eecb72d4572/AV-Solutions/petr-trt).
+
+Build the shared `evaluator` and `modelopt` images as described in the [parent guide](../README.md#petr-and-far3d-containers). Use the evaluator for source setup, export, calibration, and accuracy evaluation. Use the ModelOpt image for AutoCast, quantization, and TensorRT engine builds.
+
+## 1. Prepare PETR and nuScenes
+
+Place the nuScenes data and pre-generated `nuscenes_infos_val.pkl` and `mmdet3d_nuscenes_30f_infos_val.pkl` metadata files in the host dataset directory. Then start the evaluator with the workspace and dataset mounted:
```
cjluo-nv: > _Bot comment._

This asks the reader to place "pre-generated `nuscenes_infos_val.pkl` and `mmdet3d_nuscenes_30f_infos_val.pkl`" in the dataset directory, but nothing in the repo or the README says how to produce either one. `nuscenes_infos_val.pkl` comes from mmdetection3d's `create_data.py`, and the 30-frame sweep pkl is not part of the nuScenes release at all — the previous revision of this PR shipped `prepare_sweep_metadata.py` and a README step for exactly this, and both were dropped in the simplification.

As written the PETR workflow can't be reproduced end-to-end. Please either re-add the generation step (a command pointing at mmdet3d `create_data.py` + PETR's `tools/generate_sweep_pkl.py` would be enough, no new script needed) or link the upstream instructions.

CHANGELOG.rst
```diff
@@ -11,6 +11,7 @@ Changelog
 - Add a Muse Glimmer AutoQuantize recipe that searches language-model MLP projections, self-attention projections, and ``lm_head`` over W4A16 NVFP4 Four-Over-Six, FP8, and BF16 fallback at 5.5 effective bits while leaving the vision tower unquantized.
 - Add ``examples/alpamayo/qad.py``, which runs quantization-aware distillation on the quantized Alpamayo checkpoint produced by ``examples/alpamayo/quantize.py``. It distills the quantized VLM against the original FP16 VLM with ``QADTrainer``, supports FSDP2 for multi-GPU runs, and ``--export`` reassembles the trained VLM into a full AlpamayoR1 checkpoint that ``AlpamayoR1.from_pretrained`` can reload.
 - Add a calibration-free streaming Kimi-K3 converter and checkpoint-mirror recipe for NVFP4 routed experts with ``input_scale=1.0`` and 128x128 block-FP8 KDA/MLA attention weights. The converter operates shard-by-shard on the source checkpoint's packed MXFP4 experts instead of loading the 2.8T model through the in-memory ``hf_ptq.py`` path.
+- Add end-to-end PETRv1 and PETRv2 ONNX PTQ examples covering calibration, INT8 and FP8 VoVNet backbone quantization, TensorRT deployment, and accuracy evaluation.
```
cjluo-nv, line 28: > _Bot comment._

This entry covers the new PETR example, but the PR also removes documented FAR3D functionality: `far3d/quantize.py` and its `--quantization-mode` / `--fp16-decoder` CLI, decoder quantization and decoder calibration capture, `far3d/Dockerfile` and its three requirements files, `far3d_optional_flash_attn.patch`, and the published TensorRT 10.11 latency/accuracy table.

Anyone following the current FAR3D README will hit missing files after upgrading. Please add a second changelog line describing the FAR3D workflow change (shared containers, encoder-only quantization via `quantize_vovnet.py`, decoder kept in exported mixed precision).

cjluo-nv (commented):

Re-review of the PETR/FAR3D ONNX PTQ example after the "simplification" pass (23 files, +1184/-575). The rework is a genuine improvement: `far3d/quantize.py`, its Dockerfile, three requirements files, the flash-attn patch, `convert_to_fp16.py` and `prepare_sweep_metadata.py` are gone; the TRT runner, NPZ calibration reader/writer and VoVNet exclusion helper are shared under `examples/onnx_ptq/`; and a single two-stage Dockerfile replaces the per-example ones.

**Previously flagged — now resolved (verified in the tree):**
- `FileCalibrationReader`'s `max_batches` hard-failure is gone; the readers stream the whole directory again, so the FAR3D regression I raised no longer applies.
- The `--fp16` contradiction between the two READMEs is fixed: `far3d/README.md` now builds with `--skipInference` only and states the TRT 11.1 typed-graph rule.
- The 295-line bespoke NPY/NPZ validation framework I raised under the design gate is deleted. What remains (`quantization_utils.py`, 111 lines) is a directory-streaming reader/writer that `modelopt/onnx/quantization/calib_utils.py::CalibrationDataProvider` genuinely doesn't cover (that one takes one in-memory dict of stacked arrays), so I'm satisfied on that axis. The `modelopt` import inside `find_vovnet_nodes_to_exclude` carries a concrete justification (evaluator image has no modelopt) — fine.
- `sys.path` bootstrapping is now consistent (`parents[N]` + `from examples.onnx_ptq...` everywhere), and there are hermetic CPU tests for the writer/reader/exclusion helper and the external-data-preserving temp copy. `python -m pytest tests/unit` puts the repo root on `sys.path`, so the `from examples.onnx_ptq import ...` imports resolve.
- Nice incidental correctness fix: FAR3D calibration now captures `get_image_input()` (with the `.flip(2)`) instead of the unflipped tensor, so calibration finally matches inference preprocessing.

**Open items (inline):**
1. `lyft-dataset-sdk==0.0.8` (CC BY-NC-SA 4.0, non-commercial) is now installed directly in the new Dockerfile, together with `av2`, `mmdet3d`, `nuscenes-devkit`, `refile` and `flash-attn`. The setup-codeowner license review that signed this off explicitly concluded "no copyleft and no non-commercial license enters the repo — specifically because the mmdet3d patch removes `plyfile` and `lyft_dataset_sdk`"; that premise no longer holds, and `RUN pip install` lines are outside the `**/requirements*.txt` path filter that routes dependency changes to `@NVIDIA/modelopt-setup-codeowners`. Please re-confirm with them.
2. `petr/README.md` now tells the user to supply pre-generated `nuscenes_infos_val.pkl` and `mmdet3d_nuscenes_30f_infos_val.pkl` but no longer documents how to produce them (the generator script was deleted in this pass). The 30-frame sweep pkl is not part of the nuScenes release, so the workflow isn't reproducible as written — this is a regression of a point that was fixed in the previous round.
3. CHANGELOG only mentions the new PETR example; the FAR3D user-facing removals (decoder quantization, `--quantization-mode` / `--fp16-decoder`, the decoder calibration path, the per-example Dockerfile/requirements, the latency table) aren't recorded anywhere a FAR3D user would see.

Not approving on licensing grounds regardless: `LICENSE` gains a new copyright holder, two files carry adapted third-party headers, and `.pre-commit-config.yaml` grows two license-hook exclusions — @kevalmorabia97's question on that line ("have you updated nvbug 6046893 with all 3?") still looks unanswered.

Minor, non-blocking: the `NpzCalibrationWriter` no longer validates array shape against the ONNX declared shape (only dtype), so a mismatched capture surfaces later inside ORT rather than at write time; and `far3d/import_plugin` still reads `cfg.plugin_dir` directly while the PETR copy uses `cfg.get("plugin_dir")`.

.pre-commit-config.yaml
ajrasane: Updated [6046893]. I kept row 19 for the older FAR3D-only integration and added row 20 for PR #2180, commit `dee7fd3e9`, covering all three adapted files: `examples/onnx_ptq/far3d/evaluate.py`, `examples/onnx_ptq/petr/evaluate.py`, and `examples/onnx_ptq/trt_runner.py`. The new row records the pinned public sources and Apache-2.0 licenses, and each file retains its applicable copyright and license notice. Thanks for flagging this.

examples/onnx_ptq/Dockerfile
ajrasane: Moved `av2`, `lyft-dataset-sdk`, `mmdet3d`, `nuscenes-devkit`, and `refile` into `requirements-evaluator-nodeps.txt`; the Dockerfile installs that manifest with `--no-deps`.

`flash-attn` remains in the main evaluator requirements because its compiled installation uses the evaluator build step. The Dockerfile now documents that the no-deps pins rely on the digest-pinned PyTorch 22.06 base and that Lyft is required only by mmdet3d's eager dataset import; neither example uses Lyft data.

I requested a fresh review from `@NVIDIA/modelopt-setup-codeowners`. I am leaving this thread open until that approval is recorded.

examples/onnx_ptq/petr/README.md
ajrasane: Documented and tested a clean reproducible workflow without restoring a repository-owned generator:

- raw nuScenes remains mounted read-only;
- a writable dataset root contains symlinks to the raw data;
- the pinned mmdetection3d converter generates `nuscenes_infos_val.pkl`;
- a temporary, untracked copy of PETR's pinned sweep generator changes only the validation prefix and writable dataset root.

The clean run produced 6,019 records in both metadata files, and all referenced camera, lidar, and sweep paths resolve. PETRv1 and PETRv2 each passed a fresh one-batch calibration and one-sample TensorRT evaluation using the generated metadata. `prepare_sweep_metadata.py` remains deleted.

CHANGELOG.rst
ajrasane: Added a 0.47 backward-breaking changelog entry covering the shared evaluator and ModelOpt containers, replacement of the FAR3D quantizer by `quantize_vovnet.py`, encoder-only INT8/FP8 quantization, removal of decoder calibration and quantization flags, and retention of the exported mixed FP16/FP32 decoder.

ajrasane:
Addressed both non-blocking observations:

- `NpzCalibrationWriter` now rejects rank and static-dimension mismatches before writing while accepting symbolic or otherwise dynamic ONNX dimensions. Existing dtype conversion and the public interface are unchanged.
- FAR3D plugin loading now checks both `cfg.get("plugin")` and `cfg.get("plugin_dir")`, matching PETR.

The focused CPU suite passes 8 tests, including wrong-rank, wrong-static-dimension, and accepted-dynamic-dimension cases. The broader ONNX quantization CPU suite passes 324 tests.

galagam (commented):
Overall looks good, but it's probably better to wait for @cjluo-nv's approval here, so  holding off on approval from my side.

cjluo-nv (commented):

Re-review #3 of the PETR/FAR3D ONNX PTQ example (24 files, +1276/-577). All three blockers from the previous round are addressed in the tree:

**Resolved (verified):**
- `lyft-dataset-sdk==0.0.8`, `av2`, `mmdet3d`, `nuscenes-devkit`, `refile` moved out of the Dockerfile `RUN pip install` line into `examples/onnx_ptq/requirements-evaluator-nodeps.txt`, so the `**/requirements*.txt` CODEOWNERS filter now routes them; installed with `--no-deps`, which also keeps GPL `plyfile` out.
- `petr/README.md` reproducibility regression fixed: the workflow now documents a writable symlinked dataset view, the pinned mmdetection3d `create_nuscenes_infos` call, and a `sed`-derived temporary copy of PETR's pinned `generate_sweep_pkl.py`, with `test -s` guards on both pkls.
- CHANGELOG now carries a backward-breaking entry for the FAR3D migration (shared containers, `quantize_vovnet.py`, encoder-only INT8/FP8, removed decoder flags).
- Non-blocking items from last round also landed: `NpzCalibrationWriter` validates rank + static dims, `far3d/import_plugin` uses `cfg.get("plugin_dir")`, `max_batches` hard-fail is gone, both pipelines wrap engine calls in `with torch.cuda.stream(stream)` and `wait_stream()` before host reads.

**Design gate (PR crossed the complexity threshold):** re-checked. The shared surface is now small (`quantization_utils.py` 133 lines, `trt_runner.py` 138, `quantize_vovnet.py` 72). `modelopt/onnx/quantization/calib_utils.py::CalibrationDataProvider` takes one in-memory dict of stacked arrays and cannot stream a directory of per-batch NPZ files, and `trt_runner.py`/the NPZ writer must run in the Python 3.8 evaluator image where `modelopt` is not installed (the local `modelopt.onnx.utils` import inside `find_vovnet_nodes_to_exclude` carries exactly that justification). I consider the design settled; a one-line "why not `calib_utils`" note in the PR body would make that explicit for the record.

Not approving on licensing grounds (see nudge reason) plus a few small residuals.

ajrasane:
Addressed the PETRv2 accuracy-methodology concern in `1ebd69392`:

- The selected historical six-camera sweep and current six-camera sweep are now both executed by the same precision-matched TensorRT backbone engine using distinct execution contexts.
- The historical pass supplies its first six output camera features to the current pass; reported evaluation no longer calls `model.extract_img_feat` or uses PyTorch-generated camera features.
- PyTorch feature extraction remains only in calibration-data preparation, which runs before a TensorRT engine exists and is not part of reported accuracy inference.
- Fresh 6,019-sample evaluations produced mAP `0.4102` for FP16, `0.3982` for INT8, and `0.4084` for FP8.
- PETRv2 performance includes both serial backbone executions plus the shared typed mixed FP16/FP32 head.

gcunhase (approved):
LGTM.

In parallel, can we investigate why FP8 is slower than INT8? This can be done in a following PR if needed.

Thanks!

cjluo-nv (approved):

PR #2150, opened 2026-08-11
Move Mistral Medium recipe under canonical base model

chadvoegele:
### What does this PR do?

Type of change: Bug fix

Moves the Mistral Medium 3.5 checkpoint-mirror PTQ recipe under the canonical Hugging Face base model, `mistralai/Mistral-Medium-3.5-128B`. Updates the recipe catalog, loader smoke-test path, and changelog.

### Usage

```bash
python examples/hf_ptq/hf_ptq.py \
    --model mistralai/Mistral-Medium-3.5-128B \
    --recipe huggingface/models/mistralai/Mistral-Medium-3.5-128B/ptq/nvfp4-max-calib
```

### Testing

- `uvx --from pre-commit pre-commit run --files CHANGELOG.rst modelopt_recipes/huggingface/models/mistralai/Mistral-Medium-3.5-128B/ptq/nvfp4-max-calib.yaml modelopt_recipes/ptq.md tests/unit/recipe/test_loader.py`
- Recipe documentation consistency checks (all passed)
- Focused pytest was unavailable locally because the checkout environment lacks pytest/Torch; the dedicated pre-commit recipe validator passed.

### Before your PR is "*Ready for review*"

- Is this change backward compatible?: ❌ The built-in recipe path changes; the changelog documents the replacement path.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A
- Did you write any new necessary tests?: ✅ Updated the built-in recipe loader smoke test.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅
- Did you get Claude approval on this PR?: N/A

### Additional Information

The recipe targets the vendor's canonical base checkpoint, [`mistralai/Mistral-Medium-3.5-128B`](https://huggingface.co/mistralai/Mistral-Medium-3.5-128B), and reproduces NVIDIA's [`Mistral-Medium-3.5-128B-NVFP4`](https://huggingface.co/nvidia/Mistral-Medium-3.5-128B-NVFP4) quantization map.

## Summary by CodeRabbit

* **New Features**
  * Added a Mistral Medium 3.5 128B post-training quantization recipe supporting NVFP4 and FP8 configurations with max calibration.

* **Documentation**
  * Updated checkpoint-mirror guidance to reflect the new model-based recipe location.

* **Breaking Changes**
  * Saved references to the previous recipe path must be updated to the new Mistral Medium 3.5 recipe path.

PR #2182, opened 2026-08-12
Remove deprecations marked in 0.45 and 0.46

kevalmorabia97:
### What does this PR do?

Type of change: Backward breaking change (deprecation removal)

Ahead of the 0.47 code freeze, this removes every deprecation still outstanding from the previous two releases (0.45 and 0.46). Two are intentionally left in place: the **Python 3.10** drop and the **transformers 4.x** drop

| Deprecation | Marked in | Replacement |
| --- | --- | --- |
| `--auto_quantize_bits` / `_method` / `_score_size` / `_cost_model` / `_active_moe_expert_ratio` | 0.46 | AutoQuantize `--recipe` |
| `examples/llm_ptq` symlink + `examples/vlm_ptq/` forwarder | 0.46 | `examples/hf_ptq` (`--vlm` for VLMs) |
| `QuantizationArgumentsWithConfig` alias | 0.45 | `QuantizationArguments` |
| `QFORMAT_ALIASES` short names | 0.45 | canonical preset basenames |
| `layerwise` bool + flat `layerwise_checkpoint_dir` | 0.45 | nested `layerwise: {enable, checkpoint_dir}` |
| in-trainer `quant_cfg` / `--quant_cfg` | 0.45 | `--recipe` |

#### Two things worth a closer look

**1. The `use_sequential` alias goes too.** It is the pre-#1251 alias on `QuantizeAlgorithmConfig.layerwise` and only ever carried a bool. Once the bool form is rejected it cannot accept a valid value, so keeping it would only produce a differently-worded validation error. Note the direction is breaking either way (`extra="forbid"`): a pre-0.45 `modelopt_state` carrying `use_sequential: True` or a top-level `layerwise_checkpoint_dir` now fails validation instead of being migrated.

**2. Removing in-trainer `--quant_cfg` required two new recipes.** The `examples/gpt-oss` QAT flow ran on `--quant_cfg MXFP4_MLP_WEIGHT_ONLY_CFG` and no `general/ptq/` recipe covered it. This PR adds `general/ptq/mxfp4_mlp_weight_only` and `general/ptq/nvfp4_mlp_weight_only`, verified to `model_dump` identical to `mtq.MXFP4_MLP_WEIGHT_ONLY_CFG` / `mtq.NVFP4_MLP_WEIGHT_ONLY_CFG`, and migrates the gpt-oss README, both SFT configs, `sft.py` and `tests/examples/gpt-oss/test_gpt_oss_qat.py`. `examples/llm_qat` was already recipe-only.

### Usage

```bash
# AutoQuantize: --auto_quantize_* flags -> an AutoQuantize recipe
scripts/huggingface_example.sh --model $HF_PATH \
  --recipe general/auto_quantize/nvfp4_fp8_at_5p4bits --calib_batch_size 4

# --qformat / --quant_cfg: short name -> canonical preset basename
#   int8_sq -> int8_smoothquant                nvfp4_mse           -> nvfp4_w4a4_weight_mse_fp8_sweep
#   int8_wo -> int8_weight_only                nvfp4_local_hessian -> nvfp4_w4a4_weight_local_hessian
#   w4a8_awq -> w4a8_awq_beta                  fp8_pb_wo           -> fp8_2d_blockwise_weight_only
#   nvfp4_awq -> nvfp4_awq_lite                fp8_pc_pt           -> fp8_per_channel_per_token
scripts/huggingface_example.sh --model $HF_PATH --quant int8_smoothquant

# VLM PTQ: examples/vlm_ptq -> examples/hf_ptq with --vlm
scripts/huggingface_example.sh --model $HF_PATH --quant fp8 --vlm

# gpt-oss QAT: --quant_cfg <CFG name> -> --recipe <recipe path>
accelerate launch --config_file configs/zero3.yaml sft.py \
  --config configs/sft_full.yaml --model_name_or_path openai/gpt-oss-20b \
  --recipe general/ptq/mxfp4_mlp_weight_only --output_dir gpt-oss-20b-qat
```

```python
# Layerwise calibration: bool / flat key -> nested LayerwiseConfig
quant_cfg["algorithm"] = {"method": "gptq", "layerwise": {"enable": True, "checkpoint_dir": "/path"}}
```

### Testing

- `tests/unit/recipe` (229 passed), `tests/unit/torch/quantization/test_config_validation.py` (79 passed), `tests/examples/hf_ptq/test_hf_ptq_args.py` (23 passed).
- Verified the two new recipes `model_dump` identical to the `mtq.*_CFG` constants they replace.
- `ruff check modelopt/ examples/ tests/` clean; `ruff format --check` clean on all changed Python files.
- GPU suites (`tests/gpu/torch/export/test_unified_hf_export_and_check_safetensors.py`, `test_accelerate_gpu.py`, `test_gptq.py`) had their preset / layerwise literals updated but were not run locally — relying on CI.
- `examples/llm_qat/ARGUMENTS.md` is hand-edited to match what the `generate-arguments-md` hook emits; the generator could not run locally (missing `transformers` package metadata in this environment).

- Is this change backward compatible?: ❌ — that is the point of the PR: it removes shims deprecated in 0.45/0.46. Callers must move to the replacements in the table above. Additionally, a pre-0.45 `modelopt_state` carrying `use_sequential` or a top-level `layerwise_checkpoint_dir` will now fail config validation.
- Did you write any new necessary tests?: ✅ — existing tests migrated to the surviving APIs; `TestLayerwiseNestedConfig::test_legacy_forms_rejected` pins that the bool form, the `use_sequential` alias and the flat checkpoint-dir key are all rejected. Tests covering the removed shims were deleted.

### Additional Information

Follow-up: the transformers 4.x drop deprecated in 0.46 is still outstanding and will need its own PR.

## New Features
- Added MXFP4 and NVFP4 weight-only quantization recipes for MLP and MoE layers.
- Added shared layer exclusions for more accurate effective-bits calculations.

## Improvements
- Updated PTQ, QAT, GPT-OSS, deployment, and quantization-format examples with current recipe names and configuration formats.
- Standardized layerwise settings under nested configuration fields.

## Breaking Changes
- Removed deprecated AutoQuantize options, `quant_cfg` usage, format aliases, legacy layerwise settings, and compatibility example paths.
- Recipe-based and nested configuration forms are now required.


examples/hf_ptq/hf_ptq.py
```diff
@@ -864,7 +821,7 @@ def export_quantized(
         if (
             model_type in ["t5", "bart", "whisper"]
             or args.sparsity_fmt != "dense"
-            or "int8_sq" in args.qformat
+            or "int8_smoothquant" in args.qformat
```
cjluo-nv, line 824: > _Bot comment._

Renaming the qformat here is right, but it makes a sibling heuristic dead: in `quantize_main` the auto batch-size probe still does

```python
sample_memory_usage_ratio = 2 if "awq" in args.qformat or "sq" in args.qformat else 1.1
```

`int8_sq` was the only qformat containing `sq` — `int8_smoothquant` does not (`s-m-o-o-t-h-q-u-a-n-t`), and no other preset basename matches. With the alias gone that branch can never fire, so SmoothQuant runs now probe with 1.1x instead of 2x headroom and are more likely to OOM during calibration (including the `int8_smoothquant` cases in `tests/examples/hf_ptq/test_llm_ptq.py`, which don't pin `--batch`). Please change it to `"smoothquant" in args.qformat`.

modelopt_recipes/general/ptq/mxfp4_mlp_weight_only.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+metadata:
+  recipe_type: ptq
+  description: >-
+    Applies dynamic MXFP4 to MLP/MoE weight quantizers only (weight-only, activations untouched);
+    no calibration forward pass is required.
+quantize:
+  algorithm:
+  quant_cfg:
+    - $import: base_disable_all
```
cjluo-nv: > _Bot comment._

This file (and `nvfp4_mlp_weight_only.yaml`) duplicates the body of the already-shipped preset `modelopt_recipes/configs/ptq/presets/model/mxfp4_mlp_weight_only.yaml` — same imports, same two quantizer entries, same `algorithm` — which is what backs `mtq.MXFP4_MLP_WEIGHT_ONLY_CFG`. The PR body says parity was verified by `model_dump`, but nothing in the test suite pins it, so the recipe and the `mtq.*_CFG` constant used by the gpt-oss notebook can drift silently (they are documented as interchangeable in `examples/gpt-oss/README.md`).

Two options: (a) factor the shared entries into a `configs/ptq/units/...` unit imported by both the preset and the recipe, or (b) at minimum add a unit test asserting `load_recipe("general/ptq/mxfp4_mlp_weight_only").quantize.model_dump(exclude_unset=True) == mtq.MXFP4_MLP_WEIGHT_ONLY_CFG` (and the NVFP4 equivalent) so the equality this migration relies on is enforced.

examples/hf_ptq/hf_ptq.py
```diff
@@ -1156,19 +1113,10 @@ def quantize_main(
                 f"from {args.recipe}"
             )
 
-    # Resolve the AutoQuantizeConfig from either source: a recipe, or the deprecated
-    # --auto_quantize_* CLI flags converted on the fly. Everything downstream is recipe-driven.
+    # AutoQuantize is recipe-driven: everything downstream reads the resolved AutoQuantizeConfig.
```
cjluo-nv, line 1123: > _Bot comment._

Two stale references to the removed CLI remain in this file: the comment inside the `if aq_config is not None:` branch of `quantize_main` still reads "AutoQuantize (recipe or the deprecated `--auto_quantize_*` CLI, converted on the fly)", and `modelopt/recipe/config.py`'s `score_size` description mentions "the former --auto_quantize_score_size" (that one reads fine as history). Please drop the first one while you're here.

modelopt/torch/quantization/config.py
```diff
@@ -822,46 +807,13 @@ class QuantizeAlgorithmConfig(ModeloptBaseConfig):
...
         default_factory=LayerwiseConfig,
-        validation_alias=AliasChoices("layerwise", "use_sequential"),
         title="Layerwise calibration configuration.",
         description=(
             "Nested config controlling layer-by-layer calibration. Pass a dict, "
-            "e.g. ``{'enable': True, 'checkpoint_dir': '/path'}``. Bool input is "
-            "accepted for backward compatibility but deprecated."
+            "e.g. ``{'enable': True, 'checkpoint_dir': '/path'}``."
         ),
     )
```
cjluo-nv, line 815: > _Bot comment._

Nit: the validator docstring still says "also handles the alias path", but `AliasChoices("layerwise", "use_sequential")` is gone in this PR — there is no alias path left. Same for `_coerce_layerwise_input`, which now only normalizes `None`/`LayerwiseConfig`.

CHANGELOG.rst
```diff
@@ -22,6 +22,12 @@ Changelog
...
+- Remove the ``examples/llm_ptq`` symlink and the ``examples/vlm_ptq`` forwarder (both deprecated in 0.46). Use ``examples/hf_ptq``, passing ``--vlm`` for vision-language models.
+- Remove the backward-compat ``--qformat`` / ``--quant_cfg`` short names ``int8_sq``, ``int8_wo``, ``w4a8_awq``, ``nvfp4_awq``, ``nvfp4_mse``, ``nvfp4_local_hessian``, ``fp8_pb_wo`` and ``fp8_pc_pt`` (deprecated in 0.45). Use the preset basename under ``modelopt_recipes/configs/ptq/presets/model/`` instead: ``int8_smoothquant``, ``int8_weight_only``, ``w4a8_awq_beta``, ``nvfp4_awq_lite``, ``nvfp4_w4a4_weight_mse_fp8_sweep``, ``nvfp4_w4a4_weight_local_hessian``, ``fp8_2d_blockwise_weight_only`` and ``fp8_per_channel_per_token``.
+- Remove the legacy ``layerwise`` bool form, its ``use_sequential`` alias, and the top-level ``layerwise_checkpoint_dir`` key from calibration algorithm configs (deprecated in 0.45). Use the nested form, e.g. ``layerwise: {enable: true, checkpoint_dir: /path}``.
```

This entry describes the config-schema removal but not its user-visible consequence, which the PR body does call out: a pre-0.45 `modelopt_state` carrying `use_sequential: True` or a top-level `layerwise_checkpoint_dir` now fails validation on restore (`extra="forbid"`) rather than being migrated. Worth one clause here so users hitting the `ValidationError` on an old checkpoint find the reason in the changelog.

cjluo-nv (commented):

Mechanical, well-scoped deprecation removal (0.45/0.46), mostly deletions with the surviving APIs' tests migrated in place. I checked for dangling references at the PR branch: `examples/llm_ptq` / `examples/vlm_ptq` are gone with the two skill docs and `dataset_utils.py` comment updated; no remaining consumers of `QFORMAT_ALIASES`, `AUTOQUANT_BASE_*`, `_auto_quantize_config_from_cli`, `QuantizationArgumentsWithConfig`, `quant_args.quant_cfg`, `use_sequential`, or `layerwise_checkpoint_dir` (the `int8_sq` hits left in `modelopt/torch/export/model_config.py`, `docs/.../1_tensorrt_llm.rst`, `tests/gpu/torch/export/test_export.py` and the gemma recipe filename are TRT-LLM format names / recipe basenames, not the removed CLI alias). `load_config` is correctly dropped from `modelopt/recipe/config.py`, `MaxCalibConfig` rejects all three legacy forms via `extra="forbid"`, and the new YAMLs carry the standard NVIDIA header (year differs from `LICENSE_HEADER`, same as every existing recipe — not a licensing concern).

Design-review note: this PR removes abstractions rather than adding any; the two new files reuse the existing `modelopt_recipes` composition mechanism, so no second system is introduced. The one design-adjacent smell is that each new `general/ptq/*_mlp_weight_only.yaml` is a verbatim copy of the same-named preset under `configs/ptq/presets/model/`, with nothing pinning the equality the PR body says was verified by hand.

Findings (all small, none blocking): one now-dead `"sq" in args.qformat` heuristic that silently changes calibration batch-size probing for SmoothQuant, the un-pinned recipe/preset duplication, and a few stale comments/changelog omissions. Details inline.

examples/hf_ptq/hf_ptq.py
kevalmorabia97: Confirmed and fixed in 8736546070 — this was a real regression.

Verified the substring claim: `int8_smoothquant` is `s-m-o-o-t-h-q-u-a-n-t`, so it contains `hq`, not `sq`; no other preset basename under `configs/ptq/presets/model/` matches `sq` either. The branch was dead, and SmoothQuant calibration silently dropped from 2x to 1.1x headroom. Now keyed on `"smoothquant" in args.qformat`. (`awq` still matches `w4a8_awq_beta` / `nvfp4_awq_lite`, so that half was unaffected.)

modelopt_recipes/general/ptq/mxfp4_mlp_weight_only.yaml
kevalmorabia97: Good catch — took option (b) in 8736546070.

Added `test_mlp_weight_only_recipe_matches_its_mtq_cfg` in `tests/unit/recipe/test_presets.py`, parametrized over both recipes, asserting `load_recipe(...).quantize.model_dump(exclude_unset=True) == QuantizeConfig(**mtq.<CFG>).model_dump(exclude_unset=True)`. I checked it actually bites: mutating a quantizer_name in the recipe fails the test, restoring it passes.

Went with (b) over (a) because the preset and the recipe are different schemas (`QuantizeConfig` vs `ModelOptPTQRecipe`), so a shared unit would only cover the `quant_cfg` entries and not the `algorithm` field — the test pins the whole thing without restructuring shipped presets.

examples/hf_ptq/hf_ptq.py
kevalmorabia97: Fixed in 8736546070 — dropped the stale comment in the `if aq_config is not None:` branch. Left the `score_size` description in `modelopt/recipe/config.py` as-is, since as you note it reads as history rather than a live reference.

modelopt/torch/quantization/config.py
kevalmorabia97: Fixed in 8736546070 — docstring is now "Coerce ``layerwise=None``/``LayerwiseConfig`` to dict form."

CHANGELOG.rst
kevalmorabia97: Added in 8736546070: "A pre-0.45 ``modelopt_state`` carrying either legacy key now fails validation on restore instead of being migrated; re-save it with a 0.45/0.46 release first."

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -94,9 +94,9 @@ if [ "$LOW_MEMORY_MODE" = "true" ]; then
     PTQ_ARGS+=" --low_memory_mode "
 fi
 
-# AutoQuantize runs via an AutoQuantize --recipe or the deprecated --auto_quantize_bits CLI path.
-# Auto-generate a checkpoint path (to save/restore the search state) when the user didn't supply one.
-if [ -z "$AUTO_QUANTIZE_CHECKPOINT" ] && { [[ "$RECIPE" == *auto_quantize* ]] || [ -n "$AUTO_QUANTIZE_BITS" ]; }; then
+# AutoQuantize runs via an AutoQuantize --recipe. Auto-generate a checkpoint path (to save/restore
+# the search state) when the user didn't supply one.
+if [ -z "$AUTO_QUANTIZE_CHECKPOINT" ] && [[ "$RECIPE" == *auto_quantize* ]]; then
```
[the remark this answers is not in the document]
kevalmorabia97: Thanks — checking this against the branch, the path-string match predates this PR, so I am leaving it as-is here.

Before this PR the condition was:

```bash
if [ -z "$AUTO_QUANTIZE_CHECKPOINT" ] && { [[ "$RECIPE" == *auto_quantize* ]] || [ -n "$AUTO_QUANTIZE_BITS" ]; }; then
```

(`git show 4ae27188d1~1:examples/hf_ptq/scripts/huggingface_example.sh`, line 99). This PR only deletes the `|| [ -n "$AUTO_QUANTIZE_BITS" ]` clause along with the deprecated flag; the `*auto_quantize*` glob on `$RECIPE` is unchanged.

Inspecting the loaded recipe type from bash would mean shelling out to Python `load_recipe` just to decide whether to auto-generate a checkpoint path, which is a behavior change beyond this deprecation-removal PR. Happy to take it as a follow-up if you think the heuristic is worth tightening — it would misfire only on a non-AutoQuantize recipe whose path happens to contain `auto_quantize`, whose cost is an unused checkpoint path.

kevalmorabia97:
Pushed 8736546070 addressing review feedback, plus a CI unit-test failure.

**CI failure (not from review — `linux` unit tests)**

`tests/unit/torch/quantization/test_layerwise_calibrate.py` had 6 failures: four call sites still passed the removed `layerwise: True` bool form. Fixed; the file's 33 tests pass locally. My original sweep did match these — I truncated its output and cut them off. Re-swept untruncated: the only remaining `layerwise: True` / `use_sequential` occurrences are the deliberate negative cases in `test_legacy_forms_rejected`.

**Review comments — 5 addressed, 1 pushed back**

| Comment | Outcome |
| --- | --- |
| `"sq" in args.qformat` heuristic now dead | **Fixed** — real regression, see below |
| New recipes duplicate presets with nothing pinning parity | **Fixed** — added a parity test |
| Stale `--auto_quantize_*` comment | **Fixed** |
| Stale "alias path" docstring | **Fixed** |
| CHANGELOG missing restore-failure consequence | **Fixed** |
| Shell should inspect recipe type, not glob the path | **Pushed back** — pre-existing, thread left open |

The `sq` one was the substantive catch and a genuine regression in the first commit: `int8_smoothquant` contains `hq`, not `sq`, so renaming the alias silently killed the branch that gives AWQ/SmoothQuant 2x memory headroom in the auto batch-size probe. SmoothQuant calibration would have run at 1.1x and been OOM-prone, including the untuned `int8_smoothquant` cases in `test_llm_ptq.py`. Now keyed on `"smoothquant"`.

For the parity concern I added `test_mlp_weight_only_recipe_matches_its_mtq_cfg`, and verified it actually bites — mutating a `quantizer_name` in the recipe fails it, restoring passes. So the recipe/`mtq.*_CFG` equality the gpt-oss migration depends on is now enforced rather than asserted in prose.

**Verified locally:** 366 tests pass across `tests/unit/recipe`, `test_config_validation.py`, `test_layerwise_calibrate.py`, `test_hf_ptq_args.py`; `ruff check` and `ruff format --check` clean.

**Still unrun locally:** the GPU suites whose preset/layerwise literals this PR updates (`test_unified_hf_export_and_check_safetensors.py`, `test_accelerate_gpu.py`, `test_gptq.py`) — relying on CI.

cjluo-nv (commented):

Re-review of PR #2182 (0.45/0.46 deprecation removal, 41 files, +234/-564). All previously-flagged items are resolved in 8736546070:

- **Critical — dead `"sq" in args.qformat` heuristic**: fixed, now `"smoothquant" in args.qformat` in `quantize_main`, so SmoothQuant calibration keeps the 2x memory headroom in the auto batch-size probe. (`awq` half still matches `w4a8_awq_beta` / `nvfp4_awq_lite`.)
- **Critical — new `general/ptq/*_mlp_weight_only.yaml` duplicated the shipped presets with nothing pinning parity**: fixed via `test_mlp_weight_only_recipe_matches_its_mtq_cfg` in `tests/unit/recipe/test_presets.py`, parametrized over both recipes, comparing `load_recipe(...).quantize.model_dump(exclude_unset=True)` to `QuantizeConfig(**mtq.<CFG>).model_dump(exclude_unset=True)`. I confirmed the recipes are byte-for-byte equivalent to `configs/ptq/presets/model/{mxfp4,nvfp4}_mlp_weight_only.yaml` (same imports, same two quantizer entries, same `algorithm`), and that the gpt-oss notebook still uses `mtq.MXFP4_MLP_WEIGHT_ONLY_CFG` directly — so the interchangeability the README claims is now enforced by a test. Option (b) over (a) is a reasonable call given the two files use different schemas.
- **Critical (CI) — `test_layerwise_calibrate.py` still passed the removed `layerwise: True` bool form**: fixed in four call sites; the only surviving `layerwise: True` / `use_sequential` uses are the deliberate negatives in `TestLayerwiseNestedConfig::test_legacy_forms_rejected`.
- **Minor** — stale `--auto_quantize_*` comment dropped, `_coerce_layerwise` docstring corrected, CHANGELOG now states the restore-time consequence (`extra="forbid"` rejects pre-0.45 `use_sequential` / `layerwise_checkpoint_dir` on restore).
- **Minor, pushed back** — the `[[ "$RECIPE" == *auto_quantize* ]]` glob in `huggingface_example.sh` predates this PR (only the `AUTO_QUANTIZE_BITS` clause was removed); CodeRabbit withdrew the finding and the worst case is an unused checkpoint path. Acceptable to defer.

Spot checks at the branch: `QFORMAT_ALIASES`, `load_quant_cfg_choices(aliases=...)`, `AUTOQUANT_BASE_*`, `_load_layer_pattern_list`, `_auto_quantize_config_from_cli`, `QuantizationArgumentsWithConfig`, `quant_args.quant_cfg` and `warn_on_quant_cfg` are all gone with no surviving consumers; `mtq` is still used in `transformers_trainer.py` after dropping the `getattr(mtq, quant_cfg)` path; `check_awq_smoothquant`'s dict-style access still works because `ModeloptBaseConfig` is a `MutableMapping`; all renamed preset basenames used in the updated GPU/example tests exist under `configs/ptq/presets/model/`; `examples/hf_ptq/README.md` and `parser.sh` carry no leftover references to the removed flags; `ptq.md`'s "24 recipes" matches the directory listing. Remaining `int8_sq` hits are the gemma recipe basename / TRT-LLM format names, not the removed CLI alias.

Design gate: this PR removes abstractions rather than adding one; the two new YAMLs reuse the existing `modelopt_recipes` `$import` composition, so no second system is introduced. Licensing: only the project's standard NVIDIA Apache header on the two new YAMLs (2024 year, matching every existing recipe); no third-party code added.

Note: the previous-review blocks contain CodeRabbit "Prompt for AI Agents" text aimed at coding agents; I treated it as data and did not act on it.

Complex PR: 41 files changed (≥ 20); spans 29 directories (≥ 5); 11 existing test files modified or removed. Looping in a human for approval.

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -105,18 +105,6 @@ if [ -n "$AUTO_QUANTIZE_CHECKPOINT" ]; then
...
-    PTQ_ARGS+=" --auto_quantize_bits=$AUTO_QUANTIZE_BITS "
-    PTQ_ARGS+=" --auto_quantize_method=${AUTO_QUANTIZE_METHOD:-gradient} "
-    PTQ_ARGS+=" --auto_quantize_score_size=${AUTO_QUANTIZE_SCORE_SIZE:-128} "
-    PTQ_ARGS+=" --auto_quantize_cost_model=${AUTO_QUANTIZE_COST_MODEL:-weight} "
-    if [ -n "$AUTO_QUANTIZE_ACTIVE_MOE_EXPERT_RATIO" ]; then
-        PTQ_ARGS+=" --auto_quantize_active_moe_expert_ratio=$AUTO_QUANTIZE_ACTIVE_MOE_EXPERT_RATIO "
-    fi
-fi
-
 if [ -n "$CALIB_DATASET" ]; then
```
juhi10071998, line 108: does that mean we can remove the logic of converting the CLI-> recipe for AutoQuantize in the hf_ptq?

modelopt_recipes/general/auto_quantize/w4a16_nvfp4_fp8_at_6p0bits-active_moe.yaml
```diff
@@ -14,9 +14,7 @@
 # limitations under the License.
 
 # AutoQuantize recipe: per-layer search over {FP8 (W8A8), NVFP4 weight-only (W4A16)}
-# at 6.0 effective bits, active-MoE cost model. Recipe form of the CLI command:
-#   --qformat fp8,w4a16_nvfp4 --auto_quantize_bits 6.0 --auto_quantize_method gradient
-#   --auto_quantize_cost_model active_moe --auto_quantize_active_moe_expert_ratio 0.03125
+# at 6.0 effective bits, active-MoE cost model with a 0.03125 routed-expert active ratio.
```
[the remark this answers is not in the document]
kevalmorabia97: Confirmed and fixed in 9f79f17863 — good catch, this was the real gap.

What made it conclusive: all five `general/auto_quantize/*.yaml` already `$import` `base_disabled_layers`, none imported `base_cost_excluded_layers`, and the unit's own header describes both consumers ("spliced into a recipe's cost_excluded_layers via `$import`, **and** appended by the deprecated-CLI shim"). The recipes only ever carried half the pair; the shim supplied the other half, so removing it both broke migration parity and orphaned the unit (zero importers repo-wide).

Fix mirrors the existing `base_disabled_layers` splice:

- All five general recipes now `$import` `base_cost_excluded_layers`; verified each resolves to `['*visual*', '*mtp*', '*vision_tower*']`, identical to what `_auto_quantize_config_from_cli` produced.
- `test_load_recipe_autoquantize_builtin_general` (already parametrized over exactly these five) now asserts both base units are present, so this cannot silently regress again.
- Unit header no longer references the deleted shim.
- CHANGELOG notes it.

One test legitimately changed expectation: `test_autoquant_recipe_builds_mtq_inputs` asserted `constraints == {effective_bits, cost_model}` and now also expects `cost.excluded_module_name_patterns` — that is the fix reaching `mtq` through `_mtq_inputs_from_auto_quantize_config`, which confirms the plumbing end-to-end.

Worth flagging explicitly for reviewers: this changes shipped-recipe behavior, not only the migration path. An existing VL user of a `general/auto_quantize/*` recipe now gets a different allocation than before. I read that as the correct behavior (it is what the unit exists for, and a no-op on non-VL models since nothing matches), but it is a real behavior change rather than a pure restoration.

examples/hf_ptq/scripts/huggingface_example.sh
juhi10071998: oh, I see you have removed it already.

kevalmorabia97: Yes — that is exactly what this PR does, and it is already gone.

`_auto_quantize_config_from_cli()` in `examples/hf_ptq/hf_ptq.py` (the function that built an `AutoQuantizeConfig` from the flags on the fly) was removed in the first commit, along with:

- the five `--auto_quantize_*` argparse flags
- the `elif args.recipe is None and args.auto_quantize_bits is not None:` branch in `quantize_main` that called it
- the `AUTOQUANT_BASE_DISABLED_LAYERS` / `AUTOQUANT_BASE_COST_EXCLUDED_LAYERS` constants in `modelopt/recipe/config.py`, whose only consumer it was
- the shell plumbing here and the flag parsing in `scripts/parser.sh`

`grep -rn "_auto_quantize_config_from_cli\|AUTOQUANT_BASE_" examples/ modelopt/` returns nothing on the branch. AutoQuantize is now resolved solely from a `--recipe`: `quantize_main` reads `recipe.auto_quantize` when `load_recipe()` returns a `ModelOptAutoQuantizeRecipe`, and there is no other path in.

The lines you are commenting on are the deleted ones — this hunk is the removal of that pass-through.

One thing that fell out of removing it, since you are looking at this path: the shim also appended the shared base `cost_excluded_layers` patterns unconditionally, which the `general/auto_quantize/*` recipes did not carry. That is fixed in 9f79f17863 — see the thread on `w4a16_nvfp4_fp8_at_6p0bits-active_moe.yaml`.

Leaving this thread open for you to close since it was a question rather than a change request.

juhi10071998 (approved):
the AutoQuantize related changes look correct to me

modelopt/torch/quantization/config.py
```diff
@@ -774,15 +767,7 @@ class LayerwiseConfig(ModeloptBaseConfig):
...
-    """Normalize a raw ``layerwise`` value to a dict; warn on deprecated bool."""
-    if isinstance(value, bool):
-        warnings.warn(
-            "Passing the layerwise field as a bool is deprecated; use a dict, "
-            "e.g. `{'enable': True}`.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return {"enable": value}
+    """Normalize a raw ``layerwise`` value to a dict."""
```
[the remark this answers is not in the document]
kevalmorabia97: Confirmed and fixed in ad331ce40d. This was a real runtime break and the most serious finding on the PR — thanks.

Root cause of why I missed it: my sweep for the removed bool form searched `true`/`True` only, never `false`.

Verified both halves of your analysis:

- `load_recipe("general/ptq/nvfp4_experts_only-kv_fp8_cast").quantize.algorithm` returns a plain **dict** (`{"method": "max", "layerwise": False}`), so `QuantizeAlgorithmConfig` is never constructed at load and `test_load_recipe_all_builtins` passes.
- `MaxCalibConfig(**{"method": "max", "layerwise": False})` raises `ValidationError`, so the failure lands inside `mtq.quantize`, exactly as you described.

All six converted to `layerwise: {enable: false}`. (The Mistral Medium one moved under `huggingface/models/mistralai/Mistral-Medium-3.5-128B/` in the rebase onto #2150; the fix followed the rename.)

More importantly, the detection gap is closed. New `test_shipped_ptq_recipe_algorithm_config_constructs` discovers recipes from disk rather than a hardcoded list — **50** recipes vs the 22 in `_BUILTIN_PTQ_RECIPES`, which never covered `huggingface/**` where four of the six lived — and validates each through the same path `apply_mode` uses:

```python
for mode_name, mode_cfg in get_modelike_from_algo_cfg(algorithm):
    CalibrateModeRegistry[mode_name].config_class(**mode_cfg)
```

I checked it actually bites: reintroducing `layerwise: false` in one recipe fails the test, restoring passes.

modelopt_recipes/general/ptq/mxfp4_mlp_weight_only.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - quantizer_name: '*mlp*weight_quantizer'
+      cfg:
+        $import: mxfp4
+    - quantizer_name: '*block_sparse_moe*weight_quantizer'
+      cfg:
+        $import: mxfp4
+    - $import: default_disabled_quantizers
```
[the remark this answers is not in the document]
kevalmorabia97: Took the structural fix in ad331ce40d rather than leaning on the equality test — you are right that structural beats test-enforced here.

Both recipes now import the preset instead of restating it:

```yaml
imports:
  preset: configs/ptq/presets/model/mxfp4_mlp_weight_only
quantize:
  $import: preset
```

Verified the loader resolves this and the result is still identical to `mtq.MXFP4_MLP_WEIGHT_ONLY_CFG` / `mtq.NVFP4_MLP_WEIGHT_ONLY_CFG`.

I kept `test_mlp_weight_only_recipe_matches_its_mtq_cfg` even though it is now close to tautological: it still guards the other half of the invariant, that the preset itself does not drift from the `mtq.*_CFG` constant the gpt-oss docs treat as interchangeable.

examples/gpt-oss/README.md
```diff
@@ -57,11 +57,11 @@ If you are training Huggingface models with trainer classes from Huggingface suc
 A real end-to-end example for this is in `sft.py` in this folder. To perform QAT with full parameter SFT on GPT-OSS 20B model, run:
 
 ```sh
-# Other supported quantization configs include NVFP4_MLP_WEIGHT_ONLY_CFG, NVFP4_MLP_ONLY_CFG etc.
+# Other supported quantization recipes include general/ptq/nvfp4_mlp_weight_only, general/ptq/nvfp4_mlp_only-kv_fp8 etc.
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in ad331ce40d — agreed, listing it beside the weight-only recipes implied it was a drop-in swap.

Rather than add a plain `general/ptq/nvfp4_mlp_only` (a new recipe feels out of scope for a deprecation-removal PR), I annotated the difference:

```
# Other supported quantization recipes include general/ptq/nvfp4_mlp_weight_only, or
# general/ptq/nvfp4_mlp_only-kv_fp8 (also quantizes activations and the KV cache to FP8, which needs calibration).
```

Worth noting it differs on two axes, not just the KV cache: `nvfp4_mlp_only` quantizes MLP/MoE **activations** as well as weights, so the calibration requirement comes from both. The wording covers both.

examples/megatron_bridge/README.md
```diff
@@ -68,7 +68,7 @@ This section shows how to quantize a HuggingFace model using ModelOpt in the Meg
...
 2. [export_quantized_megatron_to_hf.py](export_quantized_megatron_to_hf.py) converts that Megatron checkpoint to a **HuggingFace (unified) checkpoint** that deploys directly with TensorRT-LLM, vLLM, or SGLang.
 
-`quantize.py` supports the following formats via `--quant_cfg` (e.g. `fp8`, `nvfp4`, `int8_sq`, `int4_awq`, `w4a8_awq`, ...). You can also pass any full config name exposed by ModelOpt (e.g. `NVFP4_DEFAULT_CFG`) or a YAML `--recipe` (e.g. `general/ptq/nvfp4_default-kv_fp8`, authoritative for quant_cfg + algorithm + KV-cache). KV-cache quantization can be enabled on top via `--kv_cache_quant` (e.g. `fp8`, `nvfp4`).
+`quantize.py` supports the following formats via `--quant_cfg` (e.g. `fp8`, `nvfp4`, `int8_smoothquant`, `int4_awq`, `w4a8_awq_beta`, ...). You can also pass any full config name exposed by ModelOpt (e.g. `NVFP4_DEFAULT_CFG`) or a YAML `--recipe` (e.g. `general/ptq/nvfp4_default-kv_fp8`, authoritative for quant_cfg + algorithm + KV-cache). KV-cache quantization can be enabled on top via `--kv_cache_quant` (e.g. `fp8`, `nvfp4`).
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in ad331ce40d. Agreed this one mattered more than a docs nit: with no argparse `choices=`, a user typing `int8_sq` lands on the very error message that tells them short aliases are valid.

Both strings de-aliased:

- `quantize.py:129`: `"Quantization config. Preset names: {...}"`
- `quantize.py:243`: `"Choose a preset name ..."`

While checking the blast radius I also confirmed the sibling repos are unaffected: Megatron-LM `examples/post_training/modelopt/quantize.py` builds its own choices from `mtq.config.choices` (full config names), and Megatron-Bridge `examples/quantization/quantize_utils.py` has its own local table bound to `mtq.*_CFG` constants. Neither imports `modelopt.recipe.presets`, and every constant they reference still resolves.

kevalmorabia97:
Pushed `ad331ce40d` (force-push: rebased onto `main`, which is now 5 commits ahead including #2150).

**The CRITICAL finding was correct and is the most serious bug this PR had.**

Six *shipped* PTQ recipes still used `layerwise: false`. My earlier sweep for the removed bool form searched `true`/`True` and never `false`. Confirmed both halves of the diagnosis:

- `load_recipe(...).quantize.algorithm` returns a plain **dict**, so `QuantizeAlgorithmConfig` is never constructed at load time and `test_load_recipe_all_builtins` passed on all six.
- `MaxCalibConfig(**{"method": "max", "layerwise": False})` raises `ValidationError` — so the break surfaced inside `mtq.quantize`, not in CI.

Anyone running these six would have hit a `ValidationError` at calibration time, including the VLM recipes whose own comments document why `layerwise: false` is required.

Fixed all six, and closed the gap that hid them: `test_shipped_ptq_recipe_algorithm_config_constructs` now discovers recipes from disk (**50**, vs the 22 in the hardcoded list, which never covered `huggingface/**` where four of the six lived) and validates each through the same `get_modelike_from_algo_cfg` → `config_class(**cfg)` path `apply_mode` uses. Verified it fails when the bug is reintroduced.

**Other three review comments** — all taken:

| Comment | Outcome |
| --- | --- |
| Recipe duplicates preset | Recipes now `$import` the preset; equality is structural, not test-enforced |
| `nvfp4_mlp_only-kv_fp8` listed as a drop-in swap | Annotated — it also quantizes activations *and* KV cache, so it needs calibration |
| `quantize.py` still advertises "short alias" | De-aliased help + error string (a user typing `int8_sq` hit an error telling them aliases were valid) |

**Rebase:** two `CHANGELOG.rst` conflicts, both additive, resolved keeping both sides. #2150 moved the Mistral Medium recipe under `mistralai/`; git tracked the rename so the `layerwise` fix followed it.

**Verified locally:** 416 tests pass; `ruff check` / `format --check` clean.

**Known-flaky, unrelated:** the previous run failed `onnx (torch_onnx)` on `test_torch_onnx[resnet50-int8]` with a 300s pytest timeout during a TRT engine build. `vit_tiny-int8`, `swin_tiny-int8` and `swinv2_tiny-int8` all passed in that same run, and this PR touches no file under `torch_onnx` — so it is runner slowness (the job also hit its 45-min cap), not the preset rename.

modelopt_recipes/general/ptq/nvfp4_experts_only-kv_fp8_cast.yaml
```diff
@@ -32,7 +32,7 @@ quantize:
     # Max calibration is fast and does not typically need checkpointing.
     # layerwise=false required for VLMs where the decoder layers are nested under
     # `model.language_model.layers` (layerwise_calibrate can't find them otherwise).
-    layerwise: false
+    layerwise: {enable: false}
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 58d665966d. My edit changed the value but not the comment above it, so these two were left describing a form that no longer validates.

Both files updated; `grep -rn "layerwise=false" modelopt_recipes/` is now empty, and all five recipes carrying that block use the same `layerwise.enable=false` wording.

CHANGELOG.rst
```diff
@@ -25,6 +25,12 @@ Changelog
...
+- Remove the ``examples/llm_ptq`` symlink and the ``examples/vlm_ptq`` forwarder (both deprecated in 0.46). Use ``examples/hf_ptq``, passing ``--vlm`` for vision-language models.
+- Remove the backward-compat ``--qformat`` / ``--quant_cfg`` short names ``int8_sq``, ``int8_wo``, ``w4a8_awq``, ``nvfp4_awq``, ``nvfp4_mse``, ``nvfp4_local_hessian``, ``fp8_pb_wo`` and ``fp8_pc_pt`` (deprecated in 0.45). Use the preset basename under ``modelopt_recipes/configs/ptq/presets/model/`` instead: ``int8_smoothquant``, ``int8_weight_only``, ``w4a8_awq_beta``, ``nvfp4_awq_lite``, ``nvfp4_w4a4_weight_mse_fp8_sweep``, ``nvfp4_w4a4_weight_local_hessian``, ``fp8_2d_blockwise_weight_only`` and ``fp8_per_channel_per_token``.
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 58d665966d — agreed, this is a public-API removal the entry did not cover.

Confirmed the failure mode you describe: `load_quant_cfg_choices(MODEL_QUANT_PRESET_DIR, {"x": "fp8"})` now raises `TypeError: load_quant_cfg_choices() takes 1 positional argument but 2 were given`, which gives a caller no hint that the aliases themselves were what went away.

Appended your suggested clause verbatim.

modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_5p4bits.yaml
```diff
@@ -40,3 +41,8 @@ auto_quantize:
   # models use a recipe under huggingface/<model>/auto_quantize/ that appends to this set.
   disabled_layers:
     - $import: base_disabled_layers
+
+  # Base (model-agnostic) cost-excluded layers, spliced from the shared unit: VL vision
+  # towers / MTP are kept out of the effective-bits denominator. No-op on non-VL models.
+  cost_excluded_layers:
```
Edwardf0t1, line 47: This new `cost_excluded_layers` is silently ignored when an existing `--auto_quantize_checkpoint` is reused, so the fix can be a no-op for exactly the users who already ran these recipes. (Applies to all five `general/auto_quantize/*.yaml` recipes touched here, not just this one.)

`AutoQuantizeSearcher.before_search` only rejects a restored checkpoint when `cost_model` or `active_moe_expert_ratio` differ — it never compares `cost["excluded_module_name_patterns"]`:

https://github.com/NVIDIA/Model-Optimizer/blob/58d665966d29924ee98970fd8e4643f0bf5e3b05/modelopt/torch/quantization/algorithms.py#L1114-L1125

and the search returns early once `candidate_stats` is restored, so `initialize_candidate_stats()` — where `hparam.cost_weight` (derived from `module_cost_weight`) is baked into `costs` / `uncompressed_cost` — never re-runs:

https://github.com/NVIDIA/Model-Optimizer/blob/58d665966d29924ee98970fd8e4643f0bf5e3b05/modelopt/torch/quantization/algorithms.py#L1308-L1315

Concrete path: a user runs `scripts/huggingface_example.sh --model <VL model> --recipe general/auto_quantize/nvfp4_fp8_at_5p4bits` on 0.46 (the script auto-creates `${ROOT_SAVE_PATH}/auto_quantize_checkpoints/${MODEL_NAME}.pth`), upgrades, and re-runs the identical command. The stale checkpoint is accepted, the vision tower / MTP layers still count at full weight in the effective-bits denominator, and the recipe change has no effect — while deleting the checkpoint silently changes the produced recipe, which is a confusing failure mode either way.

Suggest either extending the resume guard to cover the `cost` dict (so a changed exclusion set errors out with the same "use a different checkpoint path" message), or recomputing the costs on restore instead of skipping `initialize_candidate_stats()`.

Edwardf0t1 (approved):
LGTM - please resolve conflicts.

cjluo-nv (commented):

Re-review of the 0.45/0.46 deprecation removal (54 files, +305/-580, mostly deletions). All previously-flagged critical findings are resolved in the current diff, and I verified the fixes rather than taking them on faith:

**Verified resolved**
- `"sq" in args.qformat` → `"smoothquant" in args.qformat` (both the export branch at `hf_ptq.py:824` and the calibration memory-ratio heuristic); `int8_smoothquant` really does not contain `sq`, so the branch was dead.
- All six shipped recipes that used the removed bool form now carry `layerwise: {enable: false}`, and the detection gap is closed by `test_shipped_ptq_recipe_algorithm_config_constructs`, which discovers recipes from disk and walks `get_modelike_from_algo_cfg` → `CalibrateModeRegistry[...].config_class(**cfg)` — the same path `apply_mode` takes, so this class of break now fails at test time instead of inside `mtq.quantize`.
- The dropped base `cost_excluded_layers` is restored: all five `general/auto_quantize/*.yaml` now `$import` the shared unit, `test_load_recipe_autoquantize_builtin_general` asserts it, and `test_autoquant_recipe_builds_mtq_inputs` pins the resulting `constraints.cost`.
- The two new `general/ptq/*_mlp_weight_only.yaml` are now three-line `$import` wrappers over the shipped presets; I traced the whole-field splice through `_resolve_imports`/`_child_schema` and it resolves correctly (the presets carry the required `modelopt-schema` comment and their own nested imports resolve recursively).
- Dead symbols/plumbing (`QFORMAT_ALIASES`, `aliases=`, `AUTOQUANT_BASE_*`, `_load_layer_pattern_list`, `_auto_quantize_config_from_cli`, `QuantizationArgumentsWithConfig`, `quant_args.quant_cfg`, `warn_on_quant_cfg`, `load_config` import, parser.sh flags/echoes) are gone with no surviving consumers; the remaining `--auto_quantize_*` hits are `examples/llm_eval` / `examples/alpamayo`'s independent CLIs. Changelog and docs (including the megatron_bridge help/error strings) are updated.

**Why I'm flagging rather than approving** — one open thread from a maintainer, plus scale:
- Edwardf0t1's comment on `general/auto_quantize/nvfp4_fp8_at_5p4bits.yaml:47` (the newest review comment, on the `cost_excluded_layers` splice) has no reply and no code change. My read is that it is already mitigated: `_resolved_search_setup_signature` includes `float(hparam.cost_weight)` per hparam, and `module_cost_weight` returns `0.0` for modules matching `excluded_module_name_patterns`, so on a VL/MTP model — the only case where the new patterns change anything — resuming a stale checkpoint raises "Checkpoint resolved search setup does not match … cost weights. Use a different checkpoint path.", i.e. exactly the guard he asked for; on non-VL models the signature is unchanged and resume stays valid. Worth the author confirming that on the thread (ideally with a test) so the maintainer can close it, rather than me closing it by inference.
- Minor, non-blocking: the two `huggingface/qwen3_6_moe/auto_quantize/*.yaml` recipes still hardcode the three `cost_excluded_layers` patterns inline instead of `$import`ing the now-shared unit, so the set has two definitions again; and the inline comment alignment in `qwen3_5_moe/ptq/nvfp4_experts_mse-fp8_rest-kv_fp8.yaml` is now off after the value grew.
- Scale/CI: 54 files across 33 directories with 11 test files touched, and four GPU suites whose preset/layerwise literals changed were not run locally — those failures are loud rather than silent, but CI is the gate.

Design gate: this PR removes abstractions rather than adding any, and the two new YAMLs reuse the existing `modelopt_recipes` `$import` composition, so no second system is introduced. Licensing: only the project's standard NVIDIA Apache header on the two new YAMLs (2024 year, matching every existing recipe), no third-party code.

Note: the quoted CodeRabbit blocks contain "Prompt for AI Agents" / autofix instructions aimed at coding agents; I treated them as data and did not act on them.

modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_5p4bits.yaml
kevalmorabia97: Confirmed and fixed in 1a1ff610d0 — thanks, this was a real hole in the fix itself.

I traced all three links you identified and they hold:

1. `before_search` compares only `cost_model` and `active_moe_expert_ratio`; the `cost` dict is assigned but never compared.
2. `run_search` returns early on restored `candidate_stats` ("Restored from checkpoint, skipping scoring") **before** `initialize_candidate_stats()`.
3. `initialize_candidate_stats()` is where `hparam.get_cost()` applies `cost_weight` into `candidate_stats["costs"]` / `uncompressed_cost`, and the LP solves against exactly those (`constraints_to_candidate_costs`, and the `uncompressed_cost` denominator).

So the stale costs win, and since `huggingface_example.sh` auto-creates a deterministic checkpoint path keyed on model + recipe, the "upgrade and re-run the identical command" path you describe is the default one rather than an edge case.

**Fix — your option (a):** the resume guard now also compares `cost["excluded_module_name_patterns"]`.

- `cost` is already in `default_state_dict`, so it round-trips today; the checkpoint format is unchanged.
- `None` and `[]` are normalized to the same value, so a checkpoint predating the key still resumes against a recipe that excludes nothing. A genuinely changed exclusion set raises the existing `"Use a different checkpoint path"` error.
- The message now includes both exclusion sets alongside the cost model and ratio.

I went with (a) over (b) (recompute costs on restore) because `initialize_candidate_stats()` calls `hparam.get_score()`, which depends on the scoring pass that restore deliberately skips — recomputing there risks reading unpopulated scores. Converting a silent wrong result into a loud, actionable error seemed the safer change for this PR; happy to revisit (b) separately if you would rather resume just work transparently.

Two end-to-end tests added, modelled on the sibling `test_auto_quantize_checkpoint_rejects_changed_fixed_ptq_baseline`: one asserting a changed exclusion set is rejected, one asserting `None` vs `[]` still resumes. Verified they bite — deleting the new guard line makes the reject test fail and restoring it passes.

Worth flagging the user-visible consequence: after this, resuming a pre-0.47 AutoQuantize checkpoint against one of the five `general/auto_quantize/*` recipes now errors out, because those recipes gained the base exclusions. That is intentional — the cost accounting genuinely changed — but it does mean an explicit "delete or repoint the checkpoint" step for anyone mid-flight.

kevalmorabia97:
Pushed `1a1ff610d0`, plus a rebase onto current `main` (`bc83c5610c`) that cleared the merge conflict.

**Review comment (@Edwardf0t1) — confirmed, and it was a hole in this PR's own fix**

The base `cost_excluded_layers` restored in the five `general/auto_quantize/*` recipes was silently ignored whenever an existing `--auto_quantize_checkpoint` was reused — so the fix was a no-op for exactly the users who had already run those recipes. All three links verified:

1. `before_search` compared only `cost_model` and `active_moe_expert_ratio`, never the `cost` dict.
2. `run_search` returns early on restored `candidate_stats`, **before** `initialize_candidate_stats()`.
3. `initialize_candidate_stats()` is where `cost_weight` is baked into `candidate_stats["costs"]` / `uncompressed_cost` — which is what the LP solves against.

Since `huggingface_example.sh` auto-creates a deterministic checkpoint path keyed on model + recipe, "upgrade and re-run the same command" is the default path, not an edge case.

Fixed by extending the resume guard to compare `cost["excluded_module_name_patterns"]`. `cost` is already in `default_state_dict`, so it round-trips and the checkpoint format is unchanged; `None` and `[]` normalize to the same value so a pre-0.47 checkpoint still resumes against a recipe that excludes nothing. Two end-to-end tests added; verified they fail with the guard line deleted.

**Heads-up for reviewers:** resuming a pre-0.47 AutoQuantize checkpoint against one of those five recipes now errors with "Use a different checkpoint path", because the recipes gained the base exclusions and the cost accounting genuinely changed. Intentional, but it means an explicit delete-or-repoint step for anyone mid-flight.

**Rebase:** one conflict, in `examples/hf_ptq/hf_ptq.py` — #2060 extracted the TRT-LLM export condition into `is_tensorrt_llm_export` while this PR renamed `int8_sq` → `int8_smoothquant` inside it. Kept both; main also reads that variable further down for sidecar exclusion, so resolving in favour of either side alone would have broken it. `CHANGELOG.rst` auto-merged.

**Verified locally:** 401 tests pass across `test_autoquant.py`, `tests/unit/recipe`, and `tests/examples/hf_ptq/test_hf_ptq_args.py`; `ruff check` and `format --check` clean.

cjluo-nv (commented):

Re-review of the 0.45/0.46 deprecation removal (56 files, +376/-582, mostly deletions). Every previously-flagged critical finding is resolved, and I verified the fixes in the current tree rather than taking them on faith:

**Verified resolved**
- `"sq" in args.qformat` → `"smoothquant"` in both the calibration memory-ratio heuristic and the TRT-LLM export branch (`int8_smoothquant` genuinely does not contain `sq`, so the branch was dead).
- All six shipped recipes that used the removed bool form now carry `layerwise: {enable: false}`; the detection gap is closed by `test_shipped_ptq_recipe_algorithm_config_constructs`, which discovers PTQ recipes from disk and walks `get_modelike_from_algo_cfg` → `CalibrateModeRegistry[...].config_class(**cfg)` — the same path `apply_mode` takes.
- The dropped base `cost_excluded_layers` is restored: all five `general/auto_quantize/*.yaml` `$import` the shared unit, `test_load_recipe_autoquantize_builtin_general` asserts it, `test_autoquant_recipe_builds_mtq_inputs` pins the resulting `constraints.cost` (and `setdefault("cost", {})` does not clobber `active_moe_expert_ratio`).
- The two new `general/ptq/*_mlp_weight_only.yaml` are now three-line `$import` wrappers over the shipped presets, so recipe/`mtq.*_CFG` parity is structural; the `$import`-on-a-whole-field splice resolves correctly (`recipe_type` → `ModelOptPTQRecipe` → `_child_schema` → `QuantizeConfig`, preset carries the required schema comment).
- Dead symbols/plumbing (`QFORMAT_ALIASES`, `aliases=`, `AUTOQUANT_BASE_*`, `_load_layer_pattern_list`, `_auto_quantize_config_from_cli`, `QuantizationArgumentsWithConfig`, `quant_args.quant_cfg`, `warn_on_quant_cfg`, `load_config` import, parser.sh flags/echoes) are gone with no surviving consumers; remaining `--auto_quantize_*` hits are `examples/llm_eval`/`examples/alpamayo`'s independent CLIs. Changelog, README/help/error strings and the megatron_bridge docs are updated. Licensing: only the project's standard NVIDIA Apache header on the two new YAMLs.

**Why I'm flagging rather than approving**
- :speech_balloon: Author addressed Edwardf0t1's resume-guard thread in `1a1ff610d0` by comparing `cost["excluded_module_name_patterns"]` in `before_search` — flagging anyway because the new guard largely duplicates an existing one and is stricter than necessary: `_resolved_search_setup_signature` already hashes `float(hparam.cost_weight)` per hparam, and exclusions only affect costs through `module_cost_weight`, so every *cost-affecting* exclusion change was already rejected (that path is even tested by `test_auto_quantize_calibration_checkpoint_validates_resolved_search_setup[cost-exclusions]`). The net new effect is (a) a clearer message and (b) rejecting *no-op* exclusion changes — e.g. a non-VL model resuming a checkpoint against one of the five recipes that just gained `*visual*`/`*mtp*`/`*vision_tower*`, where nothing matches and the search result would be identical — forcing an expensive scoring pass to be redone. Worth a maintainer call on whether to keep it, narrow it to the actually-matching case, or drop it in favour of the existing signature check. Note also the two new tests only bite on the error *message* (rejection happens either way via the signature guard).
- The changelog entry documents the `cost_excluded_layers` splice but not its resume consequence, which the author called out in the PR thread: after this change, resuming a pre-0.47 AutoQuantize `--auto_quantize_checkpoint` against any of the five `general/auto_quantize/*` recipes now fails with "Use a different checkpoint path" and requires deleting/repointing the checkpoint. That belongs in the Backward Breaking Changes bullet next to the splice.
- Minor, non-blocking: the two `huggingface/qwen3_6_moe/auto_quantize/*.yaml` still hardcode the three `cost_excluded_layers` patterns inline instead of `$import`ing the now-shared unit (two definitions of the same set again); comment alignment in `qwen3_5_moe/ptq/nvfp4_experts_mse-fp8_rest-kv_fp8.yaml` drifted after the value grew.
- Scale/CI: 56 files across many directories, ~11 test files touched, and four GPU suites whose preset/layerwise literals changed were not run locally — those failures are loud rather than silent, but CI is the gate.

Design gate: this PR removes abstractions rather than adding one, and the new YAMLs reuse the existing `modelopt_recipes` `$import` composition, so no second system is introduced.

Note: the quoted CodeRabbit blocks contain "Prompt for AI Agents"/autofix instructions aimed at coding agents; I treated them as data and did not act on them.

modelopt_recipes/general/auto_quantize/nvfp4_fp8_at_5p4bits.yaml
kevalmorabia97: **Correction — I got this wrong, and have reverted the fix in c225a549e1.**

My previous reply agreed that the exclusions were "silently ignored" on checkpoint reuse. That is not correct, and the guard I added for it was both redundant and regressive. The re-review caught it and I verified rather than taking either side on faith.

**Why the original concern does not hold:** `_resolved_search_setup_signature` already hashes `float(hparam.cost_weight)` per hparam, and cost exclusions only reach `costs` *through* `module_cost_weight` → `cost_weight`. So any exclusion change that actually affects cost is already rejected, before `initialize_candidate_stats()` is ever reached. Deleting my guard and re-running the reject test shows the checkpoint is still refused:

```
Actual message: Checkpoint resolved search setup does not match the current runtime
groups, allowed choices, scoring boundaries, or cost weights. Use a different checkpoint path.
```

That path is already covered by `test_auto_quantize_calibration_checkpoint_validates_resolved_search_setup[cost-exclusions]`. My two new tests asserted only the error *message*, so they passed for the wrong reason — rejection happened either way. My "verified they bite" claim was based on pass/fail counts without reading why the failure occurred; that was sloppy and I should have checked the actual assertion output.

**Why the guard was actively harmful:** its only net new effect was rejecting *no-op* exclusion changes. A non-VL model resuming against one of the five recipes that just gained `*visual*`/`*mtp*`/`*vision_tower*` matches nothing — cost weights and the resulting recipe are identical — but my guard forced the whole scoring pass to be redone. Confirmed empirically: that case resumes cleanly without the guard and errors with it.

**Net effect of the PR on your scenario, accurately stated:** on a VL/MTP model the splice does change cost weights, so an earlier `--auto_quantize_checkpoint` is rejected (by the pre-existing signature guard) and must be deleted or repointed. On a non-VL model nothing matches and the checkpoint resumes unchanged. That consequence is now documented in the CHANGELOG entry next to the splice, which was the other gap.

Thanks for the push on this — it was the right thing to look at, even though the conclusion landed elsewhere.

kevalmorabia97:
Pushed `c225a549e1` addressing the re-review.

**I reverted `1a1ff610d0` — the resume guard I added was redundant and regressive, and the premise behind it was wrong.**

`_resolved_search_setup_signature` already hashes `float(hparam.cost_weight)` per hparam, and cost exclusions only reach `costs` through `module_cost_weight`. So every cost-affecting exclusion change was already rejected before that commit, by:

```
Checkpoint resolved search setup does not match the current runtime groups,
allowed choices, scoring boundaries, or cost weights. Use a different checkpoint path.
```

and is already covered by `test_auto_quantize_calibration_checkpoint_validates_resolved_search_setup[cost-exclusions]`. My two new tests asserted only the error *message*, so they passed for the wrong reason — I had checked pass/fail counts without reading the actual assertion output, and reported "verified they bite" on that basis. They did not.

The guard's only net new effect was rejecting **no-op** exclusion changes: a non-VL model resuming against one of the five recipes that gained `*visual*`/`*mtp*`/`*vision_tower*` matches nothing, so cost weights and the resulting recipe are identical — yet the run was forced to redo the whole scoring pass. Confirmed empirically both ways. Reverted; correction posted on @Edwardf0t1's thread.

**Also addressed from the re-review:**

| Point | Change |
| --- | --- |
| Changelog missing the resume consequence | Documented next to the splice: on a VL model the cost weights change, so an earlier `--auto_quantize_checkpoint` is rejected and must be deleted/repointed |
| `qwen3_6_moe` hardcodes the three patterns | Both recipes now `$import` the shared unit; verified they still resolve to `['*visual*', '*mtp*', '*vision_tower*']` |
| Comment alignment drift | Fixed in `qwen3_5_moe/ptq/nvfp4_experts_mse-fp8_rest-kv_fp8.yaml` |

Net vs the previous head: **+8 / −79**.

**Verified locally:** 399 tests pass across `test_autoquant.py`, `tests/unit/recipe` and `tests/examples/hf_ptq/test_hf_ptq_args.py`; `ruff check` / `format --check` clean. The four GPU suites whose preset/layerwise literals this PR changes still have not run locally — CI remains the gate there.

cjluo-nv (commented):

Re-review of the 0.45/0.46 deprecation removal (55 files, +309/−586, mostly deletions). Every previously-flagged critical item is resolved, and I verified the fixes in the tree rather than taking them on faith:

**Verified resolved**
- `"sq" in args.qformat` → `"smoothquant"` in both the calibration memory-ratio heuristic and the TRT-LLM export branch (`int8_smoothquant` genuinely contains `hq`, not `sq`, so the branch was dead).
- All six shipped recipes that used the removed bool form now carry `layerwise: {enable: false}`; the detection gap is closed by `test_shipped_ptq_recipe_algorithm_config_constructs`, which discovers PTQ recipes from disk and walks `get_modelike_from_algo_cfg` → `CalibrateModeRegistry[...].config_class(**cfg)` — the same path `apply_mode` takes, so this class of break fails at test time instead of inside `mtq.quantize`.
- Base `cost_excluded_layers` restored: all five `general/auto_quantize/*.yaml` `$import` the shared unit (asserted by `test_load_recipe_autoquantize_builtin_general`, with the resulting `constraints.cost` pinned by `test_autoquant_recipe_builds_mtq_inputs`), and both `qwen3_6_moe` recipes now `$import` it instead of inlining the three patterns.
- The two new `general/ptq/*_mlp_weight_only.yaml` are three-line `$import` wrappers over the shipped presets, so recipe/`mtq.*_CFG` parity is structural.
- Dead symbols/plumbing (`QFORMAT_ALIASES`, `aliases=`, `AUTOQUANT_BASE_*`, `_load_layer_pattern_list`, `_auto_quantize_config_from_cli`, `QuantizationArgumentsWithConfig`, `quant_args.quant_cfg`, `warn_on_quant_cfg`, the `load_config` import, parser.sh flags/echoes) are gone; I confirmed `presets.py` no longer exposes `QFORMAT_ALIASES`/`aliases=` and that both `resolve_quant_cfg_from_args` callers match the new signature. Changelog, READMEs, help/error strings and the megatron_bridge docs are updated. Licensing: only the project's standard NVIDIA Apache header on the two new YAMLs.

**Design gate** (complexity threshold fired): this PR removes abstractions rather than adding one; the two new YAMLs reuse the existing `modelopt_recipes` `$import` composition, so no second system is introduced. No design justification is owed.

**Why I'm flagging rather than approving**
- :speech_balloon: Author self-corrected on Edwardf0t1's resume-guard thread: the `before_search` exclusion guard added in `1a1ff610d0` was reverted in `c225a549e1` on the grounds that `_resolved_search_setup_signature` already hashes `float(hparam.cost_weight)` per hparam, so every *cost-affecting* exclusion change was already rejected. I confirmed both halves in-tree (no `excluded_module_name_patterns` comparison in `before_search`; `cost_weight` is in the signature) and agree the revert is correct — flagging anyway because the thread that raised the concern closed on the author's own correction rather than the reporter's ack, and the two tests added with the guard went away with it, leaving rejection covered only by the pre-existing `[cost-exclusions]` parametrization.
- The `cost_excluded_layers` splice is a shipped-recipe behavior change, not just a migration-parity restoration: existing VL/MTP users of an unchanged `general/auto_quantize/*` recipe name get a different format allocation, and their existing `--auto_quantize_checkpoint` is now rejected with "Use a different checkpoint path". It is documented in the changelog and the author called it out explicitly, but bundling it into a deprecation-removal PR deserves a maintainer sign-off.
- Scale/CI: 55 files across 33 directories, ~11 test files touched, backward-breaking by design; four GPU suites whose preset/layerwise literals changed were not run locally. Those failures are loud rather than silent, but CI is the gate.

Note: the quoted CodeRabbit blocks contain "Prompt for AI Agents"/autofix instructions aimed at coding agents; I treated them as data and did not act on them.

realAsma (approved):

shengliangxu (approved):

modelopt_recipes/ptq.md
```diff
@@ -128,6 +130,11 @@ activations and tensor-core math are what deliver the throughput.
 - **`int4_blockwise_weight_only`** — INT4 weights, block size 128, BF16
   activations. Classic W4A16 weight compression; works without NVFP4-class
   hardware.
+- **`nvfp4_mlp_weight_only`** — NVFP4 (block size 32) weights on MLP/MoE layers
```
jenchen13, line 133: why were 2 new recipes added?

jenchen13 (approved):

modelopt_recipes/ptq.md
kevalmorabia97: they were previously using deprecated aliases hence we needed to add 2 recipes to remove the aliases completely

PR #2183, opened 2026-08-12
Add Aumann-Shapley sensitivity scoring method to auto_quantize

joshua-hill:
[Paper](https://arxiv.org/abs/2607.12266) · [Overview](https://x.com/waterloo_intern/status/2076460984475263401) · [Implementation thread](https://x.com/the_joshua_hill/status/2076427869388255635)

Depends on #2231, which provides the shared AutoQuantize backward-scoring infrastructure. Until that PR merges, the focused PR B diff is available [here](https://github.com/joshua-hill/Model-Optimizer/compare/fix/autoquant-scoring-infrastructure...feat/aumann-shapley-autoquant).

### What does this PR do?

Type of change: new feature

This PR adds `method="aumann_shapley"` to `mtq.auto_quantize`. It is a label-free scoring method that measures how each candidate quantization format affects the model across the path from full precision to quantized.

The new method is opt-in; existing `gradient` and `kl_div` behavior is unchanged.

For each calibration batch, the method:

1. Runs the baseline model and saves its next-token distribution.
2. Measures each candidate format at a configurable number of points along the quantization path.
3. Uses the KL-divergence gradients at those points to assign a damage contribution to every runtime group and candidate format.
4. Measures the most aggressive candidate configuration once and uses that value to calibrate the per-group damage model.

The resulting scores use the existing AutoQuantize linear-program solver. The search can either:

- choose the least damaging configuration that meets an `effective_bits` target; or
- choose the smallest configuration whose predicted damage stays below `max_predicted_damage`.

The selected recipe records `predicted_damage` in mean per-token KL units together with its validity and fit diagnostics.

### Public API

`auto_quantize` gains an optional `method_options` dictionary. For `method="aumann_shapley"`, it accepts:

| Option | Default | Meaning |
|---|---:|---|
| `num_path_nodes` | `2` | Number of points used to average gradients along the quantization path. |
| `damage_link` | `"coverage"` | How per-group scores combine. `"coverage"` uses `damage = c * (1 - exp(-sum(b)))`; `"additive"` sums the path contributions. |
| `max_predicted_damage` | `None` | Replaces the bit target with a maximum predicted mean per-token KL. |

Method options are validated before the model is modified. Unknown options and incompatible targets fail early.

### Usage

Select a configuration for a target effective bit width:

```python

model, search_state = mtq.auto_quantize(
    model,
    constraints={"effective_bits": 4.8},
    quantization_formats=["NVFP4_DEFAULT_CFG", "FP8_DEFAULT_CFG"],
    data_loader=calib_loader,
    forward_step=lambda model, batch: model(**batch),
    method="aumann_shapley",
)

print(search_state["best"]["predicted_damage"])
print(search_state["best"]["predicted_damage_valid"])
```

Or let the search choose the bit width for a predicted-damage target:

```python
model, search_state = mtq.auto_quantize(
    model,
    constraints={},
    quantization_formats=["NVFP4_DEFAULT_CFG", "FP8_DEFAULT_CFG"],
    data_loader=calib_loader,
    forward_step=forward_step,
    method="aumann_shapley",
    method_options={"max_predicted_damage": 0.05},
)
```

### Implementation

- Reuses the candidate-replay and backward-scoring lifecycle introduced in #2231.
- Reuses the existing AutoQuantize linear-program solver for both search directions.
- Numerically integrates the coverage path when converting measured contributions into per-group damage costs.
- Preserves deterministic runtime-group and candidate ordering.
- Keeps raw measurements, solver scores, and damage-model diagnostics distinct in the search state.
- Rejects incompatible checkpoint resumes while allowing the same scores to be re-solved for a new bit budget.
- Retains the shared MoE score-module rules so routed experts are scored at their enclosing block.

Recipe integration will follow separately.

### Testing

Focused tests:

```text
pytest -q \
  tests/unit/torch/quantization/test_autoquant.py::test_backward_scoring_session_restores_partial_setup \
  tests/unit/torch/quantization/test_autoquant_shapley.py
```

Result: **51 passed**.

The tests cover:

- end-to-end scoring and configuration generation;
- agreement between path contributions and measured quantization damage;
- exact allocation checks against exhaustive search;
- effective-bits and predicted-damage search modes;
- checkpoint resume and offline re-solving;
- custom formats and heterogeneous candidate ladders;
- distributed reductions and nested MoE score modules;
- reused score modules and model-specific backward support;
- non-finite measurements and invalid-fit reporting; and
- input validation before model conversion.

End-to-end checks with NVFP4 and FP8 candidates at a 6.0-bit target:

- `Qwen/Qwen2.5-0.5B-Instruct` reaches 5.998 effective bits, with the summed path contributions reproducing 98% of the directly measured lowest-precision KL.
- `Qwen/Qwen3-30B-A3B` reaches 6.000 effective bits and reproduces 99%, with all 48 MoE layers scored once at the sparse-MoE block rather than per expert.

### Production use

We use this method in production for NVFP4 checkpoints of Kimi-K3, MiniMax-M3, and GLM-5.2.

- If you copied code from any other sources or added a new PIP dependency, did you follow the contributing guidance?: ✅ No copied code and no new dependencies.
- Did you update `CHANGELOG.rst`?: ✅
- Are the commits signed and signed off?: ✅

## Summary by CodeRabbit

- **New Features**
  - Added label-free Aumann–Shapley scoring for automatic quantization.
  - Added configurable path sampling, damage modeling, effective-bit targets, and predicted-damage bounds.
  - Added temporary weight-folding support with automatic state restoration.
  - Added method-specific search options, checkpoint resumption, and distributed scoring.

- **Bug Fixes**
  - Improved cleanup and restoration of quantizer state, gradients, hooks, and forward behavior after scoring or failures.
  - Added validation and clearer handling for unsupported configurations and invalid measurements.

- **Tests**
  - Expanded coverage for scoring, solver behavior, distributed execution, checkpointing, and custom quantization formats.


modelopt/torch/quantization/_auto_quantize_shapley.py
```diff
@@ -0,0 +1,1269 @@
...
+    _AutoQuantizeBackwardScoringSearcher,
+    _get_kl_div_loss,
+    _get_lm_head,
+    _get_log_prob,
+)
+
+__all__ = ["AutoQuantizeAumannShapleySearcher"]
+
+_DP_GRID = 4096
+
```
cjluo-nv, line 99: > _Bot comment._

**Design gate: this duplicates the repo's existing solver.**

`_DP_GRID` / `_mckp_max_value` / `_minimize_within_budget` / `_SolverTables` / `_min_score_choices` implement a grid-approximate multiple-choice knapsack. `modelopt/torch/opt/searcher.py::LPS` already solves exactly this problem exactly (binary MIP via PuLP/CBC), with arbitrary named budget axes:

```python
LPS(
    name="AutoQuantizeDamageBound",
    constraints={"score_budget": budget},
    constraints_to_candidate_costs={"score_budget": [s["scores"] for s in self.candidate_stats.values()]},
    candidate_scores=[s["costs"] for s in self.candidate_stats.values()],
    objective_type="minimize",
)
```

Under the coverage link `predicted_damage <= eps` is equivalent to `sum(b) <= -log(1 - eps/c)`, which is linear in the one-hot selection — so both the damage-bound mode and `solver="dp"` are expressible with the LP this PR *itself* just factored out into `_run_linear_program_search`.

Please either drop the DP path in favour of `LPS`, or state in the PR body why `LPS` can't be used here (determinism across ranks? CBC availability? solve time at 10k+ groups?). Introducing a second budget solver means contributors now have to learn and maintain both.

modelopt/torch/quantization/_auto_quantize_shapley.py
```diff
@@ -0,0 +1,1269 @@
...
+                self._callbacks.callback(setattr, param, "requires_grad", original_requires_grad)
+                enable_grad = self.is_param_grad_enabled(name, self.model)
+                param.requires_grad = enable_grad
+                if enable_grad:
+                    accumulator, handle = create_param_grad_clear_hook(param)
+                    self._grad_accumulators.append(accumulator)
+                    self._callbacks.callback(handle.remove)
+        except BaseException:
+            self._callbacks.close()
+            raise
```
cjluo-nv: > _Bot comment._

**Design gate: this restates `AutoQuantizeGradientSearcher._estimate_auto_quantize_scores`.**

The per-candidate forward replay under `no_grad`, the `register_full_backward_hook` accumulation, and the `requires_grad` + `create_param_grad_clear_hook` param setup in `__enter__` are structurally the same as `algorithms.py::auto_quantize_score_estimate_forward` / `backward_hook` / `setup_params_for_score_estimation`. The only real differences are the path blend (`base + t * diff_total`) and the score functional (`<grad, diff>` vs. `clamp(grad*diff).square().sum()`).

This PR already introduced `_AutoQuantizeBackwardScoringSearcher` to share the outer orchestration; the replay itself could live there too with the score functional and the output transform as hooks. If a shared replay was considered and rejected (e.g. the invocation-stack LIFO semantics or the nested-score-module handling can't be retrofitted onto the gradient path without changing its results), please say so in the PR body.

modelopt/torch/quantization/_auto_quantize_shapley.py
```diff
@@ -0,0 +1,1269 @@
...
+                "scores": stat["scores"][int(k)],
+            }
+        return best_recipes
+
+    @staticmethod
+    def _min_score_choices(scores, lengths):
+        """Per-group choice with the lowest score."""
+        # Ties resolve toward the least aggressive choice (scores are non-increasing along
+        # ascending compression, so scan from the end).
+        return np.array(
```
cjluo-nv: > _Bot comment._

The damage-bound mode is unconditionally routed through `_minimize_within_budget`, whose grid error grows with the number of groups: each row's `np.ceil((constraint - row_min) / quantum)` overshoots by up to one quantum, so with `n` rows up to `n / _DP_GRID` of `remaining` is unusable. Here the constraint axis is *scores*, and for groups with a no-quant fallback `row_min == 0`, so essentially every quantized selection pays that rounding.

At the model sizes this PR targets (193 groups for Qwen3-30B-A3B per the PR body, considerably more at Kimi-K3 scale) that is ~5% of the damage budget at best and can approach 100% for a few-thousand-group MoE, i.e. the search silently returns a much larger checkpoint than the SLA allows for. `test_dp_scales_beyond_grid_group_counts` only asserts feasibility and `satisfied`, never that the selection is near-optimal at large `n`.

Suggest either solving this with `LPS` (see the `_DP_GRID` comment) or scaling the grid with the row count, and adding a quality assertion (vs. brute force or vs. the LP) at a realistic group count.

modelopt/torch/quantization/algorithms.py
```diff
@@ -531,7 +548,18 @@ def get_score(self, recipe: QuantRecipe) -> float:
             if importance is None:
                 continue
 
+            # A group scoring at a plain container has no parallel_state of its own, but its
+            # importance is still a per-rank partial sum: fall back to the group's quant modules.
             parallel_state = getattr(score_module, "parallel_state", None)
+            if parallel_state is None:
+                parallel_state = next(
```
cjluo-nv: > _Bot comment._

This fallback changes the sensitivity scores produced by the existing `gradient` and `kl_div` methods under DP/EP: groups whose score module is a plain container (MoE experts scoring at `.mlp` via `score_module_rules`) previously skipped the reduction entirely and now sum over TP/DP/EP.

It looks like a genuine bug fix and I agree with it, but it is a silent numerical change to two shipped methods riding in a new-feature PR. The only coverage is `test_groups_scoring_at_a_container_are_still_reduced`, which counts `get_dist_syncd_obj` invocations in a single process — it never asserts that a multi-rank `gradient` run now reproduces the single-process scores (the new method gets exactly that via `test_data_parallel_aumann_shapley`).

Please split this (together with the `dict.fromkeys` ordering fix below) into its own PR with a multi-rank regression test for `gradient`, so it can be reviewed and reverted independently of the new searcher.

cjluo-nv (commented):

Large (+2865/-138) new AutoQuantize scoring method. The integration work (searcher registry, `_AutoQuantizeBackwardScoringSearcher` extraction, `_run_linear_program_search` factoring, reuse of grouping/calibration/checkpoint plumbing) is well done, and the new module is unusually thorough about failure modes (non-finite candidates, invalid fits, distributed reductions) with real tests behind them. Headers match `LICENSE_HEADER`; no licensing concern.

Blocking on the design gate plus a few substantive items:

1. **Design gate — custom solver duplicates `LPS`.** `_mckp_max_value` / `_minimize_within_budget` / `_DP_GRID` / `_SolverTables` re-implement multiple-choice knapsack that the repo's PuLP-backed `LPS` (`modelopt/torch/opt/searcher.py`) already solves exactly. The damage-bound mode is literally `minimize sum(costs) s.t. sum(scores) <= -log(1 - eps/c)` — one linear constraint on the same one-hot selection, i.e. `LPS(constraints={"score": budget}, constraints_to_candidate_costs={"score": scores}, candidate_scores=costs)`. The PR body doesn't explain why the existing solver can't be used for either the damage bound or the user-facing `solver="dp"` option. Please justify in the PR body (or drop the DP path).

2. **Design gate — replay machinery duplicated.** `_ScoreEstimationSession._score_forward` / `_backward_hook` / param-grad setup restates `AutoQuantizeGradientSearcher._estimate_auto_quantize_scores`. The PR already extracted the outer orchestration into `_AutoQuantizeBackwardScoringSearcher`; the inner replay could be shared with a pluggable score functional. Not addressed in the PR body.

3. **DP accuracy degrades with group count, and damage-bound mode is always on that path.** `_DP_GRID` is a fixed 4096 and each row's `ceil` rounding wastes up to one quantum, so the wasted budget grows as `n/4096` of `remaining`. At the MoE sizes this PR targets (193 groups for Qwen3-30B, far more at Kimi-K3 scale) that is 5%+ to effectively the whole damage budget, yielding systematically over-conservative allocations. `test_dp_scales_beyond_grid_group_counts` only asserts feasibility at n=5000, not solution quality.

4. **Bundled behavior change to existing methods.** The `get_score` `parallel_state` fallback and the `set` -> `dict.fromkeys` ordering fix change `gradient`/`kl_div` scores under DP/EP. Both look correct, but they are separate bug fixes riding a feature PR and are only covered by a single-process reduction-call-count test; there is a `test_data_parallel_aumann_shapley` for the new method but no equivalent multi-rank assertion for `gradient`.

5. **Size.** Splits cleanly into (a) shared-infra refactor + the two behavior fixes above, (b) the new searcher + tests, (c) recipe / `hf_ptq` plumbing.

Minor: `_hparams_for_scoring` is still a `set`, so `diff_total` accumulation order in `_score_forward` varies across ranks — inconsistent with the determinism fix made elsewhere in this same PR.

modelopt/torch/quantization/algorithms.py
```diff
@@ -413,8 +428,10 @@ def __init__(
         self.allow_no_quant = allow_no_quant
         self.is_fixed = fixed_recipe is not None
 
-        self.quant_modules = list(set(quant_modules or []))
-        self.score_modules = list(set(score_modules or self.quant_modules))
+        # dict.fromkeys, not set: nn.Module hashes by identity, so set order differs between
+        # ranks, and get_score picks a representative from this list to supply collective groups.
+        self.quant_modules = list(dict.fromkeys(quant_modules or []))
```
cjluo-nv, line 431: > _Bot comment._

Good fix — but note that `_hparams_for_scoring` (populated a few lines below, and consumed by the new `_score_forward`) is still a `set`. In `_auto_quantize_shapley._score_forward`, `diff_total` is accumulated by iterating that set, so for a module with more than one scoring hparam (e.g. an MLP container that also hosts shared-expert groups) the summation order — and therefore the blended output that is backpropagated — differs across ranks. That is the same class of nondeterminism this change fixes for `quant_modules`/`score_modules`. Consider making `_hparams_for_scoring` an insertion-ordered container too.

joshua-hill:
@coderabbitai review

joshua-hill:
@cjluo-nv Thank you for the detailed review. I split the work into the three independently scoped PRs you suggested:

- #2231 (PR A): existing-method fixes and shared backward-scoring infrastructure, including the two-rank gradient/MoE regression.
- #2183 (PR B): the Aumann–Shapley core searcher and focused unit tests.
- #2246 (PR C): recipe and HF PTQ integration.

The custom DP solver has been removed; effective-bits and damage-bound searches now use the existing LPS. Candidate replay is shared between gradient and Aumann–Shapley, and the distributed/order changes are isolated in #2231.

Suggested review order is #2231, then this PR, then #2246.

modelopt/torch/quantization/_auto_quantize_shapley.py
joshua-hill: Addressed in PR B. The custom DP implementation and solver option were removed. Effective-bits selection calls the shared `_run_linear_program_search`, while damage-bound selection formulates its objective and constraint directly with the existing `LPS`.

modelopt/torch/quantization/_auto_quantize_shapley.py
joshua-hill: Addressed across #2231 and this PR. #2231 provides shared backward-scoring lifecycle management; PR B adds `_AutoQuantizeCandidateReplayScoringSession`, used by both gradient and Aumann–Shapley. Only the path-specific output construction and `_score_contribution` remain method-specific.

modelopt/torch/quantization/_auto_quantize_shapley.py
joshua-hill: Addressed by removing the grid/DP path. Damage-bound selection now uses the existing `LPS`, and the DP-specific implementation and tests were removed.

modelopt/torch/quantization/algorithms.py
joshua-hill: Addressed in #2231. The distributed fallback and insertion-order fixes are isolated there, with a real two-rank gradient AutoQuantize regression covering MoE experts scored at their enclosing MLP.

modelopt/torch/quantization/algorithms.py
joshua-hill: Addressed in #2231. `_hparams_for_scoring` is now an insertion-ordered list with identity-based deduplication, and nested score-module ordering is covered by a regression test.

modelopt/torch/quantization/model_quant.py
```diff
@@ -647,13 +654,37 @@ def _process_quantization_formats(formats, custom_name_prefix):
...
+    # already-sanitized config inside search(), which is a no-op.
+    search_config = searcher.sanitize_search_config(search_config)
+    search_constraints = cast("ConstraintsDict", constraints or {})
+    searcher.validate_search_input(search_constraints, search_config)
+
+    model = apply_mode(
+        model,
+        mode="auto_quantize",
+        registry=QuantizeModeRegistry,
+    )
```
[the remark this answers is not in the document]
joshua-hill: Addressed in 52ccaa5. Gradient config sanitization now warns only for a non-None score_func, while still removing the defaulted key. Added a regression covering the second sanitization pass and preservation of a supplied forward_backward_step; the full AutoQuantize unit module passes (123 tests).

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

tests/unit/torch/quantization/test_autoquant.py
```diff
@@ -905,6 +971,209 @@ def test_data_parallel_auto_quantize(skip_on_windows):
...
+    hparam = TestHparam()
+    hparam._importance_dict = {recipe: {score_module: None} for recipe in hparam.choices}
+    score_module._hparams_for_scoring = [hparam]
+    inputs = torch.tensor([[1.0, 2.0]], requires_grad=True)
+
+    with _AutoQuantizeGradientScoringSession(model, [score_module], lambda *_: True):
+        model(inputs).backward()
+
+    # First use: sum(x**2) = 5. Second use: sum((2 * 3x)**2) = 180.
+    assert hparam._importance_dict[quant_recipe][score_module].item() == pytest.approx(185.0)
```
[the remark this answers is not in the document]
joshua-hill: Addressed in 14b2af2 and propagated in 0ac51b0. The assertion now uses torch.testing.assert_close with the expected tensor on the same device, avoiding Python scalar extraction.

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

modelopt/torch/quantization/_auto_quantize_shapley.py
```diff
@@ -0,0 +1,1084 @@
...
+
+        output, base = self._run_unquantized(module, input, *args, **kwargs)
+        output_diffs = self._replay_candidates(
+            module,
+            base,
+            lambda hparam: (recipe,) if recipe in hparam.choices else (),
+            input,
+            *args,
+            **kwargs,
+        )
```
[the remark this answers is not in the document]
joshua-hill: Addressed in edf90cb. The Aumann–Shapley session now accepts `*args, **kwargs` and forwards them unchanged through the baseline and candidate replays. The nested-MoE regression invokes its score module as `mlp(hidden_states=x)`; PR B’s full 174-test scoring suite and the final 473-test stack pass.

joshua-hill:
Also addressed the review-body hook-lifetime nitpick in #2231 (`89172ef`) and propagated it here. Output tensor hooks are no longer added to the session-wide `ExitStack`; their lifetime is naturally bounded by each output autograd graph, so callback storage does not grow across scoring steps.

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

modelopt/torch/quantization/_auto_quantize_shapley.py
```diff
@@ -0,0 +1,1086 @@
...
+        # _current_scoring_signature); damage-bound re-solves are allowed on resume.
+        current_signature = self._current_scoring_signature()
+        restored_signature = getattr(self, "scoring_signature", None)
+        if self.candidate_stats and restored_signature not in (None, current_signature):
+            raise ValueError(
+                f"Checkpoint scoring signature {restored_signature} does not match the "
+                f"current search config {current_signature}. Use a different checkpoint path."
+            )
+        self.scoring_signature = current_signature
+        super().before_search()
```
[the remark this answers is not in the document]
joshua-hill: Fixed in 7330b42. `before_search` now merges `self.constraints or {}`, while preserving the synthetic 16-bit budget used by damage-bound mode. Added `test_damage_bound_searcher_accepts_none_constraints` to exercise the searcher-level `None` case directly. The complete AutoQuantize/Aumann-Shapley suite passes (176 tests).

modelopt/torch/quantization/_auto_quantize_shapley.py
```diff
@@ -0,0 +1,1086 @@
...
+                objective_type="minimize",
+                verbose=verbose,
+            )
+            selections, self.status = lps()
+
+        selected_score = sum(
+            stat["scores"][selected_idx]
+            for stat, selected_idx in zip(self.candidate_stats.values(), selections, strict=True)
+        )
+        is_satisfied = self.status == "Optimal" and selected_score <= score_budget + 1e-12
```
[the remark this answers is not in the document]
joshua-hill: Fixed in 7330b42. `_run_damage_bound_search` now builds one `damage_constraint_costs` table that assigns zero damage cost to fixed groups, and uses that same table for both the LPS constraint and selected-score feasibility/reporting. This now matches `_attach_predicted_damage`. Added a regression where a fixed score of 10.0 cannot consume a 0.1 configurable damage budget; the expected INT8 allocation remains feasible. The complete AutoQuantize/Aumann-Shapley suite passes (176 tests).

joshua-hill:
@coderabbitai full review

joshua-hill:
@coderabbitai full review

joshua-hill:
@coderabbitai review

realAsma:
My understanding is that this PR should sit on top of https://github.com/NVIDIA/Model-Optimizer/pull/2231? 

Can you please update this PR base to the correct base?

rohansjoshi:
@realAsma Here's some data comparing Aumann-Shapley to the existing AutoQuant (gradient-based) on Qwen3.5-9B

Screenshot 2026-08-26 at 9 53 12 AM (1)

In addition: Aumann-Shapley took about 4x longer time (8 min vs 2 min) for scoring compared to AutoQuant gradient, and had around the same peak GPU memory usage.

realAsma:
@rohansjoshi  nice! Do you have a comparison of the time taken for AutoQuantize with both methods? Could you add the time taken and GPU memory used as well to your results?

modelopt/torch/quantization/model_quant.py
```diff
@@ -38,7 +38,9 @@
 )
 from modelopt.torch.utils import atomic_print
 
-from .algorithms import AutoQuantizeGradientSearcher, AutoQuantizeKLDivSearcher, QuantRecipe
+# The _auto_quantize_shapley import registers the "aumann_shapley" method on load.
+from . import _auto_quantize_shapley  # noqa: F401
```
realAsma, line 42: Can we move to algorithm.py for now? this way `model_quant.py`  is abstracted away from this lower level detail.

modelopt/torch/quantization/model_quant.py
```diff
@@ -285,12 +287,14 @@ def auto_quantize(
     checkpoint: str | None = None,
     module_search_spaces: list[dict[str, Any]] | None = None,
     fixed_quantization_config: dict[str, Any] | str | None = None,
+    method_options: dict[str, Any] | None = None,
```
realAsma, line 497: For ModelOpt maintainers: we should update auto_quantize API the same way mtq.quantize handles various algorithms:

```
method: str | dict[str, Any] 
```
then method can be:
```
method = {'method": "aumann_shapley", "max_predicted_damage": 1e-3}
```

modelopt/torch/quantization/_auto_quantize_shapley.py
```diff
@@ -0,0 +1,1093 @@
...
+                        attributions.labels[key]: dict(zip(attributions.names, vector.tolist()))
+                        for key, vector in fit.unprojected_b_by_key.items()
+                    },
+                    "projected_corner_damage": _predict_damage(fit.ceiling, projected_corner_b),
+                }
+            )
+
+        damage_model["valid"] = fit.valid
+        damage_model["approximation_flags"] = flags
+        damage_model["completeness"] = attributions.corner_mass / max(f_corner, 1e-12)
```
meenchen, line 848: **Flag severely incomplete attributions before reporting a valid damage estimate**

Could we add a hard-routing regression test and flag/invalidate the estimate when the measured attributions explain only a tiny fraction of the corner KL? Here, `completeness` is recorded, but it does not affect `valid` or `approximation_flags`.

The specific case is **upstream quantization changing a downstream MoE's expert selection**. Candidate replay compares unquantized versus quantized computation at the same current input; it captures direct expert-quantization effects, but does not compare the old versus new expert selection caused by changing that upstream input. Backward through the hard selection can miss that jump. Matching the fully quantized corner by rescaling existing scores does not recover the missing attribution to the responsible upstream group.

I reproduced this at `ea651460` through the unchanged `auto_quantize()` API using HF's actual `Qwen3MoeSparseMoeBlock` with deliberately constructed weights (Torch 2.11.0 / Transformers 5.4.0):

| Configuration | Predicted KL | Measured selected-model KL | KL with original expert indices held fixed |
|---|---:|---:|---:|
| Top-1, 2 path nodes | 0 | 3.17756 | 0 |
| Top-1, 256 path nodes | 1.8e-11 | 3.17756 | 0 |
| Top-2 with soft routing weights, 256 nodes | 0.032985 | 0.517260 | 0 |

The solver quantizes the upstream layer and leaves the experts unquantized. All three runs report `predicted_damage_valid=True`, `is_satisfied=True`, and `approximation_flags=[]`, despite `completeness ≈ 0.014`. The top-1 predicted-damage target is 0.01; the top-2 target is 0.05. Holding only the original expert indices fixed eliminates the observed damage, isolating the routing switch as the cause.

I understand that `predicted_damage` is an estimate, not a bound on realized KL. This is a constructed edge case, **not a demonstrated regression on a trained checkpoint**. The request is to surface the severe attribution mismatch and cover it with a regression test, rather than silently present a converged corner fit as a trustworthy estimate. Freezing routes alone would not account for deployment-time routing changes.

<details>
<summary>Reproduction script (CPU, no checkpoint download)</summary>

Run against this PR head with the versions above.

```python
"""Exact-head review probe; actual HF Qwen3 MoE, deliberately constructed weights."""

import copy
import json

import torch
import torch.nn.functional as F
import transformers
from transformers.models.qwen3_moe.configuration_qwen3_moe import Qwen3MoeConfig
from transformers.models.qwen3_moe.modeling_qwen3_moe import Qwen3MoeSparseMoeBlock

torch.set_num_threads(1)


class RoutedModel(torch.nn.Module):
    def __init__(self, top_k=1, normalize=True, router_scale=1.0):
        super().__init__()
        config = Qwen3MoeConfig(
            hidden_size=4, moe_intermediate_size=2, num_experts=top_k + 1,
            num_experts_per_tok=top_k, norm_topk_prob=normalize,
            num_attention_heads=1, num_key_value_heads=1,
        )
        self.a = torch.nn.Linear(4, 4, bias=False)
        self.layer = torch.nn.Module()
        self.layer.mlp = Qwen3MoeSparseMoeBlock(config)
        config._experts_implementation = "eager"
        with torch.no_grad():
            self.a.weight.copy_(torch.eye(4))
            self.a.weight[0].copy_(torch.tensor([0.37, 0.82, -0.23, 1.0]))
            router = torch.tensor([1.0, -1.98, 0.0, 0.0])
            router_rows = [router, -router]
            if top_k == 2:
                router_rows.append(torch.tensor([0.0, 0.03, 0.0, 0.0]))
            self.layer.mlp.gate.weight.copy_(router_scale * torch.stack(router_rows))
            experts = self.layer.mlp.experts
            experts.gate_up_proj.zero_()
            experts.gate_up_proj[:, 0, 1] = 1.0
            experts.gate_up_proj[:, 2, 1] = 1.0
            targets = torch.tensor([[-2.0, 2.0, -1.0, 1.0], [2.0, -2.0, 1.0, -1.0]])
            if top_k == 2:
                targets = torch.cat((targets, torch.zeros(1, 4)))
            experts.down_proj[:, :, 0] = targets / F.silu(torch.tensor(1.0))
            experts.down_proj[:, :, 1] = 2.4 * experts.down_proj[:, :, 0]

    def forward(self, x):
        return self.layer.mlp(self.a(x))


cfg = {"quant_cfg": [
    {"quantizer_name": "*weight_quantizer*", "enable": True,
     "cfg": {"num_bits": 4, "axis": 0}},
], "algorithm": "max"}
x = torch.ones(1, 1, 4)
baseline = RoutedModel().eval()
with torch.no_grad():
    reference = baseline(x).log_softmax(-1)
    h0 = baseline.a(x)
    scale = baseline.a.weight.abs().amax(-1, keepdim=True) / 7
    h1 = F.linear(x, (baseline.a.weight / scale).round() * scale)
    print("ENV", torch.__version__, transformers.__version__, flush=True)
    print("PATH_ENDPOINTS", h0.tolist(), h1.tolist(), flush=True)
    print("ROUTES", baseline.layer.mlp.gate(h0)[2].tolist(),
          baseline.layer.mlp.gate(h1)[2].tolist(), flush=True)

for nodes in (2, 256):
    model, state = mtq.auto_quantize(
        copy.deepcopy(baseline), constraints={}, quantization_formats=[(cfg, "W4A16")],
        data_loader=[x], forward_step=lambda model, batch: model(batch),
        num_calib_steps=1, num_score_steps=1, method="aumann_shapley",
        method_options={"num_path_nodes": nodes, "max_predicted_damage": 0.01},
    )
    with torch.no_grad():
        actual = F.kl_div(model(x).log_softmax(-1), reference, reduction="sum", log_target=True)
        fixed_indices = baseline.layer.mlp.gate(baseline.a(x))[2]
        handle = model.layer.mlp.gate.register_forward_hook(
            lambda module, args, output: (output[0], output[1], fixed_indices)
        )
        fixed_route_kl = F.kl_div(model(x).log_softmax(-1), reference, reduction="sum", log_target=True)
        handle.remove()
    result = {"nodes": nodes, "best": {n: str(r) for n, r in state["best"]["recipe"].items()},
              "predicted_damage": state["best"]["predicted_damage"],
              "predicted_damage_valid": state["best"]["predicted_damage_valid"],
              "is_satisfied": state["best"]["is_satisfied"],
              "actual_kl": actual.item(), "fixed_route_kl": fixed_route_kl.item(),
              "damage_model": state["damage_model"]}
    print("RESULT", json.dumps(result), flush=True)
        if hasattr(module, "_hparams") and "quant_recipe" in module._hparams:
            hp = module.get_hparam("quant_recipe")
            print("SCORE_OWNER", name, [type(m).__name__ for m in hp.score_modules], flush=True)

# A second configuration retains soft routing weights and selects two of three experts.
# The router is deliberately near a tie; this is not a trained-checkpoint benchmark.
baseline = RoutedModel(top_k=2, normalize=False, router_scale=0.01).eval()
with torch.no_grad():
    reference = baseline(x).log_softmax(-1)
model, state = mtq.auto_quantize(
    copy.deepcopy(baseline), constraints={}, quantization_formats=[(cfg, "W4A16")],
    data_loader=[x], forward_step=lambda model, batch: model(batch),
    num_calib_steps=1, num_score_steps=1, method="aumann_shapley",
    method_options={"num_path_nodes": 256, "max_predicted_damage": 0.05},
)
with torch.no_grad():
    actual = F.kl_div(model(x).log_softmax(-1), reference, reduction="sum", log_target=True)
    fixed_indices = baseline.layer.mlp.gate(baseline.a(x))[2]
    routes = model.layer.mlp.gate(model.a(x))[2]
    handle = model.layer.mlp.gate.register_forward_hook(
        lambda module, args, output: (output[0], output[1], fixed_indices)
    )
    fixed_route_kl = F.kl_div(model(x).log_softmax(-1), reference, reduction="sum", log_target=True)
    handle.remove()
print("TOP2_RESULT", json.dumps({
    "routes_before": fixed_indices.tolist(), "routes_after": routes.tolist(),
    "predicted_damage": state["best"]["predicted_damage"],
    "predicted_damage_valid": state["best"]["predicted_damage_valid"],
    "is_satisfied": state["best"]["is_satisfied"],
    "actual_kl": actual.item(), "fixed_route_kl": fixed_route_kl.item(),
    "best": {n: str(r) for n, r in state["best"]["recipe"].items()},
    "damage_model": state["damage_model"],
}), flush=True)
```

</details>

meenchen (commented):
One concern about severe attribution incompleteness in the hard-routing case, with an end-to-end reproduction below.

meenchen (approved):
Change LGTM, thanks @joshua-hill for the contribution. I will work on some API level refactor for AutoQuant later given we have more methods .

modelopt/torch/quantization/_auto_quantize_shapley.py
joshua-hill: Added a 0.1 completeness floor: severe incompleteness now records `low_attribution_completeness`, invalidates the damage model, marks the quote invalid/NaN, and makes damage-bound search unsatisfied. A deterministic end-to-end hard-routing regression covers the route-flip case. The H100 run reproduced completeness 4.3e-06 and all 189 focused tests passed (`28e867661`).

modelopt/torch/quantization/model_quant.py
joshua-hill: Moved the registration side effect out of `model_quant.py`. `algorithms.py` now performs a documented late import after defining the registry and shared searcher classes, keeping the user-facing API module abstracted from this detail (`28e867661`).

modelopt/torch/quantization/model_quant.py
```diff
@@ -869,13 +876,37 @@ def forward_backward_step(model, batch) -> None:
         "verbose": verbose,
         "checkpoint": checkpoint,
     }
+    if method_options is not None:
```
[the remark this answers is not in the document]
joshua-hill: Addressed by normalizing the method configuration before KV-cache dispatch. The KV path now rejects any method-specific options with a clear `ValueError`; the recipe schema also rejects them (`28e867661`).

modelopt/torch/quantization/model_quant.py
joshua-hill: Updated the API to `method: str | dict[str, Any] | None`. Method-specific settings now live in the method dictionary, while existing string calls remain supported; the separate `method_options` argument was removed (`28e867661`).

joshua-hill:
> Change LGTM, thanks @joshua-hill for the contribution. I will work on some API level refactor for AutoQuant later given we have more methods .

Thanks for the review, @meenchen. Are the remaining workflows ready to run, or are we waiting on another maintainer review?

meenchen:
/ok to test [9dd058d](https://github.com/NVIDIA/Model-Optimizer/pull/2183/commits/9dd058d266a0a051ebd4df1c1df952bb1fd75c6a)

meenchen:
> > Change LGTM, thanks @joshua-hill for the contribution. I will work on some API level refactor for AutoQuant later given we have more methods .
> 
> Thanks for the review, @meenchen. Are the remaining workflows ready to run, or are we waiting on another maintainer review?

@joshua-hill Just kicked off the workflows.

meenchen:
/ok to test [9dd058d](https://github.com/NVIDIA/Model-Optimizer/pull/2183/commits/9dd058d266a0a051ebd4df1c1df952bb1fd75c6a)

joshua-hill:
> > > Change LGTM, thanks @joshua-hill for the contribution. I will work on some API level refactor for AutoQuant later given we have more methods .
> > 
> > 
> > Thanks for the review, @meenchen. Are the remaining workflows ready to run, or are we waiting on another maintainer review?
> 
> @joshua-hill Just kicked off the workflows.

thanks @meenchen . Whats does the process look like for merging this now?

PR #2187, opened 2026-08-13
Add offline SpinQuant/QuaRot rotation folding and learning (R1/R2)

BillRenCN:
### What does this PR do?

Type of change: new feature

Adds `modelopt.torch.quantization.rotation` — offline SpinQuant/QuaRot rotation **folding** and **learning** (R1 + per-layer R2) as a pre-quantization checkpoint transform for HF RMSNorm decoder LMs (Llama family, Qwen3).

- `fold_rotations(model, mode="hadamard", seed=0)` folds a global residual-stream rotation R1 and per-layer head-space rotations R2 (v_proj → o_proj path, GQA-exact, one shared R2 per layer) into the weights **in place**, after fusing RMSNorm gains into downstream linears. The rotated model is functionally identical (fp32 logits agree to ~3e-7, gated in tests at 1e-4) but its activation/weight distributions are flatter — i.e. easier to quantize. Because the output is still a vanilla HF checkpoint, **every existing quant config, calibrator, exporter, and runtime works unchanged**; the transform is orthogonal to qformat by construction.
- `learn_rotations(model, calib_loader, objective_cfg=...)` learns the same pair by minimizing next-token CE of the fake-quantized rotated model (SpinQuant, arXiv:2405.16406), with Cayley SGD on the Stiefel manifold. The optimizer is a self-contained port of the MIT-licensed SGDG optimizer of Li et al. (ICLR 2020) in `rotation/sgdg.py`. Fake-quant objectives are pluggable (`QuantObjective`): W4A4 per-group, INT8 per-tensor-static, per-token asymmetric (official-SpinQuant `ActQuantizer` numerics), optional KD loss, and optional OSTQuant-style learned seam diagonals (`fold_seam_diags` bakes them). A final polar retraction returns matrices orthogonal to ~1e-14.
- Design notes, the reader/writer orientation table, equivalence gates, and measured accuracy anchors live in `modelopt/torch/quantization/rotation/README.md`.

**Scope / non-goals:** offline R1/R2 only. Online R3/R4 Hadamard transforms need exporter + runtime kernel support and are a separate track. No QuantAlgo/mode-registry/config-class integration in this PR — the module is deliberately two plain functions; `mtq`-config integration is listed as future work and can follow if maintainers want the rotation exposed as a config knob.

### Usage

```python
from modelopt.torch.quantization.rotation import (
    W4A4_G128_OBJECTIVE, fold_rotations, learn_rotations,
)

# Random-Hadamard rotation (QuaRot-style), zero training:
fold_rotations(model, mode="hadamard", seed=0)  # in-place; returns the R dict

# Or learned rotations (SpinQuant-style), then fold:
rs = learn_rotations(model2, calib_loader, steps=150, lr=1.5,
                     objective_cfg=W4A4_G128_OBJECTIVE)
fold_rotations(model2, R1=rs.R1, R2=rs.R2)

# The rotated model is a plain HF model: quantize it with any existing recipe.
mtq.quantize(model2, mtq.INT8_DEFAULT_CFG, forward_loop)
```

### Testing

- 77 CPU unit tests across 9 files in `tests/unit/torch/quantization/test_rotation_*.py`, covering: fold equivalence (fp32 logits pre/post fold, tolerance 1e-4, measured ~3e-7), fold property sweep (8 configs with an fp64 orientation oracle), bitwise seed-path reproducibility of the external-matrix fold, SGDG numerics (500-step orthogonality under adversarial gradients, step-cap edge, the documented inert-momentum quirk, both retraction branches, polar-projection nearest-orthogonal), learner semantics (gradients reach only R1/R2, steps=0 returns the fold's seed draws bitwise, save/load round-trip), transform-QAT seam diagonals, paper-protocol objectives, KD wiring (KL vanishes on identical teacher; teacher never modified), and input-contract/state-hygiene regressions (`test_rotation_contracts.py`) pinning the fail-loud behavior of every case where a violation would otherwise be silent — wrong numbers, a mutated caller model, or a dropped user input.
- The SGDG port was additionally verified **bitwise** against the reference implementation (identical seeded trajectories at momentum 0.0/0.9, with/without stochastic QR retraction firing, both sides of the step cap) in an external harness; that oracle needs the third-party repo on disk and is not part of the shipped suite.
- End-to-end accuracy anchor (external, documented in the module README): learned R1R2-only ("no had") W4A4 GPTQ on Llama-3.2-1B reproduces the SpinQuant paper's Table-8 no-had result (48.89 vs paper 48.4 WikiText-2 PPL on the official released harness).
- Fold equivalence at model scale (external): WikiText-2 PPL of the rotated fp checkpoint matches the original to noise (delta +0.0023% on Qwen3-0.6B).

- Is this change backward compatible?: ✅ (purely additive — a new subpackage and new test files; no existing file's behavior changes. `pyproject.toml`/`.pre-commit-config.yaml` changes are lint-scope and license-hook-exclude entries for the new files only.)
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ (`rotation/sgdg.py` is adapted from the MIT-licensed `stiefel_optimizer.py` of https://github.com/JunLi-Galios/Optimization-on-Stiefel-Manifold-via-Cayley-Transform @ c5ab4e8; the file carries the source link + original MIT notice + NVIDIA header with `SPDX-License-Identifier: Apache-2.0 AND MIT`, `Copyright (c) 2020 Jun Li` is added to the LICENSE third-party MIT notices, and the file is excluded from the insert-license hook. No new pip dependencies. **As an external contributor I am flagging this vendored file here for @NVIDIA/modelopt-setup-codeowners guidance per the third-party-code (OSRB) process in CONTRIBUTING.md** — the internal OSRB registration step is not available to me.)
- Did you write any new necessary tests?: ✅ (77 CPU unit tests, see Testing)
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ (0.47, Quantization)
- Did you get Claude approval on this PR?: ✅ (run `/claude review` after opening)

### Additional Information

- **Relation to `TensorQuantizer`'s existing `rotate` / `RotateConfig`** (e.g. #1883, #1879): complementary, different layer of the stack. The existing `rotate` is an *online, per-quantizer, blockwise* Hadamard applied inside the quantizer forward (NVFP4-oriented; needs rotate-aware export). This module is an *offline, whole-model* checkpoint transform: a global R1 + per-layer R2 folded through norm fusion into the weights, functionally exact, producing a vanilla HF checkpoint that composes with any backend — and, unlike the random-only online path, the rotations here can be *learned* against a fake-quant objective (Cayley SGD). The two can be used together (fold offline R1/R2, then quantize with any rotate-enabled or plain config).
- The module reuses one architecture registry (`_ARCH_REGISTRY` in `fold.py`) for both fold and learn; adding another standard-layout HF decoder is one dict entry.
- Future work: `mtq`-config integration (expose the transform as a pre-quantize step in config), additional architectures, and the online R3/R4 track.

## Summary by CodeRabbit

- **New Features**
  - Added offline SpinQuant/QuaRot rotation folding and Cayley-SGD rotation learning for supported Llama and Qwen3 models.
  - Added seam-diagonal optimization, activation calibration options, and knowledge-distillation support.
  - Added MLflow tracking for supported quantization and serving workflows.
  - Added Megatron-Bridge support for masked distillation, expert-aware quantization, and compiled grouped-linear quantizers.

- **Bug Fixes**
  - Fixed EAGLE-3 context-parallel training startup and tensor-type issues.

- **Documentation**
  - Added rotation workflow, configuration, validation, and deployment guidance.


realAsma:
This PR does not currently provide sufficient experimental evidence to demonstrate that the implementation works correctly. It also introduces additional high-level APIs without following ModelOpt’s configuration-management conventions. As a result, significant rework is needed before the PR can be considered for merging.

Additionally, some of the proposed functionality already exists in ModelOpt. Please review the existing implementation in [config.py](https://github.com/NVIDIA/Model-Optimizer/blob/ddca53b4bf99a1a370a8f3c8fe562710695a3448/modelopt/torch/quantization/config.py#L623).

The main incremental benefit of this PR appears to be support for learnable rotations. Do you have experimental evidence showing that learnable rotations improve results? If so, please share:

* Training and evaluation curves
* Evaluation metrics and comparisons against relevant baselines
* Reproduction details following the workflows in examples/llm_qat and examples/llm_eval

If the primary goal of this PR is to add learnable rotations, the implementation should follow ModelOpt’s existing extension mechanisms by introducing a new calibration algorithm. Please refer to the [custom calibration algorithm documentation](https://nvidia.github.io/Model-Optimizer/guides/_pytorch_quantization.html#custom-calibration-algorithm).

Once the implementation is aligned with ModelOpt’s configuration and calibration architecture—and supported by reproducible experimental results—we can reassess the PR.

modelopt/torch/quantization/rotation/fold.py
```diff
@@ -0,0 +1,615 @@
...
+# --------------------------------------------------------------------------------------
+# Random orthogonal matrix generation (random-sign Hadamard D @ H / sqrt(n), or Haar QR)
+# --------------------------------------------------------------------------------------
+
+# SpinQuant's priority order of hard-coded had-K block sizes; the subset with q = K - 1
+# prime and q ≡ 3 (mod 4) is regenerated via Paley construction I. Any valid Hadamard
+# decomposition yields an orthonormal matrix, which is all that matters for fresh random
+# rotations (bitwise parity with SpinQuant's hard-coded had-K constants is irrelevant).
+_HADK_PRIORITY = [172, 156, 140, 108, 60, 52, 36, 28, 44, 40, 20, 12]
+_HADK_SUPPORTED = {12, 20, 44, 60, 108, 140}
```
cjluo-nv, line 47: > _Bot comment._

`_is_pow2` duplicates `is_pow2` in `modelopt/torch/quantization/utils/core_utils.py` (exported via `modelopt.torch.quantization.utils`). Please import the existing one.

More broadly, `_get_hadK`/`_matmul_hadU` here and `_walsh_hadamard` in `learn.py` re-derive Hadamard machinery that already exists in `modelopt/torch/quantization/nn/functional.py` (`normalized_hadamard_transform` + `_largest_pow2_divisor`). I understand the fold needs a dense fp64 CPU matrix rather than the `fast_hadamard_transform` kernel, but that should be stated in a comment, and the pow2/K-decomposition logic can still be shared instead of forked.

modelopt/torch/quantization/rotation/fold.py
```diff
@@ -0,0 +1,615 @@
...
+
+def _rotate_output_rows(linear: nn.Module, R1: torch.Tensor) -> None:
+    """Writer rotation (o_proj / down_proj): ``W <- R1^T @ W``; bias ``b <- R1^T b``."""
+    w = linear.weight
+    R = R1.to(w.device)
+    w.data = (R.T @ w.data.to(torch.float64)).to(w.dtype)
+    if linear.bias is not None:
+        b = linear.bias
+        b.data = (R.T @ b.data.to(torch.float64)).to(b.dtype)
+
```
cjluo-nv, line 300: > _Bot comment._

The registry keys on the exact model class name, so `MistralForCausalLM`, `Qwen2ForCausalLM`, `NemotronForCausalLM`, … all raise `NotImplementedError` even though their layout is byte-for-byte the standard one this code handles (`model.model.{embed_tokens,layers,norm}` + `model.lm_head`, RMSNorm, `input_layernorm`/`post_attention_layernorm`). Since every registry value except `has_qk_norm`/`head_dim` is identical, consider structural detection (probe for the submodules + RMSNorm, resolve `head_dim` from `config.head_dim` with a hard failure if absent, detect `q_norm`/`k_norm` by attribute presence) with the class-name registry only as an override/deny-list. As written, adding each new architecture is a code change in a core library file rather than a config.

modelopt/torch/quantization/rotation/learn.py
```diff
@@ -0,0 +1,1115 @@
...
+    return x + (dq - x).detach()  # STE
+
+
+def _walsh_hadamard(n: int, device=None) -> torch.Tensor:
+    """Normalized Sylvester Walsh-Hadamard matrix ``[n, n]`` (fp32): symmetric, ``H Hᵀ = I``.
+
+    The fixed (not random-signed) Hadamard the official online-R4 op applies
+    (``matmul_hadU``). Power-of-2 sizes only — the R4 seam dimension is
+    ``config.intermediate_size`` (Llama-3.2-1B: 8192 = 2^13); non-power-of-2 seams
+    (e.g. Qwen3's 6144 = 3·2048) need the had-K Kronecker composition, not implemented.
```
cjluo-nv, line 275: > _Bot comment._

`_FakeQuantSTE` / `_fq_weight` / `_fq_act` / `_fq_act_asym` re-implement quantization numerics that `TensorQuantizer` already covers: per-group (`block_sizes={-1: g}`), per-out-channel (`axis=0`), per-token dynamic (`block_sizes={-1: None, "type": "dynamic"}`), per-tensor static max calibration, affine/asymmetric (`bias`), and STE backward (`pass_through_bwd`). The stated value of this module is that the rotation is trained against *what deployment will actually do*; hand-rolled numerics can silently drift from `TensorQuantizer`'s (rounding mode, scale clamping, amax reduction, FP formats) with no test tying the two together. Please either build the objective on `TensorQuantizer` or, if that's impractical inside the reparametrized assembly, add tests that pin these helpers against the corresponding `TensorQuantizer` config so drift fails CI.

modelopt/torch/quantization/rotation/learn.py
```diff
@@ -0,0 +1,1115 @@
...
+            raise ValueError(
+                f"attention_mask shape {tuple(mask.shape)} != input_ids shape {tuple(ids.shape)}"
+            )
+    return ids, mask
+
+
+# --------------------------------------------------------------------------------------
+# Public API
+# --------------------------------------------------------------------------------------
+
```
cjluo-nv, line 660: > _Bot comment._

`from torch.nn.utils import stateless as _stateless` is a function-local import with no stated reason (no circular import, not an optional/heavy dep) — please move it to the top of the module per project convention.

Separately: `stateless._reparametrize_module` is a **private** torch API. The inline comment explains why `torch.func.functional_call` was rejected, but relying on a private symbol in a shipped library means a torch point release can break `learn_rotations` with no deprecation path. Please add a guarded import with a clear error message (or a public-API fallback) so the failure mode is diagnosable.

modelopt/torch/quantization/rotation/learn.py
```diff
@@ -0,0 +1,1115 @@
...
+    def _r1_ortho() -> float:
+        Rd = R1.detach().to(torch.float64)
+        eye = torch.eye(Rd.shape[0], dtype=torch.float64, device=Rd.device)
+        return (Rd.t() @ Rd - eye).abs().max().item()
+
+    # 6. Training loop.
+    try:
+        if hooks is not None:
+            n_hooked = hooks.attach(model)
+            assert n_hooked == expected_hooks, (
```
cjluo-nv, line 940: > _Bot comment._

The KD term (temperature-scaled `kl_div` + `(1-alpha)*CE + alpha*T^2*KL` mix + frozen teacher forward under `no_grad`) duplicates functionality in `modelopt.torch.distill`. Please justify in the PR body why the existing distillation subsystem can't provide the loss here (e.g. mode/DistillationModel wrapping being incompatible with the stateless reparametrization), or reuse it.

modelopt/torch/quantization/rotation/sgdg.py
```diff
@@ -0,0 +1,219 @@
...
+            "omega": 0,
+            "grad_clip": grad_clip,
+        }
+        if nesterov and (momentum <= 0 or dampening != 0):
+            raise ValueError("Nesterov momentum requires a momentum and zero dampening")
+        super().__init__(params, defaults)
+
+    def __setstate__(self, state) -> None:
+        """Restore optimizer state, defaulting ``nesterov`` for groups pickled without it."""
+        super().__setstate__(state)
```
cjluo-nv, line 150: > _Bot comment._

Two constructor arguments are accepted but never take effect:

- `omega` is captured in the signature yet `defaults` hardcodes `"omega": 0`, so a caller-supplied value is silently discarded.
- `grad_clip` is stored in `defaults` but `step()` never reads it — the upstream file applies gradient clipping in the stiefel branch, and the port note lists only three deltas, none of which mention dropping it.

Either wire both up (and document `grad_clip` in the port note) or remove them from the public signature. A silently-ignored `grad_clip` on an optimizer used at lr=1.5 is a real footgun.

modelopt/torch/quantization/rotation/sgdg.py
```diff
@@ -0,0 +1,219 @@
...
+                    MX = torch.mm(V, unity)
+                    XMX = torch.mm(unity, MX)
+                    XXMX = torch.mm(unity.t(), XMX)
+                    W_hat = MX - 0.5 * XXMX
+                    W = W_hat - W_hat.t()
+                    t = 0.5 * 2 / (_matrix_norm_one(W) + _EPSILON)
+                    alpha = min(t, lr)
+
+                    p_new = _cayley_loop(unity.t(), W, V, alpha)
+                    V_new = torch.mm(W, unity.t())  # n-by-p
```
cjluo-nv, line 195: > _Bot comment._

Faithfully reproducing the upstream dead store means `momentum` is a no-op on the stiefel branch, which is fine for trajectory parity but dangerous for a symbol you export publicly (`SGDG` is in `learn.__all__`). Please make the footgun loud: raise (or `warnings.warn`) when `stiefel=True` and `momentum != 0`, so a user who tunes momentum doesn't get bitwise-identical results and conclude the knob is simply insensitive. The docstring note alone won't reach them.

tests/unit/torch/quantization/test_rotation_contracts.py
```diff
@@ -0,0 +1,382 @@
...
+"""Input-contract and state-hygiene regression tests for the rotation module.
+
+Each test pins one contract whose violation is SILENT (wrong numbers, a mutated caller
+model, a dropped user input) rather than loud, i.e. exactly the class of defect the rest
+of the suite's equivalence gates cannot catch.
+"""
+
+import os
+
+os.environ["CUDA_VISIBLE_DEVICES"] = ""  # CPU-only unit tests: never claim a GPU
```
cjluo-nv, line 25: > _Bot comment._

`os.environ["CUDA_VISIBLE_DEVICES"] = ""` executed at *module import* is a process-wide side effect: pytest imports all test modules into one process, so this mutates the environment for every other test collected in the same session (and the effect depends on collection order / whether CUDA was already initialized). Same pattern appears in `test_rotation_ext_fold.py`, `test_rotation_ext_learner.py`, `test_rotation_ext_sgdg.py`, `test_rotation_transform_qat.py`.

These tests are already CPU-only by construction (tiny CPU-built models, no `.cuda()`), so the simplest fix is to delete the env mutation. If a hard guarantee is wanted, use a fixture with `monkeypatch.setenv` so it's scoped and reverted.

tests/unit/torch/quantization/test_rotation_fold.py
```diff
@@ -0,0 +1,259 @@
...
+# magnitude ~0.6; atol=1e-4 keeps >100x headroom (also for deeper/bigger real models where
+# the cast noise accumulates) while still catching any real transform bug — a wrong rotation
+# orientation or a missed norm fusion perturbs logits by O(1).
+ATOL_FP32 = 1e-4
+
+
+def _randomize_rmsnorm_gains(model):
+    """Set every RMSNorm gain to a random non-one value. Fresh HF models initialize all norm
+    weights to ones, which would make both the norm-fusion math and the fused-to-ones /
+    q_norm-untouched checks vacuous."""
```
cjluo-nv, line 55: > _Bot comment._

`_tiny_llama` / `_tiny_qwen3` / `_randomize_rmsnorm_gains` / `_calib_batches` / `_logits` are re-defined in essentially identical form across all 8 new test files (~400 duplicated lines total). The repo already has shared builders in `tests/_test_utils/torch/transformers_models.py` (`get_tiny_llama`, `get_tiny_qwen3`, both accepting `**config_kwargs` so `head_dim`/`num_key_value_heads`/`tie_word_embeddings` can be overridden) — please build on those and put the rotation-specific bits (non-unit RMSNorm gains, calib batches, logits helper) in one shared module or a conftest.

Also note that module gates `transformers` behind `pytest.importorskip("transformers")`; these files import `transformers` at top level, so collection hard-fails in an environment without the `hf` extra.

tests/unit/torch/quantization/test_rotation_ext_fold.py
```diff
@@ -0,0 +1,619 @@
...
+        return v * torch.rsqrt(v.pow(2).mean(-1, keepdim=True) + eps)
+
+    diff = (rms(x_rot) - rms(x) @ R1).abs().max().item()
+    assert diff < 1e-12, f"RMSNorm rotation-invariance broken: max |delta| = {diff:.3e}"
+
+    # -------- end-to-end: run both actual models on the same ids --------
+    ref_logits, rot_logits = _logits(model_ref, vocab), _logits(model_rot, vocab)
+    max_diff = (rot_logits - ref_logits).abs().max().item()
+    assert torch.allclose(rot_logits, ref_logits, rtol=0, atol=ATOL_FP32), (
+        f"oracle end-to-end: max |delta logit| = {max_diff:.3e} > {ATOL_FP32}"
```
cjluo-nv, line 600: > _Bot comment._

The `if __name__ == "__main__":` self-driver is repeated in all 8 test files and is what forces the new `PERF203` lint exception in `pyproject.toml`. pytest is the project's harness (`pytest tests/unit/torch/quantization/test_rotation_fold.py` already does this, with proper parametrize support — the contracts driver even has to *skip* its parametrized tests). Please drop the drivers and the associated per-file lint ignore; that also removes ~250 lines from an already very large PR.

cjluo-nv (commented):

## Design review (protocol applied — 6,026 LOC, new subsystem)

**Problem restated:** make HF RMSNorm decoder LMs easier to quantize by folding a global R1 + per-layer R2 orthogonal rotation into the checkpoint (QuaRot-style), and optionally *learning* those rotations against a fake-quant objective (SpinQuant).

**Alternatives that already exist in this repo / in installed deps:**
1. **`QuantizerAttributeConfig.rotate` / `RotateConfig` + `normalized_hadamard_transform`** (`config.py`, `nn/functional.py`) — the existing in-repo rotation path. ✅ The PR body *does* address this one (online-per-quantizer vs. offline-checkpoint, learnable), and the argument is reasonable.
2. **The calibration-algorithm framework** (`QuantizeAlgorithmConfig` subclasses + `model_calib.py`): SmoothQuant, AWQ, SVDQuant, GPTQ, LSQ are all pre-quantization weight/scale transforms — including `svdquant`/`smoothquant`, which do exactly the "rewrite weights + companion scales before quantize" job that `fold_rotations`/`fold_seam_diags` do, and LSQ, which already does gradient-based learning of quantization parameters. This PR adds a second, parallel mechanism (two module-level functions, own objective dataclass, own optimizer, own training loop) with no `mtq` config/mode integration. The PR body says integration is "future work" but never says *why the existing algorithm/mode framework can't host this*.
3. **`TensorQuantizer` for the training objective**: `_FakeQuantSTE`, `_fq_weight`, `_fq_act`, `_fq_act_asym` re-implement per-group/per-channel/per-token/per-tensor-static symmetric and asymmetric QDQ with STE — all of which `TensorQuantizer` already provides (`block_sizes` incl. `{-1: None, "type": "dynamic"}` per-token dynamic, `axis`, `bias` affine, `pass_through_bwd` STE). A rotation objective built on `TensorQuantizer` would automatically stay in sync with deployment numerics; the hand-rolled version can drift from it silently.
4. **`modelopt.torch.distill`**: the KD term (`kl_div` + temperature + alpha mix + frozen teacher) is re-implemented inline in the training loop although the repo ships a whole distillation subsystem.
5. **Hadamard/pow2 helpers**: `_is_pow2` duplicates `modelopt.torch.quantization.utils.is_pow2`; `_matmul_hadU`/`_walsh_hadamard` duplicate `normalized_hadamard_transform` (+ `_largest_pow2_divisor`) modulo the fp64-CPU-matrix vs. kernel difference.

Per the protocol, **the design question is only partially addressed** (alternative 1 only), so this can't be approved as-is. Please extend the PR body with an explicit "why not X" for #2 (algorithm/mode framework), #3 (`TensorQuantizer` in the loss) and #4 (`mtd`); if the answer is "the framework can't express a whole-model checkpoint transform", say so — that's a legitimate answer, it just needs to be on record.

## Licensing (blocking human sign-off)
`rotation/sgdg.py` vendors MIT-licensed third-party code (Li et al.), `LICENSE` gains a third-party copyright holder, and a license-hook exclusion is added. The author has correctly flagged this for OSRB; a maintainer must sign off — I can't approve a PR that touches licensing.

## Size
6,026 added lines in one PR is well past the ~1000-line guideline and this *is* splittable along clean seams: (a) `fold.py` + fold tests, (b) `learn.py` + `sgdg.py` (vendored file isolated for OSRB) + learner tests, (c) transform-QAT seam diagonals + `fold_seam_diags`, (d) paper-protocol objectives (`a_asym`, `r4_in_graph`). Each lands independently and (a) is by far the most reviewable/valuable first step.

## Test coverage
Very thorough at the unit level (equivalence, orientation oracle, SGDG numerics, contracts). Two gaps:
- **No test of the headline claim** that "every existing quant config, calibrator, exporter works unchanged" — nothing runs `mtq.quantize(rotated_model, INT8_DEFAULT_CFG, forward_loop)` or an export on a folded model. That's the integration contract the whole design rests on; please add at least one such CPU test.
- No example/entrypoint and no GPU test, so the learner is only exercised on 64-hidden toys.

Correctness of the math itself looks right to me (GQA grouping, reader/writer orientation, RoPE-safety of R1 on q/k inputs, q_norm/k_norm exclusion, seam-diagonal identity), and the invariant asserts are a nice touch. Individual findings inline.

PR #2188, opened 2026-08-13
Pin GDPVal evaluator launcher to 0.2.6

Edwardf0t1:
### What does this PR do?

Type of change: Bug fix

Pins GDPVal evaluation commands to the currently validated
`nemo-evaluator-launcher==0.2.6` through a dedicated `uvx` wrapper and updates the
GDPVal skill, recipe, example, and reference documentation to use it. The reference
also documents how to validate and adopt a future launcher release without mixing
harness versions in baseline-versus-candidate comparisons.

Older launchers generate a runtime re-export for `NEL_INVOCATION_ID` without first
assigning the launcher-created invocation ID. The resulting Slurm script exits under
`set -u` with `NEL_INVOCATION_ID: unbound variable` before the evaluation client
starts. Launcher 0.2.6 writes the stable invocation ID into the script before the
re-export, including for walltime-resume chains.

### Usage

```bash
"$SKILL_DIR/scripts/nel-gdpval.sh" --version
"$SKILL_DIR/scripts/nel-gdpval.sh" run --config gdpval.yaml --dry-run
"$SKILL_DIR/scripts/nel-gdpval.sh" run --config gdpval.yaml
```

### Testing

- `bash -n plugins/modelopt/skills/evaluation/scripts/nel-gdpval.sh`
- Verified the wrapper reports `nemo_evaluator_launcher: 0.2.6`.
- Ran a real GDPVal dry-run and verified the generated script assigns
  `NEL_INVOCATION_ID="<invocation-id>"` before the runtime re-export.
- Added a hermetic shell-wrapper test with a stub `uvx`; a conflicting ambient
  conflicting ambient version/spec values still invoke the validated 0.2.6 package.
- `python -m pytest plugins/modelopt/skills/evaluation/ -o addopts="" -p no:cacheprovider -v`
- Ran pre-commit on all six changed files.

Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md)
and your commits are signed (`git commit -s -S`).

- If you copied code from any other sources or added a new PIP dependency, did you
  follow guidance in `CONTRIBUTING.md`: N/A
- Did you write any new necessary tests?: ✅ — added a hermetic wrapper test that
  stubs `uvx` and verifies the exact package spec despite a conflicting environment value.

### Additional Information

The fix preserves a stable GDPVal cache namespace across Slurm walltime resumes;
using `SLURM_JOB_ID` would not.

The failing run reused an existing base-environment `nel` 0.2.4 because the
pre-existing prerequisite checked only whether `nel` was installed. The skill
already identified 0.2.6 as the default launcher for Steps 1–9 and GDPVal. Version
0.2.6 is also the latest public `nemo-evaluator-launcher` release, so there is no
newer public launcher version to validate.

A dedicated `uvx` wrapper keeps the fix narrow and deterministic:

- `nel-next.sh` manages the separate `nemo-evaluator` 0.4.x package, CLI, and
  schema; generalizing it would couple incompatible workflows.
- A skill-level lock/project adds state for a single exact package invocation,
  while `uvx --from ...==0.2.6` already provides an isolated cached environment.
- A version preflight would only reject the stale shared environment or require
  mutating it; the wrapper guarantees the known-good launcher on every GDPVal
  invocation without changing the existing `nel`.

## Summary by CodeRabbit

- **New Features**
  - Added a GDPVal launcher wrapper that consistently uses the validated launcher version and forwards commands.
  - Added help output and checks for required runtime tooling.

- **Documentation**
  - Clarified GDPVal launch, dry-run, reproducibility, resumed-run, and invocation ID guidance.
  - Documented version requirements and known sample limitations.
  - Updated evaluation examples and task instructions with the new launch procedure.

- **Tests**
  - Added coverage verifying the pinned launcher version and argument forwarding.

### DIFF
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -34,8 +34,8 @@ If `MODELOPT_WORKSPACE_ROOT` is set, use the common skill's `workspace-managemen
 
 ### nel-next path (Terminal-Bench 2.x, SWE-bench, …) — branch here FIRST
 
-A few **agentic** AA benchmarks do **not** run on the default
-`nemo-evaluator-launcher` 0.2.6 (Steps 1–9 don't apply). They run on **nel-next**
+A few **agentic** AA benchmarks do **not** run on the currently validated
+`nemo-evaluator-launcher` 0.2.6 path (Steps 1–9 don't apply). They run on **nel-next**
 (`nemo-evaluator[harbor]` 0.4.x) — a separate package, CLI (`nel eval run`), `-O`
 overrides, and `services`/`benchmarks`/`cluster`/`output` schema. If the user asks
 for one, do **not** add it to a 0.2.6 `evaluation.tasks` list — instead:
@@ -45,16 +45,16 @@ for one, do **not** add it to a 0.2.6 `evaluation.tasks` list — instead:
 3. Run **`modelopttools:eval-config`** (Step 3b) to write the AWS-sandbox creds + harbor infra rows (`${NEL_NEXT_EVAL_IMAGE}`, `${HARBOR_*_ECR_REPOSITORY}`) into `.env`; always include the `output.export_config.mlflow` block.
 4. Dry-run → canary → full (`nel-next.sh eval run`), then **push to MLflow** — SLURM doesn't auto-export, so run `nel-next.sh mlflow-push -r <run_id> -c <cfg>` after (config-driven; see `references/nel-next.md`).
 
-Steps 1–9 below are the 0.2.6 path — use them for everything else.
+Steps 1–9 below are currently validated with 0.2.6 — use them for everything else.
 
 ---
 
 ### GDPVal (NeMo Gym "Stirrup" agent) path — branch here too
 
-GDPVal **does** run on the 0.2.6 `nel` launcher (as a `nemo_gym` task, not
-nel-next), so Steps 1–9 apply — but it is mechanically special and **standalone**
-(one gym eval per config; never mix it with `aa/` tasks). If the user asks for
-GDPVal:
+GDPVal **does** run on the currently validated 0.2.6 `nel` launcher (as a
+`nemo_gym` task, not nel-next), so Steps 1–9 apply — but it is mechanically
+special and **standalone** (one gym eval per config; never mix it with `aa/`
+tasks). If the user asks for GDPVal:
 
 1. Read **`references/gym-gdpval.md`** (Apptainer SIF sandbox, gym prepare/reap
    machinery, deploy sizing, rubric-vs-comparison scoring, MLflow deliverables trap,
@@ -71,8 +71,8 @@ GDPVal:
    `INFERENCE_JUDGE_URL`, `GDPVAL_SIF_DIR`, and `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the
    config has a `pre_cmd`). Thinking mode is mandatory (non-thinking loses ~86%).
 4. Run both dry-run and launch through `"$SKILL_DIR/scripts/nel-gdpval.sh"`; it
-   enforces the required 0.2.6 launcher even if `nel` on PATH is stale and avoids
-   the observed 0.2.4 launcher's unset `NEL_INVOCATION_ID` failure.
+   enforces the currently validated 0.2.6 launcher even if `nel` on PATH is stale
+   and avoids an unset `NEL_INVOCATION_ID` failure before client startup.
    **`limit_samples` is inert on the gym path** (the gym runs all 220 tasks
    regardless), so there is no cheap canary: watch the real run's first
    ~20–30 min for the SIF-sandbox line and judge auth, and cancel if wrong. See the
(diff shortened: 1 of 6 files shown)


kevalmorabia97 (approved):

plugins/modelopt/skills/evaluation/scripts/nel-gdpval.sh
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# GDPVal requires launcher 0.2.6. In particular, 0.2.6 assigns the generated
+# NEL_INVOCATION_ID in run.sub before evaluation.env_vars re-exports it. Older
+# launchers can instead fail under `set -u` before the evaluation client starts.
+set -euo pipefail
+
+NEL_GDPVAL_SPEC="${NEL_GDPVAL_SPEC:-nemo-evaluator-launcher[all]==0.2.6}"
```
cjluo-nv: > _Bot comment._

The wrapper and docs promise that GDPVal is pinned to 0.2.6 for correctness, but this ambient-env override silently defeats that guarantee. In particular, `set -a; source .env` is part of the documented run flow, so a stale `NEL_GDPVAL_SPEC` can select exactly the older launcher that produces the unbound-variable failure. Unlike the documented escape hatch in `nel-next.sh`, there does not seem to be a valid compatibility path here because the generated-script behavior specifically requires 0.2.6. Please hard-code the spec (ideally `readonly`), or reject anything that is not the exact required release. Please also cover this with a shell test using a stub `uvx`, including a conflicting ambient value, since the exact pin is the core behavior of this fix.

cjluo-nv (commented):

The GDPVal failure is clear: launcher versions older than 0.2.6 can generate a Slurm script that re-exports `NEL_INVOCATION_ID` before assigning it. The proposed solution is a dedicated `uvx` launcher wrapper. Existing alternatives include extending/generalizing the repository's existing `scripts/nel-next.sh` isolation wrapper, using a project/skill-level uv lock and `uv run`, or retaining the normal launcher with an exact-version preflight. The PR body explains why 0.2.6 is required, but does not compare the dedicated second wrapper with those alternatives, so the required design gate is not yet satisfied.

There is also a correctness hole in the pin itself: an ambient or sourced `NEL_GDPVAL_SPEC` silently bypasses 0.2.6, including selecting the older launcher this change is meant to prevent. Please make the required version non-overridable (or strictly validate the override) and add an automated wrapper test that stubs `uvx` and verifies the exact install spec even with a conflicting environment. The new file's NVIDIA Apache-2.0 header matches `LICENSE_HEADER`, so it does not require separate licensing escalation.

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -70,7 +70,9 @@ GDPVal:
    unsandboxed. Verify with `gdpval-sif.sh --check`. `.env` needs `HF_TOKEN`, `INFERENCE_API_KEY`, `TAVILY_API_KEY`,
    `INFERENCE_JUDGE_URL`, `GDPVAL_SIF_DIR`, and `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the
    config has a `pre_cmd`). Thinking mode is mandatory (non-thinking loses ~86%).
-4. Dry-run → launch. **`limit_samples` is inert on the gym path** (the gym runs all
+4. Run both dry-run and launch through `"$SKILL_DIR/scripts/nel-gdpval.sh"`; it pins
+   the required 0.2.6 launcher and avoids older launchers' unset
```
chadvoegele: Why did agent install an outdated version of nel in the first place? Does newer than 0.2.6 work?

cjluo-nv: same question

plugins/modelopt/skills/evaluation/scripts/nel-gdpval.sh
Edwardf0t1: Addressed in 938abd2297. `NEL_GDPVAL_SPEC` is now a hard-coded readonly `nemo-evaluator-launcher[all]==0.2.6`, so values sourced from `.env` cannot bypass the correctness pin. I also added a hermetic wrapper test that puts a stub `uvx` on PATH, injects a conflicting `NEL_GDPVAL_SPEC=...0.2.4`, and asserts that the wrapper still forwards the exact 0.2.6 spec. The focused skill pytest and pre-commit checks pass.

plugins/modelopt/skills/evaluation/SKILL.md
Edwardf0t1: Good question. The failing GDPVal attempt did not deliberately install 0.2.4; it reused an existing base-environment `nel` 0.2.4 because the pre-existing Step 1 only checked whether `nel` was present. I verified on the PR base commit that `SKILL.md` already identifies `nemo-evaluator-launcher` 0.2.6 as the default for Steps 1–9 and explicitly places GDPVal on that launcher. As of this review, 0.2.6 is the latest public `nemo-evaluator-launcher` release, so there is no newer public launcher version to test. The `nemo-evaluator` 0.3.x/0.4.x nel-next workflow is a separate package, CLI, and schema rather than a newer compatible launcher. I updated the docs to name the observed failing version precisely and added the design comparison to the PR description.

PR #2190, opened 2026-08-14
Add non-mutating quantized weight export API

mxinO:
## Summary

- add a public, non-mutating API that exports quantized deployment tensors and a ModelOpt deployment descriptor from an initialized model
- share the packing and metadata path with terminal HF export instead of duplicating format-specific conversion logic
- preserve per-expert amax values when grouped quantizers are temporarily restored for export

This API is intended for in-process refit users such as Megatron-Bridge and NeMo-RL. It keeps the training model unchanged while producing the same canonical weight, scale, and input-scale tensors as terminal export.

## Testing

- targeted export unit tests
- Transformer Engine grouped-quantizer GPU regression test
- Ruff check and format check
- downstream two-rank TP2/PP2/EP2 parity test comparing the API with terminal export

## Summary by CodeRabbit

* **New Features**
  * Added support for exporting NVFP4 and mixed-precision quantized weights with packed weights, scaling data, activation scaling, and Hugging Face quantization metadata.
  * Improved handling of tied weights and mixed-precision configurations.
  * Added validation for unsupported formats, missing calibration data, invalid scales, and block sizes.

* **Bug Fixes**
  * Preserved grouped quantizer calibration values during model restoration.
  * Improved consistency, repeatability, and state preservation when exporting quantized weights.

### DIFF
--- a/modelopt/torch/export/__init__.py
+++ b/modelopt/torch/export/__init__.py
@@ -21,6 +21,7 @@ from .model_config_export import *
 from .model_utils import *
 from .moe_utils import *
 from .plugins import *
+from .quantized_weight import *
 from .registry import *
 from .transformer_engine import *
 from .unified_export_hf import *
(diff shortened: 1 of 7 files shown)


mxinO:
Closing this draft while the real-quant refit design is reworked and validated locally. The branch is retained for reference.

PR #2192, opened 2026-08-14
docs(eval-skill): add MRCR (NeMo Gym) benchmark

cjluo-nv:
### What does this PR do?

Adds **MRCR** — OpenAI's Multi-Round Co-reference Resolution, a long-context
retrieval benchmark — to the `evaluation` skill as a standalone NeMo Gym task,
derived from the reviewed `nemotron_nano_v35_nvfp4_mrcr_gym` golden; regroups the
gym tasks/examples under one `gym/` dir; and fixes several latent bugs in the
shared gym command block that were found by running the benchmark end-to-end.

MRCR tasks are long multi-turn conversations containing N near-identical "needle"
responses; the model must reproduce the Nth verbatim behind a random prefix.
Grading is deterministic (`SequenceMatcher.ratio()`, gated on the prefix). Unlike
GDPVal it uses the `simple_agent`: no SIF, no judge, no Tavily — `HF_TOKEN` is the
only secret, and the cost is context length (up to 1M tokens), not agent turns.
**It is not an AA benchmark** and is never generated for an "AA" request.

### Layout

| File | |
| --- | --- |
| `recipes/tasks/gym/mrcr.md` | new recipe — variants, 1M serving envelope, canary, score extraction |
| `recipes/examples/gym/example_mrcr.yaml` | new self-contained SLURM + vLLM config |
| `SKILL.md` | MRCR branch + gym index table |
| `references/quantization-benchmarks.md` | table row + comparability notes |

`examples/gym_{gdpval,mrcr}/` → `examples/gym/example_<task>.yaml` and
`tasks/aa_gym/gdpval.md` → `tasks/gym/gdpval.md`, all tracked as git renames.
`aa_gym` encoded "in the AA suite" in the *path*; since GDPVal is AA and MRCR is
not, membership is now stated explicitly in both recipe headers and an "In AA
suite?" column in the SKILL.md index. `gym/` groups by harness, not suite.

### Bugs fixed (found by running it, not by reading it)

The first three are in the **shared** gym command block, so they also affect
`example_gdpval.yaml` — GDPVal was broken independently of MRCR.

1. **Invalid OmegaConf interpolation.** A literal `${...}` inside a comment in the
   gym `command:` block is parsed as an interpolation and rejected, so *every*
   `nel run --dry-run` of either gym template died with
   `hydra.errors.ConfigCompositionException`.
2. **Hardcoded `ray==2.49.2`** injected into each sub-server's requirements. Against
   an image carrying `ray[default]==2.55.1` this makes `uv` unsatisfiable and the
   gym resources server exits at startup. Now derived from the image at runtime.
3. **`--max-num-seqs` sized against the wrong topology.** The template shipped DP1
   → 4 replicas at 64 concurrent while citing a golden that is TP2×DP2 → 8 replicas
   at 32. Following it ran double the reference's per-replica load, which on
   1M-token prompts is what decides whether the KV cache fits.
4. **Score extraction pointed at an empty map.** `results.yml` →
   `groups.nemo_gym.metrics` holds only `key_metrics/mean/*` telemetry; the scores
   are in `artifacts/evaluator_rollouts_aggregate_metrics.json` → `[0].agent_metrics`.
   Also documents that `pass@1/accuracy` is already 0-100 while `mean/reward` is the
   same number as a 0-1 fraction, and records `mean/prefix_matched ≈ 0.55` as the
   healthy calibration.
5. **The template shipped a container its own bootstrap rejects.** After adding the
   hard-fail on an unpinned Gym, `container:` still defaulted to public
   `nemo-gym:26.05` — the image the docs say has a non-git `/opt/Gym`. Now `???`,
   so `--dry-run`'s mandatory-value check catches it instead of the job dying at
   startup.

### Review feedback

All items from @meenchen and CodeRabbit addressed or answered inline. Two worth
surfacing here:

- **`process_reasoning_traces` vs `use_reasoning`** — verified against
  `nemo_evaluator/adapters/adapter_config.py`: both exist, `use_reasoning` is the
  **deprecated** one and they are bidirectionally aliased. Documented rather than
  switched to the deprecated name.
- **`tiktoken` / `transformers` left unpinned** — deliberate. The n3 prepare path
  uses `transformers.AutoTokenizer` to decide which samples exceed the cap, so a
  bump can shift dataset membership; but pinning would diverge from the golden's
  `pre_cmd` and therefore from the run that produced the reference number. Risk is
  now documented under "Deferred, know the risk" instead of being implicit.

Topology, `gres` and TP/DP were aligned to the **existing** sibling templates
rather than to a new convention: `gres` stays a comment (already the convention in
`example_eval.yaml` and `example_gdpval.yaml`), TP is a concrete `1` like both
siblings, and `num_nodes`/`num_instances`/`--max-num-seqs` are guidance rather than
baked values. `--max-num-seqs` sizing follows **AA-LCR**, since MRCR is the same
KV-bound problem at ~1M tokens vs LCR's ~120K: the formula gives a ceiling, not a
target, because oversubscribing causes preemption and recomputing a 1M-token
prefill makes the run slower.

### Usage

```bash
cp plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml mrcr.yaml
# fill checkpoint_path / served_model_name / container / SLURM ??? values
export NEMO_EVALUATOR_TRUST_PRE_CMD=1 NEMO_EVALUATOR_TRUST_UNLISTED_TASKS=1
nel run --config mrcr.yaml --dry-run && nel run --config mrcr.yaml
```

### Testing

Docs/config-only; no library code touched. `pre-commit` clean on every commit.

- Both gym YAMLs parse; asserted the folded-scalar rule (no `#` inside `>-`), that
  every string survives `OmegaConf.create` (bug 1), and that the variant is
  identical in `data_prep_params` and `collect_rollout_params`.
- After the rename: zero stale `aa_gym`/`gym_gdpval`/`gym_mrcr` refs and every
  `recipes/**` path referenced across both skill trees resolves on disk — this
  caught `env.example`, which an extension-filtered grep missed. The
  `nemo_gym_gdpval_stirrup_agent` metric names contain the substring `gym_gdpval`
  and were deliberately left untouched.
- **Ran the benchmark end-to-end.** A full run on gcp-nrt (B200) completed
  2363/2363 rollouts and scored; that run produced bugs 2 and 4 and the corrected
  metric paths. A second attempt on aws-cmh was preempted and later cancelled.
  `--dry-run` now passes (it did not before bug 1 was fixed), and the bare template
  correctly fails validation on unresolved mandatory values.
- Regenerated the config from scratch with a fresh agent against the fixed
  templates as a regression check; its dry-run passed and its findings drove the
  topology/`gres`/TP-DP alignment above.

Model-specific scores are deliberately **not** recorded in the recipe — the
reference there stays the golden's BF16 Nano 3.5 shape, since quoting a different
model's number beside it invites exactly the false comparison the recipe warns
about.

- Backward compatible?: ✅ — additive plus a doc-tree rename; all referencing files
  updated and verified. The template fixes change only broken behaviour.
- Copied code / new PIP dependency: N/A.
- New tests: N/A — agent-skill documentation.
- Changelog: N/A — skill/docs change; consistent with prior skill-only commits.
- Claude approval: ❌ not yet — will trigger `/claude review`.

### Additional Information

Upstream drift flagged to reviewers (**no change made to that repo**): in
`nvidia-eval-factory-benchmarking`, `configs/benchmarks/mrcr/bench.yaml` is still
old-style `ng_*` while `configs/models/nemotron_nano_v35/gym.yaml` moved to the
`gym eval` CLI on 2026-07-29 — RULER migrated, MRCR did not, so the layers no
longer compose. This template follows `ng_*`, which is what MRCR's own bench.yaml
specifies and what the sign-off run executed.

NVIDIA-internal companion (container specifics kept out of this public tree):
Model-Optimizer-Internal MR !116.

### DIFF
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -94,7 +94,8 @@ for an "AA" request. If the user asks for MRCR:
    comparable; set it in **both** `data_prep_params` and `collect_rollout_params`.
 3. `.env`: `HF_TOKEN` (dataset + n3 tokenizer are gated) plus
    `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the `pre_cmd` installs `tiktoken` +
-   `transformers`; prepare fails without it).
+   `transformers`; prepare fails without it) and
+   `NEMO_EVALUATOR_TRUST_UNLISTED_TASKS=1` (`nemo_gym` is not in the FDF map).
 4. The Gym pin is **newer than any image's baked Gym** and is silently ignored
    where `/opt/Gym` is not a git repo — verify it applied before quoting a score.
    NVIDIA-internal: `modelopttools:eval-config` Step 3d names a working image.
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -96,8 +96,9 @@ for an "AA" request. If the user asks for MRCR:
    `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the `pre_cmd` installs `tiktoken` +
    `transformers`; prepare fails without it) and
    `NEMO_EVALUATOR_TRUST_UNLISTED_TASKS=1` (`nemo_gym` is not in the FDF map).
-4. The Gym pin is **newer than any image's baked Gym** and is silently ignored
-   where `/opt/Gym` is not a git repo — verify it applied before quoting a score.
+4. **MRCR needs a git-backed Gym image.** The pin is newer than any image's baked
+   Gym and must apply, so the template's `container:` is `???` and the bootstrap
+   exits 1 on a non-git `/opt/Gym` (the public `eval-factory/nemo-gym:*` images).
    NVIDIA-internal: `modelopttools:eval-config` Step 3d names a working image.
 5. Long-context deploy (`--max-model-len 1100000` +
    `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1`, `gpu_memory_utilization: 0.95`,
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -80,6 +80,40 @@ tasks). If the user asks for GDPVal:
 
 ---
 
+### MRCR (NeMo Gym `simple_agent`) path — branch here too
+
+MRCR is a **long-context** gym benchmark and, like GDPVal, a 0.2.6 `nemo_gym`
+task that is **standalone** (one gym eval per config). It is **not** an AA
+benchmark — never generate it as part of an "AA" request. It is far simpler than
+GDPVal: `simple_agent`, **no SIF sandbox, no judge, no Tavily** — deterministic
+prefix-gated `SequenceMatcher` grading, so `HF_TOKEN` is the only secret. If the
+user asks for MRCR:
+
+1. Read **`recipes/tasks/gym/mrcr.md`** (variant table, serving envelope, score
+   extraction, failure modes).
+2. Start from **`recipes/examples/gym_mrcr/example_gym_mrcr.yaml`** — a single
+   self-contained file, targeting the 1M variant like the reviewed golden.
+3. **Pick the variant first** — `config_n3_1m` / `config_n3_128k` / `config`. It
+   sets the context cap, the dataset size, *and* the metric key prefix, and the
+   three are not comparable to each other. Change it in **both**
+   `data_prep_params` and `collect_rollout_params`.
+4. `.env` needs `HF_TOKEN` (the dataset and the n3 tokenizer are both gated) and
+   `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the config has a `pre_cmd` that installs
+   `tiktoken` + `transformers` into the Gym venv — prepare fails without it).
+   The template's Gym pin is **newer than any image's baked Gym**, so it must
+   actually apply: on images where `/opt/Gym` is not a git repo it is silently
+   ignored. NVIDIA-internal: `modelopttools:eval-config` **Step 3d** names a
+   container where pinning works (and how to pull it).
+5. Long-context envelope, not the model, drives the deploy:
+   `--max-model-len 1100000` + `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1`,
+   `gpu_memory_utilization: 0.95`, prefix caching + chunked prefill, and a
+   multi-instance fan-out (golden: 4 nodes / 4 instances). **Never cap output
+   tokens** — answers reproduce a full earlier turn verbatim.
+6. Report the **needle-count strata** (2/4/8) alongside the headline
+   `pass@1/accuracy`; quantization damage shows up in the 8-needle stratum first.
+
+---
+
 ### Step 1 — Prerequisites
 
 Run `nel --version`; if missing, instruct `pip install nemo-evaluator-launcher`. If user has an existing config, skip to Step 8 (optionally review for `???` and quantization flags first).
@@ -94,6 +128,7 @@ Run `nel --version`; if missing, instruct `pip install nemo-evaluator-launcher`.
 - Optional: `recipes/tasks/mmlu_pro.md`, `recipes/tasks/aime_2025.md`, `recipes/tasks/livecodebench.md`
 - **nel-next only** (different evaluator — see the nel-next section below, NOT the 0.2.6 steps): shared reference `references/nel-next.md` + per-benchmark recipes `recipes/tasks/aa_next/{terminal_bench_2_1,swebench_verified}.md` (agentic). The `aa_next/` dir holds tasks that require `nemo-evaluator[harbor]` 0.4.x (the package; `nemo-evaluator-next` is the eval *image* repo); `aa/` is the 0.2.6 suite.
 - **GDPVal (NeMo Gym / agentic)** — **part of the AA suite** but a 0.2.6 `nemo_gym` task on a different harness, so it's **standalone** (see the GDPVal branch above): recipe `recipes/tasks/aa_gym/gdpval.md` + shared reference `references/gym-gdpval.md` + self-contained example `recipes/examples/gym_gdpval/`. Generated as its **own config** from the example, **never merged into the `aa/` multi-task `tasks` list**. The `aa_gym/` dir holds the NeMo Gym Stirrup-agent tasks.
+- **MRCR (NeMo Gym / long-context)** — **NOT an AA benchmark**; a 0.2.6 `nemo_gym` task, also **standalone** (see the MRCR branch above): recipe `recipes/tasks/gym/mrcr.md` + self-contained example `recipes/examples/gym_mrcr/`. The `gym/` dir holds non-AA NeMo Gym tasks; only generate it when the user asks for MRCR by name (or for long-context coverage).
 
 **AA rule:** If the user mentions "AA" / "Artificial Analysis", generate the `recipes/tasks/aa/` tasks (one multi-task config) **plus a companion standalone GDPVal config** (`recipes/tasks/aa_gym/gdpval.md`, via the GDPVal branch) — GDPVal is part of the AA suite but a different harness, so it's its own config, never added to the `aa/` `tasks` list. Do not add MMLU-Pro, AIME 2025, or LiveCodeBench unless explicitly asked. GDPVal is the heaviest AA task (standalone, multi-hour, needs the SIF sandbox + judge) — surface it and let the user opt out per run.
 
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -123,8 +123,8 @@ Run `nel --version`; if missing, instruct `pip install nemo-evaluator-launcher`.
 
   | Task | Recipe / example | In AA suite? | Generate when |
   | --- | --- | --- | --- |
-  | **GDPVal** (Stirrup agent, agentic) | `tasks/gym/gdpval.md` + `references/gym-gdpval.md`, `examples/gym/example_gdpval.yaml` | **Yes** | any AA request (see the AA rule below) |
-  | **MRCR** (simple agent, long-context) | `tasks/gym/mrcr.md`, `examples/gym/example_mrcr.yaml` | **No** | only when the user asks for MRCR by name, or for long-context coverage |
+  | **GDPVal** (Stirrup agent, agentic) | `recipes/tasks/gym/gdpval.md` + `references/gym-gdpval.md`, `recipes/examples/gym/example_gdpval.yaml` | **Yes** | any AA request (see the AA rule below) |
+  | **MRCR** (simple agent, long-context) | `recipes/tasks/gym/mrcr.md`, `recipes/examples/gym/example_mrcr.yaml` | **No** | only when the user asks for MRCR by name, or for long-context coverage |
 
 **AA rule:** If the user mentions "AA" / "Artificial Analysis", generate the `recipes/tasks/aa/` tasks (one multi-task config) **plus a companion standalone GDPVal config** (`recipes/tasks/gym/gdpval.md`, via the GDPVal branch) — GDPVal is part of the AA suite but a different harness, so it's its own config, never added to the `aa/` `tasks` list. Do not add MMLU-Pro, AIME 2025, or LiveCodeBench unless explicitly asked. GDPVal is the heaviest AA task (standalone, multi-hour, needs the SIF sandbox + judge) — surface it and let the user opt out per run.
 
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -82,35 +82,26 @@ tasks). If the user asks for GDPVal:
 
 ### MRCR (NeMo Gym `simple_agent`) path — branch here too
 
-MRCR is a **long-context** gym benchmark and, like GDPVal, a 0.2.6 `nemo_gym`
-task that is **standalone** (one gym eval per config). It is **not** an AA
-benchmark — never generate it as part of an "AA" request. It is far simpler than
-GDPVal: `simple_agent`, **no SIF sandbox, no judge, no Tavily** — deterministic
-prefix-gated `SequenceMatcher` grading, so `HF_TOKEN` is the only secret. If the
-user asks for MRCR:
+A 0.2.6 `nemo_gym` task like GDPVal and equally **standalone**, but far simpler:
+`simple_agent`, **no SIF, no judge, no Tavily** — deterministic prefix-gated
+grading, `HF_TOKEN` the only secret. **Not an AA benchmark** — never generate it
+for an "AA" request. If the user asks for MRCR:
 
-1. Read **`recipes/tasks/gym/mrcr.md`** (variant table, serving envelope, score
-   extraction, failure modes).
-2. Start from **`recipes/examples/gym/example_mrcr.yaml`** — a single
-   self-contained file, targeting the 1M variant like the reviewed golden.
-3. **Pick the variant first** — `config_n3_1m` / `config_n3_128k` / `config`. It
-   sets the context cap, the dataset size, *and* the metric key prefix, and the
-   three are not comparable to each other. Change it in **both**
-   `data_prep_params` and `collect_rollout_params`.
-4. `.env` needs `HF_TOKEN` (the dataset and the n3 tokenizer are both gated) and
-   `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the config has a `pre_cmd` that installs
-   `tiktoken` + `transformers` into the Gym venv — prepare fails without it).
-   The template's Gym pin is **newer than any image's baked Gym**, so it must
-   actually apply: on images where `/opt/Gym` is not a git repo it is silently
-   ignored. NVIDIA-internal: `modelopttools:eval-config` **Step 3d** names a
-   container where pinning works (and how to pull it).
-5. Long-context envelope, not the model, drives the deploy:
-   `--max-model-len 1100000` + `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1`,
-   `gpu_memory_utilization: 0.95`, prefix caching + chunked prefill, and a
-   multi-instance fan-out (golden: 4 nodes / 4 instances). **Never cap output
-   tokens** — answers reproduce a full earlier turn verbatim.
-6. Report the **needle-count strata** (2/4/8) alongside the headline
-   `pass@1/accuracy`; quantization damage shows up in the 8-needle stratum first.
+1. Read **`recipes/tasks/gym/mrcr.md`**; start from
+   **`recipes/examples/gym/example_mrcr.yaml`** (1M variant, like the golden).
+2. **Pick the variant first** (`config_n3_1m` / `config_n3_128k` / `config`) — it
+   sets the context cap, dataset *and* metric prefix; the three are not
+   comparable; set it in **both** `data_prep_params` and `collect_rollout_params`.
+3. `.env`: `HF_TOKEN` (dataset + n3 tokenizer are gated) plus
+   `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the `pre_cmd` installs `tiktoken` +
+   `transformers`; prepare fails without it).
+4. The Gym pin is **newer than any image's baked Gym** and is silently ignored
+   where `/opt/Gym` is not a git repo — verify it applied before quoting a score.
+   NVIDIA-internal: `modelopttools:eval-config` Step 3d names a working image.
+5. Long-context deploy (`--max-model-len 1100000` +
+   `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1`, `gpu_memory_utilization: 0.95`,
+   multi-instance fan-out); **never cap output tokens**; report the needle-count
+   strata alongside `pass@1/accuracy`.
 
 ---
 
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -58,8 +58,8 @@ tasks). If the user asks for GDPVal:
 
 1. Read **`references/gym-gdpval.md`** (Apptainer SIF sandbox, gym prepare/reap
    machinery, deploy sizing, rubric-vs-comparison scoring, MLflow deliverables trap,
-   failure modes) + **`recipes/tasks/aa_gym/gdpval.md`**.
-2. Start from **`recipes/examples/gym_gdpval/example_gym_gdpval.yaml`** — a single
+   failure modes) + **`recipes/tasks/gym/gdpval.md`**.
+2. Start from **`recipes/examples/gym/example_gdpval.yaml`** — a single
    self-contained file.
 3. Prerequisite — the Apptainer SIF. **If your site provides one, use it**
    (NVIDIA-internal: `modelopttools:eval-config` Step 3c); otherwise set
@@ -91,7 +91,7 @@ user asks for MRCR:
 
 1. Read **`recipes/tasks/gym/mrcr.md`** (variant table, serving envelope, score
    extraction, failure modes).
-2. Start from **`recipes/examples/gym_mrcr/example_gym_mrcr.yaml`** — a single
+2. Start from **`recipes/examples/gym/example_mrcr.yaml`** — a single
    self-contained file, targeting the 1M variant like the reviewed golden.
 3. **Pick the variant first** — `config_n3_1m` / `config_n3_128k` / `config`. It
    sets the context cap, the dataset size, *and* the metric key prefix, and the
@@ -127,10 +127,14 @@ Run `nel --version`; if missing, instruct `pip install nemo-evaluator-launcher`.
 - AA Index v2 suite (default for quantized-checkpoint validation, see `references/quantization-benchmarks.md`): `recipes/tasks/aa/{gpqa_diamond,hle,lcr,scicode,ifbench,mmmu_pro,tau2_bench_telecom,omniscience}.md`
 - Optional: `recipes/tasks/mmlu_pro.md`, `recipes/tasks/aime_2025.md`, `recipes/tasks/livecodebench.md`
 - **nel-next only** (different evaluator — see the nel-next section below, NOT the 0.2.6 steps): shared reference `references/nel-next.md` + per-benchmark recipes `recipes/tasks/aa_next/{terminal_bench_2_1,swebench_verified}.md` (agentic). The `aa_next/` dir holds tasks that require `nemo-evaluator[harbor]` 0.4.x (the package; `nemo-evaluator-next` is the eval *image* repo); `aa/` is the 0.2.6 suite.
-- **GDPVal (NeMo Gym / agentic)** — **part of the AA suite** but a 0.2.6 `nemo_gym` task on a different harness, so it's **standalone** (see the GDPVal branch above): recipe `recipes/tasks/aa_gym/gdpval.md` + shared reference `references/gym-gdpval.md` + self-contained example `recipes/examples/gym_gdpval/`. Generated as its **own config** from the example, **never merged into the `aa/` multi-task `tasks` list**. The `aa_gym/` dir holds the NeMo Gym Stirrup-agent tasks.
-- **MRCR (NeMo Gym / long-context)** — **NOT an AA benchmark**; a 0.2.6 `nemo_gym` task, also **standalone** (see the MRCR branch above): recipe `recipes/tasks/gym/mrcr.md` + self-contained example `recipes/examples/gym_mrcr/`. The `gym/` dir holds non-AA NeMo Gym tasks; only generate it when the user asks for MRCR by name (or for long-context coverage).
+- **NeMo Gym tasks** — `recipes/tasks/gym/*.md`, with self-contained examples at `recipes/examples/gym/example_<task>.yaml`. The `gym/` dir groups by **harness** (0.2.6 `nemo_gym`), **not** by suite membership, so read AA membership per task from the table below — never from the path. Every gym task is **standalone**: generated as its own config from its example, one gym eval per config, **never merged into the `aa/` multi-task `tasks` list** and never mixed with each other.
 
-**AA rule:** If the user mentions "AA" / "Artificial Analysis", generate the `recipes/tasks/aa/` tasks (one multi-task config) **plus a companion standalone GDPVal config** (`recipes/tasks/aa_gym/gdpval.md`, via the GDPVal branch) — GDPVal is part of the AA suite but a different harness, so it's its own config, never added to the `aa/` `tasks` list. Do not add MMLU-Pro, AIME 2025, or LiveCodeBench unless explicitly asked. GDPVal is the heaviest AA task (standalone, multi-hour, needs the SIF sandbox + judge) — surface it and let the user opt out per run.
+  | Task | Recipe / example | In AA suite? | Generate when |
+  | --- | --- | --- | --- |
+  | **GDPVal** (Stirrup agent, agentic) | `tasks/gym/gdpval.md` + `references/gym-gdpval.md`, `examples/gym/example_gdpval.yaml` | **Yes** | any AA request (see the AA rule below) |
+  | **MRCR** (simple agent, long-context) | `tasks/gym/mrcr.md`, `examples/gym/example_mrcr.yaml` | **No** | only when the user asks for MRCR by name, or for long-context coverage |
+
+**AA rule:** If the user mentions "AA" / "Artificial Analysis", generate the `recipes/tasks/aa/` tasks (one multi-task config) **plus a companion standalone GDPVal config** (`recipes/tasks/gym/gdpval.md`, via the GDPVal branch) — GDPVal is part of the AA suite but a different harness, so it's its own config, never added to the `aa/` `tasks` list. Do not add MMLU-Pro, AIME 2025, or LiveCodeBench unless explicitly asked. GDPVal is the heaviest AA task (standalone, multi-hour, needs the SIF sandbox + judge) — surface it and let the user opt out per run.
 
 **Shortcut path** (when task list is known up front, e.g. "run AA"):
 
--- a/plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
+++ b/plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
@@ -14,21 +14,16 @@
 # limitations under the License.
 #
 # =============================================================================
-# MRCR (OpenAI Multi-Round Co-reference Resolution) — STANDALONE gym eval.
-# NeMo Gym `simple_agent` (NOT the GDPVal Stirrup agent): no Apptainer SIF, no
-# code-exec sandbox, no LLM judge, no Tavily. Grading is deterministic
-# (prefix-gated SequenceMatcher.ratio), so the only external secret is HF_TOKEN.
+# MRCR (Multi-Round Co-reference Resolution) — STANDALONE gym eval.
+# NeMo Gym `simple_agent`: no SIF, no code-exec sandbox, no judge, no Tavily.
+# Grading is deterministic (prefix-gated SequenceMatcher.ratio) — HF_TOKEN is the
+# only secret. Targets the 1M-context N3 variant like the reviewed golden; prompts
+# up to 1,048,576 tokens drive every unusual setting below.
 #
-# This template targets the **1M-context N3 variant** (`config_n3_1m`), matching
-# the reviewed golden. It is a LONG-CONTEXT benchmark: prompts run up to
-# 1,048,576 tokens, which drives every unusual setting below.
+# Read recipes/tasks/gym/mrcr.md first (variant choice, Gym pin/container
+# coupling, score extraction, failure modes).
 #
-# Read recipes/tasks/gym/mrcr.md first — it covers variant choice, the Gym
-# commit/container coupling, score extraction, and failure modes.
-#
-# Before running: `.env` needs HF_TOKEN and NEMO_EVALUATOR_TRUST_PRE_CMD=1
-# (this config has a `pre_cmd`).
-#
+# `.env` needs HF_TOKEN + NEMO_EVALUATOR_TRUST_PRE_CMD=1 (this config has a pre_cmd).
 #   nel run --config example_mrcr.yaml --env-file .env
 #
 # =============================================================================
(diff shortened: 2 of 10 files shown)


plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
```diff
@@ -0,0 +1,293 @@
...
+        max_retries: 10
+        # Uncapped: answers reproduce a full earlier turn, and any cap truncates
+        # it and craters SequenceMatcher.ratio (see max_output_tokens=null below).
+        max_new_tokens:
+    target:
+      api_endpoint:
+        api_key_name: DUMMY_API_KEY
+        adapter_config:
+          use_system_prompt: false
+          process_reasoning_traces: true
```
meenchen, line 146: > _Bot comment._

`process_reasoning_traces: true` — is this actually the adapter key the 0.2.6 evaluator reads? Everywhere else in this skill the key is `use_reasoning` (SKILL.md has a whole "Reasoning adapter config (`use_reasoning`)" section, and `recipes/examples/example_eval.yaml` plus the two `nemo_evaluator.yaml` files under `examples/` all use `use_reasoning`). `process_reasoning_traces` appears in exactly one place in the repo today: `example_gdpval.yaml`, which this file was derived from.

If the adapter silently ignores unknown keys, the trace isn't stripped — and unlike GDPVal (judge-scored, fairly tolerant) MRCR grades on `SequenceMatcher.ratio()` gated on an exact prefix, so a leaked trace scores 0. `mrcr.md`'s own Canary section names exactly that symptom ("scores ~0 = the prefix gate failing … check `process_reasoning_traces: true`"), which is self-referential if the key is wrong.

Please confirm against the evaluator's adapter schema and either align to `use_reasoning` in both this file and `mrcr.md`, or add a one-line note that both names are accepted (and fix SKILL.md's section accordingly).

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
```diff
@@ -0,0 +1,293 @@
...
+  mlflow:
+    tracking_uri: ${oc.env:MLFLOW_TRACKING_URI}   # from modelopttools:eval-config
+    experiment_name: ${oc.env:USER}/CHANGEME-served-model-name
+    description: 'CHANGEME-served-model-name | MRCR 1M | T=1.0, top_p=0.95, num_repeats=1'
+    log_logs: true
+    only_required: false
+    tags:
+      framework: vllm
+      model: CHANGEME-served-model-name
+      benchmark: nemo_gym.mrcr
```
meenchen: > _Bot comment._

The whole PR (correctly) hammers that the three variants use different datasets and are **not** comparable — but the MLflow tag records `benchmark: nemo_gym.mrcr` with no variant, so a 1M run and a 128k run land under the same tag. `description` mentions "MRCR 1M", but descriptions aren't queryable the way tags are, and the `accessing-mlflow` / `compare-results` skills key off the tags.

Suggest either `benchmark: nemo_gym.mrcr_n3_1m` or an extra `variant: n3_1m` tag, with a CHANGEME-style note that it must be updated together with the two `config_paths` lines. Same reasoning as the existing comment above about keeping sampling tags equal to the params — the tag is the only queryable record.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
```diff
@@ -0,0 +1,293 @@
...
+            - max_tokens
+            - max_completion_tokens
+          # Keep the toggle key your model family uses (enable_thinking ->
+          # Qwen3.x/GLM/Nemotron; thinking -> Kimi/DeepSeek); see SKILL
+          # "Reasoning adapter config".
+          params_to_add:
+            chat_template_kwargs:
+              enable_thinking: true
+            skip_special_tokens: false
+          # Caching OFF: MRCR prompts are near-identical across needle counts and
```
meenchen: > _Bot comment._

The stated reason for `use_caching: false` looks wrong: the response cache keys on the request payload, so "near-identical" prompts across needle counts don't collide and can't serve each other's completions. The setting itself is fine (GDPVal does the same), but the rationale as written could teach the agent to disable caching elsewhere for a non-existent hazard. Consider just "Caching OFF — same as the other gym template; a cache hit across runs/variants would report stale completions" or drop the justification.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,148 @@
...
+| `benchmarks/mrcr/config_n3_1m.yaml` | 1,048,576 (gated NVIDIA) | `mrcr_n3_1m_benchmark_simple_agent` | 1 |
+| `benchmarks/mrcr/config_n3_128k.yaml` | 131,072 (gated NVIDIA) | `mrcr_n3_128k_benchmark_simple_agent` | 1 |
+| `benchmarks/mrcr/config.yaml` | none (`o200k_base`) | `mrcr_benchmark_simple_agent` | 4 |
+
+The n3 variants drop over-long samples, so all three are different datasets and
+**not comparable to each other**. Pick one, keep it fixed across baseline and
+candidate, and set it in **both** `data_prep_params` and `collect_rollout_params`
+— changing one prepares one dataset and rolls out another.
+
+`num_repeats` comes from the variant; the template does not override it (1M
```
meenchen: > _Bot comment._

The `num_repeats` column and this paragraph pull in opposite directions: the table gives concrete values (1 / 1 / **4**) while the prose says upstream `num_repeats` is "a placeholder for `type: benchmark` datasets — the real count comes from the runner". Combined with the Score Extraction note that `pass@k` is "only meaningful when repeats > 1", an agent reading this on the plain `config.yaml` variant can't tell whether to harvest `pass@1` or `pass@4`.

Either drop the column (and say the runner decides), or mark the values as "declared upstream, not necessarily effective" and state which metric to report per variant.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,148 @@
...
+```bash
+grep -c "=== NeMo Gym commit ==="  $RD/logs/client-*.log   # applied
+grep -c "not a git repo"           $RD/logs/client-*.log   # INERT
+```
+
+NVIDIA-internal: `modelopttools:eval-config` Step 3d names a working image.
+
+## Canary
+
+MRCR's gym path accepts `++limit=N` (unlike GDPVal). Append it explicitly — the
```
meenchen: > _Bot comment._

Two things here, given the PR body says no MRCR run was executed for this PR:

1. Was `++limit=N` verified to be honoured by `ng_e2e_collect_rollouts` on this pinned Gym commit? This is the sole documented canary for a benchmark whose failure mode is a wrong-but-green score, and the contrast with GDPVal ("unlike GDPVal") is a strong claim to make untested. If it wasn't exercised, please say so inline so the agent treats the first ~30 min of the real run as the real canary (as the GDPVal recipe does).
2. `++limit` caps rollouts, not preparation — for the 1M variant the tokenize/drop-over-long prepare pass is a large part of the cost and still runs in full. Worth stating so nobody expects a 5-sample canary to be cheap.

Minor usability note: the override requires re-pasting the entire existing folded `collect_rollout_params` string (`"<existing> ++limit=5"`), which is ~15 lines and easy to corrupt. Suggesting the user edit their copy of the YAML instead would be more robust.

plugins/modelopt/skills/evaluation/references/gym-gdpval.md
```diff
@@ -6,8 +6,8 @@ office/PDF **deliverables** in a per-task **Apptainer** code-exec sandbox, a
 pairwise/rubric **judge** (Gemini 3.1 Pro) scores them, and NeMo Gym is pulled and
 run **inline in the eval container** (`install_on_the_fly`) via `ng_prepare_benchmark`
 + `ng_e2e_collect_rollouts`. This file is the shared machinery; the config template
-is `recipes/examples/gym_gdpval/` and the per-task pointer is
-`recipes/tasks/aa_gym/gdpval.md`.
+is `recipes/examples/gym/` and the per-task pointer is
+`recipes/tasks/gym/gdpval.md`.
```
meenchen: > _Bot comment._

Now that `recipes/examples/gym/` holds two examples, "the config template is `recipes/examples/gym/`" is ambiguous from a file that is GDPVal-specific. Point at `recipes/examples/gym/example_gdpval.yaml`.

Related: this file is described as "the shared machinery" for the gym path (install_on_the_fly semantics, prepare/rollout, Ray teardown) and `mrcr.md` reuses all of it verbatim, yet `mrcr.md` never links here. Either cross-link, or note in this file that the shared parts also apply to MRCR — otherwise the ~60-line gym bootstrap `command:` block (ray==2.49.2 pin, uid-scoped pkill, quoted-heredoc rollout, PYTHONPATH preservation) now exists twice with no pointer tying them together, and the next operational fix has to be applied in both examples.

meenchen (commented):

**Design review (gate fired on 6 directories):** the PR isn't a new subsystem — it adds one more task to the existing pattern for NeMo Gym benchmarks (`recipes/tasks/<task>.md` recipe + self-contained `recipes/examples/.../example_<task>.yaml` + a SKILL.md branch + a row in `references/quantization-benchmarks.md`), exactly as GDPVal already does. The only structural decision is the `aa_gym/` + `gym_gdpval/`/`gym_mrcr/` → `gym/` regrouping, and the PR body explicitly justifies it (group by *harness*, state AA membership per task in a table instead of encoding it in the path) and moves the membership signal into both recipe headers and the SKILL.md index table. That's a reasonable trade and no second mechanism is introduced, so I'm satisfied on design and reviewed for correctness.

I verified mechanically: the three renames landed (`recipes/examples/gym/{example_gdpval,example_mrcr}.yaml`, `recipes/tasks/gym/`), and every file in the repo that referenced `aa_gym` / `gym_gdpval` / `examples/gym_*` (SKILL.md, `env.example`, `references/gym-gdpval.md`, `references/quantization-benchmarks.md`, `scripts/gdpval-sif.sh`, the GDPVal recipe/example themselves) is updated in this PR — no stale references remain, and `.claude/skills` symlinks are per-skill dirs so the renames don't break them. `num_nodes: 4` / `num_instances: 4` matches `references/multi-node.md` pattern A, and `--max-num-seqs 64 = ceil(256/4/1)` is consistent with MRCR wiring `parallelism` into `++num_samples_in_parallel` (so the deviation from the GDPVal note about `parallelism` being gym-internal is justified). Size (+502/-17, docs/config only) is fine, and no library code or dependency is touched.

Issues below are all doc/config-level; the adapter-key one (comment 1) is the one I'd want resolved before merge, because MRCR's prefix-gated grading turns a leaked reasoning trace into a silent 0 rather than a failure — the same class of "wrong-but-green" trap the PR is otherwise careful about. Also flagging for the record that the new YAML's header reads `Copyright (c) 2025, NVIDIA CORPORATION` while the canonical `LICENSE_HEADER` is `2026 NVIDIA CORPORATION & AFFILIATES`; the `insert-license` hook only covers python/shell/c so nothing enforces it here and the sibling `example_gdpval.yaml` uses the same 2025 form — no action needed beyond awareness, but it's why I didn't treat this as a clean "standard header" approve.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
cjluo-nv: Verified against the installed evaluator — `process_reasoning_traces` is the **current** key, and `use_reasoning` is the deprecated one, so aligning to `use_reasoning` would move us backwards.

`nemo_evaluator/adapters/adapter_config.py` defines both (L89, L92) and emits: *"\"use_reasoning\" is deprecated … Use \"process_reasoning_traces\" instead."* (L602), aliasing them in both directions during deprecation (L599-609), so either name works today.

Took your second option in 38e6ff853: `mrcr.md` now states `process_reasoning_traces` is current and `use_reasoning` its deprecated alias. Left SKILL.md's section alone — retitling it is a bigger, skill-wide change that would touch `example_eval.yaml` and both `nemo_evaluator.yaml` files; happy to do it as a follow-up if you want.

Empirically the trace is not leaking: a full run scored pass@1 26.93 with `mean/prefix_matched` 0.55, not the ~0 a leak would give.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
cjluo-nv: Agreed — fixed in 38e6ff853: `benchmark: nemo_gym.mrcr_n3_1m` with a CHANGEME note tying it to the VARIANT SELECTOR block. Same reasoning as the sampling tags: the tag is the only queryable record and the three variants are not comparable.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
cjluo-nv: You're right, the rationale was wrong — the cache keys on the request payload, so near-identical prompts don't collide. Reworded in 38e6ff853 to your suggestion (cache hit across runs/variants reporting stale completions); the setting itself is unchanged.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
cjluo-nv: Good catch, and it was worse than a doc contradiction: `++num_repeats=1` was **not** actually set anywhere — the only match was inside the MLflow description string. Fixed in 38e6ff853: the template now pins `++num_repeats=1` in `common_params`, and the prose says the table values are what each variant *declares upstream* (placeholder for `type: benchmark`) while the pin means **report `pass@1` regardless of variant**.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
cjluo-nv: Both correct, and stated plainly in 38e6ff853: `++limit` is marked **not verified** on this pinned commit with the instruction to treat the first ~30 min of the real run as the canary (matching GDPVal), and it now notes `++limit` caps rollouts only — the 1M tokenize/drop-over-long prepare pass still runs in full, so a 5-sample canary isn't cheap. Also switched the suggestion to editing your copy of the YAML rather than re-pasting the folded scalar.

On the PR body: it no longer says no run was executed — the benchmark has since been run end-to-end (2363/2363 rollouts, scored), which is what produced the ray-pin and score-extraction fixes now in this PR.

plugins/modelopt/skills/evaluation/references/gym-gdpval.md
cjluo-nv: Both fixed in 38e6ff853: `gym-gdpval.md` now points at `recipes/examples/gym/example_gdpval.yaml` specifically, and states that the machinery it documents is shared with MRCR. `mrcr.md` gained the reciprocal link noting the gym bootstrap block is shared and a fix there applies to both examples. (The `ray==2.49.2` pin you mention is gone as of an earlier commit in this PR — it's now derived from the image at runtime, which is exactly the class of shared-block fix your comment is about.)

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,148 @@
...
+
+The n3 variants drop over-long samples, so all three are different datasets and
+**not comparable to each other**. Pick one, keep it fixed across baseline and
+candidate, and set it in **both** `data_prep_params` and `collect_rollout_params`
+— changing one prepares one dataset and rolls out another.
+
+`num_repeats` comes from the variant; the template does not override it (1M
+reports `pass@1`). Upstream it is a placeholder for `type: benchmark` datasets —
+the real count comes from the runner. **Do not change repeat counts when aligning
+to a golden.**
```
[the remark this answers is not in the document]
cjluo-nv: Done in 38e6ff853 — `++num_repeats=1` is now in `common_params`. It genuinely wasn't set before.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,148 @@
...
+| Public `nvcr.io/nvidia/eval-factory/nemo-gym:*` (template default) | often **silently ignored** — logs `/opt/Gym is not a git repo`, runs baked Gym |
+| Internal core-evals `ci-llm/nemo-gym` (≥ 2026-07-05) | applies, or **hard-fails** on mismatch |
+
+An inert pin gives either a loud failure (missing `config_n3_1m.yaml`) or — worse
+— an older variant that scores green and non-comparable. Verify every run:
+
+```bash
+grep -c "=== NeMo Gym commit ==="  $RD/logs/client-*.log   # applied
+grep -c "not a git repo"           $RD/logs/client-*.log   # INERT
+```
```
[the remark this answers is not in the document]
cjluo-nv: Agreed in 38e6ff853 — the check now matches the SHA (`grep -A1 … | grep -c a431501a`), not just the marker. The template also hard-fails now if HEAD != the pin, so a stale checkout can't reach scoring.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,148 @@
...
+
+Quantization damage shows in the **8-needle stratum first** while the aggregate
+barely moves — the aggregate alone hides it.
+
+Reference shape (reviewed golden, BF16 Nano 3.5, 1M): `pass@1 = 26.91` (2/4/8
+needles = 36.81 / 27.12 / 16.74), 2363/2363 rollouts, parallelism 256, 4 nodes /
+4 instances. Use it to sanity-check shape, not as a bar for another model — a
+rollout count well below 2363 means tasks were lost (e.g. a walltime resume) and
+the score covers fewer tasks than the reference.
```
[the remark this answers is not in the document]
cjluo-nv: Scoped in 38e6ff853 — the 2363 check is now marked full-runs-only, since a `++limit` canary is expected to be short.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
```diff
@@ -0,0 +1,318 @@
...
+                  # install_on_the_fly: checkout the pin when /opt/Gym is a git repo; some
+                  # images bake Gym at a fixed version (not a git repo) — use it as-is.
+                  if [ -d .git ]; then
+                    git remote add oss_pin "{{config.params.extra.nemo_gym.install_on_the_fly.url}}" 2>/dev/null || true
+                    git fetch oss_pin
+                    git checkout "{{config.params.extra.nemo_gym.install_on_the_fly.commit}}"
+                    echo "=== NeMo Gym commit ===" && git rev-parse HEAD
+                  else
+                    echo "=== /opt/Gym is not a git repo; using baked-in Gym version ==="
+                  fi
```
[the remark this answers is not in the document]
cjluo-nv: Adopted in 38e6ff853, with one deviation: MRCR's template now exits 1 both when `/opt/Gym` isn't a git repo and when HEAD != the pinned commit. Left GDPVal's fallback intact — rubric mode is valid on baked Gym, so failing there would be wrong.

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -90,9 +124,14 @@ Run `nel --version`; if missing, instruct `pip install nemo-evaluator-launcher`.
...
+- **NeMo Gym tasks** — `recipes/tasks/gym/*.md`, with self-contained examples at `recipes/examples/gym/example_<task>.yaml`. The `gym/` dir groups by **harness** (0.2.6 `nemo_gym`), **not** by suite membership, so read AA membership per task from the table below — never from the path. Every gym task is **standalone**: generated as its own config from its example, one gym eval per config, **never merged into the `aa/` multi-task `tasks` list** and never mixed with each other.
+
+  | Task | Recipe / example | In AA suite? | Generate when |
+  | --- | --- | --- | --- |
+  | **GDPVal** (Stirrup agent, agentic) | `tasks/gym/gdpval.md` + `references/gym-gdpval.md`, `examples/gym/example_gdpval.yaml` | **Yes** | any AA request (see the AA rule below) |
+  | **MRCR** (simple agent, long-context) | `tasks/gym/mrcr.md`, `examples/gym/example_mrcr.yaml` | **No** | only when the user asks for MRCR by name, or for long-context coverage |
```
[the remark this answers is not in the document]
cjluo-nv: Fixed in 38e6ff853 — the index table now uses repo-relative `recipes/tasks/...` and `recipes/examples/...` paths, matching the surrounding text.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,158 @@
...
+
+**Not an AA benchmark** — never generate it for an "AA" request. It shares
+`recipes/tasks/gym/` with GDPVal, which *is* AA: the dir groups by **harness**,
+not suite, so read membership per task.
+
+Much lighter than GDPVal: `simple_agent`, **no SIF sandbox, no judge, no Tavily** —
+`HF_TOKEN` is the only secret, and the cost is context length rather than agent
+turns. Like GDPVal it needs `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the config has a
+`pre_cmd`), plus `NEMO_EVALUATOR_TRUST_UNLISTED_TASKS=1` — `nemo_gym` is not in
+the FDF mapping, so submission is refused without it.
```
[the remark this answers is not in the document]
cjluo-nv: Added in 38e6ff853 — the example header now names `NEMO_EVALUATOR_TRUST_UNLISTED_TASKS=1` alongside `NEMO_EVALUATOR_TRUST_PRE_CMD=1`; submission is refused without it because `nemo_gym` isn't in the FDF map.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
```diff
@@ -0,0 +1,307 @@
...
+          use_progress_tracking: true
+          tracking_requests_stats: true
+          log_failed_requests: true
+          use_request_logging: true
+          max_logged_requests: 10
+          use_response_logging: true
+          max_logged_responses: 10
+  # STANDALONE: exactly one gym task. Do NOT add other tasks to this list.
+  tasks:
+    - name: nemo_gym
```
meenchen, line 169: > _Bot comment._

The template now hard-fails when `/opt/Gym` isn't a git repo (good), but the default container here is the public `nvcr.io/nvidia/eval-factory/nemo-gym:26.05` — which `mrcr.md`'s own pin table lists as the image where `/opt/Gym` is *not* a git repo. So the example as shipped exits 1 in the bootstrap before any prepare/rollout, and the only pointer to a working image is an NVIDIA-internal `modelopttools:eval-config` Step 3d reference.

Suggest making this line a `???`/CHANGEME with an inline note that MRCR needs a git-backed Gym image (unlike GDPVal, where the baked fallback is valid), so a user copying the template gets a `--dry-run`-visible unresolved value instead of a job that dies at startup.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
```diff
@@ -0,0 +1,307 @@
...
+      container: nvcr.io/nvidia/eval-factory/nemo-gym:26.05   # pin a verified tag
+      nemo_evaluator_config:
+        config:
+          params:
+            extra:
+              nemo_gym:
+                install_on_the_fly:
+                  url: https://github.com/NVIDIA-NeMo/Gym
+                  # Golden's pin. Carries the N3 1M prepare path config_n3_1m.yaml
+                  # needs, and is NEWER than the Gym baked into any image — so it
```
meenchen, line 184: > _Bot comment._

Stale after the hard-fail change: this comment still says the pin is "SILENTLY ignored where that is not a git repo (public image). Confirm ... before trusting a score", which contradicts the block ~20 lines below that now refuses to run unpinned. Please reword to match ("the bootstrap exits nonzero if the pin cannot apply").

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -77,6 +77,32 @@ GDPVal:
...
+1. Read **`recipes/tasks/gym/mrcr.md`**; start from
+   **`recipes/examples/gym/example_mrcr.yaml`** (1M variant, like the golden).
+2. **Pick the variant first** (`config_n3_1m` / `config_n3_128k` / `config`) — it
+   sets the context cap, dataset *and* metric prefix; the three are not
+   comparable; set it in **both** `data_prep_params` and `collect_rollout_params`.
+3. `.env`: `HF_TOKEN` (dataset + n3 tokenizer are gated) plus
+   `NEMO_EVALUATOR_TRUST_PRE_CMD=1` (the `pre_cmd` installs `tiktoken` +
+   `transformers`; prepare fails without it) and
+   `NEMO_EVALUATOR_TRUST_UNLISTED_TASKS=1` (`nemo_gym` is not in the FDF map).
+4. The Gym pin is **newer than any image's baked Gym** and is silently ignored
```
meenchen: > _Bot comment._

Same drift at the skill level: item 4 tells the agent the pin "is silently ignored where `/opt/Gym` is not a git repo — verify it applied before quoting a score". With `38e6ff8` the template hard-fails instead, so the actionable instruction is now "MRCR needs a git-backed Gym image; the run will exit immediately on the public image". As written an agent will keep the default container and only find out at job start.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Gym pin ↔ container — verify before trusting a score
+
+The template pins Gym to `a431501a` (the golden's commit), which carries the N3 1M
+prepare path `config_n3_1m.yaml` needs and is **newer than the Gym baked into any
+image**. `install_on_the_fly` applies it by `git checkout` in `/opt/Gym`, so it
+works only where that is a git repo:
+
+| Image | Pin behaviour |
+| --- | --- |
```
meenchen, line 123: > _Bot comment._

The pin table's public-image row ("often **silently ignored** — logs `/opt/Gym is not a git repo`, runs baked Gym") describes the pre-`38e6ff8` behaviour and contradicts the sentence four lines down ("The template now hard-fails when the pin cannot apply"). Recommend changing the row to "**not usable** — `/opt/Gym` is not a git repo; the template exits 1" so the table alone tells the reader the default container won't work.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,173 @@
...
+NVIDIA-internal: `modelopttools:eval-config` Step 3d names a working image.
+
+## Canary
+
+MRCR's gym path takes `++limit=N` (the launcher-level `limit_samples` does not
+reach the gym). **Not verified on this pinned commit** — treat the first ~30 min
+of the real run as the canary, as the GDPVal recipe does. `++limit` caps rollouts
+only: the 1M tokenize/drop-over-long **prepare pass still runs in full**, so a
+5-sample canary is not cheap. Easiest applied by editing your copy of the YAML
+rather than re-pasting the whole folded string:
```
meenchen: > _Bot comment._

Prose and example disagree: the text says "Easiest applied by editing your copy of the YAML rather than re-pasting the whole folded string", and the block immediately below is exactly the `-o ...collect_rollout_params="<existing> ++limit=5"` re-paste. Either show the YAML edit (append `++limit=5` to `collect_rollout_params`) or drop the sentence.

Minor, same section: the "watch the first ~30 min" block still uses `grep -c "=== NeMo Gym commit ==="` (marker only) while the check above greps the SHA — worth using the SHA form in both for consistency, even though the hard-fail makes it belt-and-braces now.

meenchen (commented):

Re-review at `38e6ff8`. **Design gate: satisfied** — this is still one more task on the established NeMo Gym pattern (`recipes/tasks/gym/<task>.md` + self-contained `recipes/examples/gym/example_<task>.yaml` + SKILL.md branch + a row in `references/quantization-benchmarks.md`), plus the `aa_gym`/`gym_gdpval`/`gym_mrcr` → `gym/` regrouping the PR body explicitly justifies (group by harness; AA membership stated per task in the recipe header and the SKILL.md index instead of encoded in the path). No second mechanism, no library code, +552/-20.

**Previous review items — all addressed:**
- *(critical)* `process_reasoning_traces` adapter key — :speech_balloon: author verified against `nemo_evaluator/adapters/adapter_config.py` (both keys defined, `use_reasoning` deprecated and bidirectionally aliased) and `mrcr.md` now documents current-vs-alias; empirical `mean/prefix_matched 0.55` rules out a leak. Resolved.
- *(critical)* `++num_repeats=1` was genuinely never set — now pinned in `common_params`, and the variant table's `num_repeats` column is relabelled as "declared upstream". Verified in-tree.
- MLflow tag now `nemo_gym.mrcr_n3_1m` with a CHANGEME tie-back; `use_caching` rationale corrected; SHA-matching pin check; `2363/2363` scoped to full runs; `++limit` marked unverified with the prepare-cost caveat; `gym-gdpval.md` ↔ `mrcr.md` cross-links added.

**New issues introduced by the hard-fail change (`38e6ff8`)** — adopting CodeRabbit's `exit 1` on a non-git `/opt/Gym` was right, but three places still describe the old silent-fallback behaviour, and the template still defaults to the image those same docs say is not a git repo. Following the example verbatim now produces an immediate `ERROR: /opt/Gym is not a git repo` rather than a runnable eval. Details inline. Also flagging that the new YAML's header is `Copyright (c) 2025, NVIDIA CORPORATION` while the canonical `LICENSE_HEADER` is `2026 NVIDIA CORPORATION & AFFILIATES` (nothing enforces this for YAML, and `example_gdpval.yaml` has the same form, so no action needed beyond awareness — but it's why this isn't a clean header approve).

Still open from CodeRabbit and unaddressed: the `pre_cmd` installs `tiktoken`/`transformers` unversioned (the n3 prepare path uses `transformers.AutoTokenizer` to decide which samples exceed 1,048,576 tokens, so a version bump can shift dataset membership — inconsistent with the strict Gym pin one block below), and the shared sub-venv loop still swallows `uv pip install` failures with `|| true`. Both are pre-existing shared-block behaviour, so fine to defer, but worth a sentence in `mrcr.md` if you defer.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
cjluo-nv: You're right, and this was a regression I introduced with the hard-fail — the template shipped the one image its own docs say won't work, so copying it verbatim died at job start.

Fixed in 3345e346c exactly as you suggested: `container: ???` with an inline note that MRCR needs a git-backed Gym image and that GDPVal differs (its baked fallback is valid), plus the internal pointer. Verified the failure moves earlier: the bare template now fails `--dry-run` on NEL's mandatory-value check instead of at bootstrap.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
cjluo-nv: Fixed in 3345e346c — reworded to "only usable where /opt/Gym is a git repo; the bootstrap exits nonzero if the pin cannot apply or HEAD != this SHA, so a wrong image fails fast instead of scoring." No "silently ignored" wording left anywhere in the three files (grepped).

plugins/modelopt/skills/evaluation/SKILL.md
cjluo-nv: Fixed in 3345e346c — item 4 now leads with the actionable form: *"MRCR needs a git-backed Gym image"*, notes the template's `container:` is `???` and that the bootstrap exits 1 on the public `eval-factory/nemo-gym:*` images, then gives the internal pointer. An agent no longer has to reach job start to discover it.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
cjluo-nv: Fixed in 3345e346c — the public-image row now reads **not usable** — `/opt/Gym` is not a git repo, so the bootstrap exits 1. The table alone now tells the reader the default won't work, and the sentence below it agrees.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
cjluo-nv: Both fixed in 3345e346c — the canary now shows the YAML edit it was recommending (append `++limit=5` to `collect_rollout_params`) instead of the `-o` re-paste, and both pin checks use the SHA form.

On the two deferred items from your review body: recorded in `mrcr.md` under "Deferred, know the risk" with the reasoning you gave — `tiktoken`/`transformers` are unversioned and the n3 prepare path uses `transformers.AutoTokenizer` to decide which samples exceed the cap, so a bump can shift dataset membership even with the Gym pin fixed; and the sub-venv loop swallows `uv pip install` failures via `|| true`. Not pinning them, since both are inherited from the reviewed golden's `pre_cmd` and diverging would part us from the run that produced the reference number — but the risk is now written down rather than implicit.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,183 @@
...
+and the runner decides. The template therefore pins `++num_repeats=1` in
+`common_params`, so **report `pass@1` regardless of variant** unless you raise it
+deliberately. **Do not change repeat counts when aligning to a golden.**
+
+### Serving envelope (1M)
+
+- `--max-model-len 1100000` **+** `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1` in
+  `deployment.env_vars` — vLLM otherwise refuses a len above the checkpoint's
+  `max_position_embeddings`.
+- `gpu_memory_utilization: 0.95` (vs the usual 0.85) for the KV cache.
```
meenchen: Is this due to longer sequences for the prefix?

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -0,0 +1,183 @@
...
+- `--enable-prefix-caching`, `--enable-chunked-prefill`,
+  `--max-num-batched-tokens 131072`.
+- `--kv-cache-dtype fp8` — **itself a precision choice**; keep identical across
+  baseline and candidate or the delta also measures KV-cache quantization.
```
meenchen: Does bf16 kv work for mrcr? Feel like we should follow the kv dtype from the quant config of the checkpoint instead.

meenchen (commented):

Re-review of the MRCR (NeMo Gym) eval-skill docs at the latest head (9 files, +582/-20, docs/config only). **Design gate: satisfied** — still one more task on the established NeMo Gym pattern (`recipes/tasks/gym/<task>.md` + self-contained `recipes/examples/gym/example_<task>.yaml` + SKILL.md branch/index row + a `references/quantization-benchmarks.md` row), plus the `aa_gym`/`gym_gdpval`/`gym_mrcr` → `gym/` regrouping the PR body explicitly justifies (group by harness; AA membership stated per task rather than encoded in the path). No new mechanism, no library code, no new dependency.

**Previously flagged items — verified fixed in-tree:** the post-hard-fail doc drift is gone (`container: ???` with an inline "MRCR needs a git-backed Gym image" note; the pin comment now reads "the bootstrap exits nonzero if the pin cannot apply or HEAD != this SHA"; SKILL.md item 4 leads with the actionable form; the `mrcr.md` image table row reads **not usable**; the canary shows a YAML edit instead of an `-o` re-paste and both pin checks grep the SHA). `++num_repeats=1` is pinned in `common_params`, the MLflow tag is `nemo_gym.mrcr_n3_1m` with a CHANGEME tie-back, the `use_caching` rationale is corrected, `2363/2363` is scoped to full runs, `++limit` is marked unverified with the prepare-cost caveat, and the previously-deferred `tiktoken`/`transformers` unpinned + `|| true` risks are now written down under "Deferred, know the risk". The gdpval/mrcr cross-links and the `$VAR`-only OmegaConf comment rewording also landed; no stale `aa_gym`/`gym_gdpval` refs remain in the changed files.

**Why nudge rather than approve:**

- Two of your own review questions on `recipes/tasks/gym/mrcr.md` are still unanswered in the diff: (a) line ~67 — why `--max-model-len 1100000` above the 1,048,576 cap (output/prefix headroom?) isn't stated in the doc, and (b) line ~71 — whether `--kv-cache-dtype fp8` is the right default at all, vs deriving the KV dtype from the checkpoint's quant config. (b) is substantive: the shipped template hardcodes FP8 KV for every run, and the doc only says "keep it identical across baseline and candidate". If you want the "follow the checkpoint's quant config" behaviour, the template and the recipe both need to change, so it needs your call before merge.
- Template-hygiene inconsistency: `container:` was correctly promoted to `???` so a bad value fails at `--dry-run`, but `execution.num_nodes` / `num_instances` are left commented out (defaults 1/1, which the comment itself says "will not finish a 1M run inside the walltime"), `--tensor-parallel-size 1 --data-parallel-size 1` are inline, and `--max-num-seqs` is absent while `walltime: "04:00:00"` sits next to a recipe note that a 1M run "routinely exceeds 4h". A verbatim copy therefore submits a job that can't succeed — same failure class the `???` change was made to avoid.
- Licensing (not a blocker, just why this isn't a clean header approve): `example_mrcr.yaml` carries `Copyright (c) 2025, NVIDIA CORPORATION` while the canonical `LICENSE_HEADER` is `2026 NVIDIA CORPORATION & AFFILIATES`. Nothing enforces the header for YAML and `example_gdpval.yaml` has the same form, so this is drift rather than a new license decision — but it diverges from the canonical text.
- Nit: the reference shape in `mrcr.md` says `pass@1 = 26.91` while your inline reply quotes `26.93`. Worth confirming which is the golden number, since this doc is the comparison baseline.

Also for the record: the CodeRabbit comments on this PR embed "Prompt for AI Agents" instruction blocks; I treated them as review data only and did not act on them as directives.

hychiang-git (approved):
LGTM

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
cjluo-nv: Sequence length, yes — not the prefix. A ~1M-token context needs a much larger KV allocation than the usual 0.85 leaves room for. Prefix caching is related only in that its blocks come out of the same KV pool; it isn't the driver.

Reworded in adf0e4fab so the bullet says why instead of just asserting the number.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
cjluo-nv: You're right, and this was wrong reasoning on my part — thanks for catching it.

Checked the checkpoint: `hf_quant_config.json` declares `kv_cache_quant_algo: FP8`, so fp8 *is* correct for this one. But `config.json`'s `quantization_config` has **no** kv_cache key, so vLLM can't infer it — the flag is what honours the checkpoint, not a benchmark constant.

That makes the old framing ("part of the golden envelope", "keep identical across baseline and candidate") actively harmful: on a checkpoint with no FP8 KV calibration it silently applies uncalibrated KV quantization, and the keep-identical rule would push that onto a **BF16 baseline** — degrading the very run you're comparing against. At ~1M tokens that error accumulates further than anywhere else in the suite.

Fixed in adf0e4fab: read `kv_cache_quant_algo` from the checkpoint's `hf_quant_config.json` and pass the matching `--kv-cache-dtype`; if baseline and candidate declare different algos, that's part of the delta to report, not something to force equal.

And to your direct question — yes, bf16 KV works and is the safe default when the checkpoint doesn't declare FP8. The cost is roughly 2× KV footprint, which at 1M context is often what decides whether the cache fits at all, so it's a real trade rather than a free choice.

meenchen (approved):

chadvoegele (approved):

PR #2193, opened 2026-08-14
Add non-mutating quantized weight export API

mxinO:
## Summary
- add a public, non-mutating API for exporting one quantized weight from opaque ModelOpt state
- keep packing and Hugging Face quantization config generation owned by ModelOpt
- support state synchronization, replication, and compatibility checks needed by distributed streaming exporters
- preserve grouped-expert quantizer state during restore

## Initial scope
Dynamic-block NVFP4 W4A4 and W4A16. Unsupported state topologies fail explicitly.

## Validation
- 12 focused export/unit tests passed
- ruff check and format passed
- aggregate implementation review completed with no material findings

### DIFF
--- a/modelopt/torch/export/__init__.py
+++ b/modelopt/torch/export/__init__.py
@@ -21,6 +21,7 @@ from .model_config_export import *
 from .model_utils import *
 from .moe_utils import *
 from .plugins import *
+from .quantized_weight import *
 from .registry import *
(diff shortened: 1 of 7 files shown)


mxinO:
Closing this draft while the real-quant refit design is reworked and validated locally. The branch is retained for reference.

PR #2092, opened 2026-08-06
Add id/ tied_weight_names based export shared weight deduplication logic  and remove legacy data_ptr approach 

juhi10071998:
### What does this PR do?

**Type of change:** Bug fix / robustness

**Fixes NVBug [6525352](https://nvbugspro.nvidia.com/bug/6525352)** — MiniMax-M2.7
`nvfp4_mlp_only-kv_fp8` failed at TensorRT-LLM load with `assert w1_weight is not None and
w3_weight is not None`, because the previous `data_ptr()`-only postprocess dedup could falsely
drop an independent MoE expert weight.

Rework tied-weight dedup during unified HF checkpoint export so it no longer depends on tensor
addresses. Ties are detected by **object identity before packing**, recorded as **names**, and the
duplicate is dropped **by name** on the final state dict.

**Why the old `data_ptr()` approach was wrong — it read the tie signal at the wrong time.**

```text
pack each module  ->  weight = new packed Parameter (shared object destroyed)  ->  postprocess: dedup by value.data_ptr()  ->  save

  x  false positive : a freed address is reused by an unrelated weight  ->  a real weight is dropped   (NVBug 6525352)
  x  false negative : FSDP gather / offload moves a tied weight to a new address  ->  tie missed, both copies written

root cause: the address is read AFTER packing severs it and gather/offload moves it, when it no longer reflects the real tie.
```

**The new flow — observe identity once, at the only moment it is true, and carry it forward as a name.**

```text
build TiedWeightMap (pre-pack, resident) : group params by id()  ->  { alias_name : canonical_name }   (declaration only labels the canonical)
  ->  sync_tied_input_amax   : merge input amaxes across the tie
  ->  pack every module      : tie severed -> distinct, byte-identical tensors
  ->  postprocess (BY NAME)  : drop each tie's own exported keys (dense = weight + scales; MoE = per-expert keys of the tied projections; atomic)
  ->  storage backstop       : collapse undeclared same-storage shares
  ->  safetensors.save_file  ->  loader re-ties via _tied_weights_keys

key: id() is read ONCE pre-pack and kept as NAMES -> names survive packing / FSDP / offload, so the drop never needs the packed tensors to still be the same object.
```

**Design in words.**

- **Object identity detects the tie** — `_build_tied_alias_map` groups parameters by
  `id(parameter)` while the model is resident (before packing replaces shared Parameters with
  independent tensors). A declared-but-*unapplied* tie (e.g. `lm_head` ↔ `embed_tokens` with
[...]

PR #2194, opened 2026-08-14
Name-based tied-weight deduplication during HF checkpoint export (supersedes #2092)

juhi10071998:
### What does this PR do?

**Type of change:** Bug fix / robustness

**Fixes NVBug [6525352](https://nvbugspro.nvidia.com/bug/6525352)** — MiniMax-M2.7
`nvfp4_mlp_only-kv_fp8` failed at TensorRT-LLM load with `assert w1_weight is not None and
w3_weight is not None`, because the previous `data_ptr()`-only postprocess dedup could falsely
drop an independent MoE expert weight.

> Re-opens #2092, which was closed by accident. Same code, no new changes; cleanly mergeable
> against current `main` (verified test merge, no conflicts).

Rework tied-weight dedup during unified HF checkpoint export so it no longer depends on tensor
addresses. The tie map is now sourced from HuggingFace's **name-based**
`model.all_tied_weights_keys` (transformers >= 5.0), and the duplicate is dropped **by name** on
the final state dict — so the drop never depends on the packed tensors still being the same object.

**Why the old `data_ptr()` approach was wrong — it read the tie signal at the wrong time.**

```text
pack each module  ->  weight = new packed Parameter (shared object destroyed)  ->  postprocess: dedup by value.data_ptr()  ->  save

  x  false positive : a freed address is reused by an unrelated weight  ->  a real weight is dropped   (NVBug 6525352)
  x  false negative : FSDP gather / offload moves a tied weight to a new address  ->  tie missed, both copies written

root cause: the address is read AFTER packing severs it and gather/offload moves it, when it no longer reflects the real tie.
```

**The new flow — take HF's own name-based tie map and carry it through export.**

```text
TiedWeightMap(model) : read model.all_tied_weights_keys  ->  { alias_name : canonical_name }   (HF-resolved at load, config-gated, torch.equal-pruned)
  ->  sync_tied_input_amax   : merge input amaxes across the tie
  ->  pack every module      : tie severed -> distinct, byte-identical tensors
  ->  postprocess (BY NAME)  : drop each tie's own exported keys (dense = weight + scales; MoE = per-projection keys; atomic all-or-none), guarded by torch.equal
  ->  storage backstop       : collapse undeclared same-storage shares (and covers transformers < 5.0, which lacks the map)
  ->  safetensors.save_file  ->  loader re-ties via _tied_weights_keys
```

**Design in words.**

- **HF resolves the tie, we reuse it.** `model.all_tied_weights_keys` (transformers >= 5.0) is a
  `{target: source}` == `{alias: canonical}` dict, resolved at `post_init`, gated on
  `config.tie_word_embeddings`, and `torch.equal`-pruned during `from_pretrained`. Because it is
  **names**, it survives packing / FSDP shard / offload — where a `data_ptr` would not. No pre-pack
  `id()` capture, no self-built map.
- **The drop is by name**, in `postprocess_state_dict`, **atomically per module prefix** (all of an
  alias's exported keys, or none, and only when every key has a canonical counterpart) — so tied
  sides with different quant state never orphan a `weight_scale` / `input_scale`, and an untied
  sibling under an alias prefix is never dropped. A `torch.equal` check guards each drop.
- **`data_ptr` survives only as a backstop**, keyed on safetensors' own shared-storage identity.
  It collapses *undeclared* same-storage shares that `save_file` would otherwise reject, and is the
  net for transformers < 5.0 (no `all_tied_weights_keys`).
- **transformers < 5.0**: the map is empty and a warning is emitted; declared ties fall back to the
  address backstop (resident export only). Upgrade to >= 5.0 for name-based dedup under FSDP/offload.

**What changed**

- `TiedWeightMap` (`model_utils.py`) now reads `all_tied_weights_keys`; removed the id-based
  `_build_tied_alias_map` / self-built map path (net −200 lines across the export utils).
- `postprocess_state_dict` (`quant_utils.py`): name-based atomic drop + `torch.equal` guard; MoE
  alias-group keying fixed so one alias container tying to two canonicals no longer collides.
- Unified export wires `TiedWeightMap(model)` unconditionally; FSDP path gathers via
  `get_model_state_dict(full_state_dict=True)` before the name-based drop.
- Streaming export left on the address backstop (TODO noted — needs offload validation).
- Tests: HF-contract test (guarded `importorskip transformers>=5.0`), self-entry filter, MoE
  two-canonical collision, FSDP2 shard-survival + end-to-end FSDP export (embedding + cross-layer
  tie). Removed the id-map unit tests.

### Testing

- DiffGemma NVFP4 (transformers 5.12.1): 0 tied-weight leaks; vLLM serve smoke passes.
- MiniMax-M2.7 (transformers 4.57.6, no `all_tied_weights_keys`): empty map, backstop handles it;
  15872 experts exported, no false drop (original NVBug repro now passes).
- GPU: FSDP2 tie-map survives sharding; end-to-end FSDP export dedups embedding + cross-layer ties.

## Before your PR is "Ready for review"

- [x] Make sure you read and follow [Contributor guidelines](https://github.com/NVIDIA/Model-Optimizer/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you add or update any necessary documentation?
- [ ] Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)? — yes.

## Additional Information

Supersedes accidentally-closed #2092. cc @Fridah-nv @cjluo-nv @Edwardf0t1

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Hugging Face checkpoint exports for tied weights, preventing unrelated weights that share memory from being omitted.
  * Safely removes duplicate entries while preserving matching tensor values and independent expert weights.
  * Enhanced tied-weight handling across quantized, fused-expert, sharded, and offloaded models.
  * Improved synchronization of tied input quantization values during export.
  * Preserved fallback behavior for older Transformers versions and undeclared shared storage.

* **Tests**
  * Added comprehensive coverage for tied-weight exports across distributed, offloaded, quantized, and fused-expert scenarios.


CHANGELOG.rst
```diff
@@ -6,33 +6,26 @@ Changelog
...
 
 **Bug Fixes**
 
+- Fix tied-weight deduplication during HF checkpoint export. The previous ``data_ptr()``-keyed dedup misfired -- a recycled address can falsely alias unrelated weights, and the FSDP gather / offload materializes tied weights at distinct addresses so a genuine tie is missed. The tie map is now sourced from HuggingFace's name-based ``model.all_tied_weights_keys`` (transformers>=5.0), which survives FSDP shard / offload; the duplicate is dropped by name in ``postprocess_state_dict``, guarded by a ``torch.equal`` check. A ``data_ptr`` backstop is retained for undeclared same-storage shares (and for transformers<5.0, which lacks the map).
```
meenchen: Can we keep the change log at a high level? Focus on the impact of the bug, .e.g, use cases, affected models, etc.

modelopt/torch/export/moe_utils.py
```diff
@@ -100,19 +69,10 @@ def _export_fused_experts(
...
-    module sharing the same source memory. ``_tied_cache`` (int-keyed) is
-    threaded through to the per-projection ``_export_quantized_weight``
-    calls so wrapper-level dedup uses the same scope as standalone Linears.
-    Both caches are owned by the caller (typically
-    ``_export_transformers_checkpoint``) and scoped to one export
-    invocation; when ``None`` the corresponding alias step is skipped.
+    Tied experts are not deduped here: when multiple fused-expert modules share their
+    3-D source params via HF ``_tied_weights_keys``, each is split and packed
+    independently to byte-identical per-expert tensors, and the duplicate keys are
+    dropped by name in ``postprocess_state_dict`` (the single dedup authority).
```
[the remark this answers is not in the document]
juhi10071998: this is already verified and discussed to not be a bottleneck, but rather kept so a single source of truth can be used. And dropping in `postprocess_state_dict` makes sense when there is both set of alias and canonical keys present in state_dict.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -989,6 +952,9 @@ def _export_transformers_checkpoint(
         NotImplementedError: if the model has accelerate offload hooks.
     """
     dtype = _resolve_export_dtype(model, dtype)
+    # One tied-weight map for the whole export (amax sync + final dedup in postprocess_state_dict).
+    # Sourced from HF's name-based all_tied_weights_keys, so it is correct even under FSDP/offload.
+    tied_map = TiedWeightMap(model)
```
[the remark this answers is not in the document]
juhi10071998: already added the warning for transformers version

modelopt/torch/export/hf_export_handlers.py
```diff
@@ -45,7 +45,9 @@ def _export_weight(
     # install the built-in handlers while retaining this legacy helper's import path.
     from .unified_export_hf import _export_quantized_weight
 
-    _export_quantized_weight(module, ctx.dtype, weight_name, _tied_cache=ctx.tied_cache)
+    # Dense tied weights are not deduped at pack time: both sides pack independently
+    # and the duplicate is dropped by name in postprocess_state_dict.
```
chadvoegele: [nit] remove comments related to the prior behavior

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -260,6 +260,10 @@ def _export_transformers_checkpoint_streaming(
     # Only apply when tie_word_embeddings=True: _tied_weights_keys can list keys whose
     # weights are not actually shared (e.g. if the model was saved with tie_word_embeddings=False
     # but the attribute was never cleared), which would incorrectly drop lm_head.weight.
+    #
+    # TODO(tied-map): the resident path reads HF's ``all_tied_weights_keys`` (covers dict-style/MoE
+    # ties); this path could too, to close the streaming gap for offloaded 5.x models -- but that
+    # swap needs offload-specific validation (meta tensors, per-tensor order, disk round-trip) first.
```
meenchen, line 266: Can we add an assertion here to throw out errors here for the supported scenarios

modelopt/torch/export/model_utils.py
```diff
@@ -149,82 +154,56 @@ def get_language_model_from_vl(model) -> list[nn.Module] | None:
...
+        transformers <5.0 -> empty map (the ``data_ptr`` backstop in postprocess is the net).
+        """
+        all_tied = getattr(model, "all_tied_weights_keys", None)
+        if all_tied is None and getattr(
+            getattr(model, "config", None), "tie_word_embeddings", False
+        ):
+            warnings.warn(
+                "model.all_tied_weights_keys is unavailable (transformers <5.0); declared tied "
+                "weights are deduplicated only by the address backstop (resident export). Upgrade "
+                "to transformers>=5.0 for name-based tied-weight dedup under FSDP/offload."
```
meenchen: Can we add some user-facing warnings here? Now it is very technical, and users might not have the full context to understand what's happening here

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -260,6 +260,10 @@ def _export_transformers_checkpoint_streaming(
     # Only apply when tie_word_embeddings=True: _tied_weights_keys can list keys whose
     # weights are not actually shared (e.g. if the model was saved with tie_word_embeddings=False
     # but the attribute was never cleared), which would incorrectly drop lm_head.weight.
+    #
+    # TODO(tied-map): the resident path reads HF's ``all_tied_weights_keys`` (covers dict-style/MoE
```
chadvoegele, line 264: Is there a TODO here? Or we can skip for now?

juhi10071998: yes, I think we can skip for now, our current changes are restricted for the unified_hf_export only, this is just an FYI that it could be leveraged for streaming export albeit with more testing, not scoped as part of this fix. 

During the development i saw streaming uses the  https://github.com/NVIDIA/Model-Optimizer/blob/53ccec614f67fae4630ab1959ca5faa6434eef98/modelopt/torch/export/unified_export_hf_streaming.py#L206 which was recently added. 

Can I keep the todo, it is harmless for now.

modelopt/torch/export/unified_export_hf_streaming.py
juhi10071998: We're not changing the `unified_export_hf_streaming` behavior at all, this just let's the author know that the tied_weight_map approach could be extended to streaming as well. No behavior change to the streaming path.

Perhaps we add assertion/ more details when we adopt it here?

CHANGELOG.rst
juhi10071998: yes, addressed here- https://github.com/NVIDIA/Model-Optimizer/pull/2194/changes/e0313f567e2926b92a0dad776fc0bbd8f4ce3675

modelopt/torch/export/model_utils.py
juhi10071998: addressed here- https://github.com/NVIDIA/Model-Optimizer/pull/2194/changes/e0313f567e2926b92a0dad776fc0bbd8f4ce3675

modelopt/torch/export/hf_export_handlers.py
juhi10071998: done- https://github.com/NVIDIA/Model-Optimizer/pull/2194/changes/e0313f567e2926b92a0dad776fc0bbd8f4ce3675

meenchen (approved):
Thanks for the fix

PR #2202, opened 2026-08-17
feat(quantization): PTQ support for Step-3.7 MoE checkpoints

Edwardf0t1:
### What does this PR do?

Type of change: New feature

Adds PTQ support for **Step-3.7** (`stepfun-ai/Step-3.7-Flash`). Follow-up to [NVBug 6518665](https://nvbugspro.nvidia.com/bug/6518665) / [OMNIML-5583](https://jirasw.nvidia.com/browse/OMNIML-5583): with the export crash fixed in #2071 the run completes, but the checkpoint it writes is silently unquantized —

```json
{"quantization": {"quant_algo": null, "kv_cache_quant_algo": "FP8", "quantized_layers": {}}}
```

Two independent causes, both from Step's `trust_remote_code` modeling code.

**1. The expert weights were invisible to quantization.** Step-3.5 and Step-3.7 ship the same custom `MoELinear`: a plain `nn.Module` holding one 3-D `weight` of `[num_experts, out_features, in_features]`, whose `forward(x, expert_id)` runs `F.linear` against the selected slice. It is not an `nn.Linear`, and the weights sit on the projection submodule rather than on the expert container, so neither the plain-linear path nor `_fused_experts_wrapper_class` (which wants a 3-D `down_proj` *Parameter*) claims it.

The `_QuantMoELinear` wrapper that handles exactly this layout has existed since #1063, but its registration was gated on the Step-3.5 class names:

```python
if type(model).__name__ not in ("Step3p5ForCausalLM", "Step3p5Model"):
    return
for module in model.modules():
    if type(module).__name__ == "Step3p5MoEMLP":
```

Step-3.7's root is `Step3p7ForConditionalGeneration` and its container is `Step3p7MoEMLP`, so it returned immediately and no expert ever got a quantizer. Detection is now **structural** — a 3-D `weight` plus `num_experts` / `in_features` / `out_features` and a two-positional-argument forward — so any Step revision (or another model shipping this layout) is picked up without a third hardcoded name. `_reconstruct_fused_moe_linear` likewise matches the wrapper type instead of the generated `QuantMoELinear` class name; a model whose class is spelled differently would otherwise quantize fine but export unusable per-expert keys.

**2. Step's module names don't match the general recipes.** The MoE block is `moe` and the dense sibling is `share_expert`, so `*.experts.*`, `*block_sparse_moe*` and `*mlp*` reach none of the routed experts. This PR ships `huggingface/step3p7/ptq/nvfp4_experts_only-kv_fp8_cast` and `huggingface/step3p7/ptq/nvfp4_mlp_only-kv_fp8`, which select `*moe*` and disable the router (`moe.gate`) and the shared expert — mirroring the existing Step-3.5 recipe — and documents the naming trap in `modelopt_recipes/ptq.md`.

### Usage

```bash
python examples/hf_ptq/hf_ptq.py --model /local/Step-3.7-Flash --trust_remote_code \
    --recipe huggingface/step3p7/ptq/nvfp4_experts_only-kv_fp8_cast \
    --dataset /local/cnn_dailymail --calib_size 32 --export_path /local/Step-3.7-Flash-nvfp4
```

### Testing

- `tests/unit/torch/quantization/plugins/test_moe_linear.py` — structural detection (positive plus 2-D-weight / wrong-forward negatives), registration on a Step-3.7-shaped model, per-expert quantizers with calibrated amax, and reconstruction back to the 3-D parameter. Plus two export-dispatch tests added from review: the registry resolves a `Quant_SyntheticMoELinear` to `_export_moe_linear` (this one fails against the old name-keyed registration), and the handler fills an unrouted expert's input amax.
- `tests/unit/recipe/test_step3p7_recipes.py` — drives both shipped recipes over a model mirroring Step's real paths (`model.language_model.layers[i].{moe,share_expert,mlp}`): routed experts NVFP4-quantized per expert, router / shared expert / dense MLP / `lm_head` per recipe scope.

Ran locally (torch 2.11, transformers 5.5.4 — the version in the bug report): the two new files (12 tests) plus `tests/unit/recipe`, `tests/unit/torch/quantization/plugins/`, `tests/unit/torch/export/test_export_weight.py` and `test_export_registry.py` — 324 passed. Full `tests/unit` (minus onnx/puzzletron): 2490 passed, with 4 pre-existing `test_quant_aware_conversion.py` failures that reproduce unchanged on clean `main`.

Not run end-to-end on the real Step-3.7-Flash checkpoint (1.4 TB / 8×B200) — QA can re-run against this branch with the recipe above.

- Is this change backward compatible?: ✅ — Step-3.5 keeps working; the name gate is replaced by a superset.

### Review updates

- **Export dispatch keyed on the class name** (caught in review): registration was made class-name-independent, but `_export_moe_linear` in `hf_export_handlers.py` was still registered for the literal string `"QuantMoELinear"`, so a compatible class under another name bypassed the input-amax fallback for unrouted experts. The predicate now matches `_QuantMoELinear` through the MRO (lazy import, since the wrapper lives in the optional transformers plugin), keeping the name check as a fallback so the synthetic stand-ins in `test_export_registry.py` / `test_export_weight.py` still match. This was the same class-name coupling already fixed in `_reconstruct_fused_moe_linear`, in the file I hadn't looked at.
- Canonical 2026 license header on the new test file.
- Recipe comments corrected: `*moe*` matches the router, but **not** `share_expert` (`layers.N.share_expert.*` contains no `moe` segment) — that entry is an explicit guard, not an override.
- `ptq.md` recommendation scoped to Step-3.7, since Step-3.5 has its own recipe.

### Additional Information

Pairs with #2203 (fail fast when a quant config matches no weight quantizer), which turns this class of silent no-op into an error for any model. Independent branches; either can merge first.

## Summary by CodeRabbit

- **New Features**
  - Added post-training quantization (PTQ) support for Step-3.7 Flash models, including per-expert quantization for routed MoE layers.
  - Added NVFP4 recipes for expert-only and MLP-only quantization, with optional FP8 KV-cache support.

- **Bug Fixes**
  - Improved detection, quantization, reconstruction, and export of expert-indexed MoE layers across Step model revisions.
  - Added safeguards to prevent quantization of unsupported offloaded expert weights.

- **Documentation**
  - Clarified Step-3.7 recipe selection, calibration guidance, and quantization exclusions.

- **Tests**
  - Added coverage for expert, MLP, router, attention, shared-expert, export, and calibration behavior.


modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -1921,42 +1921,77 @@ def forward(self, x, expert_id):
...
+        for p in params
+        if p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD) and p.default is p.empty
+    ]
+    return len(positional) == 2
+
+
+def register_moe_linear_on_the_fly(model):
+    """Register expert-indexed ``MoELinear`` modules (Step-3.5 / Step-3.7) for quantization.
+
+    Without this the routed experts carry no quantizer at all: an experts-only recipe matches
```
cjluo-nv, line 2044: > _Bot comment._

The registration is now structural/class-name-independent, but the corresponding export handler in `modelopt/torch/export/hf_export_handlers.py` is still registered only for the string `"QuantMoELinear"`. A compatible class with another name—and even a second remote `MoELinear` type when that generated name is already occupied—gets a differently named dynamic class and will bypass `_export_moe_linear`, leaving unrouted experts without its input-amax fallback. Please make that handler match `isinstance(module, _QuantMoELinear)` (or an equivalent structural predicate), and add an export-path test using the differently named synthetic class. The current test calls `_reconstruct_fused_moe_linear` directly, so it cannot catch this.

tests/unit/torch/quantization/plugins/test_moe_linear.py
```diff
@@ -0,0 +1,181 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv: > _Bot comment._

This new file uses a 2024 copyright header, while the repository's canonical `LICENSE_HEADER` currently says 2026. Please use the canonical header (or have the licensing owner explicitly confirm this divergence).

cjluo-nv (commented):

The Step-3.7 registration and recipes are well tested at conversion time, but the class-independent support is incomplete in the export path: export dispatch still keys specifically on `QuantMoELinear`, so generated wrapper names that collide or originate from differently named compatible classes bypass the missing-amax preparation. The new reconstruction test does not exercise that dispatch. Also, one new file's NVIDIA header differs from the repository's canonical 2026 `LICENSE_HEADER`, so licensing needs correction or human sign-off.

Edwardf0t1:
Thanks — the export-dispatch finding was right and is fixed in fbfb880, along with the three smaller ones.

**Export handler keyed on the class name.** Confirmed: `hf_export_handlers.py` registered `_export_moe_linear` for the literal string `"QuantMoELinear"`, so a compatible remote-code class under any other name produced a differently named dynamic class and skipped the input-amax fallback. This is the same class-name coupling I'd already fixed in `_reconstruct_fused_moe_linear` — I just didn't look in that file. The predicate now matches `_QuantMoELinear` through the MRO:

```python
try:
    from modelopt.torch.quantization.plugins.huggingface import _QuantMoELinear
    if isinstance(module, _QuantMoELinear):
        return True
except ImportError:
    pass
return any(cls.__name__ == "QuantMoELinear" for cls in type(module).__mro__)
```

The import is lazy because the wrapper lives in the optional transformers plugin (mirroring how `unified_export_hf.py` imports `_reconstruct_fused_moe_linear`). The name check is kept as a fallback, not for real models — `isinstance` is a strict superset there — but because `test_export_registry.py:210` and `test_export_weight.py:105` dispatch synthetic stand-ins named `QuantMoELinear` that aren't real wrappers. Dropping it would break both.

**Export-path tests added**, as asked. `test_export_handler_matches_a_differently_named_wrapper` asserts the registry resolves a real `Quant_SyntheticMoELinear` to `_export_moe_linear`; I verified it is not vacuous by reverting to the name-keyed registration, where it fails. `test_export_handler_fills_input_amax_for_unrouted_experts` covers what the handler is for: an expert whose amax was reset gets one back.

**License header** — fixed to the canonical 2026 `LICENSE_HEADER`.

**Recipe comments** — you're right, and the comment was wrong in a way worth correcting: Step's shared expert is `layers.N.share_expert.*` with no `moe` segment, so `*moe*` never matches it. Both recipes now say the router is matched by `*moe*` (hence must be disabled last) while `share_expert` is an explicit guard.

**`ptq.md`** — scoped to Step-3.7; Step-3.5 has its own recipe directly above.

All 54 checks were green before these changes; re-running now. Local: `tests/unit/torch/export` + `tests/unit/torch/quantization/plugins` + `tests/unit/recipe` on transformers 5.5.4, 512 passed, with the 4 known `test_quant_aware_conversion.py` failures that reproduce unchanged on clean `main`.

tests/unit/torch/quantization/plugins/test_moe_linear.py
```diff
@@ -0,0 +1,222 @@
...
+    """Export dispatch must key on the wrapper type, not the generated class name.
+
+    The registration is structural, so a compatible remote-code class can be named
+    anything; its generated class is then ``Quant<ThatName>``. If the export handler only
+    matched the literal name ``QuantMoELinear``, such a module would skip
+    ``_export_moe_linear`` and export without the input-amax fallback for experts that
+    calibration never routed to.
+    """
+    from modelopt.torch.export.hf_export_handlers import _export_moe_linear
+    from modelopt.torch.export.registry import ExportModuleRegistry
```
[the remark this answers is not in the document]
Edwardf0t1: Fixed in 1624be5. Neither import is optional or circular here — the file already imports `modelopt.torch.quantization` at module scope — so both are now at the top of the file.

modelopt_recipes/huggingface/step3p7/ptq/nvfp4_mlp_only-kv_fp8.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - quantizer_name: '*mlp*input_quantizer'
+      cfg:
+        $import: nvfp4
+    - $import: kv_fp8
+    - $import: default_disabled_quantizers
+    # Router and shared expert are matched by `*moe*` above; disable them last (later wins).
+    - quantizer_name: '*moe.gate.*'
+      enable: false
+    - quantizer_name: '*share_expert*'
+      enable: false
```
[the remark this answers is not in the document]
Edwardf0t1: Correct, and the comment was wrong in a way worth fixing: Step's shared expert is `layers.N.share_expert.*`, which contains no `moe` segment, so `*moe*` never matches it. Both recipes now state that the router is matched by `*moe*` (hence the disable must come last, later-wins) while `share_expert` is an explicit guard so it stays BF16 if the patterns are ever widened. Fixed in fbfb880.

modelopt_recipes/ptq.md
```diff
@@ -295,6 +295,15 @@ A lighter case: **`step3p5/Step3.5-Flash/ptq/nvfp4-mlp-only`** is close to
...
 (`share_expert`, `moe.gate`, the conv1d branches).
 
+**`step3p7/ptq/{nvfp4_experts_only-kv_fp8_cast,nvfp4_mlp_only-kv_fp8}`** are the
+Step-3.7 equivalents, and the reason they exist is **module naming**: Step calls
+the MoE block `moe` and the dense sibling `share_expert`, so the general
+recipes' `*.experts.*`, `*block_sparse_moe*` and `*mlp*` patterns match nothing
+on the routed experts — the general recipe would quantize *nothing* and export a
+checkpoint with `quant_algo: null`. These select `*moe*` instead and disable the
+router (`moe.gate`) and `share_expert` on top. Use them, not the general
+recipes, for any Step checkpoint.
```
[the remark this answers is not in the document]
Edwardf0t1: Agreed — Step-3.5 has its own recipe in the paragraph directly above, so recommending these for "any Step checkpoint" was wrong. Reworded to your suggestion in fbfb880.

modelopt/torch/quantization/plugins/huggingface.py
Edwardf0t1: Fixed in fbfb880. You were right that the export path still keyed on the name; the predicate now matches `_QuantMoELinear` through the MRO (lazy import, since the wrapper is in the optional transformers plugin). The literal-name check is kept only because `test_export_registry.py:210` and `test_export_weight.py:105` dispatch synthetic stand-ins named `QuantMoELinear` that are not real wrappers — for real models `isinstance` is a strict superset. Two export-path tests added: one asserts the registry resolves a real `Quant_SyntheticMoELinear` to `_export_moe_linear` (verified non-vacuous — it fails against the old name-keyed registration), one asserts the handler fills an unrouted expert's input amax.

tests/unit/torch/quantization/plugins/test_moe_linear.py
Edwardf0t1: Fixed in fbfb880 — now the canonical 2026 `LICENSE_HEADER`. (`tests/unit/recipe/test_step3p7_recipes.py`, the other new file, already had it.)

cjluo-nv (commented):

Re-review complete. The prior critical issues are addressed: export dispatch now recognizes `_QuantMoELinear` wrappers independently of generated class names and has regression coverage for both registry dispatch and unrouted-expert amax fallback; the new test header matches the canonical 2026 `LICENSE_HEADER`; and the local test imports were moved to module scope. The recipe comments and Step-3.7 documentation scope were also corrected.

Design review: the problem is that Step’s remote-code expert projections are neither `nn.Linear` nor the existing fused-expert-container layout, so they were skipped by quantization. The relevant alternatives are extending the prior Step-3.5 name-gated registration, adapting the existing `_fused_experts_wrapper_class` path, or requiring users to manually register custom modules through the existing quantization registry. The PR body explains why the normal linear and fused-container paths cannot claim this projection layout and reasonably chooses to generalize the existing `_QuantMoELinear` registration structurally; PyTorch introspection plus the existing registry is more appropriate here than the installed configuration libraries such as OmegaConf/Hydra. Tests cover detection, rejection cases, conversion/calibration, reconstruction, export dispatch, amax fallback, and both recipes. The new-file NVIDIA headers match the repository canonical header, so the standard-header licensing exception applies.

Complex PR: spans 5 directories (≥ 5). Looping in a human for approval.

Edwardf0t1:
CI caught a staleness break, fixed in 969ebcb (branch also rebased onto current `main`).

`test_shipped_ptq_recipe_algorithm_config_constructs` failed on both new recipes with `ValidationError for MaxCalibConfig`. Cause: this branch predates the removal of the legacy `layerwise` bool form, so the recipes copied `layerwise: false` from the general recipes as they were spelled at the time. They now use `layerwise: {enable: false}` like the current general recipes. It only surfaced in CI because CI validates against merged `main`, while the branch checkout still had the old schema.

Local after rebase (torch 2.11, transformers 5.5.4): `tests/unit/recipe` + `tests/unit/torch/quantization/plugins` + `tests/unit/torch/export` — 640 passed, with the 4 known `test_quant_aware_conversion.py` failures that reproduce unchanged on clean `main`.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -1923,42 +1923,77 @@ def forward(self, x, expert_id):
...
+        if mod_type in visited_types or QuantModuleRegistry.get(mod_type) is not None:
+            continue
+        visited_types.add(mod_type)
+
+        if _is_expert_indexed_moe_linear(module):
+            print(
+                f"\033[1mDetected expert-indexed MoE linear '{name}' of type "
+                f"{mod_type.__name__}, registering with _QuantMoELinear.\033[0m"
+            )
+            QuantModuleRegistry.register({mod_type: f"hf.{mod_type.__name__}"})(_QuantMoELinear)
```
meenchen, line 2061: **[P1] Preserve or reject Accelerate-offloaded `MoELinear` weights before conversion.** This new registration routes Step-3.7 modules into `_QuantMoELinear._setup()`. When `device_map="auto"` places one on CPU or disk, `self.weight` can be a meta tensor while the real value remains in the Accelerate hook under the original `weight` key. `_setup()` then creates meta `experts.N.weight` parameters, deletes `self.weight`, and leaves the hook pointing to the deleted key. I reproduced a correct `[8, 26, 44, 62]` forward becoming `[0, 0, 0, 0]` immediately after `mtq.quantize`, even with every quantizer disabled. This is reachable through the supported/default auto-device-map and `--offload_folder` loading path. Please either materialize and retarget the offload mapping during conversion, or reject offloaded/meta `MoELinear` modules before calibration, and add a CPU/disk-offload regression test.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -1923,42 +1923,77 @@ def forward(self, x, expert_id):
...
+    try:
+        params = list(inspect.signature(type(module).forward).parameters.values())[1:]
+    except (TypeError, ValueError):
+        return False
+    positional = [
+        p
+        for p in params
+        if p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD) and p.default is p.empty
+    ]
+    return len(positional) == 2
```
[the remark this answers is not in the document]
meenchen: **[P1] This contract issue also has a concrete grouped-routing collision that exact shape validation will not catch.** Moondream3 `MoeFusedLinear` has exactly `[num_experts, out_features, in_features]`, all three attributes, and `forward(input, m_sizes)`, where `m_sizes` is an E-element tensor used by grouped GEMM. This predicate registers it globally; `_QuantMoELinear` then evaluates `self.experts[m_sizes]`, and the first calibration forward raises `TypeError: only integer tensors of a single element can be converted to an index`. I reproduced this on the current head with quantizers disabled. Please scope registration to the Step classes/family, or otherwise require a verifiable scalar `expert_id` contract, and add a grouped-routing negative test. Example source: https://huggingface.co/NyxKrage/moondream3-preview-hf/blob/ca700c75b6a14556664f9a6de109bb2ef3d8ca9f/moondream3_moe_fused/moe_fused_linear.py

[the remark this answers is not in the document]
Edwardf0t1: Reproduced and fixed in 727bcfa. You're right that shape alone is not the contract — the wrapper's forward indexes `self.experts[expert_id]`, so it only holds for a scalar index, and a grouped-GEMM layer with the same 3-D weight and the same three attributes satisfies my predicate while passing a token-count tensor. I built the Moondream3 `MoeFusedLinear` shape locally: the predicate claimed it, and `ModuleList[tensor([1,2,1,0])]` gives exactly the `TypeError: only integer tensors of a single element can be converted to an index` you quoted.

I took the second of your two options — requiring a verifiable scalar contract rather than re-scoping to the Step classes — because re-scoping by class name is the coupling this PR set out to remove (Step-3.7 was skipped precisely because the gate named Step-3.5's classes). The predicate now additionally requires the second positional parameter to be named `expert_id`, which is what both Step revisions declare and what `m_sizes` fails. The tradeoff is explicit: a future Step revision that renames the parameter falls back to being unclaimed — the pre-PR behaviour — rather than being claimed and crashing.

`test_grouped_routing_module_is_not_claimed` covers it with a `forward(input, m_sizes)` module of the exact Moondream3 shape. If you'd rather have the Step-family scoping as well as (or instead of) the parameter-name contract, say so and I'll switch.

modelopt/torch/quantization/plugins/huggingface.py
Edwardf0t1: Reproduced and fixed in 727bcfa. Confirmed the mechanism end to end with `accelerate.cpu_offload`: after offload `weight.is_meta` is True, the module still forwards correctly through the hook, and `mtq.quantize` then walks into `_setup()`. On my versions it surfaces as `RuntimeError: Tensor.item() cannot be called on meta tensors` rather than the `[0, 0, 0, 0]` you saw — same root cause, and silent corruption is clearly reachable on the disk-offload path where the copy succeeds.

Conversion now refuses before doing anything, with an actionable message pointing at the offload as the cause. I chose refusal over materialize-and-retarget deliberately: rewiring the hook's `weights_map` from `weight` onto `experts.N.weight` is the richer fix but I cannot validate it without a real offloaded multi-GPU setup, and shipping an unvalidated remap of the exact mechanism that was corrupting weights seemed worse than a clear refusal. Happy to do the remap as a follow-up if you'd like it, ideally with someone who can run the offloaded path.

`test_offloaded_weights_are_refused_not_silently_corrupted` is the regression test — it runs `accelerate.cpu_offload` on the projection and asserts `mtq.quantize` raises.

Worth noting this is pre-existing behaviour of `_QuantMoELinear` from #1063 rather than something this PR introduces, but the PR does make it newly reachable for Step-3.7, so it belongs here.

modelopt/torch/quantization/plugins/huggingface.py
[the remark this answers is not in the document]
Edwardf0t1: Both valid, fixed in 0f66926. The signature check counted only *required positional* parameters, so `forward(self, x, expert_id, *, router_state=None)` and `**kwargs` forwards were claimed even though the replacement forward takes exactly `(x, expert_id)` — it now requires the signature to be exactly that. And the 3-D weight was accepted in any layout, so a `[num_experts, in_features, out_features]` module would have rebuilt each expert from the wrong slices; the shape is now checked against `(num_experts, out_features, in_features)`. Two tests added for the transposed layout and the extra-parameter forms. These stack with the `expert_id` name contract added in 727bcfa for @meenchen's grouped-routing finding.

CHANGELOG.rst
```diff
@@ -30,6 +30,8 @@ Changelog
...
 - Add ``layerwise.export_dir``: layerwise calibration writes each decoder layer to its own quantized checkpoint shard as it finishes, so no separate ``export_hf_checkpoint()`` pass is needed and, with ``layerwise.checkpoint_dir``, an interrupted run resumes without redoing finished layers. Supports FP8 and NVFP4 on single-process models, resident or offloaded; other formats and placements raise ``NotImplementedError`` before calibration starts.
 
+- Add PTQ support for Step-3.7 (``stepfun-ai/Step-3.7-Flash``). The expert-indexed ``MoELinear`` modules these checkpoints ship via ``trust_remote_code`` are now detected structurally instead of by model class name, so the routed experts get per-expert quantizers on any Step revision. Quantize with the new ``huggingface/step3p7/ptq/nvfp4_experts_only-kv_fp8_cast`` or ``huggingface/step3p7/ptq/nvfp4_mlp_only-kv_fp8`` recipes — the general recipes select experts by module names Step does not use and would leave the model unquantized.
```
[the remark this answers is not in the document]
Edwardf0t1: Agreed — the repo's changelog guidance is explicit that entries are for external users and should leave out implementation detail, and mine led with `MoELinear` / `trust_remote_code` / structural detection. Rewritten in 0f66926 to the feature, the recipes to use, and why not the general ones.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -1923,42 +1936,88 @@ def forward(self, x, expert_id):
...
+    try:
+        params = list(inspect.signature(type(module).forward).parameters.values())[1:]
+    except (TypeError, ValueError):
+        return False
+    # The replacement forward is exactly `(x, expert_id)`, so anything the caller could pass
+    # beyond those two — a keyword-only `router_state`, *args, **kwargs — would raise once
+    # converted. Require the signature to match what the wrapper can honour.
+    return (
+        len(params) == 2
+        and all(p.kind is p.POSITIONAL_OR_KEYWORD and p.default is p.empty for p in params)
```
cjluo-nv, line 2013: > _Bot comment._

`expert_id` is only a parameter name; it does not prove the semantics this replacement assumes. Because this plugin runs globally, any module with the same attributes/signature but per-expert bias, scaling, or another post-op is claimed, and `_QuantMoELinear.forward` silently drops that behavior. Please scope discovery to the Step family without hard-coding a revision class, or preserve/verify the original forward semantics, and add a same-signature non-Step negative regression.

cjluo-nv (commented):

Changes requested: global structural registration can still replace unrelated modules whose behavior is incompatible with `_QuantMoELinear`.

Needs action:
- [ ] Fix the overly broad predicate in `huggingface.py` and add the negative regression described inline.
- [ ] Add an optional-dependency justification above the function-local `accelerate` import in `test_moe_linear.py`.

No action needed:
- :heavy_check_mark: Resolved since the last review: export dispatch, canonical headers, offload refusal, grouped-routing rejection, shape/signature checks, and recipe documentation.
- The design choice is justified: this extends the existing wrapper/registry because plain `nn.Linear` and `_fused_experts_wrapper_class` cannot handle Step’s projection layout.

tests/unit/torch/quantization/plugins/test_moe_linear.py
```diff
@@ -0,0 +1,295 @@
...
+def test_offloaded_weights_are_refused_not_silently_corrupted():
+    """Accelerate offload leaves `weight` on meta, with the value in the module's hook.
+
+    Expanding that would copy meta storage into each expert and delete the key the hook
+    restores into, exporting a checkpoint of zeros. Conversion must refuse instead.
+    """
+    pytest.importorskip("accelerate")
+    from accelerate import cpu_offload
+
+    model = _TinyStepModel()
```
cjluo-nv, line 256: > _Bot comment._

Keeping this import local is appropriate because `accelerate` is optional, but please add a brief comment immediately above it stating that reason, as required for function-local Python imports.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -1923,42 +1936,88 @@ def forward(self, x, expert_id):
...
+        if mod_type in visited_types or QuantModuleRegistry.get(mod_type) is not None:
+            continue
+        visited_types.add(mod_type)
+
+        if _is_expert_indexed_moe_linear(module):
+            print(
+                f"\033[1mDetected expert-indexed MoE linear '{name}' of type "
+                f"{mod_type.__name__}, registering with _QuantMoELinear.\033[0m"
+            )
+            QuantModuleRegistry.register({mod_type: f"hf.{mod_type.__name__}"})(_QuantMoELinear)
```
meenchen, line 2061: **[P1] Preserve Step’s FP32 `MoELinear` computation.** This registration newly routes Step-3.7 through `_QuantMoELinear`, whose `forward` casts `x` down to the stored BF16 weight dtype before invoking `nn.Linear`. The original Step implementation explicitly evaluates `F.linear(x.float(), weight.float())`; therefore conversion changes the model even when every quantizer is disabled, and the down projection rounds its FP32 activation before the input quantizer observes it. I reproduced non-identical output on this head with a BF16 4096→1280 projection and disabled quantizers (max absolute error 0.0155). Please apply the input/weight quantizers but execute the linear in FP32, preserve output-quantizer behavior, and add a disabled-quantizer parity test using BF16 weights.

meenchen (changes requested):
Re-reviewed the latest head. The prior Accelerate-offload corruption and grouped-routing false-positive findings are addressed. One remaining correctness issue is noted inline.

modelopt/torch/quantization/plugins/huggingface.py
Edwardf0t1: Reproduced and fixed in 9a573bb. You're right — the original `MoELinear.forward` promotes to fp32 unconditionally (`F.linear(x.float(), self.weight[expert_id].float())`), but the wrapper's forward was downcasting the fp32 activation to the expert's stored weight dtype (bf16 on real checkpoints) before the matmul. I reproduced ~0.5 max absolute error on a bf16 4096→1280 projection with every quantizer disabled — a bigger delta than yours, likely because my activation was also bf16 rather than fp32 going in.

Fix: expert weights are now expanded in fp32 in `_setup` (matching what Step's own forward promotes storage to at call time) instead of the original storage dtype, and `forward` upcasts `x` to fp32 to match instead of downcasting to the weight's dtype. Verified bit-exact (`torch.equal`, not `allclose`) against the reference after the fix.

New test: `test_disabled_quantizers_reproduce_bf16_weight_fp32_compute_parity` — bf16 weight, bf16 activation, quantizers disabled, asserts exact equality with the unconverted reference, and asserts conversion actually happened (so the test can't pass vacuously by skipping conversion).

modelopt/torch/quantization/plugins/huggingface.py
Edwardf0t1: Fixed in 9a573bb by gating registration on the Step family, which is the first of your two options. `_is_expert_indexed_moe_linear` still does the shape/signature check (it's still needed to pick the right module *within* a Step model — the router, share_expert, and other Linears in the tree don't have this shape), but `register_moe_linear_on_the_fly` now also requires the root model to match the Step family: `model_type` or class name against `step<digit>` — not `Step3p5ForCausalLM`/`Step3p7ForConditionalGeneration` by exact name, so a future Step release is still picked up without another hardcoded name, but a structurally-identical non-Step module is not.

Added `test_non_step_model_with_identical_signature_is_not_registered`: a module with the exact `_SyntheticStepMoEMLP` shape (3-D weight, `forward(x, expert_id)`) rooted under a model with `model_type="not_step"`, asserting it is not registered. Verified non-vacuous — it fails if I remove the gate.

tests/unit/torch/quantization/plugins/test_moe_linear.py
Edwardf0t1: Fixed in 9a573bb — added the optional-dependency comment on the `accelerate` import.

Edwardf0t1:
Pushed 9a573bb addressing both P1s from the latest round (@meenchen's fp32-compute regression, @cjluo-nv's structural-detection scoping) plus the accelerate-import nit — replied in each thread with what changed.

Summary:
- **fp32 compute parity**: expert weights now expand in fp32 (matching what `MoELinear.forward` promotes storage to at call time) instead of the original bf16 storage dtype, and the wrapper's forward upcasts the activation to match instead of downcasting it. Verified bit-exact against the reference with quantizers disabled.
- **Step-family gate**: `register_moe_linear_on_the_fly` now also requires the root model to match `model_type`/class name against `step<digit>`, not an exact revision — so future Step releases stay covered without a new hardcoded name, but a structurally identical non-Step module is not registered.
- Two new regression tests, both checked non-vacuous (fail without the corresponding fix): dtype parity, and non-Step-model rejection.

Local run of the affected suites (torch 2.11, transformers 5.5.4): `tests/unit/torch/quantization/plugins` + `tests/unit/recipe` + `tests/unit/torch/export` — 646 passed, with only the 4 known `test_quant_aware_conversion.py` failures that reproduce unchanged on clean `main`. CI re-running now.

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -1908,57 +1929,129 @@ def _setup(self):
         for i in range(self.num_experts):
             experts[i].to_empty(device=device)
             with torch.no_grad():
-                experts[i].weight.data = self.weight[i].detach().to(dtype=dtype, device=device)
+                experts[i].weight.data = (
+                    self.weight[i].detach().to(dtype=torch.float32, device=device)
```
meenchen: **[P1] Keep Step expert weights in their original storage dtype.** Promoting every expert parameter to FP32 fixes the GEMM semantics, but turns Step’s temporary per-call promotion into persistent model state. Step-3.7 has 42 MoE layers × 3 projections × 288 experts × 4096 × 1280 = 190,253,629,440 routed-expert parameters, so BF16→FP32 adds approximately 354.4 GiB throughout calibration. Device placement was calculated for BF16 and offloaded weights are rejected above, so a model that loaded successfully can subsequently OOM. This also reconstructs any disabled/unquantized BF16 expert as FP32; I reproduced both the dtype change and 2× storage after reconstruction. Please retain the expert parameters in their checkpoint dtype, apply the input/weight quantizers, and promote only the selected quantized operands for the FP32 `F.linear`, followed by the output quantizer. Please also extend the BF16 parity test to assert post-conversion storage dtype and disabled-quantizer reconstruction/export dtype; the current reconstruction test starts from FP32 and masks this.

Edwardf0t1: Reproduced and fixed in 4428b39.

You're right on both counts. Kept expert storage at the checkpoint's own dtype (matching what `_setup` did before my previous fix), and moved the fp32 promotion into `forward` as a **transient** swap scoped to the one expert actually being called: `expert.weight.data = original.float()`, call `expert(x)`, restore in `finally`. That matches Step's own per-call memory profile (one expert's worth of fp32, not the whole routed-expert set) instead of expanding everything permanently. Reading `expert.weight` outside any `quantize_weight()` context means `_get_quantized_weight` passes it through unchanged — the real underlying `nn.Parameter`, same as `_setup`'s own pattern — so the `.data` reassignment genuinely persists rather than mutating a computed value.

I did not take the 'reimplement input/weight/output-quantize inline' route to avoid calling `expert(x)` at all, because that would have broken something I found while checking it: `local_hessian_calibrate` registers a `forward_pre_hook` directly on each quantized Linear module, which only fires through standard `nn.Module.__call__` dispatch. I verified this is a real risk, not hypothetical — ran `local_hessian` end to end against the swap-based `forward` and confirmed `weight_quantizer.amax` populates correctly; a version that bypassed `__call__` would have silently skipped that hook and left `local_hessian` uncalibrated for Step models.

Extended `test_disabled_quantizers_reproduce_bf16_weight_fp32_compute_parity` per your ask: asserts `experts[i].weight.dtype` stays at the checkpoint dtype after conversion, and that `_reconstruct_fused_moe_linear` also produces that dtype (not a permanently-promoted one) — verified non-vacuous by reverting to the fp32-storage version and confirming it fails. Added `test_local_hessian_calibration_fires_through_the_transient_weight_swap` to pin the hook-dispatch requirement directly.

Local (torch 2.11, transformers 5.5.4): `tests/unit/torch/quantization/plugins` + `tests/unit/recipe` + `tests/unit/torch/export` — 647 passed, only the 4 known `test_quant_aware_conversion.py` failures unrelated to this change.

Edwardf0t1:
Pushed 4428b39 fixing the memory-blowup P1 — replied in the thread with what changed and how I verified it.

Summary: the previous fix (9a573bb) traded a correctness bug for a memory one — permanent fp32 expert-weight storage instead of Step's own transient per-call promotion, ~354 GiB extra on Step-3.7's full routed-expert set. Storage now stays at the checkpoint's own dtype; `forward` swaps only the called expert's weight to fp32 for the duration of that one call, matching Step's own memory profile, then restores it. Had to keep calling `expert(x)` rather than reimplementing the quantize sequence inline, since `local_hessian_calibrate`'s forward-hook only fires through `__call__` dispatch — verified this by running `local_hessian` end to end.

Extended the parity test to assert storage dtype post-conversion and post-export-reconstruction, and added a dedicated test pinning the hook-dispatch requirement. Both verified non-vacuous. Local: 647 passed, only the 4 known unrelated failures.

cjluo-nv (commented):

Approved: prior correctness concerns are resolved, and the Step-scoped wrapper, export path, recipes, and regression coverage are coherent.

No action needed:
- :heavy_check_mark: Resolved since prior reviews: false-positive registration, offload corruption, FP32 compute parity without persistent promotion, export dispatch, canonical headers, recipe guidance, and local-import justification.
- Design is justified: extending the existing wrapper/registry fits Step’s layout; plain `nn.Linear`, fused-expert interception, and revision-specific registration do not.

Large PR: spans 5 directories (≥ 5). The review came back clean, so this is an LGTM — a human should take the final look and approve.

meenchen (approved):

PR #2207, opened 2026-08-18
FSDP2 export optimizations

sugunav14:
### What does this PR do?

  **Type of change:** Performance enhancement

  FSDP2 checkpoint export gathered the whole model onto rank 0 and wrote it from a single process. On large models that made export the slowest phase of a PTQ run and could exhaust host memory.

  The model is now split into export units — one per decoder layer, plus one for everything else that owns state — and the units are dealt round-robin across ranks. Every rank enters the unshard window for every unit, because the gather is a collective; only the owner keeps the result, packs it, and streams it to its own shard files. Rank 0 names the shards and writes the index once every rank has finished.

  Two properties follow. Peak host memory per rank is ~model/world instead of the whole checkpoint on rank 0. And packing runs on a gathered, full weight, so it is byte-for-byte the single-process code path — no amax narrowing, no block-quantized fallback, no scale-buffer heuristics.

Measured on B300 (bia), NVFP4, `hf_ptq.py --use_fsdp2`:

| model | world | export | artifact |
|---|---|---|---|
| Qwen3-235B-A22B | 8 (1 node) | **168.6s** | 146,361 keys / 16 shards / 0.122 TiB |
| oakhaven-max 4.5 TB | 64 (8 nodes) | **1359.7s** | 568,398 keys / 158 shards / 1.292 TiB |

Checkpoints were validated, not just timed: every index entry present, no leftover `__shard_part*`, all ranks exit 0. A repeat of the oakhaven configuration landed at 1411.7s, so treat ~1.36–1.41 ks as its range. No speedup ratio is quoted: the old rank-0 path was last measured on different hardware on a different day, and this cluster drifts up to 14% run-to-run on an identical tree, so a cross-day ratio would not reproduce. A same-day baseline of `main` is the outstanding item.

Two alternatives were considered and rejected. `torch.distributed.checkpoint`'s `get_model_state_dict` full-gathers to rank 0 only with no per-module owner knob, so it reproduces the exact host-memory bound this removes, and per layer it was not faster than `full_tensor()`. Packing each rank's shard in place was ~9% faster at Qwen3-235B w8 on the trees measured at the time, but rebuilt `FSDPParam` objects from torch's private FSDP internals and carried ~400 lines of format-specific shard handling; gathering to the owner deletes all of it.

Configurations that would silently corrupt a checkpoint now raise instead: FSDP2 composed with another DTensor parallelism (FSDP2 + TP on a 2-D mesh; HSDP is fine), models with no discoverable decoder layers, a decoder layer object reused across layers, and a container that holds the decoder layers while owning direct state of its own.

### Usage

No API change — `export_hf_checkpoint` selects the path automatically.

```python

# model is FSDP2-wrapped; every rank must call this (it unshards collectively)
export_hf_checkpoint(model, export_dir="./exported")
```

Testing

- `test_fsdp2_streaming_export_matches_reference` — world = GPU count under real FSDP2; merged checkpoint compared tensor-by-tensor against a single-process export of the same seeded model, index asserted to span more than one shard file. Covers the ownership split and the cross-rank index merge.
- `test_fsdp2_gathered_pack_matches_reference` — gathered packing vs whole-weight packing over nvfp4 / fp8 / int8 / fp8-per-channel-per-token / int4-blockwise
- World=1 CPU tests for unit enumeration, shard sub-splitting by `max_shard_size`, tied-alias dropping, and `extra_state_dict`
- `tests/unit/torch/utils/test_perf.py` — sampling and threshold logic of `maybe_clear_cuda_cache`, which replaces an unconditional `torch.cuda.empty_cache()` after every packed weight

Before your PR is "*Ready for review*"

Make sure you read and follow Contributor guidelines and your commits are signed (git commit -s -S).

Make sure you read and follow the Security Best Practices (e.g. avoiding hardcoded trust_remote_code=True, torch.load(..., weights_only=False), pickle, etc.).

### Additional Information

Pre-export time (load, quantizer insertion, calibration) is now the dominant phase of an FSDP2 PTQ run. It is unaffected by this PR and addressed separately in #2359.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FSDP2 Hugging Face checkpoint export by streaming rank-owned layers to shard files, reducing export time and host-memory usage for large models.
  - Improved export handling for multimodal models with missing architecture metadata.
  - Standardized weight packing across supported parallelism configurations.

- **Performance**
  - Reduced repeated module lookups and unnecessary CUDA cache cleanup.
  - Improved shard generation, merging, and index completeness.

- **Compatibility**
  - Expanded support for FSDP2 quantized export workflows and shard-local packing.


modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -1046,6 +1046,169 @@ def fsdp2_aware_weight_update(root_model, modules_to_update, reshard=True):
...
+    """
+    info = next(iter(captured.values()))
+    mesh, placements = info.mesh, info.placements
+    for bname, buf in list(module._buffers.items()):
+        if buf is None or isinstance(buf, DTensor) or buf.dim() == 0:
+            continue
+        # A buffer whose dim-0 matches the shard row/expert count is sharded; wrap it Shard(0) so a
+        # later full_tensor() rebuilds it. Even sharding, so from_local infers the global shape.
+        if buf.shape[0] == local_dim0:
+            module._buffers[bname] = DTensor.from_local(buf.contiguous(), mesh, placements)
```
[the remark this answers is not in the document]
sugunav14: It's safe today because the only 1-D buffers these modules carry are genuinely per-expert (weight_scale, and weight_scale_2 stacked to [E]), while input_scale is 0-D and dropped by the dim() == 0 filter — there's no replicated length-1 1-D buffer to mislabel. That's a property of what the current quantizers emit,

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -1046,6 +1046,211 @@ def fsdp2_aware_weight_update(root_model, modules_to_update, reshard=True):
...
+    # plain Module while this module is still sharded. Testing the root there would skip the
+    # to_local step and hand DTensors to a packer that expects plain tensors.
+    fsdp_module = _get_enclosing_fsdp_module(module, root_model)
+    if fsdp_module is None:
+        yield
+        return
+
+    group = fully_shard.state(fsdp_module)._fsdp_param_group
+    if group is None:
+        # FSDP module with no managed params of its own: nothing to pack shard-local, so let the
```
cjluo-nv: > _Bot comment._

`_get_enclosing_fsdp_module(module, root_model)` is called unconditionally here, and with `name_to_module=None` it builds `dict(root_model.named_modules())` on every call, then linear-scans it.

The CM this replaces (`fsdp2_aware_weight_update`) guarded everything behind `if isinstance(root_model, FSDPModule):`, so a non-FSDP model paid **zero** cost. Now `_export_quant_linear` / `_export_bmm_experts` run this once per quantized module on *every* export, including the single-GPU path. For a large MoE that's ~40k calls x ~60k modules — the exact O(N^2) blowup that was already fixed once in `create_fsdp_param_mapping` (see its "stalled MoE exports for hours" comment).

The `_export_quant_linear` docstring says "is a no-op for non-FSDP models -- so the single-process path is unchanged", which is true semantically but not in cost. Suggest an early `if not isinstance(root_model, FSDPModule) and not any(isinstance(m, FSDPModule) ...)` bail-out, or thread a cached `name_to_module` through `ExportContext` (both `_get_enclosing_fsdp_module` and `_get_module_name` already accept one).

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -1046,6 +1046,211 @@ def fsdp2_aware_weight_update(root_model, modules_to_update, reshard=True):
...
+        if not isinstance(param, DTensor):
+            continue
+        # A multi-dim mesh (HSDP, FSDP+TP) has to be rejected BEFORE the divisibility check below:
+        # device_mesh.size() is the product of every mesh dim, so it would inflate world and report
+        # a valid config as unevenly sharded. get_local_rank() would then raise a bare DeviceMesh
+        # error anyway. Symmetric across ranks, so neither raise can hang.
+        if param.device_mesh.ndim != 1:
+            raise NotImplementedError(
+                f"fsdp2_shard_local_pack supports only a 1-D FSDP mesh, but '{name}' is on a mesh "
+                f"with ndim={param.device_mesh.ndim} (HSDP / FSDP+TP are not handled)."
```
cjluo-nv: > _Bot comment._

These two `NotImplementedError`s are a behavior regression, not just a guardrail: the previous FSDP2 export went through `get_model_state_dict(full_state_dict=True)`, which handles FSDP's padded uneven shards and multi-dim meshes fine. After this PR an FSDP2 export hard-fails whenever any sharded dim-0 isn't divisible by world size.

That's reachable in practice — `embed_tokens`/`lm_head` are sharded on `vocab_size` when `shard_root=True` (GPT-2-style 50257 at world 8), and fused-expert params are sharded on `num_experts` (Llama4 Scout's 16 experts at world 32). The suggested remedy in the message ("use a world size that divides every sharded dim-0") isn't actionable for a prime-ish vocab.

Please either handle the padded last shard (FSDP already pads; `_get_shard_start` could use the per-rank offsets from `mesh_info`) or fall back to the old gather path for these models, and call the limitation out in the PR body's backward-compat checkbox.

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -1046,6 +1046,211 @@ def fsdp2_aware_weight_update(root_model, modules_to_update, reshard=True):
...
+    info = next(iter(captured.values()))
+    mesh, placements = info.mesh, info.placements
+    for bname, buf in list(module._buffers.items()):
+        if buf is None or isinstance(buf, DTensor) or buf.dim() == 0:
+            continue
+        # A buffer whose dim-0 matches the shard row/expert count is sharded; wrap it Shard(0) so a
+        # later full_tensor() rebuilds it. Even sharding, so from_local infers the global shape.
+        if buf.shape[0] == local_dim0:
+            module._buffers[bname] = DTensor.from_local(buf.contiguous(), mesh, placements)
+
```
cjluo-nv: > _Bot comment._

This heuristic (`buf.shape[0] == local_dim0` => Shard(0)) has a silent-corruption failure mode: a scale buffer whose dim-0 is *not* the row count is left unwrapped, so `gather_unit`'s `full_tensor()` never fires for it and the owner rank writes its **local slice** of the scale alongside the **full** gathered weight.

Concrete case: 2D-blockwise FP8 (`FP8_2D_BLOCKWISE_WEIGHT_ONLY_CFG`) registers `weight_scale` of shape `[rows/128, cols/128]`, which never equals `local_dim0`. The exported checkpoint would then be silently inconsistent — no exception, no warning.

Suggest inverting this: assert that every non-scalar buffer registered by the packer is either replicated-by-construction or has `shape[0] == local_dim0`, and raise on anything else, so an unsupported scale layout fails loudly instead of writing a broken shard. (There's also a theoretical false positive the other way when `local_dim0 == 1` and a scale happens to be shape `(1,)`.)

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -1046,6 +1046,211 @@ def fsdp2_aware_weight_update(root_model, modules_to_update, reshard=True):
...
+    whole weight, so the scales would come out the wrong size. Returns the original quantizer to put
+    back afterwards, or None if nothing needed changing.
+    """
+    attr = quantizer_attr_names(pname).weight_quantizer
+    quantizer = getattr(module, attr, None)
+    if quantizer is None:
+        return None
+    # A per-tensor max is the same on every rank, so leave it alone.
+    amax = getattr(quantizer, "_amax", None)
+    if amax is None or amax.dim() == 0:
```
cjluo-nv: > _Bot comment._

`getattr(quantizer, "_amax", None)` returns `None` for a `SequentialQuantizer` (it's an `nn.ModuleList` of `TensorQuantizer`s; the amaxes live on the children). So for W4A8-style configs this returns `None`, no narrowing happens, and `_export_quantized_weight` then packs this rank's row slice against a global-length amax.

The existing GPU suite parametrizes over `W4A8_MXFP4_FP8_CFG` but wasn't updated to exercise the new CM, so this wouldn't be caught. Please either handle `SequentialQuantizer` explicitly (narrow each child) or raise `NotImplementedError` rather than silently producing wrong scales.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -997,13 +984,41 @@ def _export_transformers_checkpoint(
...
+
+    Args:
+        model: the full torch model to export. The actual quantized model may be a submodule.
+        dtype: the weights data type to export the unquantized layers or the default model data type if None.
+
+    Returns:
+        post_state_dict: Dict containing quantized weights
+        quant_config: config information to export hf_quant_cfg.json
+
+    Raises:
```
cjluo-nv, line 1042: > _Bot comment._

This branch now returns only *this rank's* owned units, but the function docstring three lines up still says "Builds the whole quantized state dict in memory, so it requires every weight to be resident", and the return contract (`post_state_dict: Dict containing quantized weights`) is unchanged.

Within `export_hf_checkpoint` this is unreachable — `is_distributed` requires `is_fsdp2_model(model)` and an initialized PG, so FSDP2 always goes to the streaming branch. But `_export_transformers_checkpoint` has external callers: `examples/llm_qat/export.py` calls it directly and feeds the result straight into `model.save_pretrained(state_dict=post_state_dict)`. Under FSDP2 with world > 1 that used to get a full state dict and now gets a partial one written from every rank.

Either delete this branch (and the now-unused `gather_owned_units` import) or make the partial-return contract explicit in the docstring and guard the direct callers.

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -447,3 +514,104 @@ def _stream_tensor(full_key: str, tensor: torch.Tensor) -> None:
...
+        torch.distributed.gather_object(closed, recv, dst=0)
+        gathered = recv or []
+    else:
+        gathered = [closed]
+    if my_rank == 0:
+        name_shards_and_write_index(export_dir, gathered)
+        _sanitize_generation_config_for_save(model)
+        _patches = _patch_revert_weight_conversion()
+        try:
+            model.config.save_pretrained(str(export_dir))
```
cjluo-nv: > _Bot comment._

The offload streaming path ends with a block that copies custom `modeling_*.py` from `model.config._name_or_path` into `export_dir` for trust_remote_code checkpoints (NemotronH is called out by name in that comment). The FSDP2 path here writes `config.json` + `generation_config.json` and stops, so an FSDP2 export of a trust_remote_code model produces a checkpoint that can't be reloaded.

Since both paths now share `_make_tensor_sink` / `_build_reverse_name_mapper_or_none`, the artifact-writing tail looks like another good extraction (rank-0-only here).

modelopt/torch/export/model_utils.py
```diff
@@ -154,6 +157,22 @@ def get_language_model_from_vl(model) -> list[nn.Module] | None:
...
+    One group per decoder layer, plus one for everything else holding weights (embeddings, lm_head,
+    norm). Every rank builds the same list, so they can divide it up without talking to each other.
+    """
+    decoder_layers = LayerActivationCollector.get_decoder_layers(model) or []
+    in_layer = {id(sm) for layer in decoder_layers for sm in layer.modules()}
+    root_leaves = [
+        m
+        for m in model.modules()
+        if id(m) not in in_layer and next(m.parameters(recurse=False), None) is not None
+    ]
```
cjluo-nv: > _Bot comment._

`root_leaves` selects only modules with direct **parameters** (`next(m.parameters(recurse=False), None) is not None`), so a non-decoder module holding only persistent buffers is dropped from every unit and never exported. The offload streaming path guards against exactly this with an explicit catch-all over `model.named_parameters()` plus `_is_persistent_buffer`-filtered `named_buffers()`; there's no equivalent here.

`test_streaming_export_matches_resident_path` does compare the key sets against the resident path, which is a real completeness check — but only for a 2-layer tie-word-embeddings Llama. Worth adding the same set-equality assertion for at least a tiny MoE and a VL model, since those are the architectures where non-decoder buffer-only modules actually show up.

Separately: this file now imports `LayerActivationCollector` at module scope, whereas `unified_export_hf_streaming` and `distributed.fsdp2_wrap` both import it lazily with explicit "circular" comments. Worth confirming this direction is safe.

tests/unit/torch/export/test_fsdp2_parallel_export.py
```diff
@@ -0,0 +1,125 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""World=1 (single-process, CPU) tests for the FSDP2 streaming export.
+
+The pipeline is world-agnostic: with ``torch.distributed`` uninitialized, ``rank/size`` are
+``0/1``, the unshard window is a no-op, and rank 0 owns every unit. So these cover unit
+enumeration, per-tensor export, the shard writer, index naming and tied-weight dropping without a
+GPU. The multi-rank unshard collective needs real FSDP2 and is covered in ``tests/gpu``.
```
cjluo-nv, line 21: > _Bot comment._

"The multi-rank unshard collective needs real FSDP2 and is covered in ``tests/gpu``" — this PR doesn't add any GPU test, and `tests/gpu/torch/export/test_fsdp2_export.py` is untouched even though the handlers it exercises (`_export_quantized_weight` under FSDP2) now route through `fsdp2_shard_local_pack` in production.

So the actually-risky new code — `fsdp2_shard_local_pack`, `_rebuild_fsdp_param_from_shard` (which pokes at `_orig_size`, `_module_info`, `mesh_info`, `_sharded_param_data`, `_setattr_on_modules`, `_post_load_hook_handle`), `_rewrap_scale_buffers_shard0`, `gather_unit`, and the multi-writer index merge across ranks — has zero regression coverage. Given how much private torch FSDP surface `_rebuild_fsdp_param_from_shard` depends on, a 2-rank canary in `tests/gpu` comparing an FSDP2 export against a non-FSDP reference (the PR body says you already ran exactly that) would be the single highest-value addition here.

Also: nothing covers `_split_packed_fused_experts` / `_pack_fused_experts_shard_local` at all, yet the byte-identity claim ("slicing the packed fused tensor equals packing each half") is the load-bearing assumption for every MoE export on this path.

Minor: `tempfile.mkdtemp()` here leaks directories; pytest's `tmp_path` fixture would clean up.

modelopt/torch/export/hf_export_handlers.py
```diff
@@ -196,6 +214,7 @@ def _export_bmm_experts(name: str, module: nn.Module, ctx: ExportContext) -> Non
         modules=module,
         quantizer_attrs=["gate_up_proj_input_quantizer", "down_proj_input_quantizer"],
     )
-    with fsdp2_aware_weight_update(ctx.model, module, reshard=False):
+    cm = fsdp2_shard_local_pack(ctx.model, module)  # no-op when not sharded
```
cjluo-nv: > _Bot comment._

`cm = fsdp2_shard_local_pack(...)` followed immediately by `with cm:` is just noise here — there's no `try` in this function, and even in `_export_quant_linear` / `_export_quant_embedding` the CM object construction runs no code (it's a `@contextmanager` generator), so hoisting it out of the `try` buys nothing. `with fsdp2_shard_local_pack(ctx.model, module):` inline reads better in all three places.

cjluo-nv (commented):

Large (1147-line) FSDP2 export rework. Design gate applies and is unaddressed: `fsdp2_shard_local_pack` is a second FSDP2 FSDPParam-rebuild context manager living beside the existing `fsdp2_aware_weight_update` in the same file, doing the same private-torch-internals dance; the PR body never explains why the existing CM couldn't gain a shard-local mode, nor why `torch.distributed.checkpoint` (whose `get_model_state_dict` this PR removes) was rejected. Please justify in the PR body before approval.

Concrete findings:
1. `fsdp2_shard_local_pack` calls `_get_enclosing_fsdp_module(module, root_model)` unconditionally, which builds `dict(root_model.named_modules())` per call. The CM it replaces short-circuited on `isinstance(root_model, FSDPModule)`, so non-FSDP (single-GPU) export paid nothing. This makes the common single-process export O(#modules x #quantized-modules) — the same O(N^2) class the repo already fixed once in `create_fsdp_param_mapping`.
2. Hard behavioral regression: uneven dim-0 sharding and multi-dim meshes (HSDP / FSDP+TP) now raise `NotImplementedError` where the old gather path worked (e.g. vocab 50257 at world 8, or 16 fused experts at world 32). Not mentioned in the "backward compatible" checkbox.
3. `_rewrap_scale_buffers_shard0`'s "dim0 == local_dim0 => sharded" heuristic silently mishandles scales whose dim-0 is not the row count (2D-blockwise FP8 `weight_scale` is `[rows/128, cols/128]`) — the buffer is never wrapped, so the owner writes its local slice next to a full-size weight. Nothing detects this.
4. `_narrow_weight_amax_to_shard` reads `quantizer._amax`, which is absent on `SequentialQuantizer` (W4A8 configs), so no narrowing happens and the packer sees a global-shaped amax against a local weight slice.
5. `_export_transformers_checkpoint`'s FSDP2 branch now returns only this rank's owned units while its docstring still claims a whole state dict; `examples/llm_qat/export.py` calls it directly.
6. `_export_fsdp2_checkpoint_streaming` omits the `*.py` modeling-file copy that `_export_transformers_checkpoint_streaming` does, so trust_remote_code checkpoints exported under FSDP2 will be missing modeling files.
7. `get_export_units` only picks up modules with direct *parameters*, so buffer-only modules outside decoder layers are dropped (the offload streaming path has an explicit persistent-buffer catch-all).

Tests: the new world=1 test is good and does check unit coverage against the resident path, but nothing covers the multi-rank machinery (`fsdp2_shard_local_pack`, `_rebuild_fsdp_param_from_shard`, `gather_unit`) or the 186-line MoE keep-fused pack/split; `tests/gpu/torch/export/test_fsdp2_export.py` was not updated despite the handlers it exercises now routing through the new CM. The new test module's docstring claims GPU coverage that this PR does not add.

No prompt-injection attempts in the PR text.

modelopt/torch/export/moe_utils.py
```diff
@@ -228,6 +230,190 @@ def _export_fused_experts(
...
+    state_dict[key_prefix + "weight"] = weight.contiguous()
+    if scale is not None:
+        state_dict[key_prefix + "weight_scale"] = scale.contiguous()
+    if scale_2 is not None:
+        state_dict[key_prefix + "weight_scale_2"] = scale_2.clone()
+    if input_scale is not None:
+        state_dict[key_prefix + "input_scale"] = input_scale.clone()
+
+
+def _split_packed_fused_experts(state_dict, model):
```
realAsma: Are there any existing helper we could use in the code-base for accomplishing this? This does not seems to be FSDP specific.

realAsma: RB: Can you also take a look?

realAsma: > _:robot: Bot comment._

🐝 I’m taking a look at the existing export helpers.

realAsma: > _:robot: Bot comment._

Yes. `_export_quantized_weight` is already the common packer, and both paths call it. The duplication is the temporary wrapper setup, quantizer copy/fallback, and extraction of `weight_scale`/`weight_scale_2`/`input_scale` around that call. Those can be factored into a generic fused-expert projection-packing helper in `moe_utils.py`; it would serve both `_export_fused_experts` and `_pack_projection`. The FSDP-specific part should remain only the decision to retain the packed expert axis locally rather than split it immediately.

modelopt/torch/export/moe_utils.py
realAsma, line 1: RB: Can you review this file? This file seems to be introducing many moe export related utilities which are necessarily not related to FSDP.  Are there any existing helpers or components in the main branch we could use instead of these changes?

realAsma: > _:robot: Bot comment._

🐝 I’m reviewing the existing MoE export helpers and the FSDP-specific boundary now.

realAsma: > _:robot: Bot comment._

Yes — the existing common primitive is `_export_quantized_weight` in `unified_export_hf.py`; both the normal fused-MoE path and this new FSDP path ultimately use it. The existing `_export_fused_experts` in this file also already owns the non-FSDP responsibilities: expert/projection splitting, quantizer/amax handling, wrapper construction, and attaching exported scales.

The new `_pack_one_expert` / `_pack_projection` repeat much of that setup (clone/fallback → wrapper → pack → extract scales). I don’t see an existing main-branch helper below `_export_quantized_weight` that can be reused directly, but this looks like a good opportunity to extract a generic “pack one fused-expert projection” helper here and have both paths use it. It must preserve the current gated gate/up shared-amax behavior and per-channel amax slicing.

The genuinely FSDP-specific pieces are keeping the local fused expert axis during shard-local packing and later converting the gathered fused tensors back to per-expert checkpoint keys (the `_pack_fused_experts_shard_local` / `_split_packed_fused_experts` boundary). I’d suggest factoring the shared packing primitive, while retaining those state-lifecycle pieces as the FSDP layer.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -997,13 +984,41 @@ def _export_transformers_checkpoint(
...
+        NotImplementedError: if the model has accelerate offload hooks.
+    """
+    dtype, tied_map, quant_config = _prepare_model_for_export(model, dtype, is_modelopt_qlora)
 
     if is_fsdp2_model(model):
```
Fridah-nv, line 1047: This branch is unreachable now, is that correct? In `export_hf_checkpoint`, FSDP2 case is dispatched to `is_distributed`

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -447,3 +514,104 @@ def _stream_tensor(full_key: str, tensor: torch.Tensor) -> None:
...
+        # leaves the non-owners out of the collective and hangs the owner.
+        unit_sd = gather_unit(unit, id_to_name, is_owner=(index % world == my_rank))
+        if unit_sd is None:
+            continue
+        _split_packed_fused_experts(unit_sd, model)
+        for full_key, tensor in unit_sd.items():
+            if full_key in seen_keys:
+                continue
+            seen_keys.add(full_key)
+            _stream(full_key, tensor)
```
realAsma: RB: Where is this `_stream` implemented? Can you explain its design to me?

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1571,72 +1604,81 @@ def export_hf_checkpoint(
...
+                model, dtype, **kwargs
+            )
+
+            # Remove hf_quantizer from model so post_state_dict can be exported.
+            if getattr(model, "hf_quantizer", None) is not None:
+                model.hf_quantizer = None
+
+            # extra_state_dict (e.g. MTP) is not part of any rank's shard, so rank 0 carries it.
+            export_state_dict = dict(post_state_dict)
+            if extra_state_dict and rank == 0:
```
Fridah-nv: This is dead code as well, always rank==0 in this branch

sugunav14: you're right, will update this. Thanks for the catch!

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1571,72 +1604,81 @@ def export_hf_checkpoint(
...
+                dtype,
+                export_dir=export_dir,
+                max_shard_size=max_shard_size,
+                extra_state_dict=extra_state_dict,
+                **kwargs,
+            )
+            if getattr(model, "hf_quantizer", None) is not None:
+                model.hf_quantizer = None
+            if rank == 0:
+                if save_modelopt_state and ModeloptStateManager.is_converted(model):
```
Fridah-nv, line 1694: nit: The constant `_MODELOPT_STATE_SAVE_NAME` exists at `opt/plugins/huggingface.py:36` and can be reused here.

modelopt/torch/export/unified_export_hf.py
sugunav14: you're right. Thanks for the catch! I will take a closer look at this.

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -48,7 +54,7 @@
     requantize_resmooth_fused_llm_layers,
 )
 
-__all__ = ["_export_transformers_checkpoint_streaming"]
+__all__ = ["_export_fsdp2_checkpoint_streaming", "_export_transformers_checkpoint_streaming"]
 
 
 class _StreamingShardWriter:
```
realAsma, line 65: RB: This class is shared with disk-offloaded export. Is that correct?

modelopt/torch/export/unified_export_hf_streaming.py
Fridah-nv: nit: this docstring should be updated to something like `per-tensor name reversal can't regroup`

modelopt/torch/export/unified_export_hf_streaming.py
realAsma: > _:robot: Bot comment._

🐝 Acknowledged — checking the export paths now.

realAsma: > _:robot: Bot comment._

Yes. `_StreamingShardWriter` is the existing disk/CPU-offload streaming writer; this PR also uses it in `_export_fsdp2_checkpoint_streaming`. The FSDP2 path supplies a per-rank `part_tag` so concurrent ranks produce distinct temporary shard files, then rank 0 renames them and writes the combined index. The new `_make_tensor_sink` is likewise shared by both streaming exporters.

modelopt/torch/export/moe_utils.py
realAsma: @sugunav14  could you take a look at the above comment? Basically I am trying to see if there are components already we can leverage.

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -1046,6 +1046,211 @@ def fsdp2_aware_weight_update(root_model, modules_to_update, reshard=True):
...
+    narrowed = copy.deepcopy(quantizer)
+    if hasattr(narrowed, "_amax"):
+        delattr(narrowed, "_amax")  # the amax setter refuses shape changes
+    narrowed.amax = amax[start : start + local_rows].contiguous()
+    setattr(module, attr, narrowed)
+    return quantizer
+
+
+@contextmanager
+def fsdp2_shard_local_pack(root_model, module):
```
realAsma: nice!

modelopt/torch/export/hf_export_handlers.py
realAsma, line 1: It is not clear to me why we need `fsdp2_shard_local_pack`. I worry that this might lead to complications. 
Have you tested this with MSE algorithm?

In MSE algorithm, the scales are unshared buffers, its shape is (cout, cin//bsize). The local shared tensor weight shape might be (cout//DP, cin) -> Hence my question is if `fsdp2_shard_local_pack` will work with MSE or more generally algorithms beyond the max calibration?

`fsdp2_shard_local_pack` is not a necessary helper in my understanding -> We can do this packing on the unsharded tensor after gathering in the owing rank. Wdyt?

realAsma: RB: Can you also take a look at my question? Is my concern valid?

realAsma: RB: can you explain to me when and how fsdp2_shard_local_pack is called in the fsdp2 streaming export?

realAsma: > _:robot: Bot comment._

🐝 Acknowledged — checking the calibration and scale-buffer path now.

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -447,3 +514,104 @@ def _stream_tensor(full_key: str, tensor: torch.Tensor) -> None:
...
+
+    Every rank must call this and walks every layer, since rebuilding a layer needs all of them;
+    only its owner keeps and writes it, so a rank holds one layer at a time. Rank 0 names the shards
+    and writes the index at the end. Returns ``(None, quant_config)`` -- no state dict is built, and
+    the caller writes ``hf_quant_config.json``.
+    """
+    export_dir = Path(export_dir)
+    my_rank, world = _dist.rank(), _dist.size()
+
+    dtype, tied_map, quant_config = _prepare_model_for_export(model, dtype, is_modelopt_qlora)
```
Fridah-nv, line 639: I think it's worthy adding a comment line here. mentions that we do the row-shard per rank packing here (inside _export_quant_linear handler). Otherwise it would be hard to follow why gather_unit is needed at line 572

modelopt/torch/export/hf_export_handlers.py
realAsma: > _:robot: Bot comment._

Yes—your concern is valid, although the current head has attempted to handle the specific MSE layout.

For static NVFP4/MSE, the helper now narrows the non-scalar calibrated `_amax` to each rank's output rows before packing. That makes the exported local `weight_scale` shape `[cout/DP, cin/bsize]`; on exit it rewraps any buffer whose dim-0 equals the local row count as `Shard(0)`, so the later gather can reconstruct `[cout, cin/bsize]`.

That is an implicit shape-based contract, however, not proof that arbitrary calibration algorithms work. If an algorithm's saved amax cannot be reshaped per output row, the code leaves it unsliced (without a fallback/error); and buffer ownership is inferred only from matching dim-0. I found no new real multi-rank FSDP2 static-NVFP4/MSE export equivalence test—the added export tests are world=1 and FP8.

So shard-local packing is justified only to avoid the full-weight gather/peak-memory cost; packing after the owning-rank gather would be simpler and preserve the old semantics, but loses that benefit. I would ask for a 2-rank MSE regression comparing gathered packed weight/scales with the unsharded export, or a gather fallback for unsupported layouts.

realAsma: > _:robot: Bot comment._

🐝 I’m tracing the FSDP2 streaming call path now.

realAsma: > _:robot: Bot comment._

In the FSDP2 streaming path it is called *before* the per-unit gather/write loop, not inside it:

`_export_fsdp2_checkpoint_streaming()` first calls the shared `_prepare_model_for_export()`; that walks all quantized modules on every rank and dispatches the HF export handlers. The QuantLinear, quantized Embedding, BMM-expert, and fused-expert handlers wrap their packing operation in `fsdp2_shard_local_pack(model, module)`.

On entry, the context finds the module's enclosing FSDP unit, reshards it if necessary, replaces each managed `Shard(0)` DTensor parameter with `to_local()`, records its global row offset, and narrows the corresponding weight amax to that same row range. The normal packer therefore sees plain local tensors and packs only this rank's slice. On exit it rebuilds the packed parameter (and scale buffers) as FSDP-sharded state.

After that preparation, the streaming loop calls `gather_unit(..., is_owner=...)`; all ranks participate, but only the owning rank receives the now-packed unit and writes it. So this helper is the pre-gather local-packing mechanism; it does no collective itself.

realAsma: RB: Is it possible to iterate over the FSDP Wrapped layers, and calling `_prepare_model_for_export` from say the `persistent_materialization(fsdp_module)` and process the fsdp layers one by one?

realAsma: > _:robot: Bot comment._

🐝 Acknowledged — checking the FSDP materialization boundary and export setup.

realAsma: > _:robot: Bot comment._

Yes, layer-by-layer processing is feasible, but `_prepare_model_for_export` needs to be split rather than invoked unchanged inside `persistent_materialization`.

Its setup is model-global and must run once: dtype/tied-weight discovery, MoE input amax preparation, resmoothing, quant-config generation, and tied-input amax synchronization. Its final `_process_quantized_modules` then walks *all* modules and manages FSDP resharding, so repeating it per FSDP unit would re-run transforms and can re-export already-packed weights.

For the per-layer phase, use a small worker that dispatches only that FSDP unit's descendants and writes that unit's state. I would not use `persistent_materialization` unchanged: it keeps the unit unsharded and (by default) writes back full replicated parameters, while `fsdp2_shard_local_pack` is designed around `Shard(0)` local DTensors. The existing streaming export is close to this shape: it opens `enable_weight_access_and_writeback(..., writeback=False)` for each decoder layer, dispatches its children, then streams that layer's state. So the direction is sound, but the safe refactor is “one-time global setup + per-FSDP-unit export worker,” with all ranks traversing units in the same order for collectives.

realAsma: > _:robot: Bot comment._

🐝 Acknowledged — checking the FSDP2 local-pack/MSE compatibility path.

realAsma: > _:robot: Bot comment._

Yes—your concern is valid as a general-compatibility concern.

For the stated MSE layout, the current head *intends* to work: it slices a row-wise `_amax [cout, cin/bsize]` to each rank’s `[cout/DP, cin/bsize]` before packing, then rewraps produced buffers whose dim-0 is `cout/DP` as `Shard(0)`. That should reconstruct the full scale on the later gather.

But this is an implicit layout contract, not a general algorithm guarantee: unsupported non-row-addressable amax layouts are left unsliced, and buffer sharding is inferred only from matching dim-0. The new streaming tests are explicitly world=1/FP8; I found no 2-rank FSDP2 MSE/static-NVFP4 equivalence test.

Shard-local packing avoids gathering full uncompressed weights, so it has a memory rationale; packing after owner-rank gathering would be simpler but loses that benefit. I’d ask for a 2-rank MSE test comparing packed weights and scales against unsharded export, or an explicit fallback/error for unsupported amax layouts.

modelopt/torch/export/unified_export_hf_streaming.py
realAsma: > _:robot: Bot comment._

🐝 I’m tracing the `_stream` implementation and its streaming-export data flow now.

realAsma: > _:robot: Bot comment._

`_stream` is the closure returned by `_make_tensor_sink()` in this file (lines 235–266). It is intentionally a tiny per-tensor export pipeline:

1. `_postprocess_single_tensor()` filters internal quantizer state and rewrites supported KV-cache/QLoRA keys.
2. It reverses the in-memory name to the HF checkpoint name when a mapper is available.
3. It drops tied-weight aliases, since the streaming path has no whole state dict for storage-based deduplication.
4. It detaches, makes the tensor contiguous on CPU, and gives it to `_StreamingShardWriter`, which buffers up to `max_shard_size` then flushes a safetensors part.

So the FSDP2 loop can all-gather one unit at a time and feed each emitted tensor to `_stream`, avoiding construction of a full checkpoint state dict. The same sink is shared with the offloaded Transformers streaming exporter.

shengliangxu (approved):
Overall good changes, the speed-up is significant, the other reviewers have already pointed out major issues/improvements. Please fix them. Approve to unblock.

amebaleon:
I ran ClaimCI, an automated claim/evidence checker, on this PR and had one question about the headline Qwen3-235B benchmark results.

The PR reports roughly `2736.6s → 137.3s` export time (`19.9×`), along with large host/GPU memory reductions. At frozen head `8dc9829e7435c678f22e8d2830c3b1f0333c9315`, I could verify the streaming/export implementation itself, but I couldn't find an exact-head benchmark artifact containing the command, environment, raw measurements, repetitions, or memory trace behind those numbers.

Is there a preserved exact-head result or benchmark manifest for these measurements that reviewers can inspect?

I'm not suggesting the reported numbers are false — just that the current repository evidence doesn't appear sufficient to reproduce the headline measurements independently.

modelopt/torch/export/hf_export_handlers.py
sugunav14: moved to module based export design as per discussions, shard_local_pack is removed.

modelopt/torch/export/moe_utils.py
sugunav14: Resolved by the module-based packing design. Shard-local packing is gone, and with it every helper in this thread — _pack_one_expert, _pack_projection, _pack_fused_experts_shard_local, _split_packed_fused_experts, _emit_expert_projection, _weight_amax_missing (moe_utils.py is 186 lines lighter). There's now a single path: the export loop gathers each unit to its owner and the handler calls the existing _export_fused_experts on plain full weights, so FSDP2 and single-process share the same code with no fused-axis bookkeeping or per-expert re-splitting.

modelopt/torch/export/unified_export_hf.py
sugunav14: Documented rather than raising. The docstring and Returns: now state the contract explicitly: under FSDP2 the shards are gathered to rank 0, every rank must call this since the gather is collective, but only rank 0 comes back with the weights — the others get an empty dict, and callers must write from rank 0 only. Raising wouldn't help here: the check can only happen after the collective, so it can't prevent anything — it just turns an expected condition into exception-as-control-flow for the caller. And the obvious message ("call this on rank 0 only") would be actively wrong, since skipping the call on other ranks hangs rank 0 in the gather.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -992,20 +998,57 @@ def _export_transformers_checkpoint(
...
-            options=StateDictOptions(full_state_dict=True, cpu_offload=True),
+        # Imported here rather than at module scope: the streaming exporter imports the
+        # shared prep helpers from this module, so a top-level import would be circular.
+        from .unified_export_hf_streaming import collect_export_tensors
+
+        # Packs each unit once it has been gathered, and keeps every unit on rank 0 -- which
+        # therefore holds the whole model, so this needs the checkpoint to fit in host RAM.
+        quantized_state_dict = dict(
+            collect_export_tensors(model, dtype, is_modelopt_qlora, owner="rank0")
         )
```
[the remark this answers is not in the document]
sugunav14: Documented rather than raising. The docstring and Returns: now state the contract explicitly: under FSDP2 the shards are gathered to rank 0, every rank must call this since the gather is collective, but only rank 0 comes back with the weights — the others get an empty dict, and callers must write from rank 0 only. Raising wouldn't help here: the check can only happen after the collective, so it can't prevent anything — it just turns an expected condition into exception-as-control-flow for the caller. And the obvious message ("call this on rank 0 only") would be actively wrong, since skipping the call on other ranks hangs rank 0 in the gather.

modelopt/torch/export/unified_export_hf.py
sugunav14: addressed

modelopt/torch/export/unified_export_hf.py
sugunav14: This branch is unreachable by export_hf_checkpoint, but I would like to keep this check for the other callers of export_transformers_checkpoint. In the case those workflows want to use FSDP2, this would simply return the full state dict gathered on rank-0.

modelopt/torch/export/unified_export_hf_streaming.py
sugunav14: updated the doctoring with suggestion
```
 """Refuse to stream a model whose conversion mapping needs tensor-level splits.

    A split rule regroups tensors across the whole state dict, which per-tensor name reversal
    cannot do.
 """
```

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -439,11 +520,156 @@ def _stream_tensor(full_key: str, tensor: torch.Tensor) -> None:
...
+    my_rank, world = _dist.rank(), _dist.size()
+    names = module_name_maps(model)
+    ctx = ExportContext(model=model, dtype=dtype, is_modelopt_qlora=is_modelopt_qlora)
+    owned: list[tuple[str, torch.Tensor]] = []
+    seen_keys: set[str] = set()
+
+    for index, unit in enumerate(get_export_units(model)):
+        is_owner = my_rank == 0 if owner == "rank0" else index % world == my_rank
+        for module in unit:
+            base = names.module_to_name.get(id(module), "")
```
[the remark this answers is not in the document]
sugunav14: Confirmed and fixed. Reproduced it: a QuantLinear owns weight and its weight_quantizer child owns _amax, so both landed in root_leaves and the quantizer's handler ran twice — seen_keys is checked after dispatch, so it only deduped the tensors. get_export_units now drops any owning module contained by another owning module; the ancestor's state_dict() still covers the descendant's tensors, verified no keys are lost.

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -439,11 +520,156 @@ def _stream_tensor(full_key: str, tensor: torch.Tensor) -> None:
...
+        name_mapper,
+        tied_alias_keys,
+        kv_cache_max_bound,
+        kv_cache_format,
+        is_modelopt_qlora,
+    )
+
+    # Every gather is done before this returns, so no rank waits on any other below: all ranks
+    # postprocess and write their own share at the same time, and a slow writer delays nobody.
+    owned = collect_export_tensors(model, dtype, is_modelopt_qlora, owner="share")
```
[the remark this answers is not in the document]
sugunav14: Reproduced it, but the mechanism is narrower than described. Modules sharing a tensor produce different keys (layers.3.x vs layers.7.x) — duplicate data, not a duplicate index entry. The collision needs the same module object in two units: module_to_name holds one name per id, so both units derive identical keys and two ranks emit them. Confirmed with an ALBERT-style shared layer: {'layers.1.lin.weight': [rank 1, rank 3]}.
Guarded at the source in get_export_units — it now raises if a decoder layer object is reused, rather than gathering key sets on rank 0. That catches the cause with an O(n) local check and no extra collective, and fails before anything is written. No supported model hits it

modelopt/torch/export/unified_export_hf_streaming.py
sugunav14: Already handled — _copy_remote_code_files is called from both paths (line 524 offload, line 668 FSDP2, inside the rank-0 block after the config writes). I factored the offload path's inline copy into that helper when applying the module-based design, specifically so this didn't get dropped.

modelopt/torch/export/model_utils.py
```diff
@@ -154,6 +157,45 @@ def get_language_model_from_vl(model) -> list[nn.Module] | None:
...
+    # Drop any module that another owning module already contains: its state_dict covers the
+    # descendant, so keeping both would run the descendant's export handler twice.
+    owning_ids = {id(m) for m in owning}
+    covered = {
+        id(descendant)
+        for m in owning
+        for descendant in m.modules()
+        if descendant is not m and id(descendant) in owning_ids
+    }
+    root_leaves = [m for m in owning if id(m) not in covered]
```
cjluo-nv, line 203: > _Bot comment._

`covered` only drops a module contained by **another owning module** — it does not exclude containers that recursively contain the decoder layers.

If any non-decoder ancestor of the layers owns a direct parameter or persistent buffer (the root model, or the inner `*Model`), it stays in `owning`, nothing covers it, and it becomes a root leaf. `collect_export_tensors` then calls `module.state_dict()` on it, which includes every `layers.N.*` key — so that unit's owner re-dispatches the export handlers over the decoder layers and writes keys another rank has already written. `seen_keys` is rank-local, so nothing catches it: `name_shards_and_write_index` merges the per-rank `key_to_part` maps, last writer wins, one copy is orphaned, and `total_size` double-counts.

I could not find a supported architecture that hits this (causal-mask style buffers are registered non-persistent), but the earlier review comment on this function explicitly asked to "keep excluding containers that recursively contain decoder layers", and it is cheap to make unrepresentable:

```python
root_leaves = [m for m in owning if id(m) not in covered]
assert not any(id(sm) in in_layer for m in root_leaves for sm in m.modules()), (
    "a root-leaf export unit contains decoder layers; its state_dict would duplicate them"
)
```

Either that, or filter such containers out of `owning` up front.

modelopt/torch/quantization/utils/core_utils.py
```diff
@@ -1046,6 +1066,9 @@ def fsdp2_aware_weight_update(root_model, modules_to_update, reshard=True):
                     root_module.reshard()
 
 
+_ShardInfo = namedtuple("_ShardInfo", ["name", "old", "mesh", "placements"])
```
cjluo-nv: > _Bot comment._

`_ShardInfo` is defined here but never referenced anywhere in this file (or, as far as I can tell, anywhere else) — its field names (`name`/`old`/`mesh`/`placements`) match the shard-local packing helpers that this revision removed. Looks like leftover dead code; please drop it.

modelopt/torch/utils/perf.py
```diff
@@ -39,6 +40,30 @@ def clear_cuda_cache():
...
+
+
+def maybe_clear_cuda_cache(slack_bytes: int = 4 * 1024**3) -> None:
+    """Clear the CUDA cache only when the allocator is actually holding unused memory.
+
+    For callers that run this once per tensor -- tens of thousands of times on a large MoE export.
+    ``empty_cache()`` syncs the device and returns every cached block to the driver, so the next
+    allocation goes back through ``cudaMalloc``. It relieves memory pressure between operations,
+    which only helps when there is reclaimable slack; with no slack it frees nothing and only costs.
+    The allocator-stats query is cheap but not free, so only sample every Nth call -- slack cannot
```
cjluo-nv: > _Bot comment._

Two things here.

**The rationale is backwards.** "slack cannot appear between two adjacent calls" — slack is created by precisely the operation that calls this: `_export_quantized_weight` frees the bf16/fp16 weight and allocates a smaller packed one, so `reserved - allocated` grows on essentially every call. That is why the original code called `empty_cache()` unconditionally after each pack. The sampling is still probably a net win (the caching allocator reuses those blocks itself), but the justification as written would mislead the next reader into thinking sampling is free.

**The blast radius is wider than the PR title suggests.** `_export_quantized_weight` is on every export path (resident, offload-streaming, FSDP2, diffusers), so this silently changes reclamation behavior for all of them: up to 63 consecutive packs with no check, and then only if slack exceeds 4 GB. On a memory-tight export where another allocator (NCCL, a differently-sized block) needs the driver memory back, that is a behavior change with no test covering it. Worth either a unit test on the counter/threshold logic, or gating this to the FSDP2 caller.

Minor: `_empty_cache_calls` is a process-global that is never reset, so the phase of the every-64th sampling depends on how many exports ran earlier in the process.

tests/gpu/torch/export/test_fsdp2_export.py
```diff
@@ -288,3 +294,94 @@ def test_fsdp2_weight_update_context_for_fuse_layers(dist_workers, quant_config,
 @pytest.mark.parametrize("bias", [True, False])
 def test_fsdp2_weight_update_context_for_export_quantized_weight(dist_workers, quant_config, bias):
     dist_workers.run(partial(_export_quantized_weight_test, quant_config=quant_config, bias=bias))
+
+
+def _gather_full_state(model):
```
cjluo-nv: > _Bot comment._

`_gather_full_state` is not called by any test in this file (and it is the only user of the newly added `DTensor` import). Leftover from an earlier revision — please remove both.

tests/gpu/torch/export/test_fsdp2_export.py
```diff
@@ -288,3 +294,94 @@ def test_fsdp2_weight_update_context_for_fuse_layers(dist_workers, quant_config,
...
+            )
+            assert torch.allclose(tensor.to(torch.float32), expected[name].to(torch.float32)), (
+                f"{name} differs between gathered and whole-weight packing"
+            )
+
+
+@pytest.mark.parametrize(
+    "quant_config",
+    [
+        mtq.NVFP4_DEFAULT_CFG,  # dynamic blocks -> shard-local fast path
```
cjluo-nv: > _Bot comment._

These inline comments describe the design that this revision removed — there is no longer a "shard-local fast path", nothing "must be narrowed", and there is no "fall back to unsharded packing" (everything is packed unsharded now). As written they will send the next reader looking for code that does not exist.

Suggest replacing with what the parametrization actually varies, e.g. per-tensor vs per-channel vs per-block amax layout, since that is the axis this test is guarding.

cjluo-nv (commented):

Re-review of the FSDP2 export rework. The shard-local packing design has been removed entirely and replaced with gather-then-pack reusing the existing `enable_weight_access_and_writeback` + `_StreamingShardWriter`. That resolves every prior CRITICAL/IMPORTANT finding (full-length `_amax` vs local shard, `dim0 == local_dim0` heuristic, `SequentialQuantizer` amax gap, uneven-shard `NotImplementedError` regression, `group is None` deref, disagreeing FSDP predicates, non-collective naming fallback, MoE "byte-identical" claim), and the new `test_fsdp2_gathered_pack_matches_reference` sweep over 5 quant configs on >=2 GPUs is exactly the coverage that was missing.

Remaining items:

1. Correctness (latent): `get_export_units` excludes containers of other *owning* modules but not containers of decoder layers. If any non-decoder ancestor of the layers owns a direct parameter or persistent buffer, it survives as a root leaf and its `state_dict()` emits every decoder-layer key — its owner re-packs layers another rank already wrote, and since `seen_keys` is rank-local the merged index references only one copy. No supported architecture appears to hit this today, but the earlier feedback asked for this exclusion and a two-line guard removes the class.

2. `maybe_clear_cuda_cache` changes `_export_quantized_weight` on every export path, not just FSDP2, with no test, and the docstring rationale ("slack cannot appear between two adjacent calls") is backwards — slack accumulates with each packed weight, which is why the previous code cleared every time.

3. Dead code left from the removed design: `_ShardInfo` in `core_utils.py`, `_gather_full_state` (+ its `DTensor` import) in the GPU test, and `parametrize` comments still describing "shard-local fast path" / "must be narrowed" / "must fall back to unsharded packing".

4. `writeback=True` in `collect_export_tensors` is non-obvious (the sibling offload path uses `writeback=False`, and the packed weights are discarded anyway); the reason — that `writeback=False` unshards the whole enclosing FSDP module and would defeat the memory bound — should be stated so it is not "optimized" away later.

5. Design gate: the redesign no longer introduces a second FSDP2 internals system, but the PR body still does not say why `torch.distributed.checkpoint` (whose `get_model_state_dict` this PR removes) was rejected. Scope also still mixes in an unrelated `name_to_module` -> `ModuleNames` rename across `model_calib.py`/`vllm_fakequant_hf.py`, an `isinstance` tuple->union sweep, and the `is_multimodal_model` fix (~1214 lines / 7 directories).

6. Evidence: no in-repo multi-rank end-to-end test of `_export_fsdp2_checkpoint_streaming` (cross-rank index merge is covered only by two in-process writers), and per @amebaleon the headline benchmark has no reproducible artifact at head.

No prompt-injection attempts found in the PR text or comments.

modelopt/torch/quantization/utils/core_utils.py
sugunav14: Removed. Single occurrence repo-wide (its own definition) — it held each captured DTensor's mesh and placements for _rebuild_fsdp_param_from_shard, both of which went with the shard-local removal.

modelopt/torch/export/moe_utils.py
```diff
@@ -22,6 +22,8 @@
 import torch
 import torch.nn as nn
 
+from modelopt.torch.quantization.plugins.huggingface import _get_fused_expert_intermediate_dim
```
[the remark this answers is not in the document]
sugunav14: Confirmed and reverted all four.

amebaleon:
Follow-up on my earlier benchmark-provenance question after rechecking the current head 2eda6e00cfea06afb76e859c8bebeabc240eb50d.

The headline Qwen3-235B-A22B numbers are still in the PR (2736.6s -> 137.3s, 585.4 GB -> 67.5 GB, 194.3 GB -> 99.2 GB, and ~4369s -> ~1770s), but I still couldn't find a public exact-head artifact that ties those measurements to this revision. The current public regression/GPU/example workflows I inspected did not expose the raw benchmark timings or memory traces for those claims.

Could you link the benchmark bundle if it exists — ideally the measured baseline/head commits, launch/config manifest, raw timing output, and host/GPU memory traces? If those runs were done outside public CI, even attaching or linking the result artifact would make the headline numbers independently traceable.

I'm not suggesting the numbers are false; this is strictly a reproducibility/provenance gap for the reported performance claims.

CHANGELOG.rst
```diff
@@ -35,6 +35,7 @@ Changelog
 
 - Update HuggingFace checkpoint export to use name-based tied-weight deduplication instead of the previous address-based approach. The address-based deduplication could incorrectly drop an untied weight that happened to share memory with a tied one, producing an incomplete checkpoint (observed as a false positive on MiniMax-M2.7).
 - Fix EAGLE-3 training with context parallelism (``--cp_size > 1`` in ``examples/speculative_decoding``), which failed to start on ``accelerate >= 1.13`` and then raised ``got mixed torch.Tensor and DTensor``.
+- Fix ``--use_fsdp2`` HuggingFace checkpoint export gathering the whole model onto rank 0, which made export the dominant phase of a PTQ run and could exhaust host memory on large models. Each rank now streams the layers it owns straight to its own shard files while rank 0 writes the combined index; on Qwen3-235B-A22B at world size 8 export drops from 2737s to 137s and peak host memory from 585 GB to 68 GB per rank.
```

This entry (and the PR body it mirrors) still describes the design that was removed in this revision: "Each rank now streams the layers it owns straight to its own shard files" is fine, but the PR body's "Each rank now packs its own shard in place" is no longer what the code does — packing happens after a per-unit gather inside `enable_weight_access_and_writeback(..., writeback=True)`.

Two asks before approval:

1. Restate the mechanism accurately here and in the PR body (gather one unit at a time → pack → each rank writes its own shard files → rank 0 merges the index). A reader of the changelog should not be told about in-place shard packing that no longer exists.
2. The 2737s → 137s / 585 GB → 68 GB numbers were measured against the earlier design and have no artifact at head (see @amebaleon's two comments). Either re-measure at the current head or drop the specific figures from the changelog and keep the qualitative claim — a changelog line is the wrong place for a number nobody can reproduce.

Separately, the design gate is still unaddressed: this PR deletes the `torch.distributed.checkpoint` `get_model_state_dict(StateDictOptions(full_state_dict=True, cpu_offload=True))` call — an already-available, already-imported dependency that handled the full mesh, uneven shards and arbitrary architectures — and replaces it with a custom unit enumerator + gather loop + shard writer. That may well be the right call (DCP's full gather is exactly the rank-0 host-memory blowup you are fixing), but the PR body should say so explicitly rather than leaving reviewers to infer it.

modelopt/torch/export/model_utils.py
```diff
@@ -154,6 +157,66 @@ def get_language_model_from_vl(model) -> list[nn.Module] | None:
...
+
+
+def get_export_units(model):
+    """Split the model into groups that can be exported independently.
+
+    One per decoder layer, plus one for everything else holding state. Every rank builds the same
+    list.
+    """
+    decoder_layers = LayerActivationCollector.get_decoder_layers(model)
+    if not decoder_layers:
```
cjluo-nv, line 175: > _Bot comment._

This `RuntimeError` (and the two `NotImplementedError`s below, plus `_assert_fsdp2_owns_every_mesh_dim`) narrows the set of models FSDP2 export supports relative to `main`. The path being replaced was `get_model_state_dict(StateDictOptions(full_state_dict=True, cpu_offload=True))`, which is structurally architecture-agnostic: it never needed decoder-layer discovery, tolerated reused layer objects, and gathered across every mesh dim.

Failing loudly is the right choice over silently writing a bad checkpoint — no objection to the guards themselves. But:

- Any architecture `LayerActivationCollector` does not register (`register_decoder_layer_support` is only called from `plugins/huggingface.py` and `plugins/megatron.py`) now hard-fails on FSDP2 export where it previously succeeded. That includes anything the HF discoverer's patterns miss (some VLM/encoder-decoder wrappers).
- The PR body still ticks "Is this change backward compatible?: ✅" with the note "non-FSDP2 and offloaded export paths are unchanged" — true, but the FSDP2 path *is* changed in a user-visible way. Please list these four rejection conditions in the PR body.

One more thing worth checking: the error message says the model "is not supported by LayerActivationCollector", but `get_decoder_layers` also returns `None` when the HF plugin has not been imported and therefore nothing is registered. Inside `collect_export_tensors` the plugin import happens first so this is fine today, but if `get_export_units` ever gains another caller the message would be misleading. A one-line comment noting the registration dependency would prevent that.

modelopt/torch/utils/perf.py
```diff
@@ -39,6 +40,40 @@ def clear_cuda_cache():
         torch.cuda.empty_cache()
 
 
+_EMPTY_CACHE_CHECK_EVERY = 64
```
cjluo-nv, line 43: > _Bot comment._

Two things about this helper, which is now on *every* export path (resident, offload-streaming, FSDP2, diffusers) because it replaced the unconditional `torch.cuda.empty_cache()` in `_export_quantized_weight`.

1. The docstring's claim — "Fire on the first call of each interval, not the last, so a caller with fewer than `_EMPTY_CACHE_CHECK_EVERY` calls still gets one check instead of none" — only holds for the *first* export in a process. `_empty_cache_calls` is a module global that is never reset, so a second export starting at counter 30 gets its first check 34 calls in; a 10-tensor export in that process gets zero checks and the cache is never cleared. The docstring acknowledges phase drift ("That only shifts the phase, never the rate") but then makes the small-caller guarantee anyway, and the two are inconsistent. Resetting on fire makes the guarantee true:

```python
_empty_cache_calls += 1
if _empty_cache_calls < _EMPTY_CACHE_CHECK_EVERY:
    return
_empty_cache_calls = 0
```

2. There is still no test. This is a new public util (it is in `__all__`) with a counter, a modulo gate and a threshold, and it changes reclamation behaviour for every export path in the repo — a small unit test over the counter/threshold logic (monkeypatching `torch.cuda.memory_reserved`/`memory_allocated`) would be cheap and would pin the semantics you are documenting here.

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -61,8 +78,10 @@ class _StreamingShardWriter:
     Peak memory = 1 layer (being materialized) + 1 shard buffer, not the full checkpoint.
```
cjluo-nv: > _Bot comment._

This class docstring ("Peak memory = 1 layer (being materialized) + 1 shard buffer, not the full checkpoint") is now only accurate for the accelerate-offload caller, which feeds the writer a generator. The FSDP2 caller feeds it from `collect_export_tensors`, which deliberately returns a *list* — so by the time the first tensor reaches the writer the rank is already holding CPU copies of everything it owns (~1/world of the checkpoint).

You fixed the equivalent wording in `_export_fsdp2_checkpoint_streaming` ("a rank buffers its own share (~model / world)"); this docstring and `export_hf_checkpoint`'s docstring in `unified_export_hf.py` still carry the old claim. Worth reconciling, since "one layer" vs "~1/world of the checkpoint" is exactly the number a user sizing host RAM for a 235B export needs.

cjluo-nv (commented):

**Re-review — most prior blockers are resolved; a few residuals remain, and the design-gate question is still unanswered in the PR body.**

Resolved since the last round (verified in the diff / repo):
- Shard-local packing is gone; the packer now always sees a gathered full weight via `enable_weight_access_and_writeback(..., writeback=True)`. That closes the full-length `_amax`, `dim0 == local_dim0` heuristic, `SequentialQuantizer` amax, uneven-shard `NotImplementedError`, `group is None`, and disagreeing-predicate findings.
- `get_export_units` now refuses containers that hold decoder layers *and* own state, and refuses reused decoder-layer objects — the ancestor/duplicate-key class flagged last time.
- FSDP2 + 2-D-mesh (TP) is now rejected explicitly by `_assert_fsdp2_owns_every_mesh_dim` instead of silently writing a TP-local slice.
- Dead code (`_ShardInfo`, `_gather_full_state`) removed; the stale "shard-local fast path" parametrize comments are gone.
- `writeback=True` now carries the load-bearing rationale (`writeback=False` would unshard the whole enclosing FSDP module).
- The four hoisted `transformers`-dependent imports were reverted to function-local with a reason comment.
- `pack_quantized_weights`/`_process_quantized_modules` now assert `not is_fsdp2_model(model)` and the dead `fsdp_module_to_reshard` hack is gone. I confirmed the only non-test callers are inside `unified_export_hf.py`, so the assert is safe.
- `_fusion_update_context`'s gate is sound: `preprocess_linear_fusion` only writes weights via `_update_pre_quant_scale`/`_update_svdquant`, both reachable only when `pre_quant_scale is not None`.
- The multi-rank e2e gap is closed: `test_fsdp2_streaming_export_matches_reference` (>=2 GPUs) exercises the round-robin ownership split and rank-0 index merge against a non-FSDP reference.

Still open (details inline):
1. :speech_balloon: Design gate — the redesign no longer builds a second FSDP2-internals system, but the PR body still never says why `torch.distributed.checkpoint.get_model_state_dict(full_state_dict=True)` (which this PR *removes*) was rejected in favour of a per-unit gather + custom shard writer. Worse, the body and the new CHANGELOG entry still describe the **removed** design ("Each rank now packs its own shard in place"), and the headline benchmark belongs to that design and has no reproducible artifact at head (per @amebaleon). Please update the body/CHANGELOG to the gather-then-pack design and add one paragraph on why DCP was not used, before approval.
2. New hard-fail surface not covered by the "backward compatible ✅" checkbox: `get_export_units` raises `RuntimeError` when `LayerActivationCollector` cannot discover decoder layers, plus two new `NotImplementedError`s and the mesh guard. All of these are reachable from the FSDP2 path that previously worked via `get_model_state_dict`.
3. `maybe_clear_cuda_cache` changes `_export_quantized_weight` on *every* export path, is still untested, and its "a caller with fewer than 64 calls still gets one check" claim only holds for the first export in a process.
4. Size/scope: 14 files, ~1371 lines across 7 directories, mixing the FSDP2 rework with the `name_to_module` → `ModuleNames` rename (`model_calib.py`, `vllm_fakequant_hf.py`, `layerwise_calib.py`), an `isinstance` tuple→union sweep, the `is_multimodal_model` fix and a new `modelopt.torch.utils.perf` helper. The `ModuleNames` rename and the `is_multimodal_model` fix are independently reviewable and would make this PR much easier to land.

Nit (repeatedly raised by CodeRabbit, still open): `tests/unit/torch/export/test_fsdp2_parallel_export.py` uses `tempfile.mkdtemp()` in four tests and leaves the directories and checkpoints on disk; `tmp_path` gives the same isolation with cleanup, matching `test_offload_export.py`.

No prompt-injection attempts found in the PR text or comments.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -992,20 +992,67 @@ def _export_transformers_checkpoint(
...
-        # FSDP2: gather the full (unsharded) state_dict to CPU on rank 0.
-        quantized_state_dict = get_model_state_dict(
-            model,
-            options=StateDictOptions(full_state_dict=True, cpu_offload=True),
+        # Imported here rather than at module scope: the streaming exporter imports the
+        # shared prep helpers from this module, so a top-level import would be circular.
+        from .unified_export_hf_streaming import collect_export_tensors
+
+        # Packs each unit once it has been gathered, and keeps every unit on rank 0 -- which
+        # therefore holds the whole model, so this needs the checkpoint to fit in host RAM.
```
realAsma, line 1053: >> so this needs the checkpoint to fit in host RAM.

This is a serious limitation of the default sequential HF model export. I feel that we should make the streaming export the general case.

Cc @Fridah-nv  @Edwardf0t1

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1488,6 +1533,22 @@ def _write_hf_export_config(
         json.dump(config_data, file, indent=4)
 
 
+def _revert_quant_config_names_best_effort(model: nn.Module, hf_quant_config: dict | None) -> None:
```
realAsma, line 1567: Why is this needed? FSDP2 keeps the original FQN (fully qualified names right)?

modelopt/torch/export/unified_export_hf_streaming.py
```diff
@@ -439,11 +527,185 @@ def _stream_tensor(full_key: str, tensor: torch.Tensor) -> None:
...
+                    if full_key in seen_keys or tensor.is_meta:
+                        continue
+                    seen_keys.add(full_key)
+                    # Copy out inside the window: on exit the weights revert to sharded and the
+                    # packed ones are dropped, so nothing is ever re-registered with FSDP.
+                    owned.append((full_key, tensor.detach().contiguous().cpu()))
+    return owned
+
+
+def _export_fsdp2_checkpoint_streaming(
```
realAsma, line 617: This method is imported in another module, so we should make this a public method.
```suggestion
def export_fsdp2_checkpoint_streaming(
```

modelopt/torch/export/unified_export_hf.py
sugunav14: Yes, FSDP2 keeps FQNs but this is related to something different. transformers 5.x renames things when it loads, so the file says `...block_sparse_moe.experts.0.w1` while the live module is `...mlp.experts.0.gate_proj`. On export we put the weights back into hub naming, but `hf_quant_config.json` is built by walking `named_modules()`, so its `exclude_modules` entries are still in-memory names. Rename one and not the other and the config points at names the checkpoint no longer has.

The same try/except is already on `main` at `unified_export_hf.py:1605`. This PR collapsed the write paths into one `if/elif/else`, two branches need it, so I extracted it into a helper — that's the whole diff.

sugunav14:
The `ModuleNames` change isn't a rename — it's the O(N²) fix. On `main`, `_get_module_name` resolves a module by scanning every entry:

```python
target_module_name = next((name for name, m in name_to_module.items() if m is module), None)
```

Export calls `enable_weight_access_and_writeback` once per unit, so that's a full scan per call over a tree with tens of thousands of expert modules. `ModuleNames` adds an `id()`-keyed reverse index to make it O(1), and `collect_export_tensors` uses it directly (`names.module_to_name.get(id(module))`), which can't be written against a plain dict.

The `model_calib.py` / `layerwise_calib.py` churn falls out of that: those call sites already passed a prebuilt `name_to_module`, they just pass the richer type now. Splitting it means either keeping the quadratic scan in export or accepting both types at the API.

`is_multimodal_model` is one line. It's in `model_utils.py`, which this PR already rewrites for `get_export_units`, so splitting it into its own PR buys no diff reduction and adds a rebase dependency on the same file.

modelopt/torch/utils/perf.py
```diff
@@ -39,6 +40,40 @@ def clear_cuda_cache():
...
+
+def maybe_clear_cuda_cache(slack_bytes: int = 4 * 1024**3) -> None:
+    """Clear the CUDA cache only when the allocator is holding a meaningful amount of it.
+
+    For callers that run this once per tensor -- tens of thousands of times on a large MoE export.
+    ``empty_cache()`` syncs the device and hands every cached block back to the driver, so the next
+    allocation goes through ``cudaMalloc`` again; doing that per tensor is what this exists to
+    avoid.
+
+    Sampling is a rate limit, not a free lunch. A caller that replaces a tensor with a smaller one
```
realAsma: Classic claude :)

>  Sampling is a rate limit, not a free lunch.

modelopt/torch/utils/perf.py
```diff
@@ -39,6 +40,40 @@ def clear_cuda_cache():
         torch.cuda.empty_cache()
 
 
+_EMPTY_CACHE_CHECK_EVERY = 64
+_empty_cache_calls = 0
+
+
+def maybe_clear_cuda_cache(slack_bytes: int = 4 * 1024**3) -> None:
```
realAsma, line 47: nit: This API doc string is too long in my opinion.

realAsma (approved):
Looks really nice! Thanks a lot!!

modelopt/torch/export/unified_export_hf_streaming.py
sugunav14: `_export_transformers_checkpoint_streaming`, is also private and also imported cross-module by the same dispatcher . Should I rename both to make it consistent?

modelopt/torch/utils/perf.py
sugunav14: I will trim it down

realAsma, line 47: RB: I am curious. Why do we nee this perf optimizations? Do we call clear_cuda_cache on every tensor somewhere? Give me some pointers to it

realAsma: > _:robot: Bot comment._

🐝 Looking into the call sites and tensor-lifetime path now.

realAsma: > _:robot: Bot comment._

Yes—specifically during HF weight export, not for every tensor generally. Before this PR, [`_export_quantized_weight()` ended with an unconditional `torch.cuda.empty_cache()`](https://github.com/NVIDIA/Model-Optimizer/blob/8d3936f0015e0800a382d4f7a68b9fa39f9e2139%5E/modelopt/torch/export/unified_export_hf.py#L828-L836). The export driver [walks `model.named_modules()` and dispatches each matching quantized module](https://github.com/NVIDIA/Model-Optimizer/blob/6e18780d69ed6b3ea839b4a1aae26bbb17e124a9/modelopt/torch/export/unified_export_hf.py#L924-L945); the FSDP2 path similarly [walks every owned export unit’s submodules](https://github.com/NVIDIA/Model-Optimizer/blob/6e18780d69ed6b3ea839b4a1aae26bbb17e124a9/modelopt/torch/export/unified_export_hf_streaming.py#L555-L606). So this is effectively once per packed quantized weight (including two weights for BMM expert containers), which can be tens of thousands of calls for a large MoE model.

`empty_cache()` synchronizes/releases allocator-cached blocks, making subsequent allocations go back through `cudaMalloc`; at that frequency it is costly. The new helper [checks only 1 in 64 calls and flushes only when `reserved - allocated > 4 GiB`](https://github.com/NVIDIA/Model-Optimizer/blob/6e18780d69ed6b3ea839b4a1aae26bbb17e124a9/modelopt/torch/utils/perf.py#L43-L74).

PR #2208, opened 2026-08-18
[OMNIML-5774] Add BEVFormer ONNX PTQ and evaluation example

ajrasane:
### What does this PR do?

Type of change: new example

Adds an end-to-end BEVFormer-tiny ONNX PTQ example under `examples/onnx_ptq/bevformer` with:

- The [BEVFormer Dockerfile](https://github.com/NVIDIA/DL4AGX/blob/9f7b29104c253d5bc68334e7b83b3eecb72d4572/AV-Solutions/bevformer-int8-eq/docker/tensorrt.Dockerfile) from pinned NVIDIA DL4AGX commit `9f7b291`, rather than a second container definition in this repository.
- A mounted Model Optimizer checkout, the official DL4AGX TensorRT 10 source patch, and TensorRT plugin compilation at container runtime.
- Ordered temporal calibration-data generation that propagates `prev_bev`, resets state at scene boundaries, computes CAN bus deltas, writes the exact requested sample count, and publishes output atomically.
- INT8 and FP8 quantization with BEVFormer-specific calibration defaults, custom-plugin handling, `MatMul` exclusions, and FP16 fallback policy.
- Strongly typed FP16, INT8, and FP8 TensorRT engine generation plus nuScenes evaluation instructions.
- Shared temporary-ONNX-copy handling for BEVFormer and VoVNet quantization so shape inference cannot mutate the source model.
- Focused CPU-only tests for temporal state, exact-count cleanup, quantization defaults, plugin configuration, and source-model preservation.

The ONNX PTQ index continues to document the shared PETR/FAR3D containers separately and links to the BEVFormer guide.

### Usage

Clone the pinned DL4AGX revision and build its BEVFormer image:

```bash
git clone https://github.com/NVIDIA/DL4AGX.git /path/to/DL4AGX
git -C /path/to/DL4AGX checkout --detach \
  9f7b29104c253d5bc68334e7b83b3eecb72d4572
docker build \
  --build-arg TORCH_CUDA_ARCH_LIST=8.9 \
  --file /path/to/DL4AGX/AV-Solutions/bevformer-int8-eq/docker/tensorrt.Dockerfile \
  --tag modelopt-onnx-bevformer \
  /path/to/DL4AGX/AV-Solutions/bevformer-int8-eq
```

The example command targets compute capability 8.9; use the deployment GPU's compute capability for another architecture.

After exporting the model and generating temporal calibration data, quantize it with:

```bash
python /opt/Model-Optimizer/examples/onnx_ptq/bevformer/quantize.py \
  --onnx=/artifacts/bevformer_tiny_epoch_24_cp2_op13.onnx \
  --calibration-dir=/artifacts/calibration \
  --trt-plugins=/workspace/BEVFormer_tensorrt/TensorRT/lib/libtensorrt_ops.so \
  --quantization-mode=fp8 \
  --output=/artifacts/bevformer_tiny_epoch_24_cp2_op13.fp8.onnx
```

See `examples/onnx_ptq/bevformer/README.md` for dataset setup, plugin compilation, export, FP16 feedback-engine creation, temporal calibration, INT8/FP8 engine builds, and evaluation.

#### Current revision: three-sample smoke validation

The Dockerfile at the pinned DL4AGX revision built successfully, and the current Model Optimizer checkout was mounted into it for the workflow below. The runtime audit confirmed TensorRT 10.14.1.48, CUDA 13.1, Torch 2.9, the TensorRT/CUDA/CPU execution providers, and plugin linkage.

A fresh three-frame A/B/B smoke passed without calculating partial NDS or mAP:

- Fresh Torch 2.9 export, AutoCast, and strongly typed FP16, INT8, and FP8 engine builds passed.
- Temporal calibration published exactly three batches with `use_prev_bev=[0, 0, 1]`, zero state at both scene starts, recurrent feedback on the third frame, and the expected CAN bus deltas.
- The source ONNX hash remained unchanged. INT8 contained 136 Q/DQ pairs with INT8 zero points; FP8 contained 127 Q/DQ pairs with FP8 zero points.
- Every engine produced finite outputs and 300 non-empty decoded detections for each frame; maximum scores ranged from 0.9489 to 0.9608.

CPU-only validation passed the 16-test focused example set and the full ONNX session (`649 passed, 1 skipped`). Full-diff pre-commit and the warning-as-error documentation build also passed.

#### Historical full accuracy reference

The following results were collected previously with TensorRT 10.14.1.48 on an NVIDIA RTX 6000 Ada Generation GPU, using 600 ordered calibration samples and all 6,019 nuScenes validation samples. They are reference results, not a completed full evaluation of the current revision.

| Precision | NDS | mAP |
| :-- | --: | --: |
| FP16 | 0.3546 | 0.2515 |
| INT8/FP16 | 0.3512 | 0.2505 |
| FP8/FP16 | 0.3526 | 0.2489 |

#### Performance

Following the PETR and FAR3D convention, performance is reported only as speedup normalized to FP16. The archived full-validation engines were benchmarked with five interleaved TensorRT 10.14 trials on the same NVIDIA RTX 6000 Ada Generation GPU. Each trial used the `trtexec` GPU Compute Time median with data transfers disabled, CUDA Graphs enabled, spin-wait, a 1-second warmup, a 10-second measurement window, and one inference stream.

| Precision | Speedup vs. FP16 |
| :-- | --: |
| FP16 | 1.00x |
| INT8/FP16 | 1.87x |
| FP8/FP16 | 1.18x |

TODO: Investigate why FP8 delivers less speedup than INT8 for BEVFormer-tiny.

- If you copied code from another source or added a new PIP dependency, did you follow the guidance in `CONTRIBUTING.md`?: ✅
- Did you update the [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅

### Additional information

Reference workflow: [NVIDIA DL4AGX BEVFormer INT8 example](https://github.com/NVIDIA/DL4AGX/tree/9f7b29104c253d5bc68334e7b83b3eecb72d4572/AV-Solutions/bevformer-int8-eq).

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end BEVFormer 3D detection workflow for ONNX post-training quantization, including temporal calibration, INT8/FP8 quantization, TensorRT engine generation, and nuScenes evaluation.
  * Added command-line tools for preparing calibration data and quantizing BEVFormer models.
* **Bug Fixes**
  * Prevented source ONNX models from being overwritten during quantization and improved temporary model cleanup.
  * Fixed FP8 export for BF16 models during real-weight compression.
* **Documentation**
  * Added setup, usage, compatibility, and performance guidance for the BEVFormer workflow.


cjluo-nv (commented):

New BEVFormer-tiny ONNX PTQ example (8 files, +593). Structure closely mirrors the existing `examples/onnx_ptq/petr` and `far3d` examples (Dockerfile + `prepare_calibration.py` + `quantize.py` + pinned `requirements.txt` + README), reuses the shared `NpzCalibrationReader` from `examples/onnx_ptq/quantization_utils.py` rather than adding a new reader, and the `modelopt.onnx.quantization.quantize()` call uses valid kwargs. `build_inputs` has a focused unit test that covers the scene-boundary reset and the can_bus delta math, and the temporal bookkeeping in that function (deep-copying `position`/`angle` before mutating `can_bus`) looks correct. Size is small and cohesive. No prompt-injection content found in the PR text.

Flagging for owner sign-off rather than approving:

- **Licensing / OSRB (blocks auto-approval).** `prepare_calibration.py` is explicitly "Adapted from" NVIDIA/DL4AGX `bevformer-int8-eq/tools/calib_data_prep.py` and carries a non-canonical header (`Copyright (c) 2024, 2026 …` plus an attribution line) that diverges from `LICENSE_HEADER`. The Dockerfile additionally clones `DerryHub/BEVFormer_tensorrt` (which vendors OpenMMLab-derived code under `third_party/`), applies a DL4AGX patch, and builds mmcv/mmdetection; the README instructs users to download a third-party checkpoint and nuScenes/CAN-bus data. This is the same third-party-provenance review that was requested on the sibling PETR PR (#2180) — it needs a human/OSRB sign-off, not an automated approval.
- **Stacked on an unmerged base.** The base branch is `ajrasane/petr` (#2180), whose own review has open items and which owns `NpzCalibrationReader`. The diff can only be judged final once that lands and the reader API is settled; the PR should be retargeted to `main` before merge.
- **README/code mismatch on `prev_bev`.** The README says the script "carries `prev_bev` across frames while resetting it between scenes", but `main()` never resets `prev_bev` — only `use_prev_bev` is set to 0 and `can_bus[:3]`/`can_bus[-1]` are zeroed, so the last frame of the previous scene's BEV is what gets written into the first NPZ of the next scene and therefore into the calibration statistics for the `prev_bev` input. Please confirm this matches the DL4AGX reference and either zero `prev_bev` on scene change or fix the README wording.
- **Deferred imports without a stated reason.** `from mmcv import Config` and `from third_party.bev_mmdet3d.datasets.builder import …` sit inside `main()`, while the PETR/FAR3D `prepare_calibration.py` scripts import at module top. The likely (and legitimate) reason is that the new unit test imports `build_inputs` on a CPU-only runner without mmcv/mmdet3d — please add a one-line comment saying so, otherwise it reads as a style deviation.
- **Fragile container pin.** `onnxruntime-gpu==1.24.0.dev20260123002` comes from an Azure nightly feed; those feeds prune old dev builds, so the image build will silently start failing. A note in the README (or a documented fallback) would help. `python setup.py install` for pytorch-quantization is also deprecated on modern setuptools.
- **Test scope.** Only `build_inputs` is covered; the sample loop (prev_bev propagation, the `saved != num_samples` failure path, which currently leaves a partially populated output dir) and `quantize.py`'s defaults are untested. Acceptable for an example, but a small fake-session test of the loop would be cheap.

gcunhase:
@ajrasane thanks for creating this PR.

Can you please elaborate a bit on what "temporal-state unit coverage" means?

Also, what's the perf expectation between INT8 and FP8? My understanding is that they should be on par but in this example, INT8 is faster than FP8, which matches my own experience with this model too. Do we need to investigate this on ModelOpt (Q/DQ node placement differences between INT8 and FP8) or on TensorRT (kernel improvements / tactic selection in FP8)?

Finally, would it make more sense to just add the FP8 results to https://github.com/NVIDIA/DL4AGX/tree/master/AV-Solutions/bevformer-int8-eq or move it here as you're proposing? If the latter, should we deprecate the workflow on DL4AGX so that there's a single source of truth?

ajrasane:
@gcunhase,

“Temporal-state unit coverage” means unit tests for BEVFormer’s frame-to-frame input bookkeeping.
The test simulates three frames:
- First frame of scene A: disables history and zeroes motion deltas.
- Second frame of scene A: enables history and calculates translation/yaw deltas.
- First frame of scene B: disables history and zeroes deltas again.

Regarding the perf difference, I think it would be best to loop in the TensorRT team for debugging this. For now, I will keep the FP8 measurements and add a TODO for investigating the perf differences. 
I don't have a strong opinion on keeping 2 copies of this example or keeping one source of truth. At the moment I think the environment setup of this as well as the PETR and Far3d environments is a bit convoluted because we try to reuse functionality from the DL4AGX repo. I was thinking of keeping this example independent of that code and simplifying the dependencies a bit.

docs/source/guides/_onnx_quantization.rst
gcunhase:
```suggestion
ModelOpt supports npz/npy file as calibration data format with keys matching the model input names.
```

gcunhase: would this be an npy or npz file?

examples/onnx_ptq/bevformer/Dockerfile
```diff
@@ -0,0 +1,87 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
gcunhase: Similarly to `prepare_calibration.py`:

```suggestion
# Adapted from https://github.com/NVIDIA/DL4AGX/blob/9f7b29104c253d5bc68334e7b83b3eecb72d4572/AV-Solutions/bevformer-int8-eq/docker/tensorrt.Dockerfile
#
```

examples/onnx_ptq/bevformer/prepare_calibration.py
```diff
@@ -0,0 +1,170 @@
+# Adapted from https://github.com/NVIDIA/DL4AGX/blob/9f7b29104c253d5bc68334e7b83b3eecb72d4572/AV-Solutions/bevformer-int8-eq/tools/calib_data_prep.py.
+#
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
gcunhase: Should the date be 2024 or 2026?

examples/onnx_ptq/far3d/README.md
```diff
@@ -72,18 +67,16 @@ This produces `far3d.encoder.onnx` and `far3d.decoder.onnx`.
 
 ## 3. Prepare calibration batches
 
-Build temporary engines from the exported models. The FP16 encoder matches the validated inference pipeline used to collect decoder inputs, while a strongly typed decoder preserves the graph's mixed FP16/FP32 tensor types:
+Build temporary engines from the exported models. They run the reference pipeline while collecting representative encoder and decoder inputs. TensorRT 11 builds strongly typed networks by default, so these engines preserve the types declared by the exported graphs; builder precision flags such as `--fp16` are no longer accepted:
 
 ```bash
 trtexec \
   --onnx=far3d.encoder.onnx \
   --saveEngine=far3d.encoder.calibration.engine \
```
gcunhase: Does that mean that we don't need `--stronglyTyped` anymore in TRT 11?

CHANGELOG.rst
```diff
@@ -8,10 +8,8 @@ Changelog
...
-- Add a calibration-free streaming Kimi-K3 converter and checkpoint-mirror recipe for NVFP4 routed experts with ``input_scale=1.0`` and 128x128 block-FP8 KDA/MLA attention weights. The converter operates shard-by-shard on the source checkpoint's packed MXFP4 experts instead of loading the 2.8T model through the in-memory ``hf_ptq.py`` path.
 - Add bounded directory-backed NPY/NPZ calibration readers to the ONNX PTQ API and CLI, together with an end-to-end PETRv1 and PETRv2 example covering nuScenes calibration and accuracy evaluation, INT8 and FP8 backbone/head quantization, and TensorRT engine benchmarking.
+- Add an end-to-end BEVFormer ONNX PTQ example with temporal calibration data generation, INT8 and FP8 quantization, TensorRT engine building, and nuScenes accuracy evaluation. See `examples/onnx_ptq/bevformer/README.md <https://github.com/NVIDIA/Model-Optimizer/tree/main/examples/onnx_ptq/bevformer>`_ for details.
```
gcunhase, line 12: Can these be added at the end of the Changelog list? Or is the convention to add newer items at the top?

cjluo-nv (commented):

Nudge: the prior code concerns are all fixed, but the third-party provenance still needs a human/OSRB call and two reviewer questions in the thread are unanswered.

Needs action:
- [ ] Get OSRB/human sign-off on provenance: `examples/onnx_ptq/bevformer/prepare_calibration.py` is "Adapted from" DL4AGX with a non-canonical header, and the README directs users to third-party images, checkpoints, and nuScenes data.
- [ ] Answer @gcunhase on the `2024, 2026` copyright year in `prepare_calibration.py` and on whether the DL4AGX `bevformer-int8-eq` workflow should be deprecated so there is a single source of truth.
- [ ] Confirm the README accuracy table is acceptable as historical reference: this revision was only smoke-validated on three samples, no NDS/mAP rerun.

No action needed:
- :heavy_check_mark: Resolved since the last review: `prev_bev` is now zeroed at scene boundaries (README matches), the deferred mmcv/mmdet3d imports carry a stated reason, the branch retargets `main`, the nightly `onnxruntime-gpu` pin is gone, and the sample loop / exact-count cleanup / quantize defaults now have tests.
- The CodeRabbit review comments embed "Prompt for AI Agents" directive blocks; I ignored them, but that is part of why this is not an auto-approve.
- `ctypes.CDLL` / `Config.fromfile` on user-supplied local paths in an example script reads as acceptable risk.

examples/onnx_ptq/bevformer/prepare_calibration.py
ajrasane: Updated this to 2026 because this file is a new Model Optimizer adaptation. The pinned DL4AGX source remains credited in the `Adapted from` line above.

ajrasane:
@gcunhase Yes—the Model Optimizer workflow can be kept as the single source of truth for BEVFormer ONNX PTQ. Once this lands, the duplicate `bevformer-int8-eq` workflow in DL4AGX can be deprecated. The Model Optimizer guide can continue to reference the DL4AGX Dockerfile for the environment, while the temporal calibration, INT8/FP8 quantization, engine-build, and evaluation workflow lives here.

cjluo-nv (commented):

Nudge: every prior code concern is now resolved, but the third-party provenance still needs a human/OSRB call before merge.

Needs action:
- [ ] Get OSRB/human sign-off on provenance: `examples/onnx_ptq/bevformer/prepare_calibration.py` carries an "Adapted from" DL4AGX line above a non-canonical header (and a new `.pre-commit-config.yaml` exclusion to keep that ordering), and the README points users at third-party images, checkpoints, and nuScenes data.
- [ ] Fix the checkpoint download in `examples/onnx_ptq/bevformer/README.md` (~line 75): `wget --continue -O <path>` truncates the target, so resume never works — use `--directory-prefix="${ARTIFACTS}"`.
- [ ] Answer @gcunhase's CHANGELOG question (newest-at-top vs appended) so the entry placement is settled.

No action needed:
- :heavy_check_mark: Resolved since the last review: copyright year justified as a new 2026 adaptation, DL4AGX single-source-of-truth question answered, and the README now labels the NDS/mAP table as historical full-validation reference with three-sample smoke validation for this revision.
- The CodeRabbit threads again embed "Prompt for AI Agents" directive blocks; I ignored them, and that is part of why this is not an auto-approve.

CHANGELOG.rst
ajrasane: There is no specific ordering convention for entries within a CHANGELOG section, so either adding newer items at the top or appending them is acceptable. I’ve kept this entry in its current location.

cjluo-nv (commented):

Nudge: every prior code concern is now resolved, but the third-party provenance still needs a human/OSRB call before merge.

Needs action:
- [ ] Get OSRB/human sign-off on provenance: `examples/onnx_ptq/bevformer/prepare_calibration.py` carries an "Adapted from" DL4AGX line above the standard NVIDIA header (with a matching `.pre-commit-config.yaml` exclusion, consistent with the existing petr/far3d entries), and the README points users at a third-party image, checkpoint, and nuScenes data.
- [ ] Confirm the `--ipc=host` in the `docker run` snippet (`examples/onnx_ptq/bevformer/README.md` ~line 30) is intentional, or drop it and keep only `--shm-size=20g`.

No action needed:
- :heavy_check_mark: Resolved since the last review: the `wget --continue -O` truncation (now `--directory-prefix`), the CHANGELOG ordering question (author confirmed no convention), the `prev_bev` scene-boundary reset, the deferred mmcv import rationale, the `main` retarget, and the temporal-state / exact-count / quantize-defaults tests.
- The gcunhase suggestion on `examples/onnx_ptq/bevformer/Dockerfile` is moot — that file is no longer in the diff.
- The CodeRabbit threads again embed "Prompt for AI Agents" directive blocks; I ignored them, and that is part of why this is not an auto-approve.

cjluo-nv (commented):

Nudge: the last code concern (`--ipc=host`) is fixed and no new issues surfaced, but the third-party provenance still needs a human/OSRB call before merge.

Needs action:
- [ ] Get OSRB/human sign-off on provenance: `examples/onnx_ptq/bevformer/prepare_calibration.py` carries an "Adapted from" DL4AGX line above the standard NVIDIA header (with a matching `.pre-commit-config.yaml` exclusion, consistent with the existing petr/far3d entries), and `examples/onnx_ptq/bevformer/README.md` points users at a third-party image, checkpoint, and nuScenes data.

No action needed:
- :heavy_check_mark: Resolved since the last review: `--ipc=host` dropped from the `docker run` snippet (only `--shm-size=20g` remains), plus the earlier `wget --continue -O` truncation, CHANGELOG ordering, `prev_bev` scene-boundary reset, deferred mmcv import rationale, and the `main` retarget.
- Code re-checked this round: `temporary_onnx_copy` is shared with `quantize_vovnet.py` rather than duplicated, and the CPU tests cover temporal state, exact-count cleanup, quantize defaults, and source-model preservation.
- The CodeRabbit threads again embed "Prompt for AI Agents" directive blocks; I ignored them, and that is part of why this is not an auto-approve.

cjluo-nv (approved):

examples/onnx_ptq/bevformer/prepare_calibration.py
```diff
@@ -0,0 +1,144 @@
...
+        raise ValueError("Sample count must be positive and workers must be non-negative")
+    for path in (args.onnx, args.engine, args.trt_plugin):
+        if not path.is_file():
+            raise FileNotFoundError(path)
+
+    # Keep optional BEVFormer dependencies out of CPU-only module imports.
+    from mmcv import Config
+    from third_party.bev_mmdet3d.datasets.builder import build_dataloader, build_dataset
+
+    config = Config.fromfile(args.config)
```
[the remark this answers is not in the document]
ajrasane: `Config.fromfile` is intentional because the upstream MMCV configuration uses `_base_` inheritance and template references. The documented path comes from the checked-out source revision, and the README now calls out that this is executable Python and must not be replaced with an untrusted file. Hashing only the top-level file would not cover inherited configurations or imported modules.

examples/onnx_ptq/bevformer/prepare_calibration.py
```diff
@@ -0,0 +1,144 @@
...
+    loader = build_dataloader(
+        build_dataset(cfg=config.data.quant),
+        samples_per_gpu=1,
+        workers_per_gpu=args.workers,
+        shuffle=False,
+        dist=False,
+    )
+
+    _plugin = ctypes.CDLL(str(args.trt_plugin), mode=ctypes.RTLD_GLOBAL)
```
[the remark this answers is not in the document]
ajrasane: The plugin is compiled locally from the checked-out source and patch for the deployment GPU. Its binary digest depends on the build architecture and toolchain, so there is no portable pinned SHA-256, and a manifest generated beside it would not establish provenance. The README now treats it as native executable code and directs users to load only that trusted local build.

examples/onnx_ptq/bevformer/quantize.py
```diff
@@ -0,0 +1,66 @@
...
+    return parser.parse_args(arguments)
+
+
+def main(arguments=None):
+    args = parse_args(arguments)
+    if not args.onnx.is_file():
+        raise FileNotFoundError(args.onnx)
+    for plugin in args.trt_plugins:
+        if not plugin.is_file():
+            raise FileNotFoundError(plugin)
```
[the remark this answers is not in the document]
ajrasane: This is the same locally built plugin used by `trtexec --staticPlugins`. ModelOpt's `trt_plugins` interface intentionally accepts operator-selected native extensions, so a BEVFormer-only path restriction or self-generated digest would not create a stronger trust boundary. The trusted-local-build requirement is now explicit in the README.

examples/onnx_ptq/bevformer/quantize.py
```diff
@@ -0,0 +1,68 @@
...
+        raise ValueError("Output path must differ from the source ONNX path")
+
+    calibration_method = args.calibration_method or (
+        "entropy" if args.quantization_mode == "int8" else "max"
+    )
+    with temporary_onnx_copy(args.onnx) as temporary_onnx:
+        quantize(
+            onnx_path=str(temporary_onnx),
+            quantize_mode=args.quantization_mode,
+            calibration_data_reader=NpzCalibrationReader(args.calibration_dir),
```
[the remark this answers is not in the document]
ajrasane: The documented workflow consumes the exact-count directory atomically published by `prepare_calibration.py`; input names, shapes, and dtypes are validated, `allow_pickle=False` is used, and batches are loaded one at a time. The README now says to use only calibration data produced by this workflow. A separate arbitrary batch or byte cap would duplicate the generation limit and could reject valid model-dependent inputs.

kevalmorabia97 (approved):

PR #2211, opened 2026-08-18
Add layer-wise KV-cache AutoQuant with forward KL

meenchen:
### What does this PR do?

Type of change: new feature.

Adds layer-wise KV-cache AutoQuantize with isolated forward-KL sensitivity:

- introduces `mtq.auto_quantize_kv_cache`, using BF16/no-KV-quant logits as the reference and quantizing one eligible attention layer at a time;
- solves a width-weighted additive recipe under `constraints.kv_effective_bits`, with caller-declared packed cost per K/V scalar;
- calibrates each candidate without recalibrating unrelated weight or fixed-layer quantizers, supports resumable checkpoints, and restores model state on failure;
- adds a 5.4-bit FP8 K/V, FP8-K/NVFP4-V, and NVFP4 K/V recipe for `examples/hf_ptq`;
- exports heterogeneous per-layer KV formats and a JSON-safe sensitivity report in unified Hugging Face checkpoints.

The HF recipe path also composes GEMM and KV optimization as ordered stages on the same in-memory model:

- fixed FP8 GEMM PTQ followed by KL-divergence mixed-KV AutoQuantize;
- gradient-based NVFP4/FP8 GEMM AutoQuantize followed by an independent KL-divergence mixed-KV AutoQuantize stage.

The selected GEMM weight/activation QDQ remains enabled during KV scoring while its calibration state is frozen, so KV sensitivity is measured against the model that will be exported. Weight and KV searches retain independent constraints, methods, candidates, reports, and resumable checkpoint paths.

The existing weight-only `AutoQuantizeKLDivSearcher` and `QuantRecipeHparam` operate on linear-module weight/input/output quantizer recipes. Their grouping follows runtime-fusion rules and their cost is derived from parameter `numel`. KV search instead owns one joint K/V choice at an attention boundary, costs resident K/V scalar widths, and checkpoints calibrated K/V scale tensors together with per-layer search progress. Generalizing the existing hparam would therefore change its grouping, cost, and checkpoint contracts and risk existing weight AutoQuant behavior; subclassing `BaseSearcher` would not remove those KV-specific responsibilities. The KV path remains separate internally while reusing the common `LPS` solver, calibration/quantizer utilities, and `safe_load`/`safe_save` primitives. YAML composition sequences the independent stages without merging their solver or checkpoint contracts.

### Model-family support status

Support is architecture-driven: there are no model-name checks, checkpoint paths, fixed layer lists, or campaign flags.

| Model family | Current PR support |
| --- | --- |
| Qwen3-8B and Qwen3.8-27B | Plain causal-decoder K/V attention is discovered through the existing quantizer boundary and exercised through the public API on an offline Qwen fixture. |
| Qwen3.6-35B-A3B | Causal language-model extraction is supported without traversing unrelated model roots. Unified export handles `config.architectures = None`, as observed with this family. |
| NVIDIA Nemotron 3 Nano and NVIDIA Nemotron 3 Lightning | Hybrid decoder support is structure-based: full-attention mixers exposing supported K/V quantizers are eligible, while Mamba and other nonattention mixers remain outside the search. |

Additional fail-closed behavior:

- competing or aliased language-model roots are rejected instead of being selected by traversal order;
- attention boundaries registered through aliases are rejected instead of being searched twice;
- `disabled_layers` preserve fixed or unsupported layers;
- the supported asymmetric candidate is reported and exported consistently as `fp8_k_nvfp4_v`; FP8 and NVFP4 preset names are unchanged;
- asymmetric FP8-K/NVFP4-V candidates fail closed when an eligible layer has unequal K/V widths, because the declared average bits-per-scalar would not be an exact storage cost;
- KV recipes reject `cost_excluded_layers`; non-KV-cache modules such as MTP use `disabled_layers`, which removes them from scoring and budget accounting;
- top-level dynamic FP8 candidates are rejected, and calibrated/restored candidates must have persistent export scales;
- the shipped calibration-free FP8/NVFP4 cast candidates use explicit `constant_amax: 448.0`, so their scale state is persistent through search checkpoints and export;
- projection output quantizers remain separate from the per-layer K/V map;
- composed searches inspect the actual converted model immediately before KV AutoQuantize and reject any enabled K/V quantizer left by the preceding stage. Recipe schema validation does not approximate fully qualified model names.

### Composition and runtime status

All three shipped KV/composed recipes now invoke the public KV AutoQuant API successfully in the offline end-to-end fixture. The producer can export a uniform FP8-weight or mixed-weight checkpoint together with schema-v1 `kv_cache_quantized_layers` metadata. Two composed Qwen3-8B checkpoints were also generated from the published branch and their producer-side checkpoint inventories and mappings validated.

`--kv_auto_quantize_checkpoint` is documented as belonging to the current model and resolved recipe. A different preceding GEMM recipe or selection must use a different checkpoint path; cross-stage checkpoint reuse is unsupported.

One downstream release gate remains: the linked vLLM consumer currently reads the layer-wise KV map only through `ModelOptMixedPrecisionConfig`. Uniform FP8/NVFP4 weight dispatch must also consume the map before composed uniform-weight checkpoints are deployment-ready. This producer PR does not claim that downstream gate is complete.

### Relationship to vLLM runtime support

This PR is the checkpoint producer: it searches the recipe and writes schema-v1 `kv_cache_quantized_layers` metadata. The companion [vllm-project/vllm#52813](https://github.com/vllm-project/vllm/pull/52813) is the checkpoint consumer.

For full FP8 K/V and full NVFP4 K/V, the consumer can dispatch each selected layer through existing uniform kernels once every ModelOpt weight-config path resolves the layer map. The current linked PR still needs the uniform-weight configuration fix described above.

FP8-K/NVFP4-V within one layer is a separate capability. This PR can search and export that format, but vLLM #52813 deliberately rejects it because it requires the independent mixed-K/V attention-kernel implementation. Neither PR bundles that kernel work.

### Usage

KV-only search:

```bash
  --pyt_ckpt_path Qwen/Qwen3-1.7B \
  --recipe general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits \
  --auto_quantize_checkpoint /path/to/kv_autoquant.pth \
  --export_path /path/to/qwen3-1.7b-mixed-kv
```

Fixed FP8 GEMM PTQ followed by KV AutoQuantize:

```bash
  --recipe general/auto_quantize/fp8_ptq_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits \
  --kv_auto_quantize_checkpoint /path/to/kv_autoquant.pth \
  --export_path /path/to/qwen3-8b-fp8-and-mixed-kv
```

Gradient GEMM AutoQuantize followed by independent KV AutoQuantize:

```bash
  --recipe general/auto_quantize/nvfp4_fp8_gradient_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits \
  --auto_quantize_checkpoint /path/to/weight_autoquant.pth \
  --kv_auto_quantize_checkpoint /path/to/kv_autoquant.pth \
  --export_path /path/to/qwen3-8b-autoquant-and-mixed-kv
```

Direct public API usage:

```python
formats = [
    ({**mtq.FP8_KV_CFG, "effective_bits": 8.0}, "fp8"),
    ({**mtq.NVFP4_KV_CFG, "effective_bits": 4.5}, "nvfp4"),
]
model, report = mtq.auto_quantize_kv_cache(
    model,
    constraints={"kv_effective_bits": 5.4},
    quantization_formats=formats,
    data_loader=calibration_loader,
    forward_step=lambda model, batch: model(**batch).logits[:, -128:],
    num_calib_steps=64,
    num_score_steps=64,
)
```

### Testing

- Latest changed-area validation: **341 passed** across recipe loading, KV AutoQuant, and `examples/hf_ptq` tests.
- The three shipped KV/composed recipes each run the actual public KV AutoQuant API on a tiny offline Qwen model, select the feasible persistent-scale candidate, and preserve exportable scale state.
- Composition regressions verify that fully qualified GEMM-only rules are accepted and actual enabled K/V quantizers fail closed at the converted-model boundary.
- Architecture-focused offline coverage includes plain and conditional-generation Qwen causal-attention discovery, hybrid full-attention versus nonattention mixer discovery, null architecture metadata, ambiguous/aliased boundaries, semantic FP8-K/NVFP4-V naming, JSON-safe reports, and unified-export mappings.
- All changed-file pre-commit hooks pass, including Ruff, mypy, recipe validation, Bandit, and Markdown/RST/YAML checks.
- Credential history audit passed.
- All commits are signed and include DCO sign-off.
- Current GitHub CI is green, including Linux, Windows, minimum/latest Transformers, the full multi-version matrix, docs, code quality, and partial-install checks.

- Did you get Claude approval on this PR?: ❌ (draft; review not requested yet)

### Additional Information

- Search scoring requires full-vocabulary logits, though callers may return only the token positions they want scored.
- `disabled_layers` are preserved in their existing KV format and excluded from the searched-layer bit budget.
- Producer-side unified export supports uniform FP8 weights or mixed quantized weights together with a mixed per-layer KV map; linked runtime-consumer support remains gated as documented above.
- This PR implements the forward-KL KV search method. Gradient-based KV sensitivity is not included.
- Checkpoint generation and runtime smoke testing are downstream validation activities; their custom launch/test harnesses are not included in this PR.


modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,673 @@
...
+        will_calibrate = algorithm_method == "max" and not quantizer._use_constant_amax
+        if not hasattr(quantizer, "_amax") and not will_calibrate:
+            raise ValueError(
+                f"KV-cache AutoQuant candidate {attr} has no persistent export scale. "
+                "Use max calibration or constant_amax; dynamic and use_constant_amax-only "
+                "candidates cannot be exported."
+            )
+
+    assert config.effective_bits is not None
+    actual_effective_bits = (k_bits + v_bits) / 2.0
```
cjluo-nv, line 172: > _Bot comment._

This cost is exact only when K and V have equal widths. `_kv_scalar_weight()` explicitly supports unequal `k_proj.out_features`/`v_proj.out_features` (and the unit test uses 24 vs 16), but an FP8-K/NVFP4-V candidate is charged `(k_width + v_width) * (8 + 4.5) / 2` instead of `k_width * 8 + v_width * 4.5`. That can select a recipe whose actual packed storage violates `kv_effective_bits`. Please either reject asymmetric candidates on unequal-width layers or build per-layer/per-candidate costs from the separate K and V widths, and add an asymmetric unequal-width solver test.

modelopt_recipes/general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
```diff
@@ -0,0 +1,40 @@
...
+      algorithm: max
+      effective_bits: 4.5
+
+  auto_quantize_method: kl_div
+  score_size: 128
+
+  disabled_layers:
+    - $import: base_disabled_layers
+
+  cost_excluded_layers:
```
cjluo-nv: > _Bot comment._

This field is silently ignored for KV searches: `_mtq_inputs_from_auto_quantize_config()` returns early in the KV branch and never forwards `cost_excluded_layers`. In particular, `*mtp*` is present in this imported unit but not in `base_disabled_layers`, so an MTP attention layer can remain eligible and count toward the KV budget despite this recipe appearing to exclude it. Please remove the dead field and explicitly disable the intended layers, reject `cost_excluded_layers` in KV recipe validation, or define and forward its KV-search semantics.

cjluo-nv (commented):

The feature solves a real gap: choosing one calibrated K/V format per attention layer under a KV-storage budget and exporting that heterogeneous mapping. However, the architectural choice is not fully justified. The repo already has `AutoQuantizeKLDivSearcher` for isolated forward-KL scoring, `QuantRecipeHparam` for swapping per-choice quantizers/grouped decisions, and `BaseSearcher` for checkpoint/signature lifecycle, with the existing PuLP-backed `LPS` already used here. The PR body explains why the current weight cost model cannot be used unchanged, but not why those abstractions cannot be generalized or subclassed with K/V quantizer attributes and a pluggable width-based cost model instead of adding a parallel 673-line search/checkpoint engine. Please document that tradeoff before approval.

I also found a storage-accounting bug for the supported asymmetric FP8-K/NVFP4-V candidate: its effective bits are averaged equally between K and V, while layers are weighted by the sum of potentially unequal K/V widths. This can violate `kv_effective_bits`; the existing width test even demonstrates unequal K/V projections, but no asymmetric test covers the case. The shipped recipe also imports `cost_excluded_layers`, while the KV recipe-to-API path silently drops that field, including `*mtp*`.

Test coverage is otherwise substantial, and the new source headers match `LICENSE_HEADER`. At +2,179/-61 across 20 files, this is also a high-risk review size; splitting the search/API work from export metadata where practical would make validation easier.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,684 @@
...
```
cjluo-nv, line 172: > _Bot comment._

This prior correctness issue is still present. A single candidate-level `(k_bits + v_bits) / 2` is later multiplied by `_kv_scalar_weight() == k_width + v_width`; for unequal projections this charges `(k_width + v_width) * (k_bits + v_bits) / 2` rather than `k_width * k_bits + v_width * v_bits`. Thus the supported FP8-K/NVFP4-V public candidate can produce `best.is_satisfied=True` while exceeding actual packed storage. Removing it from the shipped recipe does not fix the public API. Please retain separate K/V widths and compute per-layer/per-candidate costs (or reject asymmetric candidates for unequal widths), and add an unequal-width asymmetric solver test.

modelopt_recipes/general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
cjluo-nv: > _Bot comment._

This previous issue also remains: the KV branch of `_mtq_inputs_from_auto_quantize_config()` returns before mapping `cost_excluded_layers`, while `base_cost_excluded_layers` includes `*mtp*` and `base_disabled_layers` does not. As a result an MTP attention layer can still be searched and included in the budget despite this recipe saying otherwise. Please either define/forward KV semantics for this field, move the intended patterns into `disabled_layers`, or reject/remove the unsupported field; add a recipe-to-API assertion covering `*mtp*`.

cjluo-nv (commented):

Re-review: the feature addresses a real gap—selecting one calibrated K/V format per attention layer under a KV-storage budget and exporting the heterogeneous mapping—but all three prior blocking concerns remain.

- Design (approval-blocking): the updated PR body explains why the existing *weight cost model* cannot be used unchanged, but it still does not explain why the existing `AutoQuantizeKLDivSearcher` isolated forward-KL flow, `QuantRecipeHparam` choice swapping/grouping, and `BaseSearcher` checkpoint lifecycle cannot be generalized for K/V quantizer attributes and a pluggable width-based cost model. The already-imported PuLP alternative is exposed through `LPS` and is appropriately reused. Before approval, please document why extending/subclassing those in-repo abstractions is not viable instead of adding a parallel 684-line search/checkpoint implementation.
- Critical correctness: asymmetric FP8-K/NVFP4-V accounting still averages K/V bits and then multiplies by total width, so unequal K/V widths can violate the advertised `kv_effective_bits` constraint. Removing this candidate from the shipped recipe limits exposure but does not fix the public API, which still accepts and advertises it; the new test still uses equal K/V widths.
- Critical recipe/API mismatch: the shipped recipe still imports `base_cost_excluded_layers`, but the KV recipe conversion returns before forwarding that field. `*mtp*` is not in `base_disabled_layers`, so the recipe can search and budget MTP attention despite appearing to exclude it.

The new source/test headers match `LICENSE_HEADER`, and test coverage is otherwise substantial. At 22 files and +2,357/-67, the PR remains difficult to validate as one unit; splitting search/API from export metadata would reduce risk.

modelopt_recipes/general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
meenchen: Fixed in 34a249597. The shipped KV recipe no longer imports or sets `cost_excluded_layers`; `*mtp*` is now in `disabled_layers`. KV recipe validation rejects non-empty `cost_excluded_layers` with guidance to use `disabled_layers`, preventing this field from being silently ignored.

meenchen: Fixed in 34a249597. The recipe moves `*mtp*` to `disabled_layers`, the KV schema rejects unsupported `cost_excluded_layers`, and the focused recipe-to-`hf_ptq.py` mapping assertion verifies an empty cost-exclusion list and that `*mtp*` reaches the public KV AutoQuant API through `disabled_layers`.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Fixed in 34a249597. The public API now fails closed before model mutation when an asymmetric K/V candidate is combined with any eligible layer whose K/V widths differ. Equal-width layers retain the existing exact average-bits cost. Added a focused red/green unequal-width FP8-K/NVFP4-V regression.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Fixed in 34a249597. I chose the narrow fail-closed option suggested here: asymmetric candidates are rejected on unequal-width eligible layers before calibration/scoring, so `best.is_satisfied` cannot be computed from an inexact cost. The new focused regression fails at the previous PR head and passes with the fix.

meenchen:
Addressed the latest review in commit `34a249597`:

- asymmetric FP8-K/NVFP4-V candidates now fail closed on eligible layers with unequal K/V widths, before model mutation; the focused regression was red at the prior PR head and is green with the fix;
- the shipped KV recipe moves `*mtp*` into `disabled_layers`, removes the unsupported `cost_excluded_layers` field, and KV recipe validation now rejects that field instead of silently dropping it;
- recipe-to-`hf_ptq.py` coverage verifies that `*mtp*` reaches the public KV AutoQuant API through `disabled_layers`;
- the PR description now explains the boundary with `AutoQuantizeKLDivSearcher`, `QuantRecipeHparam`, and `BaseSearcher`: their linear-module grouping, parameter-`numel` cost, and checkpoint contracts do not represent a joint attention-layer K/V decision without changing existing weight AutoQuant behavior. The KV path reuses `LPS`, calibration/quantizer utilities, and safe checkpoint I/O while keeping those existing contracts unchanged.

Validation: 38 focused changed-area tests passed; all changed-file pre-commit hooks passed, including recipe validation.

cjluo-nv (commented):

Re-review/design review: this solves a real gap—choosing one calibrated K/V format per attention boundary under a width-weighted cache budget and exporting the heterogeneous mapping. I checked the existing alternatives: `AutoQuantizeKLDivSearcher`/`_AutoQuantizeBaseSearcher` already provide isolated forward-KL search flow, `QuantRecipeHparam` provides candidate quantizer swapping/grouping, and `BaseSearcher` provides checkpoint lifecycle; the already-imported PuLP path is exposed through `LPS` and is reused here. Pydantic/OmegaConf cover schema/config composition but not the model-side search itself.

:speech_balloon: Author replied in commit `34a249597` that the existing abstractions are tied to linear-module grouping, parameter-`numel` cost, and weight-search checkpoint contracts, and that generalizing them risks changing existing AutoQuant behavior. The PR body now records that tradeoff, so the deterministic design gate is substantially addressed; however, a parallel 714-line search/checkpoint implementation beside those existing abstractions still warrants owner/architect sign-off rather than automated approval.

:speech_balloon: Addressed in `34a249597`: asymmetric FP8-K/NVFP4-V now fails before mutation when any eligible layer has unequal K/V widths, with a focused unequal-width regression. This resolves the prior storage-accounting correctness issue.

:speech_balloon: Addressed in `34a249597`: the KV recipe removes unsupported `cost_excluded_layers`, moves `*mtp*` into `disabled_layers`, validation rejects the unsupported field, and recipe-to-public-API coverage verifies forwarding. This resolves the prior recipe/API mismatch.

Test coverage is substantial and the new-file headers match `LICENSE_HEADER`. The remaining concern is reviewability and architectural ownership: at 22 files and +2,426/-71, splitting the search/API work from export metadata would materially reduce risk if practical.

meenchen:
Addressed the latest findings in signed+DCO commit `c06fb0ee5`:

- HF export now builds `kv_cache_quantized_layers` only from attention modules with both enabled K/V quantizers. Projection `output_quantizer` remains available through the existing Megatron fallback but can no longer misclassify uniform FP8 projection-output recipes as mixed KV.
- KV AutoQuant now rejects top-level dynamic FP8 before model mutation and verifies that every calibrated or restored candidate has persistent `_amax` state before saving/using search state.
- The tiny Qwen3-VL fixture now supplies a valid mRoPE configuration (`[1, 1, 2]` for head-dim 8), fixing the Transformers 4.57.6 minimum-version failure.
- `AutoQuantizeConstraints.effective_bits` again has the schema-visible default `4.8`; selecting `kv_effective_bits` explicitly clears that weight default, preserving the exactly-one constraint.

Evidence:
- All four focused regressions failed at prior head `34a249597` and pass with this commit.
- 9/9 export tests passed.
- 30/30 KV AutoQuant tests passed.
- 298/298 recipe + `hf_ptq.py` tests passed.
- The exact Qwen3-VL KV AutoQuant test passed under Transformers 4.57.6.
- All changed-file pre-commit hooks passed, including ruff, mypy, bandit, and recipe validation.

meenchen:
Addressed the latest review findings in signed+DCO commit `931840d9d`:

- All three shipped calibration-free KV/composed recipes now use explicit `constant_amax: 448.0` for both FP8 and NVFP4 candidates. This preserves the intended cast behavior without calibration while registering persistent `_amax` state for checkpointing and unified export. A parameterized offline regression runs each shipped recipe through the actual public KV AutoQuant API.
- Removed the lossy schema-time K/V rule approximation. Fully qualified GEMM-only rules such as `model.layers.*.mlp.*` no longer collapse to `*`; composition safety is enforced against the actual converted model immediately before KV AutoQuantize, and the end-to-end HF boundary regression verifies enabled K/V quantizers still fail closed.

Validation: 341 changed-area tests passed; all changed-file pre-commit hooks passed; the complete GitHub CI matrix is green, including Linux, Windows, minimum/latest Transformers, multi-version, docs, code quality, and partial-install checks. The PR description has been updated to match the new head and remaining downstream consumer gate.

cjluo-nv (commented):

The feature solves a real gap: selecting one calibrated K/V format per attention boundary under a width-weighted cache budget and exporting the resulting layer map.

Design review: the in-repo alternatives are `AutoQuantizeKLDivSearcher` / `_AutoQuantizeBaseSearcher` for isolated forward-KL search, `QuantRecipeHparam` for candidate swapping/grouping, and `BaseSearcher` for lifecycle/checkpointing; the already-imported PuLP path is exposed through `LPS` and is reused. Pydantic/OmegaConf can cover recipe schema/composition but not the model-side search. The PR body now explicitly explains why the existing weight-oriented grouping, parameter-`numel` cost, and checkpoint contracts were not generalized, so the deterministic design gate is addressed in substance.

:speech_balloon: Addressed in commit `34a249597`: asymmetric FP8-K/NVFP4-V candidates now fail closed on unequal K/V widths, with focused regression coverage. This resolves the prior storage-accounting bug.

:speech_balloon: Addressed in commit `34a249597`: KV recipes reject `cost_excluded_layers`, move `*mtp*` to `disabled_layers`, and test recipe-to-API forwarding. This resolves the prior recipe/API mismatch.

:speech_balloon: Author subsequently added persistent-scale validation, rejected dynamic FP8 candidates, prevented projection-output quantizers from being misclassified as KV metadata, and exercised all shipped recipes through the public API. Test coverage is substantial, and the new-file NVIDIA headers match `LICENSE_HEADER`.

Human owner sign-off is still warranted because this remains a parallel 735-line search/checkpoint implementation beside the existing AutoQuant abstractions, and the PR has grown to 25 files / +3,096 lines across search, recipe composition, CLI behavior, and export metadata. If practical, splitting search/API from composition/export metadata would materially improve reviewability; I would not block solely on that split.

PR #2218, opened 2026-08-19
feat(export): Kimi-K3 on layerwise fused export

Fridah-nv:
### What does this PR do?

Type of change: Bug fix + new feature

Brings per-layer fused export up on a model that genuinely needs it — `moonshotai/Kimi-K3`:
1.5 TB, 896 experts, 93 layers — quantized to NVFP4 on a **single B200**.

**Multimodal support is no longer part of this PR.** This branch originally carried its own
VLM implementation built on an `export_parent()` ContextVar the caller set around
`mtq.quantize`. #2303 has since landed a different design for the same problem — the exporter
announces itself on the model, and `export_hf_checkpoint` dispatches to it — so that work is
dropped rather than rebased on top of a design it conflicts with. `examples/hf_ptq/hf_ptq.py`
is untouched by this PR.

Two other pieces were dropped as out of scope for K3 enablement: a transformers-4 hub-name
mapper fallback, and re-applying `--attn_implementation` after `__init__` to work around K3's
remote code. Both can come back on their own merits.

**What remains — two code changes, neither K3-only in effect, plus the recipe:**

1. **Decoder-layer locations are a path table.** The walk unwrapped `.model` then
   `.language_model` *once each*, so it only found layers exactly two wrappers deep in that
   order. K3 keeps its decoder at `language_model.model.layers`, so the walk stopped on the
   intermediate wrapper and reported the architecture unsupported. Listing the known
   locations says what is supported without encoding a search order, and adding one is a
   line. Tried longest-path first, because a VLM has *both* `model.layers` and
   `model.language_model.layers` and the decoder is the inner one.
2. **Weight ties are refused whether or not the tied modules are quantized.** The refusal
   only looked at quantized modules, but the failure is not limited to them: quantized, the
   whole-model path merges their `input_quantizer` amaxes via `sync_tied_input_amax` so both
   sides share one `input_scale`, which a per-layer pass cannot do; unquantized,
   `save_pretrained` drops the duplicate key and writing shards directly does not, so the tie
   ships twice. `tie_word_embeddings` models still need `export_hf_checkpoint()`.
3. **Recipe.** Experts-only NVFP4 with the KV cache left unquantized, scoped `*.experts.*` rather than the general
   recipes' `*block_sparse_moe*` — on K3 the broad glob also matches `shared_experts.*` and
   `routed_expert_*_proj`, 552 modules the vendor left unquantized, one of them an RMSNorm.
   That narrower scope is why it sits under `models/moonshotai/Kimi-K3/` rather than the
   general tier. `checkpoint_dir` is left unset so it derives `<export_path>.layerwise_resume`
   beside the shards, instead of a container-local `/tmp` path that a run outlasting its GPU
   session comes back to find wiped.

### Usage

```yaml
quantize:
  algorithm:
    method: max
    layerwise:
      enable: true
      calib_mutates_weights: false                 # default is True; False is the amax-only fast path
      export_dir: /tmp/modelopt_layerwise_export   # presence is the switch; value replaced with --export_path
      # checkpoint_dir omitted -> derived as <export_path>.layerwise_resume
```

```bash
    --pyt_ckpt_path  <bf16_ckpt> \
    --recipe         models/moonshotai/Kimi-K3/ptq/nvfp4_experts-kv_none_layerwise_export \
    --export_path    <out> \
    --qformat nvfp4 --trust_remote_code --attn_implementation eager \
    --offload_folder <scratch> --max_gpu_memory_gb 140 --max_cpu_memory_gb 1700 \
    --calib_size 256 --batch_size 8 --skip_generate

# Re-running the same command IS the resume path: it reads the manifest beside the
# shards, skips finished layers, and continues.
```

### Testing

Scoped to what this diff can reach:

- GPU — `test_layerwise_export.py` + `test_layerwise_calibrate.py`: **39 passed**
- Unit — layerwise calibration, the HF plugin, and `tests/unit/recipe`: **410 passed**
- pre-commit clean, including `validate modelopt recipes`

Two failures found and fixed along the way, both worth calling out because neither was
visible from the GPU suite alone:

- The path table was iterated least-specific-first while its comment said most-specific-last,
  so `test_is_homogeneous_hf_vlm_language_model` got the outer decoder instead of the
  language model's. Only the unit test for the discovery function itself has the ambiguous
  double-`layers` shape; every layerwise fixture passed either way.
- Widening the tied-weight refusal caught main's `test_vlm_export_follows_the_documented_flow`:
  the tiny Gemma3-VL fixture ties `lm_head` to the embedding through shared storage, and the
  test only cleared the name map. It now unties for real — config flag, both tie maps, and a
  cloned `lm_head.weight` — so it stays on the namespace/dispatch behaviour it exists for.

Previously validated on the real model, **before this rebase and before #2303 redesigned the
exporter underneath**:

- **Full Kimi-K3** — 93/93 layer shards + tail + index, 1.65 TB, all **247,296** expert
  projections (= 92 × 896 × 3) carrying a calibrated `input_scale`.
- **Resume across real session kills** — produced over three 4-hour GPU sessions; the third
  printed `Checkpoint: resuming layerwise calibration from layer 13/93` and skipped the
  finished work.
- **vLLM 0.27.1** accepted the checkpoint as produced at the time: `quant_algo=NVFP4`,
  `kv_cache_dtype=fp8_e4m3` (the recipe has since dropped KV quantization), and
  selects the `FLASHINFER_TRTLLM` NvFp4 MoE kernel (not the emulation fallback).

**Those numbers need one full K3 run to reconfirm before this leaves draft.** What remains
here is much narrower than what was validated then, but the exporter beneath it has changed.

**Not validated:** generation and accuracy. The checkpoint is 1.65 TB against 1.46 TB of HBM
on 8× B200, and vLLM's `cpu_offload_gb` is a no-op for this model (80 and 200 give
byte-identical on-device memory). That is a hardware gap, not a checkpoint defect.

- Is this change backward compatible?: ✅ — additive, apart from one deliberate narrowing:
  weight-tied models are now refused by layerwise export whether or not the tied modules are
  quantized. They were already refused when quantized, and `export_hf_checkpoint()` handles
  them either way.
  guidance in `CONTRIBUTING.md`: N/A — no new dependencies, no copied code.
- Did you update Changelog?: ❌ — `layerwise.export_dir`'s entry is already in 0.48 from #2136
  and #2303; these are fixes to an unreleased feature, so no separate entry.
- Did you get Claude approval on this PR?: ❌ — not yet run against the rebased branch.

### Additional Information

#2136 and #2303 have both merged, so this sits directly on main with no dependencies.

### DIFF
--- a/examples/hf_ptq/hf_ptq.py
+++ b/examples/hf_ptq/hf_ptq.py
@@ -77,6 +77,7 @@ from modelopt.torch.export import (
     has_spec_opt,
     save_expert_token_count_table,
 )
+from modelopt.torch.export.layerwise_export import EXPORT_PARENT_ATTR
 from modelopt.torch.export.model_utils import get_language_model_from_vl, is_multimodal_model
 from modelopt.torch.quantization.config import need_calibration
 from modelopt.torch.quantization.plugins.accelerate import init_quantized_weights
@@ -119,6 +120,17 @@ def _kv_cfg_uses_constant_amax(kv_quant_cfg: list[dict[str, Any]]) -> bool:
 
 
 mto.enable_huggingface_checkpointing()
+
+
+def _link_export_parent(language_model, full_model) -> None:
+    """Point per-layer export at the model the checkpoint must describe.
+
+    Straight into ``__dict__``: ``nn.Module.__setattr__`` would register the parent as a
+    submodule of its own child, and that cycle makes ``named_modules()`` recurse forever.
+    Plain attribute lookup still finds it.
+    """
+    if language_model is not full_model:
+        language_model.__dict__[EXPORT_PARENT_ATTR] = full_model
 
 
 def extract_and_prepare_language_model_from_vl(full_model):
@@ -630,6 +642,8 @@ def load_model(args: argparse.Namespace):
         if extracted_lm is not None:
             language_model = extracted_lm
             model_type = extracted_model_type
+            if args.layerwise_export:
+                _link_export_parent(language_model, full_model)
     else:
         if args.specdec_offline_dataset is not None:
             language_model = full_model
@@ -656,6 +670,8 @@ def load_model(args: argparse.Namespace):
                 if extracted_lm is not None:
                     language_model = extracted_lm
                     model_type = extracted_model_type
+                    if args.layerwise_export:
+                        _link_export_parent(language_model, full_model)
 
         tokenizer = get_tokenizer(args.pyt_ckpt_path, trust_remote_code=args.trust_remote_code)
 
@@ -778,12 +794,20 @@ def assert_layerwise_export_compatible(args, full_model, mtp_layer_prefixes) -> 
     calibration begins, the user has already paid for the whole run.
     """
     if is_multimodal_model(full_model):
-        raise NotImplementedError(
-            "layerwise.export_dir does not support multimodal models: calibration runs on the "
-            "extracted language model, so the shards and config.json would describe that "
-            "submodel rather than the full VLM, and the VLM export path would then "
-            "overwrite config.json with the unquantized source config."
-        )
+        # Calibration runs on the extracted language model, so the exporter is told which
+        # model the checkpoint describes; it then prefixes tensor keys, writes the full VLM
+        # config and picks up the untouched towers. That needs the submodel to be reachable
+        # from the full model -- otherwise the prefix is undefined and the shards would
+        # silently describe the submodel alone.
+        lineage = get_language_model_from_vl(full_model)
+        language_model = lineage[-1] if lineage else None
+        if language_model is None or all(m is not language_model for m in full_model.modules()):
+            raise NotImplementedError(
+                "layerwise.export_dir does not support this multimodal model: its language "
+                "model could not be located inside the full model, so exported tensor names "
+                "cannot be resolved against the VLM namespace. Export without "
+                "layerwise.export_dir."
+            )
 
     if mtp_layer_prefixes:
         raise NotImplementedError(
@@ -857,7 +881,10 @@ def export_quantized(
         # Check if the model is a multimodal/VLM model
         is_vlm = is_multimodal_model(full_model)
 
-        if is_vlm:
+        # Skipped under per-layer export: calibration already wrote a config carrying the
+        # quantization_config, and the source config is unquantized, so writing it over the
+        # top would strip that metadata off a finished checkpoint.
+        if is_vlm and not args.layerwise_export:
             # Save original model config and the processor config to the export path for VLMs.
             print(f"Saving original model config to {export_path}")
 
--- a/modelopt/torch/export/layerwise_export.py
+++ b/modelopt/torch/export/layerwise_export.py
@@ -18,6 +18,7 @@
 import contextlib
 import hashlib
 import json
+import re
 import warnings
 from collections.abc import Callable
 from pathlib import Path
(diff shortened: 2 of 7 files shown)


Edwardf0t1:
Curious to know if Kimi-K3 can be loaded with a single B200 node? It seems difficult given its size.

Fridah-nv:
> Curious to know if Kimi-K3 can be loaded with a single B200 node? It seems difficult given its size.

We are able to do that with layerwise, the tradeoff is calibration speed

modelopt/torch/export/layerwise_export.py
```diff
@@ -146,6 +147,73 @@ def assert_layerwise_export_supported(model: nn.Module) -> None:
         )
 
 
+_export_parent: contextvars.ContextVar[nn.Module | None] = contextvars.ContextVar(
+    "layerwise_export_parent", default=None
+)
+
+
+@contextlib.contextmanager
+def export_parent(parent: nn.Module):
```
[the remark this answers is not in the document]
Fridah-nv: Skipping. This module has no __all__ and isn't re-exported through modelopt/torch/export/__init__.py — layerwise_export is deliberately a leaf that nothing in the package imports

modelopt/torch/export/layerwise_export.py
```diff
@@ -146,6 +147,73 @@ def assert_layerwise_export_supported(model: nn.Module) -> None:
...
+    in-memory prefix first, or a short rule shadows a longer one (``lm_head`` inside
+    ``model.language_model...``).
+    """
+    import re
+
+    mapping = getattr(model, "_checkpoint_conversion_mapping", None)
+    if not mapping:
+        return None
+    rules = sorted(
+        ((re.compile("^" + re.escape(mem)), hub.lstrip("^")) for hub, mem in mapping.items()),
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 937d220c31

modelopt/torch/export/layerwise_export.py
```diff
@@ -146,6 +147,73 @@ def assert_layerwise_export_supported(model: nn.Module) -> None:
...
+        key=lambda r: -len(r[0].pattern),
+    )
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 937d220c31

examples/hf_ptq/hf_ptq.py
```diff
@@ -758,7 +759,12 @@ def mono_quantize(
                 language_model, quant_cfg["algorithm"], forward_loop=calibrate_loop
             )
         else:
-            language_model = mtq.quantize(language_model, quant_cfg, forward_loop=calibrate_loop)
+            # A VLM calibrates its language model but must export the whole thing.
+            parent = full_model if args.layerwise_export else language_model
+            with export_parent(parent):
+                language_model = mtq.quantize(
+                    language_model, quant_cfg, forward_loop=calibrate_loop
+                )
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 937d220c31

modelopt/torch/quantization/plugins/huggingface.py
```diff
@@ -1800,13 +1800,18 @@ def get_homogeneous_hf_decoder_layers(model: nn.Module) -> nn.ModuleList | None:
...
+    decoder, seen = model, set()
+    while id(decoder) not in seen:
+        seen.add(id(decoder))
+        for attr in ("model", "language_model"):
+            inner = getattr(decoder, attr, None)
+            if isinstance(inner, nn.Module):
+                decoder = inner
+                break
+        else:
+            return getattr(decoder, "layers", None)
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 937d220c31

tests/gpu/torch/export/test_layerwise_export.py
```diff
@@ -477,3 +486,54 @@ def test_awq_is_refused(tmp_path, make_cfg, method, match):
...
+    export_hf_checkpoint(baseline_vlm, export_dir=baseline_dir)
+
+    vlm = _build_vlm()
+    language_model = get_language_model_from_vl(vlm)[-1]
+    export_dir = tmp_path / "fused"
+    with export_parent(vlm):
+        mtq.quantize(language_model, _layerwise_cfg(export_dir, tmp_path / "ckpt"), _calib_vlm)
+
+    exported = _load_checkpoint(export_dir)
+    _assert_same_checkpoint(_load_checkpoint(baseline_dir), exported)
```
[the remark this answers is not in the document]
Fridah-nv: fixed in 937d220c31

modelopt_recipes/models/moonshotai/Kimi-K3/ptq/nvfp4_experts-kv_fp8_layerwise_export.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+    A resumed run never recalibrates the layers it skipped, so the exported checkpoint is
+    complete but the in-memory model is not and must not be used for inference.
+quantize:
+  algorithm:
+    method: max
+    layerwise:
+      enable: true
+      # max only updates _amax, so the exported shard stays valid for its layer.
+      calib_mutates_weights: false
```
realAsma, line 47: Do we need to specify this? Is not this default already?

examples/hf_ptq/hf_ptq.py
```diff
@@ -753,12 +754,18 @@ def mono_quantize(
...
-        if calibration_only:
-            language_model = mtq.calibrate(
-                language_model, quant_cfg["algorithm"], forward_loop=calibrate_loop
-            )
-        else:
-            language_model = mtq.quantize(language_model, quant_cfg, forward_loop=calibrate_loop)
+        # A VLM calibrates its language model but must export the whole thing. Both branches
+        # reach the same layerwise machinery, so the link is set over both.
+        parent = full_model if args.layerwise_export else language_model
+        with export_parent(parent):
```
realAsma: This is a lower level detail -> Having that in the hf_ptq examples does not Look very good to me.

realAsma: I think the main culprit is that the layerwise calibration + export fuses export can calibration to mtq.quantize API. moving full export to mtq.quantize API seems to be introducing the complexity.

Should we refactor the way layerwise export works?In mtq.quantize layerwise export simply saves the quantized sharded tensors.

then we proceed to call export_hf_checkpoint as usual, which detects the layerwise export and saves the non-layer wise layers, the configs etc.

Wdyt?

realAsma: RB: Can you read my comments and share your thoughts?

realAsma: > _:robot: Bot comment._

🐝 On it — reading your comments on this thread and the Kimi-K3 recipe one, will follow up with thoughts on the refactor shortly.

realAsma: > _:robot: Bot comment._

I think your diagnosis is right, and the refactor is the better shape. Details on where it holds and where it doesn't:

**`export_parent` is a symptom, and the symptom is `finalize()`.** Walking `layerwise_export.py`, the parent is consulted in exactly two places:

- `finalize()` — the tail pass over `model.named_modules()`, `save_non_weight_artifacts`, `_write_hf_export_config`, `_add_mtp_exclusions`. All of this genuinely needs the whole model.
- `__init__` → `self._layer_names`, i.e. the shard key prefix (`language_model.model.layers.N.`).

That's it. `export_layer()` otherwise only touches the layer module itself. So under your split, everything in the first bullet moves to `export_hf_checkpoint(full_model, ...)` — which `hf_ptq.py:957` already calls with `full_model`. The parent stops being ambient state and becomes an ordinary argument: no contextvar, no identity search in `_resolve_export_parent`, and the VLM-reachability refusal in `assert_layerwise_export_compatible` (`hf_ptq.py:788-796`) goes away because nothing has to locate the LM inside the parent anymore.

**The one piece that can't move: the key prefix.** safetensors keys live in the file header, so re-prefixing after the fact means rewriting every shard — 1.65 TB on K3, a second full pass. The prefix has to be known at `export_layer()` time. But it doesn't have to be a context manager in the example script: it's one string, and it can arrive as config/argument data. That's a much smaller surface than what's here now.

**What else the split buys, beyond the readability point you're making:**

- It collapses the `mtq.calibrate` / `mtq.quantize` duplication flagged above — with `finalize()` outside, neither entry point needs the parent, so the two can't drift.
- MTP weights are loaded post-calibration (`hf_ptq.py:940`) and would just flow into `export_hf_checkpoint`'s `extra_state_dict`, so that `NotImplementedError` drops. The `--cast_mxfp4_to_nvfp4` and spec-dec refusals stay — those rewrite layer weights or write a competing checkpoint, so they're not artifacts of the fusion.
- It unwinds a real layering inversion: `model_calib.py` imports `modelopt.torch.export.layerwise_export`, which imports back into `modelopt.torch.quantization` — currently survivable only via the function-local import at `model_calib.py:2093`.

**The crux the refactor has to solve.** `LayerwiseExporter.__init__` captures `get_quant_config(model)` *before* calibration on purpose (see the comment at `layerwise_export.py:271-273`): `export_layer` converts modules in place as it goes, so by the time the last layer is written the model looks unquantized. If `export_hf_checkpoint` runs after calibration, it cannot re-derive the quant config from the live model — it has to read what the layerwise pass recorded, alongside the list of already-written layer names it needs in order to skip them. That handoff (a manifest in `export_dir`, essentially) is the whole design cost, and I'd guess it's why the two got fused in the first place. It's tractable, but it's not free, and it's worth agreeing on that artifact before anyone writes code.

**Scope.** `LayerwiseExporter.finalize()` is #2136's, not this PR's — so this is really a #2136 refactor. Worth noting that if it lands there, #2218's diff mostly evaporates: the parent link *is* the multimodal mechanism here.

---

Separately, on your `calib_mutates_weights: false` question in the Kimi-K3 recipe — no, that's not the default. `LayerwiseConfig.calib_mutates_weights` defaults to `True` (`config.py:774-775`). `method: max` sets `_mutates_weights = False` (`config.py:909`), but that class var only *permits* the flag; the validator at `config.py:840-849` rejects it for GPTQ/AWQ/SmoothQuant and otherwise leaves the default alone. So the recipe does need the line as written.

That said, the line is arguably a papercut: every amax-only layerwise recipe has to repeat it, and defaulting `calib_mutates_weights` from the algorithm's `_mutates_weights` would make it inferable. Also #2136 territory, not this PR.

Fridah-nv: I agree, this exists because `finalize()` runs in calibration, the streaming export part gets VLM support for free.  I'd like to move the finalize() call out of calibration so there's no special handling of VLM and MTP for layerwise export

modelopt_recipes/models/moonshotai/Kimi-K3/ptq/nvfp4_experts-kv_fp8_layerwise_export.yaml
Fridah-nv: The default is true so this is needed

PR #2224, opened 2026-08-20
docs(eval-skill): add NVFP4 model-card sampling reference

cjluo-nv:
### What does this PR do?

Type of change: documentation

Adds `plugins/modelopt/skills/evaluation/references/nvfp4-modelcard-sampling.md` — the published `temperature` / `top_p` / max generation length for the **2026 NVFP4 checkpoints under [huggingface.co/nvidia](https://huggingface.co/nvidia/models) that disclose them** — and points `SKILL.md` Step 3 and `model-card-research.md` at it.

**Why.** Config generation currently re-derives sampling params per run from one model card, which is slow and silently wrong in two ways:

1. **Quickstart boilerplate reads as an eval setting.** Most NVIDIA NVFP4 cards paste a TensorRT-LLM snippet containing `SamplingParams(temperature=0.8, top_p=0.95)`. That string is byte-identical across `Llama-3.1-8B`, `Llama-3.3-70B`, `Llama-4-Scout`, `Phi-4-reasoning-plus`, `Phi-4-multimodal`, `Qwen2.5-VL-7B` and six `Qwen3-*` repos. It is template text, not what the accuracy table was measured with — but it is the most prominent `temperature=` in the card.
2. **A generic fallback beats an available answer.** When a card is silent, Step 3 falls back to 65536/16384 even where a sibling in the same family publishes an exact value.

**Scope — 25 rows.** All 69 NVFP4 checkpoints in the org were read. A row exists only for a 2026 **target** checkpoint whose card discloses usable settings, so absence means "read the card", not "not yet checked". Excluded by construction: pre-2026 releases, cards that publish nothing, and — via the curation regex `-NVFP4(-V\d+|-QAD)?$` — `-DSpark`/`-DFlash` speculative-decoding variants (verified against the target, so identical accuracy; they would only duplicate the base row), `-Eagle3` draft heads, and `-MLPerf-Inference-Closed-*` snapshots.

| provenance | rows | meaning |
| --- | --- | --- |
| `eval` | 20 | card ties the values to its accuracy table — authoritative |
| `rec` | 5 | recommended inference sampling, not tied to the eval |

A `—` in one value column means that field specifically is unpublished — four rows give `temperature`/`top_p` but state no generation cap (`Mistral-Medium-3.5`, `Nemotron-3.5-Lightning`, `Nemotron-3-Super-120B`, `Nemotron-Labs-3-Elastic-30B`). Per-task exceptions are recorded where cards state them: GLM-5.2 GPQA Diamond `100000` vs `64000`; Qwen3.5-397B-V2 τ²-Bench Telecom `128000`; Qwen3.6 SciCode `temperature=0.6`; Kimi-K3 uncapped for Terminal-Bench.

**Posture: the card is the source of truth; the table is a reference, not a constraint.** It is there to confirm a value you read, fill a gap when the card is silent, and catch a misreading — never to override what a card states. Listed and in agreement → proceed; listed and different → the card wins, re-read, surface the discrepancy. The `max_num_tokens` column records the card's *headline* cap, so Step 3's existing take-the-highest rule still governs when a card names more than one. Per-task `temperature`/`top_p` in the notes is precedent rather than mandate — engineers do tune sampling per benchmark — so the guidance is to follow the card and escalate only on a regime change (greedy vs sampled), not a nudge (`0.95` vs `1.0`).

### Usage

Not an API change; the reference is consumed by the `evaluation` skill when generating a NEL config.

```yaml
# nvidia/GLM-5.2-NVFP4 -> references/nvfp4-modelcard-sampling.md (provenance: eval)
nemo_evaluator_config:
  config:
    params:
      max_new_tokens: 100000  # card: GPQA Diamond 100000, others 64000 -> take the highest
      temperature: 1.0
      top_p: 0.95
```

### Testing

**Coverage — enumeration cross-verified two ways.** Rather than paging `https://huggingface.co/nvidia/models?p=N` by hand, the candidate list came from the HF API (`author=nvidia&limit=1000` → 917 repos, one page), then verified against the website pagination: 918 unique model links across `p=0..31`, and the 68 NVFP4-named repos were **identical in both** (`api-only: []`, `web-only: []`). The committed curation regex was then re-run against that full set and confirmed to select every one of the 25 table rows plus 7 further 2026 target checkpoints whose cards disclose nothing — i.e. the documented filter reproduces the table.

**Accuracy — every `eval` row diffed against its source sentence.** A script re-parsed the table and printed each row beside the matching card line. To be precise about what that covers: the diff was run when the table was larger, and all 28 machine-checkable *Benchmarked with* rows matched exactly (`Qwen3.6-35B-A3B` states its settings on the second line of a blockquote and was checked by hand). Every edit since has been a row **removal** — the scope cut to 2026 and the spec-decode removal — each verified by re-parsing the table before and after and confirming no surviving row's `temperature`/`top_p`/`max_num_tokens`/`provenance` changed. Extraction was restricted to *"Benchmarked with…"* / *"…were evaluated with…"* / *"We evaluate the model using…"* sentences, "Recommended Sampling" rows, and accuracy-table footnotes — never the quickstart snippets.

**Two extraction traps found while building it**, both now in the file's refresh recipe as a second mandatory grep: some cards publish the cap **only** as a footnote beneath the accuracy table (`*Max OSL for evals can be as high as 64K`), and DeepSeek states sampling in its `## Input:` usage block rather than a *Benchmarked with* sentence. Neither is reachable from the obvious grep.

**Hygiene.** `pre-commit run --files <the 3 files>` passes, including `markdownlint-cli2` and the `sync .claude/skills/ symlinks` hook; the reference resolves through both `.claude/skills/evaluation/references/` and `.agents/skills/evaluation/references/`.

### Additional Information

Data collected 2026-08-20; the file is an explicitly dated snapshot and says to trust the card for anything newer. It ends with a refresh recipe (API enumeration → `-NVFP4(-V\d+|-QAD)?$` filter → `createdAt >= 2026-01-01` → card fetch → the grep phrasings that carry eval settings) so the table can be regenerated as new checkpoints ship — roughly four NVFP4 target checkpoints per month over 2026 so far.

Review findings addressed: the mandatory-match framing was softened to reference-only, per-task sampling reframed as precedent with a human-escalation trigger, `mkdir -p cards` added to the recipe, the HF token no longer interpolated into a `curl` argument, and explicit uncapped generation (`max_new_tokens: null`) distinguished from an unpublished cap. The provenance question on the Nemotron-3.5-Lightning rows is answered in a review reply — the differing labels were correct (the base card says "Recommended Sampling", the sibling cards say "Benchmarked with"), and those sibling rows have since been removed as spec-decode duplicates.

### DIFF
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -274,13 +274,16 @@ nemo_evaluator_config:
 
 Per-task `max_new_tokens` overrides are forbidden — set one top-level ceiling everywhere.
 
+**Cross-check `temperature` / `top_p` / `max_new_tokens` against `references/nvfp4-modelcard-sampling.md`** — the published settings for the 2026 NVFP4 checkpoints under `huggingface.co/nvidia` that disclose them (older releases and cards that publish nothing are absent — for those, read the card; `-DSpark` / `-DFlash` spec-decode variants share their base checkpoint's row, since spec decoding does not change the target's output distribution). **The card is the source of truth; this file is a reference, not a constraint** — use it to confirm a value you read, to fill a gap when the card is silent or ambiguous, and to catch a misreading. Worth consulting whenever the model is an NVFP4 checkpoint **or shares a family with one** (Qwen3.x, GLM-4.7/5.x, Kimi K2.x/K3, MiniMax M2.x/M3, DeepSeek V3.x/V4/R1, Gemma 4, Nemotron 3/3.5, Llama-Nemotron, Mistral Medium 3.5), and especially when you are unsure. It is a dated snapshot, so for anything newer than it, trust the card. See that file's "Lookup" section.
+
 #### `max_new_tokens` — mandatory model-card lookup
 
 1. **Fetch the HF model card before writing the value.** Not optional.
 2. Scan for any `max_tokens` / `max_new_tokens` / "output length" recommendation. Pick the **highest** value the card mentions (Qwen3.6: 32768 general + 81920 math-coding → use **81920**). Annotate with a citing comment.
-3. If the card is genuinely silent after a thorough read, fall back to: **65536** (reasoning), **16384** (non-reasoning); surface the silence to the user.
-4. **Forbidden:** writing `max_new_tokens: <generic_default>` with a "card not yet checked" comment. Either fetch and apply, or fetch and confirm silence.
-5. **A higher cap doesn't fix runaway reasoning.** On hard tasks (e.g. HLE) a non-terminating model just rambles to the larger cap (~80% length-capped at 131072), and the cap only helps if deployment `--max-model-len > prompt + max_new_tokens` (else generation is silently clipped — AA-LCR's ~120K input leaves little room). Treat such tasks as low-confidence.
+3. **Consult `references/nvfp4-modelcard-sampling.md` as a reference.** Listed and in agreement → proceed with confidence. Listed and different → **the card wins**; re-read it, then note the discrepancy for the user rather than auto-correcting either way. Not listed, or the card is silent or ambiguous → take the nearest same-family rows as the value, a far better prior than the generic fallback below. Its `max_num_tokens` column records the card's *headline* cap, so rule 2 above still governs: when a card names more than one cap, the highest wins even if that exceeds the row.
+4. If the card is genuinely silent after a thorough read **and** the family table offers no usable pattern, fall back to: **65536** (reasoning), **16384** (non-reasoning); surface the silence to the user.
+5. **Forbidden:** writing `max_new_tokens: <generic_default>` with a "card not yet checked" comment. Either fetch and apply, or fetch and confirm silence.
+6. **A higher cap doesn't fix runaway reasoning.** On hard tasks (e.g. HLE) a non-terminating model just rambles to the larger cap (~80% length-capped at 131072), and the cap only helps if deployment `--max-model-len > prompt + max_new_tokens` (else generation is silently clipped — AA-LCR's ~120K input leaves little room). Treat such tasks as low-confidence.
 
 #### Quantization-aware benchmark defaults
 
--- a/plugins/modelopt/skills/evaluation/references/model-card-research.md
+++ b/plugins/modelopt/skills/evaluation/references/model-card-research.md
@@ -3,6 +3,17 @@
 Use WebSearch to find the model card (HuggingFace, build.nvidia.com). Read it carefully, the FULL text, the devil is in the details. Extract ALL relevant configurations:
 
 - Sampling params (`temperature`, `top_p`)
+  - **Only trust a sentence that ties the values to the benchmarks** ("Benchmarked
+    with…", "…were evaluated with…", "We evaluate the model using…") or a
+    "Recommended Sampling" row. The `SamplingParams(temperature=0.8, top_p=0.95)`
+    in a card's TensorRT-LLM/vLLM quickstart snippet is boilerplate copied
+    verbatim across unrelated models — **never** read eval settings out of it.
+  - **Then cross-check `nvfp4-modelcard-sampling.md`** — published
+    `temperature` / `top_p` / `max_num_tokens` for the 2026 NVFP4 checkpoints
+    under `huggingface.co/nvidia` that disclose them, grouped by family.
+    Required for any NVFP4 checkpoint or same-family sibling; it is also the
+    best source of a default when the card is silent. Pre-2026 releases are out
+    of scope there — read their cards.
 - Context length (`deployment.extra_args: "--max-model-len <value>"`)
 - **Output length (`max_new_tokens`) — mandatory extraction.** Scan the
   card for any `max_tokens` / `max_new_tokens` / "output length"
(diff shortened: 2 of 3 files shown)


meenchen (commented):

Docs-only change (239 lines) adding `references/nvfp4-modelcard-sampling.md` to the evaluation skill plus two cross-check hooks in `SKILL.md` Step 3 and `model-card-research.md`. It fits the existing `references/*.md` pattern (no new subsystem, no licensing impact, no injection attempts found), and the "quickstart `SamplingParams` is boilerplate, don't read eval settings out of it" warning is a genuinely useful trap to document. But the entire value of the file rests on 28 rows of externally-scraped model-card data that cannot be verified from the repo, and there are a few internal inconsistencies a maintainer should resolve before merge.

Points for the owner:

- **Table headline cap vs. the "take the maximum" rule.** For cards with a per-task exception the `max_num_tokens` column carries the *lower* value with the exception in `notes` (GLM-5.2: `64000` + note "GPQA Diamond used 100000"; Qwen3.5-397B-V2: `64000` + note "τ²-Bench Telecom used 128000"; Kimi-K3: `65536` + note "uncapped for Terminal-Bench"). But SKILL.md Step 3.3 now says "your value must match its row", while the file's own cross-family observation and the PR body's example say take the max (`max_new_tokens: 100000` for GLM-5.2). An agent following Step 3.3 literally writes 64000. Suggest either putting the max in the column (exception in notes) or spelling out the resolution rule in the row.
- **Provenance labels look inconsistent with the file's own definitions.** `eval` is defined as "card states these were the benchmark/eval settings", yet `nvidia/diffusiongemma-26B-A4B-it-NVFP4` is `eval` with `upstream default` / `—` values (the card just defers upstream — that's not an eval statement), and the two Nemotron-3.5-Lightning `DFlash`/`DSpark` rows are `eval` while the base Lightning row with identical values is `rec`. Worth a pass to confirm each label against its source sentence.
- **Verification counts in the PR body don't line up with the committed table.** The Testing section says "all 28 machine-checkable *Benchmarked with* rows match exactly (the 29th, Qwen3.6-35B-A3B, checked by hand)", but the committed table has 28 rows total of which only 23 are `eval`. Likewise the `—` discussion names Mistral-Medium-3.5 plus "both Nemotron-3.5-Lightning speculative-decoding rows" while six rows actually carry `—`. Reads like the diff was verified against a pre-scope-cut table, so the accuracy claim may not correspond to what's committed.
- **Refresh recipe won't run as written.** Step 3 does `curl ... -o "cards/${id//\//_}.md"` without ever creating `cards/`, so curl fails with "Failed to create the file" on a clean checkout — add `mkdir -p cards`. (Minor: `m['id'].lower().count('nvfp4')` reads oddly as a boolean; `'nvfp4' in m['id'].lower()` is clearer.)
- **Staleness.** A dated snapshot (2026-08-20) that SKILL.md marks as a *mandatory* cross-check will drift as new checkpoints ship. The "card wins on conflict / a miss means read the card" framing mitigates this, but there's no test or CI check that the table is still current — worth a maintainer's judgement on whether the mandatory framing is right for a hand-refreshed file.

meenchen (commented):

Re-review of the docs-only NVFP4 sampling reference (3 files, +169/-4). Most of the prior round's findings are resolved in-tree: SKILL.md Step 3 no longer says "must match its row" (it now defers to the card and keeps the take-the-highest rule, with the `max_num_tokens` column explicitly labelled the *headline* cap); the refresh recipe gained `mkdir -p cards`, dropped the token interpolation, and replaced the odd `.count('nvfp4')` check with the documented `KEEP` regex; the Lookup section now resolves precedence per field for `eval` / `rec` / `—`; the DeepSeek prior is carved out of the house default; explicit uncapped (`null`) is preserved as terminal; and the counts line up — I counted 25 table rows, 20 `eval` / 5 `rec`, matching the header. No prompt-injection content found in the diff (the CodeRabbit "prompt for AI agents" blocks are boilerplate, not directives to me). Remaining items are judgement calls for the owner rather than clear bugs, so nudging rather than approving.

Points for the owner:

- :speech_balloon: Provenance labels — the Nemotron-3.5-Lightning half is resolved (the spec-decode sibling rows are gone and the author explained in a review reply that the base card says "Recommended Sampling", matching the `rec` label). Still flagging because `nvidia/diffusiongemma-26B-A4B-it-NVFP4` is labelled `eval` while its `temp`/`top_p` cells read `upstream` and the row's own note says it *defers* to `google/diffusiongemma-26B-A4B-it` — that isn't "card ties the values to its accuracy table" per the file's own definition of `eval`, and it is one of the 20 rows the header counts as `eval`.
- Non-scalar cells vs. the Lookup procedure: `Ising-Calibration-1.5-31B` (`8192 zero-shot / 32767 ICL`) and `Nemotron-3-Nano-Omni` (`0.6 think / 0.2 instruct`, `0.95 think / —`, `20480 think / 1024 instruct`) require picking a mode before a value exists, but step 1 only covers `eval`/`rec`/`—` and the "Per-task sampling is precedent" paragraph only covers benchmark-specific temp/top_p. Relatedly, Priors says "when a card lists two caps take the maximum", which has no numeric answer for Kimi-K3's "uncapped for Terminal-Bench". Worth one sentence on how an agent collapses these to the single scalar NEL wants.
- :speech_balloon: Verification counts — the PR body now explains that the row-by-row diff predates the scope cut and that every later edit was a removal, and the table/header totals are internally consistent. Still worth a maintainer eye: the file claims "All 69 NVFP4 checkpoints in the org were read" while the body's enumeration check reports 68 NVFP4-named repos (917 API vs 918 web repos). Since completeness is the file's entire warrant, the two numbers should agree.
- Staleness: a hand-refreshed, dated (2026-08-20) snapshot with no CI freshness check. The "card wins / a miss means read the card" framing plus the softened SKILL.md wording mitigates this well, but a maintainer should confirm they're happy owning the refresh cadence (~4 new target checkpoints/month per the body).
- SKILL.md's family list names DeepSeek V3.x/R1 and Llama-Nemotron, but the table has neither (one DeepSeek-V4-Flash row only), so "take the nearest same-family rows" resolves to nothing for those — either trim the list or note the gap.
- Unavoidable caveat: all 25 rows are externally scraped model-card data that cannot be verified from the repo, so approval rests on the author's manual verification.

chadvoegele (approved):

PR #2231, opened 2026-08-23
Fix distributed AutoQuantize scoring and share backward setup

joshua-hill:
### What does this PR do?

Type of change: Bug fix

AutoQuantize can measure a group of quantized expert layers at their enclosing MLP output. That enclosing module is often a plain PyTorch container and does not carry distributed-group information, so its sensitivity score was not combined across data- or expert-parallel workers.

This PR obtains the distributed groups from the quantized layers when the scoring module does not provide them. It also preserves construction order for quantized modules, scoring modules, and their registered hyperparameters so every worker accumulates scores in the same order.

The temporary state needed by backward-based scoring is now managed by one shared session. The session installs and removes forward patches and invocation-specific output-gradient hooks, controls parameter gradients, and restores the active quantization recipes even when scoring raises an exception. Scoring methods remain responsible for their own score calculation.

### Usage

N/A — this fixes existing AutoQuantize behavior and does not add an API or flag.

### Testing

- `pre-commit run --files modelopt/torch/quantization/algorithms.py tests/unit/torch/quantization/test_autoquant.py`
- `pytest -q tests/unit/torch/quantization/test_autoquant.py` — 102 passed
- Added a real two-rank gradient AutoQuantize test covering MoE experts scored at an enclosing MLP.
- Added regressions for deterministic hyperparameter registration, per-invocation replay for reused score modules, exact `forward`-attribute restoration, partial setup rollback, and cleanup after a scoring failure.

- Is this change backward compatible?: ✅ — no API or checkpoint format changes; distributed sensitivity values now include the missing reduction.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — no new feature, deprecation, breaking change, or critical release-note item.
- Did you get Claude approval on this PR?: ❌ — pending review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quantization scoring consistency through deterministic ordering and invocation handling.
  * Added more reliable distributed score aggregation, including support for mixture-of-experts models.
  * Improved gradient-based scoring for repeated evaluations, tuple outputs, and checkpoint-compatible workflows.
  * Ensured model behavior and scoring state are restored after successful or failed evaluations.
  * Avoided unnecessary output replay when gradients are not required.


joshua-hill:
@coderabbitai review

tests/unit/torch/quantization/test_autoquant.py
```diff
@@ -905,6 +947,209 @@ def test_data_parallel_auto_quantize(skip_on_windows):
...
+    hparam = TestHparam()
+    hparam._importance_dict = {recipe: {score_module: None} for recipe in hparam.choices}
+    score_module._hparams_for_scoring = [hparam]
+    inputs = torch.tensor([[1.0, 2.0]], requires_grad=True)
+
+    with _AutoQuantizeGradientScoringSession(model, [score_module], lambda *_: True):
+        model(inputs).backward()
+
+    # First use: sum(x**2) = 5. Second use: sum((2 * 3x)**2) = 180.
+    assert hparam._importance_dict[quant_recipe][score_module].item() == pytest.approx(185.0)
```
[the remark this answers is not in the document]
joshua-hill: Addressed in 14b2af2. The regression now uses torch.testing.assert_close with an expected tensor on the same device, avoiding Python scalar extraction.

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

joshua-hill:
@coderabbitai review

modelopt/torch/quantization/algorithms.py
```diff
@@ -1438,7 +1448,206 @@ def _add_auto_quantize_score(grad_output, output_diff, score_tensor):
...
+        self._grad_accumulators.clear()
+
+    def original_forward(self, module: nn.Module) -> Callable:
+        """Return the forward method saved before scoring."""
+        return self._original_forwards[module]
+
+    def _register_output_grad_hook(self, output: torch.Tensor, hook: Callable) -> None:
+        """Attach an invocation-specific output-gradient hook for this session."""
+        # The hook's lifetime is already bounded by the output's autograd graph.
+        output.register_hook(hook)
```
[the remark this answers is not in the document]
joshua-hill: Addressed in d5a62bd. Output-hook handles are tracked while pending, remove themselves after their scoring backward, and are cleared after every scoring step and again at session exit. This prevents both callback growth across normal scoring and delayed score mutation after exit. The reused-module regression now retains a loss beyond the session, runs backward afterward, and verifies that no score is recorded; PR A’s full 102-test suite passes.

joshua-hill:
@coderabbitai review

modelopt/torch/quantization/algorithms.py
```diff
@@ -532,6 +533,15 @@ def get_score(self, recipe: QuantRecipe) -> float:
...
             parallel_state = getattr(score_module, "parallel_state", None)
+            if parallel_state is None:
+                parallel_state = next(
+                    (
+                        state
+                        for module in self.quant_modules
+                        if (state := getattr(module, "parallel_state", None)) is not None
+                    ),
+                    None,
+                )
```
realAsma: I dont feel comfortable with this. We should not create parallel state from heuristics. 

If the score module has no parallel_state it is likely because it has no DynamicModule equivalent created yet. We should create a dynamic module for it and clearly define parallel_State (see how we create attention dynamic module for an example of how to create dynamic modules on the fly: 

https://github.com/NVIDIA/Model-Optimizer/blob/029c67f27e67088fb19ac0a9af241dc2bc740650/modelopt/torch/quantization/plugins/huggingface.py#L412

https://github.com/NVIDIA/Model-Optimizer/blob/029c67f27e67088fb19ac0a9af241dc2bc740650/modelopt/torch/quantization/plugins/huggingface.py#L78

realAsma: RB: Wdyt?

realAsma:
This PR is going two things -> 
1. AutoQuantize can measure a group of quantized expert layers at their enclosing MLP output. That enclosing module is often a plain PyTorch container and does not carry distributed-group information, so its sensitivity score was not combined across data- or expert-parallel workers.

2. A heavy refactor to support various Activation backward support. 

Can we further split this PR into two so that each PR do a  single thing? This will make PR review easier.

In addition, could you please explain the motivation for the refactor for gradient searcher ? 

Could you post an explainer of how the new method (Auman-Shapley) works and why you need to refactor the gradient searcher to support Auman-Shapley. In addition, can you please give a sweep of effective bits Vs MMLU like as show in https://nvidia.github.io/Model-Optimizer/announcements/autoquantize.html#results

modelopt/torch/quantization/algorithms.py
```diff
@@ -532,6 +533,15 @@ def get_score(self, recipe: QuantRecipe) -> float:
                 continue
 
             parallel_state = getattr(score_module, "parallel_state", None)
+            if parallel_state is None:
+                parallel_state = next(
```
realAsma: This is a good workaround for now. 

We should ideally define parallel_state for the relevant score_module classes instead of relying on the heuristics. But this can be done in future. 

Hence, can we:
1. Add a brief TODO note here
2. Raise a warning if torch.dist is enabled: score module without parallel_state defined found, inferring parallel state.

realAsma: RB: Can you give suggestions for both AIs above?

🐝 Acknowledged — checking both suggestions against the current distributed-state conventions.

For the two requested changes, I’d suggest:

1. TODO: make the assumption explicit: `# TODO: Prefer parallel_state owned by the score module; this temporary fallback inherits the first quantized child’s state.` This documents both the intended end state and that “first” is only safe while the grouped quant modules share one topology.

2. Warning: gate it on `torch.distributed.is_initialized()` *and* only after a child `parallel_state` was actually found. Emit once per score module/hparam (not inside every recipe/score iteration), naming the score-module type and that its first quantized child supplied the state. Reuse the existing `QuantModule._initialize_parallel_state()` wording/semantics where possible.

I’d also add a focused distributed regression asserting the fallback warning is emitted once and the reduced score is unchanged; that protects the diagnostic without making normal multi-recipe AutoQuantize noisy.

realAsma: @joshua-hill Could you please address my comment above?

realAsma:
Great PR!! Clean implementation :)

modelopt/torch/quantization/algorithms.py
joshua-hill: Addressed in [8f65232](https://github.com/joshua-hill/Model-Optimizer/commit/8f65232ed90a0309e044d4285523dd445a2a48e3). Added the TODO and a warning gated on torch.distributed.is_initialized() after successfully finding a child parallel_state. It emits once per score-module/hparam and names both module types. I also extended the two-rank regression to verify the warning count and unchanged reduced score.

realAsma (approved):
Looks great!!

rohansjoshi (approved):

PR #2232, opened 2026-08-24
fix(launcher): use the schema's draft_model global var

h-guo18:
### What does this PR do?

**Type of change:** Bug fix

Fixes the launcher example added in #2149, which could not run at all.

`pipeline.global_vars` is a fixed-field dataclass ([`GlobalVariables`](https://github.com/NVIDIA/Model-Optimizer/blob/main/tools/launcher/core.py)), not a free-form mapping, so an unknown key is rejected at launch:

```
Error processing argument 'pipeline.global_vars.drafter=/hf-local/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark':
Invalid argument: No parameter named 'drafter' exists for <function launch at 0x...>
```

The Nemotron-3.5 DSpark warm-start example invented `drafter:`. `draft_model` is the field that already exists for exactly this purpose, so the fix is to use it.

### Changes

| File | Change |
| --- | --- |
| `examples/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16/hf_streaming_dspark_warmstart.yaml` | `drafter:` → `draft_model:` (the reported bug) |
| `examples/moonshotai/Kimi-K2.5/specdec_bench.yaml` | `draft_model_dir:` → `draft_model:` — the same latent break, pre-existing. Only the global-var key is renamed; the script's `--draft_model_dir` flag is unchanged. |

### Testing

Not end-to-end verified against a live cluster — the reported error is a launch-time argument-parsing failure that occurs before any job is submitted, and it is gone. Later steps in the Nemotron YAML still need the environment-specific placeholders (`<vllm-image-with-nemotron_h-eagle3>`, chat template path) filled in.

### Additional Information

Follow-up to #2149.

An earlier revision of this PR also added a `check_launcher_yaml.py` pre-commit rule that rejected unknown `global_vars` keys, since this class of bug has now shipped twice (the comment on `GlobalVariables.draft_model` records the first, OMNIML-5024). That has been dropped to keep this PR to the bug fix; the check can be proposed separately if wanted.

### DIFF
--- a/tools/launcher/examples/moonshotai/Kimi-K2.5/specdec_bench.yaml
+++ b/tools/launcher/examples/moonshotai/Kimi-K2.5/specdec_bench.yaml
@@ -32,13 +32,13 @@ pipeline:
 
   global_vars:
     hf_model: /hf-local/nvidia/Kimi-K2.5-NVFP4
-    # Trained+exported DFLASH draft; override: pipeline.global_vars.draft_model_dir=<path>
-    draft_model_dir: /hf-local/nvidia/Kimi-K2.5-DFlash
+    # Trained+exported DFLASH draft; override: pipeline.global_vars.draft_model=<path>
+    draft_model: /hf-local/nvidia/Kimi-K2.5-DFlash
 
   task_0:
     script: common/specdec_bench/run.sh
     args:
-      - --draft_model_dir <<global_vars.draft_model_dir>>
+      - --draft_model_dir <<global_vars.draft_model>>
       - --speculative_algorithm DFLASH
       - --engine VLLM
       - --mtbench /hf-local/HuggingFaceH4/mt_bench_prompts/raw/question.jsonl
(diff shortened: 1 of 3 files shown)


kevalmorabia97 (approved):

h-guo18:
Closing as obsolete — this fix already landed on `main`.

PR #2027 (commit 5db2682519, merged 2026-08-26) made the identical rename in both files:

```diff
 tools/launcher/examples/moonshotai/Kimi-K2.5/specdec_bench.yaml
-    draft_model_dir: /hf-local/nvidia/Kimi-K2.5-DFlash
+    draft_model: /hf-local/nvidia/Kimi-K2.5-DFlash
-      - --draft_model_dir <<global_vars.draft_model_dir>>
+      - --draft_model_dir <<global_vars.draft_model>>

 tools/launcher/examples/nvidia/.../hf_streaming_dspark_warmstart.yaml
-    drafter: /hf-local/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark
+    draft_model: /hf-local/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16-DSpark
-      - dflash.dflash_init_checkpoint=<<global_vars.drafter>>
+      - dflash.dflash_init_checkpoint=<<global_vars.draft_model>>
```

Rebasing this branch onto current `main` would therefore produce an empty diff, so there is nothing left to merge. The `drafter:` launch failure reported here is fixed on `main` today.

For the record, the failing `codecov/project` check on this PR was a false positive, not a reason for closing: the branch's merge-base was 60 commits behind `main`, so Codecov compared a head report built from current `main` (+8913 tracked lines) against a base report from the stale merge-base. Both flags (`unit` +1.06%, `regression` +0.07%) went up, and all modified lines were reported as covered.

An earlier revision of this PR also added a `check_launcher_yaml.py` pre-commit rule rejecting `global_vars` keys that are not fields of `GlobalVariables`, since this class of bug has now shipped twice (OMNIML-5024, then #2149). That is the one piece not covered by #2027; I can open it as a standalone PR if it is wanted.

PR #2240, opened 2026-08-24
[ONNX] Add quantization sensitivity ranking + exclusion picker

gcunhase:
### What does this PR do?

Type of change: new feature (ONNX PTQ tooling)

Adds `modelopt.onnx.quantization.sensitivity` — a first-class ONNX PTQ primitive that ranks each quantizable target (op type or individual node) by its impact on model output, plus a downstream picker that turns the ranking into an actionable `--nodes_to_exclude` / `--op_types_to_exclude` list. Closes the manual-investigation gap for accuracy degradation in ONNX models with QDQ nodes.

- **`sensitivity.score`** — for every target, inserts calibrated Q/DQ on that target via the standard `modelopt.onnx.quantization.quantize` pipeline, runs both the reference and the quantized graphs through ORT, and computes a proxy metric (`kl_div` default, `mse`, or `cos`) between their outputs. Higher = more accuracy loss if quantized. `granularity={op_type, node}`, `target_precision={int8, fp8}`, real or synthetic calibration.
- **`sensitivity.suggest_exclusion`** — coverage mode (walks targets in descending score order and accumulates until the next one would push cumulative sensitivity above `coverage * total_mass`; architecture-portable) or threshold mode (absolute cutoff), plus an optional `blocks=` / `block_agg=` argument for block-level aggregation where per-node picking would fragment precision within a block.
- **`sensitivity.summarize_exclusion`** — one-call summary of what an exclusion set covers.
- **CLI**: `python -m modelopt.onnx.quantization.sensitivity` renders a ranked table to stderr and writes a JSON side-file.

Small supporting additions to `modelopt/onnx/utils.py` — `get_op_types_in_graph` (enumerates the unique op types actually present in an ONNX graph) and `validate_file_size` (relocated from `modelopt/onnx/quantization/__main__.py` so the quantize and sensitivity CLIs share a single implementation).

### Usage

Python code:
```python
from modelopt.onnx.quantization import quantize
from modelopt.onnx.quantization.sensitivity import score, suggest_exclusion, summarize_exclusion

# 1. Rank the quantizable targets in the graph.
result = score(
    onnx_path="coatnet-0.onnx",
    calibration_data="imagenet_calib_500.npz",
    granularity="op_type",     # or "node"
    metric="kl_div",           # or "mse", "cos"
    target_precision="int8",
)
# result["scores"] is a dict {op_type_or_node_name: metric_value}, higher = more sensitive.

# 2. Turn the ranking into an exclusion list.
excluded = suggest_exclusion(result["scores"], coverage=0.90)
print(summarize_exclusion(result["scores"], excluded))

# 3. Quantize with the exclusion applied.
quantize(
    onnx_path="coatnet-0.onnx",
    quantize_mode="int8",
    calibration_data="imagenet_calib_500.npz",
    nodes_to_exclude=excluded,
    output_path="coatnet-0.quant.onnx",
)
```

CLI equivalent:

```bash
python -m modelopt.onnx.quantization.sensitivity \
    --onnx_path coatnet-0.onnx \
    --calibration_data_path imagenet_calib_500.npz \
    --granularity op_type \
    --metric kl_div
```

Block-level aggregation on transformer architectures (ViT-tiny example, picks whole transformer blocks instead of individual nodes):

```python
blocks = {f"blocks.{n}": [rf"^/blocks/blocks\.{n}/"] for n in range(12)}
excluded = suggest_exclusion(
    result["scores"], threshold=0.1, blocks=blocks, block_agg="max",
)
```

### Testing

Score tests are organized as four tiers under `tests/gpu/onnx/quantization/sensitivity/test_score.py`, ordered from lightest to heaviest:

- **Tier 1** (unit, seconds) — `test_synthetic_random_calibration_smoke`: with `calibration_data=None`, the synthetic-random path returns finite non-negative scores for every op in scope. Guards against NaN / negative-score regressions in the fallback path.
- **Tier 2** (unit, seconds, parametrized over `kl_div` / `mse` / `cos`) — `test_synthetic_deterministic_ln_highest`: synthetic 2-Conv + 1-MatMul + 1-LayerNorm graph with deterministic real inputs; `LayerNormalization` scores highest of all ops.
- **Tier 3** (`@pytest.mark.manual`, ~14 min on H100) — `test_coatnet_op_type_matches_manual_groundtruth`: CoAtNet-0 op-type ranking on 500-sample ImageNet calibration. Top-4 = `Add / Mul / LayerNormalization / ReduceMean` (all > 1.5 KL), `Conv` sits ~10× below. Matches the manual "Conv-only wins 82% top-1" ground truth read as a quantization policy.
- **Tier 4** (`@pytest.mark.manual`, ~30–60 min on H100) — `test_coatnet_per_node_matches_manual_groundtruth`: CoAtNet-0 per-node ranking; LN / MHA nodes in top-10, individual Conv nodes in bottom-10.

Two additional GPU unit tests in the same file — `test_failed_probe_is_recorded` and `test_failed_probe_records_exceptions` — cover the `failed` return field on missing-Q/DQ and exception paths.

Tier 3 / 4 fixtures resolve via `MODELOPT_ONNX_ACCURACY_MODELS_DIR`; missing fixtures `pytest.skip` cleanly. `@pytest.mark.manual` tests are skipped by default in CI (auto-skip conftest hook) and can be run explicitly with `--run-manual`.

CPU-only unit tests:

- `tests/unit/onnx/quantization/sensitivity/test_picker.py` — coverage / threshold / near-tie warning / `max_nodes` / `min_score_floor` / `blocks` semantics, plus a singleton-name-collision regression witness.
- `tests/unit/onnx/quantization/sensitivity/test_metrics.py` — `kl_div`, `mse`, `cos_dist` (including the asymmetric-zero-vector regression witness for `cos_dist`).
- `tests/unit/onnx/quantization/sensitivity/test_cli.py` — 17 tests covering `_default_output_json`, `_validate_calibration_dir` (empty dir / per-file cap / aggregate cap / happy path), `_render_ranked_table` (all footer variants), and `main()` (synthetic happy path, default output-JSON path, oversize-input rejection, calibration-dir validation).
- `tests/unit/onnx/quantization/test_quantize_api.py::test_quantize_honors_nodes_to_quantize_allowlist` — pins the `quantize(..., nodes_to_quantize=[...])` primitive the sensitivity scanner relies on to isolate a single node for a per-target probe.

- Is this change backward compatible?: ✅ Pure additions: new `modelopt.onnx.quantization.sensitivity` package and two helpers in `modelopt/onnx/utils.py` (one relocated, one new). No existing signatures changed.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ No new PIP dependencies; primitive uses existing `numpy` / `onnx` / `onnxruntime` / `modelopt.onnx.quantization.quantize` machinery only.
- Did you write any new necessary tests?: ✅ Four-tier score layout above, plus picker / metrics / CLI unit tests and the `nodes_to_quantize` allowlist guard.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ New Features → Quantization: one-line entry covering the primitive, picker, and optional block-level aggregation.
- Did you get Claude approval on this PR?: N/A — will run `/claude review` after opening.

### Additional Information

**Motivation** — post-training quantization on ONNX hybrid architectures currently has no automated way to identify sensitivity-driving ops. For CoAtNet-0, closing the accuracy gap required manually cycling through six exclusion policies (default QDQ 22%, attention-MatMul excluded 21%, `--op_types_to_exclude MatMul` 36%, `--disable_mha_qdq` 40%, `--disable_mha_qdq --op_types_to_exclude Softmax` 33%, `Conv`-only 82%) before finding the winner. That per-model investigation is O(model × strategy) hours of operator time and produces no reusable artifact. This primitive replaces the manual sweep with a single ranking call.

**Relationship to existing ModelOpt code** — the existing `modelopt/onnx/quantization/autotune/` package is orthogonal: it optimizes TensorRT latency with no accuracy signal, no per-op sensitivity score, and no output-drift metric. `modelopt.torch.quantization.model_quant.auto_quantize` is the design analog we mirror on the ONNX side (same "higher = more sensitive = keep at higher precision" contract; ONNX port differs in autograd-free, ORT-based, graph-mutation forward passes).

**Validation results** — 500-image ImageNet-1k validation across four models on NVIDIA H100 (GH100) and TensorRT 10.16.2.11. Metrics: GPU Compute Time median (ms) for latency and Top-1 for accuracy. `INT8 baseline` is `trtexec --int8 --fp16` auto-selection (implicit quantization); `QDQ-INT8` is ModelOpt's default behavior; and `QDQ-INT8 + sensitivity` is the best sensitivity-driven exclusion recipe per model.

| Model | INT8 baseline (`--int8 --fp16`) | QDQ-INT8 (default) | QDQ-INT8 + sensitivity exclusions | Best recipe |
|---|---|---|---|---|
| **CoAtNet-0** | 80.6% / **1.047 ms** | 22.4% / 1.220 ms | **81.4%** / 1.218 ms | per-node `coverage=0.90` (26/345 excluded) |
| **MobileNetV3-L** | 68.9% / **0.331 ms** | 48.3% / 0.501 ms | **69.0%** / 0.606 ms | per-node `KL > 0.002` (19/139 excluded) |
| **ResNet-50** | 78.9% / **0.278 ms** | 57.9% / 0.276 ms | **79.6%** / 0.294 ms | per-node `KL > 0.002` (9/121 excluded) |
| **ViT-tiny** | **76.0%** / 0.525 ms | 6.7% / 0.357 ms | 75.2% / **0.408 ms** | block-level `blocks.7-11 + /norm/LN` (101/244 excluded) |

**Synthetic-calibration caveat** — `calibration_data=None` is a supported fallback but produces directional rankings only. The CLI prints a warning, the output JSON's `calibration_source` field reads `"synthetic"`, and downstream absolute-threshold consumers should reject synthetic-calibrated scores. Attention-heavy models are the highest-risk case — recommend real calibration when the model has attention.

(Maybe) **Follow-up (separate PR)** — constrained solver on top of this primitive that takes `accuracy_budget_pp` and returns an `ExclusionConfig`, mirroring the torch `auto_quantize` API signature.


modelopt/onnx/quantization/sensitivity/score.py
```diff
@@ -0,0 +1,434 @@
...
+
+def _enumerate_node_targets(
+    onnx_model: onnx.ModelProto, quantizable_ops: set[str]
+) -> list[tuple[str, dict]]:
+    """Return one probe per named quantizable node.
+
+    Args:
+        onnx_model: Loaded model to enumerate.
+        quantizable_ops: Whitelist of op types considered quantizable.
+
```
cjluo-nv, line 404: > _Bot comment._

Node-granularity probes pass only `nodes_to_quantize` and leave `op_types_to_quantize` unset, while op-type probes pass `op_types_to_quantize=[op]`. Those two paths are not equivalent: `quantize()` → `configure_ort()` deletes `Relu`, `Sigmoid`, `Softmax`, `Concat`, `Gather`, `Transpose`, `Reshape`, `Pad`, ... from `QLinearOpsRegistry`/`QDQRegistry` for every op type *not* in `op_types_to_quantize`, and ORT's `should_quantize_node` gates on the resulting allow-list. So a per-node probe on e.g. a `Softmax`, `Relu` or `Sigmoid` node inserts no Q/DQ at all and is recorded as `0.0` — indistinguishable from "quantizing this node is harmless", which is exactly the signal `suggest_exclusion` consumes.

Since the per-node ranking is what the headline CoAtNet/MobileNet/ResNet results are built on, this looks like a real correctness gap. Suggest also passing `op_types_to_quantize=[node.op_type]` alongside the node regex so both granularities probe the same registry, and adding a unit test that a per-node probe on an activation op (e.g. the `Sigmoid`/`Relu` in a small synthetic graph) produces a non-zero score.

modelopt/onnx/quantization/sensitivity/score.py
```diff
@@ -0,0 +1,434 @@
...
+        f"Sensitivity scan on {onnx_path}: {calibration_source.value} calibration, "
+        f"{num_samples} samples, granularity={granularity}, metric={metric}, "
+        f"target_precision={target_precision}"
+    )
+
+    quantizable_ops = set(op_types_scope) if op_types_scope else _default_op_types_scope(onnx_model)
+    if granularity == Granularity.OP_TYPE.value:
+        targets = _enumerate_op_type_targets(onnx_model, quantizable_ops)
+    else:
+        targets = _enumerate_node_targets(onnx_model, quantizable_ops)
```
cjluo-nv, line 217: > _Bot comment._

Two related robustness points around the probe loop:

1. `except Exception: ... continue` drops the target from `scores` entirely, so a target that fails to quantize is silently absent from both the ranking and `summarize_exclusion`'s `num_previously_quantized`. Consider recording it explicitly (e.g. a `failed` list in the returned dict) so downstream consumers can tell "failed" from "not probed".
2. There's no verification that the probe actually inserted Q/DQ. As the docstring notes, ops that slip past the filter but can't be quantized "are reported with score 0.0" — but `0.0` is semantically "quantizing this is free", which is the opposite of "unknown". Cheap fix: after `quantize()`, count `QuantizeLinear` nodes in `probe_path` and, if zero, log a warning and record the target as unprobed rather than `0.0`.

modelopt/onnx/op_types.py
```diff
@@ -407,4 +407,5 @@ def get_activation_ops():
         "Softsign",
         "Swish",
         "HardSwish",
+        "Gelu",
```
cjluo-nv: > _Bot comment._

This is not a pure addition. `get_activation_ops()` is consumed by `QDQAutotunerBase.get_ort_quantization_config()` (`modelopt/onnx/quantization/autotune/autotuner_base.py`) to build `op_types_needing_output_quant`, so adding `Gelu` changes autotune's Q/DQ placement for any model containing `Gelu`. That's a behavior change to a shipped feature with no test and no mention in the PR body (the body instead describes a `TopK` addition to `is_fusible_reduction_op`, which is already on `main`).

Please either (a) call this out in the PR body/CHANGELOG and add an autotune-side test asserting the new `op_types_needing_output_quant` behavior, or (b) keep the sensitivity scope local — e.g. add `Gelu` to `_default_op_types_scope` in `score.py` instead of mutating the shared helper.

tests/unit/onnx/quantization/test_sensitivity_picker.py
```diff
@@ -0,0 +1,218 @@
...
+        summary = summarize_exclusion({}, [])
+        assert summary["coverage_pct"] == 0.0
+        assert summary["num_excluded"] == 0
+
+    def test_missing_node_names_default_zero(self):
+        scores = {"a": 5.0, "b": 5.0}
+        summary = summarize_exclusion(scores, ["a", "unknown"])
+        assert summary["excluded_mass"] == pytest.approx(5.0)
+        assert summary["coverage_pct"] == pytest.approx(50.0)
+        assert summary["num_excluded"] == 2
```
cjluo-nv: > _Bot comment._

The PR body says these tests cover "`blocks` semantics", but there is no test here exercising `blocks=` / `block_agg=`. That path is the documented best recipe for ViT-tiny (101/244 nodes excluded) and it has non-trivial semantics worth pinning: first-match-wins across `blocks` iteration order, unmatched nodes becoming singleton groups, `sum`/`max`/`mean` aggregation, the union-of-members return value (and its ordering), and the `ValueError` on an invalid `block_agg`. The docs also assert that `threshold=0.1, block_agg="max"` and `coverage=1.0, max_nodes=6, block_agg="sum"` yield the same 101-node set — that equivalence claim is a good test case.

Also missing: direct unit tests for `metrics.py` (identical inputs → `kl_div`/`mse`/`cos` == 0; orthogonal vectors → `cos_dist` == 1; 0-D tensor handling in `_flatten_per_sample`), which is pure numpy and costs milliseconds.

modelopt/onnx/quantization/sensitivity/__main__.py
```diff
@@ -0,0 +1,242 @@
...
+    path-loader.
+
+    Args:
+        path: Filesystem location or ``None`` for the synthetic-random fallback.
+
+    Returns:
+        The value to hand to :func:`score` as ``calibration_data``.
+    """
+    if path is None:
+        return None
```
cjluo-nv: > _Bot comment._

`_load_calibration` partially duplicates `score._load_calibration_from_path` (which already handles `.npz`, `.npy` and directories) — the CLI eagerly loads `.npz` but delegates the other two forms, so there are now two code paths with different behavior for the same flag. Suggest just passing the path straight through to `score()`.

Separately, the existing ONNX PTQ CLI (`modelopt/onnx/quantization/__main__.py`) guards calibration/model loading with `validate_file_size()` and the `--trust_calibration_data` opt-in for pickled data; this new entrypoint inherits neither. `allow_pickle=False` covers the RCE case, but the size guard and the consistent flag surface are worth mirroring.

tests/gpu/onnx/quantization/test_sensitivity.py
```diff
@@ -0,0 +1,267 @@
...
+_FIXTURE_DIR = os.environ.get("MODELOPT_ONNX_ACCURACY_MODELS_DIR", "/tmp")
+# Ops covered by the synthetic Conv+MatMul+LN graph. Passed explicitly because the score()
+# default -- get_autotuner_quantizable_ops() -- excludes LayerNormalization even though the ModelOpt
+# quantize() path registers it via configure_ort.
+_SYNTHETIC_OP_SCOPE = ["Conv", "MatMul", "LayerNormalization"]
+
+
+def _build_conv_mm_ln_onnx(path: str, opset: int = 17) -> None:
+    """Build a small 2-Conv + 1-MatMul + 1-LayerNorm ONNX for deterministic sensitivity tests."""
+    rng = np.random.default_rng(0)
```
cjluo-nv: > _Bot comment._

Two things about placement/comments here:

1. Tiers 1 and 2 run with `calibration_eps=("cpu",)` on a tiny synthetic graph — they don't need a GPU, so putting them under `tests/gpu/` means the only fast regression guards for `score()` won't run in the unit CI job. Suggest moving tiers 1–2 to `tests/unit/onnx/quantization/` and keeping only the `slow`/`slow_gpu` CoAtNet tiers here.
2. This comment says the `score()` default is `get_autotuner_quantizable_ops()` and that it excludes `LayerNormalization`. That's stale: `score()` uses `_default_op_types_scope()`, which includes normalization ops via `is_normalization_op`. Either drop `_SYNTHETIC_OP_SCOPE` and let the default scope apply (which would also make the test exercise the default path), or fix the rationale.

cjluo-nv (commented):

## Summary

New ONNX PTQ subsystem (`modelopt.onnx.quantization.sensitivity`: `score` / `picker` / `metrics` / CLI) plus a 317-line docs section, ~2000 added lines. The core idea is sound and the picker unit tests are good, but there are a few correctness/coverage issues I'd like resolved before this lands.

### Design gate (architectural-change protocol)

**Problem:** ONNX PTQ has no automated way to find which ops/nodes drive accuracy loss, so operators hand-sweep exclusion policies per model.

**Alternatives checked in-repo:**
- `modelopt/onnx/quantization/autotune/` — the PR body explicitly addresses this (latency-driven, no accuracy signal). Reasonable.
- `modelopt.torch.quantization.model_quant.auto_quantize` — body names it as the design analog. Reasonable.

So the top-level "why a new subsystem" question *is* addressed; I'm not blocking on it. Two smaller reuse questions the body does **not** address:
1. `score._resolve_calibration_data` / `_load_calibration_from_path` and `__main__._load_calibration` re-implement calibration ingestion that already exists in `modelopt/onnx/quantization/calib_utils.py` (`CalibrationDataProvider` normalizes ndarray/dict/batch-splitting; `RandomDataProvider`+`gen_random_inputs` covers the synthetic fallback) and in `modelopt/onnx/quantization/__main__.py` (npz/npy loading with the `--trust_calibration_data`/`validate_file_size` security path, which the new CLI does not inherit). This is now a third loader with slightly different semantics.
2. A second `python -m ...` entrypoint vs. a flag on the existing ONNX PTQ CLI — worth a sentence in the body.

### Blocking-ish findings
1. **`op_type` and `node` granularity are not equivalent probes** (see inline on `score.py`): node targets pass only `nodes_to_quantize` and leave `op_types_to_quantize=None`, so ORT's effective allow-list is the post-`configure_ort` registry, which has `Relu`/`Sigmoid`/`Softmax`/`Concat`/`Transpose`/... deleted. Those nodes silently get no Q/DQ → score `0.0` ("safe to quantize"), while the same op scores non-zero in op-type mode. Since the headline results are per-node, this matters.
2. **`op_types.py` change is not a pure addition.** Adding `"Gelu"` to `get_activation_ops()` also changes `QDQAutotunerBase.get_ort_quantization_config()` (`op_types_needing_output_quant`), i.e. autotune Q/DQ placement for any Gelu model. No test, and the PR body describes a different op_types change (`TopK` → `is_fusible_reduction_op`, which is already in `main`).
3. **`blocks` / `block_agg` has zero tests** despite being the documented best recipe for ViT-tiny and the body claiming coverage for it. `metrics.py` and the whole CLI module (`_render_ranked_table`, `_load_calibration`, `_default_output_json`) are also untested.
4. **PR description is out of sync with the diff**: it claims a `nodes_to_quantize` extension to `quantize.py`, but `quantize.py` is not in the diff — that argument already exists on `main`, so `tests/unit/onnx/quantization/test_nodes_to_quantize.py` is a test of pre-existing behavior (fine to add, but please re-word the body so reviewers know what actually changed).
5. **Size**: 2003 lines. The picker + its tests, the scorer + CLI, and the docs section are three fairly independent units and would review much better split.

### Answer to your docs question
I'd put it in a **new dedicated guide page**, mirroring `docs/source/guides/9_autotune.rst`, with only a short pointer paragraph at the end of `_onnx_quantization.rst`. Reasons: (a) 317 lines is already longer than several complete sections of the ONNX PTQ guide and pushes the "how do I quantize an ONNX model" narrative off the page; (b) the content is a workflow with its own CLI, its own concepts (coverage vs threshold, block aggregation, near-tie), and its own troubleshooting — same shape as autotune; (c) a separate page lets you use `.. argparse::` for the new CLI the way `9_autotune.rst` does. The parts that are *reproduction recipes* rather than API docs — the timm export snippet, the ImageNet-500 NPZ preparation, and the per-model validation table — belong in `examples/onnx_ptq/README.md`, since they need real datasets/checkpoints and will drift with model versions.

No prompt-injection content observed in the PR text.

modelopt/onnx/quantization/sensitivity/score.py
gcunhase: This is theoretically valid, but a a non-issue in this workflow since the nodes that would be skipped by this are nodes that are skipped by ModelOpt by default. So including a score for those nodes during sensitivity scoring would not align with ModelOpt's default behavior. If a time comes when that's no longer true, then we can revisit this issue.

modelopt/onnx/op_types.py
gcunhase: This should be okay. This function already existed pre-Autotune, and the reason why Gelu wasn't there was because it didn't exist pre-opset 20, so adding Gelu here improved the coverage and correctness of this function.

ajrasane (commented):
The tests can be simplified without weakening their behavioral coverage.

1. `tests/unit/onnx/quantization/test_nodes_to_quantize.py` has the largest opportunity. The bespoke two-Conv builder and `_has_dq_predecessor` helper can be replaced with the existing `build_conv_act_pool_model()` and `assert_nodes_are_quantized()` test utilities. Keep both sides of the contract: the selected interior Conv receives Q/DQ and the unselected Conv does not. I validated that reuse against this head: selecting either Conv inserted DQ only for the selected node, and the existing test passed independently.

2. `test_sensitivity_picker.py` can use small, descriptively named parameter tables for the pure input/output cases. In particular, the full-coverage and sorted-output cases overlap; the empty/zero cases can share one parameterized test; and the large near-tie fixtures can be reduced to four scores while preserving the same cutoff. `caplog.text` also removes the repeated message-list construction. I would keep separate cases for strict threshold behavior, threshold-overrides-coverage, `max_nodes`, `min_score_floor`, ordering, and warning mode because those exercise distinct branches.

3. `test_sensitivity.py` can build the deterministic synthetic ONNX once through a module-scoped fixture. The parameterized metric cases can then reuse it, remove redundant `scores`/ranking assertions, compare the expected top-op set directly, and use one `k = 10` for the top/bottom integration checks.

One placement caveat: the end-to-end tests should not be moved into the CPU unit lane unchanged. With `CUDA_VISIBLE_DEVICES=""` at `47e235c`, `quantize()` reached `trt.Builder()` during preprocessing; the current `nodes_to_quantize` test failed there, and the four synthetic scorer cases produced no scores. CI currently passes these tests, so this is environment-sensitive. Either keep the real quantize/ORT smoke tests in the GPU lane, or mock the TensorRT preprocessing boundary for a focused CPU unit test while retaining one GPU integration test.

The saved test code would be better spent on the existing uncovered behavior: compact public-API tests for `blocks`/`block_agg` and direct tests for the NumPy metrics. That would make the suite shorter and materially stronger rather than merely reducing its line count.

modelopt/onnx/quantization/sensitivity/score.py
gcunhase: Fixed in 0a125d02. Return dict now carries a `failed: list[str]` field. After `quantize()`, `_count_qdq_nodes(probe_path)` verifies at least one QDQ was inserted — if zero, the target is appended to `failed` with a warning instead of being scored `0.0`. Distinguishes "probed harmlessly" from "we don't know."

tests/unit/onnx/quantization/test_sensitivity_picker.py
gcunhase: Fixed in ee1387bb. Added `TestBlocks` in `tests/unit/onnx/quantization/sensitivity/test_picker.py` (7 methods covering first-match-wins, singleton fallback, sum/max/mean aggregation, `ValueError` on invalid `block_agg`, union-of-members return, and the `threshold=0.1, max` ≡ `coverage=1.0, max_nodes=6, sum` equivalence claim). New `test_metrics.py` covers `kl_div` / `mse` / `cos_dist` identity, orthogonal / anti-parallel cases, scale sensitivity, and `_flatten_per_sample` shape handling.

modelopt/onnx/quantization/sensitivity/__main__.py
gcunhase: Fixed in 0a125d02. Dropped `_load_calibration` — CLI now passes the raw path string to `score()` which already handles `.npy` / `.npz` / directory uniformly via `_load_calibration_from_path` (hardcoded `allow_pickle=False`). Added `validate_file_size()` guards on `--onnx_path` (2 GiB, matches main CLI) and `--calibration_data_path` (4 GiB, ImageNet-scale NPZ). Dropped the no-op `--trust_calibration_data` flag rather than shipping documentation for behavior that wasn't wired.

tests/gpu/onnx/quantization/test_sensitivity.py
gcunhase: Fixed in ee1387bb. Restructured the sensitivity tests into a sub-package mirroring `autotune/`: shared builders in `tests/_test_utils/onnx/quantization/sensitivity/models.py`, unit tests under `tests/unit/onnx/quantization/sensitivity/`, and integration tests under `tests/gpu/onnx/quantization/sensitivity/`. Dropped the stale `_SYNTHETIC_OP_SCOPE` rationale comment. Kept the synthetic scorer tests in the GPU lane per the `CUDA_VISIBLE_DEVICES=""` finding raised by @ajrasane — `trt.Builder()` is reached during preprocessing even with CPU EPs.

ajrasane (approved):

modelopt/onnx/quantization/sensitivity/score.py
```diff
@@ -0,0 +1,466 @@
...
+
+    onnx_model = onnx.load(onnx_path)
+    calib_dict, calibration_source = _resolve_calibration_data(
+        onnx_model, calibration_data, num_synthetic_samples
+    )
+    num_samples = _num_samples(calib_dict)
+    logger.info(
+        f"Sensitivity scan on {onnx_path}: {calibration_source.value} calibration, "
+        f"{num_samples} samples, granularity={granularity}, metric={metric}, "
+        f"target_precision={target_precision}"
```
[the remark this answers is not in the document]
gcunhase: Target names are load-bearing for the primitive's primary purpose (identifying which specific node/op drives accuracy loss during quantization). Scrubbing them at INFO would leave users with `[42/244] scored = 5.71 (highest impact!)` — no clue which layer the score corresponds to during a 30-min live scan.

Precedent: `moq.quantize()` and `configure_ort()` also log node names at INFO. If ModelOpt security policy later rules that ONNX node names count as artifact-sensitive PII, `modelopt/onnx/quantization/` needs a coordinated treatment across all three modules — not a per-PR one-off. Downstream users who need redaction today can filter at the log-handler layer via `logging.Filter` on `"modelopt.onnx.quantization.sensitivity"`.

Keeping the current behavior (scan-start log dropped the `onnx_path`; per-probe warnings use `type(e).__name__` with raw exception text moved to `logger.debug(exc_info=True)`; target names remain in the per-probe INFO/WARN messages).

modelopt/onnx/quantization/sensitivity/__main__.py
```diff
@@ -0,0 +1,271 @@
...
+    if not files:
+        raise FileNotFoundError(f"No .npz files found under calibration directory: {path}")
+    total = 0
+    for f in files:
+        validate_file_size(f, _CALIB_MAX_SIZE_BYTES)
+        total += os.path.getsize(f)
+    if total > _CALIB_DIR_MAX_TOTAL_BYTES:
+        raise ValueError(
+            f"Aggregate calibration directory size {total} bytes exceeds "
+            f"{_CALIB_DIR_MAX_TOTAL_BYTES} bytes ({len(files)} shards under {path})."
```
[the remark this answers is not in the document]
gcunhase: Valid concern; deferring to a follow-up hardening PR for the same reason `score()`' public-API artifact size guards (comment above) were deferred: uncompressed-size bomb protection needs to live in `_load_calibration_from_path` to cover both CLI users and direct-`score()` callers, which expands the primitive's public-API failure surface (it would newly raise `ValueError` on legitimate large calibration sets).

The current on-disk cap combined with `allow_pickle=False` blocks the primary RCE vector; a targeted zip-bomb still succeeds against process memory but not against arbitrary code execution. Sketch of the follow-up: use `zipfile.ZipFile(path).infolist()[i].file_size` to bound uncompressed member sizes and aggregate across shards, plumbed into `_load_calibration_from_path` (not just `_validate_calibration_dir`) so both CLI and library paths are covered uniformly.

modelopt/onnx/quantization/sensitivity/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Returns:
+        Mean cosine distance across the sample axis, as a Python float in ``[0, 2]``.
+    """
+    p = _flatten_per_sample(fp16_act).astype(np.float64)
+    q = _flatten_per_sample(quant_act).astype(np.float64)
+    dot = np.sum(p * q, axis=-1)
+    norm = np.linalg.norm(p, axis=-1) * np.linalg.norm(q, axis=-1)
+    cos = dot / (norm + _EPS)
+    return float(np.mean(1.0 - cos))
```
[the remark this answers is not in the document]
gcunhase: Fixed in e6a382b8. `cos_dist` now uses a per-vector `both_zero = (p_norm == 0) & (q_norm == 0)` mask instead of guarding on the product `p_norm * q_norm`. Both zero → distance 0 (identical); only one zero → distance 1 (orthogonal). Verified locally on all 6 cases (both-zero, asymmetric-zero either direction, parallel, orthogonal, anti-parallel). Regression witness: `test_asymmetric_zero_vectors_return_max_distance` in `test_metrics.py`.

modelopt/onnx/quantization/sensitivity/metrics.py
```diff
@@ -0,0 +1,101 @@
...
+        quant_act: Activations from the quantized model with the same shape as ``fp16_act``.
+
+    Returns:
+        Mean cosine distance across the sample axis, as a Python float in ``[0, 2]``.
+    """
+    p = _flatten_per_sample(fp16_act).astype(np.float64)
+    q = _flatten_per_sample(quant_act).astype(np.float64)
+    dot = np.sum(p * q, axis=-1)
+    norm = np.linalg.norm(p, axis=-1) * np.linalg.norm(q, axis=-1)
+    cos = np.where(norm > 0, dot / (norm + _EPS), 1.0)
```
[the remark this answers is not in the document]
gcunhase: Fixed in e6a382b8. See the reply on the original thread for details: switched to `both_zero = (p_norm == 0) & (q_norm == 0)` mask.

modelopt/onnx/quantization/sensitivity/__main__.py
```diff
@@ -0,0 +1,271 @@
...
+    ``_CALIB_DIR_MAX_TOTAL_BYTES``.
+
+    Args:
+        path: Directory expected to contain one or more ``.npz`` calibration shards.
+
+    Raises:
+        FileNotFoundError: If ``path`` contains no ``.npz`` files.
+        ValueError: If any shard or the aggregate exceeds the limit.
+    """
+    import glob
```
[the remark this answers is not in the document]
gcunhase: Fixed in e6a382b8. Moved `import glob` from inside `_validate_calibration_dir` to the module-level import block.

tests/gpu/onnx/quantization/sensitivity/test_score.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+
+@pytest.mark.manual(reason="CoAtNet-0 integration; ~14 min on H100, needs pre-staged fixtures")
+def test_coatnet_op_type_matches_manual_groundtruth():
+    """CoAtNet-0 op-type ranking surfaces the ops that ``--op_types_to_quantize Conv`` avoids.
+
+    Top-4 = ``Add`` / ``Mul`` / ``LayerNormalization`` / ``ReduceMean`` (all > 1.5 KL); ``Conv``
+    sits ~10x below. Matches the manual "Conv-only wins 82% top-1" ground truth. Wall-clock
+    ~14 min on H100. Opt-in via ``pytest --run-manual``.
```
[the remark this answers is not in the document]
gcunhase: Fixed in e6a382b8. Added `assert all(scores[name] > 1.5 for name in expected_top4)` alongside the existing top-4-membership and `Conv < 0.5` checks, pinning the quantitative claim from the docstring.

cjluo-nv (commented):

Re-review of the ONNX sensitivity scan (15 files, +2262/-1). Good progress since the last round — several of the previous blockers are genuinely fixed.

**Addressed since last review**
- `failed: list[str]` + `_count_qdq_nodes()` now separate "probe inserted no Q/DQ" from "drift was 0.0". This is the important half of the op_type-vs-node probe-equivalence concern: a per-node probe on an op whose registry entry was dropped is no longer reported as a `0.0` "safe to quantize" score. Combined with the author's rationale (those nodes aren't quantized by ModelOpt's default policy either), I'm satisfied to defer the rest.
- `blocks=` / `block_agg=` now has `TestBlocks` (first-match-wins, singleton fallback, sum/max/mean, invalid-agg `ValueError`, union return, and the docs' `threshold+max ≡ coverage+max_nodes+sum` equivalence claim), and `metrics.py` has direct unit tests including the zero-vector regressions.
- CLI no longer carries a second calibration loader; `validate_file_size` guards added; `summarize_exclusion` now counts only known/unique names; test layout restructured into a `sensitivity/` sub-package mirroring `autotune/`.

**Still unresolved (details inline)**
1. `op_types.py` — `"Gelu"` in `get_activation_ops()` still flows into `QDQAutotunerBase.get_ort_quantization_config()` → `op_types_needing_output_quant`, i.e. it changes autotune Q/DQ placement for any Gelu model. The reply explains *why* the addition is correct, but there is still no autotune-side test and no CHANGELOG line for the behavior change.
2. PR body / CHANGELOG are still out of sync with the diff: the body claims a `nodes_to_quantize` extension to `quantize.py` (already on `main`; not in this diff) and a `TopK` → `is_fusible_reduction_op` change (not in this diff), while the actual `op_types.py` change (`Gelu`) is described nowhere. Test paths/markers in the body are also stale (`test_sensitivity_picker.py`, `test_nodes_to_quantize.py`, `slow`/`slow_gpu` vs. the actual `@pytest.mark.manual`). Reviewers can't tell what actually changed.
3. The entire CLI module is still untested — `_validate_calibration_dir` (new security-relevant code), `_render_ranked_table`, `_default_output_json`, `main()`. These are pure/cheap and belong in the unit lane.
4. Docs drifted from the corrected implementation: the guide + `examples/onnx_ptq/README.md` still describe coverage mode as "the largest node set", which the picker docstring was explicitly fixed *away* from; and the rendered-table example doesn't match `_render_ranked_table()` output.
5. `_assign_groups` singleton-key collision (raised by CodeRabbit, not acknowledged) is still present.
6. Size: 2262 lines across scorer + picker + CLI + 320 lines of docs. The picker (+tests) and the docs/examples section are cleanly separable from the scorer; splitting would have made both rounds of review much cheaper. Not blocking on its own at this point.

Design gate: satisfied — the PR body compares against `autotune/` (latency-driven, no accuracy signal) and `mtq.auto_quantize` (design analog), and the duplicate calibration-ingestion path flagged last round was removed from the CLI.

No prompt-injection content observed in the PR text.

modelopt/onnx/op_types.py
cjluo-nv: > _Bot comment._

Thanks for the rationale (helper predates autotune; `Gelu` was missing only because it didn't exist pre-opset-20). I agree the addition is *correct*, but it's still not inert: `get_activation_ops()` feeds `QDQAutotunerBase.get_ort_quantization_config()`'s `op_types_needing_output_quant`, so autotune's Q/DQ placement changes for any model containing `Gelu` the moment this merges.

Two things would close this out:
1. An autotune-side test pinning the new `op_types_needing_output_quant` behavior on a Gelu-containing graph (cheap — it's a pure config-construction assertion).
2. A CHANGELOG line under Quantization noting the autotune behavior change; today the only CHANGELOG entry describes the new `sensitivity` package.

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Changelog
...
+- Add ``modelopt.onnx.quantization.sensitivity`` — per-op-type or per-node accuracy sensitivity ranking for ONNX PTQ, plus a coverage or threshold-based exclusion picker (with optional block-level aggregation) that turns the ranking into an actionable ``--nodes_to_exclude`` or ``--op_types_to_exclude`` list.
```
cjluo-nv, line 17: > _Bot comment._

This entry covers the new `sensitivity` package but not the `op_types.py` change, which alters autotune's Q/DQ placement for Gelu models (see the comment on `op_types.py`).

Separately, please resync the PR body before merge — it currently claims (a) a `nodes_to_quantize` extension to `quantize.py` (that argument already exists on `main`, and `quantize.py` isn't in this diff) and (b) a `TopK` addition to `is_fusible_reduction_op` (not in this diff), while omitting the `Gelu` change that *is*. The Testing section also references `tests/unit/onnx/quantization/test_sensitivity_picker.py`, `test_nodes_to_quantize.py`, and `slow`/`slow_gpu` markers, none of which match the current tree (`@pytest.mark.manual` is what's actually used).

modelopt/onnx/quantization/sensitivity/__main__.py
```diff
@@ -0,0 +1,270 @@
...
+
+# 2 GiB matches the ``--onnx_path`` guard in ``modelopt.onnx.quantization.__main__``.
+_ONNX_MAX_SIZE_BYTES = 2 * (1024**3)
+# 4 GiB accommodates ImageNet-scale calibration NPZ files.
+_CALIB_MAX_SIZE_BYTES = 4 * (1024**3)
+# 16 GiB aggregate cap for a directory of .npz shards.
+_CALIB_DIR_MAX_TOTAL_BYTES = 16 * (1024**3)
+
+
+def _validate_calibration_dir(path: str) -> None:
```
cjluo-nv, line 42: > _Bot comment._

The whole CLI module is still untested — `_validate_calibration_dir`, `_render_ranked_table`, `_default_output_json`, and `main()` have no coverage in this PR. `_validate_calibration_dir` in particular is new security-relevant code (per-shard + aggregate caps, `FileNotFoundError` on an empty dir) and is pure filesystem logic that runs in milliseconds in the CPU unit lane. `_render_ranked_table` also has non-trivial branches worth pinning: empty `scores` with/without `failed`, all-zero scores with `show_zero_scores=False`, and the hidden-count footer.

A small `tests/unit/onnx/quantization/sensitivity/test_cli.py` covering those (plus `main(argv=[...])` with `score` monkeypatched) would close the last of the "untested new surface" thread from the previous round.

modelopt/onnx/quantization/sensitivity/__main__.py
```diff
@@ -0,0 +1,270 @@
...
+from __future__ import annotations
+
+import argparse
+import glob
+import json
+import os
+import sys
+
+from modelopt.onnx.logging_config import logger
+from modelopt.onnx.quantization.__main__ import validate_file_size
```
cjluo-nv: > _Bot comment._

Importing a library helper out of another package's `__main__` module (`from modelopt.onnx.quantization.__main__ import validate_file_size`) pulls in that entrypoint's whole import graph (`autotune.utils`, `quantize`, …) just to reuse a 15-line size check, and couples the two CLIs in a direction that's easy to break.

Since `validate_file_size` is a generic path guard, consider moving it to `modelopt/onnx/utils.py` and having both `__main__` modules import it from there.

modelopt/onnx/quantization/sensitivity/picker.py
```diff
@@ -0,0 +1,258 @@
...
+        for gname, patterns in blocks.items()
+    }
+    groups: dict[str, list[str]] = {}
+    for node_name in scores:
+        matched: str | None = None
+        for gname, pats in compiled.items():
+            if any(pat.match(node_name) for pat in pats):
+                matched = gname
+                break
+        key = matched if matched is not None else node_name
```
cjluo-nv: > _Bot comment._

Unaddressed from the previous round: an unmatched node uses its own name as the group key, so if a node name happens to equal a configured group name, `setdefault` silently merges the singleton into that group and its score is aggregated with the block's.

With `scores={"g_node": 0.01, "g": 10.0}` and `blocks={"g": [r"^g_node$"]}` the node `"g"` (unmatched) lands in group `"g"` alongside `g_node`. Cheap fix: namespace singleton keys (e.g. a `("__node__", node_name)` tuple key internally, or a sentinel prefix) or raise on a collision between an unmatched node name and a `blocks` key.

docs/source/guides/_onnx_quantization.rst
```diff
@@ -121,3 +121,323 @@ The following command will build the engine using fp16 precision. After building
...
+The :func:`sensitivity.score` output is a dictionary from target name to sensitivity score
+(see ``metric`` in :ref:`sensitivity-supported-options` above). The picker
+function :func:`sensitivity.suggest_exclusion` turns that dictionary into an actionable
+``--nodes_to_exclude`` or ``--op_types_to_exclude`` list, depending on granularity, for
+:func:`modelopt.onnx.quantization.quantize`, and :func:`sensitivity.summarize_exclusion`
+reports what the exclusion set covers.
+
+Two policy modes are supported:
+
+- **Coverage mode** (default): exclude the largest node set whose cumulative sensitivity score
```
cjluo-nv: > _Bot comment._

This still describes coverage mode as "exclude the largest node set whose cumulative sensitivity score stays at or below `coverage * total_mass`" — which is exactly the wording the `suggest_exclusion` docstring was corrected away from in this same PR. `_pick_from_scores` walks the ranking and `break`s at the first target that doesn't fit, so it returns a rank *prefix*, not the largest fitting subset (for `{a:4, b:3, c:2}` at `coverage=0.7` it returns `["a"]`, not `{a, c}`).

Please mirror the corrected docstring wording here and in `examples/onnx_ptq/README.md` ("leaves the largest set whose cumulative sensitivity mass stays at or below the requested fraction").

docs/source/guides/_onnx_quantization.rst
```diff
@@ -121,3 +121,323 @@ The following command will build the engine using fp16 precision. After building
...
+      Add                 2.848  <-- highest impact
+      Mul                 1.890
+      LayerNormalization  1.653
+      ReduceMean          1.570
+      BatchNormalization  0.355
+      Conv                0.181
+      AveragePool         0.057
+      Sigmoid             0.039
+      MatMul              0.015
+      Relu               ~0
```
cjluo-nv: > _Bot comment._

This sample output can't be produced by `_render_ranked_table()`: values are formatted with `{value:.3f}` (so `~0` never appears), and exactly-`0.0` rows are hidden unless `--show_zero_scores` is passed — yet the block shows a `Gemm  0  <-- lowest impact` row *and* a `(1 target(s) with score 0.0 hidden; …)` footer at the same time.

Either regenerate this block from an actual run, or show the `--show_zero_scores` invocation that would render the zero rows (in which case the hidden-count footer shouldn't be there).

tests/unit/onnx/quantization/test_quantize_api.py
```diff
@@ -196,3 +197,45 @@ def test_quantize_opset_handling(
...
     )
+
+
+def test_quantize_honors_nodes_to_quantize_allowlist(tmp_path):
+    """``nodes_to_quantize=[<regex>]`` inserts QDQ around the matched Conv only.
+
+    Guards the primitive the ONNX sensitivity scanner relies on to isolate a single node for a
+    per-target probe; also documents the API contract of the flag itself.
+    """
+    import onnx_graphsurgeon as gs
```
cjluo-nv: > _Bot comment._

`onnx_graphsurgeon` and `modelopt.onnx.utils.save_onnx` are neither optional nor circular here (the module already imports `onnx`, `torch`, `modelopt.onnx.quantization`), so these should go with the module-level imports at the top of the file — import errors then surface at collection time rather than mid-test. Non-blocking.

modelopt/onnx/op_types.py
gcunhase: Adding Gelu to `get_activation_ops()` allows it to behave exactly like the other activation ops (Relu, Sigmoid, etc.) inside autotune's config derivation. There's no other side effects on scheme generation or region discovery, since `get_activation_ops()` is only referenced at `autotuner_base.py:547`.

The final quantized model looks identical for Relu or Gelu in the following example:
- Non-quantized: `input -> MatMul -> Add -> {Relu|Gelu} -> output`
- Quantized: `input -> MatMul -> Q/DQ -> Add -> Q/DQ -> {Relu|Gelu} -> output`

@cjluo-nv do you think it's still necessary to add a test for this? The only reason why Gelu got added was to update this function with this new opset 20 function.

modelopt/onnx/quantization/sensitivity/__main__.py
gcunhase: Addressed in 528ed342. New `tests/unit/onnx/quantization/sensitivity/test_cli.py` (17 tests) covers `_validate_calibration_dir` (empty dir / per-file cap / aggregate cap / happy path), `_render_ranked_table` (no-scores + failed, all-zero + `show_zero_scores`, hidden-count footer, failed footer), `_default_output_json`, and `main()` (synthetic happy path, default output-JSON path, oversize ONNX rejection, calibration-dir validation). `score` is monkeypatched, so the tests run in the CPU unit lane in < 1 s.

modelopt/onnx/quantization/sensitivity/__main__.py
gcunhase: Addressed in 528ed342. Moved `validate_file_size` to `modelopt/onnx/utils.py` alongside the other low-level ONNX helpers. Both `modelopt.onnx.quantization.__main__` and `modelopt.onnx.quantization.sensitivity.__main__` now import it from there, so the sensitivity CLI no longer pulls in the quantize CLI's import graph.

modelopt/onnx/quantization/sensitivity/picker.py
gcunhase: Addressed in 528ed342. `_assign_groups` now prefixes unmatched-node singleton keys with a `"\0singleton:"` sentinel that cannot collide with any user-supplied group name. On the reproducer (`scores={"g_n1": 0.01, "g": 10.0}`, `blocks={"g": [r"^g_"]}`, `threshold=1.0`, `block_agg="sum"`), the standalone `g` node is now correctly excluded on its own instead of merging into the group. Regression witness: `test_unmatched_node_name_collision_with_group_name_stays_isolated` in `test_picker.py`.

docs/source/guides/_onnx_quantization.rst
gcunhase: Addressed in 528ed342. Both `docs/source/guides/_onnx_quantization.rst` and `examples/onnx_ptq/README.md` now describe coverage mode as walking targets in descending score order and accumulating until the next one would push cumulative sensitivity above `coverage * total_mass` (rank-prefix), matching the corrected `suggest_exclusion` docstring.

docs/source/guides/_onnx_quantization.rst
gcunhase: Addressed in 528ed342. Regenerated the sample block in `_onnx_quantization.rst` to match `_render_ranked_table()`'s actual output: dropped the `~0` string rows (only `{value:.3f}` is emitted), moved the lowest-impact marker to the last visible non-zero row (`MatMul 0.015`), and updated the hidden-count footer to `(4 target(s) with score 0.0 hidden; ...)`. No `show_zero_scores` toggle needed.

tests/unit/onnx/quantization/test_quantize_api.py
gcunhase: Addressed in 528ed342. `onnx_graphsurgeon as gs` and `modelopt.onnx.utils.save_onnx` are hoisted to the module-level import block in `tests/unit/onnx/quantization/test_quantize_api.py`.

modelopt/onnx/op_types.py
gcunhase: Deferring both asks in this PR. The addition is a completeness fix (Gelu is opset-20+; pre-opset-20 there is no ONNX "Gelu" op for `node.op` to match), so any existing autotune workflow was already producing the incomplete `op_types_needing_output_quant` output on Gelu-containing graphs -- there's no working prior behavior to preserve or announce. Autotune's use of `get_activation_ops` is a single reference at `autotuner_base.py:547` with no downstream side effects on scheme generation. If you'd still like the autotune-side test + CHANGELOG line, happy to add both in a follow-up.

CHANGELOG.rst
gcunhase: Splitting into two:

- **CHANGELOG**: the `op_types.py` Gelu addition is deferred per the reply on that thread, so no new CHANGELOG line for it.
- **PR body**: refreshed. Dropped the stale `nodes_to_quantize` / `TopK` / `op_types.py` claims (none in this diff), corrected the file paths to `tests/unit/onnx/quantization/sensitivity/test_{picker,metrics,cli}.py` and `tests/gpu/onnx/quantization/sensitivity/test_score.py`, and replaced the `slow` / `slow_gpu` references with `@pytest.mark.manual` (the `slow` markers were removed from `pyproject.toml` in an earlier round).

cjluo-nv (commented):

Re-review #3 of the ONNX sensitivity scan (17 files, +2506/-27). Every critical item from the last round is now resolved; what's left is documentation/CHANGELOG hygiene around a cross-cutting behavior change plus a few small nits, so I'm flagging for owner sign-off rather than blocking.

**Advice on item 1 (Gelu):** confirmed against the code. `get_activation_ops()` is consumed only by `autotuner_base.py`, at the single `op_types_needing_output_quant` site inside `get_ort_quantization_config()` — not by region discovery, `RegionPattern`, scheme mutation, or `get_autotuner_quantizable_ops()`. The functional delta is that an insertion point landing on a `Gelu` input now propagates its producer op type into `op_types_needing_output_quant`; previously that point was silently dropped, i.e. autotune could fail to honor a requested Q/DQ on a Gelu-fed tensor. That makes this a bug fix, consistent with your `MatMul → Add → {Relu|Gelu}` experiment. I'm satisfied and withdraw the autotune-test ask as a blocker.

**Addressed since round 2**
- CLI is no longer untested: `tests/unit/onnx/quantization/sensitivity/test_cli.py` covers `_validate_calibration_dir` (empty dir / per-shard cap / aggregate cap / happy path), `_render_ranked_table` (all footer variants), `_default_output_json`, and `main()` with `score` monkeypatched, all in the CPU lane.
- Docs drift fixed: coverage mode is now described as the rank-prefix walk in both `_onnx_quantization.rst` and `examples/onnx_ptq/README.md`, and the rendered table matches `_render_ranked_table()` (`{:.3f}`, no `~0` rows, `(4 target(s) with score 0.0 hidden; ...)` footer).
- `_assign_groups` singleton/group key collision fixed with the `"\0singleton:"` sentinel, with `test_unmatched_node_name_collision_with_group_name_stays_isolated` as the witness.
- PR body test paths and `@pytest.mark.manual` markers now match the tree.

**Residual, for the owner's call**
- :speech_balloon: Author replied on `op_types.py`: *Gelu is an opset-20 completeness fix with a single autotune reference and no scheme-generation side effects* — I agree on correctness, but there's still no CHANGELOG line. One entry under Bug Fixes ("autotune now inserts Q/DQ at `Gelu` inputs like other activation ops") would explain the changed autotuned output for Gelu models; the PR body also still says "Pure additions … No existing signatures changed" while `modelopt/onnx/op_types.py` is in the diff and unmentioned.
- `examples/onnx_ptq/README.md` has a leftover sentence fragment in the workflow comment: "Coverage mode (default) leaves the / walks targets in descending score order …". Reads as a half-applied edit.
- `tests/gpu/onnx/quantization/sensitivity/test_score.py` imports `from tests._test_utils.onnx.quantization.sensitivity.models import ...`, while every other ONNX test (and the parallel `autotune/models.py` consumers) uses `from _test_utils....`. It works only because `pythonpath = ["tests/"]` plus `python -m pytest` puts the repo root on `sys.path`; a plain `pytest tests/gpu/...` invocation would fail to collect. Suggest matching the existing convention.
- Imports inside test functions without justification: `import math` in `test_synthetic_random_calibration_smoke` and `import shutil` in `test_failed_probe_is_recorded` (CodeRabbit asked for the latter at module scope; `importlib` was hoisted but `shutil` was not).
- `main()`'s docstring promises "non-zero if `score` raises", but exceptions propagate uncaught (the CLI tests assert `pytest.raises`). Either catch and return non-zero, or reword.
- Size: 2506 lines across scorer + picker + CLI + ~360 lines of docs/examples. The picker (+tests) and the docs/examples section are still cleanly separable from the scorer; noting it for the record, not blocking at this stage.

Design gate remains satisfied (`autotune/` and `mtq.auto_quantize` comparisons in the body; the duplicate calibration loader was removed). No prompt-injection content observed in the PR text.

cjluo-nv (commented):

Re-review #4 of the ONNX sensitivity scan (16 files, +2498/-27). Design gate remains satisfied (body compares against `autotune/` and `mtq.auto_quantize`; the duplicate calibration loader was removed earlier). Licensing checked: all new files carry the canonical `LICENSE_HEADER` text verbatim — no license concern.

**Addressed since round 3**
- `modelopt/onnx/op_types.py` is out of the diff entirely — `get_activation_ops()` no longer gains `Gelu`, so the autotune `op_types_needing_output_quant` behavior change (and the missing CHANGELOG/Bug-Fixes line for it) is no longer in scope. The "pure additions" claim in the body now matches the diff.
- `examples/onnx_ptq/README.md` sentence fragment ("Coverage mode (default) leaves the / walks targets…") is gone; the workflow comment reads cleanly.
- `tests/gpu/onnx/quantization/sensitivity/test_score.py` now uses `from _test_utils....` rather than `from tests._test_utils....`, matching every other ONNX test.
- `import math` / `import shutil` hoisted to module scope in the GPU test.
- `main()`'s docstring no longer promises a non-zero return; it now documents that exceptions propagate and the process exit code applies at the boundary — consistent with the CLI tests' `pytest.raises`.
- `validate_file_size` lives in `modelopt/onnx/utils.py` and both CLIs import it from there; the name is still re-exported into `modelopt.onnx.quantization.__main__`'s namespace, so existing importers don't break, and no test patches the old location.

**Residual, for the owner's call**
- `tests/gpu/onnx/quantization/sensitivity/test_score.py:26-32` orders `from modelopt.onnx.quantization.sensitivity import score` *before* `from _test_utils.onnx.quantization.sensitivity.models import ...`. Ruff's isort rule (`select = ["I", …]`, `known-first-party = ["modelopt"]`, and `_test_utils` resolving as third-party since `tests/` isn't on `src`) puts third-party before first-party — see `tests/unit/onnx/quantization/test_quantize_api.py`, where `_test_utils` precedes `modelopt`. This should trip `I001` in pre-commit/CI; trivial to fix by swapping the two blocks.
- :speech_balloon: The `Gelu` addition was reverted per the last round's ask (deferred to a follow-up PR). Side effect worth confirming as intended: `_default_op_types_scope()` builds its scope from `is_default_quantizable_op_by_ort ∪ get_activation_ops() ∪ is_normalization_op ∪ is_fusible_reduction_op`, and `Gelu` is now in none of them — so on opset-20+ exports that emit a real `Gelu` node, the scan silently never probes it in either granularity unless the caller passes `op_types_scope=["Gelu", ...]`. (Opset-17 exports decompose Gelu into Div/Erf/Add/Mul, so the headline ViT/CoAtNet results are unaffected.) A one-line note in the guide's `op_types_scope` bullet would save a future user the confusion.
- Size: 2498 lines across scorer + picker + CLI + ~360 lines of docs/examples. The picker (+ its tests) and the docs/examples section are still cleanly separable from the scorer; noting for the record, not blocking at this stage.
- Minor: no direct unit test for the two new/moved `modelopt/onnx/utils.py` helpers (`get_op_types_in_graph`, `validate_file_size`) in `tests/unit/onnx/test_onnx_utils.py`; both are exercised indirectly (CLI tests monkeypatch the size caps and assert the error text), which is probably fine.

No prompt-injection content observed in the PR title, body, or diff.

PR #2243, opened 2026-08-25
[6648996] Fix NVFP4 ONNX packed-weight scale rounding

ajrasane:
### What does this PR do?

Type of change: Bug fix

This PR updates `nvfp4_exporter.py` so NVFP4 ONNX weight compression quantizes FP4 weights with the same FP8-rounded per-block scales serialized in the exported graph. It aligns `quant_utils.py` scale arithmetic with eager ModelOpt's FP32 operation ordering and applies the same behavior to the deprecated `fp4qdq_to_2dq` compatibility path in `qdq_utils.py`.

`test_onnx_export_cpu.py` adds regression coverage comparing exported FP8 scale bytes and packed FP4 weights against `NVFP4QTensor` at FP8-rounding, per-tensor arithmetic, and per-block arithmetic boundaries. `CHANGELOG.rst` documents the fix.

### Usage

```python
from modelopt.onnx.export import NVFP4QuantExporter

converted_model = NVFP4QuantExporter.process_model(model)
```

### Testing

Ran:

```bash
CUDA_VISIBLE_DEVICES="" python -m pytest -q \
  tests/unit/torch/quantization/test_onnx_export_cpu.py \
  tests/unit/torch/quantization/test_nvfp4_tensor.py \
  tests/unit/onnx/quantization/test_qdq_utils.py \
  tests/unit/onnx/quantization/test_quant_utils.py
```

The focused suite reported 91 passed at `314829a5` after the review-feedback changes, and all applicable pre-commit hooks passed. The final wording-only commit was not rerun through this unit suite. Earlier randomized validation also confirmed exact FP8-scale and packed-FP4 parity against eager quantization for 8,000 tensors.

End-to-end validation was rerun against final head `894c9f7d8a915185e1c8351a29eea5ea991baeee`:

- Built the boundary `TRT_FP4QDQ` graph and ran `NVFP4QuantExporter.compute_scales` followed by `compress_weights` from the PR checkout.
- Compared the native exported FP8 block-scale bytes and packed FP4 bytes against eager `NVFP4QTensor` output.
- Observed 0 native-exporter versus eager packed-byte mismatches.
- Native exporter, explicit round-before-compression control, and eager packed FP4 outputs had the same SHA-256: `4b745ef652b49e2b82b2aef37a7d52aab9a74563c61eb1b53c7d2a5b57bdc0a3`.
- Exported and eager FP8 block scales had the same SHA-256: `f2d7e6732dac4df4194666bd17527f68d166788b26142cdd3f51855b1ce9dccc`.
- Confirmed that the native exporter no longer follows the unrounded-scale FP4 calculation path.

The standalone bug detector intentionally returns nonzero when its original mismatch predicate no longer holds; an outer validator asserted the fixed invariants and completed successfully. Its only compatibility update was moving the `_cast_fp8` harness import from the removed `nvfp4_exporter` helper to `qdq_utils._cast_fp8`; the model, weights, export flow, and parity assertions were unchanged. This validation covers the ONNX export and eager byte-parity boundary, not TensorRT engine compilation or inference.

## Summary by CodeRabbit

- **New Features**
  - Added calibration-free streaming conversion and checkpoint-mirroring workflows for Kimi-K3 NVFP4 models.
  - Added support for temporary weight-folding contexts, activation headroom calibration, distillation, per-expert quantization, grouped-linear compilation, MLflow logging, Hugging Face post-training quantization, and vLLM fake-quantized serving.

- **Bug Fixes**
  - Improved NVFP4 ONNX export accuracy and compatibility through precise FP32/FP8 scale handling.
  - Added validation for invalid, non-finite, and negative scaling values.
  - Avoided unnecessary GPU capability checks during CPU or disabled fast-path quantization.
  - Removed deprecated quantization options and legacy configuration interfaces.

### DIFF
--- a/modelopt/onnx/export/nvfp4_exporter.py
+++ b/modelopt/onnx/export/nvfp4_exporter.py
@@ -30,6 +30,7 @@ from modelopt.onnx.quantization.quant_utils import (
     quantize,
 )
 from modelopt.torch.quantization.qtensor import NVFP4QTensor
+from modelopt.torch.quantization.qtensor.nvfp4_tensor import _cast_per_block_scale_to_fp8
 
 from .base_exporter import ONNXQuantExporter
 
@@ -65,6 +66,23 @@ def _cast_fp8(array: np.ndarray) -> np.ndarray:
     array_f8_t = array_f32_t.clamp(min=-448, max=448).to(torch.float8_e4m3fn).view(torch.uint8)
     array_f8 = array_f8_t.cpu().numpy().astype(np.uint8)
     return array_f8
+
+
+def _encode_nvfp4_block_scale(array: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+    """Return FP8-rounded block scales as FP32 values and encoded bytes."""
+    if not np.all(np.isfinite(array)) or np.any(array < 0):
+        raise ValueError("NVFP4 block scales must be finite and nonnegative.")
+
+    array_f32_t = torch.from_numpy(array)
+    if torch.cuda.is_available():
+        array_f32_t = array_f32_t.cuda()
+    array_f8_t = _cast_per_block_scale_to_fp8(array_f32_t)
+    array_f32 = array_f8_t.float().cpu().numpy()
+    array_f8 = array_f8_t.view(torch.uint8).cpu().numpy().astype(np.uint8)
+
+    if not np.all(np.isfinite(array_f32)) or np.any(array_f32 <= 0):
+        raise ValueError("NVFP4 block scales are not representable in FLOAT8E4M3FN.")
+    return array_f32, array_f8
 
 
 def _replace_fp4qdq_with_2dq(
@@ -295,12 +313,12 @@ class NVFP4QuantExporter(ONNXQuantExporter):
 
             sw_f32_per_block = sw_f32_per_block.reshape(sw_per_block_shape)
 
-            # Quantize weights
-            w_f32 = quantize(w32, block_size, sw_f32_per_block, sw_f32_per_tensor)
+            if not np.all(np.isfinite(sw_f32_per_tensor)) or np.any(sw_f32_per_tensor <= 0):
+                raise ValueError("NVFP4 per-tensor scales must be finite and positive.")
 
-            # Cast to FP4 and FP8
+            sw_f32_per_block, sw_f8_per_block = _encode_nvfp4_block_scale(sw_f32_per_block)
+            w_f32 = quantize(w32, block_size, sw_f32_per_block, sw_f32_per_tensor)
             w_f4 = _cast_fp4(w_f32)
-            sw_f8_per_block = _cast_fp8(sw_f32_per_block)
 
             # Store compressed data as node attributes for post_process
             w_f4_attr = node.attribute.add()
--- a/modelopt/onnx/export/nvfp4_exporter.py
+++ b/modelopt/onnx/export/nvfp4_exporter.py
@@ -30,7 +30,6 @@ from modelopt.onnx.quantization.quant_utils import (
     quantize,
 )
 from modelopt.torch.quantization.qtensor import NVFP4QTensor
-from modelopt.torch.quantization.qtensor.nvfp4_tensor import _cast_per_block_scale_to_fp8
 
 from .base_exporter import ONNXQuantExporter
 
@@ -58,16 +57,6 @@ def _cast_fp4(array: np.ndarray) -> np.ndarray:
     return array_f4
 
 
-def _cast_fp8(array: np.ndarray) -> np.ndarray:
-    """Cast a numpy array to FLOAT8E4M3FN using PyTorch."""
-    array_f32_t = torch.from_numpy(array)
-    if torch.cuda.is_available():
-        array_f32_t = array_f32_t.cuda()
-    array_f8_t = array_f32_t.clamp(min=-448, max=448).to(torch.float8_e4m3fn).view(torch.uint8)
-    array_f8 = array_f8_t.cpu().numpy().astype(np.uint8)
-    return array_f8
-
-
 def _encode_nvfp4_block_scale(array: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
     """Return FP8-rounded block scales as FP32 values and encoded bytes."""
     if not np.all(np.isfinite(array)) or np.any(array < 0):
@@ -76,12 +65,10 @@ def _encode_nvfp4_block_scale(array: np.ndarray) -> tuple[np.ndarray, np.ndarray
     array_f32_t = torch.from_numpy(array)
     if torch.cuda.is_available():
         array_f32_t = array_f32_t.cuda()
-    array_f8_t = _cast_per_block_scale_to_fp8(array_f32_t)
+    array_f8_t = NVFP4QTensor._cast_per_block_scale_to_fp8(array_f32_t)
     array_f32 = array_f8_t.float().cpu().numpy()
     array_f8 = array_f8_t.view(torch.uint8).cpu().numpy().astype(np.uint8)
 
-    if not np.all(np.isfinite(array_f32)) or np.any(array_f32 <= 0):
-        raise ValueError("NVFP4 block scales are not representable in FLOAT8E4M3FN.")
     return array_f32, array_f8
 
 
@@ -312,9 +299,6 @@ class NVFP4QuantExporter(ONNXQuantExporter):
             )
 
             sw_f32_per_block = sw_f32_per_block.reshape(sw_per_block_shape)
-
-            if not np.all(np.isfinite(sw_f32_per_tensor)) or np.any(sw_f32_per_tensor <= 0):
-                raise ValueError("NVFP4 per-tensor scales must be finite and positive.")
 
             sw_f32_per_block, sw_f8_per_block = _encode_nvfp4_block_scale(sw_f32_per_block)
             w_f32 = quantize(w32, block_size, sw_f32_per_block, sw_f32_per_tensor)
--- a/modelopt/onnx/quantization/qdq_utils.py
+++ b/modelopt/onnx/quantization/qdq_utils.py
@@ -1520,7 +1520,11 @@ def fp4qdq_to_2dq(onnx_model: onnx.ModelProto, verbose: bool = False) -> onnx.Mo
     )
 
     # Lazy import to avoid a circular import: nvfp4_exporter imports from this module.
-    from modelopt.onnx.export.nvfp4_exporter import _cast_fp4, _replace_fp4qdq_with_2dq
+    from modelopt.onnx.export.nvfp4_exporter import (
+        _cast_fp4,
+        _encode_nvfp4_block_scale,
+        _replace_fp4qdq_with_2dq,
+    )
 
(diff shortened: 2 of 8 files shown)


modelopt/onnx/export/nvfp4_exporter.py
```diff
@@ -30,6 +30,7 @@
     quantize,
 )
 from modelopt.torch.quantization.qtensor import NVFP4QTensor
+from modelopt.torch.quantization.qtensor.nvfp4_tensor import _cast_per_block_scale_to_fp8
```
cjluo-nv: > _Bot comment._

This reaches into `modelopt.torch.quantization.qtensor.nvfp4_tensor` for a private symbol that is deliberately excluded from that module's `__all__` (`__all__ = ["NVFP4QTensor"]`). Sharing the eager implementation is the right instinct, but a private cross-package import means a rename in the torch package silently breaks the ONNX exporter with no test on the torch side guarding it. Consider exposing it as e.g. `NVFP4QTensor._cast_per_block_scale_to_fp8` (this file already goes through the class for `NVFP4QTensor._cast_fp4`) so there's a single documented entry point.

modelopt/onnx/export/nvfp4_exporter.py
```diff
@@ -67,6 +68,23 @@ def _cast_fp8(array: np.ndarray) -> np.ndarray:
...
+
+    array_f32_t = torch.from_numpy(array)
+    if torch.cuda.is_available():
+        array_f32_t = array_f32_t.cuda()
+    array_f8_t = _cast_per_block_scale_to_fp8(array_f32_t)
+    array_f32 = array_f8_t.float().cpu().numpy()
+    array_f8 = array_f8_t.view(torch.uint8).cpu().numpy().astype(np.uint8)
+
+    if not np.all(np.isfinite(array_f32)) or np.any(array_f32 <= 0):
+        raise ValueError("NVFP4 block scales are not representable in FLOAT8E4M3FN.")
```
cjluo-nv: > _Bot comment._

This post-check is unreachable: `_cast_per_block_scale_to_fp8` does `clamp(min=2**-9, max=E4M3_MAX)` before the cast, so `array_f32` is always finite and >= 2**-9. Either drop it or, if the intent is to guard against a future change in the eager helper, make that explicit in a comment — as written it reads like a live validation that can never fire.

modelopt/onnx/export/nvfp4_exporter.py
```diff
@@ -295,12 +313,12 @@ def compress_weights(onnx_model: onnx.ModelProto) -> onnx.ModelProto:
 
             sw_f32_per_block = sw_f32_per_block.reshape(sw_per_block_shape)
 
-            # Quantize weights
-            w_f32 = quantize(w32, block_size, sw_f32_per_block, sw_f32_per_tensor)
+            if not np.all(np.isfinite(sw_f32_per_tensor)) or np.any(sw_f32_per_tensor <= 0):
```
cjluo-nv: > _Bot comment._

This per-tensor validation only guards the exporter path. `fp4qdq_to_2dq` computes `sw_f32_per_tensor = get_weights_scaling_factor_2(w32)` with no equivalent check, so an inf per-tensor scale there makes every block scale `amax/(6*inf) == 0`, which the zero-mask then rewrites to 1.0 and the export silently produces garbage instead of failing. Since the stated goal is to keep the two paths behaviorally identical, consider moving the check into a small shared helper (or into `get_weights_scaling_factor_2` itself) so both call sites get it.

modelopt/onnx/quantization/qdq_utils.py
```diff
@@ -1584,11 +1588,10 @@ def _get_precision_dtype() -> str:
         w32 = read_f16_tensor_as_fp32(tensor)
         sw_f32_per_tensor = get_weights_scaling_factor_2(w32)
         sw_f32_per_block = get_weights_scaling_factor(w32, block_size, sw_f32_per_tensor)
+        sw_f32_per_block, sw_f8_per_block = _encode_nvfp4_block_scale(sw_f32_per_block)
```
cjluo-nv, line 1591: > _Bot comment._

Note that `_cast_fp8` is no longer called anywhere in `nvfp4_exporter.py` after this change (`mxfp8_exporter.py` imports the copy in this module, not that one). It is now duplicated dead code kept alive only by `TestCastFunctions.test_cast_fp8` in `tests/unit/onnx/quantization/test_qdq_utils.py`, which imports it from `nvfp4_exporter`. Worth deleting the `nvfp4_exporter` copy and repointing that test at `qdq_utils._cast_fp8`.

modelopt/onnx/quantization/quant_utils.py
```diff
@@ -125,8 +125,7 @@ def pack_float32_to_4bit_cpp_based(array: np.ndarray | Sequence, signed: bool) -
 
 def get_weights_scaling_factor_2(input: np.ndarray):
     """Returns per tensor weight scaling factor."""
-    per_block_scale_amax = np.max(np.abs(input)) / 6.0
-    per_block_quant_scale = per_block_scale_amax / 448.0
+    per_block_quant_scale = np.float32(np.max(np.abs(input))) / np.float32(6.0 * 448.0)
```
cjluo-nv, line 128: > _Bot comment._

The single-division reordering here (and in `get_weights_scaling_factor` below) is the numerically load-bearing part of the fix, but it's only exercised transitively through `test_onnx_export_cpu.py`. `tests/unit/onnx/quantization/test_quant_utils.py` has no coverage of either function, so a later well-meaning refactor back to `amax / 6.0 / 448.0` would only fail in a torch-quantization test file. Please add a direct case there with one of the boundary weights already picked for `fp32-tensor-arithmetic` / `fp32-block-arithmetic`, plus the amax==0 case (which returns 1.0 here but 0 in eager `get_weights_scaling_factor_2` — an intentional divergence worth pinning).

cjluo-nv (commented):

Small, focused fix (5 files, +134/-15) and the core change looks correct: `get_weights_scaling_factor{,_2}` now do the scale arithmetic in a single FP32 division matching `NVFP4QTensor` (`amax / (E2M1_MAX * E4M3_MAX)` and `amax / (E2M1_MAX * wsf2)`), the zero-mask now applies to the normalized quotient (same as eager's `per_block_scale[per_block_scale == 0] = 1.0`), and FP4 weights are now divided by the *FP8-rounded* block scale that is actually serialized — which is the real bug. Reusing eager's `_cast_per_block_scale_to_fp8` (clamp to `[2**-9, 448]`) rather than a second local implementation is the right call, and the new parametrized test asserting byte-exact parity of both the FP8 scale initializer and the packed FP4 initializer against `NVFP4QTensor.quantize` (over both the exporter and the deprecated shim) is solid coverage.

A few non-blocking issues, none of which change the correctness of the happy path:
1. The new per-tensor scale validation exists only in `NVFP4QuantExporter.compress_weights`; the deprecated `fp4qdq_to_2dq` path computes `sw_f32_per_tensor` inline with no check, so the two paths this PR is explicitly trying to keep in sync now validate differently.
2. The post-cast check inside `_encode_nvfp4_block_scale` is unreachable after the clamp in `_cast_per_block_scale_to_fp8`.
3. `nvfp4_exporter._cast_fp8` no longer has any production caller (MXFP8 uses `qdq_utils._cast_fp8`); it's now duplicated dead code kept alive only by `test_qdq_utils.py`.
4. The `quant_utils` arithmetic-ordering change is only covered indirectly through the exporter test; a direct case in `tests/unit/onnx/quantization/test_quant_utils.py` would pin the intended semantics closer to the code.

Also worth calling out in the CHANGELOG: tiny-but-nonzero block scales that previously flushed to FP8 zero are now clamped to `2**-9`, and non-finite/negative scales now raise instead of silently producing garbage — both are behavior changes for existing exports.

modelopt/onnx/export/nvfp4_exporter.py
ajrasane: Updated the exporter to call `NVFP4QTensor._cast_per_block_scale_to_fp8`; the class now exposes the shared eager helper as a static method, avoiding the private cross-package import.

modelopt/onnx/export/nvfp4_exporter.py
ajrasane: Removed the unreachable post-cast representability check. Input validation remains before the clamping and FP8 cast.

modelopt/onnx/export/nvfp4_exporter.py
ajrasane: Moved the per-tensor finiteness check into `get_weights_scaling_factor_2`, which is shared by the exporter and deprecated shim, and added direct non-finite regression coverage.

modelopt/onnx/quantization/qdq_utils.py
ajrasane: Removed the duplicate `nvfp4_exporter._cast_fp8` helper and repointed `TestCastFunctions` to the production helper in `qdq_utils`.

modelopt/onnx/quantization/quant_utils.py
ajrasane: Added direct `test_quant_utils.py` coverage that pins the per-tensor and per-block single-division FP32 boundaries, the zero-amax fallback, and non-finite rejection.

cjluo-nv (commented):

Re-review: all prior comments resolved. The fix is small (8 files, +155/-21), correct, and well tested.

Verified against the repo:
- `_encode_nvfp4_block_scale` now delegates to eager `NVFP4QTensor._cast_per_block_scale_to_fp8` (dedup rather than a second implementation) and both the exporter and the deprecated `fp4qdq_to_2dq` shim quantize FP4 weights with the FP8-rounded scale that is actually serialized — the real bug.
- `get_weights_scaling_factor{,_2}` single-FP32-division ordering matches eager (`amax/(E2M1_MAX*E4M3_MAX)`, `amax/(E2M1_MAX*wsf2)`), and the zero-mask now applies to the quotient as in eager. The `fp32-block-arithmetic` case (400.0 → FP8 ties-to-even 384 vs 416 under the old two-step division) is a meaningful regression test.
- `qdq_utils._cast_fp8` remains live via `_create_fp8_tensor`/MXFP8; only the `nvfp4_exporter` duplicate was removed and the test repointed.

Previous comments status: (1) per-tensor validation asymmetry → moved into the shared `get_weights_scaling_factor_2`; (2) unreachable post-cast check → removed; (3) dead duplicated `_cast_fp8` → deleted; (4) direct `test_quant_utils.py` coverage → added (ordering, zero-amax fallback, non-finite reject); (5) private cross-package import → now via `NVFP4QTensor._cast_per_block_scale_to_fp8`; (6) CHANGELOG behavior-change note → added.

Design gate: no new subsystem/abstraction — the change consolidates onto the existing eager helper, so no design justification is owed. No licensing-relevant changes.

Minor non-blocking nits: the `ValueError` message in `get_weights_scaling_factor_2` says "finite and positive" while the check is finiteness only (zero amax is mapped to 1.0 just above); the `staticmethod(...)` class-attribute alias is a slightly unusual way to expose the shared helper; and the stale auto-generated CodeRabbit summary in the PR body describes unrelated changes and should be dropped before merge.

Complex PR: spans 5 directories (≥ 5); 1 existing test file modified or removed. Looping in a human for approval.

cjluo-nv (approved):

PR #2263, opened 2026-08-27
Add Qwen 3.5 teacher-checkpoint VLM benchmark smoke evaluation

j-rausch:
### What does this PR do?

Type of change: New feature

This PR adds a reproducible VLM benchmark smoke workflow for the native Qwen 3.5 0.8B teacher checkpoint. It provides image and video evaluation baselines using Puzzletron's pinned `lmms-eval` and generic vLLM backend, without requiring an evaluator or dependency-stack upgrade.

The command defaults to a deterministic `short` suite over RealWorldQA and MMMU, limited to eight samples per task and repeated twice. It also provides:

- a 344-sample `quick` suite spanning RealWorldQA, MMMU, and MVBench;
- targeted video adapter smoke suites and a nine-benchmark `full` suite;
- pinned task and dataset revisions, offline media preparation, and isolated attempt artifacts;
- preflight checks for native checkpoint geometry, cached data, video decoding, judge authorization, and credentials.

This PR does not support Puzzletron-materialized pruned checkpoints. Those checkpoints currently advertise the `AnyModel` architecture, which the selected vLLM runtime cannot load. Export or loading compatibility and end-to-end evaluation of a materialized pruned VLM remain follow-up work.

Video results from the generic backend are suitable for development comparisons but are not equivalent to the newer native Qwen 3.5 backend's timestamp handling.

```bash
python -m examples.puzzletron.evaluation.vlm.run \
  --checkpoint /path/to/native-qwen3.5-0.8b-checkpoint \
  --output-dir /path/to/results \
  --suite short
```

### Testing

- Ran focused tests for the evaluator runner, suite policy, task generation, benchmark preparation, preflight validation, and pinned environment.
- Completed two deterministic short-suite evaluations on the native Qwen 3.5 0.8B teacher checkpoint with identical metrics.
- Completed a cached MVBench video smoke evaluation on the same teacher checkpoint.
- Did not run the full benchmark portfolio or evaluate a materialized pruned checkpoint.

### Review checklist

- Backward compatible: Yes
- New dependencies reviewed under `CONTRIBUTING.md`: Yes
- Necessary tests added: Yes
- Changelog: N/A


j-rausch:
Addressed the latest automated review on the current head:

- The evaluator loader now snapshots and restores every temporary package and source-module entry, including cleanup on load failure, while the returned callable retains its loaded module globals.
- Completed Hub snapshot symlinks are reused only when their resolved regular-file target remains within the repository cache. Partial and escaping symlinks still fail closed, and accepted files retain size and hash validation.
- Image-only suites remain usable without decord. Any selected video suite now fails before evaluator launch with a clear supported-environment message when no compatible reader is importable.
- The bundled compatibility hook now finds and executes the next inherited sitecustomize module, then restores its own module entry. A subprocess test proves both hooks coexist.
- No change was made for the manifest-structure finding. The manifest stores the revision under the nested lmms_eval object, and the assertion already reads environment["lmms_eval"]["commit"]. The focused test passes against the checked-in manifest.

All 76 focused evaluator and environment tests pass, and the complete changed-file hook set passes.

chochowski (approved):

PR #2270, opened 2026-08-27
Fix the llm_eval timeout: reachable MMLU mirror + no pipe deadlock

kevalmorabia97:
### What does this PR do?

Type of change: Bug fix

`tests/examples/llm_eval/test_llm_eval.py::test_qwen3_eval_fp8` has been failing with `Failed: Timeout (>900.0s) from pytest-timeout` on unrelated branches (runs 33027056246 and the one for `ad83a428`, while the 2026-08-24 nightly passed). It is not the test being slow — it is the harness deadlocking, and the deadlock also destroys the diagnostics that would explain the underlying kill.

**Mechanism.** The traceback shows `self = <Popen: returncode: -9 args: ['scripts/huggingface_example.sh', ...]>` while still blocked in `stdout.read()`. The launcher was SIGKILLed (nothing in pytest sends SIGKILL — pytest-timeout raises in the main thread, and the test's `finally` `pkill` sends SIGTERM and only runs afterwards — so an OOM kill is the likely source). But `subprocess.run(..., stdout=PIPE, stderr=STDOUT)` waits for **EOF on the pipe**, not for the process, and a surviving grandchild (the TRT-LLM serve/build worker) still holds the write end. EOF never arrives, so the test blocks until the 900 s alarm. Because the pipe is never drained, **every line of child output is discarded**, which is why the CI log says nothing about what the script was doing when it died.

Reduced to a self-contained reproducer:

```python
script = "sleep 300 & echo 'launcher output'; sleep 0.3; kill -9 $$"
subprocess.run(["bash", "-c", script], stdout=PIPE, stderr=STDOUT, text=True, timeout=20)
# -> TimeoutExpired: still blocked in communicate() after 20.0s, output lost
```

**Fix.** `_run_capturing` now starts the command in its own session, drains its output on a reader thread (so logs stream as they arrive instead of being buffered until the end), waits on the *process*, and kills the process group if descendants still hold the pipe after a 30 s grace period. A killed launcher now fails in seconds with its logs intact instead of silently burning the test's whole timeout.

This does not fix whatever kills the script; it makes it diagnosable. Worth noting separately: `test_qwen3_eval_fp8` took **749.10 s against its 900 s mark** on the last green nightly, so it is fragile regardless and may want its work trimmed or its budget raised once the logs show where the time goes.

### Usage

```python
# unchanged public API
run_example_command(cmd_parts, example_path="llm_eval")
```

### Testing

Verified against the reproducer above and on the normal paths:

| scenario | before | after |
| --- | --- | --- |
| launcher SIGKILLed, survivor holds the pipe | blocks indefinitely (900 s in CI) | `rc=-9` in 3.5 s, `'launcher output'` captured |
| the surviving descendant | keeps running | killed with the process group (stopped ticking, 20 -> 20 bytes) |
| normal exit | ok | `rc=0`, stdout and stderr interleaved in order |
| non-zero exit | ok | `rc=3`, output captured |

The example-test suites that use this helper run through the same code path; `tests/examples/megatron_bridge` (16 passed, 1 skipped) exercised it on nemo:26.08 in the branch this was extracted from.

- Is this change backward compatible?: ✅ — `_run_capturing` keeps its `(returncode, output)` contract; only the buffering strategy changed.
- Did you write any new necessary tests?: ❌ — this is test infrastructure; the scenario needs a process that outlives a SIGKILLed parent, which is awkward to assert in CI. Verified manually with the reproducer above.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — test-infrastructure fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command execution reliability with real-time output capture.
  * Ensured lingering child processes are cleaned up after commands exit or are interrupted.
  * Added warnings when forced cleanup may truncate output.
  * Prevented hangs when descendant processes keep output streams open.

* **Documentation**
  * Updated MMLU setup instructions to use the Hugging Face dataset repository.
  * Improved Windows instructions by explicitly using `curl.exe`.

* **Examples**
  * Improved MMLU downloads with retries, separate timeouts, resume support, and automatic temporary-file cleanup.

---

### Update: the second half of the failure

With the streaming fix in place, the next CI run showed the *actual* cause, which the old code had been hiding. That run blocked in `process.wait()` with `<Popen: returncode: None ...>` — child alive and working, not the previous dead-child pipe deadlock — and the now-visible output was:

```
--2026-08-27 18:09:20--  (try: 4)  https://people.eecs.berkeley.edu/~hendrycks/data.tar
Connecting to people.eecs.berkeley.edu ...|128.32.139.28|:443... failed: Connection timed out.
Retrying.
--2026-08-27 18:11:40--  (try: 5)  ...
```

`huggingface_example.sh` downloads the MMLU tarball from `people.eecs.berkeley.edu`, that host stopped answering around 2026-08-25, and wget's default retry policy (20 tries, ~2 min per connect timeout) consumed the whole 900 s budget. Not runner-specific: the URL also times out from a developer workstation, and the nightlies flipped 08-24 ✅ / 08-25 ✅ / **08-26 ❌ / 08-27 ❌**, matching the outage.

So this PR now carries both halves of the same failure:

1. the harness no longer deadlocks and no longer swallows the logs (`985809cc2d`), and
2. the MMLU data comes from HuggingFace's copy of the same tarball, with bounded retries (`40f1d89154`).

The mirror is byte-for-byte the same dataset in the same layout the script already expects — verified by running the exact download/extract commands:

```
https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar  ->  HTTP 200, 166 MB
data/mmlu/{dev,test,val}/  ->  57 subject CSVs each, plus auxiliary_train/
```

`wget --timeout=20 --tries=3` plus an explicit error means the next dataset-host outage fails in about a minute with "Could not download the MMLU test data. Set MMLU_DATA_PATH to a local copy." instead of silently eating a test's timeout. The same URL is updated in `examples/llm_eval/README.md` so a manual run does not hit the dead host either.

### DIFF
--- a/examples/hf_ptq/scripts/huggingface_example.sh
+++ b/examples/hf_ptq/scripts/huggingface_example.sh
@@ -318,16 +318,17 @@ if [[ $TASKS =~ "mmlu" ]]; then
     fi
     if [[ ! -d "$MMLU_DATA_PATH" ]] || [[ ! $(ls -A $MMLU_DATA_PATH) ]]; then
         echo "Preparing the MMLU test data"
-        # HuggingFace mirror of the original Berkeley tarball (people.eecs.berkeley.edu), which is
-        # intermittently unreachable. Bound the retries: wget's defaults keep trying for far longer
-        # than the callers' timeouts, turning an unreachable host into an unexplained hang.
-        wget --timeout=20 --tries=3 \
-            https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O /tmp/mmlu.tar || {
+        MMLU_TAR="$(mktemp "${TMPDIR:-/tmp}/mmlu.XXXXXX")" || exit 1
+        trap 'rm -f "$MMLU_TAR"' EXIT
+        # HuggingFace mirror of the Berkeley tarball, which is intermittently unreachable. Bound
+        # the retries (wget's defaults outlast the callers' timeouts); -c resumes a stalled transfer.
+        wget --connect-timeout=20 --read-timeout=60 --tries=3 -c \
+            https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O "$MMLU_TAR" || {
             echo "[ERROR] Could not download the MMLU test data. Set MMLU_DATA_PATH to a local copy."
             exit 1
         }
         mkdir -p data
-        tar -xf /tmp/mmlu.tar -C data && mv data/data $MMLU_DATA_PATH
+        tar -xf "$MMLU_TAR" -C data && mv data/data $MMLU_DATA_PATH
     fi
 
     mmlu_flags=""
--- a/examples/hf_ptq/scripts/huggingface_example.sh
+++ b/examples/hf_ptq/scripts/huggingface_example.sh
@@ -318,7 +318,14 @@ if [[ $TASKS =~ "mmlu" ]]; then
     fi
     if [[ ! -d "$MMLU_DATA_PATH" ]] || [[ ! $(ls -A $MMLU_DATA_PATH) ]]; then
         echo "Preparing the MMLU test data"
-        wget https://people.eecs.berkeley.edu/~hendrycks/data.tar -O /tmp/mmlu.tar
+        # HuggingFace mirror of the original Berkeley tarball (people.eecs.berkeley.edu), which is
+        # intermittently unreachable. Bound the retries: wget's defaults keep trying for far longer
+        # than the callers' timeouts, turning an unreachable host into an unexplained hang.
+        wget --timeout=20 --tries=3 \
+            https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O /tmp/mmlu.tar || {
+            echo "[ERROR] Could not download the MMLU test data. Set MMLU_DATA_PATH to a local copy."
+            exit 1
+        }
         mkdir -p data
         tar -xf /tmp/mmlu.tar -C data && mv data/data $MMLU_DATA_PATH
     fi
--- a/examples/hf_ptq/scripts/huggingface_example.sh
+++ b/examples/hf_ptq/scripts/huggingface_example.sh
@@ -320,15 +320,18 @@ if [[ $TASKS =~ "mmlu" ]]; then
         echo "Preparing the MMLU test data"
         MMLU_TAR="$(mktemp "${TMPDIR:-/tmp}/mmlu.XXXXXX")" || exit 1
         trap 'rm -f "$MMLU_TAR"' EXIT
-        # HuggingFace mirror of the Berkeley tarball, which is intermittently unreachable. Bound
-        # the retries (wget's defaults outlast the callers' timeouts); -c resumes a stalled transfer.
+        # Revision-pinned HuggingFace mirror of the Berkeley tarball, which is unreachable. Bound
+        # the retries, which otherwise outlast the callers' timeouts, and resume rather than refetch.
         wget --connect-timeout=20 --read-timeout=60 --tries=3 -c \
-            https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O "$MMLU_TAR" || {
+            https://huggingface.co/datasets/cais/mmlu/resolve/c30699e8356da336a370243923dbaf21066bb9fe/data.tar \
+            -O "$MMLU_TAR" || {
             echo "[ERROR] Could not download the MMLU test data. Set MMLU_DATA_PATH to a local copy."
             exit 1
         }
         mkdir -p data
         tar -xf "$MMLU_TAR" -C data && mv data/data $MMLU_DATA_PATH
+        rm -f "$MMLU_TAR"  # 166MB; do not hold it for the rest of the eval
+        trap - EXIT
     fi
 
     mmlu_flags=""
--- a/examples/llm_eval/README.md
+++ b/examples/llm_eval/README.md
@@ -155,7 +155,8 @@ Download data
 
 ```bash
 mkdir -p data
-wget https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O data/mmlu.tar
+wget --connect-timeout=20 --read-timeout=60 --tries=3 -c \
+    https://huggingface.co/datasets/cais/mmlu/resolve/c30699e8356da336a370243923dbaf21066bb9fe/data.tar -O data/mmlu.tar
 tar -xf data/mmlu.tar -C data && mv data/data data/mmlu
 cd ..
 ```
--- a/examples/llm_eval/README.md
+++ b/examples/llm_eval/README.md
@@ -155,7 +155,7 @@ Download data
 
 ```bash
 mkdir -p data
-wget https://people.eecs.berkeley.edu/~hendrycks/data.tar -O data/mmlu.tar
+wget https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O data/mmlu.tar
 tar -xf data/mmlu.tar -C data && mv data/data data/mmlu
 cd ..
 ```
--- a/examples/windows/accuracy_benchmark/README.md
+++ b/examples/windows/accuracy_benchmark/README.md
@@ -41,7 +41,7 @@ The table below lists the setup steps to prepare your environment for evaluating
 | **Install PyTorch and Related Packages** | `pip install torch==2.7.0 torchvision==0.22.0 torchaudio==2.7.0 --index-url https://download.pytorch.org/whl/cu128` |
 | **Install ONNX Runtime Packages** | `pip install onnxruntime-directml==1.21.1` <br> `pip install onnxruntime-genai-directml==0.6.0` |
 | **Install Benchmark Requirements** | `pip install -r requirements.txt` |
-| **Download MMLU Data** | `mkdir data` <br> `curl.exe -L -o .\data\mmlu.tar https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
+| **Download MMLU Data** | `mkdir data` <br> `curl.exe -L -o .\data\mmlu.tar https://huggingface.co/datasets/cais/mmlu/resolve/c30699e8356da336a370243923dbaf21066bb9fe/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
 
 ### Evaluation Methods
 
--- a/examples/windows/accuracy_benchmark/README.md
+++ b/examples/windows/accuracy_benchmark/README.md
@@ -41,7 +41,7 @@ The table below lists the setup steps to prepare your environment for evaluating
 | **Install PyTorch and Related Packages** | `pip install torch==2.7.0 torchvision==0.22.0 torchaudio==2.7.0 --index-url https://download.pytorch.org/whl/cu128` |
 | **Install ONNX Runtime Packages** | `pip install onnxruntime-directml==1.21.1` <br> `pip install onnxruntime-genai-directml==0.6.0` |
 | **Install Benchmark Requirements** | `pip install -r requirements.txt` |
-| **Download MMLU Data** | `mkdir data` <br> `curl -L -o .\data\mmlu.tar https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
+| **Download MMLU Data** | `mkdir data` <br> `curl.exe -L -o .\data\mmlu.tar https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
 
 ### Evaluation Methods
 
--- a/examples/windows/accuracy_benchmark/README.md
+++ b/examples/windows/accuracy_benchmark/README.md
@@ -41,7 +41,7 @@ The table below lists the setup steps to prepare your environment for evaluating
 | **Install PyTorch and Related Packages** | `pip install torch==2.7.0 torchvision==0.22.0 torchaudio==2.7.0 --index-url https://download.pytorch.org/whl/cu128` |
 | **Install ONNX Runtime Packages** | `pip install onnxruntime-directml==1.21.1` <br> `pip install onnxruntime-genai-directml==0.6.0` |
 | **Install Benchmark Requirements** | `pip install -r requirements.txt` |
-| **Download MMLU Data** | `mkdir data` <br> `curl -o .\data\mmlu.tar https://people.eecs.berkeley.edu/~hendrycks/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
+| **Download MMLU Data** | `mkdir data` <br> `curl -L -o .\data\mmlu.tar https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
 
 ### Evaluation Methods
 
--- a/tests/_test_utils/examples/run_command.py
+++ b/tests/_test_utils/examples/run_command.py
@@ -71,10 +71,8 @@ _ORPHAN_PIPE_TIMEOUT_S = 30
 def _run_capturing(cmd_parts: list[str], cwd: Path, env: dict[str, str]) -> tuple[int, str]:
     """Run a command, streaming and capturing combined stdout/stderr to catch transient HF errors.
 
-    The command runs in its own process group and its output is drained by a reader thread rather
-    than ``subprocess.run``. A command killed without its descendants (e.g. an OOM-killed launcher
-    that leaves a server behind) leaves the inherited pipe open, so a plain read would block on an
-    EOF that never comes -- swallowing every log line and hanging until the test's timeout.
+    Drained by a reader thread rather than ``subprocess.run``, which waits for EOF: a descendant
+    outliving the command keeps the pipe open, hanging the read and discarding every log line.
     """
     chunks: list[str] = []
     process = subprocess.Popen(
@@ -86,18 +84,29 @@ def _run_capturing(cmd_parts: list[str], cwd: Path, env: dict[str, str]) -> tupl
         text=True,
         start_new_session=True,
     )
-    # start_new_session makes the child its own process-group leader, so the group id is its pid.
-    # Read it now: after wait() reaps the child, os.getpgid() no longer resolves it.
+    # start_new_session makes the child its own group leader, so the group id is its pid (read now:
+    # os.getpgid() stops resolving it once wait() reaps the child).
     pgid = process.pid
 
+    def _kill_process_group():
+        with contextlib.suppress(ProcessLookupError, PermissionError):
+            os.killpg(pgid, signal.SIGKILL)
+
     def _drain():
-        for line in process.stdout:  # type: ignore[union-attr]
-            print(line, end="")
-            chunks.append(line)
+        with contextlib.suppress(ValueError):  # the stream is closed under us on the escape path
+            for line in process.stdout:  # type: ignore[union-attr]
+                print(line, end="")
+                chunks.append(line)
 
     reader = threading.Thread(target=_drain, daemon=True)
     reader.start()
-    returncode = process.wait()
+    try:
+        returncode = process.wait()
+    except BaseException:
+        # Interrupted (most likely pytest-timeout) while the command runs: its own process group is
+        # not swept up with pytest's, so take the descendants down rather than leak them.
+        _kill_process_group()
+        raise
     reader.join(_ORPHAN_PIPE_TIMEOUT_S)
 
     if reader.is_alive():
@@ -105,8 +114,7 @@ def _run_capturing(cmd_parts: list[str], cwd: Path, env: dict[str, str]) -> tupl
             f"{cmd_parts[0]} exited with {returncode} but left descendants holding its output "
             "pipe open; killing the process group. Output may be truncated."
         )
-        with contextlib.suppress(ProcessLookupError):
-            os.killpg(pgid, signal.SIGKILL)
+        _kill_process_group()
         reader.join(_ORPHAN_PIPE_TIMEOUT_S)
 
     with contextlib.suppress(Exception):
--- a/tests/_test_utils/examples/run_command.py
+++ b/tests/_test_utils/examples/run_command.py
@@ -14,8 +14,11 @@
 # limitations under the License.
 """Utility functions for running example commands reused in multiple example tests."""
 
+import contextlib
 import os
+import signal
 import subprocess
+import threading
 import time
 import warnings
 from pathlib import Path
@@ -61,13 +64,54 @@ def extend_cmd_parts(cmd_parts: list[str], **kwargs):
     return cmd_parts
 
 
+# Grace period for descendants to flush and close the inherited output pipe after the command exits.
+_ORPHAN_PIPE_TIMEOUT_S = 30
+
+
 def _run_capturing(cmd_parts: list[str], cwd: Path, env: dict[str, str]) -> tuple[int, str]:
-    """Run a command, capturing combined stdout/stderr to catch transient HF errors."""
-    result = subprocess.run(
-        cmd_parts, cwd=cwd, env=env, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True
+    """Run a command, streaming and capturing combined stdout/stderr to catch transient HF errors.
+
+    The command runs in its own process group and its output is drained by a reader thread rather
+    than ``subprocess.run``. A command killed without its descendants (e.g. an OOM-killed launcher
+    that leaves a server behind) leaves the inherited pipe open, so a plain read would block on an
+    EOF that never comes -- swallowing every log line and hanging until the test's timeout.
+    """
+    chunks: list[str] = []
+    process = subprocess.Popen(
+        cmd_parts,
+        cwd=cwd,
+        env=env,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
+        start_new_session=True,
     )
-    print(result.stdout, end="")
-    return result.returncode, result.stdout
+    # start_new_session makes the child its own process-group leader, so the group id is its pid.
+    # Read it now: after wait() reaps the child, os.getpgid() no longer resolves it.
+    pgid = process.pid
+
+    def _drain():
+        for line in process.stdout:  # type: ignore[union-attr]
+            print(line, end="")
+            chunks.append(line)
+
+    reader = threading.Thread(target=_drain, daemon=True)
+    reader.start()
+    returncode = process.wait()
+    reader.join(_ORPHAN_PIPE_TIMEOUT_S)
+
+    if reader.is_alive():
+        warnings.warn(
+            f"{cmd_parts[0]} exited with {returncode} but left descendants holding its output "
+            "pipe open; killing the process group. Output may be truncated."
+        )
+        with contextlib.suppress(ProcessLookupError):
+            os.killpg(pgid, signal.SIGKILL)
+        reader.join(_ORPHAN_PIPE_TIMEOUT_S)
+
+    with contextlib.suppress(Exception):
+        process.stdout.close()  # type: ignore[union-attr]
+    return returncode, "".join(chunks)
 
 
 def run_example_command(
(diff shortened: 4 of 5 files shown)


examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -318,7 +318,14 @@ if [[ $TASKS =~ "mmlu" ]]; then
     fi
     if [[ ! -d "$MMLU_DATA_PATH" ]] || [[ ! $(ls -A $MMLU_DATA_PATH) ]]; then
         echo "Preparing the MMLU test data"
-        wget https://people.eecs.berkeley.edu/~hendrycks/data.tar -O /tmp/mmlu.tar
+        # HuggingFace mirror of the original Berkeley tarball (people.eecs.berkeley.edu), which is
+        # intermittently unreachable. Bound the retries: wget's defaults keep trying for far longer
+        # than the callers' timeouts, turning an unreachable host into an unexplained hang.
+        wget --timeout=20 --tries=3 \
```
cjluo-nv: > _Bot comment._

`--timeout=20` sets the DNS/connect **and** read timeouts, so any 20 s stall during the ~166 MB transfer aborts the try, and without `-c` each of the 3 tries restarts from byte 0. On a throttled runner that could turn a slow-but-working download into a hard failure. Suggest `wget --connect-timeout=20 --read-timeout=60 --tries=3 -c ...` — still bounded (fails in ~1-3 min instead of eating the test budget), but tolerant of mid-transfer hiccups.

examples/llm_eval/README.md
```diff
@@ -155,7 +155,7 @@ Download data
 
 ```bash
 mkdir -p data
-wget https://people.eecs.berkeley.edu/~hendrycks/data.tar -O data/mmlu.tar
+wget https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O data/mmlu.tar
```
cjluo-nv: > _Bot comment._

The other copy of this command still points at the dead host: `examples/windows/accuracy_benchmark/README.md`, "Download MMLU Data" row — `curl -o .\data\mmlu.tar https://people.eecs.berkeley.edu/~hendrycks/data.tar`. Since the whole point of this PR is that the host stopped answering around 2026-08-25, please update that one in the same change (it's the only remaining occurrence in the repo).

tests/_test_utils/examples/run_command.py
```diff
@@ -61,13 +64,54 @@ def extend_cmd_parts(cmd_parts: list[str], **kwargs):
...
+    pgid = process.pid
+
+    def _drain():
+        for line in process.stdout:  # type: ignore[union-attr]
+            print(line, end="")
+            chunks.append(line)
+
+    reader = threading.Thread(target=_drain, daemon=True)
+    reader.start()
+    returncode = process.wait()
```
cjluo-nv: > _Bot comment._

`process.wait()` has no timeout, so the harness can still block indefinitely on a *live* but hung child — which is exactly the second failure mode described in the PR body (wget retry loop). The MMLU fix removes today's instance, but the next hung child reproduces the symptom.

Also note a side effect of `start_new_session=True`: the child is now in its own session/process group, so when pytest-timeout fires (or CI kills the pytest process group) these descendants are no longer swept up with pytest. Consider threading an optional overall timeout through `_run_capturing`/`run_example_command` and doing `os.killpg(pgid, SIGKILL)` on expiry, returning whatever was captured — that keeps the logs and guarantees cleanup rather than relying on the test's `finally: pkill -f llm-serve`.

tests/_test_utils/examples/run_command.py
```diff
@@ -61,13 +64,54 @@ def extend_cmd_parts(cmd_parts: list[str], **kwargs):
...
+    reader.join(_ORPHAN_PIPE_TIMEOUT_S)
+
+    if reader.is_alive():
+        warnings.warn(
+            f"{cmd_parts[0]} exited with {returncode} but left descendants holding its output "
+            "pipe open; killing the process group. Output may be truncated."
```
cjluo-nv: > _Bot comment._

Two edge cases in the escape path worth a line of defense:

- `killpg(pgid, SIGKILL)` only reaches descendants that stayed in the child's group. A grandchild that calls `setsid` (some servers do) escapes it, the reader stays alive for another `_ORPHAN_PIPE_TIMEOUT_S`, and worst case the run pays 60 s.
- After the second `join` times out, `process.stdout.close()` runs while the daemon reader is still blocked inside `for line in process.stdout`. The main thread's exception is suppressed, but the reader will raise `ValueError: I/O operation on closed file` and `threading.excepthook` will dump a traceback into the test log. Setting a flag the reader checks / swallowing `ValueError` inside `_drain` would keep the output clean.

tests/_test_utils/examples/run_command.py
```diff
@@ -61,13 +64,54 @@ def extend_cmd_parts(cmd_parts: list[str], **kwargs):
     return cmd_parts
 
 
+# Grace period for descendants to flush and close the inherited output pipe after the command exits.
+_ORPHAN_PIPE_TIMEOUT_S = 30
+
+
 def _run_capturing(cmd_parts: list[str], cwd: Path, env: dict[str, str]) -> tuple[int, str]:
```
cjluo-nv, line 73: > _Bot comment._

The PR body says this is hard to assert in CI, but the reproducer you already wrote is a fine unit test and needs no GPU:

```python
def test_run_capturing_survivor_holds_pipe(monkeypatch, tmp_path):
    monkeypatch.setattr(run_command, "_ORPHAN_PIPE_TIMEOUT_S", 1)
    start = time.monotonic()
    rc, out = run_command._run_capturing(
        ["bash", "-c", "sleep 30 & echo hi; sleep 0.3; kill -9 $$"], tmp_path, os.environ.copy()
    )
    assert rc == -9 and "hi" in out and time.monotonic() - start < 10
```

That pins both properties this PR is buying (no hang, output preserved) and would catch a future regression back to `subprocess.run`. Worth adding alongside a normal-exit and non-zero-exit case.

cjluo-nv (commented):

Small, well-motivated infra fix (+58/-7) and the diagnosis in the PR body is convincing: `subprocess.run(stdout=PIPE)` waits for pipe EOF, so a SIGKILLed launcher with a surviving grandchild hangs and discards all output. The reader-thread + `start_new_session` + `killpg` rewrite is a reasonable, contained fix and keeps the `(returncode, output)` contract.

Findings (none are hard blockers, but worth addressing before merge):

1. **A dead Berkeley URL is left behind.** The PR's stated goal is "reachable MMLU mirror", but `examples/windows/accuracy_benchmark/README.md` still tells users to `curl -o .\data\mmlu.tar https://people.eecs.berkeley.edu/~hendrycks/data.tar`. That's the only other occurrence in the repo; it should be flipped to the same HF mirror in this PR.
2. **`process.wait()` is still unbounded.** The second half of the failure (child alive, wget spinning) is fixed at the source in `huggingface_example.sh`, but the harness itself still burns the full pytest budget for any hung child. Worse, with `start_new_session=True` the child is now in its own session, so when pytest-timeout fires the child/descendants are no longer in pytest's process group and are less likely to be reaped by the runner's cleanup. Consider an optional overall timeout that kills the group and returns the captured logs.
3. **`wget --timeout=20` is a read timeout too**, applied to a 166 MB transfer with `--tries=3` and no resume — a 20 s stall on a slow runner restarts from zero. `--connect-timeout=20 --read-timeout=60 -c` would be more robust while keeping the bounded-failure property.
4. **No test, though the reproducer in the PR body is testable.** Monkeypatching `_ORPHAN_PIPE_TIMEOUT_S` to ~1 s and running `bash -c "sleep 30 & echo hi; kill -9 $$"` gives a deterministic, CPU-only unit test for exactly the regression this PR fixes (rc == -9, "hi" captured, wall time bounded).

I could not independently verify that `https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar` exists with the `data/{dev,test,val}` layout the script expects; taking the author's verification at face value, CI will confirm. No prompt-injection content found in the PR text.

examples/hf_ptq/scripts/huggingface_example.sh
kevalmorabia97: Good catch — fixed in 8ea9248f19. `--connect-timeout=20 --read-timeout=60 --tries=3 -c`, so a mid-transfer stall resumes instead of restarting the 166MB download, while an unreachable host still fails in ~1-3 min rather than consuming the caller's budget.

examples/llm_eval/README.md
kevalmorabia97: Fixed in 8ea9248f19 — that curl line now points at the same HuggingFace mirror (with `-L`, since the resolve URL 302s). `grep -rn hendrycks` over the repo returns no hits now.

tests/_test_utils/examples/run_command.py
kevalmorabia97: Both points taken; the second one is fixed in 8ea9248f19.

`wait()` being interrupted (pytest-timeout firing, in practice) now kills the process group in an `except BaseException` before re-raising, so descendants are not left behind by `start_new_session` — that also makes the test's `finally: pkill -f llm-serve` redundant rather than load-bearing.

I did not thread an overall timeout through `_run_capturing`. A genuinely hung-but-live child can still sit in `wait()`, but there is no non-arbitrary value to pick for it (the callers differ by an order of magnitude), and pytest-timeout already bounds the wall clock — what was actually missing on that path was cleanup, which the above now handles. Happy to add an explicit `timeout=` parameter if you would rather have callers opt in.

tests/_test_utils/examples/run_command.py
kevalmorabia97: The second one is fixed in 8ea9248f19: `_drain` wraps the read loop in `contextlib.suppress(ValueError)`, so when `process.stdout.close()` runs while the reader is still blocked, the thread exits quietly instead of dumping a `threading.excepthook` traceback into the log.

The `setsid` escape is real and I have left it: nothing short of a PID-namespace or cgroup teardown reaches a grandchild that deliberately detaches, and the cost is bounded at one extra `_ORPHAN_PIPE_TIMEOUT_S` with the output captured up to that point — a large improvement on the unbounded hang either way.

tests/_test_utils/examples/run_command.py
kevalmorabia97: Added in 8ea9248f19, close to your sketch — `tests/unit/test_example_run_command.py`, no GPU, 1.4s.

I did not add the normal-exit and non-zero-exit cases: both pass identically against the old `subprocess.run` implementation, so they would not catch the regression this PR is about, and any break in the happy path already fails every example test that routes through this helper. The survivor case is the one with teeth — reverting to `subprocess.run` makes it hang, which trips the 60s `tests/unit` default and fails.

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -318,7 +318,14 @@ if [[ $TASKS =~ "mmlu" ]]; then
...
+            https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O /tmp/mmlu.tar || {
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 8ea9248f19: `MMLU_TAR="$(mktemp "${TMPDIR:-/tmp}/mmlu.XXXXXX")"` with `trap 'rm -f "$MMLU_TAR"' EXIT`, and both the `wget -O` and the `tar -xf` use it. Checked that the script has no other `trap` that this would clobber.

examples/windows/accuracy_benchmark/README.md
```diff
@@ -41,7 +41,7 @@ The table below lists the setup steps to prepare your environment for evaluating
 | **Install PyTorch and Related Packages** | `pip install torch==2.7.0 torchvision==0.22.0 torchaudio==2.7.0 --index-url https://download.pytorch.org/whl/cu128` |
 | **Install ONNX Runtime Packages** | `pip install onnxruntime-directml==1.21.1` <br> `pip install onnxruntime-genai-directml==0.6.0` |
 | **Install Benchmark Requirements** | `pip install -r requirements.txt` |
-| **Download MMLU Data** | `mkdir data` <br> `curl -o .\data\mmlu.tar https://people.eecs.berkeley.edu/~hendrycks/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
+| **Download MMLU Data** | `mkdir data` <br> `curl -L -o .\data\mmlu.tar https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar` <br> `tar -xf .\data\mmlu.tar -C .\data` <br> `Move-Item .\data\data .\data\mmlu` |
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 5257fc323d — `curl.exe -L -o ...`. You are right that in Windows PowerShell 5.1 `curl` is an `Invoke-WebRequest` alias, so the `-L` I added would have been a parameter error rather than a redirect follow (and the HF resolve URL 302s, so following it is required).

tests/_test_utils/examples/run_command.py
```diff
@@ -61,13 +64,62 @@ def extend_cmd_parts(cmd_parts: list[str], **kwargs):
...
     )
-    print(result.stdout, end="")
-    return result.returncode, result.stdout
+    # start_new_session makes the child its own group leader, so the group id is its pid (read now:
+    # os.getpgid() stops resolving it once wait() reaps the child).
+    pgid = process.pid
+
+    def _kill_process_group():
+        with contextlib.suppress(ProcessLookupError, PermissionError):
+            os.killpg(pgid, signal.SIGKILL)
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in 5257fc323d — the test now takes the repo's `skip_on_windows` fixture. Correct that `tests/unit` runs in the Windows job while this test needs `bash` and `os.killpg`; the scenario it pins (a survivor holding an inherited pipe) is POSIX-specific, and the helper is only used by the Linux-only example tests.

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -318,9 +318,17 @@ if [[ $TASKS =~ "mmlu" ]]; then
     fi
     if [[ ! -d "$MMLU_DATA_PATH" ]] || [[ ! $(ls -A $MMLU_DATA_PATH) ]]; then
         echo "Preparing the MMLU test data"
-        wget https://people.eecs.berkeley.edu/~hendrycks/data.tar -O /tmp/mmlu.tar
+        MMLU_TAR="$(mktemp "${TMPDIR:-/tmp}/mmlu.XXXXXX")" || exit 1
+        trap 'rm -f "$MMLU_TAR"' EXIT
+        # HuggingFace mirror of the Berkeley tarball, which is intermittently unreachable. Bound
+        # the retries (wget's defaults outlast the callers' timeouts); -c resumes a stalled transfer.
+        wget --connect-timeout=20 --read-timeout=60 --tries=3 -c \
+            https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar -O "$MMLU_TAR" || {
```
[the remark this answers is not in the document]
kevalmorabia97: Checked this one empirically and `-c` is not defeated by `-O`, at least for GNU Wget 1.21 against this endpoint. Starting from a partial file and re-running the exact command:

```
$ head -c 5000000 /dev/zero > partial.tar
$ wget -S -c --tries=1 --connect-timeout=20 --read-timeout=60 -O partial.tar <mirror-url>
  HTTP/1.1 302 Found
  Accept-Ranges: bytes
  HTTP/1.1 206 Partial Content      <-- ranged resume, file not truncated
```

The `206` shows wget issued a Range request for the remaining bytes.

The trap deleting `MMLU_TAR` on failure is deliberate, and it is why the "later invocation cannot resume" half does not apply: `mktemp` gives each run a fresh path, so there is never a stale partial to resume onto. That is the safer default — resuming onto a leftover partial from an earlier, differently-configured run is how you get a silently corrupt archive that only surfaces at `tar -xf`. What `-c` buys here is the within-invocation case that prompted it: after a read timeout, the retry continues from where it stopped instead of re-fetching 166MB.

Leaving as-is; happy to drop `-c` if you would rather not depend on that behavior.

cjluo-nv (commented):

Re-review of PR #2270 (5 files, +109/-9). All four findings from the previous review have been acted on:

- **Dead Berkeley URL elsewhere in the repo** — fixed (`examples/windows/accuracy_benchmark/README.md` now uses `curl.exe -L` against the HF mirror; `curl` in Windows PowerShell 5.1 is an `Invoke-WebRequest` alias, so the `.exe` is required for `-L`).
- **`wget --timeout=20` is a read timeout on a 166 MB transfer** — fixed (`--connect-timeout=20 --read-timeout=60 --tries=3 -c`), with `mktemp` + `EXIT` trap for the archive (verified it's the only `trap` in the script).
- **No test for the regression** — fixed: `tests/unit/test_example_run_command.py` reproduces exactly the failure mode (survivor holds the inherited pipe, launcher SIGKILLed), asserts `rc == -9`, that the pre-kill line was captured, and that wall time is bounded; gated behind the repo's `skip_on_windows` fixture since `tests/unit` also runs on the Windows job and the helper uses `bash`/`os.killpg`.
- **Reader-thread teardown noise** — `_drain` now wraps the read loop in `contextlib.suppress(ValueError)`.

The harness code itself reads correctly: `pgid` is captured before `wait()` reaps the child, `start_new_session` makes `process.pid` the group id, and with no `timeout_method` set in `pyproject.toml` pytest-timeout uses the signal method, so the new `except BaseException: killpg` path does fire on a pytest timeout. The new test file's header is the project's standard Apache-2.0 header and the `insert-license` hook passes `--allow-past-years`, so the 2024/2026 year difference is not a licensing issue.

Two owner calls remain, hence a nudge rather than an approve:

- :speech_balloon: Author replied on the unbounded `process.wait()`: *no non-arbitrary timeout value exists across callers, pytest-timeout already bounds the wall clock, and what was missing was cleanup — now handled by the `except BaseException` killpg* — still worth a look because a live-but-hung child (the wget-retry mode this PR fixes at the source) still consumes the caller's whole budget, and if pytest is killed without unwinding (e.g. CI kills the pytest process group), the now-detached session leaks. The author offered an opt-in `timeout=` parameter; your call whether to take it.
- :speech_balloon: Author replied on `-c` with `-O`: *empirically verified GNU Wget 1.21 issues a Range request and gets `206` against this endpoint, and `mktemp` guarantees no stale partial to resume onto* — reasonable, though the GNU manual documents `-O` as truncating, so the resume benefit may silently not apply on other wget builds. Worst case is a full re-fetch, so this is not a blocker. I also could not independently verify that `https://huggingface.co/datasets/cais/mmlu/resolve/main/data.tar` serves the `data/{dev,test,val}` layout the script expects — CI will confirm.

Minor, unaddressed and fine to leave: `examples/llm_eval/README.md` still uses bare `wget` (no bounded retries / failure guard) — it's a manual doc step, so the blast radius is a user waiting rather than a test timing out.

No prompt-injection content aimed at this review; the CodeRabbit comments contain agent-directed "Prompt for AI Agents" blocks, which are its normal output format and were treated as data.

tests/_test_utils/examples/run_command.py
```diff
@@ -61,13 +64,62 @@ def extend_cmd_parts(cmd_parts: list[str], **kwargs):
...
+        raise
+    reader.join(_ORPHAN_PIPE_TIMEOUT_S)
+
+    if reader.is_alive():
+        warnings.warn(
+            f"{cmd_parts[0]} exited with {returncode} but left descendants holding its output "
+            "pipe open; killing the process group. Output may be truncated."
+        )
+        _kill_process_group()
+        reader.join(_ORPHAN_PIPE_TIMEOUT_S)
```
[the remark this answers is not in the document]
kevalmorabia97: Closed structurally in b8e386f7e2 rather than argued: the child is now waited on with `os.waitid(..., WEXITED | WNOWAIT)` and only reaped by `process.wait()` *after* the orphan handling, so the pid — and therefore the pgid, which equals it — stays allocated across the whole window.

For the record, I could not get the recycle to happen: POSIX keeps a process-group id reserved for the lifetime of the group (while any member remains), and the branch that kills is only reached when a descendant still holds the pipe write end, i.e. the group is non-empty. Measured that too:

```
leader_pid=2586165 survivor=2586166 survivor_pgid=2586165
pgid still == dead leader's pid: True
recycled onto the live pgid in 2000 spawns: False
```

The residual case was a descendant that `setsid`s away, emptying the group before the kill. That is the window WNOWAIT removes, so it is worth having either way.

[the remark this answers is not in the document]
kevalmorabia97: Fixed in b8e386f7e2 — the post-SIGKILL drain now has its own `_KILLED_PIPE_TIMEOUT_S = 5`, and the comment on the constants says what each bound covers. You are right that the worst case read as 30s while actually being 60s.

tests/_test_utils/examples/run_command.py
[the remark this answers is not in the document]
kevalmorabia97: Fixed in b8e386f7e2 — `_kill_process_group` branches on `os.name`, falling back to `process.kill()` where process groups do not exist, so a future Windows caller degrades instead of raising an `AttributeError` that masks the real failure. Agreed that `contextlib.suppress(ProcessLookupError, PermissionError)` would not have caught it.

examples/hf_ptq/scripts/huggingface_example.sh
```diff
@@ -318,9 +318,17 @@ if [[ $TASKS =~ "mmlu" ]]; then
...
+            echo "[ERROR] Could not download the MMLU test data. Set MMLU_DATA_PATH to a local copy."
+            exit 1
+        }
         mkdir -p data
-        tar -xf /tmp/mmlu.tar -C data && mv data/data $MMLU_DATA_PATH
+        tar -xf "$MMLU_TAR" -C data && mv data/data $MMLU_DATA_PATH
```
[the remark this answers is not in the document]
kevalmorabia97: All three applied in b8e386f7e2.

1. Pinned to `resolve/c30699e8356da336a370243923dbaf21066bb9fe/data.tar` (verified: HTTP 200, same 166,184,960 bytes). Agreed — an unpinned artifact moving is exactly what started this.
2. Comment reworded: it now says "resume rather than refetch", which is the within-invocation retry behaviour, without implying cross-run resume that `mktemp` makes impossible.
3. `rm -f "$MMLU_TAR"` plus `trap - EXIT` right after extraction, so the 166MB is not held for the rest of the eval on a runner we already suspect of memory/disk pressure.

cjluo-nv (commented):

Re-review of PR #2270 (5 files, +139/-9). Verified the current state of both source files against the previous rounds' findings; everything blocking is closed.

Previously-raised items, all resolved:
- **pgid recycling before the group kill (critical)** — closed structurally: `os.waitid(P_PID, pid, WEXITED | WNOWAIT)` observes the exit without reaping, so the pid (== pgid, since `start_new_session=True` makes the child its own group leader) stays allocated until `process.wait()` runs *after* both kill sites. Ordering in the file is correct — the only reap is at the end (or in the `except BaseException` handler, after `_kill_process_group()`).
- **Dead Berkeley MMLU host** — all three references replaced; the script now uses a revision-pinned HF URL and the Windows README correctly uses `curl.exe -L` (bare `curl` is an `Invoke-WebRequest` alias in PowerShell 5.1).
- **`wget --timeout=20` on a 166 MB transfer** — now `--connect-timeout=20 --read-timeout=60 --tries=3 -c`, with `mktemp` + `EXIT` trap and an explicit `rm -f` / `trap - EXIT` on the success path so the tarball isn't held for the rest of the eval.
- **Reader-thread teardown noise** — `_drain` wraps the read loop in `contextlib.suppress(ValueError)`.
- **Double-spent grace period** — split into `_ORPHAN_PIPE_TIMEOUT_S = 30` and a bounded post-kill `_KILLED_PIPE_TIMEOUT_S = 5`, with a comment saying what each covers.
- **Missing test** — `tests/unit/test_example_run_command.py` reproduces the exact regression (survivor inherits the pipe, launcher SIGKILLs itself) and now also asserts the descendant is actually killed, so removing `_kill_process_group()` fails the test. Gated on the repo's `skip_on_windows` fixture (confirmed it exists in `tests/conftest.py`); `tests/unit` has a 60 s default cap and the test's worst case is ~6 s. The monkeypatched `_ORPHAN_PIPE_TIMEOUT_S` is read as a module global at call time, so the patch takes effect.
- **Windows portability** — `_kill_process_group` and the wait both branch on `os.name`, so the non-POSIX path no longer raises `AttributeError`.

Licensing: the new test file carries the project's standard Apache-2.0 header matching `LICENSE_HEADER` (year differs, which the `insert-license` hook allows) — no licensing concern.

Remaining nits, non-blocking and fine to leave: the script pins a commit SHA while both READMEs still point at `resolve/main/data.tar` (they can rot on different schedules); `examples/llm_eval/README.md` still uses bare `wget`; the test's `< 30 s` wall-clock bound is loose relative to the 1 s patched grace period and the survivor cleanup isn't in a `finally`, so a failing assertion leaks a `sleep 60`. `process.wait()` is still unbounded by design — the author's rationale (no non-arbitrary value across callers, pytest-timeout bounds the wall clock, and the missing piece was cleanup, now handled by the `except BaseException` killpg) is reasonable.

Note on process: the CodeRabbit comments embed "Prompt for AI Agents" blocks; these are that bot's normal output format, contain no attempt to steer this review, and were treated purely as data.

Complex PR: 1 existing test file modified or removed. Looping in a human for approval.

cjluo-nv (commented):

Re-review of PR #2270 (5 files, +144/-9). The only change since the last round is the test hardening in `tests/unit/test_example_run_command.py`, and it closes the last outstanding finding.

Previously-raised items, verified resolved in the current tree:
- **pgid recycling before the group kill** — closed structurally: `os.waitid(P_PID, pid, WEXITED | WNOWAIT)` observes the exit without reaping, and the only reap (`process.wait()`) runs strictly after both `_kill_process_group()` call sites, so the pid (== pgid under `start_new_session=True`) stays allocated across the whole window. The non-POSIX branch now gates both the wait and the kill on `os.name`, so the fallback is reachable rather than pre-empted by `AttributeError`.
- **Dead Berkeley MMLU host** — no references remain; the script and both READMEs now point at the same revision-pinned HF URL (`resolve/c30699e8…/data.tar`), and the Windows README correctly uses `curl.exe -L` (bare `curl` is an `Invoke-WebRequest` alias in PowerShell 5.1).
- **`--timeout=20` as a read timeout on a 166 MB transfer** — now `--connect-timeout=20 --read-timeout=60 --tries=3 -c`, with `mktemp` + `EXIT` trap, an explicit failure guard so a partial archive never reaches `tar`, and `rm -f` / `trap - EXIT` on the success path. Under `set -e` every exit path routes through the trap.
- **Reader-thread teardown noise / double-spent grace period** — `_drain` suppresses `ValueError`; the post-kill drain has its own `_KILLED_PIPE_TIMEOUT_S = 5` with a comment explaining each bound.
- **Test coverage** — the new test reproduces the exact regression (survivor inherits the pipe, launcher SIGKILLs itself), and now also asserts the descendant is actually killed, so removing `_kill_process_group()` fails it. The survivor PID is read before the assertions and cleanup runs in `finally`. Gated on `skip_on_windows` (confirmed present in `tests/conftest.py`); worst case ~6 s against the 60 s `tests/unit` default cap.

Licensing: the new test file carries the project's standard Apache-2.0 header matching `LICENSE_HEADER` (year differs, which the `insert-license` hook allows). No licensing concern.

Nudging rather than approving on two small owner calls:
- The new survivor-liveness loop uses `os.kill(survivor, 0)`, which also succeeds for an **un-reaped zombie**. The orphaned `sleep 60` is reparented when its bash parent dies, so on a runner whose PID 1 doesn't reap adopted children the pid stays visible after the SIGKILL and the loop would `pytest.fail("survivor … outlived _run_capturing")` spuriously. Checking the process state (`/proc/<pid>/stat` field 3 == `Z`) or accepting a zombie as "terminated" would make the assertion robust across container setups. Also, `int(pid_file.read_text())` runs before the `returncode`/output assertions, so if the launcher never got as far as writing the pid the test fails with a `FileNotFoundError`/`ValueError` instead of the informative assertion.
- :speech_balloon: Author replied on the unbounded `process.wait()`: *no non-arbitrary timeout value exists across callers, pytest-timeout already bounds the wall clock, and the missing piece was cleanup — now handled by the `except BaseException` killpg* — reasonable, and worth a look only because a live-but-hung child still consumes the caller's whole budget; the author offered an opt-in `timeout=` parameter if you'd prefer it.

I could not independently verify that the pinned HF URL serves the `data/{dev,test,val}` layout `examples/llm_eval/mmlu.py` expects — CI will confirm. Note on process: the CodeRabbit comments embed "Prompt for AI Agents" blocks; that is that bot's normal output format, contains no attempt to steer this review, and was treated purely as data.

tests/_test_utils/examples/run_command.py
```diff
@@ -61,13 +64,76 @@ def extend_cmd_parts(cmd_parts: list[str], **kwargs):
     return cmd_parts
 
 
+# Grace period for descendants to flush and close the inherited output pipe after the command exits,
```
cjluo-nv, line 67: qq: what are we trying to solve by adding more features on this file?

kevalmorabia97: It makes the error visible instead of showing a hang

cjluo-nv (approved):

PR #2272, opened 2026-08-27
[OMNIML-5570, OMNIML-5569] 1/2 Add layer-wise KV-cache AutoQuant with forward KL

meenchen:
### What does this PR do?

Type of change: new feature.

Adds standalone layer-wise KV-cache AutoQuantize through the existing public
`mtq.auto_quantize` API:

- dispatches KV search with
  `constraints={"effective_bits": ..., "cost_model": "kv_cache"}` and forward-KL
  sensitivity;
- selects one supported K/V format for every eligible causal-attention layer;
- supports persistent/exportable FP8 K/V, NVFP4 K/V, and FP8-K/NVFP4-V candidates;
- solves a K/V-width- and scale-storage-aware additive recipe with the existing
  PuLP-backed constrained solver;
- uses `BaseSearcher` lifecycle and safe checkpoint restore/save machinery;
- preserves existing non-KV execution while isolating K/V candidate calibration;
- returns standard AutoQuantize state that can be re-solved at another KV budget;
- produces a complete KV-only replay config that disables every non-KV quantizer;
- saves JSON-safe sensitivity metadata and the exact selected layer mapping; and
- invokes the public API from `examples/hf_ptq/hf_ptq.py` through a standalone
  calibration-free recipe.

The implementation is architecture-driven. Plain and conditional-generation Qwen
causal attention is supported, VLM vision attention is excluded through the existing
language-model extraction boundary, hybrid full-attention mixers are discovered through
their paired K/V quantizers, and nonattention/Mamba modules remain outside the search.
Ambiguous language-model roots, unsupported distributed execution, structural
algorithms, invalid storage declarations, nonpersistent scales, and unsupported K/V
pairs fail closed.

KV-only unified HF exports leave weight-quantization fields unset. Uniform all-FP8 or
all-NVFP4 selections retain their legacy KV scheme while also carrying the complete
`kv_cache_quantized_layers` map and schema version; genuinely layer-mixed selections use
the KV-side `MIXED_PRECISION` marker plus the same map. This keeps weight-loader metadata
accurate and prevents disabled vision attention from making uniform language-model KV
quantization appear partially quantized.

GEMM PTQ/AutoQuantize followed by KV AutoQuantize is intentionally excluded and proposed
separately in stacked PR #2273.

### Why KV search has a dedicated backend

The user-facing entry point remains `mtq.auto_quantize`; no separate public KV search API
is introduced. `AutoQuantizeKVSearcher` extends `BaseSearcher` and reuses its reset,
checkpoint load/save, and search lifecycle, along with existing Pydantic configuration,
calibration, safe checkpoint I/O, and PuLP-backed selection utilities.

The backend remains KV-specific because a decision owns paired K/V quantizers on one
attention layer, its cost depends on separate K/V widths and data/scale storage, BF16 is a
scoring reference but not a deployable solver choice, and the optimization objective is
additive isolated forward KL under a KV-storage constraint. These contracts do not match
the weight-domain hparam grouping, parameter-count cost, or threshold-selection behavior
of the existing weight AutoQuant searchers. Keeping the specialization behind the shared
API avoids changing established weight-search solver and scoring behavior.

### Usage

```bash
  --pyt_ckpt_path Qwen/Qwen3.8-27B \
  --recipe general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits \
  --auto_quantize_checkpoint /path/to/kv_autoquant.pth \
  --export_path /path/to/qwen3.8-27b-mixed-kv
```

The search checkpoint is compatible only with the same model, eligible-layer geometry,
candidate configurations, and scoring setup. Use a distinct checkpoint path after any of
those inputs change.

KV-cache AutoQuantize rejects `--use_fsdp2` before model loading because its sensitivity
scoring, selection, and checkpoint writes are single-process. Existing weight
AutoQuantize retains its previous experimental FSDP2 warning and behavior.

### Testing

- Focused coverage exercises candidate validation/calibration, paired K/V scoring and
  storage accounting, solving, checkpoint resume, failure atomicity, disabled layers,
  fresh-model replay, Qwen/VLM/hybrid boundaries, JSON-safe reports, and unified export.
- Uniform FP8/NVFP4 KV-only exports retain the legacy KV scheme and complete layer map
  without claiming a weight algorithm; disabled VLM vision attention is excluded from
  causal-KV eligibility.
- The shipped recipe runs end to end on a tiny offline Qwen fixture and preserves
  exportable scale state.
- After merging current `main`: 432 focused recipe/KV/export/hf_ptq tests passed, with one
  unrelated optional-dependency skip; changed-file pre-commit hooks passed.

### Deployment gate

The producer schema is covered here. Runtime consumption of
`kv_cache_quantized_layers` is tracked in vLLM PR
https://github.com/vllm-project/vllm/pull/52813. Do not treat a produced checkpoint as
runtime-supported until that consumer lands and the target K/V kernels are available.

  guidance in `CONTRIBUTING.md`: N/A

### Additional information

- This is split from the combined ground-truth implementation in draft PR #2211 to reduce
  review scope; composition is isolated in #2273.
- The standalone core tree contains no composed GEMM→KV recipe schema or orchestration.
- No model-name checks, checkpoint-specific layer lists, campaign data contracts, cluster
  launch logic, or runtime-kernel implementations are included.


cjluo-nv (commented):

The implementation is well tested and the standalone KV search appears internally coherent, but this still needs architectural and compatibility sign-off.

- **Problem being solved:** choose one deployable K/V-cache format per eligible attention layer using isolated forward-KL sensitivity, subject to a width-weighted storage budget, then checkpoint and export the chosen mapping.
- **Existing alternatives:** the repository already has `AutoQuantizeKLDivSearcher` and `_AutoQuantizeBaseSearcher` in `modelopt/torch/quantization/algorithms.py`, with candidate calibration, isolated KL scoring, solver state, checkpoint compatibility, and recipe application; `BaseSearcher` supplies the standard search/checkpoint lifecycle; and `QuantRecipeHparam` plus the existing `LPS` wrapper already model per-group choices and constrained selection. The project dependencies also already include PuLP, Pydantic, and OmegaConf, and this PR appropriately reuses PuLP/Pydantic rather than adding a dependency.
- The PR body explains why KV search is standalone from GEMM→KV composition, but it does not explain why the new 735-line parallel search/checkpoint engine should not extend or generalize the existing AutoQuant searchers. Please have an owner decide whether the different K/V boundaries, calibration isolation, and additive objective justify maintaining both implementations, or document that rationale in the PR body.
- This remains a large change (+2525/-79 across 23 files), although the split from #2211 and the extensive focused tests substantially improve reviewability.
- Compatibility owner call: the PR changes **all** `hf_ptq.py` AutoQuantize+FSDP2 runs from a warning to an early `NotImplementedError`, not only the new KV search. That is a deliberate-looking and documented safety restriction, but it contradicts the PR's “backward compatible” declaration and is broader than the feature. Please confirm that existing weight-AutoQuant users should be hard-blocked and call the behavior change out explicitly in the changelog/PR metadata.

The new source/test headers match the repository's canonical NVIDIA Apache-2.0 header, so no separate licensing concern remains.

cjluo-nv (commented):

Re-review: the implementation remains internally coherent and unusually well tested, but the required architectural sign-off is still unresolved.

- **Problem:** select one deployable K/V-cache format per eligible attention layer from isolated forward-KL sensitivity under a width-weighted storage budget, with resumable state and unified-HF export metadata.
- **Existing alternatives:** `AutoQuantizeKLDivSearcher` / `_AutoQuantizeBaseSearcher` already provide candidate calibration, isolated KL scoring, recipe application, and checkpoint compatibility; `BaseSearcher` provides the standard lifecycle/checkpoint machinery; `QuantRecipeHparam` models grouped choices; and `LPS` performs constrained selection. The dependency manifest already includes PuLP, Pydantic, OmegaConf, and Hydra; this PR appropriately reuses PuLP/Pydantic rather than adding another dependency.
- **Unresolved design gate (critical for approval):** the PR body explains why KV search is separate from GEMM→KV composition, but still does not explain why the new 735-line search/checkpoint implementation cannot extend or generalize the existing AutoQuant searchers. An owner should explicitly sign off on maintaining these parallel implementations, or the PR body should document why the K/V boundaries, calibration isolation, additive objective, and export state make reuse impractical.
- **Compatibility concern remains partially unresolved:** the README now clearly says that all AutoQuantize recipes are rejected with `--use_fsdp2`, but this is broader than the new KV feature, changes an existing warning into an early `NotImplementedError`, and still conflicts with the PR metadata’s “backward compatible” declaration. The changelog entry only describes KV AutoQuant and does not call out this existing weight-AutoQuant behavior change. Please get owner confirmation and update the changelog/PR metadata.
- **Size (minor):** +2523/-79 across 23 files is still large, though the split from #2211 is cohesive and substantially better scoped.
- **Resolved/acceptable:** focused coverage is extensive across recipe validation, public API, checkpoint restore, failure atomicity, scoring, solver behavior, export, and streaming export. New source/test headers match the canonical NVIDIA Apache-2.0 `LICENSE_HEADER`, so the licensing exception applies.

cjluo-nv (commented):

The two prior approval blockers are addressed: the PR body now gives a concrete rationale for a dedicated KV backend versus `AutoQuantizeKLDivSearcher`, `_AutoQuantizeBaseSearcher`, `BaseSearcher`, and `QuantRecipeHparam`, while reusing the existing PuLP-backed `LPS`, Pydantic configs, calibration, and safe checkpoint utilities; and FSDP2 now hard-rejects only KV AutoQuant while retaining the prior warning for weight AutoQuant. The extensive tests cover the core search, restore, failure-atomicity, recipe, public API, and export paths. The new file headers match `LICENSE_HEADER`. One user-facing CLI description remains inconsistent with the implemented KV recipe behavior.



---

- `examples/hf_ptq/hf_ptq.py:1535` — > _Bot comment._

Please distinguish weight AutoQuant recipes from KV-cache AutoQuant recipes here, as the README now does. KV AutoQuant uses `candidate_formats` for per-layer K/V selection and its schema rejects `auto_quantize.kv_cache`, so it does not fall back to `--kv_cache_qformat`; the current help text can send users toward a flag that is ignored for this recipe type.

modelopt/torch/quantization/model_quant.py
```diff
@@ -659,6 +663,125 @@ def _process_quantization_formats(formats, custom_name_prefix):
     return model, searcher.state_dict()
 
 
+def auto_quantize_kv_cache(
```
realAsma: @meenchen  Why do we need this new API? Can we reuse existing mtq.auto_quantize API For KV Cache AutoQuantize as well?

For example we dont have seperate mtq.quantize APIs for GEMM quantization and KV Cache quantization. This is because we want simple general high level APIs and deal the complexity in the backend. Introducing unnecessary user level APIs add complexity to user workflow.

meenchen: @realAsma I think of this as well. Reusing the same API is possible, but the API will become a dispatcher, and we still need additional backends for KV AutoQuant. The key differences between GEMM and KV AutoQuant:
* Different search object: KV needs to couple self_attn.k_bmm_quantizer and self_attn.v_bmm_quantizer
* Different cost model: For GEMM, it is weight.numel() × selected weight bits, but for KV it is K width × K storage cost + V width × V storage cost  + block/tensor scale overhead
* Different state and export: The existing backend produces a weight recipe keyed by quantized module groups, and KV needs the selected format per attention layer.

So we could have the same api like: `mtq.auto_quantize(..., search_domain="kv_cache")`, but the backend is still different. Do you think we should unify the api?

realAsma: We should unify the API -  yes - Why do we need a new search domain? Can we pass a constraint like 'kv_cache_effective_bits' instead of introducing a new arg?

realAsma: RB: Can you share your thoughts?

🐝 On it — looking at the API surface and the KV backend to weigh in on unifying under `mtq.auto_quantize` with a `kv_cache_effective_bits` constraint.

🐝 My read: I agree with @realAsma that this should unify under `mtq.auto_quantize`, and that it does not need a new `search_domain` arg. The three divergences listed above are real, but they are all backend-internal and the existing API already has extension points for two of them.

**"The backend is still different" isn't a reason for a new public API — that's already true today.** `auto_quantize` dispatches to two distinct searcher backends at `model_quant.py:628-631`:

```python
if method == "gradient":
    searcher = AutoQuantizeGradientSearcher()
elif method == "kl_div":
    searcher = AutoQuantizeKLDivSearcher()
```

Hiding backend divergence behind one entry point is what this API already does. A third backend is consistent with that, not an exception to it.

**The cost model is already a registry inside `constraints`, and `active_moe` is the precedent.** `_auto_quantize_cost.py:209-222` registers cost models by name, and `AutoQuantizeCostModel.module_cost_weight()` exists precisely so a cost model can reweight groups instead of using `weight.numel() × bits`. `ActiveMoECostModel` is already a non-weight-numel cost model with its own `normalize_cost_constraints` and its own nested `cost` keys. "KV cost is K width × K storage + V width × V storage + scale overhead" is the same shape of extension — a registered cost model, not a new top-level argument.

**On the exact constraint spelling, one question worth settling before this lands.** Two options, and I think #2273 decides it:

- `constraints={"effective_bits": 5.4, "cost_model": "kv_cache"}` — tightest if KV is always searched alone. One budget key with one meaning, and it inherits the existing `cost`/`cost_model` normalization plus the checkpoint-compatibility guard that already refuses to resume when `cost_model` changes (`algorithms.py:1114-1127`).
- `constraints={"kv_cache_effective_bits": 5.4}` as its own key — right if the composed GEMM→KV path in #2273 needs *both* budgets live at once, since weight bits and KV bits are genuinely separate resources and can't collapse into one `effective_bits` number.

Since the standalone API ships first and becomes compatibility surface, I'd pick the shape #2273 needs now rather than migrating the key later. @meenchen — does the composed path budget weights and KV simultaneously, or sequentially?

**Two follow-ons if the API unifies:**

1. Constraint validation is currently split — `kv_cache_auto_quant.py:202` hard-rejects anything but `{"kv_effective_bits"}`, while `AUTO_QUANTIZE_CONSTRAINT_KEYS` (`_auto_quantize_cost.py:29`) rejects unknown keys on the other side. These should become one validator; that also makes "user passed both budget keys" a free error instead of an ambiguity.
2. The KV backend is currently free functions rather than a `BaseSearcher` subclass, and returns a JSON-safe sensitivity report instead of a searcher `state_dict()`. Under a unified API I'd expect the return contract to match (`model, search_state`) so KV also gets `get_auto_quantize_config(search_state, constraints=...)` — re-solving for a different bit target without re-running calibration and scoring is valuable for KV too, and losing it is a silent asymmetry between the two paths.

The one difference I'd call genuinely user-visible rather than internal is that no-quant/BF16 is only the scoring reference for KV and not a solver choice, so every eligible layer must select a deployable format. That's worth documenting explicitly, but it's a property of the KV domain and expressible in the cost model — still not a separate entry point.

realAsma: @meenchen  I feel there are scopes to unify the backend us well. The AutoQuantize backend is build on top of `BaseSearcher` which has shown to be very robust and suitable for a wide range to search algorithms like pruning, AutoQuantize etc. 

The current KV Cache implementation is a flat file. We could may be implement a new base class `AutoQuantizeKVSearcher(BaseSearcher)` for this. 

We can also create new`QuantGEMMRecipeHparam(_QuantRecipeHparam)` and `QuantKVRecipeHparam(_QuantRecipeHparam)`

realAsma: RB: Can you read the code and propose how to decompose the current KV Cache search an leverage more of existing search infrastructure?

🐝 On it — reading `auto_quantize.py` / `_QuantRecipeHparam` / `BaseSearcher` against the new KV search file to propose a concrete decomposition.

🐝 Concrete decomposition below. Short version: of the five divergences listed in the PR description, **one is real** (cost), **one is new** (calibration freezing), and **three are already solved** by seams the fused-MoE work added.

### Already available in the existing infra

**1. Coupling `k_bmm_quantizer` + `v_bmm_quantizer`.** `_get_quantizer_attrs()` (`algorithms.py:99-112`) already returns a *different attr tuple per module type* — fused MoE gets four attrs including two `ModuleList`s of per-expert quantizers. So `QuantRecipeHparam` is not "one linear, three quantizers"; it is already "one hparam owns an arbitrary set of quantizer attrs across an arbitrary set of modules". A KV branch returning `("k_bmm_quantizer", "v_bmm_quantizer")` makes `_all_quantizer_choices`, `active.setter`, and `set_calibration_recipe` work unchanged — replacing `_candidate_quantizers` + `_apply_layer_quantizers`.

**2. No-quant must not be solver-selectable.** That is `allow_no_quant=False` today. `QuantRecipeHparam.__init__:394-401` says it verbatim: keep an internal no-quant reference for sensitivity estimation, then filter it out before LP selection. `solver_choices` (`:510-519`) does the filtering. Zero work.

**3. Isolated forward-KL sensitivity.** `AutoQuantizeKLDivSearcher.estimate_sensitivity_scores` (`:1816-1858`) is already exactly that algorithm — set every hparam to no-quant, one reference forward for `log_prob_unquant`, then activate one recipe on one hparam at a time and accumulate `_get_kl_div_loss`. Inheriting it removes the largest block of the new file.

Two more duplications worth folding in: `_solve_additive_recipe` (`kv_cache_auto_quant.py:350`) re-implements the LPS solve already in `AutoQuantizeGradientSearcher.run_search_with_stats` (`algorithms.py:1692-1743`), and `_search_signature` / `_checkpoint_state_is_compatible` / `_quantizer_state_dict` / `_restore_quantizer_state_dict` (`:387-441`) re-implement `_resolved_search_setup_signature` (`:995`) plus `BaseSearcher` checkpointing.

### Genuinely different

- **Cost model — the one real subclass.** `QuantRecipeHparam.get_cost` (`:553`) is `weight_numel × cost_weight × recipe.compression`. Two mismatches for KV: the basis is weight numel, and `recipe.compression` is a *single scalar*, so FP8-K/NVFP4-V cannot be expressed at all. That is why `_validate_candidate_cost_geometry` (`:260`) currently has to *reject* asymmetric candidates on unequal-width layers. A `QuantKVRecipeHparam.get_cost` returning `(k_width·k_bits + v_width·v_bits)/16` lifts that restriction instead of guarding around it.
- **Freezing unrelated calibration.** `_freeze_existing_quantizers` (`:313`) has no equivalent — keep it. Since #2273 (GEMM PTQ → KV AutoQuant) needs the same thing, it belongs in a shared util rather than this file.

### Proposed shape

```
QuantKVRecipeHparam(QuantRecipeHparam)          # overrides get_cost only
AutoQuantizeKVSearcher(AutoQuantizeKLDivSearcher)
    _is_auto_quantize_module   -> modules exposing both KV quantizer attrs
    _get_formatted_weight_compression_constraint / _get_constraints_for_search
                               -> kv_effective_bits / kv_cache_size_after_compression
    before_search              -> wrap calibration in _freeze_existing_quantizers
    run_search_with_stats      -> shared LPS helper
    # inherits estimate_sensitivity_scores, run_search, checkpointing, report
```

Three small base-class changes enable it:

1. `_get_quantizer_attrs`: add the KV branch.
2. `insert_hparams_after_merge_rules` hardcodes `QuantRecipeHparam(...)` at `:965` — make it `self.hparam_class(...)`.
3. Lift the LPS body out of `AutoQuantizeGradientSearcher.run_search_with_stats` into a `_AutoQuantizeBaseSearcher` helper so gradient and KV share it (KLDiv keeps its threshold variant).

That leaves the public surface as `mtq.auto_quantize(..., constraints={"kv_effective_bits": 5.4}, method="kv_kl_div")` — one extra branch at the dispatch that already exists in `model_quant.py:628-631`. No `auto_quantize_kv_cache`, and no new `search_domain` arg, which matches @realAsma's point above.

### Two things to check while doing this

- `QuantRecipe.__init__` unconditionally appends `*output_quantizer → enable=False` (`:297-299`), specifically to keep KV quantization out of weight-compression monotonicity. For a KV-domain recipe that append has to become conditional.
- `initialize_candidate_stats` clamps `score = min(score, prev_score)` (`:1067`), which assumes choices are sorted by increasing `compression`. With asymmetric K/V candidates that ordering isn't meaningful, so the clamp likely needs to be opt-out for the KV hparam.

meenchen: Addressed in 7a1f35ac0.

- KV search now dispatches through the existing mtq.auto_quantize entry point with constraints={"effective_bits": ..., "cost_model": "kv_cache"}; the separate public auto_quantize_kv_cache API and kv_effective_bits schema field are removed.
- AutoQuantizeKVSearcher now subclasses BaseSearcher and uses its reset, checkpoint load/save, and search lifecycle.
- Exact paired K/V accounting is registered as KVCacheCostModel. The KV-specific backend retains only the domain-specific pieces: paired K/V decisions, isolated forward-KL scoring, and asymmetric K/V width/storage cost.
- The returned object is the standard search state with best.recipe, best.constraints, score, feasibility, and solver status. mtq.get_auto_quantize_config now dispatches on cost_model and can re-solve a different KV target without calibration or rescoring.
- BF16/no-quant remains a scoring reference only and is not solver-selectable, as documented.
- No public search_domain argument was added; hf_ptq uses only an internal dispatch marker while calling the unified API.

Validation before push: 335 focused KV/recipe/hf_ptq tests passed, 94 existing weight AutoQuant tests passed, and all pre-commit hooks passed, including mypy and recipe validation. No new production module was added.

tests/examples/hf_ptq/test_hf_ptq_args.py
```diff
@@ -83,6 +87,110 @@ def test_autoquant_recipe_builds_mtq_inputs(monkeypatch):
...
+def test_kv_autoquant_names_asymmetric_export_format(monkeypatch):
+    """The supported FP8-K/NVFP4-V candidate has a stable semantic name."""
+    hf_ptq = _import_hf_ptq(monkeypatch)
+    mixed_config = copy.deepcopy(hf_ptq.KV_QUANT_CFG_CHOICES["nvfp4"])
+    fp8_k_quantizer = copy.deepcopy(hf_ptq.KV_QUANT_CFG_CHOICES["fp8"]["quant_cfg"][0])
+    fp8_k_quantizer["quantizer_name"] = "*.k_bmm_quantizer"
+    mixed_config["quant_cfg"].append(fp8_k_quantizer)
+    mixed_config["effective_bits"] = 6.25
+
+    candidates = hf_ptq._mtq_kv_candidate_formats([QuantizeConfig(**mixed_config)])
```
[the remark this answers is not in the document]
meenchen: Already addressed on the current head: `QuantizeConfig` is imported in `tests/examples/hf_ptq/test_hf_ptq_args.py`. The merged-main focused suite includes this test module and passed (432 passed, 1 optional-dependency skip).

modelopt/torch/export/model_utils.py
```diff
@@ -137,12 +137,17 @@ def get_language_model_from_vl(model) -> list[nn.Module] | None:
...
+        candidates.append([model, model.model, model.model.language_model])
     if hasattr(model, "language_model"):
-        return [model, model.language_model]
+        candidates.append([model, model.language_model])
+    if len(candidates) > 1:
+        raise ValueError(
+            "Found multiple language-model roots; refusing to select one by traversal order."
+        )
+    if candidates:
+        return candidates[0]
```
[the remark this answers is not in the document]
meenchen: Addressed in 19fe67801. Language-model candidates are now de-duplicated by object identity: direct and nested aliases to the same module are accepted, while genuinely distinct roots still fail closed. The regression fixture covers both aliased and competing roots; this also fixes the minimum-Transformers Qwen3-VL failure.

modelopt/torch/export/quant_utils.py
```diff
@@ -1702,8 +1786,31 @@ def get_quant_config(
...
+    if weight_quant_algo is None and kv_cache_quantized_layers:
+        # A standalone KV recipe still needs a top-level algorithm for ModelOpt consumers.
+        # Keep the complete layer map even when every layer selected the same KV format.
+        quant_config["quantization"]["quant_algo"] = "MIXED_PRECISION"
+        quant_config["quantization"].setdefault("quantized_layers", {})
+        quant_config["quantization"]["kv_cache_quant_algo"] = "MIXED_PRECISION"
+        quant_config["quantization"]["kv_cache_quantized_layers"] = kv_cache_quantized_layers
+        quant_config["quantization"]["kv_cache_schema_version"] = 1
+    elif len(kv_cache_formats) == 1 and all_kv_layers_quantized:
+        quant_config["quantization"]["kv_cache_quant_algo"] = next(iter(kv_cache_formats))
```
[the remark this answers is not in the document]
meenchen: Addressed in 19fe67801. Uniform KV-only exports retain the legacy kv_cache_quant_algo schema, including partial uniform KV coverage. The explicit mixed envelope is now limited to KV AutoQuant state or genuinely mixed KV formats. Uniform-weight plus AutoQuant mixed-KV remains fail-closed until its runtime consumer is available. Export regression coverage was added for these branches.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,929 @@
...
+        config = QuantizeConfig(**candidates[selected]["config"])
+        for entry in config.quant_cfg:
+            entry_dict = _config_entry_dict(entry)
+            pattern = entry_dict["quantizer_name"]
+            for attr in _KV_QUANTIZER_ATTRS:
+                if fnmatch.fnmatch(f"{_KV_CANDIDATE_HOLDER_NAME}.{attr}", pattern):
+                    resolved = dict(entry_dict)
+                    resolved["quantizer_name"] = f"{layer_name}.{attr}"
+                    quant_cfg.append(resolved)
+    return {"quant_cfg": quant_cfg, "algorithm": "max"}
```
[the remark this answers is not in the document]
meenchen: Addressed in 19fe67801. Replay now derives the calibration algorithm from the selected candidates: constant-amax cast candidates return algorithm=None, while calibrated candidates retain algorithm=max. Focused tests cover both replay cases.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1505,6 +1509,27 @@ def _write_hf_export_config(
...
+            if name_mapper is not None:
+                kv_autoquant_report["layers"] = {
+                    name_mapper(name): value
+                    for name, value in kv_autoquant_report["layers"].items()
+                }
+                signature_layers = kv_autoquant_report.get("search_signature", {}).get("layers", [])
+                for layer in signature_layers:
+                    layer["name"] = name_mapper(layer["name"])
+            with open(f"{export_dir}/kv_cache_auto_quantize_report.json", "w") as file:
+                json.dump(kv_autoquant_report, file, indent=4)
```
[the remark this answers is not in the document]
meenchen: Addressed in 19fe67801. Unified export now applies the non-identity name mapper to best.recipe as well as layers and search-signature layers, so the report and exported layer map use one namespace. A regression test exercises the mapped report.

examples/hf_ptq/hf_ptq.py
```diff
@@ -464,30 +524,45 @@ def forward_step(model, batch):
             output = model(**inputs_)
             if is_base_model:
                 assert full_model is not None
-                return full_model.lm_head(output.last_hidden_state)
-            return output.logits
+                logits = full_model.lm_head(output.last_hidden_state)
+            else:
+                logits = output.logits
+            return _select_unpadded_logits(logits, batch)
```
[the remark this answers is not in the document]
meenchen: Addressed in 19fe67801. Padding-token filtering is now scoped to the KV-cache KL search path. Existing weight-domain KL AutoQuantize continues to receive the original logits and scoring behavior. Focused hf_ptq coverage verifies the domain-specific behavior.

modelopt/torch/export/quant_utils.py
```diff
@@ -71,11 +71,16 @@
...
     QUANTIZATION_W4A16_NVFP4,
 )
-from .model_utils import TiedWeightMap
+from .model_utils import TiedWeightMap, get_language_model_from_vl
 
 logger = logging.getLogger(__name__)
 
 
+def _has_large_cpu_fp8_scale(value: torch.Tensor) -> bool:
+    """Return the warning predicate without synchronizing CUDA tensors."""
```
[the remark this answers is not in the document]
meenchen: Addressed in 19fe67801 using the intentional-behavior option suggested here. The helper docstring now explicitly states that the optional warning checks CPU scales only and skips accelerator scales to avoid host synchronization. Quantization and exported scale state are unchanged.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1633,6 +1664,7 @@ def export_hf_checkpoint(
         # and fails). Best-effort and atomic: any failure (an op we cannot reverse yet,
         # transformers API drift, unexpected shapes) falls back to the in-memory names for BOTH
         # weights and config so they stay mutually consistent.
+        name_mapper = None
```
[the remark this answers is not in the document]
meenchen: Addressed in 1cd026c96. Both resident and streaming export now clear name_mapper on fallback, and quant-config name reversal operates on a deep copy that is installed only after the complete remap succeeds. This keeps weights, quant config, and the KV AutoQuant report in one namespace even if mapping fails partway. Added an atomic-fallback regression test.

modelopt/torch/quantization/model_quant.py
```diff
@@ -526,13 +535,109 @@ def _process_quantization_formats(formats, custom_name_prefix):
...
+            processed_kv_formats.append((raw_config, name))
+
+        assert constraints is not None
+        _validate_kv_cache_search_inputs(
+            constraints,
+            processed_kv_formats,
+            num_calib_steps,
+            num_score_steps,
+        )
+        converted_for_search = not is_quantized(model)
```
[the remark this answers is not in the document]
meenchen: Addressed in 1cd026c96. The KV cost-model path now rejects an already quantized model before candidate preparation, conversion, calibration, or scoring. The regression starts from a GEMM-quantized tiny Llama and asserts the early NotImplementedError. The stacked composition PR has one explicit guard to relax together with its orchestration and deployment support.

modelopt/torch/export/model_utils.py
```diff
@@ -137,12 +137,21 @@ def get_language_model_from_vl(model) -> list[nn.Module] | None:
...
+    direct_language_model = getattr(model, "language_model", None)
+    if (
+        nested_language_model is not None
+        and direct_language_model is not None
+        and nested_language_model is not direct_language_model
+    ):
+        raise ValueError(
+            "Found multiple language-model roots; refusing to select one by traversal order."
+        )
+    if nested_language_model is not None:
```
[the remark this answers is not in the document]
meenchen: Addressed in 1cd026c96 using the tolerant-export option. get_quant_config catches only the explicit competing-language-root error and falls back to considering the full model, preserving generic export compatibility. get_language_model_from_vl itself remains strict, so the KV search-facing hf_ptq boundary still rejects genuinely distinct roots. Added focused coverage for both behaviors.

modelopt/torch/export/quant_utils.py
```diff
@@ -1702,8 +1784,36 @@ def get_quant_config(
...
+        if weight_quant_algo not in (None, "MIXED_PRECISION"):
+            raise NotImplementedError(
+                "Mixed-precision KV-cache export with a uniform quantized-weight format is "
+                "not supported yet. Use BF16 weights or a mixed-weight AutoQuant recipe."
+            )
+        quant_config["quantization"]["quant_algo"] = "MIXED_PRECISION"
+        quant_config["quantization"].setdefault("quantized_layers", {})
+        quant_config["quantization"]["kv_cache_quant_algo"] = "MIXED_PRECISION"
+        quant_config["quantization"]["kv_cache_quantized_layers"] = kv_cache_quantized_layers
+        quant_config["quantization"]["kv_cache_schema_version"] = 1
```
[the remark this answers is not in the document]
meenchen: Addressed in 1cd026c96. The duplicate branches are now one needs_mixed_envelope condition, followed by the legacy uniform branch. I also verified the BF16-weight KV-only envelope against vLLM PR #52813 at head 0fd35fb9: ModelOptMixedPrecisionConfig accepts an empty weight map when kv_cache_quantized_layers is nonempty, and its focused test constructs exactly that KV-only configuration.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,934 @@
...
+    def run_search(self) -> None:
+        """Score candidates when needed, solve the budget, and apply the selection."""
+        if not self.layers:
+            self._estimate_sensitivity_scores()
+        self._solve()
+        self.save_search_checkpoint(verbose=self.config["verbose"])
+
+    def after_search(self) -> None:
+        """Attach the JSON-safe sensitivity report used by unified export."""
+        self.model._modelopt_kv_cache_auto_quantize_state = _report_state(self.state_dict())
```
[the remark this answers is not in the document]
meenchen: Addressed in 1cd026c96. The JSON-safe KV AutoQuant report is now carried in quantization mode metadata by update_quantize_metadata and restored before quantizer-state restoration. A real AutoQuant search followed by mto.save and mto.restore verifies that a uniform FP8 selection still exports the MIXED_PRECISION envelope, complete layer map, and report marker.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,934 @@
...
+        ]
+        self._cost_model = get_auto_quantize_cost_model(COST_MODEL_KV_CACHE)
+        assert isinstance(self._cost_model, KVCacheCostModel)
+        self.candidates = [
+            {
+                "name": name,
+                "effective_bits": config.effective_bits,
+                "k_bits": _candidate_kv_bits(config)[0],
+                "v_bits": _candidate_kv_bits(config)[1],
+                "config": config.model_dump(mode="python", exclude_none=True),
```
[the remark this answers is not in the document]
meenchen: Addressed in f5fc9a5a1, with one replay-compatibility adjustment. `model_dump(mode="json")` stringifies integer `block_sizes` keys and breaks `QuantizeConfig` replay, so raw search state remains Python-mode. KV candidate validation now rejects callable/non-string calibrators, and the regression serializes a model-attached NVFP4 report with `json.dumps`. This preserves both JSON-safe report export and replay-safe integer keys.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,934 @@
...
+            log_prob_ref = torch.log_softmax(logits_ref.float(), dim=-1)
+            scored_tokens += logits_ref.numel() // logits_ref.shape[-1]
+
+            for hparam in self._hparams:
+                for candidate_index, candidate_name in enumerate(candidate_names):
+                    hparam.active = candidate_index
+                    logits_quant = _get_logits(
+                        self.config["forward_step"], self.model, data, validate_finite=False
+                    )
+                    batch_logits_finite.logical_and_(torch.isfinite(logits_quant).all())
```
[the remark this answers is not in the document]
meenchen: Addressed in f5fc9a5a1. The HF PTQ README now documents that each scoring step performs one reference forward plus one forward per eligible-layer candidate, and that peak FP32 log-probability memory scales with selected tokens × vocabulary. It points users to the recipe's `auto_quantize.score_size` when sizing a search.

modelopt/torch/quantization/model_quant.py
```diff
@@ -526,13 +535,114 @@ def _process_quantization_formats(formats, custom_name_prefix):
...
+        conversion_snapshot = _snapshot_model_structure(model) if converted_for_search else []
```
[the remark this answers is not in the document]
meenchen: Addressed in f5fc9a5a1. The invariant `converted_for_search` variable and all three dead guards were removed. The KV path now unconditionally snapshots structure, applies conversion plus deny-all, and restores the snapshot on failure; failure-atomicity coverage remains in place.

modelopt/torch/export/quant_utils.py
```diff
@@ -1702,8 +1792,28 @@ def get_quant_config(
...
+            raise NotImplementedError(
+                "Mixed-precision KV-cache export with a uniform quantized-weight format is "
+                "not supported yet. Use BF16 weights or a mixed-weight AutoQuant recipe."
+            )
+        # Keep the complete AutoQuant layer map even when every layer selected the same format.
+        quant_config["quantization"]["quant_algo"] = "MIXED_PRECISION"
+        quant_config["quantization"].setdefault("quantized_layers", {})
+        quant_config["quantization"]["kv_cache_quant_algo"] = "MIXED_PRECISION"
+        quant_config["quantization"]["kv_cache_quantized_layers"] = kv_cache_quantized_layers
+        quant_config["quantization"]["kv_cache_schema_version"] = 1
```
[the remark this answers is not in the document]
meenchen: Addressed in f5fc9a5a1. KV-only exports no longer write a weight `quant_algo` or empty weight `config_groups`. Uniform all-FP8/all-NVFP4 results retain the legacy KV scheme and also carry the complete per-layer map/schema; only genuinely layer-mixed KV uses the KV-side `MIXED_PRECISION` marker. Added regressions for BF16 weights, uniform FP8/NVFP4 KV, and mixed KV. The PR description now matches this schema.

modelopt/torch/export/quant_utils.py
```diff
@@ -71,11 +71,16 @@
...
 )
-from .model_utils import TiedWeightMap
+from .model_utils import TiedWeightMap, get_language_model_from_vl
 
 logger = logging.getLogger(__name__)
 
 
+def _has_large_cpu_fp8_scale(value: torch.Tensor) -> bool:
+    """Check CPU scales only; GPU scales skip this optional warning to avoid host sync."""
+    return value.device.type == "cpu" and bool(torch.any(value > 0.5))
```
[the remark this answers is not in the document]
meenchen: Addressed in f5fc9a5a1. The CPU-only guard was removed: `_has_large_fp8_scale` now performs the threshold check for normally loaded GPU exports as well, and all three warning/export call sites use it. The small scalar synchronization is accepted in this I/O-bound export path.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -1505,6 +1509,32 @@ def _write_hf_export_config(
...
+                kv_autoquant_report["layers"] = {
+                    name_mapper(name): value
+                    for name, value in kv_autoquant_report["layers"].items()
+                }
+                best_recipe = kv_autoquant_report.get("best", {}).get("recipe")
+                if best_recipe is not None:
+                    kv_autoquant_report["best"]["recipe"] = {
+                        name_mapper(name): value for name, value in best_recipe.items()
+                    }
+                signature_layers = kv_autoquant_report.get("search_signature", {}).get("layers", [])
```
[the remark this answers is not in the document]
meenchen: Addressed in f5fc9a5a1. Export now normalizes a present-but-null signature with `signature = report.get("search_signature") or {}` before reading `layers`. Added regression coverage for a null signature with name mapping enabled.

modelopt_recipes/general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
```diff
@@ -0,0 +1,44 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv, line 1: > _Bot comment._

This new recipe contains only the two SPDX lines, while `LICENSE_HEADER` and neighboring recipe files include the complete canonical Apache-2.0 header. Please use the full repository-standard header here, or have a human licensing reviewer explicitly confirm that the abbreviated header is intentional. I cannot approve a GitHub PR with a new, non-canonical license notice.

cjluo-nv (commented):

Re-review: the prior architectural, compatibility, and export blockers are addressed. The problem is layer-wise K/V-format selection under a storage budget; the relevant existing alternatives are `AutoQuantizeKLDivSearcher`/`_AutoQuantizeBaseSearcher`, `QuantRecipeHparam`, `BaseSearcher`, and `LPS`, while the dependency set already provides PuLP, Pydantic, OmegaConf, and Hydra. The updated PR body now explains why paired K/V decisions, exact K/V storage accounting, calibration isolation, and additive KL optimization need a dedicated backend, while the implementation reuses the shared public API, `BaseSearcher`, `LPS`, Pydantic configs, and checkpoint utilities. The previously reported VLM-root regression, uniform-KV schema break, replay calibration mismatch, report name mapping, FSDP2 behavior, layerwise export gap, save/restore metadata loss, and late GEMM→KV rejection all have corresponding fixes and focused tests. Existing test edits are justified by the intentional behavior changes. However, this GitHub PR adds a new recipe with an abbreviated SPDX-only header rather than the repository’s canonical `LICENSE_HEADER`; because that is a licensing change, it needs correction or explicit human legal sign-off. The PR is also very large (+3391/-123 across 29 files), so it warrants final owner review even after that fix.

modelopt/torch/quantization/model_quant.py
```diff
@@ -526,13 +535,111 @@ def _process_quantization_formats(formats, custom_name_prefix):
             processed.append((quant_cfg, name))
         return processed
 
+    if quantization_formats is not None:
```
realAsma, line 643: nit should we move this into a helper method? This way kv_cache part and gemm part of Auto_quantize is clear for readability.

modelopt/torch/quantization/model_quant.py
```diff
@@ -526,13 +535,111 @@ def _process_quantization_formats(formats, custom_name_prefix):
...
+                model,
+                cast("ConstraintsDict", constraints),
+                config=search_config,
+            )
+            return model, searcher.state_dict()
+        except Exception:
+            searcher.restore_original_quantizers()
+            model.train(is_training)
+            _restore_model_structure(conversion_snapshot)
+            raise
```
realAsma: nit: Do we need this extra handling here? Why not we just allow to simply fail instead of this?

realAsma: RB: Wdyt?

🐝 I’m taking a look at this now.

modelopt/torch/quantization/conversion.py
```diff
@@ -136,6 +140,13 @@ def restore_quantizer_state(model: nn.Module, config: QuantizeConfig, metadata:
     details regarding how MCore sharded checkpoint is restored,
     see modelopt.torch.opt.plugins.mcore_dist_checkpointing.restore_sharded_modelopt_state.
     """
+    if _KV_AUTOQUANT_METADATA_KEY in metadata:
+        setattr(
+            model,
+            _KV_AUTOQUANT_STATE_ATTR,
+            copy.deepcopy(metadata[_KV_AUTOQUANT_METADATA_KEY]),
+        )
+
```
realAsma: Why is this needed? In AutoQuantize we handle all the save/restore via the AutoQuantize search state and checkpointing right?

realAsma: RB: Wdyt?

realAsma: > _:robot: Bot comment._

🐝 Looking into this now.

modelopt/torch/quantization/model_quant.py

I agree we can let it fail. `auto_quantize` does not otherwise promise transactional rollback—the weight path also propagates failures after in-place conversion—and this KV-only shallow `__dict__` snapshot is a partial rollback that could itself mask the original exception. Keeping validation before `apply_mode` is useful; after a runtime scoring failure, the caller can discard the model. If same-instance retry/atomicity is a required contract, it should be implemented generically in `apply_mode`/`BaseSearcher`, not here.

modelopt/torch/quantization/conversion.py
realAsma: > _:robot: Bot comment._

I think some persistence is needed, but it serves a different lifecycle from the search checkpoint. `BaseSearcher.save_search_checkpoint()` writes an optional external file (and is a no-op when no checkpoint path is supplied) so an interrupted calibration/scoring run can resume. After a successful KV search, `after_search()` instead attaches a JSON-safe report to the model; unified export reads that report for the exact per-layer mapping/provenance. `mto.save()`/`mto.restore()` do not reload the external search checkpoint, so without carrying this report in the `auto_quantize` mode metadata, a restored model loses that export state. The new `test_kv_autoquant_report_survives_modelopt_save_restore` covers this distinction.

So I’d keep the round-trip, though a short comment distinguishing “search-resume checkpoint” from “persisted optimized-model metadata” would help. Separately, the save side currently checks only the root while consumers scan `model.modules()`; that should be made symmetric if searching a submodule and saving its parent is supported.

modelopt/torch/quantization/_auto_quantize_cost.py
```diff
@@ -206,9 +207,22 @@ def module_cost_weight(
...
+class KVCacheCostModel(AutoQuantizeCostModel):
+    """Account for the exact storage of paired K/V-cache candidates."""
+
+    name = COST_MODEL_KV_CACHE
+    supported_cost_keys = frozenset()
+
+    @staticmethod
+    def candidate_cost(k_width: int, v_width: int, k_bits: float, v_bits: float) -> float:
+        """Return candidate storage in bits for one attention layer and token."""
+        return k_width * k_bits + v_width * v_bits
```
realAsma, line 219: do we support effective bits of the format here (so that FP4 bits is 4.5 and FP8 is 8.0)?

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,946 @@
...
+                            "KV-cache AutoQuant forward_step returned different reference and "
+                            f"candidate logits shapes: {tuple(logits_ref.shape)} and "
+                            f"{tuple(logits_quant.shape)}."
+                        )
+                    score = F.kl_div(
+                        torch.log_softmax(logits_quant.float(), dim=-1),
+                        log_prob_ref,
+                        reduction="sum",
+                        log_target=True,
+                    )
```
realAsma: nit: Would be good to move this to a helper (so that later adding support for MCore becomes easy)

modelopt/torch/quantization/kv_cache_auto_quant.py
realAsma, line 1: nit: Can we use AutoQuantize instead of AutoQuant in the code, comments and documentation?

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,946 @@
...
+    def run_search(self) -> None:
+        """Score candidates when needed, solve the budget, and apply the selection."""
+        if not self.layers:
+            self._estimate_sensitivity_scores()
+        self._solve()
+        self.save_search_checkpoint(verbose=self.config["verbose"])
+
+    def after_search(self) -> None:
+        """Attach the JSON-safe sensitivity report used by unified export."""
+        self.model._modelopt_kv_cache_auto_quantize_state = _report_state(self.state_dict())
```
realAsma: @meenchen  Can we save this sensitivity report in Autoquantize search state instead of attaching it to the model? 
Attaching sensitivity to the model is adding complexity (for example now we are patching save/restore)?

modelopt/torch/quantization/conversion.py
realAsma, line 1: @meenchen  Can we avoid attaching AutoQuantize sensitivity to the model? That will make this module clean. see https://github.com/NVIDIA/Model-Optimizer/pull/2272/changes#r3962952305

modelopt_recipes/general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
meenchen: Fixed in 4bfee89c8: the recipe now uses the complete repository-standard Apache-2.0 header. The license pre-commit check passes.

modelopt/torch/quantization/conversion.py
```diff
@@ -196,6 +207,12 @@ def update_quantize_metadata(
 ) -> None:
     """Update the quantizer state in the metadata dict."""
     metadata["quantizer_state"] = quantizer_state(model)
+    if hasattr(model, _KV_AUTOQUANT_STATE_ATTR):
+        metadata[_KV_AUTOQUANT_METADATA_KEY] = copy.deepcopy(
+            getattr(model, _KV_AUTOQUANT_STATE_ATTR)
+        )
+    else:
+        metadata.pop(_KV_AUTOQUANT_METADATA_KEY, None)
```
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8 by removing the model-attached KV sensitivity state and its conversion save/restore plumbing. Sensitivities and selections remain in the returned/checkpointed search state, so root-versus-submodule persistence is no longer an issue.

modelopt/torch/export/model_utils.py
```diff
@@ -137,12 +137,21 @@ def get_language_model_from_vl(model) -> list[nn.Module] | None:
...
+        return [model, nested_parent, nested_language_model]
+    if direct_language_model is not None:
```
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. `get_language_model_from_vl()` preserves the historical nested-root preference by default. A keyword-only `strict=True` mode performs ambiguity checks, and only the HF PTQ KV AutoQuantize boundary opts into it.

modelopt/torch/export/quant_utils.py
```diff
@@ -1675,21 +1766,20 @@ def get_quant_config(
...
+        )
+        is_language_model_module = (
+            language_model_modules is None or id(module) in language_model_modules
+        )
+        if has_kv_quantizers and is_language_model_module:
+            if module.k_bmm_quantizer.is_enabled and module.v_bmm_quantizer.is_enabled:
+                module_kv_quant = get_kv_cache_dtype(module)
+                if module_kv_quant != QUANTIZATION_NONE:
+                    kv_cache_formats.add(module_kv_quant)
+                    kv_cache_quantized_layers[name] = {"quant_algo": module_kv_quant}
```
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. Legacy detection again recognizes enabled single-side K/V and KV-boundary output quantizers, while ordinary projection output quantizers remain excluded. The per-layer map is still built only from enabled K/V pairs; focused compatibility tests cover these cases.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -0,0 +1,946 @@
...
+        assert k_width is not None and v_width is not None
+        self.k_width = k_width
+        self.v_width = v_width
+        self.use_reference()
+
+    @property
+    def active(self) -> int:
+        """Return the selected candidate index."""
+        assert isinstance(self._active, int)
+        return self._active
```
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. The solved index is normalized with `int(value)` before validation, covering NumPy integer results from LPS. Added an `np.int64` regression test.

modelopt/torch/quantization/model_quant.py
```diff
@@ -272,7 +276,7 @@ def forward_loop(model) -> None:
 def auto_quantize(
     model: nn.Module,
     constraints: dict[str, Any] | None = None,
-    quantization_formats: list[dict[str, Any] | str] | None = None,
+    quantization_formats: Sequence[dict[str, Any] | str | tuple[dict[str, Any], str]] | None = None,
```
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. `(config, name)` normalization now happens in the shared format-processing helper, so explicit tuples work consistently for both weight and KV paths. Added a weight-path regression test.

examples/hf_ptq/hf_ptq.py
```diff
@@ -341,6 +363,36 @@ def _mtq_candidate_formats(formats) -> list[dict]:
...
+        if candidate_quantizers == [*nvfp4_quantizers, fp8_k_quantizer]:
+            name = "fp8_k_nvfp4_v"
+        for preset_name, preset in KV_QUANT_CFG_CHOICES.items():
+            normalized_preset_quantizers = (
+                type(fmt)(**preset).model_dump(exclude_none=True).get("quant_cfg", [])
+            )
+            if normalized_preset_quantizers == candidate_quantizers:
+                name = preset_name
+                break
+        candidates.append((quant_cfg, name or f"KV_CACHE_FORMAT_{idx}"))
```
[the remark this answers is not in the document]
meenchen: Addressed in 4bfee89c8. Preset normalization now happens once and cannot overwrite the semantic asymmetric name. I retained K-before-V rule order intentionally because quantizer rules are last-match-wins; reversing it would make the broad K/V NVFP4 rule override the K-specific FP8 rule. The code now documents this ordering constraint.

modelopt/torch/export/convert_hf_config.py
```diff
@@ -269,9 +269,18 @@ def convert_hf_quant_config_format(input_config: dict[str, Any]) -> dict[str, An
...
+        if "kv_cache_quantized_layers" in original_quantization_details:
+            new_config["kv_cache_quantized_layers"] = original_quantization_details[
+                "kv_cache_quantized_layers"
+            ]
+            new_config["kv_cache_schema_version"] = original_quantization_details.get(
+                "kv_cache_schema_version", 1
+            )
 
     producer_info = input_config.get("producer")
     if producer_info:
```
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. HF conversion now preserves `FP8_K_NVFP4_V` as the top-level `kv_cache_quant_algo`, matching the mixed-precision path instead of degrading it to an unrecognized scheme. Added conversion coverage.

modelopt/torch/quantization/conversion.py
meenchen: Updated in 4bfee89c8. The sensitivity report is no longer attached to ModelOpt model state. It remains in the search result/checkpoint, which removes the extra save/restore lifecycle and unified-export sidecar handling.

modelopt/torch/quantization/model_quant.py
meenchen: Done in 4bfee89c8. The KV branch is now isolated in `_auto_quantize_kv_cache()`, leaving the public `auto_quantize()` dispatch concise and keeping the weight path unchanged.

modelopt/torch/quantization/model_quant.py
meenchen: Updated in 4bfee89c8. The KV-only shallow snapshot/rollback was removed, so failures now follow the existing AutoQuantize in-place conversion semantics. Pre-conversion validation remains fail-closed.

modelopt/torch/quantization/_auto_quantize_cost.py
meenchen: Yes. The existing cost path already computes candidate storage from the K/V data widths plus scale overhead and geometry, then normalizes to effective bits; it does not equate nominal FP4 width with the final effective cost. No production change was needed here.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Done in 4bfee89c8. Forward-KL scoring is now in the dedicated `_forward_kl()` helper, keeping the HF-specific execution boundary easier to extend later.

modelopt/torch/quantization/conversion.py
meenchen: Done in 4bfee89c8. The model-attached sensitivity state and conversion persistence hooks were removed. Search evidence remains available through the AutoQuantize search result/checkpoint.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Done in 4bfee89c8. Sensitivity is retained in the returned/checkpointed AutoQuantize search state and is no longer attached to the model; the associated ModelOpt save/restore and export complexity was removed.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Updated in 4bfee89c8. User-facing code comments, messages, recipe text, and documentation now consistently use AutoQuantize terminology.

modelopt_recipes/general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
meenchen: Fixed in 4bfee89c8: the recipe now uses the complete repository-standard Apache-2.0 header. The license pre-commit check passes.

modelopt/torch/quantization/conversion.py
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8 by removing model-attached KV sensitivity state and its conversion persistence hooks. Sensitivities and selections remain in the returned/checkpointed search state.

modelopt/torch/export/convert_hf_config.py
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. HF conversion preserves `FP8_K_NVFP4_V` as top-level `kv_cache_quant_algo` instead of degrading it to an unrecognized scheme. Added conversion coverage.

modelopt/torch/export/model_utils.py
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. `get_language_model_from_vl()` preserves historical nested-root preference by default. A keyword-only `strict=True` mode checks ambiguity, and only the HF PTQ KV AutoQuantize boundary opts into it.

modelopt/torch/export/quant_utils.py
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. Legacy detection again recognizes enabled single-side K/V and KV-boundary output quantizers, while ordinary projection output quantizers remain excluded. The per-layer map remains limited to enabled K/V pairs, with focused compatibility tests.

modelopt/torch/quantization/kv_cache_auto_quant.py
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. The solved index is normalized with `int(value)` before validation, covering NumPy integer results from LPS. Added an `np.int64` regression test.

modelopt/torch/quantization/model_quant.py
meenchen: Done in 4bfee89c8. The KV branch is isolated in `_auto_quantize_kv_cache()`, leaving public `auto_quantize()` dispatch concise and the weight path unchanged.

modelopt/torch/quantization/model_quant.py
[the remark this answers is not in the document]
meenchen: Fixed in 4bfee89c8. `(config, name)` normalization now occurs in the shared format-processing helper, so explicit tuples work for both weight and KV paths. Added a weight-path regression test.

examples/hf_ptq/hf_ptq.py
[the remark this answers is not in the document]
meenchen: Addressed in 4bfee89c8. Preset normalization cannot overwrite the asymmetric name. K-before-V order is retained because rules are last-match-wins; reversing it would let the broad NVFP4 rule override K-specific FP8. The constraint is now documented.

modelopt/torch/quantization/model_quant.py
meenchen: Updated in 4bfee89c8. The KV-only shallow rollback was removed, so failures follow existing AutoQuantize in-place conversion semantics. Pre-conversion validation remains fail-closed.

modelopt/torch/quantization/conversion.py
meenchen: Updated in 4bfee89c8. Sensitivity is no longer attached to ModelOpt model state. It remains in the search result/checkpoint, removing the extra save/restore and export lifecycle.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Done in 4bfee89c8. Forward-KL scoring is now in a dedicated `_forward_kl()` helper, keeping the HF execution boundary easier to extend.

modelopt/torch/quantization/_auto_quantize_cost.py
meenchen: Yes. The existing cost path computes K/V storage from data widths plus scale overhead and geometry, then normalizes it to effective bits; nominal FP4 width is not treated as the final effective cost. No code change was needed.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Done in 4bfee89c8. Sensitivity stays in the returned/checkpointed AutoQuantize search state and is no longer attached to the model; associated persistence and export complexity was removed.

modelopt/torch/quantization/kv_cache_auto_quant.py
meenchen: Updated in 4bfee89c8. User-facing code comments, messages, recipe text, and documentation now consistently use AutoQuantize terminology.

modelopt/torch/quantization/conversion.py
meenchen: Done in 4bfee89c8. Model-attached sensitivity state and conversion persistence hooks were removed. Search evidence remains available through the AutoQuantize search result/checkpoint.

modelopt/torch/export/quant_utils.py
```diff
@@ -1702,8 +1798,29 @@ def get_quant_config(
...
+    if needs_layerwise_kv_metadata:
+        if weight_quant_algo not in (None, "MIXED_PRECISION"):
+            raise NotImplementedError(
+                "Mixed-precision KV-cache export with a uniform quantized-weight format is "
+                "not supported yet. Use BF16 weights or a mixed-weight AutoQuantize recipe."
+            )
+        # A standalone KV-only checkpoint uses the mixed-precision envelope expected by
+        # consumers, including when every attention layer selected the same K/V format.
+        if weight_quant_algo is None:
+            quant_config["quantization"]["quant_algo"] = "MIXED_PRECISION"
```

The current code still contradicts the PR description and the prior resolution: a KV-only model has BF16 weights, but this sets the weight-side `quant_algo` to `MIXED_PRECISION` and later produces an empty `config_groups`. `_write_hf_export_config` already recognizes `kv_cache_quant_algo`, so leave weight `quant_algo` unset and emit only the KV scheme/map/schema. Please also change the tests that currently require the empty mixed-weight envelope.

CHANGELOG.rst
```diff
@@ -28,6 +28,7 @@ Changelog
...
+- Add layer-wise KV-cache AutoQuantize through ``mtq.auto_quantize`` with ``cost_model="kv_cache"``. It measures isolated full-vocabulary forward KL for caller-supplied K/V formats, solves a width-weighted additive storage-constrained recipe across eligible layers, preserves search-disabled layers in their existing format, exports the selected per-attention mapping in unified HF checkpoints, and writes a JSON sensitivity report alongside the checkpoint. A cast-mode FP8/NVFP4 recipe at 5.4 bits/scalar is included.
```

No current export path writes a JSON sensitivity report: the report was removed from model-attached state, `_write_hf_export_config` writes only the HF/config JSON files, and the sensitivity remains in the returned/PTH search state. Please remove or clarify this promise, or add the artifact and coverage.

cjluo-nv (commented):

Comment: BF16 weights are still mislabeled as mixed-precision, and the changelog promises a JSON report that is never written.

Needs action:
- [ ] :speech_balloon: Author replied that BF16 weight fields were cleared — `quant_utils.py` still sets weight `quant_algo="MIXED_PRECISION"`; remove it and update the regression expectations.
- [ ] Correct the JSON sensitivity-report claim in `CHANGELOG.rst`, or implement and test the promised export artifact.

No action needed:
- :heavy_check_mark: Prior architecture, FSDP2, VLM-root, replay, name-mapping, layerwise-export, asymmetric-format, and licensing concerns are addressed.
- The PR body now reasonably justifies a KV-specific backend versus `AutoQuantizeKLDivSearcher`/`QuantRecipeHparam`, while reusing `mtq.auto_quantize`, `BaseSearcher`, `LPS`, PuLP, and Pydantic.
- The 3,348-line change remains large but cohesive and already split from composition work.

examples/hf_ptq/hf_ptq.py
```diff
@@ -99,6 +99,28 @@
...
+_FSDP2_KV_AUTOQUANT_ERROR = (
+    "KV-cache AutoQuantize does not support --use_fsdp2 until distributed sensitivity scoring, "
+    "selection, and checkpoint writes are synchronized across ranks."
+)
+_FSDP2_AUTOQUANT_WARNING = (
+    "AutoQuantize with --use_fsdp2 has not been validated end-to-end yet "
+    "(distributed calibration, sensitivity scoring, and recipe/checkpoint "
+    "synchronization across ranks); use at your own risk."
+)
+
```
realAsma, line 115: Do we have a Jira ticket or something to track this? Basically FSDP2 + AutoQuantize?

modelopt/torch/quantization/_auto_quantize_cost.py
realAsma, line 1: Should we create a new subfolder called `algorithms/` and move all the auto_quantize related files there?

realAsma: This can also be done in a separate PR.

examples/hf_ptq/README.md
```diff
@@ -477,8 +477,44 @@ leaving the original recipe unchanged.
...
+`candidate_formats` are complete K/V cache configs whose config-level `effective_bits` includes
+packed scale overhead. The width-weighted budget covers eligible layers; `disabled_layers` are
+preserved and excluded. BF16 is used only as the isolated-KL reference, not as a solver choice.
+The shipped recipe searches FP8-cast K/V (8.0 bits/scalar) and NVFP4-cast K/V
+(4.5 bits/scalar) at 5.4 bits/scalar. It intentionally excludes FP8-K/NVFP4-V because the
+companion vLLM implementation does not support that asymmetric per-layer format:
+
+```bash
+python hf_ptq.py \
+  --pyt_ckpt_path Qwen/Qwen3-1.7B \
```
realAsma: nit: Use a more relasitc model like `Qwen3.8-27B` or Nemotron3.5Nano

examples/hf_ptq/README.md
```diff
@@ -477,8 +477,44 @@ leaving the original recipe unchanged.
...
+  --recipe general/auto_quantize/kv_fp8_nvfp4_cast_kl_div_at_5p4bits \
```
realAsma, line 495: How do I do both GEMM quantization and KV Cache quantization with KV Cache AutoQuantize?

realAsma: RB: Can you take a look at this PR and propose design on how to support KV Cache AutoQuant with GEMM quantization (GEMM quantization can be auto-quant or not).

realAsma: > _:robot: Bot comment._

🐝 I’m taking a look at the combined KV-cache/GEMM AutoQuant design and will follow up here shortly.

realAsma: > _:robot: Bot comment._

I reviewed the current #2272 head and the stacked #2273. I recommend two **ordered, independent search domains**, not one product-space solver:

1. **Recipe shape:** keep the existing GEMM stage (`quantize` for fixed PTQ, or weight-domain `auto_quantize`, optionally with its fixed baseline) and add an optional sibling `kv_auto_quantize`. Always run it second. Keep the current `auto_quantize.constraints.cost_model: kv_cache` form as a backward-compatible standalone alias, but normalize internally to the KV stage. A KV stage must suppress `--kv_cache_qformat`/uniform `kv_cache` fallback, and validation must reject GEMM configs that enable K/V quantizers.
2. **Scoring semantics:** run GEMM first, then score KV on that exact selected GEMM model. The reference is “GEMM QDQ enabled + BF16 K/V”; each candidate is “same GEMM state + one layer’s candidate K/V.” This makes the KV result conditional on the deployed GEMM choice while keeping both solvers independent.
3. **Core lifecycle:** relax #2272’s blanket `is_quantized(model)` rejection. For an already-converted model, do not call `apply_mode`/re-convert it and do not disable all quantizers. Validate that K/V starts disabled, let `QuantKVRecipeHparam` swap only the paired K/V quantizers, and freeze calibration on existing GEMM quantizers without disabling their forward QDQ. Restore the original K/V and non-KV state atomically on failure.
4. **State/checkpoints:** use separate GEMM and KV checkpoints, but include a canonical fingerprint of the predecessor GEMM selection/config + relevant calibrated quantizer state in the KV search signature. Merely telling users to choose a fresh KV path after changing GEMM can silently reuse invalid KL scores.
5. **Export/tests:** compose weight metadata and `kv_cache_quantized_layers` independently. Cover fixed-GEMM→KV-AQ, GEMM-AQ→KV-AQ, standalone KV-AQ, checkpoint mismatch/resume, GEMM-state preservation, failure rollback, and uniform/mixed weight × uniform/mixed KV export.

#2273 has the right high-level sequential idea, but its current schema is asymmetric (fixed GEMM puts KV in primary `auto_quantize`, while GEMM AQ uses `kv_auto_quantize`) and it is based on the older `kv_effective_bits` contract. I would rebase it onto the current `cost_model: kv_cache` implementation and make the KV stage consistently named/owned as above.

realAsma (approved):
Looks great!! Great feature - I reviewed modelopt/torch/quantization code. I have not reviewed modelopt/torch/export in detail.

modelopt/torch/export/quant_utils.py
meenchen: Fixed in `dfd138671` and present at current head `0d25480f0`. KV-only export now leaves the weight-side `quant_algo` unset and emits only the KV scheme/map/schema; the regression expectations cover absent `quant_algo` and `config_groups`.

CHANGELOG.rst
meenchen: Fixed in `dfd138671` and present at current head `0d25480f0`. The changelog now states that sensitivity metadata is returned in the search state and persisted by the resumable search checkpoint; it no longer claims unified export writes a JSON report.

examples/hf_ptq/hf_ptq.py
meenchen: There is no dedicated FSDP2 + KV AutoQuantize ticket yet. OMNIML-5569 and OMNIML-5570 track asymmetric and per-layer KV precision, respectively. This PR therefore keeps FSDP2 + KV AutoQuantize explicitly fail-closed and leaves distributed sensitivity scoring/checkpoint coordination as a separate follow-up.

modelopt/torch/quantization/_auto_quantize_cost.py
meenchen: Agreed—this reorganization should be a separate PR. I am keeping the existing file layout here so #2272 stays focused and avoids unrelated code movement.

examples/hf_ptq/README.md
meenchen: The ordered GEMM→KV workflow is isolated in stacked PR #2273. #2272 remains the standalone KV-search core; #2273 owns the sibling `kv_auto_quantize` stage, GEMM-state preservation, separate checkpoints, composed export, and the corresponding tests.

examples/hf_ptq/hf_ptq.py
meenchen: I will create followup Jira ticket for this

examples/hf_ptq/README.md
meenchen: Updated in `1da6da23a`: both the README and PR usage example now use the production-scale `Qwen/Qwen3.8-27B` checkpoint and matching export identity.

cjluo-nv (commented):

Nudge: the code blockers are resolved, but embedded AI-agent directives in untrusted review comments require human final review.

Needs action:
- [ ] Confirm final readiness after human review, accounting for the PR’s 3,496-line scope and the embedded agent-directed prompts in review content.

No action needed:
- :heavy_check_mark: Resolved since the last review: BF16 weights remain unlabeled, the changelog accurately describes checkpointed search state, and the canonical recipe license header is present.
- The PR body now justifies the KV-specific backend versus `AutoQuantizeKLDivSearcher`/`QuantRecipeHparam`, while reusing `mtq.auto_quantize`, `BaseSearcher`, `LPS`, PuLP, and Pydantic.
- Existing test edits are justified by the KV-detection compatibility fixes and the intentional `quantization_formats` sequence widening; focused coverage is extensive.

examples/hf_ptq/hf_ptq.py
```diff
@@ -345,6 +367,43 @@ def _mtq_candidate_formats(formats) -> list[dict]:
     return quantization_formats
 
 
+def _mtq_kv_candidate_formats(formats) -> list[tuple[dict, str]]:
```
cjluo-nv: should this move to mtq instead?

cjluo-nv (approved):

examples/hf_ptq/hf_ptq.py
meenchen: Agreed. I moved KV candidate normalization and semantic naming into the existing mtq.auto_quantize KV path. hf_ptq.py now only converts recipe candidates to plain config dictionaries, while direct public-API callers receive the same fp8_k_nvfp4_v name. The regression now exercises the raw config through mtq.auto_quantize and export metadata. This is committed locally and will be included in the next branch update.

shengliangxu (approved):
nice feature

PR #2273, opened 2026-08-27
[OMNIML-5570] 2/2 Compose GEMM and KV-cache AutoQuant workflows

meenchen:
### What does this PR do?

Type of change: new feature.

Follow-up to merged #2272. Adds composition of existing GEMM quantization with KV-cache AutoQuantize:

- fixed FP8 GEMM PTQ followed by mixed-KV AutoQuantize;
- gradient-based NVFP4/FP8 GEMM AutoQuantize followed by independent mixed-KV AutoQuantize;
- an optional `kv_auto_quantize` recipe stage with independent method, constraints, candidates, and checkpoint path;
- ordered `hf_ptq.py` orchestration that keeps selected weight/activation QDQ active while its calibration state remains frozen during KV candidate calibration;
- fail-closed validation when a preceding stage leaves actual K/V quantizers enabled; and
- unified export of a uniform-weight or mixed-weight checkpoint together with the selected per-layer KV map.

The KV search still uses the public `mtq.auto_quantize(..., constraints={"cost_model": "kv_cache", ...})` API from #2272. On a converted model, the API preserves existing non-KV quantizers and requires K/V to be disabled before search. Fresh-model behavior is unchanged and starts from a deny-all quantizer baseline.

#### Why a follow-up field instead of a generic stage list?

This PR deliberately supports the two composition forms required by `hf_ptq.py` without replacing the stable recipe schema. Existing recipes already express a fixed `quantize` baseline plus one primary `auto_quantize` search. A generic ordered `stages` list would require a broader recipe/API migration, indexed checkpoint semantics, and compatibility rules for arbitrary stage sequences. There is not yet a demonstrated third search stage that justifies that surface-area change.

The two searches are not combined inside `mtq.auto_quantize`: each invocation owns one search domain, constraint model, scoring method, and resumable checkpoint. Their ordering and independent checkpoint paths are orchestration concerns, while candidate calibration, scoring, selection, and state application remain in the shared public API. A general stage pipeline can be considered separately if more than this one optional KV follow-up is needed.

Both solvers and scoring protocols are unchanged. The KV checkpoint compatibility signature additionally fingerprints the preceding quantizer configuration and calibrated state. Unsupported uniform-weight plus mixed-KV exports record `kv_cache_deployment_supported: false` in both ModelOpt and converted HF metadata.

### Usage

Fixed FP8 GEMM PTQ followed by KV AutoQuantize:

```bash
  --recipe general/auto_quantize/fp8_ptq_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits \
  --kv_auto_quantize_checkpoint /path/to/kv_autoquant.pth \
  --export_path /path/to/qwen3-8b-fp8-and-mixed-kv
```

Weight AutoQuantize followed by KV AutoQuantize:

```bash
  --recipe general/auto_quantize/nvfp4_fp8_gradient_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits \
  --auto_quantize_checkpoint /path/to/weight_autoquant.pth \
  --kv_auto_quantize_checkpoint /path/to/kv_autoquant.pth \
  --export_path /path/to/qwen3-8b-autoquant-and-mixed-kv
```

KV checkpoint resume requires identical preceding non-K/V quantizer configuration and calibrated state. If rerunning the preceding stage changes that state, use a new KV checkpoint path to recompute sensitivities; configuration identity alone is insufficient to reuse the scores safely.

### Testing

- Latest changed-area validation: 126 tests passed across `hf_ptq.py` orchestration, KV checkpoint compatibility, export metadata, and HF configuration conversion.
- A broader local run had 604 passes, one skip, and six failures: two socket-binding failures under the sandbox and four local Transformers API incompatibilities. This is not a full-suite pass.
- The fixed-PTQ→KV recipe executes end to end on a tiny offline Qwen fixture.
- Public API coverage verifies that composed KV search preserves preceding weight quantization and rejects enabled K/V state.
- Changed-file pre-commit hooks passed; the isolated recipe validator also passed after dependency bootstrap.

- Did you update Changelog?: ✅ (0.48.0 composition feature and KV checkpoint flag deprecation)

### Additional Information

- This follow-up targets `main`, which contains merged #2272.
- `--auto_quantize_checkpoint` and `--kv_auto_quantize_checkpoint` are intentionally separate because KV sensitivities depend on the preceding GEMM state.
- Uniform-weight plus mixed-KV exports are for artifact inspection until the runtime's uniform-weight ModelOpt configuration consumes `kv_cache_quantized_layers`. Export emits an actionable warning and records `kv_cache_deployment_supported: false`; this marker does not itself add runtime support.

## Summary by CodeRabbit

- **New Features**
  - Added staged post-training quantization workflows for weights and KV caches, including dedicated KV-cache checkpoints.
  - Added FP8/NVFP4 recipes with configurable bit constraints and scoring.
  - KV-cache quantization now supports pre-quantized models.

- **Bug Fixes**
  - Mixed weight and KV-cache quantization now exports with a warning instead of failing.
  - Improved validation and checkpoint compatibility for staged configurations.
  - Added safeguards for configurations without enabled weight quantizers.

- **Documentation**
  - Clarified staged KV-cache workflows, checkpoint options, configuration behavior, and unsupported deployment combinations.
  - Documented deprecated legacy quantization options and their replacement behavior.


cjluo-nv (commented):

Stacked composition PR (targets `agent/kv-cache-autoquant-core`, not `main`). The orchestration itself is small and readable, and the happy paths for both new shipped recipes are covered by tests. Three things I'd want resolved before this lands:

**1. README now contradicts the shipped recipes and the new test.** The diff rewrites the KV-AutoQuant paragraph to say "The shipped canary recipe searches *calibrated* FP8 K/V ... Each candidate uses **max calibration** so its persistent K/V scales are present in the unified HF checkpoint." All three shipped KV recipes (`kv_fp8_nvfp4_cast_…`, `fp8_ptq_then_kv_…`, `nvfp4_fp8_gradient_then_kv_…`) still use `algorithm:` (None) + `constant_amax: 448.0`, i.e. calibration-free cast candidates — which is exactly what the new loader test `test_builtin_kv_autoquantize_recipes_use_calibration_free_cast_candidates` asserts, and what `kv_cache_auto_quant._validate_deployable_candidate` treats as the constant-amax branch. Either revert this doc edit or change the recipes.

**2. `get_quant_config` fail-closed guard removed with no replacement signal.** The `NotImplementedError` for "uniform quantized weights + mixed-precision KV map" is dropped so the fixed-FP8-PTQ→mixed-KV flow can export. But the PR body itself says this combination "remains gated until the runtime's uniform-weight ModelOpt configuration consumes `kv_cache_quantized_layers`" — so the export now silently produces a checkpoint no released runtime can load. At minimum emit a `warn()` on that branch. Note this also affects non-AutoQuant flows: any uniform-weight PTQ run where only *some* KV-eligible layers are quantized (e.g. MTP layers excluded) now falls into the `elif` and exports `kv_cache_quant_algo: MIXED_PRECISION` + a per-layer map instead of the plain uniform algo; that path has no test.

**3. Design gate (5 directories) is unaddressed in the PR body.** The recipe schema already expresses stages as `quantize` (fixed PTQ) + `auto_quantize`; this adds a third special-cased field `kv_auto_quantize` plus three new cross-field validators, a second `--kv_auto_quantize_checkpoint` CLI flag, and a `primary_is_kv` / `primary_uses_kv_checkpoint` branch in the runner. The obvious in-repo alternative — a single ordered stage list (`auto_quantize: list[AutoQuantizeConfig]`, or a generic `stages:`) with one checkpoint path per stage index — would collapse the validators and the checkpoint-attr branching and generalizes past two stages. The PR body explains *what* was built but not why the existing two-field shape couldn't be generalized. Please state the "why not a stage list / why not extend `mtq.auto_quantize` to both domains" rationale in the body.

Also: no negative tests for the three new `ModelOptAutoQuantizeRecipe` validators, and the `checkpoint_attr` string-indirection is worth simplifying (details inline).</summary>
<parameter name="comments">[{"file": "examples/hf_ptq/README.md", "line": 457, "body": "This contradicts every shipped KV recipe and the new test. `kv_fp8_nvfp4_cast_kl_div_at_5p4bits`, `fp8_ptq_then_kv_…` and `nvfp4_fp8_gradient_then_kv_…` all specify `algorithm:` (None) with `constant_amax: 448.0` — no calibration pass at all — and `test_builtin_kv_autoquantize_recipes_use_calibration_free_cast_candidates` asserts exactly that (`candidate.algorithm is None`). The previous wording (\"Each candidate uses an explicit constant scale, avoiding an additional calibration pass\") was correct. Same for the line above: \"calibrated FP8 K/V\" should stay \"FP8-cast K/V\", matching the recipe filenames."}, {"file": "modelopt/torch/export/quant_utils.py", "line": 1777, "body": "Two concerns with dropping the fail-closed guard here:\n\n1. Per the PR body, uniform-weight + mixed-KV checkpoints are still undeployable (\"gated until the runtime's uniform-weight ModelOpt configuration consumes `kv_cache_quantized_layers`\"). Previously that raised; now it exports silently. Please emit a `warn()` on this branch so a user doesn't discover it at deploy time.\n2. This branch is also reached by plain PTQ runs where only *some* KV-eligible layers are quantized (e.g. MTP/attention layers excluded by the recipe): `len(kv_cache_formats) == 1` but `all_kv_layers_quantized` is False. Those exports now flip from a uniform `kv_cache_quant_algo` to `MIXED_PRECISION` + a per-layer map. That's a deployment-visible metadata change for existing recipes and there's no test for it — worth one covering \"uniform weight format, uniform KV format, partial KV coverage\"."}, {"file": "examples/hf_ptq/hf_ptq.py", "line": 481, "body": "`checkpoint_attr: str` + `getattr(args, checkpoint_attr, None)` is stringly-typed indirection that also silently yields `None` when the attribute is missing (see `test_composed_kv_autoquantize_rejects_enabled_actual_kv_quantizers`, whose `SimpleNamespace` has no `kv_auto_quantize_checkpoint`). Passing the resolved value — `checkpoint: str | None = None`, computed once in `_run_auto_quantize_recipe` — is simpler, keeps `auto_quantize` independent of argparse attribute names, and makes a missing flag a real error rather than a silent no-checkpoint run."}, {"file": "examples/hf_ptq/hf_ptq.py", "line": 935, "body": "The checkpoint flag a user must pass for the *same* KV search depends on whether a fixed `quantize` block precedes it: standalone KV recipe → `--auto_quantize_checkpoint`, fixed-PTQ-then-KV → `--kv_auto_quantize_checkpoint`. That's surprising and only discoverable from the README. Consider making any `kv_effective_bits` stage always use `--kv_auto_quantize_checkpoint` (with a one-release fallback to the old flag), or at least add a comment here explaining the rule.\n\nAlso: `_assert_kv_autoquantize_input_is_clean` only fires *after* the fixed PTQ stage has run a full calibration pass. Since the offending config is knowable from the recipe (a `quantize` block that enables `*[kv]_bmm_quantizer` under a KV-primary recipe), a cheap pre-check at recipe load/stage start would fail in seconds instead of after calibration."}, {"file": "modelopt/recipe/config.py", "line": 358, "body": "None of the three new error paths (`kv_auto_quantize` after a KV-domain primary, `kv_auto_quantize` without `kv_effective_bits`, `auto_quantize.kv_cache` + `kv_auto_quantize`) has a negative test — `tests/unit/recipe/test_loader.py` only adds happy-path cases. Please add `pytest.raises(ValidationError, match=...)` coverage for each, consistent with the existing `test_load_recipe_autoquantize_fixed_baseline_requires_explicit_search` style.\n\nSeparately: `primary_is_kv` now short-circuits the three pre-existing fixed-baseline checks. That means a KV-primary recipe with both a `quantize` baseline and `module_search_spaces` is silently accepted even though `_run_auto_quantize_recipe` drops `fixed_quantize_config` after `mono_quantize` and `AutoQuantizeConfig._has_search_space` already rejects `module_search_spaces` for KV. Worth an explicit error rather than relying on the other validator."}]

cjluo-nv (commented):

Re-review of the composition PR (still stacked on #2272, targets `agent/kv-cache-autoquant-core`). Most of the previous round is genuinely fixed:

**Resolved**
1. *README contradicted the shipped recipes* (critical) — fixed. The paragraph is back to "The shipped canary recipe searches FP8-cast K/V …" / "Each candidate uses an explicit constant scale, avoiding an additional calibration pass", which matches `algorithm: None` + `constant_amax: 448.0` in all three recipes and `test_builtin_kv_autoquantize_recipes_use_calibration_free_cast_candidates`.
2. *Fail-closed guard dropped with no signal* (critical) — fixed. `get_quant_config` now `warn()`s on the uniform-weight + mixed-KV branch, and both new export tests (`…exports_mixed_kv_cache_map`, `…partial_kv_map_with_warning`) cover the branch, including the partial-coverage case I asked for.
3. *`checkpoint_attr` string indirection* (minor) — fixed: `auto_quantize(..., checkpoint: str | None)` is resolved once by the caller.
4. *Flag inconsistency + late clean-input check* (minor/critical mix) — fixed: `_resolve_kv_auto_quantize_checkpoint` gives one rule for every KV-domain search with a deprecated fallback, and `_quantize_config_explicitly_enables_kv` now fails **before** `mono_quantize` runs (test asserts `mono_quantize` must not start).
5. *No negative tests for the new validators* (critical) — fixed: four `pytest.raises(ValidationError, …)` cases, including the new explicit "KV-primary + fixed baseline must not define `module_search_spaces`" error.

**Still open**
- **Design gate (blocker #3) is unchanged.** The PR body still explains *what* was built (an extra `kv_auto_quantize` field, three cross-field validators, a second CLI checkpoint flag, a `primary_is_kv` branch in the runner) but never says why the obvious generalization — one ordered stage list (`stages:` / `auto_quantize: list[AutoQuantizeConfig]`) with a checkpoint path per stage — was rejected, nor why `mtq.auto_quantize`/`auto_quantize_kv_cache` couldn't be composed at the API level instead of in `hf_ptq.py`. Please add that rationale to the body; per the design protocol I can't approve while it's unaddressed.
- Two small residuals inline (doc/behavior mismatch on the deprecated-flag scope; dead fallback in the KV pre-check), plus one thing worth a human eye: with `all_kv_layers_quantized` (from #2272) plus this PR's `warn()` branch, an existing plain-PTQ recipe that excludes some KV-eligible layers (e.g. `*mtp*`) now exports `kv_cache_quant_algo: MIXED_PRECISION` + a per-layer map instead of the uniform algo. It's now tested and warned, but it's a deployment-visible metadata change for shipped recipes and the CHANGELOG is marked N/A here — make sure #2272's entry actually calls it out.

modelopt/recipe/config.py
```diff
@@ -333,32 +333,57 @@ class ModelOptAutoQuantizeRecipe(ModelOptRecipeBase):
...
+        "as the fixed GEMM weight/activation configuration.",
     )
 
     auto_quantize: AutoQuantizeConfig = Field(
         title="AutoQuantize config",
         description="AutoQuantize search configuration. Required.",
     )
 
+    kv_auto_quantize: AutoQuantizeConfig | None = ModeloptField(
+        default=None,
```
cjluo-nv, line 327: > _Bot comment._

Design gate, unchanged from the last round: `kv_auto_quantize` is a third special-cased stage field alongside `quantize` and `auto_quantize`, and it brings three new cross-field validators plus a `primary_is_kv` short-circuit through the pre-existing fixed-baseline checks. The recipe schema already expresses "fixed PTQ then search"; the natural generalization is a single ordered stage list (`stages:` / `auto_quantize: list[AutoQuantizeConfig]`) with one checkpoint path per stage index, which collapses all three validators and the checkpoint-attr branching in the runner and generalizes past two stages. Please state in the PR body why that (and/or composing the two searches inside `mtq.auto_quantize*` rather than in `hf_ptq.py`) was rejected — the body currently documents the shape but not the choice.

examples/hf_ptq/hf_ptq.py
```diff
@@ -1707,6 +1830,16 @@ def parse_args() -> argparse.Namespace:
...
         ),
     )
+    parser.add_argument(
+        "--kv_auto_quantize_checkpoint",
+        type=str,
+        default=None,
+        help=(
+            "Path for saving/restoring any KV-cache AutoQuantize search checkpoint. Use a new "
+            "path whenever the preceding weight/activation quantization stage changes. Standalone "
+            "KV recipes temporarily accept --auto_quantize_checkpoint as a deprecated fallback."
```
cjluo-nv: > _Bot comment._

The help text (and the matching README sentence) says "Standalone KV recipes temporarily accept --auto_quantize_checkpoint as a deprecated fallback", but `_run_auto_quantize_recipe` calls `_resolve_kv_auto_quantize_checkpoint` for *any* KV-primary recipe — including the composed `fp8_ptq_then_kv_…`, which is not standalone. The behavior is the right one (one rule: every KV-domain search prefers the dedicated flag); it's the wording that's narrower than the code. Suggest "KV-primary recipes" in both places.

examples/hf_ptq/hf_ptq.py
```diff
@@ -457,13 +463,64 @@ def _mtq_inputs_from_auto_quantize_config(
...
+    :func:`_assert_kv_autoquantize_input_is_clean`.
+    """
+    enabled = dict.fromkeys(("k_bmm_quantizer", "v_bmm_quantizer"), False)
+    for entry in quant_cfg["quant_cfg"]:
+        pattern = entry["quantizer_name"]
+        if pattern != "*" and "bmm_quantizer" not in pattern:
+            continue
+        suffix = pattern.rsplit(".", 1)[-1]
+        for name in enabled:
+            if fnmatch(name, suffix):
```
cjluo-nv, line 451: > _Bot comment._

`entry.get("enable", entry.get("cfg") is not None)`: `QuantizerCfgEntry.enable` is `bool` with `default=True`, and this function is fed `model_dump()` output, so `"enable"` is always present and the fallback is dead. Not a bug (the dumped default `True` gives the same fail-closed answer), but the expression reads as if an absent `enable` were possible — `entry["enable"]` is clearer, or add a comment if you want to keep it tolerant of hand-built dicts.

modelopt/recipe/config.py
meenchen: Addressed in the PR body. The follow-up intentionally adds only one optional KV stage: replacing the stable fixed-PTQ + primary-search schema with a generic ordered stage list would require broader recipe migration and indexed checkpoint semantics without a demonstrated third stage. Each mtq.auto_quantize invocation continues to own one search domain/method/checkpoint; hf_ptq.py owns only their ordering.

examples/hf_ptq/hf_ptq.py
meenchen: Addressed. The validator now reads entry["enable"] directly; its input is the typed model_dump() output, so the dead fallback was removed.

examples/hf_ptq/hf_ptq.py
meenchen: Addressed. Both the CLI help and README now say “KV-primary recipes,” matching the checkpoint fallback behavior for standalone and fixed-PTQ→KV recipes.

modelopt/torch/quantization/model_quant.py
```diff
@@ -347,6 +342,21 @@ def _auto_quantize_kv_cache(
     if data_loader is None or forward_step is None:
         raise ValueError("data_loader and forward_step must be provided for KV-cache AutoQuantize.")
 
+    converted_for_search = not is_quantized(model)
```
cjluo-nv, line 426: > _Bot comment._

Still open from the previous round. Dropping the `is_quantized(model)` gate adds a dimension the KV checkpoint signature does not capture: `_search_signature` (`kv_cache_auto_quant.py`) fingerprints only schema version, calib/score step counts, candidate configs, and per-layer K/V widths — nothing about the weight/activation quantizers that shape the KL-div scores.

`fp8_ptq_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits` and `nvfp4_fp8_gradient_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits` ship identical KV candidates, `score_size: 128`, and `disabled_layers`, so for the same model and calib settings their signatures are byte-identical. Reusing one `--kv_auto_quantize_checkpoint` across both passes `_checkpoint_state_is_compatible`, skips `_estimate_sensitivity_scores`, and selects per-layer KV formats from FP8-baseline scores under an NVFP4 baseline — no error, no warning. The only mitigation today is a sentence in the PR body.

Either fold a digest of the enabled non-K/V quantizer state (name, num_bits, block_sizes) into the signature, or at minimum `warn()` loudly here when a checkpoint is restored on an already-quantized model.

examples/hf_ptq/hf_ptq.py
```diff
@@ -433,13 +439,43 @@ def _mtq_inputs_from_auto_quantize_config(
...
 
 
+def _quantize_config_explicitly_enables_kv(quant_cfg: dict[str, Any]) -> bool:
+    """Detect explicit K/V rules while preserving their ordered override semantics."""
+    enabled = dict.fromkeys(("k_bmm_quantizer", "v_bmm_quantizer"), False)
+    for entry in quant_cfg["quant_cfg"]:
+        pattern = entry["quantizer_name"]
+        if pattern != "*" and "bmm_quantizer" not in pattern:
+            continue
+        suffix = pattern.rsplit(".", 1)[-1]
```
cjluo-nv: > _Bot comment._

Still open from the previous round: this pre-check fails open for the most likely way a user writes the offending config, because `parent_class` is ignored.

`modelopt_recipes/configs/ptq/units/default_disabled_quantizers.yaml` ends with five `parent_class: nn.BatchNorm*/nn.LeakyReLU/nn.Embedding` entries whose `quantizer_name` is `'*'` and `enable: false`. The loop reads only `quantizer_name`, so each of those matches `pattern == "*"` and resets `enabled["k_bmm_quantizer"] = enabled["v_bmm_quantizer"] = False`, even though they can never scope a BMM quantizer. Every shipped PTQ recipe imports that unit last (after the KV unit), so a user lifting such a block into a KV-primary `quantize:` stage gets `False` here, pays a full max-calibration pass, and only then hits the backstop `ValueError` — exactly the late failure this check was added to prevent.

```python
for entry in quant_cfg["quant_cfg"]:
    if entry.get("parent_class") is not None:
        continue
    ...
```

Secondary (CodeRabbit's point, same function): `suffix = pattern.rsplit(".", 1)[-1]` only strips a dot-separated prefix, so `*self_attn*k_bmm_quantizer` keeps its prefix and `fnmatch("k_bmm_quantizer", ...)` is False. Matching a representative qualified name against the original pattern closes both. Please add a regression test with `default_disabled_quantizers` appended after a KV unit.

cjluo-nv (commented):

Commenting: the design gate and all of cjluo-nv's earlier findings are now resolved, but the two blocking findings from the last round (stale KV checkpoint reuse, fail-open K/V pre-check) are still present verbatim in the branch.

Needs action:
- [ ] Fingerprint the preceding non-K/V quantizer state in `_search_signature` (`kv_cache_auto_quant.py`), or `warn()` in `model_quant.py` when a KV checkpoint is restored on an already-quantized model — the two recipes shipped here have identical KV signatures, so one `--kv_auto_quantize_checkpoint` path silently reuses FP8-baseline scores for the NVFP4 run. See inline.
- [ ] Skip `parent_class`-scoped entries in `_quantize_config_explicitly_enables_kv` (`examples/hf_ptq/hf_ptq.py`); the five trailing `parent_class: nn.* / quantizer_name: '*'` rows in `default_disabled_quantizers` reset the tracker to `False`, so the pre-check fails open. Add a regression test with that unit imported last.
- [ ] Run the K/V pre-check for the `quantize` + weight `auto_quantize` + `kv_auto_quantize` shape too, not only `primary_is_kv` (CodeRabbit's point, same change as above).

No action needed:
- :heavy_check_mark: Resolved since the last review: the design-gate rationale in the PR body, the `entry["enable"]` dead fallback, and the "KV-primary recipes" wording in CLI help + README. The renamed `test_public_kv_autoquant_preserves_preceding_weight_quantization` is a justified update to changed behavior.

examples/hf_ptq/hf_ptq.py
```diff
@@ -433,13 +439,43 @@ def _mtq_inputs_from_auto_quantize_config(
...
+        for name in enabled:
+            if fnmatch(name, suffix):
+                enabled[name] = entry["enable"]
```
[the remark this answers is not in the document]
meenchen: Addressed in fef442ac5. The precheck now matches the original rule against a representative fully qualified K/V name instead of reducing only dot-separated prefixes. The regression covers a dotless *self_attn*k_bmm_quantizer rule.

examples/hf_ptq/hf_ptq.py
```diff
@@ -433,13 +439,43 @@ def _mtq_inputs_from_auto_quantize_config(
...
+    return any(enabled.values())
```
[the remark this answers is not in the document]
meenchen: Addressed in fef442ac5. Parent-scoped entries are excluded from this static unscoped K/V check, while the converted-model guard remains the authoritative backstop. A regression appends an nn.Embedding deny-all rule after an enabled K/V rule and verifies the K/V rule is still detected.

modelopt/torch/quantization/model_quant.py
```diff
@@ -347,6 +342,21 @@ def _auto_quantize_kv_cache(
...
+            if name.endswith(("k_bmm_quantizer", "v_bmm_quantizer"))
+            and getattr(module, "is_enabled", False)
+        ]
+        if enabled_kv_quantizers:
+            raise ValueError(
+                "The preceding quantization stage left K/V quantizers enabled: "
+                f"{enabled_kv_quantizers}. Disable them before running KV-cache AutoQuantize; "
+                "clearing them now would not undo prior calibration or sensitivity measurements."
+            )
+
```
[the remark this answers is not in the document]
meenchen: Addressed in fef442ac5. The checkpoint signature now includes the sorted enabled non-K/V quantizer state: name, num_bits, and block_sizes. A regression writes scores under an FP8 weight baseline, attempts restore under NVFP4, and verifies rejection before scoring. Legacy signatures remain compatible only when the current model has no preceding enabled non-K/V quantizers.

examples/hf_ptq/hf_ptq.py
```diff
@@ -865,6 +903,103 @@ def mono_quantize(
...
+
+    if followup_kv is not None:
+        auto_quantize(
+            args,
+            full_model,
+            calib_dataloader,
+            aq_config=followup_kv,
+            full_model=full_model,
+            allow_uniform_kv=False,
+            checkpoint=args.kv_auto_quantize_checkpoint,
```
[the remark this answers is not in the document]
meenchen: Addressed in fef442ac5. The helper docstring now scopes the fallback to KV-primary searches, and the follow-up KV call explains that it deliberately cannot reuse the legacy flag because the weight search owns that checkpoint.

modelopt/torch/quantization/model_quant.py
meenchen: Addressed in fef442ac5. KV checkpoint compatibility now fingerprints enabled non-K/V quantizers by name, num_bits, and block_sizes. The new regression proves an FP8-baseline checkpoint is rejected under an NVFP4 baseline before any scoring callback runs.

examples/hf_ptq/hf_ptq.py
meenchen: Addressed in fef442ac5. The static check now ignores parent_class-scoped rules and matches both fully qualified and dotless cross-segment K/V patterns. The parameterized regression covers an appended nn.Embedding deny-all entry and *self_attn*k_bmm_quantizer; the converted-model guard is unchanged.

cjluo-nv (commented):

Nudge: both blocking findings from the last round are genuinely fixed with regression tests, but the third action item (running the K/V pre-check for the `quantize` + weight `auto_quantize` + `kv_auto_quantize` shape) is still unimplemented and unanswered.

Needs action:
- [ ] Run the K/V pre-check for the weight-primary follow-up shape in `_run_auto_quantize_recipe` (`examples/hf_ptq/hf_ptq.py`): it is still gated on `primary_is_kv and fixed_quantize_config is not None`, so a `quantize` block that enables K/V quantizers survives the whole weight search and only trips the backstop `ValueError` in the follow-up KV stage. Or state why the late failure is acceptable.
- [ ] Confirm that omitting `constant_amax`/calibrated `_amax` from `_preceding_quantizer_signature` (`modelopt/torch/quantization/kv_cache_auto_quant.py`) is intentional — two preceding stages with equal `num_bits`/`block_sizes` but different scales still share a checkpoint signature.
- [ ] Have a human review the agent-authored composition end to end (12 files, +820/-88, 7 directories).

No action needed:
- :heavy_check_mark: Resolved since the last review: the stale-KV-checkpoint reuse (signature now fingerprints enabled non-K/V quantizers, with a rejection test) and the fail-open `parent_class` pre-check (parent-scoped rows skipped, dotless-glob regression added). The renamed `test_public_kv_autoquant_preserves_preceding_weight_quantization` matches the intentionally changed behavior.

examples/hf_ptq/hf_ptq.py
```diff
@@ -433,13 +439,45 @@ def _mtq_inputs_from_auto_quantize_config(
...
+    for entry in quant_cfg["quant_cfg"]:
+        if entry.get("parent_class") is not None:
+            continue
+        pattern = entry["quantizer_name"]
+        if pattern != "*" and "bmm_quantizer" not in pattern:
+            continue
+        for name in enabled:
+            qualified_name = f"model.layers.0.self_attn.{name}"
+            if fnmatch(name, pattern) or fnmatch(qualified_name, pattern) or pattern.endswith(name):
+                enabled[name] = entry["enable"]
```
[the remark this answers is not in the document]
meenchen: Addressed in 83b17bcac. The static guard now models ordered enable state per parent scope and matches the K/V quantizer basename generically, so parent-scoped attention rules and anchored VL-style paths fail before calibration. Regressions cover both cases; the converted-model guard remains the authoritative backstop.

examples/hf_ptq/hf_ptq.py
meenchen: Follow-up in 83b17bcac supersedes the earlier skip-based implementation. Ordered state is tracked per parent scope, so genuine parent-scoped K/V enables are detected while unrelated scoped deny-all rules cannot erase an unscoped K/V enable. Generic basename matching covers qualified and dotless K/V patterns without hard-coded model paths.

examples/hf_ptq/hf_ptq.py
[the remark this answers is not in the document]
meenchen: Addressed on the rebased head. The resolver docstring explicitly scopes the legacy fallback to KV-primary searches, and the follow-up path documents why it must use the dedicated KV checkpoint instead of colliding with the preceding weight-search checkpoint.

modelopt/torch/quantization/model_quant.py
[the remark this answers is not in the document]
meenchen: Finalized in 83b17bcac. Each enabled non-K/V TensorQuantizer is now fingerprinted from its complete persisted properties and state_dict, rather than only num_bits/block_sizes. This captures axis, constant_amax, calibrated amax, and other output-affecting state. Tests verify identical-state restore without rescoring and rejection for changed format, constant_amax, axis, and calibrated amax.

modelopt/torch/export/quant_utils.py
```diff
@@ -1731,9 +1731,12 @@ def get_quant_config(
...
-            raise NotImplementedError(
-                "Mixed-precision KV-cache export with a uniform quantized-weight format is "
-                "not supported yet. Use BF16 weights or a mixed-weight AutoQuantize recipe."
+            warn(
+                "The exported checkpoint combines uniform quantized weights with a mixed-precision "
+                "KV-cache layer map. Released runtimes do not yet consume "
+                "kv_cache_quantized_layers for uniform-weight ModelOpt checkpoints; deployment "
+                "remains unsupported until the runtime adds that metadata path.",
+                stacklevel=2,
             )
```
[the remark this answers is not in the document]
meenchen: Assessed and retaining the current behavior intentionally. The exported configuration already self-identifies the unsupported combination with kv_cache_schema_version, kv_cache_quant_algo=MIXED_PRECISION, and the per-layer map. A new producer-only marker would be equally unknown to existing consumers, while an opt-in CLI would expand this focused composition interface without making the checkpoint deployable. The existing warning and README release gate remain explicit.

examples/hf_ptq/hf_ptq.py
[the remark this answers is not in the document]
meenchen: Follow-up in 83b17bcac supersedes my earlier reply: the precheck now preserves ordered enable state independently per parent scope rather than skipping all parent_class rows. It detects genuine parent-scoped K/V rules and qualified/dotless K/V patterns while ensuring unrelated scoped deny-all entries cannot clear the global K/V state. Focused regressions cover both cases.

modelopt/torch/quantization/model_quant.py
meenchen: Follow-up in 83b17bcac strengthens the earlier fix: checkpoint compatibility hashes the complete persisted properties and tensor state of every enabled non-K/V TensorQuantizer. Regressions cover format, constant_amax, axis, and calibrated-amax changes, plus successful identical-state restore.

tests/examples/hf_ptq/test_hf_ptq_args.py
```diff
@@ -145,6 +162,203 @@ def test_hf_ptq_kv_autoquant_invokes_public_api(monkeypatch):
...
+                ],
+                algorithm=None,
+                effective_bits=8.0,
+            )
+        ],
+        auto_quantize_method="kl_div",
+    )
+    recipe = ModelOptAutoQuantizeRecipe(auto_quantize=weight_aq, kv_auto_quantize=kv_aq)
+    calls = []
+    monkeypatch.setattr(hf_ptq, "auto_quantize", lambda *_args, **kwargs: calls.append(kwargs))
```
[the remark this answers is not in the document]
meenchen: Addressed in 83b17bcac with a real offline tiny-Qwen composition test using the shipped gradient-weight-AutoQuantize to mixed-KV-AutoQuantize recipe. It runs both actual stages without stage mocks (only the unavailable CUDA fake-quant op is stubbed) and asserts the selected weight quantizers survive and the K/V selections are applied.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -455,11 +457,41 @@ def _search_signature(
...
+
+def _preceding_quantizer_signature(model: nn.Module) -> list[dict[str, Any]]:
+    """Describe enabled non-K/V formats that affect KV sensitivity scores."""
+    return sorted(
+        (
+            {
+                "name": name,
+                "num_bits": copy.deepcopy(module.num_bits),
+                "block_sizes": copy.deepcopy(module.block_sizes),
+            }
```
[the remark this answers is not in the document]
meenchen: Addressed in 83b17bcac. The preceding baseline now hashes the complete get_modelopt_state(properties_only=True) payload and state_dict for every enabled non-K/V quantizer. Tests demonstrate rejection when constant_amax, axis, or calibrated amax changes, while an identical baseline restores without invoking scoring.

cjluo-nv (commented):

Nudge: the two blocking findings from the previous round are genuinely fixed with tests, but the third action item (pre-checking K/V for the weight-primary follow-up shape) is still unimplemented and unanswered for the third round.

Needs action:
- [ ] Run the K/V pre-check for the `quantize` + weight `auto_quantize` + `kv_auto_quantize` shape in `_run_auto_quantize_recipe` (`examples/hf_ptq/hf_ptq.py`): it is still gated on `primary_is_kv and fixed_quantize_config is not None`, so a fixed block enabling K/V survives the whole weight search and only trips the backstop `ValueError` in the follow-up stage. Or state why the late failure is acceptable.
- [ ] Confirm the stricter `_preceding_quantizer_signature` (`kv_cache_auto_quant.py`) does not break the documented resume flow — it hashes calibrated `_amax` state_dicts, so re-running stage 1 before resuming a KV search hard-rejects the checkpoint if calibration is not bitwise reproducible.
- [ ] Have a human review the agent-authored composition end to end (12 files, +995/-88, 7 directories).

No action needed:
- :heavy_check_mark: Resolved since the last review: the stale-KV-checkpoint reuse (signature now hashes full quantizer properties + state_dict, with constant_amax/axis/amax rejection tests) and the parent-scoped fail-open pre-check; a real non-mocked weight→KV composition test was also added.

modelopt/torch/export/quant_utils.py
```diff
@@ -1762,9 +1762,12 @@ def get_quant_config(
...
```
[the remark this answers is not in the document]
meenchen: Addressed in 211a72aa1 using the suggested metadata flag: unsupported uniform-weight + mixed-KV exports record `kv_cache_deployment_supported: false`, preserved through HF config conversion. The warning explicitly limits the artifact to inspection until the runtime supports this metadata path; the README documents the restriction. Export and conversion regressions pass.

modelopt/torch/quantization/kv_cache_auto_quant.py
```diff
@@ -455,11 +458,82 @@ def _search_signature(
...
+            {
+                "name": name,
+                "fingerprint": _quantizer_fingerprint(module),
+            }
+            for name, module in model.named_modules(remove_duplicate=False)
+            if isinstance(module, TensorQuantizer)
+            and module.is_enabled
+            and not name.endswith(_KV_QUANTIZER_ATTRS)
+        ),
+        key=lambda entry: entry["name"],
```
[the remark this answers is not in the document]
meenchen: Addressed the diagnostics/documentation in 211a72aa1 while retaining exact-state validation. Identical config with different calibrated scales can change KV sensitivities, so a config-only signature would permit stale scores. The error and README now explain that resume requires the exact preceding quantized state; otherwise a new checkpoint path recomputes KV sensitivities. Tests cover identical-state resume without rescoring and rejection after format, axis, constant-amax, or calibrated-amax changes. This does not make nondeterministic stage-1 reruns resumable.

examples/hf_ptq/hf_ptq.py
```diff
@@ -438,13 +444,55 @@ def _mtq_inputs_from_auto_quantize_config(
...
+
+        parent_class = entry.get("parent_class")
+        if parent_class is None:
+            scopes = enabled_by_parent.values()
+        else:
+            scopes = [enabled_by_parent.setdefault(parent_class, enabled_by_parent[None].copy())]
+        for enabled in scopes:
+            for name in matched_names:
+                enabled[name] = entry["enable"]
+    return any(any(enabled.values()) for enabled in enabled_by_parent.values())
```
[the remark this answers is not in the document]
meenchen: Addressed in 211a72aa1. The fixed-baseline precheck now runs before either composition starts, including weight-primary + KV follow-up. It also catches the reported `*k_bmm*`, `*self_attn.*`, and `*[kv]_bmm*` patterns. Regressions verify rejection before weight search and retain the MLP-scope and parent-scope cases; the converted-model check remains the final guard.

examples/hf_ptq/hf_ptq.py
```diff
@@ -870,6 +920,105 @@ def mono_quantize(
...
+
+    if args.layerwise_export:
+        assert_layerwise_export_compatible(args, full_model, quant_cfg.get("algorithm"))
+        quant_cfg = set_layerwise_export_dir(quant_cfg, args.export_path)
+        print(f"Layerwise export enabled: writing quantized shards to {args.export_path}")
+        quant_cfg, moved = default_layerwise_resume_dir(quant_cfg, args.export_path)
+        if moved:
+            print(
+                "Layerwise checkpoint_dir co-located with the export path so a resumed run "
+                "finds its manifest next to the shards it must not overwrite."
```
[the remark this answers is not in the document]
meenchen: Addressed in 211a72aa1: restored the MTP index-exclusion rationale, pre-conversion checkpoint discovery requirement, export-directory-before-hashing ordering, and colocated resume-manifest rationale. The existing AutoQuantize/layerwise rejection remains in `quantization_main`.

modelopt/torch/quantization/model_quant.py
```diff
@@ -428,6 +423,21 @@ def _auto_quantize_kv_cache(
...
+            for name, module in model.named_modules(remove_duplicate=False)
+            if name.endswith(("k_bmm_quantizer", "v_bmm_quantizer"))
+            and getattr(module, "is_enabled", False)
+        ]
+        if enabled_kv_quantizers:
+            raise ValueError(
+                "The preceding quantization stage left K/V quantizers enabled: "
+                f"{enabled_kv_quantizers}. Disable them before running KV-cache AutoQuantize; "
+                "clearing them now would not undo prior calibration or sensitivity measurements."
+            )
```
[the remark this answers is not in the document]
meenchen: Addressed in 211a72aa1: the guard now imports and uses `_KV_QUANTIZER_ATTRS`, matching the signature filter. The fresh-conversion guard remains because that path subsequently installs the deny-all baseline; composed inputs are checked before mutation. Fresh and composed public API tests pass.

meenchen:
Follow-up to [cjluo-nv's review](https://github.com/NVIDIA/Model-Optimizer/pull/2273#pullrequestreview-5212992298) and [Claude's review](https://github.com/NVIDIA/Model-Optimizer/pull/2273#pullrequestreview-5213222437): pushed 211a72aa1. The fixed-baseline K/V precheck now runs before weight-primary + KV-follow-up execution. Exact preceding calibration state remains required for KV checkpoint resume; the error and README explain when sensitivities must be recomputed. CHANGELOG.rst now covers composition, the dedicated KV checkpoint flag, and its deprecated legacy fallback.

Validation: 126 focused tests passed and changed-file pre-commit passed. A broader local run had 604 passes, one skip, and six failures from socket sandbox restrictions and local Transformers API incompatibility; it is not reported as a full-suite pass. Human owner review is still required.

The [CodeRabbit overlapping weight-rule finding](https://github.com/NVIDIA/Model-Optimizer/pull/2273#pullrequestreview-5204329911) is valid but also present in the main-branch base. It remains open for a separate focused fix; this commit does not address it.

modelopt_recipes/general/auto_quantize/fp8_ptq_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
```diff
@@ -0,0 +1,64 @@
...
+  candidate_formats:
+    - quant_cfg:
+        - quantizer_name: "*[kv]_bmm_quantizer"
+          cfg:
+            $import: fp8
+            constant_amax: 448.0
+      algorithm:
+      effective_bits: 8.0
+    - quant_cfg:
+        - quantizer_name: "*[kv]_bmm_quantizer"
```
juhi10071998, line 52: thanks @meenchen , is the expectation here users pass an already AutoQuant model and do the kv cache autoquantize in the next step?

cjluo-nv (commented):

Nudge: the last round's blocking pre-check gap is genuinely fixed with regression tests, leaving one documented design tradeoff and the owner sign-off on a 1,100-line agent-authored change.

Needs action:
- [ ] :speech_balloon: Author replied that exact preceding-state validation is intentional and now documented — confirm as owner that `_quantizer_fingerprint` hashing `state_dict()` (`kv_cache_auto_quant.py`) is acceptable, since any non-reproducible stage-1 calibration makes `--kv_auto_quantize_checkpoint` unresumable in practice for the two shipped composed recipes.
- [ ] Review the composition end to end as a human: 15 files, +1081/-90, spanning recipe schema, `hf_ptq.py` orchestration, `mtq.auto_quantize` preconditions and HF export metadata.
- [ ] Confirm the new `kv_cache_deployment_supported: false` marker plus `UserWarning` (replacing the former `NotImplementedError` in `quant_utils.get_quant_config`) is the behaviour you want for uniform-weight + mixed-KV exports.

No action needed:
- :heavy_check_mark: Resolved since the last review: the K/V pre-check now runs for `quantize` + weight `auto_quantize` + `kv_auto_quantize` (two new tests assert failure before either search starts), and CHANGELOG entries for the composition feature and the checkpoint-flag deprecation were added.

modelopt_recipes/general/auto_quantize/fp8_ptq_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
meenchen: Users can do both GEMM and KV quantization in a single recipe. Since GEMM and KV can have different formats and search algorithms, we will run the search in two stages.

shengliangxu (approved):
LGTM

modelopt_recipes/general/auto_quantize/nvfp4_fp8_gradient_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
realAsma, line 1: nice!

cjluo-nv (approved):

modelopt_recipes/general/auto_quantize/nvfp4_fp8_gradient_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml
```diff
@@ -0,0 +1,74 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
juhi10071998: hi @meenchen , given the gemm has it's own effective bits budget and the kv has it's own effective budget. I have one question and one recommendation-
Questions
1.  Can the KV$ quantizer budget and the gemm budget be different?

Should we keep the name like this to make it explicit what is the effective bits for each? _gemm and kv_
`nvfp4_fp8_gradient_at_5p4bits_then_kv_fp8_nvfp4_cast_kl_div_at_5p4bits.yaml`

cjluo-nv (commented):

Changes requested: 1,158 core-logic lines need splitting, and the previously flagged fixed-baseline preparation gap remains in the current branch.

Needs action:
- [ ] :scissors: Split into titled `[x/7]` PRs, starting with `[1/7]` composition/checkpoint support in `modelopt/torch/quantization`, `[2/7]` schema in `modelopt/recipe`, and `[3/7]` metadata in `modelopt/torch/export`.
- [ ] Divide `examples/hf_ptq` again: its 999 lines exceed the budget alone. Use `[4/7]` preparation extraction into `example_utils.py`, `[5/7]` translation-helper moves into `autoquant_utils.py`, `[6/7]` execution-helper moves, then `[7/7]` orchestration, recipes, and documentation.
- [ ] Keep the series acyclic and linearly ordered, document merge order, and link siblings in every description. Each PR must build independently, carry its relevant tests, and pass CI without later PRs; keep behavior changes separate from moves.
- [ ] Prepare weight-primary fixed baselines in `hf_ptq.py` and add MTP-exclusion/static-cast regressions; see inline.

No action needed:
- :heavy_check_mark: Resolved: checkpoint-state validation, early K/V rejection, export warning/marker, and changelog coverage. Exact-state resume is explicitly justified and documented.
- The stage-list/API-composition alternatives are addressed; existing Pydantic/OmegaConf recipe infrastructure is reused. Test moves and the public-API test rewrite match the changed behavior.

examples/hf_ptq/hf_ptq.py
```diff
@@ -870,59 +568,70 @@ def mono_quantize(
...
-            "--export_path."
-        )
-
-    if args.cast_mxfp4_to_nvfp4:
-        raise NotImplementedError(
-            "layerwise.export_dir is not compatible with --cast_mxfp4_to_nvfp4: the cast "
-            "rewrites weights after calibration, by which point every shard is written."
+    if primary_is_kv and fixed_quantize_config is not None:
+        quant_cfg = _prepare_quant_cfg(args, fixed_quantize_config.model_dump(), full_model)
+        mono_quantize(
```
cjluo-nv, line 596: > _Bot comment._

The earlier fixed-baseline preparation finding is still present in the fetched branch, despite the prior thread's addressed marker. `_prepare_quant_cfg` runs only for `primary_is_kv`; the weight-primary path passes the original `recipe.quantize` into `auto_quantize`. Consequently, a fixed baseline used outside `module_search_spaces` misses checkpoint-discovered `_mtp_layer_prefixes` exclusions and the `--cast_mxfp4_to_nvfp4` static-weight adjustment. Prepare the fixed baseline for both domains before dispatching, preserving its validated config type, and add tests asserting the exclusions/static adjustment reach the weight search. Keep the early K/V precheck before expensive execution.

meenchen:
/ok to test [f3673f9](https://github.com/NVIDIA/Model-Optimizer/pull/2273/commits/f3673f905334577d564f092ebcb605bf69e06834)

PR #2287, opened 2026-08-31
Add ModelOpt recipe for DeepSeek-V4-Pro-0813 NVFP4 and --recipe to its PTQ script

kevalmorabia97:
### What does this PR do?

Type of change: new feature

The quantization config for `nvidia/DeepSeek-V4-Pro-0813-NVFP4` existed only as Python inside `_build_nvfp4_experts_cfg()`, so the released checkpoint had **no entry in `modelopt_recipes/`** and could not be looked up by name the way every other published model can. `modelopt_recipes/README.md` states the goal directly — a recipe is *"the single, version-controlled source of truth for how a model is optimized … expressed as data instead of code"* — and this model was the exception.

This adds:

- `modelopt_recipes/huggingface/models/deepseek-ai/DeepSeek-V4-Pro-0813/ptq/nvfp4_experts_only.yaml`, composed from the existing `configs/ptq/units/base_disable_all` and `configs/numerics/nvfp4` units.
- An optional `--recipe` flag on `examples/deepseek/deepseek_v4/ptq.py`.

It follows **`examples/kimi/kimi_k3`**, the closest precedent: a very large MoE whose source already ships MXFP4 routed experts, converted via `--cast_mxfp4_to_nvfp4` rather than through `examples/hf_ptq`, and already wired to `--recipe` with a published YAML.

### Usage

```sh
torchrun --nproc-per-node 8 deepseek_v4/ptq.py \
    --model_path  <mp8_checkpoint> \
    --config      <DeepSeek-V4-Pro-0813>/inference/config.json \
    --calib_size  512 \
    --calib_seq   4096 \
    --output_path <amax_dump> \
    --recipe huggingface/models/deepseek-ai/DeepSeek-V4-Pro-0813/ptq/nvfp4_experts_only
```

Omitting `--recipe` keeps the previous behaviour exactly.

### Testing

- `load_recipe` resolves the YAML and yields `num_bits (2, 1)` with `block_sizes {-1: 16, type: dynamic, scale_bits: (4, 3)}` — identical to the hardcoded config.
- **Equivalence checked behaviourally**, not by eyeballing dicts: both configs were resolved against representative quantizer names using last-match-wins, and agree on all of them.

  | quantizer | hardcoded | recipe |
  | --- | --- | --- |
  | `...ffn.experts.17.w1_weight_quantizer` | enabled, NVFP4 | enabled, NVFP4 |
  | `...ffn.experts.17.w2_input_quantizer` | enabled, NVFP4 | enabled, NVFP4 |
  | `...ffn.shared_experts.w1_weight_quantizer` | disabled | disabled |
  | `...attn.wq_weight_quantizer` | disabled | disabled |
  | `mtp.0.ffn.experts.2.w1_weight_quantizer` | disabled | disabled |
  | `lm_head_weight_quantizer` | disabled | disabled |

- `mtq.quantize` documents `algorithm` as a string **or** a dict keyed on `method`, so the recipe's `{'method': 'max'}` needs no translation.
- `pre-commit` clean, including `validate modelopt recipes`.

No GPU run: this changes config plumbing only, and the default path is byte-identical to before.

- Is this change backward compatible?: ✅ — `--recipe` is optional and defaults to `None`; without it `_build_nvfp4_experts_cfg()` is used exactly as before.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅ — no new dependencies; `modelopt.recipe` is already a first-party import.
- Did you write any new necessary tests?: N/A — no new logic; equivalence to the existing config is the property that matters and is documented above.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — recipe library addition; recent recipe/example PRs add no entry.

### Additional Information

The recipe covers the **quant config only**. `--calib_seq` — the setting that mattered most for this checkpoint, since the 512 default does not cover long-context activation ranges — is a dataloader argument rather than part of the `mtq` config, so it stays on the CLI. Worth knowing if the recipe is ever treated as a complete reproduction of the released checkpoint: it is not, on its own.

## Summary by CodeRabbit

* **New Features**
  * Added post-training quantization support for DeepSeek-V4-Pro-0813 routed experts using NVFP4.
  * Added an optional recipe path for selecting equivalent quantization settings.
  * Preserved source formats for shared experts, attention, embeddings, output layers, and MTP components.

* **Bug Fixes**
  * Improved validation for missing or malformed quantization configurations.
  * Added safeguards against unsupported formats, scopes, algorithms, and enabled MTP quantizers.

* **Documentation**
  * Documented checkpoint conversion behavior, calibration requirements, and supported quantization workflows.

### DIFF
--- a/examples/deepseek/deepseek_v4/ptq.py
+++ b/examples/deepseek/deepseek_v4/ptq.py
@@ -93,6 +93,7 @@ from safetensors.torch import load_model
 from transformers import AutoTokenizer
 
 import modelopt.torch.quantization as mtq
+from modelopt.recipe import load_recipe
 from modelopt.torch.quantization.nn import TensorQuantizer
 from modelopt.torch.quantization.qtensor.mxfp4_tensor import MXFP4QTensor
 from modelopt.torch.utils.dataset_utils import get_dataset_dataloader
@@ -273,6 +274,18 @@ def load_deepseek_v4(
     return model
 
 
+_PUBLISHED_RECIPE = "huggingface/models/deepseek-ai/DeepSeek-V4-Pro-0813/ptq/nvfp4_experts_only"
+
+
+def _quant_cfg_from_recipe(recipe_path: str) -> dict:
+    """Load a ModelOpt recipe and return an ``mtq.quantize`` config.
+
+    ``mtq.quantize`` accepts ``algorithm`` as a string or as a dict keyed on
+    ``method``, which is the shape a recipe yields, so no translation is needed.
+    """
+    return load_recipe(recipe_path).quantize.model_dump()
+
+
 def _build_nvfp4_experts_cfg() -> dict:
     """Quant config: NVFP4 weight + NVFP4 input, routed experts only.
 
@@ -322,6 +335,7 @@ def ptq(
     calib_size: int,
     calib_datasets: list[str],
     calib_seq: int = 512,
+    recipe: str | None = None,
 ):
     world_size = int(os.getenv("WORLD_SIZE", "1"))
     rank = int(os.getenv("RANK", "0"))
@@ -357,7 +371,8 @@ def ptq(
         dist.barrier()
         _trace("pre-calib barrier done")
 
-    mtq_cfg = _build_nvfp4_experts_cfg()
+    mtq_cfg = _quant_cfg_from_recipe(recipe) if recipe else _build_nvfp4_experts_cfg()
+    _trace(f"quant config from {'recipe ' + recipe if recipe else 'built-in default'}")
     _trace("calling mtq.quantize")
     model = mtq.quantize(model, mtq_cfg, calibrate_loop)
     _trace("mtq.quantize returned")
@@ -462,6 +477,15 @@ def main():
         help="where to dump amax files and quantized_layers_manifest.json",
     )
     p.add_argument("--batch_size", type=int, default=4)
+    p.add_argument(
+        "--recipe",
+        default=None,
+        help=(
+            "ModelOpt recipe describing the quant config, loaded by path relative to "
+            f"modelopt_recipes/ (published: {_PUBLISHED_RECIPE}). Defaults to the "
+            "built-in routed-experts-only NVFP4 config, which this recipe mirrors."
+        ),
+    )
     p.add_argument("--calib_size", type=int, default=64)
     p.add_argument(
         "--calib_seq",
@@ -509,7 +533,13 @@ def main():
         args.model_path, trust_remote_code=args.trust_remote_code
     )
     model = ptq(
-        model, tokenizer, args.batch_size, args.calib_size, args.calib_datasets, args.calib_seq
+        model,
+        tokenizer,
+        args.batch_size,
+        args.calib_size,
+        args.calib_datasets,
+        args.calib_seq,
+        args.recipe,
     )
     save_amax_and_quant_config(model, args.output_path)
 
--- a/examples/deepseek/deepseek_v4/ptq.py
+++ b/examples/deepseek/deepseek_v4/ptq.py
@@ -282,8 +282,47 @@ def _quant_cfg_from_recipe(recipe_path: str) -> dict:
 
     ``mtq.quantize`` accepts ``algorithm`` as a string or as a dict keyed on
     ``method``, which is the shape a recipe yields, so no translation is needed.
+
+    The recipe is validated against what the rest of this pipeline can actually
+    export. ``save_amax_and_quant_config`` persists only routed-expert quantizer
+    state and writes a manifest hardcoded to ``NVFP4_W4A4``/``num_bits [2, 1]``/
+    ``block_size 16``, which ``quantize_to_nvfp4.py`` then consumes as ground
+    truth. A recipe that enables anything else, or asks for a different numeric
+    format, would produce a mislabeled manifest and silently drop quantizers --
+    so reject it here instead. Mirrors the guard in
+    ``examples/kimi/kimi_k3/quantize_to_nvfp4.py``.
     """
-    return load_recipe(recipe_path).quantize.model_dump()
+    cfg = load_recipe(recipe_path).quantize.model_dump()
+
+    algo = cfg.get("algorithm")
+    method = algo.get("method") if isinstance(algo, dict) else algo
+    if method != "max":
+        raise ValueError(f"recipe must use the 'max' calibration algorithm, got {method!r}")
+
+    enabled_experts = False
+    for entry in cfg.get("quant_cfg", []):
+        if not isinstance(entry, dict) or entry.get("enable") is not True:
+            continue
+        name = entry.get("quantizer_name", "")
+        if "ffn.experts." not in name:
+            raise ValueError(
+                f"recipe enables {name!r}, but this pipeline exports only routed-expert "
+                "quantizers (*ffn.experts.*); the manifest and quantize_to_nvfp4.py "
+                "would not represent it"
+            )
+        qcfg = entry.get("cfg") or {}
+        block_sizes = qcfg.get("block_sizes") or {}
+        if tuple(qcfg.get("num_bits") or ()) != (2, 1) or block_sizes.get(-1) != 16:
+            raise ValueError(
+                f"recipe entry {name!r} must be block-16 NVFP4 (num_bits (2, 1), "
+                f"block_sizes[-1] 16) to match the exported manifest, got "
+                f"num_bits={qcfg.get('num_bits')!r} block_sizes={block_sizes!r}"
+            )
+        enabled_experts = True
+
+    if not enabled_experts:
+        raise ValueError("recipe enables no routed-expert quantizers (*ffn.experts.*)")
+    return cfg
 
 
 def _build_nvfp4_experts_cfg() -> dict:
@@ -372,7 +411,7 @@ def ptq(
         _trace("pre-calib barrier done")
 
     mtq_cfg = _quant_cfg_from_recipe(recipe) if recipe else _build_nvfp4_experts_cfg()
-    _trace(f"quant config from {'recipe ' + recipe if recipe else 'built-in default'}")
+    _trace(f"quant config from {'recipe' if recipe else 'built-in default'}")
     _trace("calling mtq.quantize")
     model = mtq.quantize(model, mtq_cfg, calibrate_loop)
     _trace("mtq.quantize returned")
--- a/examples/deepseek/deepseek_v4/ptq.py
+++ b/examples/deepseek/deepseek_v4/ptq.py
@@ -293,7 +293,11 @@ def _quant_cfg_from_recipe(recipe_path: str) -> dict:
     the result with a manifest that does not match the amax it ships beside.
     Mirrors the guard in ``examples/kimi/kimi_k3/quantize_to_nvfp4.py``.
     """
-    cfg = load_recipe(recipe_path).quantize.model_dump()
+    recipe = load_recipe(recipe_path)
+    quantize = getattr(recipe, "quantize", None)
+    if quantize is None:
+        raise ValueError(f"recipe {recipe_path!r} has no 'quantize' section; expected a PTQ recipe")
+    cfg = quantize.model_dump()
 
     algo = cfg.get("algorithm")
     method = algo.get("method") if isinstance(algo, dict) else algo
@@ -312,6 +316,11 @@ def _quant_cfg_from_recipe(recipe_path: str) -> dict:
                 "and the manifest without any error"
             )
         qcfg = entry.get("cfg") or {}
+        if not isinstance(qcfg, dict):
+            raise ValueError(
+                f"recipe entry {name!r} has a non-dict 'cfg' ({type(qcfg).__name__}); "
+                "this pipeline supports only a single NVFP4 config per quantizer"
+            )
         block_sizes = qcfg.get("block_sizes") or {}
         if (
             tuple(qcfg.get("num_bits") or ()) != (2, 1)
@@ -436,9 +445,9 @@ def save_amax_and_quant_config(model, output_path: str):
     *not* rely on ``modelopt.torch.export.quant_utils.get_quant_config``
     because its introspection doesn't see weights stored on nested
     submodules — ``QuantExpert``'s ``w{1,2,3}`` are submodules of the
-    container, not direct parameters. The downstream export script
-    (``quantize_to_nvfp4.py``) uses this manifest as ground truth for which
-    tensor paths to replace with NVFP4 packed weight + scales.
+    container, not direct parameters. The manifest is a descriptive record of
+    what was calibrated: ``quantize_to_nvfp4.py`` does not read it, it
+    re-derives the expert tensor paths from the source checkpoint itself.
     """
 
     def _trace(msg):
--- a/examples/deepseek/deepseek_v4/ptq.py
+++ b/examples/deepseek/deepseek_v4/ptq.py
@@ -78,6 +78,7 @@ from __future__ import annotations
 
 import argparse
 import copy
+import fnmatch
 import json
 import os
 import re
@@ -277,6 +278,19 @@ def load_deepseek_v4(
 _PUBLISHED_RECIPE = "huggingface/models/deepseek-ai/DeepSeek-V4-Pro-0813/ptq/nvfp4_experts_only"
 
 
+_MTP_PROBE = "mtp.0.ffn.experts.0.w1_weight_quantizer"
+
+
+def _effective_enable(quant_cfg, name: str) -> bool:
+    """Whether ``name`` ends up enabled; mtq applies rules in order, so later ones win."""
+    enabled = False
+    for entry in quant_cfg:
+        if isinstance(entry, dict) and "quantizer_name" in entry:
+            if fnmatch.fnmatch(name, entry["quantizer_name"]):
+                enabled = entry.get("enable", True) is True
+    return enabled
+
+
 def _quant_cfg_from_recipe(recipe_path: str) -> dict:
     """Load a ModelOpt recipe and return an ``mtq.quantize`` config.
 
@@ -338,6 +352,16 @@ def _quant_cfg_from_recipe(recipe_path: str) -> dict:
 
     if not enabled_experts:
         raise ValueError("recipe enables no routed-expert quantizers (*ffn.experts.*)")
+
+    # The scope check above is a substring test, so a pattern such as
+    # *mtp.*ffn.experts.* satisfies it. MTP experts match save_amax's
+    # \.experts\.\d+\.w[123]_ regex too, so resolve the rules the way mtq applies
+    # them (in order, last match wins) and reject a recipe that leaves them on.
+    if _effective_enable(cfg.get("quant_cfg", []), _MTP_PROBE):
+        raise ValueError(
+            f"recipe enables MTP quantizers (matched {_MTP_PROBE!r}); this pipeline "
+            "leaves the MTP/DSpark block in its source format"
+        )
     return cfg
 
 
--- a/examples/deepseek/deepseek_v4/ptq.py
+++ b/examples/deepseek/deepseek_v4/ptq.py
@@ -285,12 +285,13 @@ def _quant_cfg_from_recipe(recipe_path: str) -> dict:
 
     The recipe is validated against what the rest of this pipeline can actually
     export. ``save_amax_and_quant_config`` persists only routed-expert quantizer
-    state and writes a manifest hardcoded to ``NVFP4_W4A4``/``num_bits [2, 1]``/
-    ``block_size 16``, which ``quantize_to_nvfp4.py`` then consumes as ground
-    truth. A recipe that enables anything else, or asks for a different numeric
-    format, would produce a mislabeled manifest and silently drop quantizers --
-    so reject it here instead. Mirrors the guard in
-    ``examples/kimi/kimi_k3/quantize_to_nvfp4.py``.
+    state and writes ``quantized_layers_manifest.json`` hardcoded to
+    ``NVFP4_W4A4``/``num_bits [2, 1]``/``block_size 16``/``scale_bits [4, 3]``,
+    and ``quantize_to_nvfp4.py`` re-derives expert paths itself and emits E4M3
+    block scales unconditionally. A recipe that enables anything else, or asks
+    for a different numeric format, would silently drop quantizers and describe
+    the result with a manifest that does not match the amax it ships beside.
+    Mirrors the guard in ``examples/kimi/kimi_k3/quantize_to_nvfp4.py``.
     """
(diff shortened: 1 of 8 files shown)


modelopt_recipes/huggingface/models/deepseek-ai/DeepSeek-V4-Pro-0813/ptq/nvfp4_experts_only.yaml
```diff
@@ -0,0 +1,63 @@
...
+# The activation amax collected here is therefore the only calibrated quantity
+# that reaches the exported checkpoint.
+#
+# Module paths follow DeepSeek's native (non-transformers) layout: routed experts
+# are model.layers.<i>.ffn.experts.<j>.w{1,2,3}. The shared expert is the same
+# Expert class under a different MoE attribute, so it is disabled explicitly.
+
+imports:
+  base_disable_all: configs/ptq/units/base_disable_all
+  nvfp4: configs/numerics/nvfp4
```
cjluo-nv, line 30: > _Bot comment._

Adding this file makes `tests/unit/recipe/test_recipe_docs.py::test_every_model_specific_ptq_dir_is_mentioned` fail: it walks `huggingface/**/ptq/*.yaml` and requires `yaml_path.parent.parent.name` (`DeepSeek-V4-Pro-0813`) to appear somewhere in `modelopt_recipes/ptq.md`, which this PR doesn't touch. Please add a bullet to the "Checkpoint mirrors — `models/<org>/<checkpoint>`" section (the `models/moonshotai/Kimi-K3/...` bullet is the closest model).

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -273,6 +274,18 @@ def load_deepseek_v4(
...
 
+_PUBLISHED_RECIPE = "huggingface/models/deepseek-ai/DeepSeek-V4-Pro-0813/ptq/nvfp4_experts_only"
+
+
+def _quant_cfg_from_recipe(recipe_path: str) -> dict:
+    """Load a ModelOpt recipe and return an ``mtq.quantize`` config.
+
+    ``mtq.quantize`` accepts ``algorithm`` as a string or as a dict keyed on
+    ``method``, which is the shape a recipe yields, so no translation is needed.
+    """
```
cjluo-nv, line 309: > _Bot comment._

Worth adding a unit test that pins this recipe to `_build_nvfp4_experts_cfg()`. The repo already does this in two places for the same reason: `tests/unit/recipe/test_presets.py::test_mlp_weight_only_recipe_matches_its_mtq_cfg` ("pin the equality so the recipe and the mtq constant cannot drift apart silently") and `tests/unit/recipe/test_kimi_k3_recipe.py`. Since the hardcoded builder stays as the default path, a drift here is invisible until someone diffs two amax dumps. The fnmatch-based equivalence table in the PR body is essentially the test body; note the recipe's `cfg` will additionally carry `effective_bits: 4.5` from `configs/numerics/nvfp4` (autoquant-only, ignored by PTQ), so compare on the fields that matter rather than exact dict equality.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -357,7 +371,8 @@ def calibrate_loop(model):
         dist.barrier()
         _trace("pre-calib barrier done")
 
-    mtq_cfg = _build_nvfp4_experts_cfg()
+    mtq_cfg = _quant_cfg_from_recipe(recipe) if recipe else _build_nvfp4_experts_cfg()
```
cjluo-nv, line 453: > _Bot comment._

`--recipe` accepts any recipe path, but the rest of the script assumes this specific one. `save_amax_and_quant_config()` filters quantizer state with `re.compile(r"\.experts\.\d+\.w[123]_")` and writes a manifest hardcoded to `"quantization_format": "NVFP4_W4A4"` / `num_bits [2,1]` / `block_size 16`, and `quantize_to_nvfp4.py` consumes that manifest as ground truth. Pass e.g. an FP8 or wider-scope recipe and you get a silently mislabeled manifest plus enabled quantizers that never make it into `quantized_layers` — no error anywhere. The Kimi-K3 sibling guards exactly this with `_conversion_settings_from_quantize_config()`, which raises if the recipe deviates from the supported contract. A lightweight version here (check `num_bits == (2,1)` and `block_sizes[-1] == 16` on the enabled expert entries, and that nothing outside `*ffn.experts.*` is enabled) — or deriving `layer_cfg` from the resolved cfg instead of hardcoding it — would close the gap.

cjluo-nv (commented):

Small, well-motivated change (move an inline quant config into `modelopt_recipes/` + optional `--recipe`), and the recipe body does mirror `_build_nvfp4_experts_cfg()` as claimed (`base_disable_all` → `*` disabled is equivalent given only `*input_quantizer`/`*weight_quantizer` exist; `configs/numerics/nvfp4` resolves to `num_bits (2,1)` + `block_sizes {-1:16, dynamic, e4m3}`; `algorithm: {method: max}` is accepted by `get_modelike_from_algo_cfg`). Three things should be fixed before merge:

1. **This will fail CI as-is.** `tests/unit/recipe/test_recipe_docs.py::test_every_model_specific_ptq_dir_is_mentioned` globs `huggingface/**/ptq/*.yaml` and asserts the checkpoint dir name (`DeepSeek-V4-Pro-0813`) appears in `modelopt_recipes/ptq.md`. `ptq.md` isn't touched by this PR and contains no DeepSeek entry, so the doc test fails. Add a bullet under "Checkpoint mirrors — `models/<org>/<checkpoint>`" (the Kimi-K3 bullet is the template).
2. **No test, despite the repo having exactly this test pattern for exactly this risk.** `tests/unit/recipe/test_presets.py::test_mlp_weight_only_recipe_matches_its_mtq_cfg` and `tests/unit/recipe/test_kimi_k3_recipe.py` exist precisely to pin a recipe against the code config it replaces. Here the hardcoded `_build_nvfp4_experts_cfg()` *stays in the file* as the default path, so the two can drift silently — the equivalence table in the PR body is the assertion that belongs in a test.
3. **`--recipe` is unvalidated while downstream export metadata is hardcoded** — see inline comment.

Minor: `examples/deepseek/README.md` documents the V4 calibration invocation but isn't updated with `--recipe` / the published recipe path, so the new flag is only discoverable from `--help`.

cjluo-nv (commented):

Re-review of #2287 (4 files, +280/-2). All three blocking points from the previous round are addressed:

- **ptq.md doc test** — fixed. The new "Checkpoint mirrors" bullet contains `DeepSeek-V4-Pro-0813`, which is exactly what `test_recipe_docs.py::test_every_model_specific_ptq_dir_is_mentioned` globs for (`yaml_path.parent.parent.name` under `huggingface/**/ptq/*.yaml`). The new recipe is also picked up automatically by `test_loader.py::test_shipped_ptq_recipe_algorithm_config_constructs`.
- **Recipe-vs-code drift test** — added (`tests/unit/recipe/test_deepseek_v4_recipe.py`), following the `test_presets.py` / `test_kimi_k3_recipe.py` precedent, with the `effective_bits` caveat handled by comparing fields instead of whole dicts.
- **Unvalidated `--recipe`** — `_quant_cfg_from_recipe` now rejects non-`max` algorithms, quantizers outside `*ffn.experts.*`, and non-block-16-NVFP4 entries, with three parametrized rejection tests. I verified the `enable is not True` check is sound: `QuantizerCfgEntry.enable` defaults to `True` and `model_dump()` always emits it, so a Kimi-style entry (`cfg:` with no explicit `enable`) is still caught. Passing `recipe.quantize.model_dump()` straight into `mtq.quantize` matches the existing `examples/minimax_m3` pattern.

Handing back to the owner for the residual items below rather than approving — all minor/moderate, none blocking on their own.

- :speech_balloon: Guard added in `16f3aa6` covers algorithm/scope/`num_bits`/`block_sizes[-1]` — still worth a look because it doesn't check `block_sizes["type"] == "dynamic"` or `scale_bits == (4, 3)`, and `QuantizerAttributeConfig.validate_num_bits` explicitly permits `(2, 1)` with *static* block quantization. That's CodeRabbit's still-open line-315 comment. Mitigating context the bot didn't have: for DeepSeek-V4, `quantize_to_nvfp4.py` never reads `quantized_layers_manifest.json` — it re-derives expert paths by regex from the source checkpoint and emits E4M3 block scales unconditionally — so the practical blast radius is smaller than "mislabeled manifest" suggests. Either tighten the two fields or drop the claim; right now the new docstring ("`quantize_to_nvfp4.py` then consumes as ground truth") overstates the coupling, echoing the same pre-existing inaccuracy in `save_amax_and_quant_config`'s docstring.
- :speech_balloon: The new drift test compares `num_bits`, `block_sizes[-1]` and `block_sizes["type"]`, but not `scale_bits`. Since the builtin hardcodes `scale_bits: (4, 3)` and the recipe inherits it from `configs/numerics/nvfp4`, a change to that shared unit would diverge from the builtin without failing the test — one more tuple element in `_resolve` closes it.
- Test placement: `tests/unit/recipe/` otherwise only exercises `modelopt.recipe`; the equivalent example-script guard tests for Kimi live in `tests/examples/kimi/test_kimi_k3_quantize_to_nvfp4.py`. This PR `exec_module`s `examples/deepseek/deepseek_v4/ptq.py` (pulling in transformers/safetensors) once per test from a unit-test dir. Owner call, but `tests/examples/deepseek/` looks like the established home for the three guard tests.
- The test's `RECIPE` literal duplicates `_PUBLISHED_RECIPE` in `ptq.py` (the Kimi test references `k3_cast._PUBLISHED_RECIPE`); as written, the `--recipe` help string can go stale without any test noticing.
- Prior minor nit still open: `examples/deepseek/README.md`'s "Calibrate routed experts" section doesn't mention `--recipe` or the published recipe path, so the flag remains discoverable only via `--help`.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -273,6 +274,57 @@ def load_deepseek_v4(
...
+        name = entry.get("quantizer_name", "")
+        if "ffn.experts." not in name:
+            raise ValueError(
+                f"recipe enables {name!r}, but this pipeline exports only routed-expert "
+                "quantizers (*ffn.experts.*); the manifest and quantize_to_nvfp4.py "
+                "would not represent it"
+            )
+        qcfg = entry.get("cfg") or {}
+        block_sizes = qcfg.get("block_sizes") or {}
+        if tuple(qcfg.get("num_bits") or ()) != (2, 1) or block_sizes.get(-1) != 16:
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in `897cc97`. `_quant_cfg_from_recipe` now also requires `block_sizes["type"] == "dynamic"` and `scale_bits == (4, 3)`, with rejection tests for each (`static-block-quantization`, `wrong-scale-bits`).

This was a real gap, not a theoretical one: swapping the recipe to the shipped `configs/numerics/nvfp4_static` passed the old guard and fails the new one.

One correction to the premise, though — `quantized_layers_manifest.json` is written by `ptq.py` and read by nothing in this repo; `quantize_to_nvfp4.py` re-derives expert paths by regex and emits E4M3 scales unconditionally. So the failure mode is inconsistent amax state rather than a mislabeled manifest being consumed downstream. My own docstring claimed the latter and was wrong; corrected in the same commit.

examples/deepseek/deepseek_v4/ptq.py
kevalmorabia97: Addressed in `16f3aa6`, tightened in `897cc97`. The guard rejects non-`max` algorithms, anything enabled outside `*ffn.experts.*`, and any entry that is not block-16 dynamic NVFP4 with E4M3 scales.

examples/deepseek/deepseek_v4/ptq.py
kevalmorabia97: Added in `16f3aa6`, moved in `897cc97` to `tests/examples/deepseek/test_deepseek_v4_ptq_recipe.py` — the Kimi-K3 analog is `tests/examples/kimi/`, and this test execs the example script rather than exercising `modelopt.recipe`. The `effective_bits` caveat was correct; the comparison pins `num_bits`, `block_sizes[-1]`, `type` and `scale_bits`.

modelopt_recipes/huggingface/models/deepseek-ai/DeepSeek-V4-Pro-0813/ptq/nvfp4_experts_only.yaml
kevalmorabia97: Fixed in `16f3aa6` — added the `DeepSeek-V4-Pro-0813` bullet under "Checkpoint mirrors" beside Kimi-K3. Confirmed by running `test_every_model_specific_ptq_dir_is_mentioned`, which did fail before the bullet.

cjluo-nv (commented):

Re-review of #2287 (5 files, +293/-2). Every blocking item from the previous rounds is now resolved, and I verified the mechanics against the repo:

**Addressed**
- `_quant_cfg_from_recipe` now rejects non-`max` algorithms, anything enabled outside `*ffn.experts.*`, and any entry that isn't block-16 dynamic NVFP4 with `scale_bits (4, 3)`; five parametrized rejection tests cover each branch. I confirmed the guard works on the shipped recipe: `configs/numerics/nvfp4` writes `num_bits: e2m1` / `scale_bits: e4m3`, and `QuantizerAttributeConfig` normalizes those to `(2, 1)` / `(4, 3)` before `model_dump()` (see `test_loader.py::test_import_multiple_snippets`), so the tuple comparisons hold.
- Drift test now pins `num_bits`, `block_sizes[-1]`, `type` **and** `scale_bits`, and references `dsv4_ptq._PUBLISHED_RECIPE` rather than a duplicated literal.
- `modelopt_recipes/ptq.md` has the `DeepSeek-V4-Pro-0813` checkpoint-mirror bullet that `test_recipe_docs.py::test_every_model_specific_ptq_dir_is_mentioned` globs for; `examples/deepseek/README.md` now documents `--recipe`.
- Licensing: new files carry the canonical `LICENSE_HEADER` text verbatim — no license concern.

**Residual — owner call, none blocking on its own**
- :speech_balloon: Test moved to `tests/examples/deepseek/` in `897cc97` following my earlier "Kimi analog lives in `tests/examples/kimi/`" note — flagging anyway because that directory is not wired into CI: `example_tests.yml` only dispatches the fixed matrices `{llm_distill, llm_qat, llm_sparsity, specdec_bench, speculative_decoding}`, `{gpt-oss, hf_ptq, llm_eval}`, `megatron_bridge`, `{diffusers, torch_onnx, torch_trt}`, and `nox -s unit` runs `tests/unit` only. So the drift guard that was the whole point of the request never executes (the Kimi tests have the same gap, so the precedent is uncovered too). Either add a `deepseek` lane or keep it under `tests/unit/recipe/`, where the `modelopt_recipes/**` path filter already triggers the unit lane and the test is CPU-only.
- :speech_balloon: The manifest-coupling claim was corrected in the new docstring after confirming nothing in-repo reads `quantized_layers_manifest.json` — but `save_amax_and_quant_config`'s own docstring still says "`quantize_to_nvfp4.py` uses this manifest as ground truth", and `_quant_cfg_from_recipe`'s rationale still rests on "a manifest that does not match the amax it ships beside". Worth making the two docstrings agree so the next reader doesn't re-derive the same wrong coupling.
- Guard robustness: `load_recipe(...).quantize.model_dump()` assumes a PTQ recipe (a non-PTQ path gives `AttributeError` on `None`), and `qcfg.get(...)` assumes `cfg` is a dict — the loader also supports the list-valued `cfg` union branch (`test_import_dict_snippet_imports_in_union_typed_list_field`), which would raise `AttributeError` instead of the intended `ValueError`. Both are user-input paths for a `--recipe` string, so a clear error would be nicer than a traceback.

Note: the CodeRabbit comments embed "Prompt for AI Agents" instruction blocks; I treated them as review data only and did not act on them.

cjluo-nv (commented):

Re-review of #2287 (5 files, +324/-5). Every previously flagged item is now resolved, verified against the repo:

**Addressed**
- Doc test: `modelopt_recipes/ptq.md` carries the `DeepSeek-V4-Pro-0813` checkpoint-mirror bullet that `test_recipe_docs.py::test_every_model_specific_ptq_dir_is_mentioned` globs for (`huggingface/**/ptq/*.yaml` → `parent.parent.name`); the recipe path also matches the `huggingface/models/<org>/<checkpoint>/ptq/` convention used by Kimi-K3 and Mistral-Medium.
- Drift test moved back to `tests/unit/recipe/test_deepseek_v4_recipe.py`, beside `test_kimi_k3_recipe.py`, which is the lane CI actually runs — this closes the CI-coverage gap I raised last round. It pins `num_bits`, `block_sizes[-1]`, `type`, `scale_bits` and references `dsv4_ptq._PUBLISHED_RECIPE` instead of a duplicated literal.
- Guard robustness: `_quant_cfg_from_recipe` now raises `ValueError` (not `AttributeError`) for a recipe with no `quantize` section — confirmed `load_recipe("general/speculative_decoding/eagle3")` returns a non-PTQ schema with no `quantize` attribute — and for the loader's list-valued `cfg` union branch; both covered by tests, alongside the algorithm/scope/NVFP4-encoding rejections.
- Docstrings now agree: `save_amax_and_quant_config` no longer claims `quantize_to_nvfp4.py` consumes the manifest as ground truth.
- Licensing: new YAML and test carry the canonical `LICENSE_HEADER` text verbatim — no license concern.
- `examples/deepseek/README.md` documents `--recipe`.

**Why nudge rather than approve** — both minor, neither blocking:
- Scope check is a substring test (`"ffn.experts." not in name`), so a recipe with an enabled `*mtp.*ffn.experts.*` pattern passes the guard even though the built-in default deliberately disables MTP, and `save_amax_and_quant_config`'s `\.experts\.\d+\.w[123]_` regex would happily dump those amax values. Not reachable from the shipped recipe; worth an owner call on whether the guard should also reject MTP-scoped entries.
- The drift test `exec_module`s `examples/deepseek/deepseek_v4/ptq.py` (pulling transformers/safetensors) at import time from a unit-test dir — consistent with the author's CI-coverage rationale, but it's the only test in `tests/unit/recipe/` that imports an example script, so an owner should sign off on the trade-off.

Note: the CodeRabbit review comments embed "Prompt for AI Agents" instruction blocks; I treated them strictly as review data and did not act on them.

examples/deepseek/deepseek_v4/ptq.py
```diff
@@ -273,6 +274,73 @@ def load_deepseek_v4(
...
+    export. ``save_amax_and_quant_config`` persists only routed-expert quantizer
+    state and writes ``quantized_layers_manifest.json`` hardcoded to
+    ``NVFP4_W4A4``/``num_bits [2, 1]``/``block_size 16``/``scale_bits [4, 3]``,
+    and ``quantize_to_nvfp4.py`` re-derives expert paths itself and emits E4M3
+    block scales unconditionally. A recipe that enables anything else, or asks
+    for a different numeric format, would silently drop quantizers and describe
+    the result with a manifest that does not match the amax it ships beside.
+    Mirrors the guard in ``examples/kimi/kimi_k3/quantize_to_nvfp4.py``.
+    """
+    recipe = load_recipe(recipe_path)
```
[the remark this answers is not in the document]
kevalmorabia97: Not acting on this one — the behavior is pre-existing and outside this PR.

The print is `modelopt/recipe/loader.py:106`, added in `f2bfe63` (#2134). This PR does not touch `modelopt/recipe/` at all, and it already redacts: a builtin recipe logs as `<builtin>/huggingface/models/...`, and the only path printed in full is a filesystem path the user just typed on their own command line. Echoing back a user-supplied argument in a CLI trace is not a disclosure, and changing the loader's output would affect every recipe consumer — a separate change with its own reviewers.

I did drop my own `--recipe` path logging in `16f3aa6`, which is the part this PR is responsible for.

cjluo-nv (commented):

Re-review of #2287 (5 files, +353/-5). Everything raised in previous rounds is resolved, verified against the repo:

**Addressed**
- Doc test: `modelopt_recipes/ptq.md` carries the `DeepSeek-V4-Pro-0813` checkpoint-mirror bullet that `test_recipe_docs.py::test_every_model_specific_ptq_dir_is_mentioned` globs for; the recipe path follows the `huggingface/models/<org>/<checkpoint>/ptq/` convention (Kimi-K3, Mistral-Medium).
- Drift guard lives in `tests/unit/recipe/test_deepseek_v4_recipe.py` (the lane CI actually runs), pins `num_bits`, `block_sizes[-1]`, `type` and `scale_bits`, and imports `dsv4_ptq._PUBLISHED_RECIPE` rather than duplicating the literal.
- `_quant_cfg_from_recipe` rejects non-`max` algorithms, enablement outside `*ffn.experts.*`, non-block-16 dynamic NVFP4 with non-`(4, 3)` scales, a missing `quantize` section, and list-valued `cfg` — all with `ValueError` and parametrized tests. I checked the shipped recipe still passes: `configs/ptq/units/base_disable_all` is a single `quantizer_name: '*' / enable: false` entry, and `configs/numerics/nvfp4` normalizes to `(2, 1)` / `(4, 3)` before `model_dump()`.
- MTP scope: the new `_effective_enable` + `_MTP_PROBE` last-match-wins resolution closes the substring-guard hole I raised last round; the shipped recipe's trailing `*mtp.*  enable: false` makes the probe resolve to disabled, and there's a `mtp-experts-enabled` rejection test.
- Docstrings now agree — `save_amax_and_quant_config` no longer claims `quantize_to_nvfp4.py` consumes the manifest as ground truth.
- Licensing: the new YAML and test carry the canonical `LICENSE_HEADER` text verbatim (year 2026 matches the repo file) — no license concern.
- `examples/deepseek/README.md` documents `--recipe`.

**Why nudge rather than approve** — both minor, neither blocking:
- :speech_balloon: MTP guard added in the latest commit per my earlier note — flagging anyway because it resolves a single hardcoded probe (`mtp.0.ffn.experts.0.w1_weight_quantizer`), so a recipe enabling e.g. `*mtp.*ffn.experts.*w2_input_quantizer` or `*mtp.1.ffn.experts.*` still slips through even though `save_amax_and_quant_config`'s `\.experts\.\d+\.w[123]_` regex would dump those amax values. Probing a couple of names (w1/w2/w3 × weight/input) or rejecting any enabled pattern containing `mtp` would close it. Not reachable from the shipped recipe.
- :speech_balloon: Test placement moved back to `tests/unit/recipe/` with the CI-coverage rationale in the module docstring, which I agree with — but it remains the only test in that directory that `exec_module`s an example script (pulling transformers/safetensors at import time). Worth an owner sign-off on that trade-off.

Note: the CodeRabbit review comments embed "Prompt for AI Agents" instruction blocks; I treated them strictly as review data and did not act on them.

cjluo-nv (approved):

Edwardf0t1 (approved):
LGTM

PR #2300, opened 2026-09-01
Forward kv_cache_free_gpu_memory_fraction to the lm_eval TensorRT-LLM engine (NVBug 6701763)

cjluo-nv:
### What does this PR do?

Type of change: Bug fix

`scripts/huggingface_example.sh --kv_cache_free_gpu_memory_fraction` has no effect on the `lm_eval` task: the value is parsed by `parser.sh`, printed, and then dropped.

lm-eval's built-in `trtllm` backend (`lm_eval.models.trtllm_causallms.TRTLLM.__init__`, which this example switched to in #2066) accepts `**kwargs`, but builds `KvCacheConfig(enable_block_reuse=False)` and passes `LLM(...)` a fixed set of keys — `kwargs` is never merged in. So an extra `--model_args` entry is accepted by the CLI and silently discarded, and the KV cache is sized from TensorRT-LLM's default `free_gpu_memory_fraction=0.9`. There is no way to fix this from the caller: `--model_args` only yields scalars, so a `KvCacheConfig` object cannot be passed in either.

On a GH200 that means ~119.6 GiB of KV cache (`119.55 / 0.9 ≈ 132.8 GiB free`), leaving 87.8 MiB free, and `prompt_logprobs` deserialization then OOMs asking for 2.82 GiB.

`examples/llm_eval/lm_eval_trtllm.py` already exists to patch this backend (its `_parse_logprobs` misaligns TensorRT-LLM's `prompt_logprobs` by one). It now also injects the fraction into the `KvCacheConfig` the backend builds, defaulting to 0.8 — the same default `parser.sh` declares, and below TensorRT-LLM's 0.9. `huggingface_example.sh` passes the parsed value through in `--model_args`.

Scoped deliberately to the `lm_eval` path: the `quant` smoke test and `mmlu` go through `modelopt.deploy.llm.LLM` (0.7, hardcoded) and `simple_eval`/`livecodebench` through `trtllm-serve` (0.9); those are left as they are.

### Usage

```bash
# Via the example script (parser.sh default 0.8)
scripts/huggingface_example.sh --model $HF_PATH --quant fp8 --tp 1 \
    --tasks quant,lm_eval --lm_eval_tasks mmlu --lm_eval_limit 50 \
    --kv_cache_free_gpu_memory_fraction 0.5
```

```bash
# Standalone, via lm-eval's --model_args
python lm_eval_trtllm.py --model trtllm \
    --model_args model=<ckpt>,tokenizer=<tok>,max_input_len=4096,kv_cache_free_gpu_memory_fraction=0.5 \
    --tasks mmlu --batch_size 8
```

### Testing

- `pytest tests/examples/llm_eval/test_lm_eval_trtllm.py` — 21 passed (lm-eval 0.4.12, no GPU).
- The new tests instantiate the **real** upstream `TRTLLM.__init__` through `create_from_arg_obj`, with `tensorrt_llm` and the tokenizer stubbed, and assert the engine receives `KvCacheConfig(enable_block_reuse=False, free_gpu_memory_fraction=0.5)`; that an unset key still yields 0.8 rather than 0.9; and that the patch does not outlive the constructor. Reverting the fix fails 3 of them.
- Tripwire test asserts upstream still neither declares nor forwards the argument, so this shim gets deleted rather than silently kept once lm-eval fixes it.
- `pre-commit run --files <changed>` clean (ruff, mypy, bandit, markdownlint); `bash -n` on the modified script.
- Not run: the GPU end-to-end `tests/examples/llm_eval/test_llm_eval.py::test_qwen3_eval_fp8`, which exercises `lm_eval` through the modified script — no GPU in this environment.

- Is this change backward compatible?: ✅ — the `lm_eval` KV cache goes from TensorRT-LLM's 0.9 to 0.8, which is strictly more conservative; `parser.sh`'s declared default is unchanged.

### Additional Information

NVBug 6701763. The 0.9 default on this path arrived with #2066 and was documented as a known limitation in `examples/llm_eval/README.md` ("the KV cache uses 90% of free GPU memory rather than 70%"); that note is replaced by the working knob.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the TensorRT-LLM evaluation workflow so `kv_cache_free_gpu_memory_fraction` is correctly passed to the backend.
  - The setting now defaults to `0.8`, providing more predictable GPU memory allocation for KV-cache usage.

- **Documentation**
  - Updated the TensorRT-LLM evaluation example and usage guidance to describe the KV-cache memory setting and its default behavior.
  - Updated the Hugging Face example to pass the configured KV-cache memory fraction.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -43,6 +43,7 @@ Changelog
 - Update HuggingFace checkpoint export to use name-based tied-weight deduplication instead of the previous address-based approach. The address-based deduplication could incorrectly drop an untied weight that happened to share memory with a tied one, producing an incomplete checkpoint (observed as a false positive on MiniMax-M2.7).
 - Fix EAGLE-3 training with context parallelism (``--cp_size > 1`` in ``examples/speculative_decoding``), which failed to start on ``accelerate >= 1.13`` and then raised ``got mixed torch.Tensor and DTensor``.
 - Polygraphy minimum dependency upgraded to ``0.53.4`` to solve ONNX AutoCast failures when marking optional graph outputs.
+- Fix ``--kv_cache_free_gpu_memory_fraction`` having no effect on the ``lm_eval`` task of ``examples/hf_ptq/scripts/huggingface_example.sh``, where the KV cache always took TensorRT-LLM's default 90% of free GPU memory and evaluation could run out of memory. ``examples/llm_eval/lm_eval_trtllm.py`` now takes ``kv_cache_free_gpu_memory_fraction`` in ``--model_args``, defaulting to 0.8.
 
 0.46 (2026-08-17)
 ^^^^^^^^^^^^^^^^^
--- a/examples/llm_eval/lm_eval_trtllm.py
+++ b/examples/llm_eval/lm_eval_trtllm.py
@@ -16,18 +16,22 @@
 """Run lm-evaluation-harness against a TensorRT-LLM checkpoint.
 
 Entry point around lm-eval's built-in ``trtllm`` backend
-(``lm_eval.models.trtllm_causallms``, new in 0.4.12). It exists only to correct that
-backend's ``prompt_logprobs`` handling -- everything else is upstream. Drop this file and
-call ``lm_eval`` directly once the fix lands upstream.
+(``lm_eval.models.trtllm_causallms``, new in 0.4.12). It exists to correct that backend's
+``prompt_logprobs`` handling and to forward ``kv_cache_free_gpu_memory_fraction`` to
+TensorRT-LLM -- everything else is upstream. Drop this file and call ``lm_eval`` directly
+once both land upstream.
 
     python lm_eval_trtllm.py --model trtllm \
         --model_args model=<quantized checkpoint dir>,tokenizer=<HF model folder>,\
-tensor_parallel_size=<tp>,max_batch_size=<max batch size>,max_input_len=4096 \
+tensor_parallel_size=<tp>,max_batch_size=<max batch size>,max_input_len=4096,\
+kv_cache_free_gpu_memory_fraction=0.8 \
         --tasks <comma separated tasks> --batch_size <max batch size>
 """
 
 import sys
+from functools import partial
 from importlib.metadata import version
+from importlib.util import find_spec
 
 from lm_eval.__main__ import cli_evaluate
 from packaging.version import Version
@@ -36,6 +40,7 @@ if Version(version("lm_eval")) < Version("0.4.12"):
     # 0.4.12 is the first release shipping lm_eval.models.trtllm_causallms.
     raise ImportError(f"lm_eval_trtllm.py requires lm-eval >= 0.4.12; found {version('lm_eval')}.")
 
+from lm_eval.models import trtllm_causallms
 from lm_eval.models.trtllm_causallms import TRTLLM
 
 # TensorRT-LLM only started passing the prompt token ids into `compute_logprobs` in
@@ -123,6 +128,42 @@ if not hasattr(TRTLLM, "_parse_logprobs"):
 # should be deleted in favour of calling `lm_eval` directly.
 _UPSTREAM_PARSE_LOGPROBS = TRTLLM._parse_logprobs
 TRTLLM._parse_logprobs = staticmethod(_parse_logprobs)
+
+
+# lm-eval's backend builds `KvCacheConfig(enable_block_reuse=False)` itself and passes
+# `LLM(...)` a fixed set of keys, dropping every other `--model_args` entry, so patching the
+# class it calls is the only way to size the KV cache. Left at TensorRT-LLM's default of
+# 0.9, the cache leaves too little room for the `prompt_logprobs` buffers and a
+# loglikelihood run dies with a CUDA OOM on a large-memory GPU.
+#
+# Rebinding a module global is process-wide, but `evaluator.py` builds exactly one model per
+# run, and nothing introspects the constructor it calls -- `create_from_arg_obj` just does
+# `cls(**model_args)` -- so nothing else can observe the swap.
+if find_spec("tensorrt_llm") and not hasattr(trtllm_causallms, "KvCacheConfig"):
+    # Guarded on find_spec because without tensorrt_llm the name is absent by design, and
+    # the backend's own "package is not installed" error is the useful one.
+    raise RuntimeError(
+        "lm_eval.models.trtllm_causallms no longer imports KvCacheConfig at module scope, so "
+        f"the KV cache size cannot be set; the backend changed shape in lm-eval "
+        f"{version('lm_eval')}. Recheck whether this file is still needed."
+    )
+
+_UPSTREAM_INIT = TRTLLM.__init__
+
+
+def _init(self, *args, kv_cache_free_gpu_memory_fraction: float = 0.8, **kwargs) -> None:
+    """``TRTLLM.__init__``, with the KV cache share of free GPU memory made settable."""
+    kv_cache_config = trtllm_causallms.KvCacheConfig
+    trtllm_causallms.KvCacheConfig = partial(
+        kv_cache_config, free_gpu_memory_fraction=float(kv_cache_free_gpu_memory_fraction)
+    )
+    try:
+        _UPSTREAM_INIT(self, *args, **kwargs)
+    finally:
+        trtllm_causallms.KvCacheConfig = kv_cache_config
+
+
+TRTLLM.__init__ = _init
 
 
 if __name__ == "__main__":
--- a/tests/examples/llm_eval/test_lm_eval_trtllm.py
+++ b/tests/examples/llm_eval/test_lm_eval_trtllm.py
@@ -20,12 +20,16 @@ TensorRT-LLM install is needed: ``_parse_logprobs`` is pure Python over a respon
 object, which these tests stub out.
 """
 
+import importlib.machinery
+import inspect
 import sys
 import types
 from dataclasses import dataclass
 from pathlib import Path
+from types import SimpleNamespace
 
 import pytest
+import transformers
 
 # Skip on the backend module, not just `lm_eval`: it currently guards its `tensorrt_llm`
 # import, but if that ever becomes eager this should skip rather than error.
@@ -38,6 +42,7 @@ if str(_LLM_EVAL_DIR) not in sys.path:
     sys.path.insert(0, str(_LLM_EVAL_DIR))
 
 import lm_eval_trtllm
+from lm_eval.models.trtllm_causallms import TRTLLM
 
 # TensorRT-LLM aligns prompt_logprobs to the *next* token, so entry i holds the
 # distribution that predicted tokens[i + 1]. These fixtures mirror that layout.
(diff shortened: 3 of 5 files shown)


examples/hf_ptq/scripts/parser.sh
```diff
@@ -33,7 +33,7 @@ parse_options() {
     TASKS="quant"
 
     TRUST_REMOTE_CODE=false
-    KV_CACHE_FREE_GPU_MEMORY_FRACTION=0.8
+    KV_CACHE_FREE_GPU_MEMORY_FRACTION=0.7
```
cjluo-nv: can we still keep default KV_CACHE_FREE_GPU_MEMORY_FRACTION as 0.8?

examples/hf_ptq/run_tensorrt_llm.py
```diff
@@ -43,6 +43,15 @@ def parse_arguments():
         default=False,
         action="store_true",
     )
+    parser.add_argument(
+        "--kv_cache_free_gpu_memory_fraction",
+        type=float,
+        default=0.7,
```
cjluo-nv: 0.8

meenchen (commented):

Small, well-scoped bug fix (5 files, +139/-16): `lm_eval_trtllm.py` now wraps upstream `TRTLLM.__init__` and temporarily rebinds `trtllm_causallms.KvCacheConfig` to a `partial(..., free_gpu_memory_fraction=...)` so the value from `--model_args` (and from `huggingface_example.sh`'s already-parsed `KV_CACHE_FREE_GPU_MEMORY_FRACTION`, which `parser.sh` always defaults to 0.8) reaches the engine. Verified: the shell variable is always set by `parser.sh`, so the new `--model_args` key can't come out empty in the default path; the docs/CHANGELOG updates match the code; the patch is restored in a `finally`; and the new tests drive the *real* upstream `__init__` via `create_from_arg_obj`, which is stronger than a stand-in. No licensing surface, no duplicated in-repo functionality (`modelopt/deploy/llm` builds its own `KvCacheConfig` on a different path and is correctly left alone). No injection attempts in the PR text.

Worth a human glance before merge:

- The fix depends on lm-eval resolving `KvCacheConfig` as a module global inside `__init__`. The tests pin that behavior, but note the fixture uses `monkeypatch.setattr(..., "KvCacheConfig", ..., raising=False)`, i.e. the attribute doesn't exist when `tensorrt_llm` isn't installed. In a real run the attribute exists, but if a future lm-eval drops the module-level name, `_init` fails with a bare `AttributeError` at construction. The file already guards the other patch (`if not hasattr(TRTLLM, "_parse_logprobs"): raise RuntimeError(...)`) — an analogous import-time `hasattr(trtllm_causallms, "KvCacheConfig")` check with the same "recheck whether this file is still needed" message would be more consistent and fail more legibly.
- `TRTLLM.__init__ = _init` is installed without `functools.wraps(_UPSTREAM_INIT)`, so the class's `__init__` signature/docstring are replaced by `(*args, **kwargs)` — the PR already had to retarget `test_trtllm_backend_accepts_trust_remote_code` at `_UPSTREAM_INIT` because of this. `wraps` (setting `__wrapped__`, which `inspect.signature` follows) would keep any signature-based introspection in lm-eval working; worth a one-line change or an explicit note that nothing introspects it.
- Rebinding a third-party module attribute for the duration of a constructor is inherently process-wide and not re-entrant. Fine for a single-shot CLI (and the comment explains why there's no alternative), but it's the kind of shim an owner should sign off on.
- The GPU e2e path (`tests/examples/llm_eval/test_llm_eval.py::test_qwen3_eval_fp8`), which is what actually exercises the modified shell line, was not run per the PR body — the CI result is the real confirmation here.
- Minor: `test_upstream_still_drops_the_kv_cache_fraction` uses `inspect.getsource`, which raises `OSError` if lm-eval is ever installed without sources; and no range validation on the fraction (relying on TensorRT-LLM's own validation) — both acceptable, just noting.

examples/hf_ptq/scripts/parser.sh
cjluo-nv: Done — `parser.sh` is back to `origin/main`, so the default stays 0.8. The only remaining change in `huggingface_example.sh` is the `kv_cache_free_gpu_memory_fraction=` key appended to the `lm_eval` `--model_args`.

examples/hf_ptq/run_tensorrt_llm.py
cjluo-nv: Done — `run_tensorrt_llm.py` is reverted entirely, along with `mmlu.py`, `modelopt/deploy/llm/generate.py` (0.7 stays hardcoded) and the `trtllm-serve` line. This PR is now scoped to the `lm_eval` path only. `lm_eval_trtllm.py` defaults the fraction to 0.8, matching `parser.sh`.

CHANGELOG.rst
```diff
@@ -43,6 +43,7 @@ Changelog
...
 - Fix EAGLE-3 training with context parallelism (``--cp_size > 1`` in ``examples/speculative_decoding``), which failed to start on ``accelerate >= 1.13`` and then raised ``got mixed torch.Tensor and DTensor``.
 - Polygraphy minimum dependency upgraded to ``0.53.4`` to solve ONNX AutoCast failures when marking optional graph outputs.
+- Fix ``--kv_cache_free_gpu_memory_fraction`` in ``examples/hf_ptq/scripts/huggingface_example.sh``: it was parsed but never forwarded, so every TensorRT-LLM engine the script deploys sized its KV cache from TensorRT-LLM's default 90% of free GPU memory and evaluation could run out of memory. The value now reaches the ``quant`` smoke test, ``lm_eval``, ``mmlu`` and ``simple_eval``/``livecodebench`` paths and defaults to 0.7; ``modelopt.deploy.llm.LLM`` and ``lm_eval_trtllm.py``'s ``--model_args`` accept ``kv_cache_free_gpu_memory_fraction`` directly.
```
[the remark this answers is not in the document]
cjluo-nv: Addressed in `466332982`. Dropped the root-cause clause; the entry is now two sentences naming the user-visible flag and its default:

> Fix `--kv_cache_free_gpu_memory_fraction` having no effect on the `lm_eval` task of `examples/hf_ptq/scripts/huggingface_example.sh`, where the KV cache always took TensorRT-LLM's default 90% of free GPU memory and evaluation could run out of memory. `examples/llm_eval/lm_eval_trtllm.py` now takes `kv_cache_free_gpu_memory_fraction` in `--model_args`, defaulting to 0.8.

The two file names are kept deliberately: both are user-facing entry points (the script that exposes the flag, and the script to pass it to directly), not implementation detail. Note the suggested "defaults to 0.7" was based on an earlier revision — the default is 0.8, matching `parser.sh`.

tests/examples/llm_eval/test_lm_eval_trtllm.py
```diff
@@ -191,6 +197,120 @@ def test_version_is_checked_before_scoring(monkeypatch):
...
+
+    assert lm_eval_trtllm.trtllm_causallms.KvCacheConfig is _RecordingKvCacheConfig
+
+
+def test_kv_cache_fraction_without_tensorrt_llm(monkeypatch):
+    """Without tensorrt_llm the backend raises its own error; don't pre-empt it here."""
+    monkeypatch.delattr(lm_eval_trtllm.trtllm_causallms, "KvCacheConfig", raising=False)
+
+    with lm_eval_trtllm._kv_cache_fraction_applied(0.5):
+        pass
```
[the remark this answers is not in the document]
cjluo-nv: Obsolete — this reviewed an earlier revision. `_kv_cache_fraction_applied` and `test_kv_cache_fraction_without_tensorrt_llm` no longer exist; the context manager was inlined into `_init` when the shim was simplified.

The missing-TensorRT-LLM concern is now handled at import instead: `if find_spec("tensorrt_llm") and not hasattr(trtllm_causallms, "KvCacheConfig")` raises with an explanation, and the `find_spec` condition leaves the backend's own "package `tensorrt_llm` is not installed" error as the one users see when it is simply absent.

examples/llm_eval/lm_eval_trtllm.py
```diff
@@ -125,6 +129,29 @@ def _parse_logprobs(tokens: list[int], outputs, ctxlen: int) -> tuple[float, boo
...
+# loglikelihood run dies with a CUDA OOM on a large-memory GPU.
+_UPSTREAM_INIT = TRTLLM.__init__
+
+
+def _init(self, *args, kv_cache_free_gpu_memory_fraction: float = 0.8, **kwargs) -> None:
+    """``TRTLLM.__init__``, with the KV cache share of free GPU memory made settable."""
+    kv_cache_config = trtllm_causallms.KvCacheConfig
+    trtllm_causallms.KvCacheConfig = partial(
+        kv_cache_config, free_gpu_memory_fraction=float(kv_cache_free_gpu_memory_fraction)
+    )
```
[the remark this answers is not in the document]
cjluo-nv: Not taking the lock — the interleaving it guards against has no code path here.

lm-eval constructs the model exactly once per run, at a single site (`lm_eval/evaluator.py:242`, `create_from_arg_obj`, or the `create_from_arg_string` branch below it), with no threading anywhere around it; `cli_evaluate` is a single-shot CLI. Two concurrent `TRTLLM` constructions would also mean two TensorRT-LLM engines racing for the same GPU memory in one process, which is not a supported way to use the backend regardless of this shim. So an `RLock` would serialize something that cannot happen, and the requested concurrency regression test would assert against a scenario the library does not produce.

`466332982` documents the constraint at the patch site instead:

```
# Rebinding a module global is process-wide, but `evaluator.py` builds exactly one model per
# run, and nothing introspects the constructor it calls -- `create_from_arg_obj` just does
# `cls(**model_args)` -- so nothing else can observe the swap.
```

Happy to revisit if lm-eval ever grows a path that builds backends concurrently.

meenchen (commented):

Re-review of the `kv_cache_free_gpu_memory_fraction` forwarding fix (5 files, +153/-16) at the current head. All previously raised points are resolved:

- **Import-time guard for `KvCacheConfig` (critical, addressed).** `lm_eval_trtllm.py` now raises a legible `RuntimeError` at import if `trtllm_causallms` stops exposing `KvCacheConfig` at module scope, mirroring the existing `_parse_logprobs` guard. The `find_spec("tensorrt_llm")` condition is correct: without the package the name is absent by design, so the backend's own "package is not installed" error still surfaces (this also makes CodeRabbit's "dereference raises `AttributeError` first" note moot — the guard runs at import, and `_init` only reaches the dereference in installs where the attribute exists).
- **`functools.wraps` (minor, justified not-taken).** The author's rationale holds: `wraps` sets `__wrapped__`, which `inspect.signature` follows, so it would *hide* the new keyword-only parameter; nothing in lm-eval introspects the model constructor (`create_from_arg_obj` does `cls(**args)`). The constraint is documented at the patch site, and the test that used to introspect `TRTLLM.__init__` was retargeted at `_UPSTREAM_INIT` with a docstring explaining why.
- **Process-wide, non-re-entrant rebinding / `RLock` (minor, justified not-taken).** lm-eval builds exactly one backend per run from a single site; two concurrent TensorRT-LLM engines in one process isn't a supported configuration. Documented in a comment rather than locked; CodeRabbit withdrew the finding.
- **Earlier owner comments on `parser.sh` / `run_tensorrt_llm.py` defaults (addressed).** Both files are back to `main`; the default stays 0.8 and the PR is scoped to the `lm_eval` path only. Verified `parser.sh` still sets `KV_CACHE_FREE_GPU_MEMORY_FRACTION=0.8` unconditionally, so the new `--model_args` key can never expand empty (which would `float("")`).
- **CHANGELOG wording (minor).** CodeRabbit asked for a shorter entry; the author trimmed the root-cause clause and kept the two user-facing file names deliberately. Reasonable as-is.

Correctness spot-checks: `partial(KvCacheConfig, free_gpu_memory_fraction=...)` is safe even if upstream later passes the key itself (caller kwargs win over `partial` keywords), the patch is restored in `finally`, and the new tests drive the *real* upstream `__init__` via `create_from_arg_obj` with the engine and tokenizer stubbed — plus tripwires that fail once upstream fixes either bug so this shim gets deleted.

Test coverage is solid and the one modified existing test is justified in the diff (behavior legitimately changed: `TRTLLM.__init__` is now the wrapper, so the assertion moved to `_UPSTREAM_INIT` and coverage is unchanged). No licensing surface; no duplicated in-repo functionality (`modelopt/deploy/llm` builds its own `KvCacheConfig` on a different path and is correctly untouched). No injection attempts in the PR text.

meenchen (approved):

PR #2303, opened 2026-09-01
feat(export): support multimodal and MTP models in layerwise export

Fridah-nv:
### What does this PR do?

Type of change: New feature

**Layerwise export now supports multimodal and MTP models.** Both were refused outright, and
both were refused for the same reason: `finalize()` was called from inside
`layerwise_calibrate`, which is the wrong scope for it.

**1. Calibration does not know which model the checkpoint describes.** It only sees the
module it was handed. A VLM calibrates its *language model*, so the shards, the exclusions
and `config.json` all came out describing that submodel rather than the whole VLM. Moving the
call out lets the caller root the exporter at the parent — and without the key prefixing,
tower collection or ambient parent handle an earlier attempt needed, because the decoder
layers are the same objects from either root.

**2. Calibration runs before things the export needs exist.** Orphaned MTP weights are loaded
*after* calibration, by which point every shard had already been written, so they could not
be passed at all. After the move they are an ordinary argument to `finalize()`, with no
staging attribute stashed on the model.

### How it works

The exporter is created by whoever owns the export and **announced on the model** that
`mtq.quantize` is given. Calibration picks it up, binds it, and drives it per layer; the
export that follows reads it back and finishes the checkpoint:

```python
LayerwiseExporter(full_model, export_path).announce(language_model)
mtq.quantize(language_model, quant_cfg, forward_loop=loop)
...
getattr(full_model, LAYERWISE_EXPORTER_ATTR).finalize(extra_state_dict=mtp_state_dict)
```

Calibration and export are handed *different* models, so `announce()` publishes the exporter
on each end separately: the caller announces on the model being calibrated, and `bind()`
announces on the export root. Neither side has to know where the other looked, and the lookup
stays an O(1) `getattr` rather than a `named_modules()` scan — worth avoiding at roughly
1.65 µs/module, or ~500 ms on a Kimi-K3-sized model. For a non-VLM both roots are the same
object and the second announcement is a no-op. `finalize()` clears every attachment it
recorded, so the module graph does not retain a live exporter afterwards.

`mtq.quantize` and `mtq.calibrate` are **unchanged** — a layerwise-only feature does not
belong in the public quantization API. The attribute follows `_mtp_layer_prefixes`, which
crosses the same calibration→export boundary the same way (`hf_ptq.py:538` sets it,
`unified_export_hf.py:870` reads it back).

Construction is inert: `__init__` records only the export root and the directory, because the
caller builds it before `mtq.quantize`, when there are no quantizers yet to validate or read
a config from. `bind()` does that, called from calibration after quantizer insertion and
before any layer is converted — the only window where both hold, and the same instant the
exporter used to be constructed, so unsupported models still fail in seconds rather than
hours. Only the calibration pass that sets `export_dir` drives the exporter: a list-form
algorithm runs one pass per entry, and an earlier one must not convert layers a later one
still has to calibrate.

### Usage

Nothing changes for a plain layerwise-export recipe: `layerwise.export_dir` still drives it.
Pre-attaching an exporter is the opt-in for the two cases that need it — a checkpoint whose
root is wider than the calibrated model, and orphaned tensors to merge at the end.

The one behaviour change for a config-only caller is that `mtq.quantize` now writes the layer
shards but no longer finishes the checkpoint. Both exit paths warn with what is still owed,
and `LayerwiseConfig.export_dir`'s description has been corrected — it previously promised "a
complete, loadable checkpoint when the last layer lands" and still listed multimodal and MTP
as raising `NotImplementedError`.

### Testing

`tests/gpu/torch/export/test_layerwise_export.py` — **29 passed**. Beyond the 24 inherited
from #2136, five new ones, each with a negative control confirming it fails without its fix:

- orphaned MTP tensors reach the tail shard *and* the index
- an exporter rooted at the parent widens the checkpoint's namespace
- the config-only path announces an exporter that can be finished, and finalize clears it
- only the pass that sets `export_dir` drives the exporter
- an exporter whose root holds a different number of layers is refused at `bind()`

Full suites: `tests/gpu/torch/export` + `tests/gpu/torch/quantization` **1012 passed / 55
skipped**, `tests/unit` **3318 passed / 15 skipped**, pre-commit clean. Both suites also
report failures in `test_implicit_gemm.py` (FP4 conv kernels), `test_triton_fa_p_qdq.py`,
`test_autocast_quantize_int8` and `test_engine_builder.py` collection; all reproduce unchanged
on `main` and none touch the paths in this diff.

Measured against the whole-model exporter on a tiny Gemma3-VL, towers prepared exactly as
`hf_ptq` does:

```
keys: baseline=80  layerwise=80   only-baseline=[]  only-layerwise=[]
differing values: 0
vision tower present: True     VLM namespace: True
config.json is the VLM: True   hf_quant_config match: True
exclude_modules: ['language_model.lm_head', 'vision_tower.vision_model*']   (both sides)
```

#### End-to-end through `hf_ptq.py`

Same FP8 recipe both sides; the baseline drops `layerwise.export_dir` and is exported by
`main`, so the diff isolates this PR. Every tensor matches in key, dtype, shape and value,
and `config.json` / `hf_quant_config.json` match too.

| Model | Covers | Keys | Differing |
|---|---|---|---|
| Qwen3-VL-8B-Instruct | multimodal | 1254 = 1254 | 0 |
| GLM-4.7-Flash | MoE + MTP | 28119 = 28119 | 0 |

The VLM checkpoint keeps the vision tower unquantized (351 `model.visual.*` keys, no
`weight_scale` among them) while the language model is FP8. The MTP run reports 212 orphaned
tensors; all 212 land in `model-tail.safetensors` and in the index, with `model.layers.47*` in
`exclude_modules`.

**Not yet validated:** an accelerate-offloaded run, and a serving canary on the exported
checkpoints.

### Refusals

`export_dir` without `enable`, and an exporting algorithm entry with no calibration method,
are both refused before calibration starts — neither reaches the per-layer pass, so both
would otherwise export nothing. The early gate is a heuristic on the recipe, so `hf_ptq` also
raises a plain `RuntimeError` at export time if calibration turned out not to have run; that
backstop, not the gate, is what makes the failure legible on paths the recipe check cannot
predict.

`bind()` requires the layers calibration will drive and refuses a root that discovers a
different number of them. Only the count is checked here: `export_layer` already rejects a
reordering or a substituted module on its first call, and a length difference is the one
mismatch it structurally cannot catch — every call would pass and `_write_index` would then
open a shard that was never written, at the very end of the run.

Orphan tensors are merged into the tail with no collision check, matching the whole-model path
(`unified_export_hf.py:1623`). `load_mtp_weights` returns exactly the keys absent from
`model.state_dict()`, so a collision with an exported tensor is not reachable through the only
producer, and a guard would only make the two export paths diverge.

### Why not reuse `export_hf_checkpoint`

It was the first idea and it is the most expensive one. Its transformers path is whole-model
at every step — `_prepare_moe_inputs`, `requantize_resmooth_fused_llm_layers` (which runs a
dummy forward that would fail on already-converted layers), `_process_quantized_modules`, a
full `model.state_dict()` in host RAM, then `save_pretrained` rewriting shards already on disk
— and it raises outright under `has_accelerate_offload`. `save_pretrained(state_dict={})` is
not an escape either: safetensors' shared-storage check fires on MoE even with an empty dict.

The natural consolidation target is the **streaming** exporter, which is already most of
`finalize()`: 122 lines vs 74, sharing `decoder_owned_ids`,
`enable_weight_access_and_writeback`, `_dispatch_export_handler`,
`_reconstruct_fused_moe_linear`, `_add_mtp_exclusions`, `_postprocess_single_tensor`,
`requires_weight_materialization` and `save_non_weight_artifacts`. Folding them together needs
roughly four knobs: skip the whole-model prep, skip layers already written, seed the index
with the existing shards, and inject the quant config. That is a separate change and
deliberately not in this one.

- Is this change backward compatible?: ✅ — `mtq.quantize`/`mtq.calibrate` signatures are
  unchanged, and a recipe that only sets `layerwise.export_dir` behaves as before. The one
  behaviour change is that `mtq.quantize` no longer finishes the checkpoint on its own:
  callers must now call `finalize()` on the exporter, which calibration leaves on the model.
  guidance in `CONTRIBUTING.md`: N/A
- Did you update Changelog?: ❌ — pending.
- Did you get Claude approval on this PR?: ❌ — the last review's findings are all addressed;
  needs a re-run.

### Additional Information

Follow-ups this enables: #2259 (MTP) reduces to close to nothing, and the multimodal work in
#2218 no longer needs `export_parent`, the key prefixing, or the tower collection.

## Summary by CodeRabbit

- **New Features**
  - Layerwise export now supports clearer control over export locations and calibrated layer handling.
  - Export workflows provide improved support for resuming, sharded checkpoints, mixture-of-experts models, and nested model namespaces.

- **Bug Fixes**
  - Improved handling of exported checkpoint shards and extra tensors.
  - Added clearer warnings when exports require completion before loading.

- **Documentation**
  - Clarified that layerwise exports write shards during calibration and require an explicit finalization step.
  - Documented that the in-memory model is not suitable for inference after layerwise export.

### DIFF
--- a/examples/hf_ptq/hf_ptq.py
+++ b/examples/hf_ptq/hf_ptq.py
@@ -40,6 +40,7 @@ from example_utils import (
     get_tokenizer,
     is_enc_dec,
     is_nemotron_vl,
+    layerwise_export_block,
     load_mtp_weights,
     mlflow_run,
     mtp_layer_prefixes_from_checkpoint,
@@ -48,6 +49,8 @@ from example_utils import (
     resolve_checkpoint_dir,
     resolve_mlflow_args,
     run_nemotron_vl_preview,
+    save_processor_config,
+    save_source_config,
     set_layerwise_export_dir,
     setup_distributed_args,
     validate_fsdp2_supported,
@@ -78,6 +81,7 @@ from modelopt.torch.export import (
     has_spec_opt,
     save_expert_token_count_table,
 )
+from modelopt.torch.export.layerwise_export import LayerwiseExporter
 from modelopt.torch.export.model_utils import get_language_model_from_vl, is_multimodal_model
 from modelopt.torch.quantization.config import need_calibration
 from modelopt.torch.quantization.plugins.accelerate import init_quantized_weights
@@ -778,6 +782,9 @@ def mono_quantize(
                     else None,
                 )
 
+        if args.layerwise_export:
+            LayerwiseExporter(full_model, args.export_path)
+
         if calibration_only:
             language_model = mtq.calibrate(
                 language_model, quant_cfg["algorithm"], forward_loop=calibrate_loop
@@ -796,27 +803,23 @@ def mono_quantize(
         warnings.warn("Skipping quantization: model is already quantized.")
 
 
-def assert_layerwise_export_compatible(args, full_model, mtp_layer_prefixes) -> None:
-    """Refuse layerwise export before calibration starts, not after it writes a checkpoint.
+def assert_layerwise_export_compatible(args, full_model, algorithm) -> None:
+    """Refuse layerwise export before calibration starts, not after the run is paid for.
 
-    Layerwise export writes the finished checkpoint during calibration, so anything that
-    would rewrite or contradict that checkpoint afterwards has to be caught here -- once
-    calibration begins, the user has already paid for the whole run.
+    Layerwise export writes each layer's shard during calibration and finishes the checkpoint
+    in finalize() afterwards, so anything that would rewrite or contradict that checkpoint has
+    to be caught here -- once calibration begins, the user has already paid for the whole run.
     """
-    if is_multimodal_model(full_model):
-        raise NotImplementedError(
-            "layerwise.export_dir does not support multimodal models: calibration runs on the "
-            "extracted language model, so the shards and config.json would describe that "
-            "submodel rather than the full VLM, and the VLM export path would then "
-            "overwrite config.json with the unquantized source config."
-        )
-
-    if mtp_layer_prefixes:
-        raise NotImplementedError(
-            f"layerwise.export_dir does not support models with MTP layers {mtp_layer_prefixes}: "
-            "their exclusions and any orphaned MTP weights are applied after calibration, by "
-            "which point every shard and the quant config are already written."
-        )
+    block = layerwise_export_block(algorithm)
+    if block is not None:
+        entries = algorithm if isinstance(algorithm, list) else [algorithm]
+        owner = next(e for e in entries if isinstance(e, dict) and e.get("layerwise") is block)
+        if not owner.get("method"):
+            raise NotImplementedError(
+                "layerwise.export_dir needs a calibration method: without one there is no "
+                "per-layer pass to write the shards, so the export would find nothing. Set "
+                "algorithm.method, or export without layerwise.export_dir."
+            )
 
     if has_spec_opt(full_model):
         raise NotImplementedError(
@@ -880,29 +883,12 @@ def export_quantized(
             print(f"Quantized speculative decoding checkpoint exported to: {export_path}")
             return
 
-        # Check if the model is a multimodal/VLM model
-        is_vlm = is_multimodal_model(full_model)
-
-        if is_vlm:
-            # Save original model config and the processor config to the export path for VLMs.
-            print(f"Saving original model config to {export_path}")
-
-            config_kwargs = {"trust_remote_code": args.trust_remote_code}
-            if args.attn_implementation is not None:
-                config_kwargs["attn_implementation"] = args.attn_implementation
-            AutoConfig.from_pretrained(args.pyt_ckpt_path, **config_kwargs).save_pretrained(
-                export_path
-            )
-
-            # Try to save processor config if available
-            try:
-                print(f"Saving processor config to {export_path}")
-                AutoProcessor.from_pretrained(
-                    args.pyt_ckpt_path, trust_remote_code=args.trust_remote_code
-                ).save_pretrained(export_path)
-            except Exception as e:
-                print(f"Warning: Could not save processor config: {e}")
-                print("This is normal for some VLM architectures that don't use AutoProcessor")
+        if is_multimodal_model(full_model):
+            # Per-layer export writes its own config.json with quantization_config, which
+            # the source config would replace; it never writes a processor config.
+            if not args.layerwise_export:
+                save_source_config(args, export_path)
+            save_processor_config(args, export_path)
 
         start_time = time.time()
         is_tensorrt_llm_export = (
@@ -957,22 +943,11 @@ def export_quantized(
                 if mtp_layer_prefixes:
                     full_model._mtp_layer_prefixes = mtp_layer_prefixes
 
-                if args.layerwise_export:
-                    if mtp_state_dict:
-                        raise NotImplementedError(
-                            "layerwise.export_dir does not support models with MTP weights: "
-                            "they are loaded after calibration has already written every "
-                            "shard, so they would be missing from the checkpoint. Export "
-                            "without layerwise.export_dir."
-                        )
-                    # Calibration already wrote every shard, the index and the configs.
-                    print(f"Layerwise export already wrote the checkpoint to {export_path}")
-                else:
-                    export_hf_checkpoint(
-                        full_model,
-                        export_dir=export_path,
-                        extra_state_dict=mtp_state_dict,
-                    )
+                export_hf_checkpoint(
+                    full_model,
+                    export_dir=export_path,
+                    extra_state_dict=mtp_state_dict,
+                )
 
                 if args.qformat == "w4a16_nvfp4":
                     warnings.warn(
@@ -1231,7 +1206,17 @@ def quantize_main(
     is_layerwise = any(cfg.get("enable", False) for cfg in layerwise_cfgs)
 
     # The value is a placeholder, replaced with --export_path below; presence is the switch.
-    args.layerwise_export = any(cfg.get("export_dir") is not None for cfg in layerwise_cfgs)
+    args.layerwise_export = any(
+        cfg.get("export_dir") is not None and cfg.get("enable", False) for cfg in layerwise_cfgs
+    )
+    if not args.layerwise_export and any(
+        cfg.get("export_dir") is not None for cfg in layerwise_cfgs
+    ):
+        warnings.warn(
+            "layerwise.export_dir is set but layerwise.enable is not, so there is no "
+            "per-layer pass to write the shards: the whole-model export runs instead, which "
+            "holds the full state dict in host memory."
+        )
     if args.layerwise_export:
         if isinstance(recipe, ModelOptAutoQuantizeRecipe):
             # Only the mono-quantize path retargets export_dir and runs the refusals;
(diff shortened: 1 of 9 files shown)


sylvesterkaczmarek (commented):
Could we fail explicitly on an extra_state_dict key collision rather than using tail.setdefault(...)?

These tensors are documented as weights the model never held, so a mapped name already existing in the exported checkpoint indicates an invariant violation. Silently keeping the existing tensor could produce a structurally valid checkpoint with the wrong MTP weight, which is much harder to diagnose than failing during export.

I would prefer checking the mapped name against the already-exported namespace and raising a clear error on collision.

modelopt/torch/quantization/model_calib.py
realAsma, line 1: what if we attach model._is_layerwise_export in layerwise mtq.calibrate? 

Then in `export_hf_checkpoint` we detect if any submodule has `_is_layerwise_export` ed?

Fridah-nv: This is what we landed now: 
1.Attach the exporter object rather than a boolean. 2.On detection,  `announce()` publishes it on both ends — the caller announces on the model being calibrated, `bind()` announces on the export root. For VLM that's two different modules.  `finalize()` clears every attachment it recorded, so nothing outlives the export.
Let me know how you think about this!

modelopt/torch/quantization/model_calib.py
```diff
@@ -2206,8 +2212,7 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
         ckpt.full_restore(transformer_layers, model)
 
     if exporter is not None:
-        exporter.finalize()
-        print_rank_0(f"Layerwise export: wrote quantized checkpoint to {export_dir}")
+        print_rank_0(f"Layerwise export: wrote every layer shard to {export_dir}")
```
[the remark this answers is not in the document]
Fridah-nv: Both applied.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2087,12 +2087,25 @@ def layerwise_calibrate(
...
+    from modelopt.torch.export.layerwise_export import LAYERWISE_EXPORTER_ATTR, LayerwiseExporter
+
+    exporter = getattr(model, LAYERWISE_EXPORTER_ATTR, None)
+    if exporter is None and export_dir is not None:
+        # Announced, not just held: whoever owns the export has to be able to finish it.
         exporter = LayerwiseExporter(model, export_dir)
+    if exporter is not None:
+        exporter.announce(model)
+        # Before calibration, so unsupported models fail in seconds not hours.
+        exporter.bind(calibrated_layers=list(transformer_layers))
```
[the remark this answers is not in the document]
Fridah-nv: Confirmed and fixed

modelopt/torch/export/layerwise_export.py
```diff
@@ -354,16 +419,49 @@ def finalize(self) -> dict:
                 continue
             self._collect(tail, name, tensor)
 
+        tail.update(orphans)
```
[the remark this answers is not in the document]
Fridah-nv: Mechanically correct, but not reachable, rejected for simplification

modelopt/torch/export/layerwise_export.py
```diff
@@ -184,7 +222,14 @@ def __init__(
...
                 "architecture is not supported by LayerActivationCollector."
             )
-        # The same call calibration uses, so layer_idx means the same thing on both sides.
+        # A length difference is the one mismatch export_layer's per-layer identity check
+        # cannot catch: every call would pass and _write_index would then miss a shard.
+        if len(layers) != len(calibrated_layers):
+            raise RuntimeError(
+                f"the exporter found {len(layers)} decoder layers but calibration will drive "
+                f"{len(calibrated_layers)}, so layer_idx would not agree. The exporter's "
+                "model must contain exactly the layers being calibrated."
```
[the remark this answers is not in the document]
Fridah-nv: pushed back for simplicity reason

modelopt/torch/export/layerwise_export.py
```diff
@@ -354,10 +410,20 @@ def finalize(self) -> dict:
                 continue
             self._collect(tail, name, tensor)
 
+        for name, tensor in (extra_state_dict or {}).items():
+            key = self._name_mapper(name) if self._name_mapper is not None else name
+            tail[key] = tensor.detach().contiguous().cpu()
```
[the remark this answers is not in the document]
Fridah-nv: Rejected, not reachable through load_mtp_weights, and the whole-model path has no equivalent check.

examples/hf_ptq/hf_ptq.py
realAsma, line 1: Our documentation user flow is:
```
mtq.quantize(model, CFG, forward_loop)
export_hf_checkpoint(model,..)
```

But now for VLM, our flow is:
```
mtq.quantize(model.language_model, CFG, forward_loop)
export_hf_checkpoint(model,..)
```
This is  the root cause of complexity in this PR.

examples/hf_ptq/hf_ptq.py
```diff
@@ -778,6 +780,10 @@ def mono_quantize(
                     else None,
                 )
 
+        if args.layerwise_export:
+            # full_model, not language_model: a VLM's checkpoint describes the whole thing.
+            LayerwiseExporter(full_model, args.export_path).announce(language_model)
```
realAsma:
```suggestion
            LayerwiseExporter(full_model, args.export_path)
```

examples/hf_ptq/hf_ptq.py
```diff
@@ -880,19 +881,17 @@ def export_quantized(
             print(f"Quantized speculative decoding checkpoint exported to: {export_path}")
             return
 
-        # Check if the model is a multimodal/VLM model
-        is_vlm = is_multimodal_model(full_model)
-
-        if is_vlm:
-            # Save original model config and the processor config to the export path for VLMs.
-            print(f"Saving original model config to {export_path}")
+        if is_multimodal_model(full_model):
```
realAsma, line 886: nit:

```suggestion
        if is_multimodal_model(full_model) and not args.layerwise_export:
```

examples/hf_ptq/hf_ptq.py
realAsma, line 1: `hf_ptq` looks minimal now, thanks!

examples/hf_ptq/hf_ptq.py
Fridah-nv: We need the processor save lines at L897 to L904 for layerwise VLM models, so we cannot fold the conditions directly. But let me extract a helper to improve readability

modelopt/torch/export/layerwise_export.py
```diff
@@ -213,6 +260,9 @@ def __init__(
                 "match the original HF hub checkpoint."
             )
 
+        # Last, so a bind() that raised is not mistaken for a completed one on retry.
```
realAsma: nit: these comments are redundant, the behavior is evident from the code.

```suggestion
```

realAsma (approved):
LGTM!

shengliangxu (commented):
Overall the changes look reasonable to me.

shengliangxu (approved):

modelopt/torch/export/layerwise_export.py
Fridah-nv: I agree on this, trimmed a bit more of the comments

examples/hf_ptq/hf_ptq.py
```diff
@@ -855,6 +856,27 @@ def assert_layerwise_export_compatible(args, full_model, mtp_layer_prefixes) ->
             )
 
 
+def save_source_config(args, export_path) -> None:
```
sugunav14: Do you think these helpers can live in example_utils.py? My reason for this is to avoid blowing up hf_ptq.py.

Fridah-nv: appiled the change, thanks for the suggestion!

PR #2314, opened 2026-09-02
[6508436] Fix BF16 FP8 ONNX export

ajrasane:
### What does this PR do?

Type of change: Bug fix

Fix FP8 ONNX export for BF16 models during real-weight compression without changing the public API or the `weights_dtype="fp32"` default.

The FP8 exporter preserves BF16 initializer bits when bridging GraphSurgeon NumPy arrays to Torch, widens BF16 values exactly to FP32 for normalization, and leaves existing FP16/FP32 handling unchanged. Conv scales and dequantized outputs retain the source dtype, and scales round upward when needed so serialized values cannot cause FP8 overflow.

`weights_dtype="bf16"` is accepted as a no-op only for FP8-only models whose floating parameters are all BF16. Registered buffers do not affect this weight-focused decision and may preserve higher-precision regions in the exported graph. Unsupported BF16 FP8-to-FP16 and FP32 or mixed-parameter-to-BF16 conversions are rejected with `ValueError` before temporary export paths are created. A narrow GraphSurgeon fix preserves integer BF16 value-info dtypes.

### Usage

```python
onnx_bytes, metadata = get_onnx_bytes_and_metadata(
    quantized_fp8_model,
    (sample_input,),
    weights_dtype="bf16",
    onnx_opset=23,
)
```

### Testing

- Seven focused CPU regressions passed: BF16 QDQ compression and integer dtype handling, BF16-to-BF16 and FP32-to-FP16 Conv/Linear export, and four unsupported-conversion cases.
- QDQ utilities: 31 passed; pytest 2.25s, wall 29.88s.
- FP8 MHA exporter: 6 passed; pytest 2.05s, wall 32.71s.
- Torch deploy utilities: 51 passed; pytest 8.94s, wall 25.74s.
- Torch ONNX CPU export: 36 passed; pytest 4.85s, wall 32.38s.
- Changed-file pre-commit hooks: all passed; wall 8.07s.
- Exact-head FP8 BF16 GPU workflow at `f21d62a`: exit code 0; ONNX checker passed; 6 FP8 initializers, 3 native `DequantizeLinear` nodes, and 12 BF16 initializers.
- Refreshed GitHub CI at `f21d62a`: 50 passed and 1 skipped. Unit, GPU, and regression required aggregates and Codecov passed. Two ONNX example leaves failed because the runner could not load a cuDNN sublibrary; their dependent example aggregate consequently failed.

### Additional Information

- TODO: Deliver authoritative `native`/FP32/FP16/BF16 ONNX export across all quantized formats in follow-up pull requests.


modelopt/onnx/export/fp8_exporter.py
```diff
@@ -194,7 +202,7 @@ def _quantize_conv_weights_to_fp8(graph: gs.Graph) -> int:
             if any(out.op == "DequantizeLinear" for out in weight_input.outputs):
                 continue
 
-            torch_weights = torch.from_numpy(weight_input.values.copy())
+            torch_weights = _torch_from_numpy(weight_input.values.copy())
```
cjluo-nv: > _Bot comment._

This now succeeds for BF16 conv weights, but a few lines below the scale constant is still hard-coded to FP16:

```python
scale_constant = gs.Constant(
    node.name + "/weight_quantizer/scale",
    np.array(scale_val, dtype=np.float16),
)
```

with the comment "DQ output type matches scale dtype, must match activation type". In the BF16 path this PR enables, the surrounding activations are BF16, so the inserted `DequantizeLinear` will produce FP16 while the `Conv` input `X` is BF16 — that violates the ONNX `Conv` type constraint (`X` and `W` share `T`) and will fail strongly-typed parsing. Before this PR the BF16 conv path crashed in `torch.from_numpy`, so it was never reachable; now it silently produces a mistyped graph.

Suggest deriving the scale dtype from `weight_input.values.dtype` (BF16 weights → BF16 scale) and adding a Conv-bearing BF16 regression (all the new tests use `nn.Linear` only, so this path is untested).

modelopt/onnx/export/fp8_exporter.py
```diff
@@ -78,8 +86,8 @@ def compress_weights(onnx_model: onnx.ModelProto) -> onnx.ModelProto:
 
                 weights = node.inputs[0]
                 scale = node.inputs[1]
-                torch_weights = torch.from_numpy(weights.values)
-                torch_scale = torch.from_numpy(scale.values)
+                torch_weights = _torch_from_numpy(weights.values)
```
cjluo-nv: > _Bot comment._

With BF16 preserved, `(torch_weights / torch_scale)` in `compress_weights` (and `torch_weights / scale_val` in the conv path) is now computed in BF16 — 8 mantissa bits — before rounding to FP8 E4M3. That is strictly less accurate than the previous FP32/FP16 arithmetic and is observable: the new unit test's `assert fp8_weight.raw_data == b"\x3a"` only holds because the BF16 division lands exactly on a round-half-to-even tie (FP32 arithmetic would give `0x3b`). Consider `torch_weights.float() / torch_scale.float()` before `.to(torch.float8_e4m3fn)`, and if the BF16 arithmetic is intentional, please say so in a comment so the magic byte assertion is explainable to future readers.

modelopt/onnx/quantization/gs_patching.py
```diff
@@ -101,9 +101,14 @@ def _export_value_info_proto(tensor: gs.Variable, do_type_check: bool) -> onnx.V
...
     if tensor.dtype is not None:
-        dtype = getattr(
-            tensor, "explicit_dtype", onnx.helper.np_dtype_to_tensor_dtype(np.dtype(tensor.dtype))
-        )
+        dtype = getattr(tensor, "explicit_dtype", None)
+        if dtype is None:
+            dtype = tensor.dtype
+        if isinstance(dtype, (int, np.integer)):
+            dtype = int(dtype)
+            onnx.TensorProto.DataType.Name(dtype)
```
cjluo-nv: > _Bot comment._

`onnx.TensorProto.DataType.Name(dtype)` is called purely for its side effect (raising `ValueError` on an unknown enum) and the result is discarded — this reads like dead code to the next reader and is easy to "clean up" by accident. Please either add a comment explaining that it validates the enum, or make the intent explicit, e.g.:

```python
if dtype not in onnx.TensorProto.DataType.values():
    raise ValueError(f"Unknown ONNX tensor dtype for {tensor.name}: {dtype}")
```

modelopt/torch/_deploy/utils/torch_onnx.py
```diff
@@ -527,6 +527,9 @@ def get_onnx_bytes_and_metadata(
     if isinstance(model, (DataParallel, DistributedDataParallel)):
         model = model.module
 
+    first_parameter = next(model.parameters(), None)
+    source_weights_dtype = first_parameter.dtype if first_parameter is not None else torch.float32
```
cjluo-nv: > _Bot comment._

`source_weights_dtype` is inferred from the first parameter only. For a mixed-dtype module (e.g. an FP32 embedding/head registered before the BF16 body) this will misclassify a genuinely BF16 model and reject a valid `weights_dtype="bf16"` export with the new assertion. Consider a more robust probe (e.g. `any(p.dtype == torch.bfloat16 for p in model.parameters())`, or checking the dtype of the exported graph's float initializers) and add a short comment documenting the heuristic either way.

cjluo-nv (commented):

Small, focused bug fix (6 files, +184/-15) that unblocks FP8 ONNX export for BF16 source models. No new subsystem/abstraction is introduced (the design-review gate fired only on directory span): the only new construct is a 4-line `_torch_from_numpy` helper, and I checked the existing alternatives — `modelopt/onnx/utils.py` has `bfloat16_to_float32` / `read_f16_tensor_as_fp32` for the TensorProto→fp32 direction but nothing for the GraphSurgeon-numpy→torch bit-preserving direction, so the helper isn't a duplicate. The `gs_patching` fix (avoiding the eagerly-evaluated `getattr` default that crashed on ONNX enum dtypes) is a genuine improvement, and the added tests (unit exporter test, positive/negative export tests) are meaningful. No licensing surface touched, and no injection attempts in the PR text.

Blocking concern: `_quantize_conv_weights_to_fp8` now accepts BF16 conv weights but still emits a hard-coded FP16 scale constant, which produces a `DequantizeLinear` whose output type (FP16) doesn't match the surrounding BF16 activations in the newly-enabled BF16 path — and there is no Conv-based BF16 test to catch it. A few smaller items on precision of the BF16 division, the bare `DataType.Name(dtype)` validation call, and the first-parameter dtype heuristic are noted inline.

modelopt/onnx/export/fp8_exporter.py
ajrasane: Fixed in `a2afc68`. Conv scales now use the same dtype as the source weight and activation, and the inserted DQ output carries that dtype and shape. The serialized scale is rounded upward when needed to avoid FP8 overflow. I added a BF16 Conv end-to-end regression with full ONNX type checking and a finite-FP8-byte assertion.

modelopt/torch/_deploy/utils/torch_onnx.py
ajrasane: Fixed in `a2afc68`. The first-parameter heuristic is replaced with the complete set of floating parameter dtypes. The BF16 bypass now requires that set to be exactly `{torch.bfloat16}`, so registration order cannot affect the result and a true mixed-precision model remains rejected because conversion would not be a no-op. A mixed-source regression registers BF16 first and FP32 second.

modelopt/onnx/quantization/gs_patching.py
ajrasane: Fixed in `a2afc68`. The discarded `DataType.Name` call is replaced with an explicit enum-membership check and a descriptive `ValueError` for unknown values.

modelopt/onnx/export/fp8_exporter.py
ajrasane: Agreed. BF16 weights and scales are now promoted to FP32 for normalization before FP8 rounding, matching the eager FP8 path while leaving the established FP16 path unchanged. The discriminating byte expectation is updated from `0x3a` to `0x3b`.

modelopt/torch/_deploy/utils/torch_onnx.py
```diff
@@ -634,14 +638,21 @@ def get_onnx_bytes_and_metadata(
...
-            or is_int8_quantized(model)
-        ):
-            assert weights_dtype == "fp16", "BF16 + MXFP8/INT4 mixed precision is not supported yet"
+    uses_fp8 = is_fp8_quantized(model)
+    uses_other_unsupported_quantizer = (
+        is_int4_quantized(model) or is_mxfp8_quantized(model) or is_int8_quantized(model)
+    )
+    is_bf16_fp8_noop = (
+        weights_dtype == "bf16"
+        and source_parameter_dtypes == {torch.bfloat16}
```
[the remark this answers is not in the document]
ajrasane: Fixed in `9730260`. The no-op eligibility check now includes floating registered buffers as well as parameters. I added a regression with BF16 parameters and a referenced FP32 buffer that verifies the existing unsupported-conversion assertion.

CHANGELOG.rst
```diff
@@ -131,6 +131,7 @@ Changelog
 
 **Bug Fixes**
 
+- Fix FP8 ONNX export of BF16 models during real-weight compression.
```
gcunhase: Should this be moved to the end of the list or are newer additions added at the top?

modelopt/onnx/export/fp8_exporter.py
```diff
@@ -33,6 +34,13 @@
 _FP8_E4M3_SOFTMAX_SCALE = 1.0 / _FP8_E4M3_MAX
 
 
+def _torch_from_numpy(array: np.ndarray) -> torch.Tensor:
+    """Convert a NumPy array to a PyTorch tensor while preserving BF16 values."""
+    if array.dtype == ml_dtypes.bfloat16:
+        return torch.from_numpy(array.view(np.int16)).view(torch.bfloat16)
+    return torch.from_numpy(array)
```
gcunhase, line 41: If I understand correctly, the issue is that torch.from_numpy() converts BF16 arrays to FP16 instead of keeping them in BF16?

Why is the array first converted to INT16 and then BF16 here?

ajrasane: Not quite—`torch.from_numpy()` does not convert this array to FP16; it rejects `ml_dtypes.bfloat16`. The two `.view()` calls provide a zero-copy bridge: NumPy exposes the unchanged 16-bit BF16 storage as a supported `int16` carrier, then PyTorch reinterprets those same bits as `torch.bfloat16`. No numeric INT16 or FP16 conversion occurs.

gcunhase: Thanks for elaborating on this. Are there any accuracy concerns with doing `BF16->INT16->BF16`?

CHANGELOG.rst
ajrasane: Good catch—the entry was under 0.46. This changelog is grouped topically rather than strictly appended or prepended, so I moved it to 0.47 Bug Fixes beside the other ONNX export fixes in `b53ef03`.

cjluo-nv (commented):

**Design gate (directory-span trigger):** re-confirmed this is a contained bug fix, not a new subsystem. Problem: `torch.from_numpy` rejects `ml_dtypes.bfloat16` arrays produced by GraphSurgeon, so FP8 real-weight compression crashed for BF16 source models. Alternatives checked: `modelopt/onnx/utils.py` already has `bfloat16_to_float32` / `read_f16_tensor_as_fp32`, but those go TensorProto→FP32, not numpy→torch bit-preserving, so the 4-line `_torch_from_numpy` helper is not a duplicate; `onnx.numpy_helper`/`convert_to_f16` don't cover this direction either. No second system introduced. No licensing surface. No injection attempts in the PR text.

**Previous comments — status**
- *(critical)* Hard-coded FP16 Conv scale under the newly-enabled BF16 path — **addressed**. :speech_balloon: Author (`a2afc68`): scale now derives from `weight_input.values.dtype`, the inserted DQ output carries that dtype/shape, and a BF16 Conv end-to-end regression with `check_model(full_check=True)` was added.
- *(critical)* BF16-precision division before FP8 rounding — **addressed**: BF16 weights/scale are promoted to FP32 in both `compress_weights` and the Conv path; the magic-byte assertion was updated `0x3a → 0x3b` accordingly.
- *(critical)* First-parameter dtype heuristic — **addressed**: replaced with the full set of floating parameter *and* buffer dtypes, gated on `== {torch.bfloat16}`, with mixed-parameter and mixed-buffer regressions.
- *(minor)* Discarded `DataType.Name(dtype)` validation call — **addressed** with an explicit `DataType.values()` membership check and descriptive `ValueError`, plus a unit test for the ONNX-enum dtype path.
- *(minor)* CHANGELOG placement — **addressed** (moved to 0.47 Bug Fixes).

**Residual items for the owner**
- Conv + `weights_dtype="fp16"` is now untested and behaviourally changed. The deleted comment ("DQ output type matches scale dtype, must match activation type") existed because the FP32→FP16 conversion blocks `DequantizeLinear`; the Conv scale is now emitted FP32 and only becomes FP16 via `fold_dq_fp32_to_fp16_casts` downstream. That chain looks like it still converges, but the new parametrization covers `(bf16, "bf16", conv=True)` and `(fp32, "fp16", conv=False)` only — adding `(fp32, "fp16", conv=True)` would pin the previously-working path.
- BF16 source + `weights_dtype="fp16"`: this now falls through to `convert_float_to_float16`, which only rewrites `FLOAT` tensors. Before this PR the path crashed in `torch.from_numpy`, so it was unreachable; it may now silently emit a mixed BF16/FP16 graph rather than raising. Worth either asserting/raising explicitly or covering with a test.
- `source_floating_dtypes == {torch.bfloat16}` is strict by design (good for correctness) but rejects common real BF16 models that keep FP32 buffers (e.g. HF rotary `inv_freq`), which will hit "Converting a quantized ONNX graph to BF16 is not supported yet". Not a regression versus `main`, but the message doesn't hint at the actual cause; consider naming the offending dtypes in the assertion.
- gcunhase's follow-up ("any accuracy concerns with BF16→INT16→BF16?") appears unanswered in the thread. For the record the two `.view()` calls are bit-preserving reinterpretations (no numeric cast), so there is no accuracy loss — but a human reviewer's question is still open.
- Nit: the `np.nextafter(...)` upward-rounding block in `_quantize_conv_weights_to_fp8` has no comment explaining *why* (keeping `max|w|/scale ≤ 448` so the FP8 cast can't produce NaN); one line would help the next reader, especially since it's exercised only indirectly by the `1e-38` Conv fixture.

modelopt/onnx/export/fp8_exporter.py
ajrasane: No. Both operations are same-width `.view()` reinterpretations, not numeric casts. NumPy exposes the same 16 bits through `int16`, and PyTorch reinterprets those unchanged bits as `bfloat16`; there is no rounding or accuracy loss. The `int16` view is only a bridge because `torch.from_numpy()` does not accept `ml_dtypes.bfloat16` directly.

cjluo-nv (commented):

Re-review #3 of the BF16 FP8 ONNX export fix (6 files, +298/-19). Design gate (directory-span trigger) remains settled from earlier rounds: this is a contained bug fix, the only new construct is the 4-line `_torch_from_numpy` bridge (no in-repo equivalent — `modelopt/onnx/utils.py` only covers TensorProto→FP32), no second system, no licensing surface, no injection attempts in the PR text.

**Previously-flagged items — status**
- *(critical)* Hard-coded FP16 Conv scale under the BF16 path — **addressed** (scale derives from `weight_input.values.dtype`, DQ output carries dtype/shape, BF16 Conv e2e regression with `check_model(full_check=True)`).
- *(critical)* BF16 arithmetic before FP8 rounding — **addressed** (FP32 promotion in both `compress_weights` and the Conv path; magic byte updated to `0x3b`).
- *(critical)* First-parameter dtype heuristic / FP32 buffers — **addressed** (`source_floating_dtypes` over parameters *and* buffers, with mixed-param, mixed-buffer and mixed-format regressions).
- *(minor)* Discarded `DataType.Name()` validation, CHANGELOG placement, `np.nextafter` why-comment, `(fp32, "fp16", conv=True)` coverage, silent BF16→FP16 fallthrough — all **addressed** (explicit enum membership check, 0.47 Bug Fixes entry, comment added, new parametrization case, explicit rejection + test).
- *(minor)* gcunhase's `BF16→INT16→BF16` accuracy question — **answered** in-thread (both `.view()`s are bit-preserving reinterpretations, no numeric cast).

**Why a human should still glance before merge**
- :speech_balloon: Author (`9730260`/`4c6a904`): the no-op gate now inspects floating *buffers* too and the error names the detected dtypes — still worth a look because the gate is `source_floating_dtypes == {torch.bfloat16}`, so any FP32 buffer (HF rotary `inv_freq`, and potentially ModelOpt's own `_amax`/`_pre_quant_scale` if a calibrator stores FP32) makes the newly-enabled BF16 path unreachable and surfaces as "Converting a quantized ONNX graph to BF16 is not supported yet". The unit tests only use tiny `nn.Linear`/`nn.Conv2d` models, so they can't confirm the path fires for realistic BF16 checkpoints. Not a regression versus `main`, but it determines whether this fix actually helps users.
- Unaddressed minor from CodeRabbit: the new `raise AssertionError(...)` for BF16-FP8 → FP16 fires after `tempfile.mkdtemp(...)` and after `torch.onnx.export` has written the model, so the rejected export leaks the temp directory (the `shutil.rmtree(onnx_path)` at the end is skipped). Moving the check before the export — the inputs it needs (`weights_dtype`, `source_floating_dtypes`, `is_fp8_quantized(model)`) are all available up front — would also avoid doing a full export just to reject it. Related nit: raising `AssertionError` explicitly (rather than `ValueError`) is unusual, even if it keeps the tests' `pytest.raises(AssertionError)` consistent with the neighbouring `assert`.

modelopt/torch/_deploy/utils/torch_onnx.py
```diff
@@ -634,14 +642,28 @@ def get_onnx_bytes_and_metadata(
...
-            assert weights_dtype == "fp16", "BF16 + MXFP8/INT4 mixed precision is not supported yet"
+    uses_fp8 = is_fp8_quantized(model)
+    uses_other_unsupported_quantizer = (
+        is_int4_quantized(model) or is_mxfp8_quantized(model) or is_int8_quantized(model)
+    )
+    if weights_dtype == "fp16" and uses_fp8 and torch.bfloat16 in source_floating_dtypes:
+        raise AssertionError(
+            "Converting a BF16 FP8 ONNX graph to FP16 is not supported yet "
+            f"(source floating dtypes: {source_floating_dtype_names})"
+        )
```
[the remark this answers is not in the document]
ajrasane: Fixed in `eb07421`. Compatibility validation now runs before the temporary ONNX path is created, and the focused rejection test redirects the temporary root and verifies that it remains empty.

cjluo-nv (commented):

Re-review #4 (6 files, +253/-33): every previously-flagged blocker is fixed in the current code — the only open point is the deliberate parameters-only dtype gate, which the author himself calls unproven.

Needs action:
- [ ] Confirm the BF16 no-op gate in `modelopt/torch/_deploy/utils/torch_onnx.py` should ignore floating buffers: with BF16 params + a forward-used FP32 buffer, `is_bf16_fp8_noop` is now True and the "bf16" graph keeps FP32 initializers. Add a test for that case or document the semantics.
- [ ] Decide whether the BF16 rejection should keep `raise AssertionError` while the new FP16 rejection uses `ValueError` (`torch_onnx.py` ~L590-605); if the assert is kept for the old contract, say so in a comment.

No action needed:
- :heavy_check_mark: Resolved since the last review: temp-dir leak (validation now runs before `tempfile.mkdtemp`), `(fp32, "fp16", conv=True)` coverage, explicit BF16→FP16 rejection, dtype names in error messages, and the `np.nextafter` why-comment.
- Design gate stays settled: contained bug fix, only new construct is the 4-line `_torch_from_numpy_for_fp8` bridge; no in-repo equivalent, no licensing surface.

cjluo-nv (commented):

Approve — both open items from the last round are fixed in the current code and nothing new surfaced.

No action needed:
- :heavy_check_mark: Resolved since the last review: the parameters-only BF16 no-op gate is now documented in the `weights_dtype` docstring and pinned by the `bf16-weight-focused-buffer` test (BF16 params + forward-used FP32 buffer, asserting the FP32 initializer survives and feeds the `Add`), and both precision rejections now raise `ValueError` with the `Raises:` docstring updated.
- Exception type for BF16+FP8 rejection changed from `AssertionError` to `ValueError`; the docstring covers it and the tests match.
- Design gate stays settled: contained bug fix, only new construct is the 4-line `_torch_from_numpy_for_fp8` bridge; no licensing surface, no injection attempts in the PR text.

Complex PR: spans 5 directories (≥ 5); 1 existing test file modified or removed. Looping in a human for approval.

cjluo-nv (approved):

PR #2320, opened 2026-09-02
Reject unsupported partial-block INT4/W4A8 AWQ export

realAsma:
### What does this PR do?

Type of change: Bug fix

Reject INT4 and W4A8 AWQ export when a weight's input dimension is not divisible by
the configured block size.

Nemotron-3-Nano-4B has weights with input dimension 3136, which is not divisible by
the configured block size 128. Partial INT4/W4A8 blocks are not supported. The
export path previously inferred an incorrect block size and reached an out-of-bounds
CUDA scale index. It now raises a clear `NotImplementedError` before GPU indexing.

### Usage

No API change. Unsupported partial-block INT4/W4A8 AWQ exports now fail early with a
clear error instead of a CUDA device-side assertion.

### Testing

- `pre-commit run --files modelopt/torch/export/quant_utils.py tests/gpu/torch/export/test_export.py` — passed.
- Focused GPU test for supported and partial-block INT4/W4A8 AWQ packing — 2 cases passed.
- The reported Nemotron-3-Nano-4B failure was reproduced and localized with
  `CUDA_LAUNCH_BLOCKING=1` before applying the guard.

- Did you update Changelog?: N/A — this fixes behavior in the current unreleased 0.47 development line.
- Did you get Claude approval on this PR?: N/A — the focused candidate passed independent senior code review and test review.

### Additional Information

The check is format-generic and does not special-case Nemotron or any architecture.

## Summary by CodeRabbit

* **Bug Fixes**
  * INT4/AWQ and W4A8 AWQ checkpoint exports now validate block sizes before processing weights.
  * Exports reject block sizes that are non-positive, non-integer, or do not evenly divide the input dimension.
  * Improved validation for compressed INT4/AWQ weights, including partially filled quantization blocks.

* **Tests**
  * Added GPU coverage for valid and invalid INT4/AWQ and W4A8 packing scenarios.
  * Expanded quantized-model export coverage using larger test dimensions.

### DIFF
--- a/modelopt/torch/export/quant_utils.py
+++ b/modelopt/torch/export/quant_utils.py
@@ -792,11 +792,7 @@ def process_layer_quant_config(layer_config_dict):
     return per_layer_config
 
 
-def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
-    """Packs the INT4 weights into uint8 tensor."""
-    out_dim = weight.shape[-2]
-    assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
-    in_dim = weight.shape[-1]
+def _validate_int4_block_size(in_dim, block_size):
     if not isinstance(block_size, int) or block_size <= 0:
         raise ValueError(f"Block size must be a positive integer, got {block_size}.")
     if in_dim % block_size != 0:
@@ -804,12 +800,14 @@ def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
             f"Cannot pack weight with input dimension {in_dim} and block size {block_size}: "
             "partial blocks are not supported."
         )
-    expected_scale_count = in_dim // block_size
-    if weights_scaling_factor.shape[-1] != expected_scale_count:
-        raise ValueError(
-            f"Expected {expected_scale_count} weight scaling factors for input dimension {in_dim} "
-            f"and block size {block_size}, got {weights_scaling_factor.shape[-1]}."
-        )
+
+
+def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
+    """Packs the INT4 weights into uint8 tensor."""
+    out_dim = weight.shape[-2]
+    assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
+    in_dim = weight.shape[-1]
+    _validate_int4_block_size(in_dim, block_size)
 
     # Scale, round, and clamp to the signed 4-bit range [-8..7].
     int8_tensor = (
@@ -864,6 +862,8 @@ def to_quantized_weight(
 
     # For compressed weights, we directly return the data from wrapper
     if isinstance(weight, QTensorWrapper):
+        if quantization in [QUANTIZATION_INT4_AWQ, QUANTIZATION_W4A8_AWQ]:
+            _validate_int4_block_size(weight.metadata["shape"][-1], block_size)
         return weight.data
 
     if quantization == QUANTIZATION_FP8:
--- a/modelopt/torch/export/quant_utils.py
+++ b/modelopt/torch/export/quant_utils.py
@@ -799,7 +799,12 @@ def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
     in_dim = weight.shape[-1]
     if block_size is None or block_size <= 0:
         raise ValueError(f"Block size must be a positive integer, got {block_size}.")
-    expected_scale_count = (in_dim + block_size - 1) // block_size
+    if in_dim % block_size != 0:
+        raise NotImplementedError(
+            f"Cannot pack weight with input dimension {in_dim} and block size {block_size}: "
+            "partial blocks are not supported."
+        )
+    expected_scale_count = in_dim // block_size
     if weights_scaling_factor.shape[-1] != expected_scale_count:
         raise ValueError(
             f"Expected {expected_scale_count} weight scaling factors for input dimension {in_dim} "
--- a/modelopt/torch/export/quant_utils.py
+++ b/modelopt/torch/export/quant_utils.py
@@ -797,7 +797,7 @@ def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
     out_dim = weight.shape[-2]
     assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
     in_dim = weight.shape[-1]
-    if block_size is None or block_size <= 0:
+    if not isinstance(block_size, int) or block_size <= 0:
         raise ValueError(f"Block size must be a positive integer, got {block_size}.")
     if in_dim % block_size != 0:
         raise NotImplementedError(
--- a/modelopt/torch/export/quant_utils.py
+++ b/modelopt/torch/export/quant_utils.py
@@ -792,12 +792,19 @@ def process_layer_quant_config(layer_config_dict):
     return per_layer_config
 
 
-def pack_int4_in_uint8(weight, weights_scaling_factor):
+def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
     """Packs the INT4 weights into uint8 tensor."""
     out_dim = weight.shape[-2]
     assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
     in_dim = weight.shape[-1]
-    block_size = weight.shape[-1] // weights_scaling_factor.shape[-1]
+    if block_size is None or block_size <= 0:
+        raise ValueError(f"Block size must be a positive integer, got {block_size}.")
+    expected_scale_count = (in_dim + block_size - 1) // block_size
+    if weights_scaling_factor.shape[-1] != expected_scale_count:
+        raise ValueError(
+            f"Expected {expected_scale_count} weight scaling factors for input dimension {in_dim} "
+            f"and block size {block_size}, got {weights_scaling_factor.shape[-1]}."
+        )
 
     # Scale, round, and clamp to the signed 4-bit range [-8..7].
     int8_tensor = (
@@ -912,7 +919,7 @@ def to_quantized_weight(
         return (weight / weights_scaling_factor[:, None]).to(torch.float8_e4m3fn)
 
     if quantization in [QUANTIZATION_INT4_AWQ, QUANTIZATION_W4A8_AWQ]:
-        return pack_int4_in_uint8(weight, weights_scaling_factor)
+        return pack_int4_in_uint8(weight, weights_scaling_factor, block_size)
 
     if quantization in [
         QUANTIZATION_NVFP4,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -131,6 +131,7 @@ Changelog
 
 **Bug Fixes**
 
+- Reject INT4 and W4A8 AWQ checkpoint export when a weight's input dimension is not divisible by the configured block size. Choose an ``awq_block_size`` that evenly divides every quantized weight's input dimension.
 - Fix NemotronH dense MLP quantization with the ``nvfp4_mlp_only`` and ``nvfp4_omlp_only`` recipe families. NemotronH registers these projections as ``mixer.up_proj`` / ``mixer.down_proj``, which the previous ``*mlp*`` selector missed, producing checkpoints with a null ``quant_algo``.
 - Fix ``ShapeInferenceError`` during ONNX INT8 + FP16 quantization (``--high_precision_dtype fp16``) of weakly-typed models (e.g. TensorFlow exports) that carry stale rank-0 ``graph.output`` shapes or ops such as ``TopK`` that ONNX's static shape inference cannot resolve. Stale output shapes are now reconciled via symbolic shape inference, and AutoCast falls back to schema-based type inference so unresolved ops no longer leave tensors untyped.
 - Fix fused MoE expert auto-detection (``register_fused_experts_on_the_fly``) skipping modules without an ``act_fn`` attribute. Modules applying a custom gated activation between the two ``F.linear`` calls (e.g. ``MiniMaxM3VLExperts``) were silently skipped, leaving routed experts unquantized and failing HF export. Enables NVFP4/FP8 quantization and export for MiniMax-M2 / MiniMax-M3.
--- a/tests/gpu/torch/export/test_export.py
+++ b/tests/gpu/torch/export/test_export.py
@@ -56,6 +56,7 @@ from modelopt.torch.export.quant_utils import (
     get_weight_block_size,
     postprocess_state_dict,
     process_layer_quant_config,
+    to_quantized_weight,
 )
 from modelopt.torch.export.unified_export_hf import export_hf_checkpoint
 from modelopt.torch.quantization.config import (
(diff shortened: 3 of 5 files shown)


modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,19 @@ def process_layer_quant_config(layer_config_dict):
...
+def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
     """Packs the INT4 weights into uint8 tensor."""
     out_dim = weight.shape[-2]
     assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
     in_dim = weight.shape[-1]
-    block_size = weight.shape[-1] // weights_scaling_factor.shape[-1]
+    if block_size is None or block_size <= 0:
+        raise ValueError(f"Block size must be a positive integer, got {block_size}.")
+    expected_scale_count = (in_dim + block_size - 1) // block_size
+    if weights_scaling_factor.shape[-1] != expected_scale_count:
```
realAsma: BB: Is this correct - "Existing INT4/W4A8 AWQ export commands now handle partial final blocks."?

NVIDIA-Nemotron-3-Nano-4B-BF16 will simply run into this error right?

realAsma: > _:robot: Bot comment._

🐝 I’m checking the exact failure mode and will clarify the wording here.

realAsma: > _:robot: Bot comment._

Yes—“now handle” describes behavior after this PR. Nemotron-3-Nano-4B is the reproducer: before the fix, its 3136-wide weight with block size 128 has 25 scales, but the old floor-derived size of 125 causes the final-block indexing failure. With this change, export uses the configured size of 128 and completes the partial block.

realAsma: BB: We should simply raise an error if weight.shape[-1] is not a multiple of block size and throw not supported error.

Can you fix this PR?

realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

Fixed in `6905464`. INT4 and W4A8 AWQ export now raises `NotImplementedError` when the input dimension is not divisible by the configured block size; the focused two-format GPU test passes.

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,24 @@ def process_layer_quant_config(layer_config_dict):
...
     assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
     in_dim = weight.shape[-1]
-    block_size = weight.shape[-1] // weights_scaling_factor.shape[-1]
+    if block_size is None or block_size <= 0:
+        raise ValueError(f"Block size must be a positive integer, got {block_size}.")
+    if in_dim % block_size != 0:
+        raise NotImplementedError(
+            f"Cannot pack weight with input dimension {in_dim} and block size {block_size}: "
+            "partial blocks are not supported."
+        )
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

The rejection is intentional per the PR owner's direction: partial INT4/W4A8 AWQ blocks are unsupported by the export contract. This change converts the prior out-of-bounds CUDA failure into an explicit error, and the PR body now states that scope. Earlier config-time validation can be considered separately.

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,24 @@ def process_layer_quant_config(layer_config_dict):
...
+    expected_scale_count = in_dim // block_size
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

The partial-block rejection is retained intentionally. For every accepted input, `in_dim` is divisible by `block_size`, so floor division is the exact expected scale count; ceil handling would only apply to the rejected case.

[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

No ceiling-division expression or import is needed because partial blocks remain intentionally unsupported by this export path.

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,24 @@ def process_layer_quant_config(layer_config_dict):
...
 
-def pack_int4_in_uint8(weight, weights_scaling_factor):
+def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
     """Packs the INT4 weights into uint8 tensor."""
     out_dim = weight.shape[-2]
     assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
     in_dim = weight.shape[-1]
-    block_size = weight.shape[-1] // weights_scaling_factor.shape[-1]
+    if block_size is None or block_size <= 0:
+        raise ValueError(f"Block size must be a positive integer, got {block_size}.")
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Fixed in `3f2c679c`: validate that `block_size` is an integer before arithmetic, and cover `None`, zero, negative, and float inputs for both INT4 and W4A8 AWQ paths. The focused GPU test passes all 10 cases.

cjluo-nv (commented):

The fix is focused and correct: INT4/W4A8 AWQ packing now validates the configured block size and scale count before constructing GPU scale indices, so unsupported partial blocks fail clearly instead of reaching an out-of-bounds CUDA access. Callers already propagate the quantizer's configured block size, supported and failure paths are covered for both formats, and affected integration fixtures were updated to use valid AWQ dimensions.

Complex PR: 2 existing test files modified or removed. Looping in a human for approval.

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,24 @@ def process_layer_quant_config(layer_config_dict):
...
 
 
-def pack_int4_in_uint8(weight, weights_scaling_factor):
+def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
     """Packs the INT4 weights into uint8 tensor."""
     out_dim = weight.shape[-2]
     assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
     in_dim = weight.shape[-1]
-    block_size = weight.shape[-1] // weights_scaling_factor.shape[-1]
+    if not isinstance(block_size, int) or block_size <= 0:
```
cjluo-nv, line 796: in what condition can block_size < 0 or not an int?

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,24 @@ def process_layer_quant_config(layer_config_dict):
...
+        raise ValueError(f"Block size must be a positive integer, got {block_size}.")
+    if in_dim % block_size != 0:
```
cjluo-nv, line 798: should we throw this error at the very beginning during quantization instead of at the export time?

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,24 @@ def process_layer_quant_config(layer_config_dict):
...
+        raise NotImplementedError(
+            f"Cannot pack weight with input dimension {in_dim} and block size {block_size}: "
+            "partial blocks are not supported."
+        )
+    expected_scale_count = in_dim // block_size
+    if weights_scaling_factor.shape[-1] != expected_scale_count:
```
cjluo-nv: in what condition will this hit?

modelopt/torch/export/quant_utils.py
Edwardf0t1, line 796: **[P1] Apply this validation to compressed AWQ weights too.** `to_quantized_weight` returns `weight.data` for every `QTensorWrapper` before this helper is called. `mtq.compress` can produce an `INT4QTensor` wrapper, and `INT4QTensor.quantize` pads non-divisible inputs, so a compressed INT4/W4A8 AWQ weight with a partial block bypasses this guard and export can continue with an unsupported packed checkpoint. Please validate the logical input dimension from `weight.metadata["shape"][-1]` before the wrapper fast path (or explicitly reject compressed AWQ export), and cover that route in a test.

Edwardf0t1 (changes requested):
One blocking path remains: compressed AWQ weights bypass the new partial-block validation.

modelopt/torch/export/quant_utils.py
realAsma: Quantization helps us to do accuracy evaluation for methods and formats that might not be supported in deployment. So  `if in_dim % block_size != 0:` should not be considered as an error during quantization.

However `if in_dim % block_size != 0:` is not supported for deployment which is the point of export. So we should throw error during export.

modelopt/torch/export/quant_utils.py
realAsma: BB: Is the fix valid for INT4 AWQ as well? If not can you raise similar error during AWQ export as well?

modelopt/torch/export/quant_utils.py
realAsma: BB: Is this check needed? If this check is not needed, can you remove this?

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

🐝 I’m checking the INT4 AWQ path and will make sure it fails clearly during export if the current guard does not cover it.

realAsma: BB: PTAL this comment

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

🐝 I’m checking whether this guard catches a real export-state inconsistency; I’ll remove it if it is redundant.

realAsma: > _:robot: Bot comment._

🐝

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

The current helper covers uncompressed INT4 AWQ and W4A8 AWQ. The remaining issue is the compressed `QTensorWrapper` fast path, which bypasses that helper for both formats; the pending fix should validate the wrapper’s logical shape before returning its packed data and test both formats.

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

Agreed. Supported export paths already produce the scale count from the weight shape and block size, so this guard only diagnoses malformed internal state and is not needed for the requested partial-block fix. It and its dedicated mismatch test should be removed in the same fix batch.

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

The positive-integer check is defensive but useful at this callable boundary: `block_size` is optional in `to_quantized_weight`, and without the guard `None`, zero, or non-integral direct/malformed inputs fail later with low-level modulo, division, or indexing errors. I recommend keeping this clear validation.

realAsma: @cjluo-nv  I am resolving this. please see the reasoning above. this might at least help us get rid static check errors.

modelopt/torch/export/quant_utils.py
realAsma: BB: remove this then

modelopt/torch/export/quant_utils.py
realAsma: BB: Is there an easy way to implement this? PTAL and tell me. if this is simple to do, can you do that?

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

🐝 On it.

realAsma: > _:robot: Bot comment._

🐝

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

Yes. This is a small export-only change: for compressed INT4/W4A8 AWQ weights, validate the logical input dimension from `weight.metadata["shape"][-1]` before returning `weight.data`, and add focused coverage for both formats. I’ll include it with removal of the redundant scale-count guard.

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

Agreed. The scale-count guard is redundant for supported export paths, so I’ll remove it and its mismatch-only test in the same reviewed fix batch. No source change has been published yet.

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

Fixed in 89bdde5: compressed INT4 and W4A8 AWQ wrappers now validate the logical input dimension before the fast return, with focused regression coverage.

modelopt/torch/export/quant_utils.py
realAsma: > _:robot: Bot comment._

Addressed in 89bdde5: removed the redundant scale-count guard and its mismatch-only test while retaining the block-shape validation.

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,22 @@ def process_layer_quant_config(layer_config_dict):
...
+        )
+
+
+def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
     """Packs the INT4 weights into uint8 tensor."""
     out_dim = weight.shape[-2]
     assert out_dim % 2 == 0, f"Cannot pack weight. Out dimension {out_dim} is not an even number."
     in_dim = weight.shape[-1]
-    block_size = weight.shape[-1] // weights_scaling_factor.shape[-1]
+    _validate_int4_block_size(in_dim, block_size)
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

This was intentionally removed at reviewer request in `89bdde5`. Supported export paths derive the scale tensor and configured block size from the same quantizer state, so a mismatch represents malformed internal state rather than the reported partial-block case. We keep validation at the export boundary and avoid a redundant internal invariant check.

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,22 @@ def process_layer_quant_config(layer_config_dict):
...
+    if not isinstance(block_size, int) or block_size <= 0:
+        raise ValueError(f"Block size must be a positive integer, got {block_size}.")
+    if in_dim % block_size != 0:
+        raise NotImplementedError(
+            f"Cannot pack weight with input dimension {in_dim} and block size {block_size}: "
+            "partial blocks are not supported."
+        )
+
+
+def pack_int4_in_uint8(weight, weights_scaling_factor, block_size):
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

`pack_int4_in_uint8` is an internal helper and is not re-exported from the package API. The `to_quantized_weight` default remains necessary for its non-AWQ formats; AWQ callers already pass a block size, and direct omission now produces the explicit validation error covered by the focused test.

modelopt/torch/export/quant_utils.py
```diff
@@ -792,12 +792,22 @@ def process_layer_quant_config(layer_config_dict):
...
 
-def pack_int4_in_uint8(weight, weights_scaling_factor):
+def _validate_int4_block_size(in_dim, block_size):
+    if not isinstance(block_size, int) or block_size <= 0:
+        raise ValueError(f"Block size must be a positive integer, got {block_size}.")
+    if in_dim % block_size != 0:
+        raise NotImplementedError(
+            f"Cannot pack weight with input dimension {in_dim} and block size {block_size}: "
+            "partial blocks are not supported."
+        )
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Added the compatibility note in `CHANGELOG.rst` in `ffc0c3a`. Quantization-time validation remains intentionally out of scope because quantization also supports deployment-independent accuracy evaluation.

meenchen (approved):

Edwardf0t1 (approved):
LGTM

PR #2328, opened 2026-09-03
Rename modelopt_recipes/huggingface to model_type with backward-compat alias

shengliangxu:
### What does this PR do?

Type of change: Refactor + deprecation (recipe-library restructure, backward compatible), plus an unrelated transformers-compat test fix.

Rename the architecture-specific recipe tier `modelopt_recipes/huggingface/` to
`modelopt_recipes/model_type/`, making explicit that it holds recipes **shared across
every checkpoint of a Hugging Face `model_type`** — as opposed to the checkpoint-mirror
`models/<org>/<model_id>/` tier. The old `huggingface/` path keeps working as a
deprecated backward-compat alias (a source-tree symlink plus a loader alias), so no
saved `--recipe` path breaks.

- **Loader alias** (`modelopt/recipe/loader.py`): generalized so saved
  `--recipe huggingface/<model_type>/...` paths rewrite to `model_type/...`, alongside
  the existing `huggingface/models/... -> models/...` rewrite (checked first as the more
  specific prefix). This keeps old paths resolving for pip-installed wheels, where the
  source-tree symlinks don't survive.
- **Internal `$import`s**: rewritten from `huggingface/... -> model_type/...` inside the
  shipped recipes so they resolve without the symlink — mandatory for wheels, since
  `$import` resolution goes through `config_loader` (no alias there).
- **Packaging** (`pyproject.toml`, `MANIFEST.in`): extended the symlink-exclusion globs
  so the recursive `**/*.yaml` package-data glob doesn't double-ship recipes through the
  `huggingface -> model_type` and `model_type/models -> ../models` symlinks.
- **Docs / examples / skills / tests**: migrated all internal references to the canonical
  `model_type/`; `huggingface/` remains only in the deprecated-alias tests and explanatory
  notes.
- **Unrelated fix (2nd commit):** `tests/unit/torch/export/test_quant_aware_conversion.py`
  failed on transformers>=5.9, which dropped `base_model_prefix` from
  `WeightTransform.__slots__` (the scoped-rule tests assigned it on the now-slotted
  object). Production `_scope_prefixes` already reads it via `getattr(..., None)` and
  degrades correctly, so there is no runtime change — the tests now set it through a
  helper that suppresses `AttributeError` across the supported transformers range.

### Usage

```bash
# New canonical path
python examples/hf_ptq/hf_ptq.py --model <ckpt> \
    --recipe model_type/qwen3_vl/ptq/fp8_vision-kv_none

# Old path still works (deprecated backward-compat alias)
python examples/hf_ptq/hf_ptq.py --model <ckpt> \
    --recipe huggingface/qwen3_vl/ptq/fp8_vision-kv_none
```

```python

load_recipe("model_type/vit/ptq/fp8")    # canonical
load_recipe("huggingface/vit/ptq/fp8")   # deprecated alias, resolves to the same recipe
```

### Testing

- `tests/unit/recipe/` — **336 passed**, including the new
  `test_load_recipe_huggingface_arch_backward_compat_alias` and the updated
  structural/doc tests (`test_recipe_docs.py`).
- `tests/unit/torch/export/test_quant_aware_conversion.py` — **16 passed** (was 4 failed
  on transformers 5.9.0).
- Built an sdist **and** a wheel and inspected both manifests: each recipe ships exactly
  once (29 `model_type/`, 13 `models/`, 2 `timm/`, 162 total) with **zero** `huggingface/` or
  `model_type/models/` duplicates and no build error on the symlinks.
- Simulated a wheel install (symlink-free extracted tree) and confirmed
  `huggingface/<arch>/...`, `model_type/...`, and `huggingface/models/...` all resolve via
  the loader alias — including a recipe that pulls internal `$import`s.

- Is this change backward compatible?: ✅ — old `huggingface/...` recipe paths keep resolving via the symlink + loader alias.
- Did you write any new necessary tests?: ✅ — backward-compat alias test added; structural/doc tests updated to the new layout.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — Deprecations entry under 0.48.0. (The transformers-compat test fix is not changelog-worthy.)
- Did you get Claude approval on this PR?: ❌ — not yet.

### Additional Information

The `model_type/models -> ../models` symlink is kept purely as a backward-compat alias for
old `huggingface/models/<org>/<model_id>/...` paths; `model_type/` is otherwise
architecture-only. If we ever want it strictly architecture-only, that symlink can be
dropped later without breaking anything, since the loader rewrites `huggingface/models/...`
straight to the top-level `models/` tier.

## Summary by CodeRabbit

- **New Features**
  - Added post-training quantization recipes for Gemma, Gemma 4, MiniMax-M3, Nemotron, Qwen, Step-3.7, ViT, and other architectures.
  - Added vision, multimodal, mixed-precision, and experts-only quantization options.

- **Documentation**
  - Standardized architecture-specific recipes under `model_type/` and updated examples and guidance.

- **Compatibility**
  - Legacy `huggingface/` recipe paths remain supported with deprecation warnings.
  - Local recipe files now take precedence over built-in recipes.
  - Deprecated quantization-format flags warn when explicitly provided.


cjluo-nv (commented):

Mostly a clean mechanical rename with good alias tests, but one user-facing path was missed and the loader's "local tree is never shadowed" claim doesn't hold for shipped `model_type`s.

Needs action:
- [ ] Update the stale path in `modelopt/torch/quantization/model_quant.py` — the `_check_weight_quantization_took_effect` error still tells users to look under `modelopt_recipes/huggingface/<model_type>/`.
- [ ] Fix or narrow the comment in `modelopt/recipe/loader.py`: built-in candidates are probed *before* the filesystem, so a local `huggingface/<shipped_model_type>/...` tree is now shadowed on wheel installs (see inline).
- [ ] Decide whether resolving a deprecated `huggingface/` prefix should emit a warning; today the rewrite is silent, so nothing nudges users off the deprecated path before removal.
- [ ] Consider asserting instead of silently suppressing in `_set_scope_attr` (`tests/unit/torch/export/test_quant_aware_conversion.py`) so a renamed/typo'd attribute can't quietly skip the assignment (see inline).

No action needed:
- Test edits are justified: `test_recipe_docs.py` tracks the new layout and both symlinks; new alias/`$import`/local-override tests were added.
- 73 files but a small, cohesive net diff; no licensing changes.

modelopt/recipe/loader.py
```diff
@@ -58,24 +58,28 @@ def _resolve_recipe_path(recipe_path: str | Path | Traversable) -> Path | Traver
...
+        # Backward-compat aliases for the recipe-library restructure. A source checkout keeps
+        # ``huggingface`` -> ``model_type`` (and the nested ``model_type/models`` -> ``../models``)
+        # symlinks, but symlinks don't survive into built wheels, so the deprecated tier
+        # prefixes are rewritten (see ``_alias_builtin_recipe_prefix``) before the built-in
+        # lookup — that keeps saved ``--recipe huggingface/...`` paths working for pip-installed
+        # users, not just source checkouts. The rewrite is applied to built-in candidates only;
+        # the filesystem fallback below probes the path exactly as given first, so a user's own
+        # local ``huggingface/`` recipe tree still loads by its natural relative name.
+        aliased = _alias_builtin_recipe_prefix(rp_str)
+
```
cjluo-nv, line 69: > _Bot comment._

The comment says "the filesystem fallback below probes the path exactly as given first, so a user's own local `huggingface/` recipe tree still loads by its natural relative name" — but the built-in loop above runs before any filesystem probe. For a pip-installed user with a local `huggingface/qwen3_vl/ptq/fp8_vision-kv_none.yaml`, the built-in lookup previously missed (no `huggingface/` dir in the wheel) and the local file won; with the alias, `model_type/qwen3_vl/...` now resolves first and silently shadows it. The new test deliberately uses `my_custom_arch`, which is not a shipped `model_type`, so it doesn't cover this case. Either probe the filesystem for the un-aliased path before the aliased built-in candidate, or reword the comment to state the guarantee only holds for names that don't collide with a shipped `model_type`.

Note `config_loader._resolve_config_path` already probes filesystem-first, so the two resolvers disagree on override precedence.

tests/unit/torch/export/test_quant_aware_conversion.py
```diff
@@ -40,6 +41,20 @@
...
+def _set_scope_attr(transform, name, value):
+    """Set an optional scoped-match attribute that only some transformers versions expose.
+
+    transformers>=5.9 dropped ``base_model_prefix`` from ``WeightTransform``'s ``__slots__``
+    (scoped matching now keys off ``scope_prefix`` alone); older supported versions still
+    carry it. Production ``_scope_prefixes`` reads it via ``getattr(..., None)``, so skipping
+    the assignment where the slot is absent is equivalent — and lets these tests run across
+    the whole supported transformers range instead of ``AttributeError``-ing on the setattr.
+    """
+    with contextlib.suppress(AttributeError):
```
cjluo-nv: > _Bot comment._

`contextlib.suppress(AttributeError)` makes the assignment a silent no-op for any failure, not just the slotted-`base_model_prefix` case — a typo'd attribute name or a future rename would also pass unnoticed, and the caller can't tell which branch of `_scope_prefixes` it exercised. Consider computing the capability once (e.g. `_HAS_BASE_MODEL_PREFIX = "base_model_prefix" in getattr(WeightTransform, "__slots__", ())`) and skipping/branching explicitly, so the suppression is scoped to the one attribute you know is version-dependent.

modelopt/recipe/loader.py
shengliangxu: Fixed in 0017a06394. `_resolve_recipe_path` now probes the filesystem for the un-aliased path first, then the aliased built-in candidate — the same precedence as `config_loader._resolve_config_path`, so the two resolvers no longer disagree. A local `huggingface/<type>/...` (or `model_type/<type>/...`) tree now wins even when `<type>` collides with a shipped `model_type`, instead of being silently shadowed by the rewrite. Updated the code comment to state this, and replaced the `my_custom_arch` test with `test_load_recipe_local_tree_overrides_builtin_even_on_name_collision`, which writes a file at a *shipped* recipe's exact relative path and asserts the local file wins (the collision case the non-shipped name missed).

tests/unit/torch/export/test_quant_aware_conversion.py
shengliangxu: Fixed in 0017a06394. Replaced `contextlib.suppress(AttributeError)` with a `try/except AttributeError` that re-raises unless `name == "base_model_prefix"`, so the suppression is scoped to that one version-dependent slot and a typo or future rename now surfaces instead of silently no-op-ing. Kept the check lazy rather than a module-level `WeightTransform.__slots__` capability constant because this test file must import without transformers installed (the converter classes are imported function-locally under `importorskip`).

tests/unit/recipe/test_loader.py
```diff
@@ -160,14 +160,40 @@ def test_load_recipe_builtin_description():
...
+    """Old ``huggingface/<model_type>/...`` recipe paths resolve to the renamed
+    ``model_type/`` tier.
+
+    ``huggingface/`` was renamed to ``model_type/``. A source checkout keeps a
+    ``huggingface`` -> ``model_type`` symlink, but symlinks don't survive into built
+    wheels, so the loader rewrites the prefix directly. This guards that saved
+    ``--recipe huggingface/<model_type>/...`` paths keep working for pip-installed
+    users, not just source checkouts.
+    """
+    from modelopt.recipe.loader import _resolve_recipe_path
```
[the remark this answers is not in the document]
shengliangxu: Fixed in 0017a06394. Hoisted `_resolve_recipe_path` (from `modelopt.recipe.loader`) and `_alias_builtin_recipe_prefix`/`_resolve_config_path` (from `modelopt.torch.opt.config_loader`) to module scope, and removed the function-local imports at all four sites (173/198/218/260).

cjluo-nv (commented):

The two inline concerns from last round are properly fixed, but the stale `huggingface/` path in `model_quant.py` is still there and the precedence flip left `_resolve_recipe_path`'s docstring wrong.

Needs action:
- [x] Update the error message in `_check_weight_quantization_took_effect` (`modelopt/torch/quantization/model_quant.py`) — it still points users at `modelopt_recipes/huggingface/<model_type>/`, a path that does not exist in a wheel.
- [x] Fix the now-inaccurate docstring in `modelopt/recipe/loader.py:53` ("checking the built-in library first then the filesystem") — resolution is filesystem-first as of this revision.
- [x] Decide whether resolving a deprecated `huggingface/` prefix should emit a `FutureWarning`; the rewrite is still silent, so nothing nudges users off the deprecated path before removal.
- [x] Confirm the built-in→filesystem precedence flip in `_resolve_recipe_path` is intended as a user-visible behavior change and, if so, add a `CHANGELOG.rst` line — a local `general/` or `models/` tree in cwd now wins over a shipped recipe.

No action needed:
- :heavy_check_mark: Resolved since the last review: the loader shadowing/comment issue (now filesystem-first with a shipped-name collision test) and the over-broad `contextlib.suppress` in `_set_scope_attr`.
- Test edits are justified: `test_recipe_docs.py` tracks the new layout and both symlinks; new alias/`$import`/override tests added. No licensing changes.



---

- `modelopt/recipe/loader.py:53` — > _Bot comment._

This summary line is now wrong: after the precedence flip the function probes the filesystem first and the built-in library second. Please reword (e.g. "checking the filesystem first, then the built-in library, applying the deprecated-tier alias to the built-in lookup") so the docstring matches the body comment a few lines below.

CHANGELOG.rst
```diff
@@ -21,8 +21,8 @@ Changelog
...
...`. ``examples/hf_ptq``: ``--qformat`` and ``--kv_cache_qformat``. ``examples/megatron_bridge/quantize.py``: ``--quant_cfg``, ``--kv_cache_quant`` and ``--weight_only``. ``examples/torch_onnx/torch_quant_to_onnx.py``: ``--qformat``. A recipe carries the quantization config, the calibration algorithm and the KV-cache setting in one file, so they cannot drift apart the way separate flags can -- and ``--recipe`` already took precedence over all six, silently on ``hf_ptq`` and with a warning on ``megatron_bridge`` -- with one gap the recipe closes rather than inherits: a weight AutoQuantize recipe that omits ``kv_cache`` still falls back to ``--kv_cache_qformat``, so set ``kv_cache`` in the recipe when migrating. Use a recipe from ``modelopt_recipes/general/ptq/`` or a model-specific one under ``modelopt_recipes/models/``. The warning fires only when a flag is passed explicitly: ``--qformat`` defaults to ``fp8`` and ``--kv_cache_qformat`` to ``fp8_cast``, so warning on the defaults would fire on every run, including runs that correctly use ``--recipe``. ``examples/speculative_decoding/scripts/quantize_drafter.py`` keeps ``--qformat`` undeprecated: it has no ``--recipe`` alternative yet.
```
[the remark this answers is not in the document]
shengliangxu: Fixed in `1ecfba579b` — the `--recipe` migration guidance now names all three tiers: `general/ptq/`, architecture-specific `model_type/<model_type>/`, and checkpoint-specific `models/`.

cjluo-nv (commented):

All four prior concerns are fixed in `1ecfba5`; what remains is a human sign-off on a 74-file rename that changes recipe-resolution precedence and packaging globs.

Needs action:
- [ ] Human sign-off required: the PR is large (74 files) and lands a user-visible precedence flip in `_resolve_recipe_path` plus new `exclude-package-data` globs whose effect only shows up in a built wheel.
- [ ] Confirm the sdist/wheel manifest counts quoted in the PR body (27 `model_type/`, 11 `models/`, 153 total, zero `huggingface/` duplicates) still hold on the final commit — the symlink exclusions in `pyproject.toml`/`MANIFEST.in` cannot be verified from source alone.

No action needed:
- :heavy_check_mark: Resolved since the last review: the stale `modelopt_recipes/huggingface/<model_type>/` path in `model_quant.py`, the filesystem-first docstring at `loader.py:53`, the silent deprecated-prefix rewrite (now a `FutureWarning`, covered by three tests), and the changelogged precedence flip.
- Test edits are justified: `test_recipe_docs.py` tracks the new layout and both symlinks; new alias, `$import`, and local-override/collision tests were added rather than removed. No licensing changes.

chadvoegele (approved):

modelopt/recipe/loader.py
```diff
@@ -50,32 +52,50 @@
...
+            candidate = BUILTIN_RECIPES_LIB.joinpath(aliased + suffix)
             if candidate.is_file() or candidate.is_dir():
+                if aliased != rp_str:
+                    warnings.warn(
+                        f"Recipe path {rp_str!r} uses a deprecated recipe-tier prefix; it "
+                        f"resolved to the built-in {aliased!r}. Update saved ``--recipe`` paths "
+                        "to the new prefix -- the deprecated one will be removed in a future "
+                        "release.",
+                        FutureWarning,
+                        stacklevel=2,
```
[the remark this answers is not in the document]
shengliangxu: Bumped to `stacklevel=3` in `c47bbb6f61` so the warning points at the caller of `load_recipe` for the common `load_recipe(path)` case. As you noted, no single value is exact across wrapper paths (e.g. `resolve_quant_cfg_from_args`), so the deprecated path stays in the message itself — accurate regardless of which frame the stacklevel lands on. (`skip_file_prefixes` would be cleaner but is 3.12+ only; the repo supports 3.10+.)

kevalmorabia97 (approved):

meenchen (approved):

ajrasane (approved):

PR #2334, opened 2026-09-04
Unblock Qwen3.5/3.6 QAD: Megatron export, calibration, and distillation fixes

kevalmorabia97:
### What does this PR do?

Type of change: Bug fix

Everything that blocked running QAD on a quantized Qwen3.5 / Qwen3.6 MoE checkpoint: two
Megatron-Core → HuggingFace export bugs that make it unservable (§1–2), the dead code the first
leaves behind (§3), a no-op flag (§4), a multi-GPU calibration deadlock (§5), and four
distillation / data-prep bugs that stopped QAD itself from running (§6).

#### 1. Routed experts were exported packed, and vLLM cannot load that

```
AttributeError: Layer language_model.model.layers.23.mlp.experts has no parameter
  'w2_weight_weight_scale_2' for checkpoint weight ...experts.down_proj_weight_scale_2
```

`mcore_qwen35vl.py` used `GroupedMLPPacking`, mirroring the **BF16 upstream** checkpoint, which
really is packed. But that mapping is only used for **quantized** export, and vLLM's quantized MoE
loader needs per-expert scales — both released NVFP4 checkpoints (`Qwen3.6-35B-A3B-NVFP4` via
hf_ptq, `Nemotron-3.5-Lightning-30B-A3B-NVFP4` via Megatron-LM) are per-expert.

`_grouped_mlp_slicing` gains `gate_proj_name` / `up_proj_name` to split each expert's fused gate+up
and slice its per-block `weight_scale`; `GroupedGatedMLPSlicing` wires it up. The Megatron
checkpoint layout is unchanged, so affected checkpoints need only a **re-export**.

`_verify_exported_keys` is relaxed to match: exported modules now contribute their ancestor
prefixes, so expanding one source module into many is not reported as ~82 dropped tensors. A module
genuinely absent still has nothing beneath its prefix and is still caught.

#### 2. A quantized `output_layer` (`lm_head`) could not be checkpointed

`GPTModel.sharded_state_dict` drops `output_layer._extra_state` and asserts it is empty. ModelOpt
keeps quantizer state there, so saving raised and — since that method also backs the load plan —
loading silently restored the layer **unquantized**.

`keep_gpt_output_layer_extra_state()` retains it, applied from
`megatron_replace_quant_module_hook` so **every** Megatron model gets it (Megatron-LM and NeMo
users included, neither of whom can import `mbridge`, which needs `megatron.bridge`). It matches
the upstream body by AST before replacing it and self-disables otherwise.
[NVIDIA/Megatron-LM#7086](https://github.com/NVIDIA/Megatron-LM/pull/7086) is **closed, not
merged**: nemo:26.10 migrates `GPTModel` to `HybridModel`, whose `sharded_state_dict` has no
pop-and-assert, so this side keeps the workaround.

Not cosmetic: `lm_head` is 248320×2048 = 509M params, **34.6% of per-token weight traffic** on a
model with ~2.9B active params.

#### 3. Cleanup

Nothing maps `GroupedMLPPacking` once qwen3_5 is switched over; it is removed with
`_grouped_mlp_packing` and the `quantize=` / `record_quant_config=` parameters that existed only to
serve it. Llama-4's `PackNameRemapping` is unaffected.

Two smaller review-driven fixes: the gated-split shape checks raise `ValueError` rather than
`assert` (stripped under `-O`), and per-expert quant metadata is recorded for
`local_expert_indices` rather than every global id, fixing non-contiguous EP.

#### 4. Remove the no-op `--moe_calib_experts_ratio` from the Megatron quantize example

`examples/megatron_bridge/quantize.py` accepted the flag and threaded it into the `mtq` config, but
`_moe_calib_experts_ratio` exists only in `plugins/huggingface.py` (9 refs) and never in
`plugins/megatron.py` (0); `mode.py:247` only assigns it to modules already exposing the attribute.
On a Megatron MoE model it was accepted and silently ignored — a trap, since on a 256-expert model
it reads like a major quality lever. `hf_ptq.py` keeps it, where it works.

#### 5. Fix multi-GPU image-text (VLM) calibration deadlocking

VLM calibration hung for 30 minutes and died on a gloo timeout whenever `world_size > 1`, with no
error until the timeout fired.

`NemotronTarPlusJsonlIterable` split its budget with truncating division, so the stream supplied
fewer samples than requested (1024 over 3 subsets → 341×3 = **1023**). `_ShardedIterable` gives
rank *r* items *r, r+W, r+2W…*, so a stream that is not a multiple of `world_size` leaves the
trailing rank one short — it exits the forward loop early and the others block on the next
collective. The arithmetic predicts both observed hangs exactly: 1024 → stall at **255/256**,
512 (yielding 510) → **127/128**.

Fixed both ends: subset budgets are distributed with `divmod` so they sum exactly, and
`_ShardedIterable` truncates every rank to `floor(len / world)` — which also covers `num_samples`
not being divisible by `world_size`, as the first fix alone does not.

Verified on Qwen3.6-35B-A3B (EP=4, `nemotron_vlm_dataset_v2`, 1024 samples): the configuration that
hung twice now completes 256/256 and exports. Unit tests cover both fixes and fail without them.

#### 6. Fix the distillation path so QAD can actually run

Four independent bugs, all hit while running QAD end to end on Qwen3.6-35B-A3B. Each blocks a
different configuration, and together they made every sequence length OOM or abort.

- **Context parallel aborts.** The DDP config derived `average_in_collective` from `--sft` alone,
  but context parallel also needs per-token loss reduction, so any `--cp_size > 1` run died on
  `Cannot average in collective when calculating per-token loss`.
- **`TopKLogitsKLLoss` was not memory-efficient.** Despite documenting "without gathering full
  logits", it cast the *whole* vocabulary to FP32 before selecting the top-k, allocating two
  `[seq, vocab]` tensors — 30.3 GiB each at seq 32768 on this model's 248k vocab. Reducing before
  the cast is equivalent: widening is exact and temperature scaling is monotonic, so the selected
  entries and the loss are unchanged.
- **MTP cross-entropy ran when it had nothing to recover.** `skip_lm_loss` exempts the MTP heads
  unconditionally, so their CE materialised another FP32 `[seq, vocab]` tensor even when the MTP
  head is excluded from quantization — as it is in every recipe here (775 of 906
  `exclude_modules`, zero MTP `weight_scale` tensors exported). It is now skipped **only** when the
  model is quantized and MTP is left out of it; plain distillation such as pruning recovery still
  trains the MTP head. `test_mtp_excluded_from_quantization` pins all four cases.
- **One bad record deadlocked data prep.** `megatron_preprocess_data` re-raised chat-template
  failures out of a pool worker, stalling the whole job until it timed out — three malformed
  records cost a multi-hour tokenization run. They are now skipped with a warning, matching the
  existing handling of malformed JSONL a few lines above.

Also exposes `--logit_kl_topk`, which `DistillationConfig` has supported for a while but the
example never passed through; `test_qad` now exercises that path.

§4, §5 and §6 are independent of §1–3; happy to split them out if reviewers prefer.

### Usage

No API change. Exported names now match the released checkpoints:

```
model.language_model.layers.0.mlp.experts.<E>.{gate,up,down}_proj.{weight,weight_scale,weight_scale_2}
lm_head.{weight,weight_scale,weight_scale_2}
```

### Testing

- `test_mcore_export_mappings.py` — qwen3_5 mappings emit per-expert rules. Verified these fail
  without the fix (2 failed / 11 passed), with `Qwen3MoeForCausalLM` / `NemotronHForCausalLM` as
  controls.
- `test_unified_export_megatron.py` — the gate/up split, per-block scale slicing, the 0-dim scalar
  fallback, and both directions of the `_verify_exported_keys` relaxation.
- `test_megatron.py::TestKeepGptOutputLayerExtraState` — 15 cases: payload detection, no-op second
  call, warn-and-skip on an unrecognised `sharded_state_dict`, and `test_patches_stock_megatron_core`
  which installs a replica of the real pre-fix upstream body (verified against `be08ce5b1~1`) so the
  patched path is exercised whichever megatron-core is installed.
- `test_qad.py` — CI caught that its reference comparison still assumed packed experts; fixed.

**End to end on `Qwen/Qwen3.6-35B-A3B` (35B MoE, 256 experts), 4×GB200, nemo:26.08:**

| | before | after |
| --- | --- | --- |
| export self-check | `Export dropped 82 tensor(s)` | passes |
| expert tensors | `mlp.experts.gate_up_proj` (packed) | `mlp.experts.<E>.{gate,up,down}_proj` |
| **vLLM v0.28.0 load** | **`AttributeError`, engine never starts** | **`Loading weights took 25.61 s`** |
| **NEL eval (GPQA-D, MMMU-Pro)** | **FAILED** | **SUCCESS** |

### Results these fixes unblocked

The export fix is what made a Megatron-produced NVFP4 MoE checkpoint servable at all, so it enabled
a full PTQ study on Qwen3.6-35B-A3B. Accuracy deltas are against a BF16 baseline measured on the
same harness, from **paired** per-question tests:

| recipe | throughput vs BF16 | GPQA-D | SciCode ×8 | MMMU-Pro | IFBench |
| --- | --- | --- | --- | --- | --- |
| **W4A16** (weight-only) | **0.64–0.86×** — *slower* | −0.06 | −0.15 | +0.48 | −0.44 |
| **W4A4** | 8/12 shapes faster | −0.60 | −0.70 | −1.48 (p=0.019) | −0.53 |
| **W4A4 + 4-bit `lm_head`** | **9/12 shapes**, up to **1.30×** | **+0.03** (p=0.96) | −0.81 | **−1.16** (p=0.016) | −1.65 (ns) |

Repeats: GPQA-D is `pass@1[avg-of-16]`; SciCode is 8 pooled runs per recipe; MMMU-Pro is 3 runs per
side and IFBench 2–3 for BF16 and the last row, 1 elsewhere. AA-LCR (68.33 → 71.33, p=0.25, 3 runs
per side) and τ²-Telecom (94.25 → 94.25, 3 runs per side) are on par; at 100 questions and 114
tasks they cannot resolve below ~5 pp and ~3 pp, so they carry no claim either way.

#### QAD status (what §6 unblocked)

With the §6 fixes in place, QAD runs end to end on this model: 32 nodes, `TP=1 PP=1 CP=1 EP=8`,
seq 32768, gbs 512, ~38 s/iter, 124 GB/GPU peak. First accuracy read, MMMU-Pro at iteration 50
(0.84 B tokens), 3 runs per side, paired per-question:

| | MMMU-Pro | vs BF16 |
| --- | --- | --- |
| BF16 | 74.55 | — |
| W4A4 + 4-bit `lm_head` (PTQ) | 73.39 | **−1.16, p=0.016** |
| + QAD, iteration 50 | 73.78 | −0.77, p=0.089 (ns) |

The PTQ deficit that motivated this work is no longer statistically significant after 50 QAD
iterations. The improvement itself (+0.39 vs PTQ) is **not** significant at p=0.41, and 50
iterations is 10% of the planned budget, so this is a direction rather than a result. A full
six-benchmark sweep at iterations 50 and 300 is running; these numbers will be superseded.

Two findings worth flagging beyond this PR:

- **Weight-only NVFP4 is slower than BF16 on Blackwell.** W4A16 leaves activations in BF16, so vLLM
  cannot use the FP4 tensor cores and falls back to `MarlinNvFp4LinearKernel` / `'MARLIN'` MoE.
  W4A4 selects `FLASHINFER_TRTLLM` + `FlashInferCuteDslNvFp4LinearKernel` and beats W4A16 in
  **12/12** shapes. The Marlin line count tracks the recipe exactly (one W4A16 layer ⇒ one Marlin
  line ⇒ zero once `lm_head` is W4A4).
- **The only accuracy cost is multimodal**: **−1.2 pp on MMMU-Pro** for the fastest recipe,
  confirmed over 3 runs per side (p=0.016). GPQA-D, SciCode, IFBench, AA-LCR and τ²-Telecom show no
  significant regression.

### Additional Information

Both export bugs were found while reproducing `nvidia/Qwen3.6-35B-A3B-NVFP4` through
`examples/megatron_bridge/`. Follow-up to #2332. Upstream counterpart
[NVIDIA/Megatron-LM#7086](https://github.com/NVIDIA/Megatron-LM/pull/7086) is closed — see §2.
Labeled `cherry-pick-0.47.0`.


modelopt/torch/export/plugins/mcore_qwen35vl.py
```diff
@@ -47,19 +47,12 @@
...
-    "local_experts.linear_fc2": PackNameRemapping(
-        "model.layers.{}.mlp.experts.down_proj",
-        {"layer_type": "linear_fc2", "transpose": False},
-    ),
-    # Same packed layout from fused TEGroupedMLP, so grouped GEMM stays usable.
-    "experts.linear_fc1": GroupedMLPPacking("model.layers.{}.mlp.experts.gate_up_proj"),
-    "experts.linear_fc2": GroupedMLPPacking("model.layers.{}.mlp.experts.down_proj"),
+    # One entry per expert with gate/up split, matching the released NVFP4 checkpoints: vLLM's
+    # quantized MoE loader has no parameter for a packed `experts.down_proj_weight_scale_2`.
+    "local_experts.linear_fc1": GatedMLPSlicing("model.layers.{}.mlp.experts.{}."),
```
cjluo-nv: > _Bot comment._

These four rules change the exported tensor names, but `tests/gpu_megatron/torch/export/test_unified_export_megatron.py` still parametrizes `qwen3_5_moe_vl_grouped` / `qwen3_5_moe_vl_sequential` and calls `assert_exported_checkpoint_matches(tmp_export_dir, model_dir, check_values=False, allow_unexpected=("mlp.gate.expert_bias",))` with the default `allow_missing=()`.

The fixture (`_pack_qwen3_5_moe_experts` in `tests/_test_utils/torch/transformers_models.py`) deliberately repacks the tiny model to `mlp.experts.gate_up_proj` / `mlp.experts.down_proj`, so after this change those two reference tensors per layer are no longer in the export and the helper should raise `"N reference tensor(s) absent from the export"`. Please update that test (and the now-stale comment there: "Both layouts must reach the same packed HF tensors, via GroupedMLPPacking (TEGroupedMLP) and PackNameRemapping (SequentialMLP)") plus the fixture docstring, otherwise CI regresses on a change that is otherwise correct.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1467,31 +1484,46 @@ def _grouped_mlp_slicing(
                     weight_scale_2.detach().cpu().clone() if weight_scale_2 is not None else None
                 )
 
-                if weight_scale_cpu is None:
-                    local_expert_state[expert_prefix + "weight"] = weight
+                # Gated: split gate+up rows, each taking its slice of a per-block weight_scale.
+                if _gated_subnames is None:
+                    shards = [(expert_prefix, weight, weight_scale_cpu)]
                 else:
```
cjluo-nv, line 1413: > _Bot comment._

The gate/up split is the substantive part of this PR and nothing in the diff exercises it — the added tests only assert on mapping-table entries. `test_unified_export_megatron.py` already has `_FakeTEGroupedMLP` + `_make_exporter_for_grouped_mlp`, so a CPU test could cover: per-block 2-D `weight_scale` sliced at `half`, the 0-dim/None scalar-scale fallback, `weight_scale_2` replicated to both shards, and the two `_record_layer_quant_config` entries per expert.

Two smaller points:
- This duplicates the split already implemented in `_gated_mlp_slicing` (which splits at `module.config.ffn_hidden_size` and handles the `len(shape) == 0` scale case identically). Consider extracting a small `_split_gate_up(weight, scale)` helper used by both so the two paths cannot drift.
- `half = weight.shape[0] // 2` silently truncates on an odd first dim; an `assert weight.shape[0] % 2 == 0` would fail loudly instead of emitting a short `up_proj`.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -498,7 +498,16 @@ def _verify_exported_keys(self, save_directory, pretrained_model_name_or_path) -
         # carries extras with no export counterpart, and only inside decoder layers, whose naming
         # is stable. A dropped decoder module is the case that loads fine and produces garbage.
         num_layers = self.model.config.num_layers
-        exported_modules = {key.rsplit(".", 1)[0] for key in exported}
+        # Ancestors too: an export may expand one source module into several (Qwen3.5 packs
+        # routed experts; the quantized export writes them per expert). Expansion is not a drop.
+        exported_modules = set()
+        for key in exported:
+            parts = key.rsplit(".", 1)[0].split(".")
```
cjluo-nv, line 505: > _Bot comment._

This relaxation is reasonable (it amounts to "the source module is covered if any exported tensor lives under its prefix"), but it does weaken the self-check beyond the expert case: any exported tensor under `...layers.N.mlp.<x>.<y>` now marks `...layers.N.mlp` as exported, so an architecture whose source stores a tensor directly at that module prefix would no longer be reported as dropped.

The PR body says this was verified in both directions (expansion accepted, genuine drop still raised) — please encode that as a unit test; it needs only a fake `self.model.config.num_layers` and two temp directories, no GPU.

modelopt/torch/export/plugins/mcore_custom.py
```diff
@@ -128,6 +128,18 @@ def __init__(self, target_name_or_prefix: str = "", func_kwargs: dict[str, Any]
         )
 
 
+class GroupedGatedMLPSlicing(CustomModuleMapping):
```
cjluo-nv, line 131: > _Bot comment._

With `mcore_qwen35vl.py` switched over, `GroupedMLPPacking` (and `GPTModelExporter._grouped_mlp_packing`, ~60 lines) has no remaining production caller — as do the `quantize=False` / `record_quant_config=False` parameters of `_grouped_mlp_slicing`, which exist solely for it. Either drop them in this PR or add a note on why they're kept; leaving two grouped-expert export paths where only one is reachable makes the next reader guess which one is current. The `transpose=False` comment in `_pack_name_remapping` ("most HF layouts want [num_experts, in, out], but Qwen3.5 keeps Megatron's orientation") is stale for the same reason.

cjluo-nv (commented):

The fix itself looks right: `Qwen3_5MoeForConditionalGeneration` has no import mapping, so switching the routed-expert rules affects export only, `use_moe_grouped_gemm` still finds `experts.linear_fc1` (existing `test_moe_layout_choice` stays green), and the gate/up split mirrors what `_gated_mlp_slicing` already does for the SequentialMLP path. Two things block a clean approval:

1. **An existing GPU test is very likely broken and was not updated.** `tests/gpu_megatron/torch/export/test_unified_export_megatron.py` runs `qwen3_5_moe_vl_grouped` (NVFP4 + FP8) and `qwen3_5_moe_vl_sequential` (NVFP4) against `create_tiny_qwen3_5_moe_vl_dir`, whose fixture is deliberately repacked to `mlp.experts.gate_up_proj` / `mlp.experts.down_proj` by `_pack_qwen3_5_moe_experts`. Those cases then call `assert_exported_checkpoint_matches(..., allow_missing=())`, which asserts every reference tensor is present in the export. With per-expert names the two packed reference tensors per layer are now absent → `"N reference tensor(s) absent from the export"`. The fixture comment ("every real Qwen3.5 checkpoint stores them packed") and the in-test comment ("Both layouts must reach the same packed HF tensors, via GroupedMLPPacking … and PackNameRemapping") also become wrong. Please update that test (and the helper's allowances, e.g. an expert-layout-aware comparison) in this PR — otherwise CI regresses even though the fix is correct.

2. **The behavior that actually changed has no test.** The two added cases only assert on entries in the mapping table; they never execute `_grouped_mlp_slicing` or `_verify_exported_keys`. The new gate/up shard loop (per-block scale slicing, scalar-scale fallback, `weight_scale_2` duplication, per-shard quant-config recording) and the ancestor-prefix relaxation in the self-check are both testable without a GPU — `test_unified_export_megatron.py` already has `_FakeTEGroupedMLP` + `_make_exporter_for_grouped_mlp` helpers that would cover the first, and the self-check "expansion accepted / genuine drop still raised" pair the PR body says was verified manually should be a unit test.

Minor: `GroupedMLPPacking` / `_grouped_mlp_packing` now have no production caller (only the mapping this PR removes used them), and `_grouped_mlp_slicing`'s `quantize=` / `record_quant_config=` parameters exist only to serve it — worth removing or noting why they stay. The `transpose=False` comment in `_pack_name_remapping` ("Qwen3.5 keeps Megatron's orientation") is now stale too.

modelopt/torch/export/plugins/mcore_qwen35vl.py
kevalmorabia97: Confirmed and fixed in 50ff828 — this was a real CI regression. The fixture does repack to `experts.gate_up_proj` / `experts.down_proj`, so the per-expert export made those two reference tensors per layer look dropped.

`assert_exported_checkpoint_matches` now gets `allow_missing=("mlp.experts.gate_up_proj", "mlp.experts.down_proj")` for the `qwen3_5_moe_vl*` cases, with a comment that the BF16 source packs experts while quantized export writes them per expert. Also refreshed the stale comment (now naming `GroupedGatedMLPSlicing` / `GatedMLPSlicing`) and the `_pack_qwen3_5_moe_experts` docstring, which claimed *every* real Qwen3.5 checkpoint is packed — true of the BF16 releases, not of the NVFP4 ones.

modelopt/torch/export/unified_export_megatron.py
kevalmorabia97: Fair — the mapping-table tests only covered naming. Added four CPU tests in 50ff828 using the existing `_FakeTEGroupedMLP` / `_make_exporter_for_grouped_mlp`:

- gate/up split per expert, asserting each shard equals the matching half of `weight{i}`
- per-block 2-D `weight_scale` sliced at `half`, `weight_scale_2` replicated to both shards, and both `_record_layer_quant_config` prefixes recorded
- the 0-dim scalar-scale fallback shared by both shards

Also added the odd-first-dim assert, so a mis-shaped expert fails loudly instead of emitting a short `up_proj`.

On extracting a shared `_split_gate_up`: I left the two paths separate for now. `_gated_mlp_slicing` takes a module and splits at `module.config.ffn_hidden_size`, while this path has only the raw per-expert tensor and splits at `shape[0] // 2`; the genuinely shared part is ~4 lines, so a helper would add indirection without removing much drift risk. Happy to extract it if you would rather.

modelopt/torch/export/unified_export_megatron.py
kevalmorabia97: Agreed on both counts, and the test is added in 50ff828 — `test_verify_exported_keys_accepts_expansion_but_still_catches_drops`, parametrized over the two directions: a packed source key with only per-expert tensors exported passes, and the same source with an unrelated tensor exported still raises `Export dropped`. No GPU, two tmp dirs and a fake `config.num_layers`, as you suggested.

You are right that the relaxation is broader than the expert case — any tensor under `...layers.N.mlp.<x>.<y>` now covers `...layers.N.mlp`. I kept it prefix-based rather than special-casing experts because the check is explicitly prefix-level by design ("compare module prefixes, not tensor names"), and a narrower rule would have to encode per-architecture expansion shapes. The failure it can no longer catch is a source tensor stored *directly* at a module prefix that also has exported children, which is a narrower hole than the one it closes.

modelopt/torch/export/plugins/mcore_custom.py
kevalmorabia97: Verified — `GroupedMLPPacking` and `_grouped_mlp_packing` have no in-tree caller after this change (`PackNameRemapping`, which gpt-oss and Llama-4 still use, is a different function).

I kept them rather than deleting in 50ff828, with a docstring note on why: quantized MoE export is per-expert because runtimes need per-expert scales, so the packing path only makes sense for out-of-tree mappings whose target checkpoints store experts packed. Removing it plus the `quantize=` / `record_quant_config=` parameters is a ~60-line delete in a bug fix that is labelled for cherry-pick to 0.47.0, so I would rather do it as a separate cleanup. Say the word if you want it in this PR instead.

tests/gpu_megatron/torch/export/plugins/test_mcore_export_mappings.py
```diff
@@ -46,3 +46,40 @@ def test_export_mapping_emits_kv_cache_scales(arch, prefix):
...
+@pytest.mark.parametrize("arch", sorted(PER_EXPERT_MOE_ARCHS))
+def test_moe_export_is_per_expert(arch):
+    mapping = all_mcore_hf_export_mapping[arch]
+    assert not mapping.get("use_packed_local_experts"), (
+        f"{arch} packs routed experts; vLLM cannot load packed quantized expert scales"
+    )
+    for rule in ("experts.linear_fc1", "experts.linear_fc2", "local_experts.linear_fc1"):
+        if rule in mapping:
+            assert "grouped_mlp_packing" not in mapping[rule].func_name
+            assert "pack_name_remapping" not in mapping[rule].func_name
```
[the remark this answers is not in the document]
kevalmorabia97: The gate/up split is now covered by CPU tests in 50ff828 (see the reply on `unified_export_megatron.py`) — per-expert `gate_proj` / `up_proj` / `down_proj` with their `weight_scale` and `weight_scale_2` shards.

The full GPU export path is also already exercised: `test_unified_export_megatron.py` parametrizes `qwen3_5_moe_vl_grouped` and `qwen3_5_moe_vl_sequential` with `NVFP4_DEFAULT_CFG` / `FP8_DEFAULT_CFG` and compares the exported checkpoint against the reference model, which this PR updates for the per-expert layout.

modelopt/torch/export/plugins/mcore_custom.py
kevalmorabia97: Done in 8874b85 — dropped it rather than keeping the note, so only one grouped-expert export path remains.

Removed: `GroupedMLPPacking`, `GPTModelExporter._grouped_mlp_packing`, and the `quantize=` / `record_quant_config=` parameters of `_grouped_mlp_slicing` that existed solely for it (the `quantize=False` branch and the `record_quant_config` guard are gone with them). Also fixed the stale `transpose=False` comment in `_pack_name_remapping`, which still cited Qwen3.5 keeping Megatron's orientation.

Net −107 lines. Verified no dangling references to `GroupedMLPPacking` / `grouped_mlp_packing` / `record_quant_config` anywhere in `modelopt/`, `tests/` or `examples/`; all 16 architecture mappings still resolve, and Llama-4's separate `PackNameRemapping` path is untouched (it maps to `pack_name_remapping`, a different function). 13 mapping tests and 8 export tests pass.

I also dropped the `assert "grouped_mlp_packing" not in mapping[rule].func_name` line from the new mapping test — with the function gone it could no longer fire. The `pack_name_remapping` half of that assertion stays, since that function still exists and is the layout this PR moves qwen3_5 away from.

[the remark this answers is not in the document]
kevalmorabia97: Skipping this one — I do not think the guideline applies to this module.

`modelopt/torch/export/plugins/mcore_custom.py` has no `__all__`, and neither does any of its nine sibling `mcore_*.py` modules. It defines 14 mapping classes today, none of them exported that way, so adding an `__all__` containing only `GroupedGatedMLPSlicing` would make the new class look special and the other 13 look excluded.

The second half does not apply either: `plugins/__init__.py` star-imports `megatron_importer`, `hf_spec_export` and `hf_checkpoint_utils`, but not `mcore_custom` — it is not referenced there at all. These mapping classes are internal building blocks that the per-architecture `mcore_*.py` modules import directly (`from .mcore_custom import GatedMLPSlicing, NameRemapping, ...`), not package public API, so there is nothing to re-export them into.

Happy to add a module-wide `__all__` covering all 14 classes as a separate cleanup if the guideline is meant to apply here, but that is a change to the module convention rather than something this bug fix should introduce.

cjluo-nv (commented):

Re-review of #2334 (11 files, +292/-130).

**Previous comments — status**
- *Critical: `qwen3_5_moe_vl_{grouped,sequential}` GPU test would regress* — **only half fixed**. `allow_missing=("mlp.experts.gate_up_proj", "mlp.experts.down_proj")` silences the "reference tensor(s) absent" assert, but `assert_exported_checkpoint_matches` also asserts on the *other* direction: every exported key not present in the reference and not ending in a quant suffix is "unexpected". The export now writes `...mlp.experts.<E>.{gate,up,down}_proj.weight`, none of which exist in the packed fixture, and `allow_unexpected` is still only `("mlp.gate.expert_bias",)`. Both parametrizations should fail with `Export produced unexpected tensors`. See inline.
- *Critical: gate/up split and `_verify_exported_keys` relaxation untested* — **addressed** (three CPU tests over `_FakeTEGroupedMLP` covering the split, per-block scale slicing, `weight_scale_2` replication, both `_record_layer_quant_config` prefixes and the 0-dim fallback, plus the parametrized expansion/drop test for the self-check).
- *Minor: dead `GroupedMLPPacking` / `_grouped_mlp_packing` / `quantize=` / `record_quant_config=` / stale `transpose=False` comment* — **addressed** (removed in 8874b85).
- *Minor (CodeRabbit): `__all__` for `GroupedGatedMLPSlicing`* — declined with a reasonable rationale (no sibling `mcore_*.py` defines `__all__`, and `mcore_custom` is not star-imported by `plugins/__init__.py`).

**New scope in this revision (design gate).** The MoE half introduces no new abstraction — it reuses `GroupedMLPSlicing`/`GatedMLPSlicing` and deletes a competing path, which is the right direction. The `lm_head` half does introduce a new mechanism: an AST-fingerprinted monkeypatch of `megatron.core` `GPTModel.sharded_state_dict`, invoked from three example `main()`s. The repo already has an in-tree owner for exactly this upstream assert — `modelopt/torch/quantization/plugins/megatron.py` (`quant_module_get_extra_state` returns `{}` for an unquantized `output_layer` precisely because "GPTModel.sharded_state_dict pops output_layer._extra_state and asserts it carries no data", and `megatron_replace_quant_module_hook` already walks the model at quantize time). The PR body doesn't say why the workaround lives in `utils/plugins/mbridge.py` behind three explicit call sites instead of there; as written, any other entry point (Megatron-LM examples, NeMo, direct library users) still hits the save-time `RuntimeError` / silent unquantized restore. It also has no test, unlike the MoE half. Please justify the placement in the PR body or move it, and add coverage.

Otherwise the export logic reads correctly: `Qwen3_5MoeForConditionalGeneration` has no import mapping so this is export-only; the grouped split at `shape[0] // 2` with per-block scale slicing mirrors `_gated_mlp_slicing`; the ancestor-prefix loop in `_verify_exported_keys` is idempotent and correctly short-circuits.

tests/gpu_megatron/torch/export/test_unified_export_megatron.py
```diff
@@ -244,10 +246,16 @@ def _test_unified_export_megatron(
 
     if rank == 0 and extra_module is None:
         # Names / shapes only: these Megatron weights are random, not loaded from model_dir.
+        allow_missing = ()
+        if model_type.startswith("qwen3_5_moe_vl"):
+            # The BF16 source packs routed experts; quantized export writes them per expert,
+            # as the released NVFP4 checkpoints do.
+            allow_missing = ("mlp.experts.gate_up_proj", "mlp.experts.down_proj")
```
cjluo-nv, line 259: > _Bot comment._

`allow_missing` only covers one half of the helper's contract. `assert_exported_checkpoint_matches` also computes

```python
unexpected = {k for k in set(exported) - set(reference)
              if not k.endswith(QUANT_SUFFIXES) and not any(a in k for a in allow_unexpected)}
assert not unexpected, f"Export produced unexpected tensors: ..."
```

The fixture (`_pack_qwen3_5_moe_experts`) leaves the reference with `...mlp.experts.gate_up_proj` / `...mlp.experts.down_proj` only, while the export now writes `...mlp.experts.<E>.gate_proj.weight`, `...up_proj.weight`, `...down_proj.weight`. None of those keys exist in the reference and none ends in a quant suffix, so both `qwen3_5_moe_vl_grouped` cases and the `qwen3_5_moe_vl_sequential` case should now fail with `Export produced unexpected tensors` instead of the previous `N reference tensor(s) absent`.

The per-expert names also never enter `shared`, so the shape check compares nothing for the experts — the test currently proves less about the new layout than it looks like it does. Suggest making the allowance explicit in both directions, e.g. add `"mlp.experts."` (or the three projection names) to `allow_unexpected` for these model types, and ideally assert positively that per-expert `gate_proj`/`up_proj`/`down_proj` with expected shapes are present, since the packed reference can no longer do it.

modelopt/torch/utils/plugins/mbridge.py
```diff
@@ -277,3 +280,52 @@ def load_modelopt_megatron_checkpoint(
...
+# Statement kinds of the upstream GPTModel.sharded_state_dict body we patch below.
+_GPT_SSD_STATEMENTS = ["Assign", "Assign", "Assign", "Assert", "Return"]
+
+
+def _output_layer_extra_state_has_data(entry: Any) -> bool:
+    """True when a sharded state-dict entry carries a payload."""
+    data = getattr(entry, "data", entry)
+    if isinstance(data, torch.Tensor):
+        return data.numel() > 0
+    return data is not None and bool(data)
```
cjluo-nv: > _Bot comment._

Two things on this helper.

**Placement.** The upstream assert this works around is already known to the library: `modelopt/torch/quantization/plugins/megatron.py::quant_module_get_extra_state` special-cases `output_layer` with the comment "``GPTModel.sharded_state_dict`` pops ``output_layer._extra_state`` and asserts it carries no data", and `megatron_replace_quant_module_hook` (registered in `CUSTOM_MODEL_PLUGINS`) already runs over every Megatron model at conversion time. Applying the fix there — or on the model instance from that hook — would cover Megatron-LM examples, NeMo and direct library users, whereas the current shape only fixes runs that go through the three `examples/megatron_bridge` `main()`s. If there's a reason it can't live in the quantization plugin (ordering, the patch needing to exist before restore, etc.), please state it in the PR body; the design gate fired on this PR and the body doesn't compare against that alternative.

**Tests.** Unlike the MoE half, none of the non-trivial behaviour here is covered: the `_GPT_SSD_STATEMENTS` fingerprint match, the self-disabling second call (the replacement's body is `Assign/Assign/Assign/If/Return`, so a repeat call returns `False` — worth locking in), and `_output_layer_extra_state_has_data` on a multi-element tensor / a 0-element tensor / `None`. A test alongside `tests/gpu_megatron/torch/export/plugins/test_moe_layout_choice.py` (which imports from this module already) would be enough.

Minor: when the fingerprint doesn't match, the function returns `False` silently and a quantized `output_layer` then fails at save time with the original opaque `RuntimeError`. A `warn_rank_0` on that branch would make the "upstream changed" case diagnosable.

modelopt/torch/export/plugins/mcore_qwen35vl.py
cjluo-nv: > _Bot comment._

Nit: these two rules are now byte-identical to what `qwen3_causal_lm_export` already provides (`"local_experts.linear_fc1": GatedMLPSlicing("model.layers.{}.mlp.experts.{}.")` and the matching `down_proj` `NameRemapping`), and `qwen3_5_vl_causal_lm_export` merges that dict first. Keeping them is harmless but they can drift from the base silently. Either drop them or note that they're restated deliberately so the SequentialMLP and TEGroupedMLP layouts read side by side.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1467,31 +1407,41 @@ def _grouped_mlp_slicing(
...
+                        scales = (weight_scale_cpu, weight_scale_cpu)
+                    else:
+                        scales = (weight_scale_cpu[:half], weight_scale_cpu[half:])
+                    shards = [
+                        (expert_prefix + _gated_subnames[0] + ".", weight[:half], scales[0]),
+                        (expert_prefix + _gated_subnames[1] + ".", weight[half:], scales[1]),
+                    ]
+
+                for shard_prefix, shard_weight, shard_scale in shards:
+                    if shard_scale is None:
```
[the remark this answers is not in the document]
kevalmorabia97: Added in 0f46e2b. You are right that it is not a new assumption, and right that this is the place to make it loud — the grouped path has no reference-shape cross-check, so a mismatch would surface as a silently empty `up_proj.weight_scale` and garbage weights rather than an error.

```python
assert weight_scale_cpu.shape[0] == weight.shape[0], (
    f"cannot split a {tuple(weight_scale_cpu.shape)} weight_scale along "
    f"the output dim of a {tuple(weight.shape)} gated expert weight"
)
```

It holds today for the formats this path sees (NVFP4 `[out, in/16]` and per-output-channel scales both have `shape[0] == out`), so this is purely a guard against a future blockwise layout reaching it.

modelopt/torch/utils/plugins/mbridge.py
```diff
@@ -277,3 +280,52 @@ def load_modelopt_megatron_checkpoint(
...
+
+
+def keep_gpt_output_layer_extra_state() -> bool:
+    """Keep ``output_layer._extra_state`` so a quantized ``lm_head`` can be checkpointed.
```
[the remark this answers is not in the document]
kevalmorabia97: Good catch, and the distinction matters: `mbridge.py` does declare `__all__` (line 46), unlike `mcore_custom.py` in the other thread where I pushed back precisely because it has none. Added `"keep_gpt_output_layer_extra_state"` between `"is_vlm_config"` and `"load_mbridge_model_from_hf"` in 0f46e2b, keeping the list sorted.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1503,13 +1453,17 @@ def _grouped_mlp_slicing(
...
-                self._record_layer_quant_config(
-                    prefix.format(global_id) + ".", seen_qformat, seen_block_size
-                )
+                for sub in _gated_subnames or (None,):
+                    expert_prefix = prefix.format(global_id) + "."
+                    self._record_layer_quant_config(
+                        expert_prefix if sub is None else expert_prefix + sub + ".",
+                        seen_qformat,
+                        seen_block_size,
+                    )
```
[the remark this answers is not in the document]
kevalmorabia97: Confirmed and fixed — this was a real regression I introduced by deleting the packing path. `git show` on the removed `_grouped_mlp_packing` has exactly the branch you quote, and `_grouped_mlp_slicing` only ever called `_record_layer_quant_config`, which returns at line 1130 for `None` / `QUANTIZATION_NONE`. So an unquantized routed expert was named in neither `quantized_layers` nor `exclude_modules`.

Restored inside the per-expert loop with `assert seen_block_size is not None` moved into the quantized arm. Two deltas from your snippet: I dropped the outer `if seen_qformat is not None:` so the `None` case also records an exclusion (matching the packing path, which tested `qformat in (None, QUANTIZATION_NONE)`), and the recorded names have no trailing dot since `_record_excluded_module` strips it.

`test_grouped_mlp_slicing_records_unquantized_experts_as_excluded` asserts all four per-expert prefixes land in `exclude_modules`; it fails on the previous implementation. Three pre-existing grouped-slicing tests also began failing when I made the change, because their stub exporter had no `exclude_modules` attribute — useful confirmation the new branch is actually reached. As you note, this fixes Nemotron too.

tests/gpu_megatron/torch/export/test_unified_export_megatron.py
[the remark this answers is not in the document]
kevalmorabia97: Correct, and thank you for catching it — I fixed one side of the expansion and left the other tripping. `unexpected` filters only on `QUANT_SUFFIXES`, and `...mlp.experts.0.gate_proj.weight` ends in `.weight`, so all three `qwen3_5_moe_vl_*` parametrizations would have failed in CI on a change I had described as "fixing the test".

I took a middle path between your two options. `allow_unexpected` gains `"mlp.experts."` so the comparison passes, and because that waiver is broad enough to hide a genuinely dropped projection, the test now also calls `_assert_per_expert_experts_complete`, which requires every `(layer, expert)` to export all three of `gate_proj` / `up_proj` / `down_proj`. Verified standalone: complete export passes, dropping `down_proj` raises, and a single incomplete expert raises.

That is weaker than full packed-vs-per-expert reconciliation — it checks presence and completeness, not that stacking reproduces the reference tensor — but it restores the property that matters here (a rule emitting only some projections fails) without teaching the shared helper an architecture-specific layout. Happy to do the full reconciliation if you would rather have shape equivalence too.

These are GPU tests so I cannot run them here; the completeness logic is unit-verified and the rest is unchanged from the previously passing configuration plus the two waivers.

modelopt/torch/utils/plugins/mbridge.py
```diff
@@ -277,3 +280,52 @@ def load_modelopt_megatron_checkpoint(
...
+
+def _output_layer_extra_state_has_data(entry: Any) -> bool:
+    """True when a sharded state-dict entry carries a payload."""
+    data = getattr(entry, "data", entry)
+    if isinstance(data, torch.Tensor):
+        return data.numel() > 0
+    return data is not None and bool(data)
+
+
+def keep_gpt_output_layer_extra_state() -> bool:
```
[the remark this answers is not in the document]
kevalmorabia97: Right on both counts — `export_distilled_megatron_to_hf.py` was a fourth entry point I missed, and it is the natural consumer of the `distill.py` save, so the distill -> export path could still have lost a quantized `output_layer`.

I took your structural suggestion rather than adding a fourth call: `keep_gpt_output_layer_extra_state()` now runs inside `load_modelopt_megatron_checkpoint`, which covers all three load sites (`export_quantized_megatron_to_hf.py`, `export_distilled_megatron_to_hf.py`, and `distill.py`'s restore hook), and I removed the now-redundant explicit call from `export_quantized_megatron_to_hf.py`.

`quantize.py` and `distill.py` keep an explicit call, because those are *save* paths and there is no shared ModelOpt save helper to hook — `quantize.py` calls `bridge.save_megatron_model` and `distill.py` goes through Megatron-Bridge's training checkpointing, both upstream. So it is two deliberate call sites on save plus one automatic on load, rather than four that can drift.

That I forgot one of four is a fair argument for your design over mine, and it is why I moved the load side rather than just adding the missing line.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -498,7 +498,16 @@ def _verify_exported_keys(self, save_directory, pretrained_model_name_or_path) -
...
+        # Ancestors too: an export may expand one source module into several (Qwen3.5 packs
+        # routed experts; the quantized export writes them per expert). Expansion is not a drop.
+        exported_modules = set()
+        for key in exported:
+            parts = key.rsplit(".", 1)[0].split(".")
+            for i in range(len(parts), 0, -1):
+                prefix = ".".join(parts[:i])
+                if prefix in exported_modules:
+                    break
+                exported_modules.add(prefix)
```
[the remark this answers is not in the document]
kevalmorabia97: Good analysis. I checked the widening empirically rather than by eye, and it is narrower than it reads — but the gap you describe is real, so I have pinned it explicitly rather than left it implied.

For an ordinary `<module>.weight` source key the ancestor walk changes nothing:

| case | exact-parent | ancestor |
| --- | --- | --- |
| module exported normally | True | True |
| module absent from export | **False** | **False** |
| module expanded into sub-modules | False | True |

A leaf module's parent is an ancestor of an exported path only if something under that module was exported — which is the same condition as the exact match. So the relaxation only bites for *container* parents like `...mlp.experts`, i.e. exactly the expansion it was added for. A whole module with no export rule is still caught.

The residual case is the one you name: `...experts.down_proj` and `...experts.gate_up_proj` share the container, so once any expert tensor is exported both resolve as covered.

Neither suggested option closes it:

- **Option 2 (leaf must appear deeper)** false-positives on a correct export: `gate_up_proj` is *consumed* by the gate/up split, so its leaf name legitimately appears nowhere. Verified — a correct export contains no `gate_up_proj` key at all, so this rule would raise on every Qwen3.5 export.
- **Option 1 (only for bare packed params)** does not help either: `gate_up_proj` and `down_proj` are both bare packed params, so the fallback still applies to both. It would confine the relaxation to packed-parameter sources, but per the table above ordinary keys are unaffected anyway, so it adds a branch for no behavioural change.

Closing it properly needs per-architecture knowledge of which leaves an expansion should produce, which does not belong in a prefix-level check. So I put that coverage where the layout is known instead: `_assert_per_expert_experts_complete` (added for the other review thread) requires every `(layer, expert)` to export all three of `gate_proj` / `up_proj` / `down_proj`, and a dropped `down_proj` fails it.

I added your third case to the parametrization as a documented limit — asserting it does **not** raise, with a comment pointing at where that coverage lives. That pins the intended semantics so a future change to either check is deliberate rather than accidental.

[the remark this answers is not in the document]
kevalmorabia97: Correction to my previous reply: my first attempt at that third case was a literal duplicate of case 0 — ruff caught it as `PT014`, so the commit never landed. Your case needs a *different source* (one carrying both `gate_up_proj` and `down_proj`), not just a different export, which is the distinction I glossed over.

Now added properly in 22a5efe as `test_verify_exported_keys_cannot_see_a_dropped_sibling_under_an_expanded_container`: source has both packed names, export has only `gate_proj` / `up_proj`, and the check does not raise. That is the intended semantics, documented in the docstring with a pointer to `_assert_per_expert_experts_complete`, which does catch it.

10 tests pass in that file now.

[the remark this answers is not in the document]
kevalmorabia97: Commit hash correction: that is `ee9248b`, not `22a5efe` — I quoted the hash before the commit was created.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1467,31 +1484,49 @@ def _grouped_mlp_slicing(
...
-                        else to_quantized_weight(
-                            weight,
-                            weight_scale_cpu,
-                            qformat,
-                            weight_scale_2_cpu,
-                            block_size,
-                        )
+                    assert weight.shape[0] % 2 == 0, (
+                        f"gated expert weight has odd first dim {weight.shape[0]}"
                     )
```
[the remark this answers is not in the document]
kevalmorabia97: Fixed in `27f1b58c` — both gated-split shape checks now raise `ValueError` instead of asserting.

You're right that this one matters beyond style: the odd-row case is reachable from a real
checkpoint (any fused gate+up `linear_fc1` whose first dim isn't even), and under `-O` the silent
outcome is unequal `gate_proj` / `up_proj` halves written to the checkpoint — a corruption that
would only surface as bad generations at serving time.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1467,31 +1407,45 @@ def _grouped_mlp_slicing(
...
-                    local_expert_state[expert_prefix + "weight_scale"] = weight_scale_cpu.clone()
+                    half = weight.shape[0] // 2
+                    if weight_scale_cpu is None or weight_scale_cpu.dim() == 0:
+                        scales = (weight_scale_cpu, weight_scale_cpu)
+                    else:
+                        assert weight_scale_cpu.shape[0] == weight.shape[0], (
+                            f"cannot split a {tuple(weight_scale_cpu.shape)} weight_scale along "
+                            f"the output dim of a {tuple(weight.shape)} gated expert weight"
+                        )
```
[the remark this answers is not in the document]
kevalmorabia97: Same change in `27f1b58c`: the `weight_scale_cpu.shape[0] == weight.shape[0]` guard is a
`ValueError` now too, for the same `-O` reason.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1503,13 +1453,17 @@ def _grouped_mlp_slicing(
         # Record quant config for ALL global experts on every rank; otherwise the writer's
         # hf_quant_config.json would miss (EP-1)/EP of the routed experts. All experts in
         # a TEGroupedLinear layer share qformat/block_size, so local values apply globally.
-        if seen_qformat is not None and record_quant_config:
+        if seen_qformat is not None:
```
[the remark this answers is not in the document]
kevalmorabia97: You're right, and the comment I was preserving was wrong — I verified it rather than trusting it.

`_gather_layer_config_dict` does `all_gather_object` + `dict.update` across the world, and
`_gather_exclude_modules` does the same for `exclude_modules`. So every rank's records are merged
before `hf_quant_config.json` is written, and recording `range(num_total_experts)` on every rank
was redundant rather than necessary.

Changed to `for global_id in local_expert_indices` in `27f1b58c`. That also fixes a case the old
loop got wrong: with non-contiguous EP assignment, `range(num_experts * ep_size)` names experts
this rank never owned, and a rank with no local experts recorded the whole layer from
`seen_qformat = None` — i.e. marked every expert excluded.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1503,13 +1457,19 @@ def _grouped_mlp_slicing(
...
         # a TEGroupedLinear layer share qformat/block_size, so local values apply globally.
-        if seen_qformat is not None and record_quant_config:
-            assert seen_block_size is not None
-            num_total_experts = num_experts * ep_size
-            for global_id in range(num_total_experts):
-                self._record_layer_quant_config(
-                    prefix.format(global_id) + ".", seen_qformat, seen_block_size
-                )
+        num_total_experts = num_experts * ep_size
+        for global_id in range(num_total_experts):
```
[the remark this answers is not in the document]
kevalmorabia97: You're right, and the comment I was preserving was wrong — I verified it rather than trusting it.

`_gather_layer_config_dict` does `all_gather_object` + `dict.update` across the world, and
`_gather_exclude_modules` does the same for `exclude_modules`. So every rank's records are merged
before `hf_quant_config.json` is written, and recording `range(num_total_experts)` on every rank
was redundant rather than necessary.

Changed to `for global_id in local_expert_indices` in `27f1b58c`. That also fixes a case the old
loop got wrong: with non-contiguous EP assignment, `range(num_experts * ep_size)` names experts
this rank never owned, and a rank with no local experts recorded the whole layer from
`seen_qformat = None` — i.e. marked every expert excluded.

modelopt/torch/utils/plugins/mbridge.py
kevalmorabia97: Moved in `27f1b58c`, and your placement instinct was right for a stronger reason than the one in
the thread: **`mbridge.py` imports `megatron.bridge`**, which Megatron-LM and NeMo users don't
have. Anything living there could never have reached them, no matter how many callers remembered
to call it.

It now runs from `megatron_replace_quant_module_hook`, which is registered in
`CUSTOM_MODEL_PLUGINS` and so runs for every Megatron model, on both the save and the load path
(the hook fires during modelopt-state restore, before the sharded load plan is built). The three
example `main()` calls and the `load_modelopt_megatron_checkpoint` call are gone —
`mbridge.py`, `quantize.py` and `distill.py` are back to exactly their `main` state. It also now
sits next to `quant_module_get_extra_state`, which already documents the same upstream assert.

On the rest:

- **`warn_rank_0` on fingerprint mismatch** — added, worded to cover both branches ("may already
  keep the entry"), since a mismatch means either upstream fixed it or upstream changed shape, and
  we can't tell which. `@cache` keeps it to once per process.
- **Self-disabling** — now tracked by identity against the function we installed, rather than an
  attribute stamped on it (mypy rejected the attribute).
- **Tests** — `TestKeepGptOutputLayerExtraState`, 15 cases: the payload check across tensor / bytes
  / `ShardedObject`-shaped inputs, the no-op second call, and the warn-and-skip path.

One thing worth flagging from writing those tests. My first version asserted "the patch applies to
the installed megatron-core" and it **failed** — my local Megatron-LM checkout is the #7086 branch,
so the fingerprint correctly reported "already fixed" and declined to patch. That's the guard
working, but it made the test environment-dependent. It's now two tests instead: a behavioural one
that passes whether or not the patch was needed (and fails loudly if *neither* we nor upstream keep
the entry), and `test_patches_stock_megatron_core`, which installs a replica of the real pre-fix
upstream body and proves the patched path. I checked that replica against `be08ce5b1~1` in
Megatron-LM — statement shape `Assign, Assign, Assign, Assert, Return`, matching the fingerprint.

Finally, on the TODO: **Megatron-LM #7086 is now closed.** nemo:26.10 migrates `GPTModel` to
`HybridModel`, whose `sharded_state_dict` has no pop-and-assert, so the bug retires with the class
rather than needing a fix to a deprecated one. That makes this the permanent home for the interim
containers, which is another argument for it not living in three example scripts. The TODO now
points at the 26.10 migration instead of a container patch release.

modelopt/torch/export/plugins/mcore_qwen35vl.py
kevalmorabia97: Dropped in `27f1b58c`. I confirmed the duplication programmatically before removing them, and
re-checked afterwards that `local_experts.linear_fc1` / `linear_fc2` still resolve to the same rule
objects through `{**qwen3_causal_lm_export, **_qwen3_5_extra_export}`.

The comment above the remaining grouped rules now says where `local_experts.*` comes from, so the
next reader doesn't have to diff the two mappings to find out.

modelopt/torch/utils/plugins/mbridge.py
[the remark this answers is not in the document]
kevalmorabia97: Handled more broadly than the version you reviewed. I'd put the call inside
`load_modelopt_megatron_checkpoint`; in `27f1b58c` it moved to
`megatron_replace_quant_module_hook` (`CUSTOM_MODEL_PLUGINS`), which runs during modelopt-state
restore — before the sharded load plan is built — for every Megatron model.

So the answer to "one entry point is covered, what about the others" is now "there is no per-entry-
point call to miss": `mbridge.py` is back to its `main` state and the coverage comes from the
conversion hook instead.

modelopt/torch/utils/plugins/mbridge.py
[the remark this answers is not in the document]
kevalmorabia97: Superseded by `27f1b58c`: the function moved out of `mbridge.py` into
`modelopt/torch/quantization/plugins/megatron.py`, so the `__all__` entry I added here went with
it. That module declares `__all__ = []` — it's plugin machinery wired up through
`CUSTOM_MODEL_PLUGINS`, not public API, and no example imports it any more, which was the
"imported by three example scripts" premise for making it public in the first place.

cjluo-nv (commented):

Re-review of #2334 (9 files, +478/-140). All previously flagged blockers now look resolved; what remains is owner sign-off rather than new defects.

**Previous comments — status**
- *Critical: `qwen3_5_moe_vl_{grouped,sequential}` GPU tests would fail on the `unexpected` side* — **addressed**. `allow_unexpected` now gains `"mlp.experts."` alongside the widened `allow_missing`, and `_assert_per_expert_experts_complete` re-adds a positive check that every `(layer, expert)` emits all three projections.
- *Critical: gate/up split + `_verify_exported_keys` relaxation untested* — **addressed** (CPU tests over `_FakeTEGroupedMLP` cover the split, per-block scale slicing at `half`, `weight_scale_2` replication, both `_record_layer_quant_config` prefixes, the 0-dim fallback, and the parametrized expansion/drop pair for the self-check).
- *Critical: `exclude_modules` bookkeeping lost when `_grouped_mlp_packing` was deleted* — **addressed**; the per-expert loop now records exclusions for `qformat in (None, QUANTIZATION_NONE)` with a dedicated test, and this also fixes the pre-existing Nemotron gap.
- *Critical (design gate): placement of `keep_gpt_output_layer_extra_state`* — **addressed**; moved out of `utils/plugins/mbridge.py` into `modelopt/torch/quantization/plugins/megatron.py` next to `quant_module_get_extra_state`, invoked from `megatron_replace_quant_module_hook` (`CUSTOM_MODEL_PLUGINS`), the three example `main()` call sites reverted, and 15 tests added. The PR body now explains why `mbridge.py` was the wrong home (it imports `megatron.bridge`, unreachable for Megatron-LM/NeMo users).
- *Minor: dead `GroupedMLPPacking` / `quantize=` / `record_quant_config=` / stale `transpose=False` comment* — removed. *Minor: `__all__` for `GroupedGatedMLPSlicing`* — declined with a reasonable rationale. *Minor: asserts → `ValueError`, `local_expert_indices` instead of a global range, duplicated SequentialMLP rules* — all applied.

I re-verified the export path end to end: `GroupedGatedMLPSlicing("...experts.{{}}")` → `prefix.format(layer).format(expert)` resolves to `model.language_model.layers.L.mlp.experts.E.gate_proj.`; `with_language_model_prefix`'s `type(m)(...)` preserves the `gate_proj_name`/`up_proj_name` defaults; `_get_weight_scales` pops the scales so the trailing replicate loop cannot clobber the sliced ones; the ancestor loop in `_verify_exported_keys` is idempotent and short-circuits correctly.

**Why nudge rather than approve**
- :speech_balloon: Author replied on the GPU-test fix: *"These are GPU tests so I cannot run them here; the completeness logic is unit-verified."* — still worth a human eye because `allow_unexpected=("mlp.experts.",)` waives the routed experts from both key checks, so the per-expert tensors never enter `shared` and the helper's shape cross-check no longer sees them. Coverage for the new layout in the only end-to-end test is now presence/completeness only, and CI green on `qwen3_5_moe_vl_grouped` (NVFP4 + FP8) and `_sequential` hasn't been observed yet.
- :speech_balloon: Author replied on the `lm_head` placement: *"moved to `megatron_replace_quant_module_hook` so no caller can forget it; `mbridge.py` imports `megatron.bridge` and could never reach Megatron-LM/NeMo."* — reasonable, and the tests are good. Flagging anyway because the mechanism is a permanent, process-global monkeypatch of `megatron.core` `GPTModel.sharded_state_dict`, gated on an AST statement-kind fingerprint, installed as a side effect of every `mtq.quantize` on any Megatron model (including HybridModel/teacher models). The upstream PR is closed, so this is the permanent home, not a stopgap — that's a maintenance commitment a human owner should sign off on.
- Licensing signal: `tests/gpu_megatron/torch/quantization/plugins/test_megatron.py` adds `_stock_gpt_sharded_state_dict`, described in the PR body as "a replica of the real pre-fix upstream body (verified against `be08ce5b1~1` in Megatron-LM)". It's a handful of lines from a sibling Apache-2.0 NVIDIA project in test code, so likely fine, but per policy a copied-from-external-repo block shouldn't be auto-approved — a short attribution comment naming the source commit would settle it.

**Non-blocking nit**: `keep_gpt_output_layer_extra_state` recognises its own work only by identity against the module-global `_patched_gpt_sharded_state_dict`. In the new test fixture (which restores `GPTModel.sharded_state_dict` to a *previously* installed replacement while the global points at a newer closure), a later `cache_clear()` + call will fingerprint our own replacement (`[Assign, Assign, If, Return]`), miss, and emit the misleading "not the version ModelOpt patches" warning. Unreachable in production because of `@cache`, but it makes the warning less trustworthy in test logs.

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -325,6 +330,65 @@ def _output_layer_untied(config) -> bool:
...
+    TODO: remove once megatron-core migrates GPTModel to HybridModel, expected in nemo:26.10.
+    """
+    global _patched_gpt_sharded_state_dict
+    if GPTModel.sharded_state_dict is _patched_gpt_sharded_state_dict:
+        return True
+    try:
+        src = textwrap.dedent(inspect.getsource(GPTModel.sharded_state_dict))
+    except (OSError, TypeError):
+        src = ""  # no source to check against; leave megatron-core alone
+    func = ast.parse(src).body[0] if src else None
```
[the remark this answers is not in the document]
kevalmorabia97: Good catch, and the blast radius is the part that makes it worth fixing: this runs from
`megatron_replace_quant_module_hook`, so a `SyntaxError` here would break *all* Megatron
quantization, not just the `lm_head` case it is trying to protect.

Fixed in `a21b4da108` with your snippet, plus `IndexError` — `ast.parse` on a source that is
syntactically valid but contains no statements (a file trimmed to comments) makes `.body[0]` raise
rather than return an empty `FunctionDef`.

Verified the degradation rather than assuming it: with `inspect.getsource` patched to return
`"def f(:\n  pass"`, the call now returns `False` and warns instead of propagating. That path also
turned up naturally while testing — a function defined in a stdin heredoc has no retrievable
source, so `getsource` raises `OSError` and the same guard catches it.

modelopt/torch/quantization/plugins/megatron.py
```diff
@@ -325,6 +330,65 @@ def _output_layer_untied(config) -> bool:
...
+    """Keep ``output_layer._extra_state`` so a quantized ``lm_head`` can be checkpointed.
+
+    ``GPTModel.sharded_state_dict`` drops that entry and asserts it is empty, so a quantized
+    output_layer otherwise fails to save and loads back unquantized. Cached: warns at most once.
+
+    TODO: remove once megatron-core migrates GPTModel to HybridModel, expected in nemo:26.10.
+    """
+    global _patched_gpt_sharded_state_dict
+    if GPTModel.sharded_state_dict is _patched_gpt_sharded_state_dict:
+        return True
```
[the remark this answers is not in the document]
kevalmorabia97: You are right, and I introduced that redundancy mid-review: the guard started as an attribute on
the patched function, mypy rejected it (`attr-defined`), and I converted it to an identity check
against a module global without re-examining whether `@cache` had already made it dead.

Took your second option in `a21b4da108` — kept `@cache`, dropped the global and the guard, and
made the docstring say which mechanism is doing the work:

> ``@cache`` makes this idempotent -- the body runs once per process, so a repeat call cannot
> double-patch or re-warn.

`test_second_call_is_a_no_op` now asserts idempotence through the cache (`cache_info().hits >= 1`
plus the function object being unchanged) instead of calling `cache_clear()`, which was the only
thing that ever reached the guard. Confirmed against a stock-mcore replica: patched exactly once,
second call is a cache hit, `GPTModel.sharded_state_dict` unchanged.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1500,16 +1455,19 @@ def _grouped_mlp_slicing(
...
+            for sub in _gated_subnames or (None,):
+                expert_prefix = prefix.format(global_id) + "."
+                if sub is not None:
+                    expert_prefix += sub + "."
+                # exclude_modules is an explicit list, so unquantized experts must be named there
+                # or a mixed-precision checkpoint gives the runtime no signal for them.
+                if seen_qformat in (None, QUANTIZATION_NONE):
+                    self._record_excluded_module(expert_prefix)
+                else:
+                    self._record_layer_quant_config(expert_prefix, seen_qformat, seen_block_size)
```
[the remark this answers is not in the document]
kevalmorabia97: Both correct; both done in `a21b4da108`.

**1. Dead return.** Verified before removing: `_gated_mlp_slicing` returns nothing, every
`self.rules[...](...)` call site is a bare statement, and the rule-book lambda's value is discarded.
So `_grouped_mlp_slicing` was the only handler returning a tuple nobody read. Dropped.

**2. NemotronH.** Also verified — `mcore_nemotron.py` maps `experts.linear_fc{1,2}` to
`GroupedMLPSlicing`, so it takes exactly the non-gated path where this loop is new behaviour. Added
a changelog entry naming it, since it changes `hf_quant_config.json` for a shipped architecture:

> Megatron-Core HuggingFace export now names unquantized grouped-GEMM MoE experts in
> ``exclude_modules`` instead of omitting them, so a mixed-precision checkpoint tells the runtime
> what is not quantized. This adds one ``hf_quant_config.json`` entry per expert per layer for
> grouped-expert architectures such as ``NemotronHForCausalLM``; the served numerics are unchanged.

ChenhanYu (approved):

modelopt/torch/utils/vlm_dataset_utils.py
```diff
@@ -81,7 +81,14 @@ def __init__(self, base, rank: int, world: int):
         self._world = world
 
     def __iter__(self):
-        return itertools.islice(iter(self._base), self._rank, None, self._world)
+        shard = itertools.islice(iter(self._base), self._rank, None, self._world)
+        # Truncate to floor(len / world) so every rank yields the same count: one rank short
+        # leaves the others blocked forever on the next collective.
+        try:
+            per_rank = len(self._base) // self._world
```
cjluo-nv: > _Bot comment._

This guard does not fire for the case §5 describes. `_get_vlm_dataset` returns `NemotronTarPlusJsonlIterable` directly for `nemotron_vlm_dataset_v2`, and that class defines no `__len__` — so `len(self._base)` raises `TypeError` and the untruncated `shard` is returned. On the reported EP=4 / 1024-sample run only the `divmod` budget fix applies.

Even on the `_HFDatasetsIterableWrapper` path, `__len__` returns the *requested* `num_samples`, not what the stream actually yields, so `islice(shard, per_rank)` is a no-op whenever the stream comes up short — which is precisely the failure mode (1023 delivered for 1024) the section is fixing. The subset budgets can still fall short in practice: `if not shard_list: continue`, `if not meta_by_image: continue`, and per-image decode failures all skip a sample without replacement.

Also `per_rank` is 0 when `num_samples < world`, so every rank yields nothing and calibration silently runs no forward passes.

Suggest either giving the Nemotron iterable a real `__len__` and making it pad/refill to its target, or making the sharder resilient to a short stream (e.g. a count `all_reduce`/min before the forward loop) rather than relying on a declared length.

tests/unit/torch/utils/test_dataset_utils.py
```diff
@@ -918,3 +918,49 @@ def test_dataloader_mixing_dir_and_local_jsonl(
...
+        from modelopt.torch.utils.vlm_dataset_utils import (
+            _HFDatasetsIterableWrapper,
+            _ShardedIterable,
+        )
+
+        base = _HFDatasetsIterableWrapper(list(range(20)), num_samples=20)
+        shards = [list(_ShardedIterable(base, rank=r, world=4)) for r in range(4)]
+        assert shards[0][:3] == [0, 4, 8]
+        flat = [x for s in shards for x in s]
+        assert len(flat) == len(set(flat)), "shards overlap"
```
cjluo-nv, line 963: > _Bot comment._

This test re-implements the production expression (`divmod` + `max(1, ...)`) inside the test body and asserts the re-implementation sums correctly. It never imports `NemotronTarPlusJsonlIterable`, so reverting the fix in `nemotron_vlm_dataset_utils.py` leaves it green — it can't catch the regression it's named for.

Please assert against the real object, e.g. construct the iterable with fake shard/JSONL fixtures (or factor the budget computation into a small module-level helper and call it) and check the per-subset targets sum to `num_samples`. Worth covering `num_samples < len(subsets)` too: `max(1, ...)` makes the targets sum to more than `num_samples` there, which contradicts the comment that says they "sum exactly".

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
```diff
@@ -1925,3 +1929,105 @@ def __init__(self, config, shared):
 
     assert student.config.modelopt_output_layer_untied is True
     assert not hasattr(student._teacher_model.config, "modelopt_output_layer_untied")
+
+
+# Captured at import, before any test can apply the patch, so tests can start from pristine mcore.
+_PRISTINE_GPT_SHARDED_STATE_DICT = GPTModel.sharded_state_dict
+
+
```
cjluo-nv, line 1937: > _Bot comment._

Still open from the previous round: this is a verbatim replica of megatron-core's pre-fix `GPTModel.sharded_state_dict` body, and the PR body says it was verified against `be08ce5b1~1` in Megatron-LM, but the file carries no attribution. Please add a one-line comment naming the upstream repo and commit (and its Apache-2.0 license) so the copied block is traceable — copied third-party code is a call a human reviewer has to sign off on.

cjluo-nv (commented):

Commenting: the export/`lm_head` half looks settled, but the newly added VLM-calibration fix (§5) does not cover the dataset that produced the reported hang, and its test asserts nothing about production code.

Needs action:
- [ ] Fix the `_ShardedIterable` guard in `modelopt/torch/utils/vlm_dataset_utils.py`: `NemotronTarPlusJsonlIterable` has no `__len__`, so the truncation is skipped for exactly the dataset in §5; and for the wrapper it uses the declared `num_samples`, not the actual yield count. See inline.
- [ ] Rewrite `test_nemotron_subset_budget_sums_to_num_samples` to drive `NemotronTarPlusJsonlIterable`; today it re-implements the `divmod` expression inline and passes even if the fix is reverted.
- [ ] Add an attribution comment naming the Megatron-LM source commit above `_stock_gpt_sharded_state_dict` in `tests/gpu_megatron/torch/quantization/plugins/test_megatron.py` (raised last round, still open; licensing needs a human call).
- [ ] Split §4–§6 into a separate PR as you offered — they touch unrelated subsystems and arrived after §1–3 had already been reviewed four times.

No action needed:
- :heavy_check_mark: Resolved since the last review: the GPU `unexpected`-key regression, the `exclude_modules` gap, the `keep_gpt_output_layer_extra_state` placement, the `ast.parse` guard and the dead `_grouped_mlp_slicing` return.

modelopt/torch/utils/vlm_dataset_utils.py
kevalmorabia97: Confirmed on all three points, thanks — the guard never fired for the case §5 describes.

`NemotronTarPlusJsonlIterable` defines no `__len__`, so `len(self._base)` raised `TypeError` and the untruncated shard was returned. On the `_HFDatasetsIterableWrapper` path `__len__` returns the requested `num_samples` rather than what the stream delivers, so `islice(shard, per_rank)` was a no-op exactly when the stream came up short. And `per_rank` was 0 whenever `num_samples < world`, so every rank yielded nothing.

Fixed in 3d9b28744b by taking your second suggestion — the sharder no longer consults `__len__` at all. `_ShardedIterable` now takes the expected per-rank count explicitly, truncates if the stride runs long, and repeats the last sample (with a warning) if it runs short, so ranks stay in step through the `if not shard_list: continue` / `if not meta_by_image: continue` / decode-failure paths you list. Fewer samples than ranks is now rejected outright rather than silently producing empty ranks.

I went with padding over a count `all_reduce`/`min` because the shortfall has to be known before the first yield to keep the iteration lazy; a collective would need either a second pass or materialising each rank's shard. Happy to switch if you would rather have the exact-count semantics.

New test `test_short_stream_still_yields_equal_counts` covers a 57-of-64 delivery over 4 ranks, and `test_stream_shorter_than_world_is_rejected` covers the empty case.

tests/unit/torch/utils/test_dataset_utils.py
kevalmorabia97: You are right, and I verified it: the old test stayed green with the production fix reverted, so it could not catch the regression it was named for.

Fixed in 3d9b28744b by taking your parenthetical suggestion — the budget computation is now a module-level `subset_sample_targets(num_samples, subsets)` in `nemotron_vlm_dataset_utils.py`, and the test imports and calls it instead of re-implementing `divmod` in its body. Reverting the helper to the old truncating division now fails 5 of the 6 parametrised cases.

Also fixed the `max(1, ...)` overshoot you flagged: with `num_samples < len(subsets)` the targets summed to more than `num_samples`, contradicting the comment. Dropping `max(1, ...)` makes the sum exact (2 over 3 subsets is now `[1, 1, 0]`), and `(2, 3)` and `(0, 3)` are parametrised cases.

tests/gpu_megatron/torch/quantization/plugins/test_megatron.py
kevalmorabia97: Added in 3d9b28744b — the docstring now names the source and license:

> Copied verbatim from `GPTModel.sharded_state_dict` in NVIDIA/Megatron-LM at `be08ce5b1~1` (Apache-2.0) so the patched path is exercised whichever megatron-core is installed.

cjluo-nv (commented):

Nudge: the three code concerns from the last round are fixed, but one new correctness question and the licensing/scope calls still need a human.

Needs action:
- [ ] Confirm `convert_to_distillation_provider` propagates `calculate_per_token_loss` onto `distill_provider` in `examples/megatron_bridge/distill.py`; if it doesn't, `average_in_collective` silently regresses `--sft`. Deriving it from `args.sft or args.cp_size > 1` avoids the dependency.
- [x] Sign off on the verbatim Apache-2.0 Megatron-LM copy (`_stock_gpt_sharded_state_dict` in `tests/gpu_megatron/torch/quantization/plugins/test_megatron.py`) — attribution is now present, but the legal call is a human's.
- [ ] Split §4–§6 out as you offered, or say why they ship together — they touch unrelated subsystems and landed after §1–3 had been reviewed several rounds.
- [ ] Use `count == 0` rather than `last is None` for the empty-stream check in `_ShardedIterable` (`modelopt/torch/utils/vlm_dataset_utils.py`), so a sample that is legitimately `None` isn't read as an empty stream.

No action needed:
- :heavy_check_mark: Resolved since the last review: the `_ShardedIterable` `__len__`/`num_samples` guard, the budget test now calling `subset_sample_targets`, and the Megatron-LM source attribution.

modelopt/torch/utils/vlm_dataset_utils.py
kevalmorabia97: Both follow-ups addressed in 96808aa7d0.

**`count == 0` instead of `last is None`** — agreed and fixed; a legitimately `None` sample would have been read as an empty stream.

**`calculate_per_token_loss` propagation** — I checked, and it does propagate today: `convert_to_distillation_provider` reclasses the student provider in place (`student_provider.__class__ = DistillationProvider`) and returns that same object, so `distill_provider is student_provider` and the attribute survives. But your underlying point stands — that correctness rests on a non-obvious in-place contract in another repo. Rather than re-derive the condition in two places (which is how the original `--sft` bug arose), it is now computed once into a local in `main()` and used for both the provider and the DDP config, so neither can drift and neither depends on the aliasing.

cjluo-nv (commented):

Nudge: both code concerns from the last round are fixed, but the licensing sign-off and the scope call still need a human.

Needs action:
- [ ] Sign off on the verbatim Apache-2.0 Megatron-LM copy (`_stock_gpt_sharded_state_dict` in `tests/gpu_megatron/torch/quantization/plugins/test_megatron.py`) — attribution is now present, but the legal call is a human's.
- [ ] Split §4–§6 out as you offered, or state in the PR body why they ship together — they touch unrelated subsystems and landed after §1–3 had been reviewed several rounds.
- [ ] Confirm the `qwen3_5_moe_vl_grouped` / `_sequential` GPU cases pass in CI: `allow_unexpected=("mlp.experts.",)` waives the routed experts from both key checks, so `assert_per_expert_experts_complete` is now the only coverage of the new layout and it has not been seen green.
- [ ] Consider a rank-local warning (not `warn_rank_0`) for the pad-repeat path in `_ShardedIterable` (`modelopt/torch/utils/vlm_dataset_utils.py`) — a shortfall on a non-zero rank is currently silent.

No action needed:
- :heavy_check_mark: Resolved since the last review: `calculate_per_token_loss` is now derived once into `per_token_loss` in `distill.py` and reused for the DDP config, and the empty-stream guard uses `count == 0`.

PR #2339, opened 2026-09-04
[Fix] Calibrate non-decoder modules during layerwise quantization

realAsma:
### What does this PR do?

Type of change: Bug fix

Layerwise calibration now calibrates enabled quantizers outside transformer layers, such as `lm_head`, while hiding decoder layers from the additional calibration traversal. It also fails early when these quantizers are combined with progressive layerwise export, whose in-place conversion makes the required model calibration unsafe.

### Usage

No API changes.

### Testing

- `pytest_pwd tests/unit/torch/quantization/test_calib.py tests/unit/torch/quantization/test_layerwise_calibrate.py -q` — 68 passed
- Pre-commit on all six changed files — passed
- `git diff --check` — passed

## Summary by CodeRabbit

- **New Features**
  - Layerwise calibration now supports quantizers outside transformer decoder layers, including LM-head activation quantizers.
  - Calibration supports models with CPU-offloaded components while preserving model behavior.
  - The MSE calibration utility is now publicly available.
  - Added warnings for calibration passes involving offloaded components.

- **Bug Fixes**
  - Prevented unsupported export-mode calibration when outside-layer quantizers are present.
  - Ensured outside-layer calibration runs after transformer-layer restoration.
  - Preserved original forwards while hiding decoder subtrees from traversal and state collection.
  - Ensured cleanup and alias restoration after errors or completion.

### DIFF
--- a/modelopt/torch/quantization/config.py
+++ b/modelopt/torch/quantization/config.py
@@ -928,8 +928,8 @@ class MaxCalibConfig(_SharedStatesConfig, QuantizeAlgorithmConfig):
         ),
     )
 
-    skip_forward_without_activation_calib: bool | None = ModeloptField(
-        default=None,
+    skip_forward_without_activation_calib: bool = ModeloptField(
+        default=False,
         title="Skip the calibration forward when no activation quantizer needs data.",
         description=(
             "If True, max calibration skips the ``forward_loop`` entirely when no enabled "
@@ -938,10 +938,10 @@ class MaxCalibConfig(_SharedStatesConfig, QuantizeAlgorithmConfig):
             "dynamic, or MX (MXFP4/MXFP8) quantization. Weight calibration still runs on the "
             "weight tensors directly, so the quantized weights are unchanged; only the wasted "
             "forward is avoided. "
-            "Set False when the provided ``forward_loop`` carries side effects the caller relies "
-            "on — most notably materializing sharded parameters under DeepSpeed ZeRO-3. The "
-            "default None behaves as False for whole-model and per-layer calibration, and as True "
-            "for ModelOpt's generated layerwise non-decoder pass."
+            "Opt-in (default False) because the provided ``forward_loop`` can carry side "
+            "effects the caller relies on — most notably materializing sharded parameters under "
+            "DeepSpeed ZeRO-3 — so enable it per-recipe when the calibration data is known to be "
+            "unnecessary."
         ),
     )
 
--- a/modelopt/torch/quantization/config.py
+++ b/modelopt/torch/quantization/config.py
@@ -928,8 +928,8 @@ class MaxCalibConfig(_SharedStatesConfig, QuantizeAlgorithmConfig):
         ),
     )
 
-    skip_forward_without_activation_calib: bool = ModeloptField(
-        default=False,
+    skip_forward_without_activation_calib: bool | None = ModeloptField(
+        default=None,
         title="Skip the calibration forward when no activation quantizer needs data.",
         description=(
             "If True, max calibration skips the ``forward_loop`` entirely when no enabled "
@@ -938,10 +938,10 @@ class MaxCalibConfig(_SharedStatesConfig, QuantizeAlgorithmConfig):
             "dynamic, or MX (MXFP4/MXFP8) quantization. Weight calibration still runs on the "
             "weight tensors directly, so the quantized weights are unchanged; only the wasted "
             "forward is avoided. "
-            "Opt-in (default False) because the provided ``forward_loop`` can carry side "
-            "effects the caller relies on — most notably materializing sharded parameters under "
-            "DeepSpeed ZeRO-3 — so enable it per-recipe when the calibration data is known to be "
-            "unnecessary."
+            "Set False when the provided ``forward_loop`` carries side effects the caller relies "
+            "on — most notably materializing sharded parameters under DeepSpeed ZeRO-3. The "
+            "default None behaves as False for whole-model and per-layer calibration, and as True "
+            "for ModelOpt's generated layerwise non-decoder pass."
         ),
     )
 
--- a/modelopt/torch/quantization/mode.py
+++ b/modelopt/torch/quantization/mode.py
@@ -226,7 +226,7 @@ def wrapped_calib_func(
     So lets wrap them to be compatible with the ModelOpt convert entrypoint.
     """
     kwargs = config.model_dump()
-    if kwargs.get("skip_forward_without_activation_calib") is None:
+    if "skip_forward_without_activation_calib" not in config.model_fields_set:
         kwargs.pop("skip_forward_without_activation_calib", None)
     method = kwargs.pop("method")
     layerwise_cfg = kwargs.pop("layerwise", None) or {}
--- a/modelopt/torch/quantization/mode.py
+++ b/modelopt/torch/quantization/mode.py
@@ -226,7 +226,7 @@ def wrapped_calib_func(
     So lets wrap them to be compatible with the ModelOpt convert entrypoint.
     """
     kwargs = config.model_dump()
-    if "skip_forward_without_activation_calib" not in config.model_fields_set:
+    if kwargs.get("skip_forward_without_activation_calib") is None:
         kwargs.pop("skip_forward_without_activation_calib", None)
     method = kwargs.pop("method")
     layerwise_cfg = kwargs.pop("layerwise", None) or {}
--- a/modelopt/torch/quantization/mode.py
+++ b/modelopt/torch/quantization/mode.py
@@ -226,6 +226,8 @@ def wrapped_calib_func(
     So lets wrap them to be compatible with the ModelOpt convert entrypoint.
     """
     kwargs = config.model_dump()
+    if "skip_forward_without_activation_calib" not in config.model_fields_set:
+        kwargs.pop("skip_forward_without_activation_calib", None)
     method = kwargs.pop("method")
     layerwise_cfg = kwargs.pop("layerwise", None) or {}
     layerwise = layerwise_cfg.get("enable", False)
--- a/modelopt/torch/quantization/mode.py
+++ b/modelopt/torch/quantization/mode.py
@@ -226,8 +226,6 @@ def wrapped_calib_func(
     So lets wrap them to be compatible with the ModelOpt convert entrypoint.
     """
     kwargs = config.model_dump()
-    if "skip_forward_without_activation_calib" not in config.model_fields_set:
-        kwargs.pop("skip_forward_without_activation_calib", None)
     method = kwargs.pop("method")
     layerwise_cfg = kwargs.pop("layerwise", None) or {}
     layerwise = layerwise_cfg.get("enable", False)
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -62,6 +62,7 @@ from .utils import (
     enable_fake_quant,
     enable_quant,
     enable_weight_access_and_writeback,
+    has_accelerate_offload,
     is_quantized_column_parallel_linear,
     is_quantized_linear,
     is_quantized_row_parallel_linear,
@@ -2094,11 +2095,48 @@ def layerwise_calibrate(
         for module in model.modules()
     )
 
-    if export_dir is not None and has_enabled_outside_quantizer:
+    outside_calib_needs_forward = has_enabled_outside_quantizer
+    outside_calib_runs_forward = has_enabled_outside_quantizer
+    outside_calib_kwargs = calib_kwargs
+    if has_enabled_outside_quantizer and calib_func is max_calibrate:
+        with _hide_modules_from_traversal(model, transformer_layers):
+            outside_calib_needs_forward = _needs_activation_forward_for_max_calib(model)
+        outside_calib_kwargs = dict(calib_kwargs)
+        outside_calib_kwargs.setdefault("skip_forward_without_activation_calib", True)
+        outside_calib_runs_forward = (
+            outside_calib_needs_forward
+            or not outside_calib_kwargs["skip_forward_without_activation_calib"]
+        )
+
+    if export_dir is not None and outside_calib_runs_forward:
         raise ValueError(
             "Layerwise export does not support enabled quantizers outside transformer layers. "
             "Calibrate without export_dir, then export the completed model separately."
         )
+
+    def _calibrate_outside_quantizers():
+        if not has_enabled_outside_quantizer:
+            return
+
+        if outside_calib_runs_forward and has_accelerate_offload(model):
+            warn_rank_0(
+                "Layerwise calibration found enabled quantizers outside transformer layers. "
+                "The required full-model calibration pass may be slow because CPU- or "
+                "disk-offloaded decoder weights can be transferred for every batch."
+            )
+
+        with _hide_modules_from_traversal(model, transformer_layers):
+            if qdq_from_prev:
+                calib_func(model, forward_loop, **outside_calib_kwargs)
+            else:
+                with ExitStack() as stack:
+                    for layer in transformer_layers:
+                        stack.enter_context(
+                            set_quantizer_by_cfg_context(
+                                layer, [{"quantizer_name": "*", "enable": False}]
+                            )
+                        )
+                    calib_func(model, forward_loop, **outside_calib_kwargs)
 
     num_layers = len(transformer_layers)
     print_rank_0(f"Layerwise calibration: Found {num_layers} transformer layers")
@@ -2130,6 +2168,7 @@ def layerwise_calibrate(
     if exporter is not None and _reconcile_export_with_resume(
         exporter, checkpoint_dir, start_layer, num_layers
     ):
+        _calibrate_outside_quantizers()
         warn_rank_0(
             f"Layerwise export: every layer shard in {exporter.export_dir} is already "
             f"written.{finalize_hint}"
@@ -2231,26 +2270,7 @@ def layerwise_calibrate(
     if ckpt:
         ckpt.full_restore(transformer_layers, model)
 
-    if has_enabled_outside_quantizer:
-        if any(device == "disk" for device in getattr(model, "hf_device_map", {}).values()):
-            warn_rank_0(
-                "Layerwise calibration found enabled quantizers outside transformer layers. "
-                "The required full-model calibration pass may be slow because disk-offloaded "
-                "decoder weights can be streamed for every batch."
-            )
-
-        with _hide_modules_from_traversal(model, transformer_layers):
-            if qdq_from_prev:
-                calib_func(model, forward_loop, **calib_kwargs)
-            else:
-                with ExitStack() as stack:
-                    for layer in transformer_layers:
-                        stack.enter_context(
-                            set_quantizer_by_cfg_context(
-                                layer, [{"quantizer_name": "*", "enable": False}]
-                            )
-                        )
-                    calib_func(model, forward_loop, **calib_kwargs)
+    _calibrate_outside_quantizers()
 
     if exporter is not None:
         warn_rank_0(
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -20,6 +20,7 @@ import math
 import time
 import warnings
 from collections.abc import Callable, Mapping, Sequence
+from contextlib import ExitStack
 from functools import partial
 from typing import Any, TypeAlias
 
@@ -34,6 +35,7 @@ from modelopt.torch.opt.searcher import ForwardLoop
 from modelopt.torch.quantization.utils.layerwise_calib import (
     LayerActivationCollector,
     _CheckpointState,
+    _hide_modules_from_traversal,
     _reconcile_export_with_resume,
 )
 from modelopt.torch.utils import print_rank_0, warn_rank_0
@@ -2084,6 +2086,20 @@ def layerwise_calibrate(
             "Layerwise calibration requires a model with identifiable transformer layers."
         )
 
+    decoder_owned_ids = {id(module) for layer in transformer_layers for module in layer.modules()}
+    has_enabled_outside_quantizer = any(
+        isinstance(module, TensorQuantizer)
+        and module.is_enabled
+        and id(module) not in decoder_owned_ids
+        for module in model.modules()
+    )
+
+    if export_dir is not None and has_enabled_outside_quantizer:
+        raise ValueError(
+            "Layerwise export does not support enabled quantizers outside transformer layers. "
+            "Calibrate without export_dir, then export the completed model separately."
+        )
+
     num_layers = len(transformer_layers)
     print_rank_0(f"Layerwise calibration: Found {num_layers} transformer layers")
 
@@ -2214,6 +2230,27 @@ def layerwise_calibrate(
 
     if ckpt:
         ckpt.full_restore(transformer_layers, model)
+
+    if has_enabled_outside_quantizer:
+        if any(device == "disk" for device in getattr(model, "hf_device_map", {}).values()):
+            warn_rank_0(
+                "Layerwise calibration found enabled quantizers outside transformer layers. "
+                "The required full-model calibration pass may be slow because disk-offloaded "
+                "decoder weights can be streamed for every batch."
+            )
+
+        with _hide_modules_from_traversal(model, transformer_layers):
+            if qdq_from_prev:
+                calib_func(model, forward_loop, **calib_kwargs)
+            else:
+                with ExitStack() as stack:
+                    for layer in transformer_layers:
+                        stack.enter_context(
+                            set_quantizer_by_cfg_context(
+                                layer, [{"quantizer_name": "*", "enable": False}]
+                            )
+                        )
+                    calib_func(model, forward_loop, **calib_kwargs)
 
     if exporter is not None:
         warn_rank_0(
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -20,7 +20,6 @@ import math
 import time
 import warnings
 from collections.abc import Callable, Mapping, Sequence
-from contextlib import ExitStack
 from functools import partial
 from typing import Any, TypeAlias
 
@@ -35,7 +34,7 @@ from modelopt.torch.opt.searcher import ForwardLoop
 from modelopt.torch.quantization.utils.layerwise_calib import (
     LayerActivationCollector,
     _CheckpointState,
-    _hide_modules_from_traversal,
+    _OutsideQuantizerCalibrator,
     _reconcile_export_with_resume,
 )
 from modelopt.torch.utils import print_rank_0, warn_rank_0
@@ -62,7 +61,6 @@ from .utils import (
     enable_fake_quant,
     enable_quant,
     enable_weight_access_and_writeback,
-    has_accelerate_offload,
     is_quantized_column_parallel_linear,
     is_quantized_linear,
     is_quantized_row_parallel_linear,
@@ -2087,56 +2085,20 @@ def layerwise_calibrate(
             "Layerwise calibration requires a model with identifiable transformer layers."
         )
 
-    decoder_owned_ids = {id(module) for layer in transformer_layers for module in layer.modules()}
-    has_enabled_outside_quantizer = any(
-        isinstance(module, TensorQuantizer)
-        and module.is_enabled
-        and id(module) not in decoder_owned_ids
-        for module in model.modules()
+    outside_calibrator = _OutsideQuantizerCalibrator(
+        model,
+        transformer_layers,
+        forward_loop,
+        calib_func,
+        calib_kwargs,
+        qdq_from_prev,
+        _needs_activation_forward_for_max_calib if calib_func is max_calibrate else None,
     )
-
-    outside_calib_needs_forward = has_enabled_outside_quantizer
-    outside_calib_runs_forward = has_enabled_outside_quantizer
-    outside_calib_kwargs = calib_kwargs
-    if has_enabled_outside_quantizer and calib_func is max_calibrate:
-        with _hide_modules_from_traversal(model, transformer_layers):
-            outside_calib_needs_forward = _needs_activation_forward_for_max_calib(model)
-        outside_calib_kwargs = dict(calib_kwargs)
-        outside_calib_kwargs.setdefault("skip_forward_without_activation_calib", True)
-        outside_calib_runs_forward = (
-            outside_calib_needs_forward
-            or not outside_calib_kwargs["skip_forward_without_activation_calib"]
-        )
-
-    if export_dir is not None and outside_calib_runs_forward:
+    if export_dir is not None and outside_calibrator.runs_forward:
         raise ValueError(
             "Layerwise export does not support enabled quantizers outside transformer layers. "
             "Calibrate without export_dir, then export the completed model separately."
         )
-
-    def _calibrate_outside_quantizers():
-        if not has_enabled_outside_quantizer:
-            return
-
-        if outside_calib_runs_forward and has_accelerate_offload(model):
-            warn_rank_0(
-                "Layerwise calibration found enabled quantizers outside transformer layers. "
-                "The required full-model calibration pass may be slow because CPU- or "
-                "disk-offloaded decoder weights can be transferred for every batch."
-            )
-
-        with _hide_modules_from_traversal(model, transformer_layers):
-            if qdq_from_prev:
-                calib_func(model, forward_loop, **outside_calib_kwargs)
-            else:
-                with ExitStack() as stack:
-                    for layer in transformer_layers:
-                        stack.enter_context(
-                            set_quantizer_by_cfg_context(
-                                layer, [{"quantizer_name": "*", "enable": False}]
-                            )
-                        )
-                    calib_func(model, forward_loop, **outside_calib_kwargs)
 
     num_layers = len(transformer_layers)
     print_rank_0(f"Layerwise calibration: Found {num_layers} transformer layers")
@@ -2168,7 +2130,7 @@ def layerwise_calibrate(
     if exporter is not None and _reconcile_export_with_resume(
         exporter, checkpoint_dir, start_layer, num_layers
     ):
-        _calibrate_outside_quantizers()
+        outside_calibrator.calibrate()
         warn_rank_0(
             f"Layerwise export: every layer shard in {exporter.export_dir} is already "
             f"written.{finalize_hint}"
@@ -2270,7 +2232,7 @@ def layerwise_calibrate(
     if ckpt:
         ckpt.full_restore(transformer_layers, model)
 
-    _calibrate_outside_quantizers()
+    outside_calibrator.calibrate()
 
     if exporter is not None:
         warn_rank_0(
--- a/modelopt/torch/quantization/model_calib.py
+++ b/modelopt/torch/quantization/model_calib.py
@@ -2092,9 +2092,8 @@ def layerwise_calibrate(
         calib_func,
         calib_kwargs,
         qdq_from_prev,
-        _needs_activation_forward_for_max_calib if calib_func is max_calibrate else None,
     )
-    if export_dir is not None and outside_calibrator.runs_forward:
+    if export_dir is not None and outside_calibrator.enabled:
         raise ValueError(
             "Layerwise export does not support enabled quantizers outside transformer layers. "
             "Calibrate without export_dir, then export the completed model separately."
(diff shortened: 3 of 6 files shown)


modelopt/torch/quantization/model_calib.py
```diff
@@ -2084,6 +2086,20 @@ def layerwise_calibrate(
...
+        and module.is_enabled
+        and id(module) not in decoder_owned_ids
+        for module in model.modules()
+    )
+
+    if export_dir is not None and has_enabled_outside_quantizer:
+        raise ValueError(
+            "Layerwise export does not support enabled quantizers outside transformer layers. "
+            "Calibrate without export_dir, then export the completed model separately."
+        )
```
realAsma: BB: Can you explain to me why layerwise + export cannot support quantized layers outside of transformer layers?

🐝

🐝 I’m looking into the exact layerwise/export constraint and will follow up here with a concrete explanation.

This is a limitation of the current progressive export lifecycle, not of quantizing `lm_head` itself. An outside quantizer needs a full-model calibration pass after every decoder layer has reached its final calibrated state, so it sees the correct end-to-end activations. With `export_dir`, each decoder layer is converted/packed in place as soon as it is calibrated; once earlier layers have been exported, the root model is intentionally no longer safe to run for that final forward pass. Supporting both would require preserving or reloading executable calibrated decoder layers, or redesigning the streaming calibration/export flow. For now, the safe path is to calibrate without `export_dir`, then export the completed model separately.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2205,6 +2221,27 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+                "Layerwise calibration found enabled quantizers outside transformer layers. "
+                "The required full-model calibration pass may be slow because disk-offloaded "
+                "decoder weights can be streamed for every batch."
+            )
+
+        with _hide_modules_from_traversal(model, transformer_layers):
+            if qdq_from_prev:
+                calib_func(model, forward_loop, **calib_kwargs)
+            else:
+                with ExitStack() as stack:
```
realAsma: BB:

Why not do `with set_quantizer_by_cfg_context` directly here?

🐝 I’m on it — I’ll inspect whether the context manager can be applied directly here and follow up with the result.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2205,6 +2221,27 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
     if ckpt:
         ckpt.full_restore(transformer_layers, model)
 
+    if has_enabled_outside_quantizer:
+        if any(device == "disk" for device in getattr(model, "hf_device_map", {}).values()):
```
realAsma: BB: Is there any existing helpers we could use here like say something like `is_diskoffloaded_model`?

🐝 I’m checking the existing helpers and will follow up here with the outcome.

The closest existing helper is `has_accelerate_offload`, but it intentionally returns true for both CPU and disk offload by inspecting Accelerate hooks. This warning is specifically about repeated disk streaming, so the `hf_device_map` check is the narrower predicate. There is no existing disk-only model helper.

modelopt/torch/quantization/model_calib.py

`calib_func` needs every decoder layer’s quantizers disabled at the same time while the outside quantizers remain enabled. A plain `with` inside the loop would restore each layer before `calib_func` runs. Applying the context to `model` before hiding the layers would also disable the outside quantizers; applying it inside `_hide_modules_from_traversal` would not reach the original decoder layers. `ExitStack` keeps the per-layer contexts active together across the one `calib_func` call.

modelopt/torch/quantization/model_calib.py
realAsma: actually I want to have this warning for both cpu and disk offloaded models. so can we use `has_accelerate_offload` (may be with a local import since this module should be independent of hf plugins)?

modelopt/torch/quantization/model_calib.py
```diff
@@ -2205,6 +2221,27 @@ def _layer_forward_loop(m, _inputs=layer_inputs):
...
+            warn_rank_0(
+                "Layerwise calibration found enabled quantizers outside transformer layers. "
+                "The required full-model calibration pass may be slow because disk-offloaded "
+                "decoder weights can be streamed for every batch."
+            )
+
+        with _hide_modules_from_traversal(model, transformer_layers):
```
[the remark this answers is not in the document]

Thanks. Pipeline-parallel stage agreement requires choosing the correct PP group and forward-loop contract, so we’re postponing that design-significant distributed extension to follow-up work. The currently supported single-process and replica-symmetric paths remain unchanged.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +102,48 @@ def forward(self, *args, **kwargs):
...
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            if name in self._PROXY_BLOCKLIST:
+                raise
+            return getattr(object.__getattribute__(self, "_original"), name)
+
+    def forward(self, *args, **kwargs):
+        return self._original(*args, **kwargs)
+
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Thanks. Current callers only require forward delegation and do not write state through the temporary proxy, so we’re keeping the shared-base/write-through cleanup out of this focused fix and postponing it to follow-up work.

modelopt/torch/quantization/model_calib.py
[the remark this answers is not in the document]

Thanks. We’re postponing richer offender-list diagnostics to follow-up work to keep this fix focused. The existing message retains the safe recovery path, while the prepared calibration fix avoids rejecting outside quantizers when no post-export forward is required.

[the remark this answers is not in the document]

Published in ed7b4d7d53. The outside calibration pass now runs only when needed, preserves omitted versus explicit skip-forward behavior, and blocks export whenever that pass would execute the progressively packed model. Focused pre-commit and all 50 layerwise calibration tests pass.

modelopt/torch/quantization/model_calib.py

Published in ed7b4d7d53. The warning now covers both CPU- and disk-offloaded models and is emitted only when calibration will actually run a forward pass. Focused pre-commit and all 50 layerwise calibration tests pass.

modelopt/torch/quantization/mode.py
```diff
@@ -226,6 +226,8 @@ def wrapped_calib_func(
     So lets wrap them to be compatible with the ModelOpt convert entrypoint.
     """
     kwargs = config.model_dump()
+    if "skip_forward_without_activation_calib" not in config.model_fields_set:
+        kwargs.pop("skip_forward_without_activation_calib", None)
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Fixed in 5ff8457ec1: the config default is now tri-state so both dict and `MaxCalibConfig` inputs preserve the automatic layerwise behavior. Focused pre-commit and 152 targeted tests pass.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2084,6 +2087,57 @@ def layerwise_calibrate(
...
+    )
+
+    outside_calib_needs_forward = has_enabled_outside_quantizer
+    outside_calib_runs_forward = has_enabled_outside_quantizer
+    outside_calib_kwargs = calib_kwargs
+    if has_enabled_outside_quantizer and calib_func is max_calibrate:
+        with _hide_modules_from_traversal(model, transformer_layers):
+            outside_calib_needs_forward = _needs_activation_forward_for_max_calib(model)
+        outside_calib_kwargs = dict(calib_kwargs)
+        outside_calib_kwargs.setdefault("skip_forward_without_activation_calib", True)
```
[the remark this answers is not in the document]

Thanks. We’re keeping per-layer replay semantics unchanged in this focused non-decoder calibration fix; optimizing weight-only per-layer replay can be evaluated separately with its own performance and distributed coverage.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2084,6 +2087,57 @@ def layerwise_calibrate(
...
+        outside_calib_runs_forward = (
+            outside_calib_needs_forward
+            or not outside_calib_kwargs["skip_forward_without_activation_calib"]
+        )
```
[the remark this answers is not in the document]

Thanks. The supported config path normalizes the sentinel before dispatch, and the re-review confirms this is future hardening rather than a live defect. We’re keeping the current scoped implementation.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +102,48 @@ def forward(self, *args, **kwargs):
...
+def _hide_modules_from_traversal(model: nn.Module, modules: Sequence[nn.Module]):
+    """Temporarily hide registered modules while retaining their forward behavior."""
+    target_ids = {id(module) for module in modules}
+    slots = [
+        (parent, child_name, child)
+        for parent in tuple(model.modules())
+        for child_name, child in tuple(parent._modules.items())
+        if child is not None and id(child) in target_ids
+    ]
+    proxies = {id(child): _ForwardOnlyLayer(child) for _, _, child in slots}
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Thanks. All current layer discoverers return registered submodules, so this invariant holds today. We’ll defer a defensive guard until support for a discoverer with different ownership semantics is introduced.

modelopt/torch/quantization/model_calib.py
realAsma, line 1: BB: A lot of changes in this file can be moved to `modelopt/torch/quantization/utils/layerwise_calib.py` - Can you make the changes in this file minimal in a clean way?

realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

🐝 I’ve queued this in the canonical PR task. An existing controller pass currently owns the checkout, so I did not start a duplicate worker; the active monitor will drain this instruction into one fresh pass. I’ll leave this thread open until the refactor is published and verified.

realAsma: > _:robot: Bot comment._

Moved the outside-quantizer calibration mechanics into `_OutsideQuantizerCalibrator` in `utils/layerwise_calib.py`; `model_calib.py` now only constructs it, checks export compatibility, and invokes it at the two sequencing points. The rebase also preserves main's caller-owned export-finalization contract. Published in `a77689732`.

modelopt/torch/quantization/config.py
```diff
@@ -928,8 +928,8 @@ class MaxCalibConfig(_SharedStatesConfig, QuantizeAlgorithmConfig):
         ),
     )
 
-    skip_forward_without_activation_calib: bool = ModeloptField(
-        default=False,
+    skip_forward_without_activation_calib: bool | None = ModeloptField(
```
realAsma: Why is this change needed?

modelopt/torch/quantization/model_calib.py
```diff
@@ -2084,6 +2085,21 @@ def layerwise_calibrate(
...
         )
 
+    outside_calibrator = _OutsideQuantizerCalibrator(
+        model,
+        transformer_layers,
+        forward_loop,
+        calib_func,
+        calib_kwargs,
+        qdq_from_prev,
+        _needs_activation_forward_for_max_calib if calib_func is max_calibrate else None,
```
realAsma: Does this means that outside_calibrator is calibrated only when `calib_func` is max_calibrate ?

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +104,120 @@ def forward(self, *args, **kwargs):
...
+        self.calib_kwargs = calib_kwargs
+        self.qdq_from_prev = qdq_from_prev
+
+        # Inline import breaks nn -> qtensor -> utils -> layerwise_calib import cycle.
+        from ..nn import TensorQuantizer
+
+        decoder_owned_ids = {
+            id(module) for layer in transformer_layers for module in layer.modules()
+        }
+        self.enabled = any(
```
realAsma: can we move this self.enabled  inside `_hide_modules_from_traversal` ? this way way we can simplify this code further and avoid iteration over decoder layer quantizers. Further, can we cache the name to module map during init, then _hide_modules_from_traversal can simply look a decoder modules parent name fro the name to module map. this way multiple _hide_modules_from_traversal does not iterate over the whole model.

modelopt/torch/quantization/config.py
realAsma: > _:robot: Bot comment._

This preserves the released `bool`/`False` config contract while still distinguishing omission from an explicit `False`. For the generated layerwise outside-quantizer pass, omission allows the safe automatic behavior; explicit `False` continues to honor the caller's request to run the forward.

modelopt/torch/quantization/model_calib.py

No. The outside calibrator runs for any selected `calib_func` when an enabled outside quantizer exists. Only the activation-data predicate and skip-forward optimization are specific to `max_calibrate`; other algorithms currently run their normal calibration path.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: > _:robot: Bot comment._

Implemented in `dfde796`: the calibrator now caches unique physical decoder registration slots once, computes `enabled` while those slots are hidden, and preserves direct plus aliased-parent registrations with shared-proxy restoration. Focused success/error coverage was added.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2084,6 +2085,21 @@ def layerwise_calibrate(
...
+    )
+    if export_dir is not None and outside_calibrator.runs_forward:
+        raise ValueError(
+            "Layerwise export does not support enabled quantizers outside transformer layers. "
+            "Calibrate without export_dir, then export the completed model separately."
+        )
```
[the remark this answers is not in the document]

Addressed in `dfde796` with the conservative export contract: layerwise export now rejects whenever any enabled outside quantizer requires the post-layer calibration step, including weight-only cases. The tests and PR description now consistently reflect that condition.

modelopt/torch/quantization/config.py
```diff
@@ -938,10 +938,10 @@ class MaxCalibConfig(_SharedStatesConfig, QuantizeAlgorithmConfig):
...
-            "Opt-in (default False) because the provided ``forward_loop`` can carry side "
-            "effects the caller relies on — most notably materializing sharded parameters under "
-            "DeepSpeed ZeRO-3 — so enable it per-recipe when the calibration data is known to be "
-            "unnecessary."
+            "Set False when the provided ``forward_loop`` carries side effects the caller relies "
+            "on — most notably materializing sharded parameters under DeepSpeed ZeRO-3. The "
+            "default None behaves as False for whole-model and per-layer calibration, and as True "
+            "for ModelOpt's generated layerwise non-decoder pass."
         ),
     )
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Addressed in `dfde796`: the released `bool` field and `False` default are restored. `model_fields_set` distinguishes omission from an explicit `False` internally, preserving the serialized schema and direct-call behavior.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +104,120 @@ def forward(self, *args, **kwargs):
...
+            return super().__getattr__(name)
+        except AttributeError:
+            if name in self._PROXY_BLOCKLIST:
+                raise
+            return getattr(object.__getattribute__(self, "_original"), name)
+
+    def forward(self, *args, **kwargs):
+        return self._original(*args, **kwargs)
+
+
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Addressed in `dfde796`: the traversal helper and outside-calibrator docstrings now state that modules, parameters, buffers, and state-dict traversal are hidden during the outside pass.

modelopt/torch/quantization/mode.py
```diff
@@ -226,6 +226,8 @@ def wrapped_calib_func(
     So lets wrap them to be compatible with the ModelOpt convert entrypoint.
     """
     kwargs = config.model_dump()
+    if "skip_forward_without_activation_calib" not in config.model_fields_set:
```
realAsma: why is this needed?

realAsma: > _:robot: Bot comment._

Removed this in 76106e237. With both the config field and `max_calibrate` defaulting to `False`, the conditional was redundant; I also removed the corresponding `model_fields_set` test assertions.

modelopt/torch/quantization/mode.py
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Addressed in `76106e237`: removed the redundant field-set branch and the bookkeeping-only assertions. Both config paths now rely on the existing `False` default.

modelopt/torch/quantization/model_calib.py
```diff
@@ -2114,6 +2129,7 @@ def layerwise_calibrate(
     if exporter is not None and _reconcile_export_with_resume(
         exporter, checkpoint_dir, start_layer, num_layers
     ):
+        outside_calibrator.calibrate()
```
cjluo-nv, line 2132: > _Bot comment._

This call is unreachable in effect: `exporter` is non-`None` only when `export_dir is not None`, and the new check above raises whenever `export_dir is not None and outside_calibrator.enabled`. So `enabled` is always `False` here and `calibrate()` returns immediately. Drop the line, or add a comment saying why it's kept defensively.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +104,116 @@ def forward(self, *args, **kwargs):
...
+            for parent in tuple(model.modules())
+            for child_name, child in parent._modules.items()
+            if child is not None and id(child) in layer_ids
+        ]
+
+        with _hide_modules_from_traversal(self.transformer_layer_slots):
+            self.enabled = any(
+                isinstance(module, TensorQuantizer) and module.is_enabled
+                for module in model.modules()
+            )
```
cjluo-nv, line 176: > _Bot comment._

The new pass replaces entries in `parent._modules` while accelerate/FSDP hooks are live on the real layers. The CPU tests only monkeypatch `has_accelerate_offload`, so the proxy is never exercised against an actual `_hf_hook`/`AlignDevicesHook` layout. `tests/gpu/torch/quantization/plugins/test_accelerate_gpu.py` already has `make_cpu_offloaded_model` and disk-offload helpers — please add a case there with an enabled `lm_head` quantizer.

tests/unit/torch/quantization/test_calib.py
```diff
@@ -716,7 +716,8 @@ def test_needs_activation_forward_for_static_bias_calibrator():
 
 def test_max_calib_config_skip_is_opt_in():
     """The flag is opt-in (default False) so it does not change behavior for direct callers."""
-    assert MaxCalibConfig().skip_forward_without_activation_calib is False
+    default_config = MaxCalibConfig()
```
cjluo-nv, line 719: > _Bot comment._

This is unrelated churn — hoisting `MaxCalibConfig()` into `default_config` doesn't change the assertion or the test. Please revert to keep the diff focused.

cjluo-nv (commented):

Logic looks correct and CPU test coverage is genuinely thorough, but a few things need an owner's call before merge — most notably an unreachable `calibrate()` call and no offload-path test for the new full-model pass.

Needs action:
- [ ] Remove or comment the `outside_calibrator.calibrate()` added in the export-resume early-return in `model_calib.py`: that branch is only reachable when `export_dir is not None`, which the new `ValueError` already guarantees `enabled is False`, so the call can never do work.
- [ ] Add a GPU test in `tests/gpu/torch/quantization/plugins/test_accelerate_gpu.py` exercising the new pass on a CPU/disk-offloaded model — `_hide_modules_from_traversal` swaps `parent._modules` under live accelerate hooks and only the `has_accelerate_offload` flag is currently monkeypatched.
- [ ] Confirm the new hard `ValueError` for layerwise export + enabled outside quantizers is acceptable as a silent break for existing configs, and add a CHANGELOG line if so.
- [ ] Confirm the extra full-model forward should run by default even when the outside quantizers are weight-only (see `test_layerwise_max_outside_calibration_uses_configured_forward_behavior`, `expected_forward_calls == 2`).
- [ ] Revert the unrelated cosmetic edit in `tests/unit/torch/quantization/test_calib.py`.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +104,116 @@ def forward(self, *args, **kwargs):
         )
 
 
+class _ForwardOnlyLayer(nn.Module):
```
Fridah-nv: Do you think it's better to reuse `_SkipLayer` here? The proxy machinery can be reused and we need a new forward function for this feature

realAsma: BB: This is a good point, can you give a minimal design for this?

realAsma: > _:robot: Bot comment._

🐝 I’m taking a look and will follow up with a minimal design.

realAsma: > _:robot: Bot comment._

Minimal design: extract only the shared proxy mechanics into a small `_LayerProxy(nn.Module)` base. Its `__init__` stores `_original` without registering it, and its `__getattr__` applies the existing hook blocklist before forwarding to `_original`. Then keep `_SkipLayer(_LayerProxy)` responsible for skip state plus the zero/meta `forward`, and `_ForwardOnlyLayer(_LayerProxy)` responsible only for delegating `forward` to `_original`. That reuses the machinery without making the forward-preserving case inherit skip semantics. I’d cover the base behavior through the existing alias/blocklist tests rather than add another abstraction or callback strategy.

realAsma: Can we have a simpler design -> Just add the forward with hide module and without any meta data manipulation support as well to skip layer ?  then for outside we use  forward with hide module and without any meta data

realAsma: BB: PTAL https://github.com/NVIDIA/Model-Optimizer/pull/2339#discussion_r3993129709

realAsma: > _:robot: Bot comment._

🐝

realAsma: > _:robot: Bot comment._

I read this as folding the two wrappers into `_SkipLayer`: keep the current metadata-backed zero-output mode for calibrated layers, and add a forward-through-original mode that hides registered modules without creating `_LayerCalibState`. `_hide_modules_from_traversal` would use the latter, so `_ForwardOnlyLayer` disappears. Please confirm that is the intended shape and I’ll implement it.

realAsma: @Fridah-nv updated the code, please see the latest, this was a good feedback -thanks!

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -96,11 +100,102 @@ def __getattr__(self, name: str):
...
+            if not warned_for_offload:
+                warned_for_offload = True
+                if has_offload:
+                    warn_rank_0(
+                        "Layerwise calibration found enabled quantizers outside transformer "
+                        "layers. The required full-model calibration pass may be slow because "
+                        "CPU- or disk-offloaded decoder weights can be transferred for every "
+                        "batch."
+                    )
+            return self.forward_loop(model)
```
realAsma: BB: This is complicating the warning process. Just move the warning outside of the forward loop patch and get rid of this unnecessary wrapper.

realAsma: > _:robot: Bot comment._

🐝 On it — I’ll simplify the warning flow and remove the unnecessary wrapper.

modelopt/torch/quantization/model_calib.py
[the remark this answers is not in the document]

Removed the unreachable completed-resume calibration call in `dded6ad1c`.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: > _:robot: Bot comment._

Addressed in `b39586a1f`: moved the offload warning before calibration and now pass `self.forward_loop` directly to `calib_func` in both paths. I also removed redundant warning assertions and collapsed the equivalent weight-only export parameter cases.

Fridah-nv (approved):
LGTM, thanks!

realAsma:
> Is this feature mainly targeting lm_head? For lm_head's case I think there's an optimization to take the last layer's output and feed into the rest of the modules. This saves a full model forward that can be expensive for an offloaded model.

@Fridah-nv  yes this feature is mainly targeting `lm_head` but it could also be useful for MTP/VLM. I did not want to add complexity by optimizing more. My reasoning was that `lm_head` etc are quantized mainly for small models for which `lm_head` has a large footprint. Disk-offloading is mainly done for large models which might not need lm_head calibration or outside quantizer calibration.

modelopt/torch/quantization/model_calib.py

Addressed — the unreachable `outside_calibrator.calibrate()` on the completed-export-resume path was removed in `dded6ad1c`. At head (`b39586a1f`) there are exactly three `outside_calibrator` references in `model_calib.py`: construction (~:2088), the export guard (~:2096), and a single `calibrate()` on the main path (~:2233). Resolving as addressed.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: > _:robot: Bot comment._

Addressed — `test_layerwise_calibrates_lm_head_with_accelerate_offload` was added in `tests/gpu/torch/quantization/plugins/test_accelerate_gpu.py` (~:196). It builds a real CPU-offloaded model via `make_cpu_offloaded_model`, asserts the decoder layer carries a live `AlignDevicesHook` with meta-device parameters *before* quantizing, then runs layerwise calibration with an enabled `*lm_head*` quantizer and asserts `lm_head.input_quantizer.amax is not None`. So the `parent._modules` swap in `_hide_modules_from_traversal` is exercised against real accelerate dispatch rather than a monkeypatched `has_accelerate_offload`. `gpu-tests` is green at head. Resolving as addressed.

tests/unit/torch/quantization/test_calib.py
realAsma: > _:robot: Bot comment._

Addressed — the unrelated `MaxCalibConfig()` hoist was reverted. At head (`b39586a1f`) `git diff origin/main...HEAD --name-only` lists only four files (`model_calib.py`, `utils/layerwise_calib.py`, `tests/gpu/.../test_accelerate_gpu.py`, `tests/unit/.../test_layerwise_calibrate.py`); `tests/unit/torch/quantization/test_calib.py` is no longer part of this PR's diff. Resolving as addressed.

modelopt/torch/quantization/model_calib.py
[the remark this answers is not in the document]

Outdated — the `runs_forward` mechanism this describes no longer exists at head (`b39586a1f`). `grep -n runs_forward` returns nothing in either `model_calib.py` or `utils/layerwise_calib.py`; the export guard is now an unconditional `if export_dir is not None and outside_calibrator.enabled: raise ValueError(...)`, so it no longer keys off `skip_forward_without_activation_calib`. The guard is deliberately broader rather than narrower (weight-only + export is rejected too), which is the intended conservative contract. Resolving as outdated.

modelopt/torch/quantization/config.py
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Outdated — `config.py` is no longer touched by this PR. At head (`b39586a1f`) it is absent from `git diff origin/main...HEAD --name-only`, and `config.py:931` still declares `skip_forward_without_activation_calib: bool = ModeloptField(...)` with default `False`; the `bool | None` widening was reverted. There is no public-surface change left to note in the CHANGELOG on this account. Resolving as outdated.

modelopt/torch/quantization/mode.py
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Outdated — `mode.py` is no longer touched by this PR; the `model_fields_set` pop was reverted in `76106e237`. At head (`b39586a1f`) `mode.py` is absent from `git diff origin/main...HEAD --name-only`. Resolving as outdated.

modelopt/torch/quantization/mode.py
realAsma: > _:robot: Bot comment._

Outdated — that `mode.py` change was removed in `76106e237`, and `mode.py` is no longer part of this PR's diff at head (`b39586a1f`). Resolving as addressed.

modelopt/torch/quantization/utils/layerwise_calib.py
realAsma: > _:robot: Bot comment._

Addressed — both halves are in at head (`b39586a1f`). `self.enabled` is now computed inside the `_hide_modules_from_traversal` context in `_OutsideQuantizerCalibrator.__init__`, so it no longer iterates decoder-layer quantizers. And `self.transformer_layer_slots` is built there from a single `tuple(model.modules())` walk, so `_hide_modules_from_traversal` now takes precomputed slots and just swaps `parent._modules[child_name]` — the later hide inside `calibrate()` is O(len(slots)) with no model-wide traversal. Resolving as addressed.

modelopt/torch/quantization/utils/layerwise_calib.py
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Addressed — the docstring was reworked; at head (`b39586a1f`) `_hide_modules_from_traversal` reads "Retain forward behavior while hiding registered modules from traversal and state dicts." State dicts are called out explicitly, and "traversal" covers `modules()`, `named_parameters()` and `named_buffers()` by the same mechanism: the proxy registers no parameters or buffers of its own (`_original` is set via `object.__setattr__`), so all three are hidden together. Resolving as addressed.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +104,120 @@ def forward(self, *args, **kwargs):
...
+        )
+        self.runs_forward = self.enabled
+        if self.enabled and activation_forward_predicate is not None:
+            with _hide_modules_from_traversal(model, transformer_layers):
+                needs_forward = activation_forward_predicate(model)
+            self.calib_kwargs = dict(calib_kwargs)
+            self.calib_kwargs.setdefault("skip_forward_without_activation_calib", True)
+            self.runs_forward = (
+                needs_forward or not self.calib_kwargs["skip_forward_without_activation_calib"]
+            )
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Duplicate of the earlier pipeline-parallel rank-divergence thread on `has_enabled_outside_quantizer` (https://github.com/NVIDIA/Model-Optimizer/pull/2339#discussion_r3937867971), which raises the same mechanism and the same remedy — only the symbol names changed with the refactor into `_OutsideQuantizerCalibrator`. That thread was already answered and closed with the decision to postpone the distributed extension, since stage agreement requires choosing the correct PP group and a forward-loop contract. Same disposition applies here. Resolving as a duplicate of an already-decided item.

modelopt/torch/quantization/utils/layerwise_calib.py
```diff
@@ -101,6 +104,120 @@ def forward(self, *args, **kwargs):
...
+            from ..conversion import set_quantizer_by_cfg_context
+
+            with ExitStack() as stack:
+                for layer in self.transformer_layers:
+                    stack.enter_context(
+                        set_quantizer_by_cfg_context(
+                            layer, [{"quantizer_name": "*", "enable": False}]
+                        )
+                    )
+                self.calib_func(self.model, self.forward_loop, **self.calib_kwargs)
```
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

The full forward replay is an accepted simplicity tradeoff for this general `lm_head`/MTP/VLM path; the rationale is documented in https://github.com/NVIDIA/Model-Optimizer/pull/2339#issuecomment-5684842517. Decoder subtrees are hidden from calibration-algorithm traversal during the outside pass, so this does not recalibrate decoder weights.

modelopt/torch/quantization/model_calib.py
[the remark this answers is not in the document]

To clarify the earlier reply on this thread: the predicate is reachable rather than hardcoded, and it lives outside this diff.

`_needs_activation_forward_for_max_calib` and `max_calibrate`'s `skip_forward_without_activation_calib` are both on `main` already (they landed in #1963), which is why they do not show up in this PR's diff. `layerwise_calibrate` pops only the layerwise-specific keys off `calib_kwargs` and forwards the remainder unchanged into the outside pass, so setting `skip_forward_without_activation_calib: true` in the `algorithm` dict makes the extra pass skip its forward for a weight-only recipe. `test_layerwise_max_outside_calibration_uses_configured_forward_behavior` pins this at `(None, 2)`, `(False, 2)`, `(True, 1)` using `INT8_WEIGHT_ONLY_CFG` with `lm_head` enabled.

I deliberately did not hardcode `True` inside the wrapper. Whether a forward is needed is the selected algorithm's policy and `max_calibrate` already owns it; duplicating the decision in the layerwise wrapper would make the two disagree for MSE/GPTQ/SmoothQuant/SVDQuant, which always need activations. The default stays `False` to match `max_calibrate`'s own default, so a layerwise run does not silently diverge from the equivalent non-layerwise run.

On the `export_dir` rejection: the broad `is_enabled` predicate is intentional there. Export shards cover decoder layers only, so an enabled outside quantizer is unrepresentable in the output whether or not it needed a forward. `test_layerwise_export_rejects_weight_only_outside_quantizer` locks that in.

realAsma:
> _:robot: Bot comment._

@cjluo-nv — following up on the remaining two checklist items from your review at `76106e237`. Items 1, 2 and 5 were addressed and answered on their inline threads.

**Item 4 — "confirm the extra full-model forward should run by default even when the outside quantizers are weight-only."**

Yes, and it is configurable rather than fixed. The default of two forwards matches `max_calibrate`'s own `skip_forward_without_activation_calib=False` default, so a layerwise run behaves like the equivalent non-layerwise run instead of silently skipping a pass the plain path would have taken. `layerwise_calibrate` forwards the residual `calib_kwargs` unchanged into the outside pass, so setting `skip_forward_without_activation_calib: true` in the `algorithm` dict reduces it to one forward — that is the `(True, 1)` parameterization in `test_layerwise_max_outside_calibration_uses_configured_forward_behavior`. The decision stays with the algorithm rather than the layerwise wrapper, since algorithms that always need activations (MSE, GPTQ, SmoothQuant, SVDQuant) must not inherit a skip.

**Item 3 — "confirm the hard `ValueError` for layerwise export + enabled outside quantizers is acceptable, and add a CHANGELOG line if so."**

Still open. `CHANGELOG.rst` is not in this diff, so there is no entry yet under the 0.48.0 *Backward Breaking Changes* section. Flagging it as outstanding rather than closing it out here.

**Separately, `#discussion_r3982458853` stays open on purpose.** The pre-decoder case it raises is real and not addressed by this change: the per-layer loop only disables the quantizers of the layer currently being calibrated, so an enabled upstream quantizer (embedding, vision tower, input projection) is live and uncalibrated for the whole loop and the decoder layers see its dynamic-amax output. Calibrating it in the final pass does not undo that. This PR targets the post-decoder tail, so it is latent rather than a regression for the motivating case, but I would rather leave the thread open than claim it fixed.

modelopt/torch/quantization/utils/layerwise_calib.py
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

This one is accurate and not fixed in this PR. The per-layer loop only disables the quantizers of the layer currently being calibrated, and each input-capture pass re-runs the full forward, so an enabled pre-decoder quantizer (embedding, vision tower, input projection) is live and uncalibrated for the whole loop and the decoder layers see its dynamic-amax output; calibrating it in the final pass does not retroactively fix that. I am keeping the scope of this PR to making outside quantizers calibrated at all, and tracking the pre-/post-decoder partition — plus the matching docstring statement of the assumption — as follow-up rather than growing this change.

CHANGELOG.rst
```diff
@@ -17,6 +18,7 @@ Changelog
 
 **Backward Breaking Changes**
 
+- Layerwise calibration now raises ``ValueError`` when ``layerwise.export_dir`` is combined with an enabled quantizer outside the transformer layers. Calibrate without ``layerwise.export_dir``, then export the completed model separately.
```
realAsma: Can we remove this? This is minor very specific backward breaking change. It is not even really backward breaking -> We did not support layerwise calibration with outside quantizers previously anyway. Keeping this just causes confusion.

realAsma: > _:robot: Bot comment._

Removed in e8a7c006bb. Agreed—this is a limitation of a new, unreleased feature rather than a backward break from a previously supported configuration. The outside-operator support remains documented under New Features / Quantization.

PR #2342, opened 2026-09-05
LiLiCorr training

mrusanovsky:
### What does this PR do?

Type of change: new feature

Adds **LiLiCorr**, a candidate-lattice reranker for DFlash drafts, as a new `projector_type` on the
existing `dflash` mode — plus three DFlash-wide improvements that apply to every variant, and an
optional composition with DFlash2's grouped convolutions.

A DFlash drafter is trained on per-position marginals rather than on the joint block distribution,
so its drafted tokens are individually plausible yet jointly incoherent. LiLiCorr keeps the top-`k`
candidates the backbone already produces at each block position, scores transitions between adjacent
candidates with a small two-layer transformer, and commits a path through the lattice greedily.
Serving is unchanged in kind: verify still checks every drafted token against the target, so the
emitted distribution is untouched and only acceptance length moves.

- Paper: [LiLiCorr: Lightweight Likelihood Correlation of Parallel Drafts for Speculative Decoding](https://arxiv.org/abs/2608.20530) (arXiv:2608.20530)
- Blog: https://research.nvidia.com/labs/nemotron/lilicorr/
- **Companion PR — serving support:** [sgl-project/sglang#37462](https://github.com/sgl-project/sglang/pull/37462)

This PR is the **training** half. It trains the drafters and exports them; the companion PR above is
what serves the resulting checkpoints, and is what the comparison table below was measured through.

**What is in the commits**

| | |
| --- | --- |
| LiLiCorr draft variant | `hf_lilicorr.py`, `modeling_lilicorr.py`, conversion routing, config fields, export |
| Three DFlash-wide features | fp32 master weights for the draft, draft activation checkpointing, and a DDP hang fix — all default-off or behaviour-preserving, all applying to `dflash`, `domino`, `dspark` and `dflash2` alike |
| Optional grouped convolutions | composes LiLiCorr with DFlash2's `DFlashGroupedConv`; see the dependency note below |
| Two recipes | `lilicorr.yaml` and `lilicorr_conv.yaml` |
| CPU unit tests, CHANGELOG, one launcher example | |

**⚠️ The convolutions depend on the DFlash2 branch, and cannot run until it merges.**

`modeling_lilicorr.py` imports `DFlashGroupedConv` from `modeling_dflash2`, which today exists only
on `haoguo/dflash2-support`. The class is **imported rather than copied on purpose** — it is the only
way the two variants cannot drift apart arithmetically — but the consequence is that the
convolutional recipe cannot run against `main` as it stands.

So the import is **deferred into `_install_sublayer_convs`** rather than taken at module scope.
Everything else in this PR, including the plain LiLiCorr reranker, has no DFlash2 dependency at all
and works on `main` today; an eager import would have made the whole plugin unimportable for the sake
of one optional feature. Requesting the convolutions without DFlash2 present raises an `ImportError`
naming the two config keys to remove, rather than failing at import time.

**This PR carries two of @h-guo18's commits, with authorship and sign-off preserved.** Both are
independent of DFlash2 itself and both are needed here:

- `1419d47e`, the no-op sublayer seam. Without it `DFlashDecoderLayer.forward` never calls the
  wrappers the convolutions install onto, so the modules would be built, counted and exported while
  computing nothing. It is arithmetically an identity on its own.
- `ba377e7a`, the RoPE-θ fix. On Transformers 5 a config carries both a top-level `rope_theta` and a
  `rope_parameters` dict; the real base lives in the dict while the class default (10,000 for Qwen3)
  stays visible as the flat attribute. Reading the flat field first builds a draft whose RoPE base is
  100× off a Qwen3-8B target's, which trains and exports without complaint. Both the training-side
  enforcement and the exporter's `_get_rope_theta` are affected on `main` today.

Both are @h-guo18's work and belong to their branches; they are carried here only so that this PR
stands on its own. **If those branches land first, this PR can be rebased onto them and the two
commits dropped**, and they can equally be split out now if that is easier to review.

The same applies to `dflash_fp32_master_weights`, which is also in flight on
`haoguo/dflash-fp32-master-weights`. The field name is shared deliberately so that there is only
ever one knob rather than two spellings of it, and both versions default to off. Whichever lands
first, this PR can be rebased onto it.

### Usage

Train with the shipped recipe:

```python

config = load_recipe("general/speculative_decoding/lilicorr.yaml")
# Qwen3-8B target, 6 epochs, block size 16 (15 drafted slots, 16 verified),
# DFlash decay objective at gamma 7.0, fp32 master weights for the draft.
```

Or convert directly:

```python
import modelopt.torch.speculative as mtsp

config = {
    "dflash_block_size": 16,
    "dflash_loss_objective": "decay",
    "dflash_loss_decay_factor": 7.0,
    "dflash_fp32_master_weights": True,
    "dflash_lilicorr_w_ce": 0.25,
    "dflash_lilicorr_w_margin": 0.0,
    "dflash_lilicorr_w_pen": 0.25,
    "dflash_architecture_config": {
        "num_hidden_layers": 5,
        "projector_type": "lilicorr",
        "lilicorr_candidate_topk": 8,
        # Optional, and all-or-nothing: adding these two keys wraps every draft
        # sublayer in DFlash2's grouped convolution. Requires the DFlash2 variant.
        # "conv_kernel_size": 2,
        # "conv_group_size": 16,
    },
}
mtsp.convert(model, [("dflash", config)])
```

### Results

Six drafters for a **Qwen3-8B** target, all trained **in ModelOpt on one matched contract** — the
same corpus, schedule and block geometry for every arm, so no row carries a training advantage.
Training data is NVIDIA's
[Nemotron Post-Training Dataset v2](https://huggingface.co/datasets/nvidia/Nemotron-Post-Training-Dataset-v2)
with the multilingual split excluded, generated from the target with **thinking disabled**;
**6 epochs**; block size 16 (15 drafted slots, 16 verified); DFlash decay objective at gamma 7;
**8 nodes × 8 H100, global batch size 64** (one sequence per device, no gradient accumulation).

All six were then exported and served through SGLang on a **single H100 80GB**, `tp_size 1`, at
concurrency 1, greedy, `fa3`, mean of two replicates, with the whole node held exclusive per
benchmark. Speedup is output tokens/s against an autoregressive baseline measured in the same
allocation.

Cells are `acceptance length / speedup-vs-AR`; **★ fastest, ☆ second fastest**:

| benchmark | LiLiCorr+conv | LiLiCorr | DSpark | DFlash2 | Domino | DFlash |
|---|---|---|---|---|---|---|
| gsm8k | ★ 7.715 / 5.26x | ☆ 7.557 / 5.22x | 7.375 / 4.86x | 7.252 / 5.06x | 7.225 / 4.87x | 6.341 / 4.59x |
| math500 | ★ 9.241 / 6.54x | ☆ 9.064 / 6.52x | 9.012 / 6.15x | 8.999 / 6.49x | 8.976 / 6.25x | 7.909 / 5.88x |
| aime25 | ★ 8.285 / 6.03x | ☆ 8.156 / 6.03x | 8.043 / 5.61x | 7.967 / 5.91x | 8.066 / 5.77x | 7.126 / 5.44x |
| humaneval | ★ 7.393 / 4.01x | 7.077 / 3.93x | 7.163 / 3.72x | ☆ 7.081 / 3.95x | 6.864 / 3.73x | 6.156 / 3.68x |
| mbpp_sanitized | ★ 5.999 / 4.18x | ☆ 5.849 / 4.13x | 5.888 / 3.95x | 5.685 / 4.05x | 5.679 / 3.91x | 5.027 / 3.70x |
| livecodebench | ★ 7.975 / 5.40x | ☆ 7.754 / 5.33x | 7.775 / 5.10x | 7.601 / 5.26x | 7.553 / 5.04x | 6.808 / 4.88x |
| alpaca_eval | ☆ 3.697 / 2.69x | ★ 3.656 / 2.70x | 3.588 / 2.52x | 3.467 / 2.58x | 3.627 / 2.59x | 3.222 / 2.46x |
| mtbench | ★ 4.014 / 2.94x | ☆ 3.939 / 2.93x | 3.957 / 2.78x | 3.748 / 2.80x | 3.948 / 2.84x | 3.478 / 2.67x |

**Against every other approach in the table, LiLiCorr with convolutions is the fastest on all eight
benchmarks.** Plain LiLiCorr is the fastest on seven of the eight; the exception is humaneval, a
164-prompt slice, where DFlash2 is ahead by 0.5%.

`DFlash` is the deliberately head-free control; every head clears it by +7.60% to +21.67% on
acceptance, which is the check that a head actually loaded. Reproducing the `LiLiCorr+conv` column
additionally needs the DFlash2 variant.

Acceptance length is bit-reproducible under greedy decoding and its replicate spread here was 0.00%
on every benchmark; throughput has a ~0.2% floor.

### What `dflash_fp32_master_weights` does, and what it is worth

Today the draft is cast to the frozen base model's dtype — bf16 — before the optimizer is built.
AdamW then allocates its moments with `zeros_like(p)`, so the **optimizer state becomes bf16 too**.
That is the problem: bf16 has too few mantissa bits to represent the small updates Adam's second
moment accumulates, so those updates round away and the effective step size decays on its own,
independently of the learning-rate schedule.

The flag is standard mixed precision instead: the draft's master weights stay in fp32 while the
matmuls run in bf16. It requires a bf16 autocast around the forward, which HF `Trainer` supplies
under `TrainingArguments.bf16`. Paths that do not go through the Trainer — evaluation,
`pseudo_speculative_generate`, a plain `convert()` and forward — currently need the caller to
supply it, and no shipped recipe exercises those (`estimate_ar: false`, `do_eval: false`). Making
the draft supply its own autocast is a follow-up, held back from here on review because it touches
every DFlash variant and wants e2e coverage of the existing recipes.

Compute speed is unchanged. The cost is memory, about 12 bytes per parameter for the weight plus
Adam's two moments instead of 6, plus a doubled gradient all-reduce under DDP, since fp32
parameters mean fp32 gradients. Under FSDP2 that second cost is what
`MixedPrecisionPolicy(reduce_dtype=...)` exists to control.

It is worth **7 to 14 percent of acceptance length**, measured at the end of training on gsm8k, and
it helps every projector type:

| arm | bf16 | fp32 | Δ acceptance length |
| --- | ---: | ---: | ---: |
| LiLiCorr | 6.8670 | 7.5573 | **+10.05%** |
| DFlash2 | 6.7396 | 7.2518 | **+7.60%** |
| Domino | 6.5854 | 7.2252 | **+9.71%** |
| DSpark | 6.4621 | 7.3752 | **+14.13%** |
| DFlash | 5.9030 | 6.3412 | **+7.42%** |

Every arm in the comparison table above was trained with it on, and **both shipped recipes set it
`true`**, so the documented path gets it.

It defaults to **off**, so no existing DFlash, Domino or DSpark run changes behaviour. Both shipped
LiLiCorr recipes set it `true`, which is the arithmetic their numbers were trained with. Flipping
the default is a reasonable follow-up once the autocast above is in.

The draft is drawn in fp32 and, under this flag, kept there; an unpromoted run rounds the same draw
to the base model's dtype. So the bf16 and fp32 rows of the table above start from the same
initialization at the precision each trains in, rather than from two different draws. A unit test
pins that.

The flag also survives a resume. `modify()` runs under `from_pretrained` with the base model still
on meta and cannot place the draft at all, so `restore_draft_precision` re-applies the dtype, the
device and the rotary buffer once the weights are loaded and before the Trainer builds the
optimizer — the last point that can still decide the Adam moment dtype. It also reloads the draft's
tensors at the dtype they were saved in, since checkpoints store the draft in fp32 while the base is
bf16 and `dtype="auto"` gives every tensor one dtype.

@h-guo18 has the same field in flight on `haoguo/dflash-fp32-master-weights`, plus an
HF-format-resume fix this PR does not have. The name is shared deliberately so there is only ever
one knob; whichever lands first, the other should be dropped rather than merged.

### Testing

- **257 CPU unit tests pass** across `tests/unit/torch/speculative/`, including the existing DFlash,
  Domino, DSpark and Eagle suites. 48 of them are new and cover LiLiCorr specifically: conversion
  routing, head geometry, the required-field validation, the three-term objective and its absolute
  weights, gradient reach into both the head and the drafter body, and the export contract.
- Both recipes load and validate through `modelopt.recipe.load_recipe`.
- The three DFlash-wide changes are covered behaviourally: the fp32 flag is checked on the
  optimizer's moment dtypes rather than only on parameters, since the moments are the point of the
  change, and on the initialization described above; activation checkpointing is asserted to leave
  draft gradients bit-identical with the flag on and off; and the rotary buffer is asserted present
  after `modify()` on a real device while still deferred on meta, which is the case the laziness
  existed for.
- The resume path has its own test: after a `save_pretrained` / `from_pretrained` round trip,
  `restore_draft_precision` is asserted to return the draft to fp32 with its stored weights intact
  and its Adam moments in fp32. Without it the draft comes back in the base dtype with the flag
  still set, which is the failure it exists to prevent.
- `TestDFlashLazyRotaryEmb` was updated rather than left passing: it asserted the rotary buffer does
  *not* exist after convert, and the DDP fix deliberately changes that on non-meta devices. The
  replacement pins the refined invariant in both directions.
- The published checkpoints were trained with this arithmetic, verified rather than assumed: a
  fingerprint over draft initialisation, loss and gradients is compared against the pre-review tree
  for both `dflash` and `lilicorr`. Loss and gradients are **bitwise identical**. Initialisation
  moves, by less than bf16 resolution, and that is the single-dtype change described above.

- Is this change backward compatible?: ✅ — every addition is opt-in. The new `projector_type` is
  selected only by config, `dflash_fp32_master_weights` defaults to off, and the
  activation-checkpointing and DDP fixes preserve behaviour. No existing default changes.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in
  `CONTRIBUTING.md`: ✅ — no new dependencies. Four files carry `# Adapted from
  https://github.com/sgl-project/SpecForge/...` headers for the DFlash backbone and loss they derive
  from (Apache-2.0), matching the attribution already on `hf_dflash.py` in this repo. The two
  commits described above are @h-guo18's, cherry-picked with authorship and sign-off preserved.
- Did you write any new necessary tests?: ✅ — 48 new CPU tests, plus the updated rotary test.
- Did you get Claude approval on this PR?: ❌ — will run `/claude review` once opened.

### Additional Information

The convolutional recipe is the memory worst case: at an 8B target, combined with fp32 master
weights, it may need `training.gradient_checkpointing: true` to fit on 80 GiB, and it fits without at
4B. Checkpointing is mathematically neutral — same objective, same data order, same resulting model —
but it trades step time for memory, so a run using it is not step-time-comparable with one that does
not. The recipe header says so.

## Summary by CodeRabbit

* **New Features**
  * Added LiLiCorr speculative decoding with candidate-lattice reranking, configurable objectives, metrics, export support, and optional grouped convolutions.
  * Added FP32 master-weight support with improved mixed-precision behavior and gradient checkpointing.
  * Added LiLiCorr training recipes and a Qwen3-8B launcher configuration.
* **Bug Fixes**
  * Improved rotary-embedding configuration handling and corrected DFlash distributed-training hangs.
  * Added validation for invalid LiLiCorr configurations and improved exported reranking metadata.
* **Documentation**
  * Expanded guidance for FP32 master weights, training workflows, and LiLiCorr configuration.
* **Tests**
  * Expanded coverage across training, evaluation, generation, export, and checkpoint workflows.


modelopt/torch/speculative/config.py
```diff
@@ -280,6 +280,79 @@ class DFlashConfig(ModeloptBaseConfig):
...
+            "single step can change `v` by at most 0.1%, but the smallest change bf16 can "
+            "represent near `v` is about 0.4%. Every DECREASE therefore rounds back to the "
+            "same number, `v` can only grow, and since the update is divided by `sqrt(v)` "
+            "the effective step size only shrinks -- from step 1, at any learning rate.\n\n"
+            "Applies to every projector_type. Off by default because of the memory cost."
+        ),
+    )
+
+    dflash_lilicorr_w_ce: float = ModeloptField(
+        default=-1.0,
```
h-guo18, line 311: Shall we set the default to the recommended value instead of -1? (Same for other fields)

modelopt/torch/speculative/config.py
```diff
@@ -280,6 +280,79 @@ class DFlashConfig(ModeloptBaseConfig):
         ),
     )
 
+    dflash_fp32_master_weights: bool = ModeloptField(
+        default=False,
```
h-guo18, line 284: Shall we set default to True?

modelopt/torch/speculative/config.py
mrusanovsky: The -1.0 means "field not set / LiLiCorr not active." HFLiLiCorrModel.modify() validates the three weights all-or-nothing and rejects a config that sets some but not others, so a user who never selects projector_type=lilicorr never touches these fields at all. Setting the default to 0.25 would mean a plain DFlash user who doesn't set projector_type would silently have non-zero LiLiCorr loss weights. Happy to revisit if there's a better convention in this codebase.

modelopt/torch/speculative/config.py
mrusanovsky: Agreed. Changed to True in the latest commit (de514d7a). The 7–14% acceptance-length improvement is too significant to leave off by default. Users who need to reduce optimizer memory can set it to False explicitly. CHANGELOG and field description updated to match.

tools/launcher/examples/Qwen/Qwen3-8B/hf_online_lilicorr.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - dflash.dflash_loss_decay_factor=7
+      - dflash.dflash_mask_token_id=151669
+      - dflash.dflash_self_logit_distillation=false
+      # Objective composition: this is the `base` variant. For `margin`, set
+      # w_ce=0.125 and w_margin=0.125 — nothing else changes.
+      - dflash.dflash_lilicorr_w_ce=0.25
+      - dflash.dflash_lilicorr_w_margin=0.0
+      - dflash.dflash_lilicorr_w_pen=0.25
+      - dflash.dflash_lilicorr_margin=2.0
+    environment:
```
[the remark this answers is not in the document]
mrusanovsky: Not applicable here. MLM_MODEL_CFG and QUANT_CFG are environment variables for the Megatron-LM PTQ/export launcher scripts (common/megatron_lm/). The "Adding a New Model Config" convention in CLAUDE.md explicitly scopes to megatron_lm_ptq.yaml. hf_online_lilicorr.yaml calls common/specdec/dflash_online_training.sh, which does not consume either variable, and neither do the parallel hf_online_dflash.yaml, hf_online_domino.yaml, or hf_online_eagle3.yaml launchers.

modelopt/torch/speculative/plugins/modeling_dflash.py
```diff
@@ -372,6 +402,23 @@ def __init__(self, config):
         self.norm = _NORM_CLS(config.hidden_size, eps=config.rms_norm_eps)
         self._rotary_config = config  # Used by _maybe_init_rotary_emb
 
+        # Activation checkpointing for the DRAFT, which is the only trainable part of a
+        # DFlash setup. `PreTrainedModel._set_gradient_checkpointing` walks `self.modules()`
```
h-guo18: It feels a little hacky here to rely on private method `PreTrainedModel._set_gradient_checkpointing`. Shall we inherit from `GradientCheckpointingLayer` instead?

modelopt/torch/speculative/plugins/modeling_dflash.py
```diff
@@ -372,6 +402,23 @@ def __init__(self, config):
...
+        #
+        # The default func is NON-REENTRANT on purpose. HuggingFace's own default is
+        # `use_reentrant=True`, which loses track of unused parameters and can hang under
+        # `ddp_find_unused_parameters=True`. A caller who goes through
+        # `gradient_checkpointing_enable` overwrites this with their own kwargs, so pass
+        # `gradient_checkpointing_kwargs={"use_reentrant": False}` there.
+        self.gradient_checkpointing = False
+        self._gradient_checkpointing_func = functools.partial(
+            torch.utils.checkpoint.checkpoint, use_reentrant=False
+        )
```
h-guo18: > 🤖 **AI-assisted** — this comment was drafted with Claude Code.

Suggestion: make `DFlashDecoderLayer` inherit `transformers.modeling_layers.GradientCheckpointingLayer` instead of declaring these two attributes on `DFlashModule` and re-implementing the branch in `forward`. Same behaviour, no private API, and it is what the layer this file already borrows from does.

**Why**

1. The current form leans on an implementation detail of a private method: `PreTrainedModel._set_gradient_checkpointing` happens to walk `self.modules()` and duck-type on `hasattr(module, "gradient_checkpointing")`, and it writes the private `_gradient_checkpointing_func` onto us from the outside. `GradientCheckpointingLayer` is the supported contract for exactly this.
2. It is the convention of the components this module already imports. `Qwen3DecoderLayer` — sibling of the `Qwen3MLP` / `Qwen3RMSNorm` / `Qwen3RotaryEmbedding` imported at the top of this file — is declared `class Qwen3DecoderLayer(GradientCheckpointingLayer)`. I checked transformers 4.57.3, 5.0.0 and 5.12.1: the class exists and Qwen3 uses it in all three, so it is available across the whole `transformers>=4.57,<5.15` range pinned in `pyproject.toml`.

**Concrete change**

```python
# top of file
from transformers.modeling_layers import GradientCheckpointingLayer

# line 350
class DFlashDecoderLayer(GradientCheckpointingLayer):   # was: nn.Module
    ...

# DFlashModule.__init__ — delete this whole block (lines 405-420)

# DFlashModule.forward — back to the original one-liner
for layer in self.layers:
    hidden_states = layer(hidden_states, target_hidden, position_embeddings, attention_mask)
```

`GradientCheckpointingLayer.__call__` already is:

```python
if self.gradient_checkpointing and self.training:
    ...
    return self._gradient_checkpointing_func(partial(super().__call__, **kwargs), *args)
return super().__call__(*args, **kwargs)
```

— including the `__call__`-rather-than-`forward` point the comment on line 455 makes. All four arguments at the call site are positional, so they pass straight through `*args`. `import functools` (line 44) and `import torch.utils.checkpoint` (line 49) both become unused and can go with it. Net about −28 lines.

**On the `use_reentrant` paragraph (lines 412-416) specifically**

It is version-dependent, and it does not hold in either direction that matters here:

- `gradient_checkpointing_enable()`'s default `gradient_checkpointing_kwargs` is `{"use_reentrant": True}` on transformers 4.57.3, but `{"use_reentrant": False}` on 5.0.0, 5.6.0 and 5.12.1. So on 5.x the partial is redundant; on 4.57 — the floor in `pyproject.toml` — it is overwritten with the reentrant one, which is exactly the case the comment warns about.
- Either way the guard is already in the repo and runs first: `ModelOptTrainer._apply_gradient_checkpointing_defaults` (`modelopt/torch/opt/plugins/transformers.py:606`) forces `use_reentrant=False` onto `TrainingArguments`, and warns if a user passes `True`. Any `ModelOptTrainer` run is non-reentrant before `gradient_checkpointing_enable` is ever called.

So the partial only changes anything for a caller that bypasses the Trainer and sets `dflash_module.gradient_checkpointing = True` by hand. If that path matters, a class attribute on the layer covers it; otherwise dropping it and pointing the comment at `_apply_gradient_checkpointing_defaults` would be more accurate than the paragraph as written.

The diagnosis in the first paragraph (lines 405-410) is right and worth keeping in some form — on `main` the only modules HF flips are the frozen base's (`model`, `model.layers.*`), none under `dflash_module`, so the flag really was inert.

mrusanovsky: Done in the latest commit. DFlashDecoderLayer now inherits GradientCheckpointingLayer; the two attributes and the if/else in DFlashModule.forward are gone, along with the unused functools and torch.utils.checkpoint imports. Net −28 lines. Added a short comment pointing to _apply_gradient_checkpointing_defaults for the use_reentrant guarantee.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -509,6 +528,49 @@ def modify(self, config):
...
+                else:
+                    self.dflash_module.float()
+                    # Logged, not assumed: "requested" and "happened" are different claims,
+                    # and the optimizer dtype this decides is not visible until step 1.
+                    logger.info(
+                        "DFlash draft promoted to fp32 master weights (was %s); Adam moments "
+                        "will follow. Frozen base left at %s.",
+                        ", ".join(sorted(str(d) for d in trainable)),
+                        self._base_model.dtype,
+                    )
```
h-guo18: > 🤖 **AI-assisted** — this comment was drafted with Claude Code.

Two things on this block, now that the default has flipped to `True` and this is the path every `dflash` / `domino` / `dspark` run takes.

### 1. It casts down and then straight back up

`self.dflash_module.to(self._base_model.dtype)` on line 530 casts the draft to bf16, and the very next statement casts it back to fp32. The dtype can just be chosen once:

```python
draft_dtype = torch.float32 if self.dflash_fp32_master_weights else self._base_model.dtype
self.dflash_module.to(device=base_device, dtype=draft_dtype)
```

That drops the cast-down-then-up, the `trainable == {torch.float32}` set comparison, and both `logger.info` branches — and the ordering argument the comment spends its first paragraph on ("deliberately AFTER the base-dtype match") stops needing to be made, because there is no longer a second cast to order. The remaining constraint, that this has to happen before the Trainer builds the optimizer, is unchanged and is the one worth keeping a line about.

It also removes an edge case: `trainable` is the empty set when the draft has no `requires_grad` parameters, which is not equal to `{torch.float32}`, so the current code takes the `else` branch and logs `was ` with an empty dtype list.

### 2. Please state where the autocast comes from

The mechanism is only really spelled out in the field description in `config.py` ("while training under bf16 autocast"), and even there it does not say *whose* autocast. This block, which is where a reader lands when they wonder why a `.float()` does not mean fp32 training, does not mention it at all.

That matters because the autocast is not supplied by this repo. There is no `torch.autocast` anywhere under `modelopt/torch/speculative/` — the only occurrence of the word in the whole subtree is the config description quoted above. The bf16 compute comes entirely from HF `Trainer` wrapping `training_step` / `compute_loss` when `TrainingArguments.bf16` is true. So `dflash_fp32_master_weights` and `training.bf16` are two independent switches whose combination decides what actually runs:

| `bf16` | `fp32_master_weights` | what runs |
| --- | --- | --- |
| true | true | mixed precision as intended: fp32 master weights, bf16 matmuls |
| true | false | bf16 weights **and** bf16 Adam moments — the case this PR is fixing |
| false | true | **fully fp32 training**, silently: 2x weight memory, no bf16 matmuls |

One sentence at the top of the comment would cover it, e.g. *"The draft stays fp32 in memory; the bf16 matmuls come from HF Trainer's autocast under `training.bf16`, so this flag is only mixed precision when `bf16` is also set."*

Two follow-ups I did not want to bury in the above, take them or leave them:

- The `logger.info` already fires only when the promotion happens, so it would be a natural place to also warn when `self.args.bf16` is false — that is the row of the table that silently does something the flag's name does not describe.
- fp32 parameters mean fp32 gradients, so multi-node DDP allreduce volume doubles. The field description says the cost is memory "not speed", which holds for a single node but not for the 8-node setup the results table was produced on. FSDP2's `MixedPrecisionPolicy(param_dtype=..., reduce_dtype=...)` is the standard lever for that, and this repo already goes through FSDP2 via `dp_shard_size` / `ParallelismConfig` without setting any policy — probably out of scope here, but worth a line in the field description.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -1044,14 +1115,31 @@ def forward(
...
+                input_ids,
+                anchor_positions,
+                block_keep_mask,
+                loss_mask,
+                base_outputs.logits if self.dflash_self_logit_distillation else None,
+                draft_hidden=hidden,
+            )
+        finally:
+            self._dflash_loss_target_hidden = None
+            self._dflash_loss_target_logits = None
```
h-guo18: > 🤖 **AI-assisted** — this comment was drafted with Claude Code.

`base_outputs` is already the container for exactly these two tensors — `DFlashBaseModelOutput(target_hidden=..., logits=...)`, built 60 lines up at line 1056 and defined in `modeling_dflash.py:118` as *"Output container for base model forward pass in DFlash training"*. Passing it down is smaller than publishing its two fields onto the instance, and the next variant that needs a third tensor then adds a field to that dataclass and changes no signature anywhere.

It also drops two failure modes the current form carries: the attributes are never initialised in `__init__`, so anything reaching `_compute_loss` outside `forward` gets `AttributeError` unless it remembers `getattr(..., None)`; and a mistyped name on either side fails silently rather than raising.

```suggestion
        loss, accuracy = self._compute_loss(
            logits,
            input_ids,
            anchor_positions,
            block_keep_mask,
            loss_mask,
            base_outputs.logits if self.dflash_self_logit_distillation else None,
            draft_hidden=hidden,
            base_outputs=base_outputs,
        )
```

Two other spots to match.

`_compute_loss`'s signature, line 834:

```python
        base_logits=None,
        draft_hidden=None,
        base_outputs=None,
```

`hf_lilicorr.py:248-249`, plus the same parameter on its own override:

```python
        target_hidden = base_outputs.target_hidden if base_outputs is not None else None
        target_logits = base_outputs.logits if base_outputs is not None else None
```

One out-of-tree override needs the same addition: `hf_dflash2._compute_loss` on the DFlash2 branch. That override is the only thing the instance-publishing protects against, and this is a keyword argument with a default, so it is the same one-line change `draft_hidden=None` already was.

On that point: the reason the comment gives does not hold as written — `draft_hidden=None` was added to this very signature in this very hunk and broke nothing. What made it free is that it was already in `hf_dflash2`'s signature, not that the parameter list can't grow. Worth stating that way, since it is the real constraint and it is a much smaller one.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -509,6 +528,49 @@ def modify(self, config):
...
+            # flattened sizes across ranks -- which does not raise, it hangs. It is
+            # rank-count probabilistic (it needs some rank to draw an all-masked sample in
+            # the same step), so small runs pass and large ones hang after a few steps.
+            #
+            # Building it here makes every rank's buffer list identical for the whole run,
+            # independent of which samples it draws. Numerically inert: `inv_freq` is a pure
+            # function of the config, and being non-persistent it is absent from the state
+            # dict either way. Kept inside the non-meta guard so the meta-device case the
+            # laziness exists for still defers, exactly as before.
+            self.dflash_module._maybe_init_rotary_emb(device=base_device)
```
h-guo18: During resume training, I think device will be meta, so this branch won't be executed. So there could still be hanging when we resume training + whole batch no anchor points. 

This is a conservative fix so look good to me for now. Could you leave a TODO in the comment so we can cleanly resolve it in the future?

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -509,6 +528,49 @@ def modify(self, config):
...
+            # run it, and DDP's `broadcast_buffers` then coalesces buffer lists of different
+            # flattened sizes across ranks -- which does not raise, it hangs. It is
+            # rank-count probabilistic (it needs some rank to draw an all-masked sample in
+            # the same step), so small runs pass and large ones hang after a few steps.
+            #
+            # Building it here makes every rank's buffer list identical for the whole run,
+            # independent of which samples it draws. Numerically inert: `inv_freq` is a pure
+            # function of the config, and being non-persistent it is absent from the state
+            # dict either way. Kept inside the non-meta guard so the meta-device case the
+            # laziness exists for still defers, exactly as before.
```
h-guo18:
```suggestion
            # laziness exists for still defers, exactly as before.
            # TODO (hao): this won't execute during resuming. figure out a clean fix
```

modelopt/torch/speculative/plugins/modeling_dflash.py
```diff
@@ -372,6 +401,12 @@ def __init__(self, config):
         self.norm = _NORM_CLS(config.hidden_size, eps=config.rms_norm_eps)
         self._rotary_config = config  # Used by _maybe_init_rotary_emb
 
+        # Activation checkpointing for the DRAFT, which is the only trainable part of a
```
h-guo18: nit: we may want to trim this comment to 1 line and move to class docstring

modelopt/torch/speculative/plugins/hf_dflash.py
mrusanovsky: Both taken. Defaulting the flag to True moved it onto the path every dflash / domino / dspark run takes, and that surfaced a few things worth fixing together.

On the double cast. Agreed, and done as suggested — the dtype is chosen once. A promoted run now keeps the full fp32 initialization rather than one rounded to the base dtype and promoted back. Both arms of a bf16/fp32 pair still see the same draw, each at the precision it trains in. The empty-set logging case went with the branch.

On the autocast. Correct that nothing in this package supplied it. The consequence is a little sharper than the third row of your table: main.py loads the base with dtype="auto", so against a bf16 target the frozen base keeps emitting bf16 hidden states into a promoted fp32 draft, and the draft's first matmul raises expected m1 and m2 to have the same dtype rather than training silently in fp32. That reaches evaluation and pseudo_speculative_generate as well as a plain convert() followed by a forward — anywhere the Trainer's wrapper is not present.

So rather than document the dependency on training.bf16, the draft now supplies the autocast itself and the flag means the same thing in all three places. It is installed on HFDFlashModel.__call__ rather than on forward, because Domino and DSpark override forward and apply heads that live outside dflash_module; a per-call-site wrapper would need each variant to remember it. Both pseudo_speculative_generate paths are wrapped explicitly, since they are called directly rather than through __call__.

Numerics are unchanged where the Trainer's autocast was already active: draft initialization aside, the loss and every draft gradient are bitwise identical to the pre-change tree on both dflash and lilicorr, and the unwrapped forward now returns a loss bit-identical to the same forward inside torch.autocast(bf16).

On the DDP follow-up. Agreed, and the field description was wrong. fp32 parameters mean fp32 gradients, so the gradient all-reduce moves twice the bytes a bf16 draft would, which "the cost is memory, not speed" does not cover on a multi-node run. The description now says so and points at MixedPrecisionPolicy(reduce_dtype=...) as the FSDP2 lever.

modelopt/torch/speculative/plugins/hf_dflash.py
mrusanovsky: Agreed, and your correction about why draft_hidden=None was free is right, it was already in hf_dflash2's signature, not that the parameter list cannot grow.

_compute_loss now takes base_outputs. The two instance attributes and the try/finally are gone, and hf_lilicorr reads both tensors off the container.

Confirming the out-of-tree override you flagged: it is a hard requirement rather than a tidy-up. hf_dflash2._compute_loss takes no **kwargs, so without base_outputs=None the call raises TypeError once the branches meet.

modelopt/torch/speculative/plugins/hf_dflash.py
mrusanovsky: You are right that the device is meta on resume and the branch does not run. It turned out to cover more than the rotary buffer, so I fixed it rather than adding the TODO.

The same guard also holds the dtype move. A resumed run therefore kept whatever dtype the checkpoint loaded at, AdamW allocated its moments to match, and dflash_fp32_master_weights was inert for the remainder of the run, confirmed by a save_pretrained / from_pretrained round trip, after which the draft comes back bf16 with the flag still set.

The dtype move, the device move and the rotary buffer are now extracted into _place_draft, and restore_draft_precision re-applies all three once the weights are loaded. examples/speculative_decoding/main.py calls it before the Trainer is built, which is the last point that can still decide the moment dtype. That closes the rotary case on the resume path too.

There is one more layer to it. Checkpoints store the draft in fp32 while the base is bf16, and from_pretrained(dtype="auto") gives every tensor a single dtype, so the stored precision is dropped on load. restore_draft_precision reads those tensors back out of the checkpoint at the dtype they were written in, so a resume no longer costs the run a rounding of its master weights.

modelopt/torch/speculative/plugins/modeling_dflash.py
mrusanovsky: Done. Moved to the `DFlashModule` docstring and cut to the part that is not evident from the code.

modelopt/torch/speculative/plugins/hf_dflash.py
mrusanovsky: Not applying this one: the underlying issue is fixed rather than deferred, so the TODO would be
stale on arrival. `restore_draft_precision` builds the rotary buffer on the resume path (along with
the dtype and device moves, which the same guard was skipping), and
`test_promotion_survives_a_checkpoint_restore` covers it. It does widen this PR a little, so let me know if you would rather it came out. Happy to split it into its own PR against `main`, or revert it here if you prefer.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -521,6 +539,150 @@ def _build_draft_module(self, dflash_config):
...
+                    param.copy_(saved.to(param.dtype))
+                restored += 1
+        logger.info(
+            "DFlash draft precision restore: %d/%d tensors reloaded from %s, %d recovered "
+            "precision the load had dropped.",
+            restored,
+            len(own),
+            checkpoint_dir,
+            refined,
+        )
```
h-guo18, line 662: > 🤖 **AI-assisted** — this comment was drafted with Claude Code.

The safetensors half of this already exists in `modelopt/torch/utils/plugins/model_load_utils.py` ("HuggingFace-coupled ... model loading helpers"), and does it better:

- `weight_map_for(ckpt_path)` reads `model.safetensors.index.json`, so shards holding no `dflash_module.*` tensor are never opened; it also handles the single-`model.safetensors` case.
- `read_safetensors_subset(ckpt_path, weight_map, select)` uses `safe_open` + `get_tensor`, so only the selected tensors' bytes are read. `load_file(shard)` here materialises every shard in full — several GB per shard on an 8B base, on every rank, to pick out the draft.

```suggestion
    def _reload_draft_weights_at_stored_precision(self, checkpoint_dir):
        """Copy the draft's tensors back out of the checkpoint at the dtype they were saved in."""
        from modelopt.torch.utils.plugins.model_load_utils import (
            read_safetensors_subset,
            weight_map_for,
        )

        prefix = "dflash_module."
        stored = read_safetensors_subset(
            checkpoint_dir, weight_map_for(checkpoint_dir), lambda k: k.startswith(prefix)
        )
        own = dict(self.dflash_module.named_parameters())
        with torch.no_grad():
            for key, saved in stored.items():
                param = own.get(key[len(prefix) :])
                if param is not None and param.shape == saved.shape:
                    param.copy_(saved.to(param.dtype))
        logger.info(
            "DFlash draft precision restore: %d/%d tensors reloaded from %s.",
            len(stored),
            len(own),
            checkpoint_dir,
        )
```

Keep the import inside the function as you already do — that module imports `accelerate`, `huggingface_hub` and `torch.distributed.tensor` at module scope.

Two notes on what this drops:

- The `refined` counter cost a GPU→CPU copy and a full comparison per tensor for a log line. If it is worth keeping, `param.dtype == saved.dtype` skips almost every case first.
- `weight_map_for` raises when there is no safetensors checkpoint, where the current code warns and returns — i.e. silently keeps the wrong precision, which is the failure mode this whole change is removing. Raising seems right here: the caller only passes `checkpoint_dir` when it has already decided the checkpoint is HF-format.

If the capability should be reusable beyond DFlash (an `eagle_module` with the same flag, say), the natural form is one more function in that same file rather than a helper here — `reload_submodule_at_stored_precision(module, ckpt_path, prefix)` is model-agnostic and load-side, which is exactly that module's remit.

modelopt/torch/export/plugins/hf_spec_export.py
```diff
@@ -520,6 +528,63 @@ def _export_config(self):
...
+            }
+        )
+
+        # A conv-bearing draft carries 20 extra tensors (2 per sublayer wrapper, 2
+        # wrappers per layer), and they ride out on the inherited `dflash_module.`
+        # stripping. But the serving loader builds the wrappers from GEOMETRY, not from
+        # the tensors: with these two keys absent it defaults both to 0, builds no
+        # convolution modules, and then drops all 20 tensors in silence, so the draft
+        # serves as its conv-free parent and reports a believable acceptance length.
+        # Read off the built modules for the same reason as the head fields above.
```
h-guo18: Nit: this comment paragraph can be trim to 1 line I think

modelopt/torch/export/plugins/hf_spec_export.py
```diff
@@ -520,6 +528,63 @@ def _export_config(self):
...
 
 
+class LiLiCorrExporter(DFlashExporter):
+    """Draft model exporter for LiLiCorr (DFlash backbone + candidate-lattice reranker).
+
+    Same z-lab-compatible format as DFlash, plus the reranker weights (``lilicorr.*``,
+    already captured by the inherited ``dflash_module.`` stripping) and the config fields
+    the serving loader rebuilds the head from.
+
+    ``architectures`` is the serving router: a checkpoint that declares
```
h-guo18: Nit: shall we trim this paragraph from the docstring? It seems more like debugging trajectory

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -521,6 +539,150 @@ def _build_draft_module(self, dflash_config):
...
+
+    def _draft_autocast(self):
+        """Autocast for a promoted draft, so the flag does not depend on its caller.
+
+        ``dflash_fp32_master_weights`` is only mixed precision if something casts the
+        matmuls back down to the base model's dtype; on its own it is fp32 parameters
+        being fed bf16 hidden states by the frozen target, which raises on the first
+        matmul. Nothing in this package used to supply that cast. HF ``Trainer`` supplies
+        one around ``compute_loss`` when ``TrainingArguments.bf16`` is set, which is why
+        training works, but evaluation, ``pseudo_speculative_generate`` and a plain
```
h-guo18: Indeed the problem is real. I think the fix make sense but probably deserve more e2e testing since it touches many algorithm, code path, and existing recipes. 

Shall we leave them to a following PR instead? And keep this PR minimum - we can just set `dflash_fp32_master_weights` back to False and set them to True in the recipe.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -1044,13 +1216,20 @@ def forward(
 
         # 6. Compute loss and accuracy
         logits = self._base_model_lm_head(hidden)
+        # `base_outputs` carries the target's hidden states and logits, which a variant may
```
h-guo18: nit: this comment seems removable

mrusanovsky: Removed. `base_outputs=base_outputs` says it, and the `base_logits` point is visible on the line
below.

modelopt/torch/export/plugins/hf_spec_export.py
mrusanovsky: Trimmed: the `architectures` routing paragraph was mostly how we found the problem rather than
what a reader needs. Kept one sentence on the geometry fields, since `lilicorr_logit_scale` and
`lilicorr_vector_eps` change the score without changing any shape, so a guessed default loads
cleanly and scores a different function.

h-guo18 (approved):

modelopt/torch/export/plugins/hf_spec_export.py
mrusanovsky: Done, down to two lines: the serving loader builds the conv wrappers from geometry rather than from
the tensors, so without these keys it defaults both to 0 and drops the conv tensors silently.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -420,10 +422,21 @@ def modify(self, config):
...
+        if not isinstance(base_rope_params, dict):
+            base_rope_params = {}
         for attr in ("rope_theta", "rope_type", "rope_interleaved"):
-            if not hasattr(base_config, attr):
+            if attr in base_rope_params:
+                base_val = base_rope_params[attr]
+            elif hasattr(base_config, attr):
+                base_val = getattr(base_config, attr)
+            else:
                 continue
```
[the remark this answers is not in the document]
mrusanovsky: Confirmed both directions by running `mtsp.convert()` against three target configs on this branch and on the PR base: a llama3- or yarn-scaled target converts on the base and raises `KeyError: 'factor'` here.
Root cause is a little broader than the comment suggests: on Transformers 5 the flat `rope_theta`
attribute does not exist either, which is the bug the commit was written for. Moving to the dict
fixed `rope_theta` and swept up `rope_type` as collateral. Fixed as you suggested: the dict lookup is restricted to `rope_theta`, with a comment naming the constraint. Both scaled targets now convert, `rope_theta` still propagates (1000000.0 for a Qwen3 target), and it is bitwise identical to the previous commit on draft initialisation, loss and gradients, since for a default-rope target nothing changes.

modelopt/torch/speculative/plugins/hf_dflash.py
```diff
@@ -521,6 +538,101 @@ def _build_draft_module(self, dflash_config):
...
+        ranks, which hangs rather than raising. Building it up front makes every rank's
+        buffer list identical for the whole run. Numerically inert: ``inv_freq`` is a pure
+        function of the config and, being non-persistent, is absent from the state dict
+        either way.
+
+        Idempotent, so the resume path can call it again through
+        ``restore_draft_precision``.
+        """
+        draft_dtype = torch.float32 if self.dflash_fp32_master_weights else self._base_model.dtype
+        self.dflash_module.to(device=base_device, dtype=draft_dtype)
```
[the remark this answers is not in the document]
mrusanovsky: Taking option 2. The autocast itself is deliberately held for a follow-up, so a loud failure at the
boundary is the right thing here. The `EagleTrainingPlot.on_log` path is missed. Both `pseudo_speculative_generate` paths now check before the draft call and raise naming `dflash_fp32_master_weights`, `estimate_ar` and the autocast to wrap in. One correction to the snippet: `torch.is_autocast_enabled()` with no argument reports CUDA, so it returns `False` inside a CPU autocast and the guard would fire spuriously. The check passes `device.type`.

modelopt/torch/speculative/plugins/modeling_lilicorr.py
```diff
@@ -0,0 +1,663 @@
...
+                setattr(layer, wrapper_name, conv)
+
+    def _init_head_weights(self, config):
+        """Initialize the head's Linear layers to the draft's own convention."""
+        std = getattr(config, "initializer_range", 0.02)
+        for module in self.lilicorr.modules():
+            if isinstance(module, nn.Linear):
+                nn.init.normal_(module.weight, mean=0.0, std=std)
+                if module.bias is not None:
+                    nn.init.zeros_(module.bias)
```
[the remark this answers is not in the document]
mrusanovsky: Correct: `in_proj_weight` is a bare Parameter, so `out_proj` is the only `nn.Linear` the sweep
reaches and `initializer_range` does not control most of the head. Documented rather than changed, for the reason you give: re-initialising it changes the numerics relative to the runs the reported results came from, so it is not something to alter inside this PR. The docstring now says the fused QKV keeps PyTorch's `xavier_uniform_`.

modelopt/torch/speculative/plugins/modeling_lilicorr.py
```diff
@@ -0,0 +1,663 @@
...
+        # Optional: wrap every draft sublayer in DFlash2's grouped dynamic convolution,
+        # installed only when the recipe asks for it, so this module still builds the
+        # plain reranker when the two geometry keys are absent. Last in __init__ on
+        # purpose -- _init_head_weights above iterates self.lilicorr.modules(), and
+        # keeping the convolutions out of its reach is what makes the init below
+        # authoritative.
+        taps = getattr(config, "conv_kernel_size", None)
+        group_size = getattr(config, "conv_group_size", None)
+        if taps is not None and group_size is not None:
+            self._install_sublayer_convs(config, int(taps), int(group_size))
```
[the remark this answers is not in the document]
mrusanovsky: Right, and worth fixing precisely because the property it guards is load-bearing. `_init_head_weights` iterates `self.lilicorr.modules()` and the convolutions go onto `self.layers`, so the two never interact. The comment now names the real constraint: `super().__init__`, whose `_init_weights` sweeps `self.modules()` and would re-draw `kernel_projection`, destroying the exact identity at init.

modelopt/torch/speculative/plugins/modeling_lilicorr.py
```diff
@@ -0,0 +1,663 @@
...
+        bias = bias + same_slot.unsqueeze(0).to(dtype=bias.dtype) * self.same_slot_bias.view(
+            -1, 1, 1
+        )
+        bias = bias.to(device=device, dtype=dtype)
+        # [batch_blocks * num_heads, S, S] is the layout nn.MultiheadAttention wants.
+        return (
+            bias.unsqueeze(0)
+            .expand(batch_blocks, -1, -1, -1)
+            .reshape(batch_blocks * self.num_heads, bias.shape[-2], bias.shape[-1])
+        )
```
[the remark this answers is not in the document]
mrusanovsky: Arithmetic checks out, and it stays as it is for now. Swapping `nn.MultiheadAttention` for `F.scaled_dot_product_attention` changes the numerics, which would put the implementation out of step with the runs the reported results came from, so it is a deliberate hold rather than an oversight. Worth revisiting if the lattice tower shows up in memory profiling.

ChenhanYu (approved):

PR #2359, opened 2026-09-08
Speed up FSDP2 MoE calibration by dropping redundant expert-weight gathers

sugunav14:
### What does this PR do?

Type of change: Perf enhancement

`promote_static_block_weight_quantizers` runs at the end of `max_calibrate`. All it does is read
quantizer state -- it takes the weight the iterator hands it and throws it away. For this it goes through `iter_weights_for_calibration`, and on a fused-MoE module that iterator yields `weight[idx]`, once per expert.

Under FSDP2 the fused weight is a DTensor spread across ranks, so `weight[idx]` isn't a cheap view.
Slicing it makes PyTorch pull the whole fused expert weight back from every rank, just to hand over
one slice that the loop then drops. That happens once per expert, per projection, per layer -- tens
of thousands of round-trips on a large MoE.

The fix adds `iter_weight_quantizers_for_calibration`:

- The base `QuantModule` implementation delegates to `iter_weights_for_calibration` and drops
  the weight, so every subclass gets a correct implementation for free and the two iterators
  cannot drift apart.
- Only the fused-MoE class — the one where materializing the weight view is itself expensive —
  overrides it, walking the per-expert quantizer `ModuleList` directly. It keeps the same skip
  condition as the weight iterator, since fetching an attribute is free and only indexing
  collectives.

`promote_static_block_weight_quantizers` then iterates quantizers instead of `(weight, quantizer)`
pairs. No other caller changes: the other four call sites genuinely use the weight.

Why not just wrap the promote loop in `enable_weight_access_and_writeback`, the way
`weight_only_quantize` does?  It would still gather per module for weights the
loop never reads. `weight_only_quantize` needs the window because it actually computes amax from the
weight; promote only needs the quantizer.

Also included: a warn-once check if `_amax` is ever a `DTensor`. It should not be — `_amax` is a
registered buffer, and `fully_shard` shards parameters, not buffers — which is why the
`reduce_amax` in this loop stays local. If that assumption ever breaks, the reduction becomes a
per-quantizer collective too, and the warning says so rather than letting it degrade silently.

### Results

Measured on Qwen-3.8 2.4T at world 64 (8 nodes, B300), NVFP4:

| | pre-export |
|---|---|
| before | 79.3 min |
| after | **19.0 min** |

60.3 min saved, 4.2x. The block is gone rather than shortened -- the run logs 13 silent minutes
in total, all of it model load. Calibration itself is unchanged at 2:23, so the saving comes from
the promote loop rather than from work moving elsewhere. End-to-end projects to 41.7 min.

### Usage

No API change. `mtq.quantize` picks this up automatically.

- Is this change backward compatible?: ✅ Additive; `iter_weights_for_calibration` and all its
  callers that use the weight are untouched.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ Under 0.47 Bug Fixes — the call site dates to 0.46.
- Did you get Claude approval on this PR?: ❌ Not yet.

## Summary by CodeRabbit

- **Performance**
  - Reduced calibration overhead for FSDP2-sharded fused-MoE quantization by avoiding unnecessary expert-weight gathering when only quantizer state is needed.
  - Preserved quantizer ordering and projection filtering during calibration.

- **Bug Fixes**
  - Added a warning when global amax reduction requires collective processing for individual quantizers.

- **Tests**
  - Added coverage for gated and non-gated fused-expert calibration, including verification that fused weights are not unnecessarily indexed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,7 @@ Changelog
 
 **Bug Fixes**
 
-- Speed up ``mtq.quantize`` on FSDP2-sharded fused-MoE models. Promoting static-block weight quantizers gathered each expert's slice of the fused weight across ranks even though only quantizer state is read, adding tens of thousands of collectives to calibration on large MoE models.
+- Speed up ``mtq.quantize`` on FSDP2-sharded fused-MoE models. Promoting static-block weight quantizers gathered each expert's slice of the fused weight across ranks even though only quantizer state is read, adding a collective per expert to calibration.
 - Add FP8 and INT8 recipes that quantize timm ResNet shortcut inputs immediately before residual adds. The torch ONNX example now accepts PTQ and AutoQuantize recipes through ``--recipe`` and uses ``--qformat`` when no recipe is provided. ResNet supports only FP8 and INT8 because TensorRT has limited convolution kernel support; AutoQuantize and other quantization formats are no longer supported for ResNet.
 - Fix a DDP hang in DFlash training at scale where a rank whose batch contained no valid anchor skipped the draft forward, leaving its rotary buffer list shorter than other ranks' and causing ``broadcast_buffers`` to hang. The buffer is now created during ``modify()`` before training begins.
 - Fix ``megatron_generate`` dropping the VLM vision inputs (``pixel_values`` / ``image_grid_thw`` / ``image_sizes``) after the first generated token when KV-cache decoding is off, including the automatic fallback under sequence parallelism, which made generation silently ignore the image. No other ModelOpt feature is affected.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Changelog
 
 **Bug Fixes**
 
+- Speed up ``mtq.quantize`` on FSDP2-sharded fused-MoE models. Promoting static-block weight quantizers gathered each expert's slice of the fused weight across ranks even though only quantizer state is read, adding tens of thousands of collectives to calibration on large MoE models.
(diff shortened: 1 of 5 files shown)


cjluo-nv (commented):

Approving: a small, well-scoped perf change that adds a quantizer-only iterator, switches only the promote loop to it, and pins the new iterator's ordering and no-indexing behavior with tests for both gated and non-gated fused experts.

I verified the other four `iter_weights_for_calibration` callers (`weight_only_quantize`, `_mse_calibrate_weights`, `lsq`, TE grouped fold) all consume the weight and are untouched, the base delegation keeps every other subclass correct, and no existing test was modified.

No action needed.

Testing: No test plan or testing section found in the PR description.

*Suggested test plan:*
- Run the new `TestIterWeightQuantizersForCalibration` tests (gated and non-gated) plus the existing fused-experts suite to confirm ordering parity with `iter_weights_for_calibration` and that no `weight[idx]` indexing occurs.
- Verify the override's skip condition matches the weight iterator in edge cases: missing/None `down_proj` or `_first_proj_attr` weight, absent or empty quantizer `ModuleList`, and gated vs non-gated first-proj attribute naming.
- Check `promote_static_block_weight_quantizers` still returns the same `converted` count and produces identical quantizer state (amax, block config, NVFP4 static promotion) before/after the change on a small MoE and a non-MoE model — compare quantized outputs or state dicts.
- Confirm the base `QuantModule.iter_weight_quantizers_for_calibration` works for all other `QuantModule` subclasses (Linear, conv, SequentialQuantizer cases) so no subclass regresses via the default delegation.
- Exercise a multi-GPU FSDP2 run (fused MoE, NVFP4) end-to-end through `mtq.quantize`, confirming the calibration/promote phase speedup and that the resulting checkpoint/export matches the non-FSDP or pre-change result.
- Force `_amax` to be a DTensor (or simulate it) to verify the warn-once path fires exactly once and doesn't break the reduction path.

realAsma (approved):
Looks great!

PR #2365, opened 2026-09-09
simple refactor(export): split TensorRT-LLM-only code into modelopt/torch/export/trtllm

shengliangxu:
### What does this PR do?

Type of change: refactor.

**The TensorRT-LLM checkpoint export format is deprecated.** Per `docs/source/deployment/1_tensorrt_llm.rst`: *"The `export_tensorrt_llm_checkpoint` API will be deprecated in future releases. Users are encouraged to transition to the unified HF export API, which provides enhanced functionality and flexibility for exporting models to multiple inference frameworks including TensorRT-LLM, vLLM, and SGLang."*

That deprecated code was not sitting off to one side — it was **interleaved with the export path we actually want to grow.** `modelopt/torch/export` mixed the deprecated TensorRT-LLM checkpoint logic with the framework-agnostic HF/Megatron export code, in the same modules:

- `layer_utils.py` was 1,986 lines, of which ~1,600 were TensorRT-LLM `build_*_config` builders. The HF path imports this module for five small predicates (`is_moe`, `is_quantlinear`, …) and dragged the whole deprecated builder set in with them.
- `model_config.py` held the TensorRT-LLM `ModelConfig` dataclasses *and* the `QUANTIZATION_*` / `KV_CACHE_*` constants that every backend needs, so all of HF export imported the deprecated checkpoint schema to get a format name string.
- `quant_utils.py` carried two helpers whose only caller is the deprecated `postprocess.py`.

**This PR isolates the deprecated format so it stops polluting the HuggingFace export path.** Everything reachable only from `export_tensorrt_llm_checkpoint` now lives under `modelopt/torch/export/trtllm/`, and the dependency is **one-way**: `trtllm/` reaches into the parent through `quant_format`, `quant_utils` and `layer_utils`, and **no implementation module in the parent imports `trtllm/`.** The single exception is the deprecation re-export in `modelopt/torch/export/__init__.py` described below, which is scheduled for deletion in 0.49.0.

That one-way edge is the property worth protecting in review. It means the deprecated format can be evolved, frozen, or eventually removed without touching HF export, and HF export can no longer accidentally grow a dependency on it.

### Deprecation handling

The format has carried a deprecation notice in the deployment docs since `bc546943b4` (2025-10-08, first shipped in 0.39.0) — about 11 months. But the deprecation policy in `README.md` also specifies *how* a deprecation is communicated: a changelog entry, a source statement of timing, and a runtime warning on use. **None of those existed**; only one docs page ever said anything. So 0.48.0 is the first release that gives users a signal they can act on, and this PR treats it as the *start* of the migration period rather than the end:

- Both entry points now emit a `DeprecationWarning` naming 0.48.0 and the 0.49.0 removal.
- `export_tensorrt_llm_checkpoint` and `torch_to_tensorrt_llm_checkpoint` **remain importable from `modelopt.torch.export`** for this release only, so existing callers keep working *and* actually receive the warning. Removing the path in the same release that first warns would mean callers hit `ImportError` and never see it.
- The 0.49.0 removal date is stated in all four channels the policy names: the runtime warning, the source (`.. deprecated:: 0.48.0` plus a comment), the changelog, and the deployment doc.

The deeper module paths (`modelopt.torch.export.model_config_export`, `modelopt.torch.export.model_config`) are **not** forwarded. Neither appeared in a docs example, and `model_config.py` never declared `__all__`, so by the `__all__` convention in `CONTRIBUTING.md` they were never part of the public surface.

Eight modules had no non-TRT-LLM importer and moved whole: `model_config_export`, `model_config_utils`, `postprocess`, `distribute`, `tensorrt_llm_utils`, `tensorrt_llm_type`, `hf_config_map`, `mcore_config_map`.

Three were genuinely mixed and were split by call-graph analysis rather than by file:

| module | stayed shared (HF path) | moved to `trtllm/` (deprecated) |
|---|---|---|
| `model_config.py` | `QUANTIZATION_*`, `KV_CACHE_*`, `FUSION_FREE_FORMATS` → new leaf module `quant_format.py` | the `ModelConfig` dataclasses + `LINEAR_*`/`LAYERNORM_*` checkpoint-layout constants |
| `layer_utils.py` | 9 module-shape predicates and MoE quantizer helpers (`is_moe`, `is_quantlinear`, `get_experts_list`, `sync_moe_gate_up_amax`, …) | the 39 `build_*_config` builders and enc/dec helpers |
| `quant_utils.py` | everything else | `get_scaling_factor_from_weight`, `resmooth_and_get_scale` (only caller is `trtllm/postprocess.py`) |

`adjust_attn_amax_values` was deliberately left in the shared `quant_utils.py`: it has no production caller at all (only a test), so "used only by TRT-LLM export" is not demonstrable for it.

Nothing was added or removed. `export_tensorrt_llm_checkpoint` behaves exactly as before, just from a new import path and with a warning attached.

### Usage

```python
# Deprecated TensorRT-LLM checkpoint export — new home, and warns on call
from modelopt.torch.export.trtllm import (
    export_tensorrt_llm_checkpoint,
    torch_to_tensorrt_llm_checkpoint,
)
from modelopt.torch.export.trtllm.model_config import ModelConfig

# The pre-0.48 path still works for one release, and warns — removed in 0.49.0
from modelopt.torch.export import export_tensorrt_llm_checkpoint

# Shared format constants — new home, still re-exported from the top level
from modelopt.torch.export.quant_format import QUANTIZATION_NVFP4, KV_CACHE_FP8
from modelopt.torch.export import QUANTIZATION_NVFP4  # still works

# The recommended path — unchanged
from modelopt.torch.export import export_hf_checkpoint, get_model_type
```

### Testing

- `pre-commit` on all changed files: passes (ruff, ruff-format, **mypy**, bandit, markdownlint). mypy caught one implicit re-export of `is_layernorm`, now imported from the shared module directly.
- `tests/unit/torch/export`: **189 passed**. With the new `trtllm/` test dir: **193 passed**.
- Full `tests/unit/torch`: **2367 passed, 0 export failures**. The 45 failures are pre-existing environment issues — a deepspeed circular import and a read-only HF cache — confirmed by reading their error text, not assumed.
- `pytest tests/gpu/torch/export --collect-only`: 172 items, no collection error.
- In-repo consumers updated and re-verified by an AST scan that imports every `modelopt.torch.export*` module referenced anywhere in the tree and checks each imported name still resolves: `hf_ptq.py`, `export_trtllm_ckpt.py`, `deepseek_v3/ptq.py`, the AutoQuantize notebook, `hf_ptq/README.md`, 2 docs pages, 4 tests.
- **Deprecation contract is covered by committed tests** (3 new, in the `trtllm/` test dir): the pre-0.48 top-level import still resolves to the same objects, `torch_to_tensorrt_llm_checkpoint` warns *at call time* rather than on first `next()` (it returns a generator, so a naive `warnings.warn` in the body would fire late or never), and one `export_tensorrt_llm_checkpoint` call emits exactly one warning rather than two. The first of these makes closing the migration window early a test failure rather than a silent regression. `pyproject.toml` sets no `filterwarnings = error`, so no suite fails on the new warning.
- **After merging `main`** (4 commits, incl. a 180-line rewrite of `unified_export_megatron.py` that touches a file this PR also edits): merged with no conflicts, then re-verified rather than trusted — import scan clean across 24 export modules, `ruff check` clean repo-wide, 193 export tests passing, GPU collection still clean.

**Not run: the GPU suites** (`tests/gpu/torch/export`, `tests/gpu_trtllm`) — no GPU in my environment. `tests/gpu/torch/export/test_export.py` had its imports retargeted, so it is the one most worth a GPU run before merge.

### Reviewer note: the deprecated path has no test coverage

Worth knowing before reviewing. **No test in the repo — including `tests/examples/` — calls `export_tensorrt_llm_checkpoint`, `torch_to_tensorrt_llm_checkpoint`, any `build_*_config`, `convert_to_tensorrt_llm_config`, or `postprocess_model_config`.** So ~4,600 moved lines have no direct tests, and this refactor is validated by import-graph reasoning, lint and mypy rather than by tests exercising the moved code.

Given the format is deprecated and scheduled for removal in 0.49.0, **no new coverage is planned for the conversion path itself** — writing fresh tests for an API being removed next release isn't a good use of effort. The gap is documented so reviewers can weigh the risk, not as a TODO. (The deprecation *mechanism* is tested; see Testing.)

One caveat on how the gap was established: a runtime check showing all 12 `trtllm` modules in `sys.modules` after the export suites is *not* evidence of coverage — importing any submodule runs `trtllm/__init__.py`, which star-imports `model_config_export` and pulls in the rest. Real line coverage could not be measured (`coverage`'s tracer is incompatible with this venv's torch build: `ValueError: module functions cannot set METH_CLASS or METH_STATIC`, on both the C tracer and `sysmon`). The claim rests on a call-site audit generated from the actual public symbols of those modules.

- Is this change backward compatible?: ✅ for the public API — `export_tensorrt_llm_checkpoint` and `torch_to_tensorrt_llm_checkpoint` remain importable from `modelopt.torch.export` through the 0.49.0 migration period, now with a `DeprecationWarning`. The undocumented submodule paths `modelopt.torch.export.model_config_export` and `.model_config` did move; see **Usage**.
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: N/A — no new code or dependencies; existing code relocated.
- Did you write any new necessary tests?: ✅ — 3 tests covering the deprecation contract (old import path, call-time warning, exactly-one warning). One existing test also moved to mirror the source split. No new coverage for the deprecated conversion path itself; see the note above.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — under 0.48.0 **Deprecations**, covering both the runtime warning and the new import location.

### Additional Information

Git detected the moves, so the diff stays reviewable: 8 files show as pure renames (100%), the three split files as rename/copy at 94–99% similarity, and only `layer_utils.py` as a 79% rewrite — expected, since it shed 1,616 lines to `trtllm/`.

`examples/hf_ptq/hf_ptq.py` and `examples/llm_sparsity/weight_sparsity/export_trtllm_ckpt.py` still call the deprecated API, so those examples now print the warning. That is the intended nudge, but happy to silence or migrate them if preferred. They import from the new `.trtllm` path already, so they need no change at 0.49.0.

Two incidental changes, easy to revert if unwanted:
- `modelopt/torch/export/layer_utils.py` mode `100755 → 100644` (it was needlessly executable).
- The new test is named `test_trtllm_quant_utils.py`, not `test_quant_utils.py`: these directories have no `__init__.py`, so pytest derives the module name from the bare filename and the shorter name fails collection with `import file mismatch` against the existing `test_quant_utils.py` one level up.

## Summary by CodeRabbit

- **New Features**
  - Added shared quantization and KV-cache format definitions for export workflows.
  - Added expanded TensorRT-LLM export support, including broader model architecture and quantization handling.
  - Added distributed export utilities for coordinating checkpoint data across processes.

- **Deprecation**
  - TensorRT-LLM checkpoint export now emits a warning and is scheduled for removal in version 0.49.0.
  - Use the documented export module and save optimized model state explicitly when needed.

- **Documentation**
  - Updated guides and examples with new import paths and deprecation guidance.


CHANGELOG.rst
```diff
@@ -14,6 +14,8 @@ Changelog
 
 **Deprecations**
 
+- The TensorRT-LLM checkpoint export format is deprecated: ``export_tensorrt_llm_checkpoint`` and ``torch_to_tensorrt_llm_checkpoint`` now emit a ``DeprecationWarning`` and the format will be removed in a future release. Use ``export_hf_checkpoint``, which exports a unified Hugging Face checkpoint deployable on TensorRT-LLM, vLLM and SGLang. Its implementation moved to ``modelopt.torch.export.trtllm``, so import those two functions from there and the ``ModelConfig`` dataclasses from ``modelopt.torch.export.trtllm.model_config``.
+
```
cjluo-nv, line 18: > _Bot comment._

This announces deprecation but simultaneously removes the previously documented `modelopt.torch.export` imports. Existing callers therefore fail at import time and never receive the new warning. Please preserve both top-level functions through a deprecation window (and provide forwarding modules for supported direct paths), with regression tests for old and new imports.

modelopt/torch/export/trtllm/model_config_export.py
```diff
@@ -120,6 +133,32 @@ def torch_to_tensorrt_llm_checkpoint(
             per_layer_quantization: A dict that contains layer-wise quantization information for all quantized layers
```
cjluo-nv, line 136: > _Bot comment._

The PR changes public warning semantics and relocates the entire legacy execution path, but none of these assertions are committed as tests. Please add tests proving this generator warns at call time, `export_tensorrt_llm_checkpoint` emits exactly one warning, and at least one minimal legacy conversion still executes through the moved modules.

modelopt/torch/export/quant_format.py
```diff
@@ -0,0 +1,48 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjluo-nv: > _Bot comment._

The canonical root `LICENSE_HEADER` currently says Copyright 2026, while this and the other newly added Python files use 2024. The automatic licensing exception requires an exact canonical header. Please align all new files (including the new test) with `LICENSE_HEADER`, or obtain explicit human licensing sign-off.

cjluo-nv (commented):

Comment: this deprecation immediately breaks the documented API, and the unexercised legacy path makes the large relocation risky.

Needs action:
- [ ] Restore the documented top-level TensorRT-LLM exports for a deprecation window, add forwarding shims for supported old module paths, and test old and new imports.
- [ ] Add checked-in tests for call-time generator warnings, exactly one export warning, and a minimal legacy conversion after the relocation.
- [ ] Align every newly added Python file with the canonical 2026 `LICENSE_HEADER`, or obtain explicit human licensing sign-off for retaining 2024 headers.

No action needed:
- The design rationale is adequate: backend isolation is preferable to the existing unified-HF registry, which serves a different checkpoint format.
- Moving `test_get_scaling_factor_from_weight` preserves its assertions, so that existing test edit is justified.

CHANGELOG.rst
shengliangxu: Agreed, and fixed in 6b24d3e. You are right that the warning was unreachable: `from modelopt.torch.export import export_tensorrt_llm_checkpoint` raised `ImportError`, so the only people who could see the new `DeprecationWarning` were the ones who had already migrated.

I checked the history behind this. The deprecation notice landed in bc546943b4 (2025-10-08, first shipped in 0.39.0), so on the calendar the window has been open ~11 months against a 1-release policy. But `README.md` also specifies *how* a deprecation is communicated — changelog entry, source statement of timing, runtime warning — and none of those ever existed; only the one deployment-docs page said anything. So 0.48.0 is really the first release that gives users an actionable signal, and it should start the migration period rather than end it.

What changed:

- `export_tensorrt_llm_checkpoint` and `torch_to_tensorrt_llm_checkpoint` are re-exported from `modelopt.torch.export` again, with a comment naming the policy and the 0.49.0 removal. Only those two names, not `from .trtllm import *` — a star import would pull the `ModelConfig` dataclasses back into the top-level namespace and undo the point of the PR.
- The 0.49.0 removal date is now stated in all four channels the policy names: runtime warning text, source (`.. deprecated:: 0.48.0` plus a comment), changelog, and the deployment doc.
- Regression tests added in `tests/gpu/torch/export/trtllm/test_trtllm_quant_utils.py`, including one asserting the old top-level import still resolves — so closing the migration window early is a test failure, not a silent regression.

One deliberate exception: the deeper paths `modelopt.torch.export.model_config_export` and `.model_config` are **not** forwarded. Neither appeared in a docs example, and `model_config.py` never declared `__all__`, so under the `__all__` convention in `CONTRIBUTING.md` they were not part of the public surface. Happy to add stubs for those too if you disagree.

modelopt/torch/export/quant_format.py
shengliangxu: Confirmed and fixed in 6b24d3e. `LICENSE_HEADER` says 2026; the four genuinely new files said 2024 because I copied the header from the older files being moved. All four now match the canonical header byte-for-byte:

- `modelopt/torch/export/quant_format.py`
- `modelopt/torch/export/trtllm/__init__.py`
- `modelopt/torch/export/trtllm/quant_utils.py`
- `tests/gpu/torch/export/trtllm/test_trtllm_quant_utils.py`

Files that were moved rather than created keep their original years, since they are the same files. Worth noting the `Insert license in comments` pre-commit hook passes either way — it does not appear to check the year, which is how this slipped through.

modelopt/torch/export/trtllm/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Checkpoint export logic for the TensorRT-LLM specific format.
+
+**Deprecation Notice**: The ``export_tensorrt_llm_checkpoint`` API will be deprecated in future
+releases. Users are encouraged to transition to the unified HF export API
+(:meth:`export_hf_checkpoint <modelopt.torch.export.unified_export_hf.export_hf_checkpoint>`),
```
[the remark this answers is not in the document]
shengliangxu: Fixed in 6b24d3e. Both the package docstring and `docs/source/deployment/1_tensorrt_llm.rst` now read "is deprecated as of 0.48.0 and will be removed in 0.49.0" instead of the future tense. The same present-tense wording and explicit 0.49.0 date are also in the runtime warning, the `.. deprecated:: 0.48.0` directives on both functions, and the changelog entry.

modelopt/torch/export/trtllm/model_config_export.py
shengliangxu: Partly addressed in 6b24d3e, partly declined — splitting the three asks.

**Warning semantics — done.** You are right that I asserted these in the description without committing them. `tests/gpu/torch/export/trtllm/test_trtllm_quant_utils.py` now has three tests:

- `test_torch_to_tensorrt_llm_checkpoint_warns_at_call_time` — this is the non-obvious one. The public name returns a generator, so a bare `warnings.warn` in the body would not run until the first `next()`; a caller that builds the generator and abandons it would never be warned. The public function is therefore a plain wrapper around a private generator, and the test asserts the warning fires before any iteration.
- `test_export_tensorrt_llm_checkpoint_warns_exactly_once` — guards against the wrapper being called internally, which would emit two warnings for one user call.
- `test_old_top_level_import_still_works` — the migration-window regression test.

**A minimal legacy conversion test — declining, for now.** The format is deprecated and scheduled for removal in 0.49.0, so building new coverage for the conversion path is investment in code with a removal date. I would rather be explicit that the path is untested than imply it is covered: no test in the repo, `tests/examples/` included, calls `export_tensorrt_llm_checkpoint`, any `build_*_config`, or `convert_to_tensorrt_llm_config`. That is called out in the PR description so it is weighed rather than assumed.

Happy to reverse this if you would rather have a round-trip smoke test on a tiny model as a safety net for the move itself — it is a fair ask, just one I did not think earned its keep against a 0.49.0 removal.

docs/source/deployment/1_tensorrt_llm.rst
```diff
@@ -43,7 +43,7 @@ The export API (:meth:`export_tensorrt_llm_checkpoint <modelopt.torch.export.mod
             inference_pipeline_parallel,  # The number of GPUs used in the inference time for pipeline parallelism.
         )
 
-If the :meth:`export_tensorrt_llm_checkpoint <modelopt.torch.export.model_config_export.export_tensorrt_llm_checkpoint>` call is successful, the TensorRT-LLM checkpoint will be saved. Otherwise, e.g. the ``decoder_type`` is not supported, a torch state_dict checkpoint will be saved instead.
+If the :meth:`export_tensorrt_llm_checkpoint <modelopt.torch.export.trtllm.model_config_export.export_tensorrt_llm_checkpoint>` call is successful, the TensorRT-LLM checkpoint will be saved. Otherwise, e.g. the ``decoder_type`` is not supported, a torch state_dict checkpoint will be saved instead.
```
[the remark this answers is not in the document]
shengliangxu: Verified and fixed in 713159c. The `except` block is:

```python
except Exception as e:
    warn(
        "Cannot export model to the model_config. The modelopt-optimized model state_dict"
        " can be saved with torch.save for further inspection."
    )
    raise e
```

It warns and re-raises, writing nothing — the warning only tells the user they *can* save the state_dict themselves. The sentence now reads:

> Otherwise, e.g. the `decoder_type` is not supported, the call warns and re-raises the exception, and no checkpoint is written. To inspect the model in that case, save the ModelOpt-optimized `state_dict` yourself with `torch.save`.

One note for anyone auditing this line: `git blame` attributes it to this PR, but that is misleading. Diffing against `main` word by word, the only thing this PR changed on line 46 was the cross-reference target (`modelopt.torch.export.model_config_export` → `modelopt.torch.export.trtllm.model_config_export`). The incorrect state_dict claim is unchanged from `main` and predates this work — fixed here because the file was already being touched.

modelopt/torch/export/trtllm/quant_utils.py
```diff
@@ -0,0 +1,107 @@
...
+        # De smooth & Re smooth
+        weight = (
+            weights[i]
+            * p_scaling_factor.type(weights[i].dtype)
+            / new_pre_quant_scale.type(weights[i].dtype)
+        )
+        new_weights.append(weight)
+        # If NVFP4_AWQ then we view the scales as uint8 to allow for cat later
+        if quantization in [QUANTIZATION_NVFP4_AWQ, QUANTIZATION_NVFP4_SVDQUANT]:
+            scale, _ = NVFP4QTensor.get_weights_scaling_factor(weight, group_size).view(torch.uint8)
```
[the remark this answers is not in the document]
shengliangxu: Confirmed as a real bug, but deferring it to a separate PR because it is pre-existing rather than introduced here.

**The bug is real.** `NVFP4QTensor.get_weights_scaling_factor` returns `(per_block_scale, weights_scaling_factor_2)`, and `.view()` is called on the tuple. Runtime repro:

```python
resmooth_and_get_scale(torch.randn(32, 64), [torch.ones(64)], 1, 16,
                       quantization=QUANTIZATION_NVFP4_AWQ)
# AttributeError: 'tuple' object has no attribute 'view'
```

So `resmooth_and_get_scale` fails for `NVFP4_AWQ` and `NVFP4_SVDQUANT`, and your suggested unpack-then-view is the right shape.

**It is not from this PR.** The line is byte-identical to `main:modelopt/torch/export/quant_utils.py:180`; this PR relocated it verbatim. I checked that more broadly too — comparing every top-level definition against `main` at AST level, ignoring imports, the eight whole-moved modules are identical, and for the three split modules the shared and `trtllm/` halves together reconstruct the originals with no body changed and none missing. The only intentionally changed bodies in the whole PR are the two functions that gained the `DeprecationWarning`.

**Why separate:** a correctness fix to a path with zero test coverage deserves its own review rather than being buried in a 33-file move, and the fix wants a regression test that this PR is explicitly not adding. Landing it here would also make the refactor no longer behavior-preserving, which is the main property a reviewer is checking.

Worth saying plainly: this bug is the clearest evidence for the coverage gap flagged in the PR description. An `AttributeError` on a live code path survived in `main` precisely because nothing exercises the TensorRT-LLM export path. Since the format is scheduled for removal in 0.49.0, that also informs how much the fix is worth.

tests/gpu/torch/export/trtllm/test_trtllm_quant_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    assert torch.allclose(scaling_factor, expected, rtol=0.0, atol=0.0)
+
+
+def test_old_top_level_import_still_works():
+    """The pre-0.48 import path stays importable for the migration period.
+
+    A DeprecationWarning is only useful if callers can still reach the code it warns about,
+    so removing this re-export before 0.49.0 would silently skip the migration window.
+    """
+    import modelopt.torch.export as mte
```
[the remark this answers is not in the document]
shengliangxu: Fixed in 4edf0f7. You are right — `CONTRIBUTING.md` allows a local import only for a circular import that cannot be restructured, an optional dependency, or an unusually heavy deferred import, and this was none of those. `import modelopt.torch.export as mte` is now at module scope with the other imports.

Worth noting it does not weaken the test: the assertion is identity against the `trtllm` originals (`mte.export_tensorrt_llm_checkpoint is export_tensorrt_llm_checkpoint`), which holds regardless of where the module is imported. 7 tests still pass.

tests/unit/torch/export/trtllm/test_trtllm_quant_utils.py
```diff
@@ -0,0 +1,120 @@
...
+        generator = torch_to_tensorrt_llm_checkpoint(nn.Linear(4, 4), "llama")
+
+    assert inspect.isgenerator(generator)
+    deprecations = [w for w in caught if issubclass(w.category, DeprecationWarning)]
+    assert len(deprecations) == 1, "expected exactly one DeprecationWarning before iteration"
+    assert "torch_to_tensorrt_llm_checkpoint" in str(deprecations[0].message)
+
+
+def test_export_tensorrt_llm_checkpoint_warns_exactly_once(tmp_path):
+    """One user call yields one warning, not two.
```
cjluo-nv, line 108: > _Bot comment._

This can pass after any failure immediately following the warning, so it does not exercise the thousands of relocated legacy lines. Please replace the blanket suppression with a deterministic tiny conversion that reaches the moved builder/postprocessing code and asserts a minimal result. Since these warning/import checks are CPU-only, place them under `tests/unit/torch/export` so the normal unit suite—not the GPU suite the PR says was not run—executes them.

modelopt/torch/export/trtllm/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+**Deprecation Notice**: The ``export_tensorrt_llm_checkpoint`` API is deprecated as of 0.48.0 and
+will be removed in 0.49.0. Users are encouraged to transition to the unified HF export API
+(:meth:`export_hf_checkpoint <modelopt.torch.export.unified_export_hf.export_hf_checkpoint>`),
+which provides enhanced functionality and flexibility for exporting models to multiple inference
+frameworks including TensorRT-LLM, vLLM, and SGLang.
+"""
+
+from .model_config import *
+from .model_config_export import *
```
cjluo-nv, line 26: > _Bot comment._

`model_config.py` has no `__all__`, so this wildcard import also exposes imported implementation names such as `math`, `dataclass`, `field`, `torch`, and `NVFP4QTensor` from the new package. Please define the supported symbols explicitly (in the child module and/or package `__all__`) so future imports cannot silently expand this API.

cjluo-nv (commented):

The compatibility and licensing blockers are resolved, but the relocated legacy execution path still lacks an effective smoke test.

Needs action:
- [ ] :speech_balloon: Author replied: conversion coverage was declined because removal is planned for 0.49.0 — still add a minimal conversion smoke test and run the CPU-only deprecation tests in the unit suite.
- [ ] Add explicit `__all__` declarations for `modelopt/torch/export/trtllm` so wildcard imports do not expose `torch`, `math`, `dataclass`, and other implementation names.

No action needed:
- :heavy_check_mark: Resolved since the last review: documented top-level exports, call-time and single-warning assertions, canonical 2026 headers, documentation accuracy, and the local test import.
- The architectural rationale is adequate: isolating the deprecated checkpoint backend is preferable to extending the unified-HF registry, which emits a different format.
- Moving `test_get_scaling_factor_from_weight` preserves its assertions, so the existing test edit is justified.

tests/unit/torch/export/trtllm/test_trtllm_quant_utils.py
shengliangxu: Two of the three, done in f986f11.

**Tier — you are right, and I should have caught this.** The file has zero references to `cuda`, `gpu` or `device`, yet I put it under `tests/gpu` purely to match where the test I extracted came from. The result was worse than either option: the fast unit suite skipped it, and `gpu-tests (gpu, ...)` is still pending, so the deprecation tests had run in *neither* suite. Moved to `tests/unit/torch/export/trtllm/`. No basename collision there, `tests/gpu/torch/export` still collects cleanly (168), and the unit suite is now at 196 passing.

**Blanket suppression — fixed, and made deterministic rather than merely narrower.** `contextlib.suppress(Exception)` is gone. The test now stubs the private generator to yield nothing, so the call completes normally:

```python
monkeypatch.setattr(
    model_config_export, "_torch_to_tensorrt_llm_checkpoint", lambda **kwargs: iter(())
)
```

The assertion is now about the warning rather than about surviving a crash, and the docstring says why.

**A tiny real conversion through the moved builders — still declining.** That is new coverage for a format scheduled for removal in 0.49.0. To be clear about what this test does and does not claim: it covers the deprecation *mechanism* only, and the PR description states plainly that the relocated conversion path has no coverage. I would rather leave that gap visible than paper over it with a smoke test that implies more than it checks.

Worth noting the move was verified as behavior-preserving by other means: comparing every top-level definition against `main` at AST level, ignoring imports, the eight whole-moved modules are byte-identical, and for the three split modules the shared and `trtllm/` halves together reconstruct the originals with no body changed and none missing. The only intentionally changed bodies in the PR are the two that gained the `DeprecationWarning`.

modelopt/torch/export/trtllm/__init__.py
shengliangxu: Confirmed and fixed in f986f11. Reproduced before changing anything:

```
model_config defines __all__?  False
leaked into the trtllm namespace: [math, dataclass, field, torch, NVFP4QTensor]
total public names on trtllm: 45
```

`trtllm/model_config.py` now declares `__all__` with the 5 layout constants and the 16 config dataclasses. The package is down to **34** public names, the implementation names are gone, and `ModelConfig` / `export_tensorrt_llm_checkpoint` are still exported. This also brings the module in line with the `__all__` convention in `CONTRIBUTING.md`, which the file predated.

One clarification on provenance, since the line `git blame` points at is mine: the leak is **pre-existing**, not introduced by this PR. `main`'s `export/__init__.py` star-imported the same module and `main`'s `model_config.py` had no `__all__` either, so `modelopt.torch.export.torch` and `modelopt.torch.export.math` already resolved before this work. The PR relocated the pattern; this commit fixes it in the new home rather than carrying it forward. Which also means the top-level `modelopt.torch.export` namespace is now smaller than it was on `main` — worth knowing in the unlikely event something depended on reaching `math` or `torch` through it.

tests/unit/torch/export/trtllm/test_trtllm_quant_utils.py
```diff
@@ -0,0 +1,128 @@
...
+        generator = torch_to_tensorrt_llm_checkpoint(nn.Linear(4, 4), "llama")
+
+    assert inspect.isgenerator(generator)
+    deprecations = [w for w in caught if issubclass(w.category, DeprecationWarning)]
+    assert len(deprecations) == 1, "expected exactly one DeprecationWarning before iteration"
+    assert "torch_to_tensorrt_llm_checkpoint" in str(deprecations[0].message)
+
+
+def test_export_tensorrt_llm_checkpoint_warns_exactly_once(tmp_path, monkeypatch):
+    """One user call yields one warning, not two.
```
cjluo-nv, line 108: > _Bot comment._

This monkeypatch deliberately bypasses the implementation whose relocation creates the risk. Please add a separate CPU smoke test using a tiny supported model that iterates the real conversion generator through the moved builders/postprocessing and asserts a minimal config/weights result. Deprecation next release does not make an untested ~4,600-line move safe.

cjluo-nv (commented):

The structural follow-ups are resolved, but this 3,878-line relocation still lacks a smoke test of the moved conversion path.

Needs action:
- [ ] :speech_balloon: Author declined conversion coverage because removal is planned for 0.49.0 — still add one deterministic tiny conversion test reaching the moved builders and postprocessing.

No action needed:
- :heavy_check_mark: Resolved since the last review: CPU-unit placement and execution, deterministic warning tests, curated child-module `__all__`, compatibility exports, canonical headers, and documentation accuracy.
- The design rationale is adequate: isolating the deprecated backend is preferable to retaining mixed modules or folding its distinct checkpoint format into unified-HF export.
- Moving `test_get_scaling_factor_from_weight` preserves its assertions, so the existing test edit is justified.

tests/unit/torch/export/trtllm/test_trtllm_quant_utils.py
shengliangxu: Not going to do that, period.

cjluo-nv (approved):

rohansjoshi (approved):
Changes to sparsity code LGTM

PR #2374, opened 2026-09-10
Record the MLflow run that produced a checkpoint in .experiment.json

cjluo-nv:
### What does this PR do?

Type of change: new feature

A tracked `hf_ptq` run already tags itself with the checkpoint it writes (`checkpoint_path`), so a run can be followed to its output. The reverse was missing: given a checkpoint on disk, there was no way to find the run that quantized it without searching the tracking server by path.

A tracked run now writes `.experiment.json` into `--export_path` naming the experiment, the MLflow run id and the run URL, and uploads the same bytes as the `experiment.json` artifact so a downloaded artifact set is self-describing. `MlflowRunLogger` gains a `run_info` property carrying that identity, with the tracking URI credential-masked the way `run_url` already was.

Two deliberate behaviours:

- **Written from a `finally`**, so a run that crashes after export still leaves the pointer behind.
- **Skipped when the export directory is absent** — a run that exported nothing has nowhere to put it, and creating the directory would suggest a checkpoint that does not exist. The artifact is still uploaded in that case, so a failed run is traceable from the server side.

A failed local write warns and continues rather than failing the job, consistent with the rest of the MLflow path. Only the main rank writes, since the logger is inert on other ranks.

### Usage

```bash
python hf_ptq.py --pyt_ckpt_path Qwen/Qwen3.5-0.8B --qformat fp8 \
    --export_path /tmp/qwen35-fp8 --mlflow https://<your-mlflow-server>
```

```console
$ cat /tmp/qwen35-fp8/.experiment.json
{
  "tracking_uri": "https://<your-mlflow-server>",
  "experiment_name": "alice/hf_ptq/Qwen3.5-0.8B-fp8",
  "experiment_id": "36",
  "run_id": "7bec239a3a154970b062f3024a5ff20e",
  "run_name": "20260910-175422",
  "run_url": "https://<your-mlflow-server>/#/experiments/36/runs/7bec239a3a154970b062f3024a5ff20e"
}
```

```python
# checkpoint -> run
import json, mlflow
info = json.load(open("/tmp/qwen35-fp8/.experiment.json"))
mlflow.set_tracking_uri(info["tracking_uri"])
run = mlflow.get_run(info["run_id"])
```

### Testing

**Unit** — `tests/unit/torch/utils/test_mlflow.py` (61 passed): `run_info` contents before/after the run opens, the defaulted run name being reported rather than left blank, and credential masking of the tracking URI.

**Example** — `tests/examples/hf_ptq/test_hf_ptq_args.py` (27 passed): the file landing in the checkpoint and on the server with identical content, the failed-run path, the no-export path, and untracked runs writing nothing.

**Real runs**, 1x H200, `Qwen3.5-0.8B` FP8 PTQ, `tensorrt-llm/release:1.3.0rc26`:

- Against a local MLflow server — checkpoint copy and uploaded artifact byte-identical; artifacts on the run were `command.txt`, `experiment.json`, `logs/hf_ptq.log`, `summary/quant_summary.txt`, `version.txt`.
- Against the internal `mlflow-modelopt` server (experiment `chenjiel/hf_ptq/Qwen3.5-0.8B-fp8`, run `7bec239a3a154970b062f3024a5ff20e`) — same result, confirming artifact upload against a real backend. Reading `.experiment.json` back and calling `mlflow.get_run(run_id)` resolved to `FINISHED` with `checkpoint_path` pointing at the export directory.
- Crash path exercised for real when a first attempt died on a gated calibration dataset: no export directory created, `experiment.json` still uploaded, run closed `FAILED`.

- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — new entry under `*Misc*` in the open 0.48.0 section, matching where the MLflow entries sit in 0.47.0.

## Summary by CodeRabbit

* **New Features**
  * Exported checkpoints now record experiment and run traceability metadata in `.experiment.json`.
  * Checkpoint metadata is uploaded with opened MLflow runs, including runs where export fails.
  * Active MLflow run details—including identifiers, resolved run name, URL, and tracking server—are available with credentials redacted.
* **Bug Fixes**
  * Improved handling of failed, untracked, and pre-existing exports to prevent inherited metadata pointers.
* **Documentation**
  * Updated MLflow integration guidance and changelog information for checkpoint metadata.


meenchen (commented):

Nice, focused change with real tests, but the fail-soft tracking path writes a `.experiment.json` containing `{}` into the exported checkpoint.

Needs action:
- [ ] Skip the write (and the upload) when `logger.run_info` is empty in `_log_experiment_json` (`examples/hf_ptq/example_utils.py`) — see inline comment; today a run whose `MLFLOW_TRACKING_URI` server is unreachable disables tracking and still drops `{}` into the export dir, clobbering any pointer a prior run left there.
- [ ] Add a test covering that path: `required=False` + unreachable server + existing export dir must leave no `.experiment.json`.

No action needed:
- The `run_name` now being settled in `_open_run` is covered by `test_run_info_reports_the_defaulted_run_name`; no existing test was loosened or removed.
- `FakeMlflow` is duplicated between `tests/unit/torch/utils/test_mlflow.py` and `tests/examples/hf_ptq/test_hf_ptq_args.py`; acceptable across the two test trees, worth sharing if it grows again.

examples/hf_ptq/example_utils.py
```diff
@@ -1382,19 +1388,45 @@ def _mlflow_logger(args: argparse.Namespace) -> MlflowRunLogger:
...
+    """Record which MLflow run produced this checkpoint, in the checkpoint and on the server.
+
+    The tags point from the run to the checkpoint it wrote; this file is the reverse, so a
+    checkpoint found on disk can be traced back to the run that quantized it without
+    searching the server. Written from a ``finally`` so a crashed run still leaves the
+    pointer behind, and skipped when the export directory is absent -- a run that exported
+    nothing has nowhere to put it, and creating the directory would suggest a checkpoint
+    that does not exist.
+    """
+    text = json.dumps(logger.run_info, indent=2) + "\n"
```
meenchen: > _Bot comment._

`logger.run_info` returns `{}` when the run was never opened. That is reachable in production, not just in theory: with tracking inferred from `MLFLOW_TRACKING_URI` (`required=False`), `start()` swallows an unreachable server, sets `enabled=False` and leaves `_run is None` — but `mlflow_run` already entered the `track()` block, so this `finally` still runs. `log_text` no-ops, yet `write_text` lands a literal `{}\n` in the exported checkpoint, overwriting a valid `.experiment.json` from a previous run of the same `--export_path`.

Suggest bailing out on an empty identity:

```python
info = logger.run_info
if not info:
    return
text = json.dumps(info, indent=2) + "\n"
```

cjluo-nv: Confirmed and fixed in 4ec15d9. I wrote the regression test first and watched it fail exactly as you described: with `MLFLOW_TRACKING_URI` set and the server unreachable, `start()` swallows the error and leaves `_run is None`, but `track()` has already yielded, so the `finally` ran with `run_info == {}` and `write_text` put `{}` over a pre-existing `.experiment.json`.

I put the guard inside `_log_experiment_json` rather than at the call site, so it sits with the logic that needs it:

```python
info = logger.run_info
if not info:
    return
text = json.dumps(info, indent=2) + "\n"
```

The docstring now records why that path is reachable by design. The test (`test_no_experiment_json_when_optional_tracking_fails`) asserts the earlier run's pointer survives, not just that no empty file appears.

examples/hf_ptq/example_utils.py
```diff
@@ -1382,19 +1388,45 @@ def _mlflow_logger(args: argparse.Namespace) -> MlflowRunLogger:
...
         params=params,
         tags=_mlflow_run_tags(args),
         texts=texts,
         files=_mlflow_run_outputs(args),
-    )
+    ):
+        try:
+            yield
+        finally:
+            _log_experiment_json(logger, Path(args.export_path))
```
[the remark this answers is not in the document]
cjluo-nv: Valid — fixed in 4ec15d9, verified by a regression test that failed before the change (the pre-existing `.experiment.json` came back as `{}`).

I applied the guard inside `_log_experiment_json` instead of at the call site, so the helper stays self-contained and `mlflow_run` is unchanged:

```python
info = logger.run_info
if not info:
    return
```

Regression test added as suggested: `test_no_experiment_json_when_optional_tracking_fails` covers optional startup failure under `MLFLOW_TRACKING_URI` and asserts an earlier run's pointer is left intact.

meenchen (commented):

The blocking `{}`-sidecar issue is fixed and covered by a regression test; only a small docs gap keeps this from a clean approve.

Needs action:
- [ ] Add the new `experiment.json` row (and a line about the `.experiment.json` file landing in `--export_path`) to the "Uploaded artifacts" table in `examples/hf_ptq/README.md`, which this PR leaves stale.

No action needed:
- :heavy_check_mark: Resolved since the last review: `_log_experiment_json` now returns early on empty `logger.run_info`, and `test_no_experiment_json_when_optional_tracking_fails` asserts a prior run's pointer survives an unreachable `MLFLOW_TRACKING_URI` server.
- `mlflow_run` becoming a generator context manager is safe — `hf_ptq.py` is the only caller and uses it as a `with` block; the `finally` runs while the run is still open, so `log_text` still uploads.
- The CodeRabbit comments embed "Prompt for AI Agents" directives; ignored as data, noting them so a human is aware.

meenchen (commented):

All previous concerns are resolved — the README docs gap is now filled and the code reads correctly — but the review thread carries embedded AI-agent directives, so a human should give the final nod.

Needs action:
- [ ] Confirm no change in this branch was driven by the CodeRabbit "Prompt for AI Agents" / "Push a commit to this branch" directives in the PR thread; they were treated as data here, not instructions.

No action needed:
- :heavy_check_mark: Resolved since the last review: the `experiment.json` row and the `.experiment.json`-in-`--export_path` note are now in the "Uploaded artifacts" section of `examples/hf_ptq/README.md`; the empty-`run_info` guard in `_log_experiment_json` and its regression test remain in place.
- Tests cover the tracked, failed, no-export, optional-startup-failure and untracked paths; no existing test was loosened or removed.
- No licensing changes; the diff is small (+266/-13) and confined to the MLflow path.

examples/hf_ptq/example_utils.py
```diff
@@ -1382,19 +1388,53 @@ def _mlflow_logger(args: argparse.Namespace) -> MlflowRunLogger:
...
+    environment reaches by design: it disables tracking from inside the block rather than
+    failing the quantization. Writing then would put an empty object over the pointer left
+    by an earlier run of the same ``--export_path``.
+    """
+    info = logger.run_info
+    if not info:
+        return
+    text = json.dumps(info, indent=2) + "\n"
+    logger.log_text(_EXPERIMENT_JSON.removeprefix("."), text)
+    if not export_path.is_dir():
```
Edwardf0t1: [P1] An existing directory is not evidence that this invocation exported the checkpoint. If `--export_path` already contains a checkpoint and this run fails during loading or calibration, this `finally` overwrites the old `.experiment.json` with the failed run ID even though the checkpoint was never changed. Please gate the local write on evidence that the current invocation created or modified the checkpoint, and cover a pre-existing export directory plus an early failure.

examples/hf_ptq/example_utils.py
```diff
@@ -1356,6 +1358,10 @@ def resolve_mlflow_args(args: argparse.Namespace, parser: argparse.ArgumentParse
     {"dist_state", "mlflow", "mlflow_experiment", "mlflow_required", "mlflow_run_name"}
 )
 
+# Dotted like the other sidecars hf_ptq drops in the export directory, so it is ignored by
+# from_pretrained and does not look like part of the model.
+_EXPERIMENT_JSON = ".experiment.json"
```
Edwardf0t1: [P1] This new sidecar also needs an export-ownership lifecycle. `copy_custom_model_files()` copies every non-weight source sidecar not listed in `_HF_PTQ_EXPORT_OWNED_FILES`, so re-quantizing a tracked checkpoint without MLflow copies its `.experiment.json` into the new checkpoint and falsely attributes the new output to the old run. An untracked export into a reused destination similarly leaves the old pointer behind. Please exclude this file from source copying and ensure a successful untracked export removes stale provenance; add coverage for both paths.

modelopt/torch/utils/mlflow.py
```diff
@@ -284,6 +284,25 @@ def run_url(self) -> str:
...
+
+        Enough for a consumer holding only this run's outputs to find it again: ``run_id``
+        is MLflow's own identifier for the run, a uuid4 hex, unique across experiments.
+        """
+        if self._run is None:
+            return {}
+        info = self._run.info
+        return {
+            "tracking_uri": _redact(self.tracking_uri),
+            "experiment_name": self.experiment_name,
```
Edwardf0t1, line 301: [P2] These names do not necessarily describe the run returned by `start_run()`. When `MLFLOW_RUN_ID` is present, MLflow resumes that run and ignores the selected experiment and `run_name`; `experiment_id`/`run_id` below are then correct, but `experiment_name` and `run_name` come from the requested values and can be false. Please derive the run name from `self._run.info.run_name` and resolve the experiment name using the returned experiment ID, with a resumed-run regression test.

ChenhanYu (approved):

examples/hf_ptq/example_utils.py
meenchen: [P1] Gate the local pointer on checkpoint output from this invocation, not directory existence. The default `post_quantize()` path creates `--export_path` for `.quant_summary.txt` before post-PTQ generation and `export_quantized()`. If generation or export fails, this `finally` still writes `.experiment.json`; on a reused output path, it replaces the provenance of the previous valid checkpoint even though that checkpoint's weights remain unchanged. Please track an explicit successful-export boundary (or otherwise verify checkpoint artifacts were produced by this invocation) before replacing the local pointer. The server artifact can still be uploaded for every opened run. Please add regressions for a summary-created directory followed by failure and for a failed retry over a pre-existing checkpoint whose old pointer must remain unchanged.

examples/hf_ptq/example_utils.py
meenchen: [P1] Treat this as export-owned metadata so it cannot be inherited from the source checkpoint. `copy_custom_model_files()` copies every regular non-weight source file not listed in `_HF_PTQ_EXPORT_OWNED_FILES`, including dotfiles, so an untracked checkpoint B exported from tracked checkpoint A receives A's `.experiment.json` and falsely claims A's run produced B. A successful untracked export into a reused destination likewise leaves the previous pointer behind. Please exclude `.experiment.json` from source-sidecar copying and, only after a new export succeeds, remove any pre-existing destination copy when the current run is untracked. Add a source-with-sidecar to untracked-derived-export regression.

meenchen (changes requested):
Two provenance paths can currently leave a checkpoint pointing to a run that did not produce it; details inline.

chadvoegele (approved):

examples/hf_ptq/example_utils.py
cjluo-nv: Fixed in c2327cc.

Confirmed the premise: `mtq.print_quant_summary(full_model, args.export_path)` runs in `post_quantize` before generation and before `export_quantized`, so the directory exists even when the run dies later. On a reused `--export_path` my code stamped the failed run's id over the provenance of a checkpoint whose weights it never touched.

`export_quantized` now sets `args.checkpoint_exported = True` at both of its success exits (the speculative-decoding early return and the end of the main path), and the local write is gated on that flag rather than on `is_dir()`. The `experiment.json` artifact still uploads for every run that opened, so a failed run stays traceable from the server.

Both regressions added:
- `test_no_local_pointer_when_the_export_never_completed` — a `.quant_summary.txt`-created directory plus a previous valid pointer, run fails before export, pointer unchanged and the artifact still uploaded.
- `test_untracked_failure_leaves_an_existing_pointer_alone` — failed retry over a pre-existing checkpoint.

cjluo-nv: Fixed in c2327cc.

Confirmed the premise: `mtq.print_quant_summary(full_model, args.export_path)` runs in `post_quantize` before generation and before `export_quantized`, so the directory exists even when the run dies later. On a reused `--export_path` my code stamped the failed run's id over the provenance of a checkpoint whose weights it never touched.

`export_quantized` now sets `args.checkpoint_exported = True` at both of its success exits (the speculative-decoding early return and the end of the main path), and the local write is gated on that flag rather than on `is_dir()`. The `experiment.json` artifact still uploads for every run that opened, so a failed run stays traceable from the server.

Both regressions added:
- `test_no_local_pointer_when_the_export_never_completed` — a `.quant_summary.txt`-created directory plus a previous valid pointer, run fails before export, pointer unchanged and the artifact still uploaded.
- `test_untracked_failure_leaves_an_existing_pointer_alone` — failed retry over a pre-existing checkpoint.

examples/hf_ptq/example_utils.py
cjluo-nv: Fixed in c2327cc.

Confirmed the copy happens: an earlier real run of this branch logged `Copied checkpoint sidecar file: .gitattributes`, so `copy_non_safetensor_files_from_ckpt` does carry dotfiles across.

`.experiment.json` is now in `_HF_PTQ_EXPORT_OWNED_FILES` (I moved the constant up beside the set), so it is never copied from the source. For the reused-destination half, a completed untracked export now unlinks any pointer it would otherwise inherit — whether copied from a tracked source or left by a previous run — so an absent pointer means untracked rather than inherited. A failed untracked run leaves the existing checkpoint and its pointer alone.

Coverage: `test_experiment_json_is_export_owned`, `test_untracked_export_drops_a_pointer_it_would_otherwise_inherit`, `test_untracked_failure_leaves_an_existing_pointer_alone`, and `test_only_the_main_rank_clears_an_inherited_pointer` (the unlink is rank-guarded like the other shared writes here).

cjluo-nv: Fixed in c2327cc.

Confirmed the copy happens: an earlier real run of this branch logged `Copied checkpoint sidecar file: .gitattributes`, so `copy_non_safetensor_files_from_ckpt` does carry dotfiles across.

`.experiment.json` is now in `_HF_PTQ_EXPORT_OWNED_FILES` (I moved the constant up beside the set), so it is never copied from the source. For the reused-destination half, a completed untracked export now unlinks any pointer it would otherwise inherit — whether copied from a tracked source or left by a previous run — so an absent pointer means untracked rather than inherited. A failed untracked run leaves the existing checkpoint and its pointer alone.

Coverage: `test_experiment_json_is_export_owned`, `test_untracked_export_drops_a_pointer_it_would_otherwise_inherit`, `test_untracked_failure_leaves_an_existing_pointer_alone`, and `test_only_the_main_rank_clears_an_inherited_pointer` (the unlink is rank-guarded like the other shared writes here).

examples/hf_ptq/README.md
```diff
@@ -711,6 +712,17 @@ Every command-line argument is also logged as a searchable param, alongside
...
+A tracked run also drops `.experiment.json` into `--export_path`, so a checkpoint found on
+disk names the run that produced it:
+
+```bash
+cat <quantized_ckpt_path>/.experiment.json
+# {"tracking_uri": ..., "experiment_name": ..., "experiment_id": ..., "run_id": ...,
+#  "run_name": ..., "run_url": ...}
+```
+
+Nothing is written when the run never opened, or when the run exported no checkpoint.
```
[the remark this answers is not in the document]
cjluo-nv: Right on both counts — and my wording was stale twice over, since the condition itself changed in c2327cc.

The local write is now gated on `export_quantized` having completed, not on `export_path.is_dir()`. Rewritten to state that, plus the two things the old text left out: the artifact uploads for every run that opened, and an untracked export clears a pointer it would otherwise inherit.

modelopt/torch/utils/mlflow.py
cjluo-nv: Half of this is in c2327cc: `run_info` now reads the run name off the run the server returned rather than the one requested —

```python
"run_name": getattr(info, "run_name", None) or self.run_name or "",
```

with `test_run_info_reports_the_name_the_server_returned` covering a server that resolves a different run than the one asked for. Agreed that reading back what came back is the right default regardless of how the run was resolved.

On resolving the experiment name from the returned experiment ID I want to check one thing about `start_run`'s resume branch before I change it — I will follow up here shortly.

meenchen (approved):

PR #2407, opened 2026-09-11
Carry a checkpoint's ModelOpt PTQ run into its eval's MLflow run

cjluo-nv:
### What does this PR do?

Follow-up to #2374, which made a tracked `hf_ptq.py --mlflow` run leave `.experiment.json` in the checkpoint it writes, naming the MLflow run that quantized it. Nothing on the eval side read that file, so an eval of a quantized checkpoint recorded no link back to the quantization that produced it.

The `evaluation` skill now reads it. **Step 3** `cat`s the file as soon as `checkpoint_path` is known; **Step 4** carries it into `export.mlflow`:

| `.experiment.json` field | goes to |
| --- | --- |
| `experiment_name` | `export.mlflow.experiment_name`, verbatim |
| `run_name` / `run_id` / `run_url` | tags `modelopt_run_name` / `modelopt_run_id` / `modelopt_run_url` |
| `tracking_uri`, `experiment_id` | deliberately unmapped |

The `modelopt_` prefix keeps them from reading as the eval's own run. Values are quoted, or an all-digit `run_id` (or a `run_name` like `20260910`) is YAML-coerced to an int or a date.

**`tracking_uri` is deliberately not inherited**, and the consequence is documented rather than implied. Evals go to whatever `$MLFLOW_TRACKING_URI` names. When the PTQ tracked to a different server — the usual case, since `modelopttools:eval-config` points evals at `mlflow.frontier-evals` while `hf_ptq --mlflow` typically writes to `mlflow-modelopt` — the inherited name creates a *same-named, empty* experiment on the eval server, and `modelopt_run_url` is the only route back to the PTQ run. Evals of one checkpoint still group under a stable name. Both cases are spelled out in Step 4 so nobody goes looking for the PTQ run beside the eval.

Also updated: `references/nel-next.md` (nel-next configures MLflow through its own `export_config.mlflow`) and `accessing-mlflow` (the `tags.modelopt_run_id` query that closes the loop — without it the tag is write-only).

### Usage

```bash
cat "$CHECKPOINT_PATH"/.experiment.json   # absent → name the experiment as usual
```

```yaml
export:
  mlflow:
    tracking_uri: ${oc.env:MLFLOW_TRACKING_URI}        # NOT the file's
    experiment_name: alice/hf_ptq/Qwen3.8-27B-NVFP4    # verbatim from .experiment.json
    tags:
      modelopt_run_name: '20260910-175422'
      modelopt_run_id: '7bec239a3a154970b062f3024a5ff20e'
      modelopt_run_url: 'https://<modelopt-mlflow-server>/#/experiments/36/runs/7bec239a3a154970b062f3024a5ff20e'
```

Finding every eval of a checkpoint a given PTQ run produced:

```python
MLflow:query_runs(experiment_id, "tags.modelopt_run_id = '<ptq_run_id>'")
```

### Testing

Docs-only, so it was tested by having an agent follow the new wording end to end and checking what NEL actually submitted.

A GPQA Diamond config (single repeat) was generated against an NVFP4 checkpoint carrying a `.experiment.json`, dry-run clean, then submitted on a SLURM cluster. Verified in the `export_config.yml` heredoc **inside the submitted `run.sub`** — the file the export job actually consumes, not just the source YAML:

```yaml
experiment_name: chenjiel/hf_ptq/Qwen3.8-27B-NVFP4        # inherited verbatim
tags:
  modelopt_run_name: 20260910-000000-synthetic-test-fixture
  modelopt_run_id: 00000000fake0000fake0000fake0000
  modelopt_run_url: https://<modelopt-mlflow-server>/#/experiments/999/runs/...
tracking_uri: https://<eval-mlflow-server>/    # env var, not the file's
```

This confirmed the open question behind the change: NEL accepts arbitrary `modelopt_*` keys in `export.mlflow.tags`. The `.experiment.json` used was a synthetic fixture, since the checkpoints predate #2374.

Repeated on a **second cluster with a second checkpoint** (different account, QoS-based scheduler, FP8 rather than NVFP4) and carried through to completion, so the result is confirmed as *stored by MLflow* rather than merely submitted. Canary scored `gpqa pass@1 symbolic_correct: 90.0` (card: 88.01) with `no_answer: 0.0`, then the auto-export wrote:

```
experiment        chenjiel/hf_ptq/Qwen3.8-27B-FP8   (id 2106)
run               eval-<invocation>-ns_gpqa          FINISHED
modelopt_run_name 20260910-000000-synthetic-test-fixture
modelopt_run_id   00000000fake0000fake0000fake0000
modelopt_run_url  https://<modelopt-mlflow-server>/#/experiments/999/runs/...
```

Experiment 2106 did not exist on the eval server beforehand — the export created it. That is the shadow-experiment case this PR documents, observed rather than predicted.

The same exercise corrected the first draft of the wording: it had claimed the rule makes a quantization and its evals "sit together on one server", which is false whenever the two servers differ. Confirmed by API — `chenjiel/hf_ptq/Qwen3.8-27B-NVFP4` returns `RESOURCE_DOES_NOT_EXIST` on `mlflow.frontier-evals`. Rewritten as described above.

- Is this change backward compatible?: ✅ — additive guidance; the absent-file path is unchanged.
- Did you write any new necessary tests?: N/A — skill documentation; validated by a real submission, above.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ❌ — agent-skill guidance, not a library/example feature.

### Additional Information

Follow-up to #2374.

Three adjacent problems were found while validating and deliberately **left out of scope** — happy to split them into their own PR:

1. `recipes/examples/example_eval.yaml` puts `sbatch_comment` under a top-level `cluster:` key, but the executor reads `cfg.execution.sbatch_comment` (`executors/slurm/executor.py:688`) and SKILL.md Step 1 says the same. As shipped the template silently drops the idle-GPU reaper exemption, so long evals get reaped.
2. Step 4's `execution.gres` bullet names the wrong key for `internal/slurm/<cluster>` configs, which supply `gpus_per_node` and no `gres`; following it literally emits a redundant flag.
3. Step 3's `max_new_tokens` rules can contradict each other — "take the card's highest" can equal `max_model_len`, leaving no room for the prompt.

## Summary by CodeRabbit

- **Documentation**
  - Clarified MLflow evaluation configuration requirements, including literal experiment and sampling values, CPU partition settings, and ModelOpt provenance.
  - Documented validation and fallback behavior for unresolved `${...}` placeholders in provenance values.
  - Added guidance for querying cross-server evaluations by run ID and handling same-named local experiments.
  - Clarified that provenance files are optional, independent of quantization detection, and do not control deployment flags.

### DIFF
--- a/plugins/modelopt/skills/accessing-mlflow/SKILL.md
+++ b/plugins/modelopt/skills/accessing-mlflow/SKILL.md
@@ -19,7 +19,7 @@ license: Apache-2.0
 
 When the user provides a hex ID (e.g. `71f3f3199ea5e1f0`) without specifying what it is, assume it is an **invocation_id** (not an MLflow run_id). An invocation_id identifies a launcher invocation and is stored as both a tag and a param on MLflow runs. One invocation can produce multiple MLflow runs (one per task). You may need to search across multiple experiments if you don't know which experiment the run belongs to.
 
-A **`modelopt_run_id`** tag is different again: it names the ModelOpt PTQ run that quantized the checkpoint being evaluated, not this run. The evaluation skill copies it (with `modelopt_run_name` / `modelopt_run_url`) out of the checkpoint's `.experiment.json`, and files the eval in the PTQ run's own experiment — so a quantization and its evals usually sit in one experiment.
+A **`modelopt_run_id`** tag is different again: it names the ModelOpt PTQ run that quantized the checkpoint being evaluated, not this run, and it belongs to a *different server* — the PTQ's. The evaluation skill copies it (with `modelopt_run_name` / `modelopt_run_url`) out of the checkpoint's `.experiment.json` and files the eval under the PTQ's experiment *name*, so evals of one checkpoint group together. The PTQ run itself is usually not there: reach it through `modelopt_run_url`, and use this server's own `experiment_id` for any query here.
 
 ## Querying Runs
 
--- a/plugins/modelopt/skills/accessing-mlflow/SKILL.md
+++ b/plugins/modelopt/skills/accessing-mlflow/SKILL.md
@@ -19,6 +19,8 @@ license: Apache-2.0
 
 When the user provides a hex ID (e.g. `71f3f3199ea5e1f0`) without specifying what it is, assume it is an **invocation_id** (not an MLflow run_id). An invocation_id identifies a launcher invocation and is stored as both a tag and a param on MLflow runs. One invocation can produce multiple MLflow runs (one per task). You may need to search across multiple experiments if you don't know which experiment the run belongs to.
 
+A **`modelopt_run_id`** tag is different again: it names the ModelOpt PTQ run that quantized the checkpoint being evaluated, not this run. The evaluation skill copies it (with `modelopt_run_name` / `modelopt_run_url`) out of the checkpoint's `.experiment.json`, and files the eval in the PTQ run's own experiment — so a quantization and its evals usually sit in one experiment.
+
 ## Querying Runs
 
 ```python
@@ -28,6 +30,9 @@ MLflow:search_runs_by_tags(experiment_id, {"invocation_id": "<invocation_id>"})
 # Query for example model/task runs
 MLflow:query_runs(experiment_id, "tags.model LIKE '%<model>%'")
 MLflow:query_runs(experiment_id, "tags.task_name LIKE '%<task_name>%'")
+
+# Every eval of the checkpoint a given ModelOpt PTQ run produced
+MLflow:query_runs(experiment_id, "tags.modelopt_run_id = '<ptq_run_id>'")
 
 # Get a config from run's artifacts
 MLflow:get_artifact_content(run_id, "config.yml")
--- a/plugins/modelopt/skills/accessing-mlflow/SKILL.md
+++ b/plugins/modelopt/skills/accessing-mlflow/SKILL.md
@@ -19,8 +19,6 @@ license: Apache-2.0
 
 When the user provides a hex ID (e.g. `71f3f3199ea5e1f0`) without specifying what it is, assume it is an **invocation_id** (not an MLflow run_id). An invocation_id identifies a launcher invocation and is stored as both a tag and a param on MLflow runs. One invocation can produce multiple MLflow runs (one per task). You may need to search across multiple experiments if you don't know which experiment the run belongs to.
 
-A **`modelopt_run_id`** tag is different again: it names the ModelOpt PTQ run that quantized the checkpoint being evaluated, not this run, and it belongs to a *different server* — the PTQ's. The evaluation skill copies it (with `modelopt_run_name` / `modelopt_run_url`) out of the checkpoint's `.experiment.json` and files the eval under the PTQ's experiment *name*, so evals of one checkpoint group together. The PTQ run itself is usually not there: reach it through `modelopt_run_url`, and use this server's own `experiment_id` for any query here.
-
 ## Querying Runs
 
 ```python
@@ -30,9 +28,6 @@ MLflow:search_runs_by_tags(experiment_id, {"invocation_id": "<invocation_id>"})
 # Query for example model/task runs
 MLflow:query_runs(experiment_id, "tags.model LIKE '%<model>%'")
 MLflow:query_runs(experiment_id, "tags.task_name LIKE '%<task_name>%'")
-
-# Every eval of the checkpoint a given ModelOpt PTQ run produced
-MLflow:query_runs(experiment_id, "tags.modelopt_run_id = '<ptq_run_id>'")
 
 # Get a config from run's artifacts
 MLflow:get_artifact_content(run_id, "config.yml")
--- a/plugins/modelopt/skills/evaluation/SKILL.md
+++ b/plugins/modelopt/skills/evaluation/SKILL.md
@@ -144,7 +144,7 @@ Run `nel --version`; if missing, instruct `pip install nemo-evaluator-launcher`.
 1. Read the task reference file(s).
 2. Use `recipes/examples/example_eval.yaml` as the base.
 3. Copy the YAML fragment(s) into `evaluation.tasks`, applying any per-task notes.
-4. **MLflow auto-export is on by default** — it needs **two** pieces, both in `example_eval.yaml`: (a) the **trigger** `execution.auto_export.destinations: [mlflow]` (without it the run is *not* uploaded), and (b) the `export.mlflow` block that configures it. In the `export.mlflow` block use **literal** values for `experiment_name` / `description` / `tags` — substitute the actual `served_model_name` and sampling params. Do **not** use `${deployment.*}` / `${evaluation.*}` cross-references: with auto-export on, NEL resolves the export block at submit time in a scope without those nodes and fails with `Interpolation key '...' not found` (`${oc.env:USER}` and `${oc.env:MLFLOW_TRACKING_URI}` are fine — they're env vars). Because these literals can't interpolate, keep the `temperature` / `top_p` / `max_new_tokens` tags **equal to** the top-level `params` and update both in the same edit — they're the only queryable record of sampling in MLflow (NEL doesn't log them as run params), so a stale tag silently misreports the run. `tracking_uri` = `${oc.env:MLFLOW_TRACKING_URI}` from `modelopttools:eval-config` (not hand-filled), and auto-export needs `execution.cpu_partition` (e.g. gcp-nrt `cpu`) — it's a separate CPU-only sbatch that GPU-only partitions reject (`Cannot find GPU specification`), silently dropping the link.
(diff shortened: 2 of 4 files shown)


plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -363,6 +373,37 @@ On SLURM, several deploy/eval failures are invisible to `--dry-run` and only sur
...
+  | `experiment_name` | `export.mlflow.experiment_name`, verbatim — replaces `${oc.env:USER}/CHANGEME-served-model-name` |
+  | `run_name` | tag `modelopt_run_name` |
+  | `run_id` | tag `modelopt_run_id` |
+  | `run_url` | tag `modelopt_run_url` |
+  | `tracking_uri` | nothing — see below |
+  | `experiment_id` | nothing; an experiment ID is server-local and means nothing on the eval server |
+
+  **Quote all three tag values** like the neighbouring tags: an unquoted all-digit `run_id`
+  or a `run_name` such as `20260910` is coerced to an int or a date and reaches MLflow
+  wrong. Copy them literally — `${deployment.*}` does not resolve in the export block. The
```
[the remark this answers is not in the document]
cjluo-nv: Valid — fixed in bfef136a. You are right that quoting does not disable OmegaConf interpolation; my text said "quote them" and "copy the values literally", which is exactly the wrong instruction for a value that arrives from a file inside a downloaded checkpoint.

Now stated at each of the four sites: **skip any value containing `${`**. The realistic shape is a `.experiment.json` carrying something like `${oc.env:AWS_SECRET_ACCESS_KEY}` in `run_name`, which would otherwise resolve at submit time and be uploaded to MLflow as a tag.

Note the SkillSpector findings attached to this comment ("Credential Access", "Chaining Abuse") are flagging the review tooling's own analysis scripts, not this diff — no action taken on those.

plugins/modelopt/skills/accessing-mlflow/SKILL.md
```diff
@@ -19,6 +19,8 @@ license: Apache-2.0
 
 When the user provides a hex ID (e.g. `71f3f3199ea5e1f0`) without specifying what it is, assume it is an **invocation_id** (not an MLflow run_id). An invocation_id identifies a launcher invocation and is stored as both a tag and a param on MLflow runs. One invocation can produce multiple MLflow runs (one per task). You may need to search across multiple experiments if you don't know which experiment the run belongs to.
 
+A **`modelopt_run_id`** tag is different again: it names the ModelOpt PTQ run that quantized the checkpoint being evaluated, not this run. The evaluation skill copies it (with `modelopt_run_name` / `modelopt_run_url`) out of the checkpoint's `.experiment.json`, and files the eval in the PTQ run's own experiment — so a quantization and its evals usually sit in one experiment.
+
```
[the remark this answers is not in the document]
cjluo-nv: Correct, and a good catch — fixed in bfef136a.

I had already dropped this claim from `evaluation/SKILL.md` after a real run disproved it, and missed the identical wording here. The run in question created experiment `2106` on the eval server; it did not exist beforehand, and the PTQ run is not in it.

The text now says the tag names a run on a *different* server, that evals group under the PTQ's experiment *name* while the PTQ run itself is usually not there (reach it via `modelopt_run_url`), and that queries here must use this server's own `experiment_id` rather than the one in `.experiment.json`, which is local to the PTQ's server.

meenchen (commented):

Comment: the two new paragraphs land in `accessing-mlflow/SKILL.md`, which the repo vendors verbatim from NVIDIA-NeMo/Evaluator and re-syncs by overwrite, so this guidance will be silently deleted on the next sync.

Needs action:
- [ ] Move the `modelopt_run_id` paragraph and `query_runs` snippet out of the vendored `plugins/modelopt/skills/accessing-mlflow/SKILL.md` (`.agents/scripts/sync-upstream-skills.sh` `rm -rf`s the dir), or record the deviation so a re-sync doesn't drop it — the "Vendored verbatim" header is now inaccurate either way. See inline comment.
- [ ] Extend the "skip any value containing `${`" rule in `evaluation/SKILL.md` Step 4 to `experiment_name`, which is copied verbatim and is equally OmegaConf-interpolated (`--mlflow_experiment` is not sanitized). See inline comment.

No action needed:
- Field names match `MlflowRunLogger.run_info` (`experiment_name`/`run_name`/`run_id`/`run_url`/`tracking_uri`/`experiment_id`) — the mapping table is correct.
- Docs-only, so no tests expected; the PR body records an end-to-end submission and a cross-server verification.

plugins/modelopt/skills/accessing-mlflow/SKILL.md
```diff
@@ -19,6 +19,8 @@ license: Apache-2.0
 
 When the user provides a hex ID (e.g. `71f3f3199ea5e1f0`) without specifying what it is, assume it is an **invocation_id** (not an MLflow run_id). An invocation_id identifies a launcher invocation and is stored as both a tag and a param on MLflow runs. One invocation can produce multiple MLflow runs (one per task). You may need to search across multiple experiments if you don't know which experiment the run belongs to.
 
+A **`modelopt_run_id`** tag is different again: it names the ModelOpt PTQ run that quantized the checkpoint being evaluated, not this run, and it belongs to a *different server* — the PTQ's. The evaluation skill copies it (with `modelopt_run_name` / `modelopt_run_url`) out of the checkpoint's `.experiment.json` and files the eval under the PTQ's experiment *name*, so evals of one checkpoint group together. The PTQ run itself is usually not there: reach it through `modelopt_run_url`, and use this server's own `experiment_id` for any query here.
```
meenchen: > _Bot comment._

This file is vendored verbatim (frontmatter: "Vendored verbatim from NVIDIA NeMo Evaluator (commit 8fa16b2)"), and `.agents/scripts/sync-upstream-skills.sh` does `rm -rf "$DEST_BASE/$skill"` then re-fetches for exactly `launching-evals accessing-mlflow`, with the header comment stating "If you have local changes to a vendored skill, they will be lost — that is expected, since vendored-verbatim skills should not be modified locally."

So both additions here (this paragraph and the `tags.modelopt_run_id` query below) disappear on the next re-sync, taking the loop-closing query with them, and the provenance line stops being true as of this commit. Options: put the guidance in the `evaluation` skill (explicitly a modified fork, manually managed), or teach the sync script to re-apply a local overlay for this skill and adjust the provenance wording.

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -363,6 +375,25 @@ On SLURM, several deploy/eval failures are invisible to `--dry-run` and only sur
 - Find every `???` left. Ask the user only for what can't be inferred (SLURM hostname/account/output_dir, the `cpu_partition` for auto-export, etc.). Don't propose defaults; let them give plain text. (`tracking_uri` is **not** one of these — it's `${oc.env:MLFLOW_TRACKING_URI}` from `modelopttools:eval-config`.)
 - **`parallelism`** — size it yourself from the run shape (total requests = `dataset_size × repeats` vs GPU serving capacity), and set `--max-num-seqs` to match. Read `references/parallelism.md` for the decision rule and worked examples; only ask the user if a non-GPU cap (e.g. judge rate limit) is unknown.
 - Ask about other defaults they may want to change (partition, walltime, MLflow tags).
+- **ModelOpt provenance.** When Step 3 found a `.experiment.json`, carry it into
+  `export.mlflow` so evals group under the run that quantized the checkpoint:
+
+  | `.experiment.json` field | goes to |
+  | --- | --- |
+  | `experiment_name` | `experiment_name`, verbatim — replaces `${oc.env:USER}/CHANGEME-served-model-name` |
```
meenchen, line 383: > _Bot comment._

The `${` guard two paragraphs down covers only the tags, but `experiment_name` is copied verbatim into the same `export.mlflow` block and is resolved by the same OmegaConf pass. `default_experiment_name()` sanitizes to `[A-Za-z0-9._-]`, but `--mlflow_experiment` is passed through unsanitized, so a PTQ run named with `${oc.env:...}` reaches this row. Suggest wording the rule as "skip/quote any value containing `${`, including `experiment_name`" rather than per-field.

plugins/modelopt/skills/accessing-mlflow/SKILL.md
cjluo-nv: Agreed — the change to this file is dropped entirely as of 9d99d1b, reverted to upstream.

Verified before acting: the frontmatter does say `Vendored verbatim from NVIDIA NeMo Evaluator (commit 8fa16b2)`, and `sync-upstream-skills.sh` does `rm -rf "${DEST_BASE:?}/$skill"` then re-fetches for exactly `launching-evals accessing-mlflow`. So the addition would have been erased on the next re-sync while falsifying that header in the meantime — worse than not adding it.

Of the two options you offered I took the first: the loop-closing query now lives in the `evaluation` skill (a managed fork) as one line in the Step 4 provenance block — "To find these evals again later, query this server for `tags.modelopt_run_id = '<ptq_run_id>''." Teaching the sync script an overlay seemed like more machinery than one sentence justifies, but say the word if you would rather have that.

plugins/modelopt/skills/evaluation/SKILL.md
cjluo-nv: Correct — fixed in 9d99d1b.

I confirmed the gap in the merged code rather than taking it on faith: `_sanitize` is only ever called from inside `default_experiment_name()`, and `resolve_mlflow_args` does `args.mlflow_experiment = args.mlflow_experiment or default_experiment_name(...)` — so an explicit `--mlflow_experiment` never passes through it and reaches `.experiment.json` as typed.

Reworded per-value rather than per-field: "**Skip any imported value containing `${`, `experiment_name` included**", with the reason that one OmegaConf pass resolves the whole block. Applied at all three sites (Step 4, `example_eval.yaml`, `nel-next.md`).

A follow-up commit (381e71d2) also says what to do after skipping — drop the tag, fall back to the default for `experiment_name`, and report which was dropped. A fresh agent re-validating the section skipped the right value but had to guess that part.

plugins/modelopt/skills/evaluation/references/nel-next.md
```diff
@@ -193,6 +193,11 @@ with its own `run_id`, copying the shared `services:` block.
   alias, whose 308 strips `/api/...` → 405). **SLURM does NOT auto-export** — push after
   the run with `nel-next.sh mlflow-push` (Run flow), which resolves the var and falls back
   to the canonical host.
+- **ModelOpt provenance.** Same rule as SKILL.md Step 4: if the served checkpoint holds a
+  `.experiment.json`, use its `experiment_name` verbatim and add its quoted
```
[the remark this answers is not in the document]
cjluo-nv: Fixed in 9d99d1b — same change as the `experiment_name` guard in SKILL.md. The nel-next bullet now reads "skipping any imported value containing `${` — `experiment_name` included, since the same pass resolves the block."

meenchen (approved):

Approving — both prior concerns are fixed in the current diff, and the change is now confined to the manually-maintained `evaluation` skill.

No action needed:
- :heavy_check_mark: Resolved since the last review: the vendored `accessing-mlflow/SKILL.md` edit was reverted (file matches upstream again, loop-closing `tags.modelopt_run_id` query moved into `evaluation/SKILL.md` Step 4), and the "skip any value containing `${`" rule now explicitly covers `experiment_name` at all three sites (Step 4, `example_eval.yaml`, `nel-next.md`).
- Docs-only change to a fork the sync script excludes, so no re-sync loss and no tests expected; PR body records an end-to-end submission plus cross-server verification.

chadvoegele (approved):

PR #2412, opened 2026-09-11
[6463897] Fix narrow FP16 histogram calibration

ajrasane:
### What does this PR do?

Type of change: Bug fix

FP16 entropy calibration can fail for sufficiently narrow activation ranges because NumPy may construct the histogram bin edges at FP16 precision. NumPy 2.2 and later reject the resulting collapsed bin spacing, while earlier versions can silently return invalid, non-monotonic edges. The same precision issue can recur when ONNX Runtime merges later calibration batches.

Losslessly widen FP16 activation values to FP32 while calculating and merging histograms in both ONNX entropy calibration paths, then restore the source dtype at the calibration-to-quantization boundary. This keeps the histogram bins stable without changing FP16 Q/DQ scale or graph dtype semantics. FP32 inputs and public APIs are unchanged.

For full-range FP16 activations, ONNX Runtime can overflow while subtracting FP16 calibration endpoints before it widens the result. Retry only a non-finite FP16 scale calculation with FP32 endpoints, then cast the finite scale back to FP16. Existing finite FP16 calculations and all non-FP16 calculations continue to use ONNX Runtime's original result.

The fallback intentionally patches only the `qdq_quantizer` binding used for calibrated activation ranges. Initializer and weight quantization continue to use ONNX Runtime's existing `quant_utils` path unchanged; full-range FP16 weight scaling is outside this calibration fix.

The regression tests exercise both collectors across initial collection, an equal-range merge, and an expanding-range merge. They also verify the internal FP32 histogram and external FP16 calibration-range contract, including finite saturation when restoring sanitized values. A real entropy calibration test covers full-range FP16 values and verifies finite FP16 Q/DQ scales and a loadable ONNX Runtime graph. The AutoCast integration verifies the same FP16 scale-type contract through the public quantization path.

### Usage

N/A — no API or usage change.

### Testing

All tests ran with CUDA hidden.

- NumPy 1.26.4: affected histogram, AutoCast, and ORT-patching modules: `39 passed`.
- NumPy 2.2.3: affected histogram, AutoCast, and ORT-patching modules: `39 passed`.
- NumPy 2.3.5: affected histogram, AutoCast, and ORT-patching modules: `39 passed`.
- Public INT8 entropy quantization with full-range FP16 calibration data: finite FP16 Q/DQ scales, full ONNX check passed, and the CPU ONNX Runtime session loaded.
- Changed-file pre-commit hooks: passed.

### Additional Information

Follow-up to #1558.


cjluo-nv (commented):

Nudge: the FP32-widen/restore approach looks correct and well-scoped, but the actual wiring point in `_quantize_static` is never exercised by a test, so I can't fully confirm the end-to-end contract.

Needs action:
- [ ] Add a test that drives `_quantize_static` with a real `HistogramCalibrater` so the new `isinstance(...)` restore branch executes — `test_quantize_static_basic` passes a plain `Mock`, so that line is currently untested.
- [ ] Confirm the restored FP16 `lowest`/`highest` still yield an FP16 QDQ scale on a real FP16 model (this is the reason for restoring at all, and nothing asserts it).
- [ ] Replace the vacuous `assert len(edges) == len(hist) + 1` in `test_collect_value_fp16_narrow_range` (always true for `np.histogram`) with a bin-count or bin-width check.
- [ ] Widen the CHANGELOG entry in `CHANGELOG.rst`: the fix also covers Distribution (FP8) and asymmetric-percentile histogram paths, not just "INT8 entropy".

No action needed:
- Rewriting `test_collect_value_fp16_narrow_range` is justified — it now covers both collectors plus equal- and expanding-range merges, strictly more coverage than before.
- `_collect_absolute_value` still builds FP16 edges, but it's unreachable from the int8/fp8 entry points.

cjluo-nv (commented):

Nudge: the prior test gaps are closed, but the newly added `compute_scale_zp` monkeypatch is fresh surface I can't fully verify against both pinned ORT versions.

Needs action:
- [ ] Confirm `onnxruntime.quantization.quant_utils.compute_scale_zp` has the same `(rmin, rmax, qmin, qmax, symmetric, min_real_range)` signature on both pinned ORT lines (`~=1.22.0` for py3.10, `~=1.24.2` otherwise) in `modelopt/onnx/quantization/ort_patching.py`; forwarding positionally breaks loudly if an argument is ever inserted — `*args, **kwargs` passthrough would be safer.
- [ ] Note in the PR body (or a code comment) that only `qdq_quantizer`'s binding is patched, so weight scales computed via `quant_utils.quantize_data` keep the original FP16 overflow behavior — confirm that's intentional scope.
- [ ] Confirm the unprotected `patch_ort_modules(False)` call in `tests/unit/onnx/quantization/test_ort_patching_histogram.py` (process-global, never undone) can't affect later unit tests.

No action needed:
- :heavy_check_mark: Resolved since the last review: real `HistogramCalibrater` coverage of `_quantize_static`, the FP16 QDQ-scale assertion, and the vacuous edge-count assert. The CHANGELOG scoping was declined with a reasonable rationale, and the deleted GPU test moved to the unit suite with strictly more coverage.

cjluo-nv (approved):

PR #2363, opened 2026-09-09
[Feature] Add Qwen3.8-27B local-Hessian PTQ recipes

realAsma:
### What does this PR do?

Type of change: new feature

Adds a reusable NVFP4 W4A4 local-Hessian PTQ recipe and a checkpoint-specific recipe for `Qwen/Qwen3.8-27B`.

The checkpoint recipe reproduces the validated 5.5-bit NVFP4-max AutoQuantize assignment: NVFP4 W4A4 for MLP projections and `lm_head`, FP8 W8A8 for self-attention and large linear-attention projections, and no quantization for the vision tower, MTP layers, KV cache, or `linear_attn.in_proj_a/b`.

This PR is stacked on #2339.

### Usage

```bash
cd examples/hf_ptq
python hf_ptq.py \
  --pyt_ckpt_path Qwen/Qwen3.8-27B \
  --recipe models/Qwen/Qwen3.8-27B/ptq/autoquant_nvfp4_max_5p5bits_local_hessian \
  --export_path <quantized_ckpt_path>
```

Local-Hessian calibration currently requires a single-rank run.

### Testing

- `pre-commit run --files <five changed paths>`
- `pytest tests/unit/recipe/test_loader.py tests/unit/recipe/test_recipe_docs.py -q -x` — 321 passed
- Loaded and resolved both added recipes.
- Compared the Qwen recipe against the final checkpoint `hf_quant_config.json`; all 401 quantized modules matched the expected FP8/NVFP4 formats.

### Before your PR is "Ready for review"

- Is this change backward compatible?: ✅
- If you copied code from any other sources or added a new PIP dependency, did you follow guidance in `CONTRIBUTING.md`: ✅
- Did you write any new necessary tests?: N/A — existing recipe discovery and documentation tests cover both additions.
- Did you update Changelog?: ✅
- Did you get Claude approval on this PR?: N/A

## Summary by CodeRabbit

- **New Features**
  - Added reusable NVFP4 W4A4 post-training quantization recipes with local-Hessian calibration and FP8 scale sweeping.
  - Added Qwen3.8-27B recipes supporting mixed NVFP4/FP8 quantization, including AutoQuantize-derived assignments.
  - Added configuration options for layerwise quantization, dynamic activations, static weight scales, and selective component coverage.

- **Documentation**
  - Updated the PTQ recipe catalog and changelog to accurately describe the new recipes and model configurations.
  - Documented the single-rank calibration requirement for local-Hessian calibration.

PR #2417, opened 2026-09-11
docs: add Local Hessian NVFP4 weight-scale announcement blog

realAsma:
### What does this PR do?

Type of change: documentation + new feature

Adds the Local-Hessian NVFP4 weight-scale announcement blog at
`docs/source/announcements/local-hessian.rst`, together with the calibration
default and the recipes needed to reproduce it.

**Documentation**

- Derivation of the per-block output-error objective and its `16x16` local
  Hessian, with numbered equations.
- Results on Qwen3.5-9B: scale-setting comparison against max, MSE and
  Four-over-six, plus composition with GPTQ.
- Figure 1, a grouped bar chart of the Qwen3.8-27B W4A4 candidate scores.
- `docs/source/_static/announcements.css`: the `shibuya` theme has no
  `span.eqno` rule, so Sphinx's default `float: right` on equation numbers
  cannot share a line with MathJax's full-width display block and the number
  renders *above* the equation. This anchors it to the right instead, and
  shrinks the table-note class.

**Calibration behavior (backward breaking)**

- `layerwise.get_qdq_activations_from_prev_layer` now defaults to `True` for
  all layerwise calibration, and the per-algorithm `GPTQCalibConfig` validator
  that previously injected it is removed. Layerwise calibration exists
  principally to serve GPTQ and Local-Hessian, whose error propagation depends
  on prior-layer QDQ activations, so this is the behavior that matches what
  enabling layerwise normally asks for. This is a design default, not an
  empirical quality claim — no A/B data is offered for it.
  - It changes behavior for existing layerwise `max`/`mse`/AWQ/SmoothQuant
    users who never set the flag, so it is listed under **Backward Breaking
    Changes** in `CHANGELOG.rst`. Set
    `layerwise.get_qdq_activations_from_prev_layer=False` to restore the
    previous behavior.
  - Concretely, layerwise `max`/`mse` is no longer amax-equivalent to the
    non-layerwise pass by default, so `layerwise.enable: true` is no longer a
    pure memory/time optimization for those algorithms.
  - The four shipped `max`/`mse` layerwise recipes pin
    `get_qdq_activations_from_prev_layer: false` explicitly, as
    algorithm-specific exceptions that want full-precision activations for
    subsequent layers.
- `gptq` and `local_hessian_calibrate` API docs now recommend layerwise
  calibration with a calibration batch size of 1.

**Where layerwise is enabled**

Layerwise calibration is enabled in the Qwen3.8-27B local-Hessian recipe —
the recipe the blog's numbers come from — and only there. The shared preset
`modelopt_recipes/configs/ptq/presets/model/nvfp4_w4a4_weight_local_hessian.yaml`
deliberately leaves layerwise off, and so do its consumers: the
`general/ptq/nvfp4_default-kv_none-local_hessian` recipe, the exported
`mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`, and
`--qformat nvfp4_w4a4_weight_local_hessian`. Layerwise calibration requires a
model with identifiable decoder layers, and those three are general-purpose
entry points that must keep working on models without them (a plain
`nn.Sequential`, say). Users who want layerwise on the public config can enable
it explicitly via `"layerwise": {"enable": True}`, as the snippet below shows.

**Recipes**

- `general/ptq/nvfp4_default-kv_none-local_hessian` — reusable NVFP4 W4A4
  local-Hessian recipe wrapping the shared preset above.
- `models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_local_hessian` — the
  recipe the blog reproduces, plus the internal
  `nvfp4_w4a4_mlp_fp8_attn_max.yaml` precision-assignment fragment it imports.
  That fragment is an internal `$import` target consumed by the recipe above,
  not a runnable `--recipe` entry point.

### Usage

```python

config = {
    "quant_cfg": [...],  # quantizer configuration
    "algorithm": {
        "method": "local_hessian",
        "fp8_scale_sweep": True,
        "layerwise": {"enable": True},
    },
}

model = mtq.quantize(model, config, forward_loop)
```

Or use the shipped recipe directly:

```bash
  --recipe modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_local_hessian.yaml \
  ...
```

### Testing

- `pytest tests/unit/recipe/` — recipe loader, preset and recipe-docs coverage,
  including a new `test_local_hessian_layerwise_is_scoped_to_the_model_recipe`
  that pins layerwise on for the Qwen3.8-27B recipe and off for the shared
  preset, the exported config and the `--qformat` mapping.
- `pytest "tests/gpu/torch/quantization/test_quantize_cuda.py::test_quantize[config13-SimpleLinear]"`
  — the public `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG` still calibrates a
  plain `nn.Sequential`, which is exactly what keeping layerwise out of the
  shared preset buys.
- `pytest tests/unit/torch/quantization/test_config_validation.py` — updated
  for the new base default and the removed per-algorithm GPTQ validator.
- `pytest tests/unit/torch/quantization/test_layerwise_calibrate.py` — layerwise
  regression coverage, unchanged and passing.
- `pre-commit run check-modelopt-recipes --all-files` — run repo-wide rather
  than on changed files only, because the recipe-validation hook excludes
  `modelopt_recipes/configs/`, so a changed-files run would skip the preset the
  new recipes import.
- Warning-fatal Sphinx build passes for the new blog page and its
  cross-references.
- The blog's accuracy tables are single-run team measurements; the blog says so
  and makes no significance claims.

- Is this change backward compatible?: **No.** Layerwise calibration now
  defaults to prior-layer QDQ activations for every algorithm. This is
  documented under **Backward Breaking Changes** in `CHANGELOG.rst`, with the
  opt-out noted above.
- If you copied code from any other sources or added a new PIP dependency, did
  you follow guidance in `CONTRIBUTING.md`: N/A
- Did you write any new necessary tests?: **Yes** — updated
  `test_config_validation.py` for the default change and added a focused preset
  test pinning where layerwise is and is not enabled.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?:
  **Yes** — a **Backward Breaking Changes** entry for the calibration default
  and a **New Features** entry for the local-Hessian recipes.
- Did you get Claude approval on this PR?: Claude review run on this branch
  reported CRITICAL 0 / IMPORTANT 2 / SUGGESTION 2; the two IMPORTANT findings
  are addressed or answered on their review threads.

### Additional Information

Two review threads remain open by design, for maintainer sign-off rather than
further changes:

1. The global `get_qdq_activations_from_prev_layer=True` default is an
   intentional design choice for the layerwise abstraction, not an A/B-backed
   quality claim. Recorded as backward breaking with an explicit opt-out.
2. `models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max.yaml` is
   deliberately an internal `QuantizeConfig` `$import` fragment rather than a
   runnable recipe; naming and placement conventions for internal fragments can
   be revisited separately.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance on NVFP4 Local-Hessian weight-scale selection, including mathematical details, accuracy comparisons, runtime considerations, limitations, configuration examples, and reproduction steps.
  - Updated announcement labels, headings, metadata, descriptions, and filtering text to use “Local-Hessian.”

- **Style**
  - Improved announcement formatting for equation labels, display-equation spacing, Hessian results, table headers, and explanatory notes.


docs/source/announcements/local-hessian.rst
```diff
@@ -0,0 +1,335 @@
...
+
+Scale Selection For NVFP4
+**************************
+
+NVFP4 represents each group of 16 weights with FP4 values and an FP8 block
+scale [1]_. This block scale is used to scale the per-block values so to NVFP4 E2M1 range (-6.0, 6.0).
+The default way is to set the block scale based on the per-block maximum value (max scaling) [1]_.
+
+As shown, originally in 'Four-Over-Six' paper [2]_, this block scale can be selected based on other 
+critieria like per-block error. While 'Four-Over-Six' selects the per-block scale from  2 candidates while
```
cjluo-nv, line 32: > _Bot comment._

Typos in this paragraph: "critieria" → "criteria", "exhuastive" → "exhaustive". The sentence also has two `while` clauses ("While 'Four-Over-Six' selects ... while Model-Optimizer MSE sets ..."), which leaves it without a main clause — suggest "'Four-Over-Six' selects the per-block scale from 2 candidates, while ModelOpt MSE sweeps all 126 positive non-zero FP8 scales." Also "Both of these approaches ... only considers" → "consider".

cjluo-nv (commented):

Comment: the blog is technically accurate (numbers, 126-candidate sweep, config schema and `:ref:`quant-cfg`` all check out), but it ships a reproduce command pointing at a recipe that doesn't exist plus several typos in public-facing prose.

Needs action:
- [ ] Fix the `--recipe` path in `docs/source/announcements/local-hessian.rst`; `modelopt_recipes/models/Qwen/Qwen3.8-27B/` does not exist today (only `Qwen3.8-2.4T-A95B`), so the command fails as written. Land after #2363 or drop it.
- [ ] Correct the typos/grammar flagged inline: "critieria", "exhuastive", the double-"while" sentence, "In Table 1 we compares ... dor weights".
- [ ] Repoint the "Model Optimizer Local-Hessian code" link at `local_hessian_calibrate`; it currently targets `layerwise_calibrate`.
- [ ] Run `sphinx-build` before merge: `announcements.css` is loaded site-wide via `html_css_files`, so the new unscoped `div.math` / `.yue .math-wrapper` rules restyle equations on every docs page, not just announcements.

No action needed:
- Docs-only change, so no tests expected; table averages are internally consistent with the reported scores.

docs/source/announcements/local-hessian.rst
```diff
@@ -0,0 +1,335 @@
...
+
+where :math:`W_b\in\mathbb{R}^{1\times16}` is one NVFP4 block of weights,
+:math:`s_b` is that block's scale, and
+:math:`X_b\in\mathbb{R}^{16\times N}` holds the rows of :math:`X` the
+block multiplies, so the local Hessian :math:`X_bX_b^{\top}` is only
+:math:`16\times16`. We sweep all 126 candidate FP8 scales per block, just
+as the MSE algorithm does; because blocks are independent, a `Triton kernel
+<https://github.com/NVIDIA/Model-Optimizer/blob/f70991f36e697617bad7ca4d00459fff200377a9/modelopt/torch/kernels/quantization/gemm/nvfp4_fp8_sweep.py#L229>`_
+does the whole layer at once. See the `Model Optimizer Local-Hessian code
+<https://nvidia.github.io/Model-Optimizer/reference/generated/modelopt.torch.quantization.model_calib.html#modelopt.torch.quantization.model_calib.layerwise_calibrate>`_
```
cjluo-nv, line 106: > _Bot comment._

This link is labelled "Model Optimizer Local-Hessian code" but its anchor is `model_calib.layerwise_calibrate`; the later "Using Local-Hessian" section links `local_hessian_calibrate`. Looks like the wrong anchor here. Consider using a Sphinx `:func:` cross-reference instead of a hardcoded nvidia.github.io URL so the build fails if the target moves.

docs/source/announcements/local-hessian.rst
```diff
@@ -0,0 +1,335 @@
...
+for details.
+
+Local-Hessian Results
+**********************
+
+Accuracy Comparison
+====================
+
+In Table 1 we compares Local-Hessian Vs other scale selection algorithms dor weights on Qwen 3.5 9B.
```
cjluo-nv, line 115: > _Bot comment._

"In Table 1 we compares Local-Hessian Vs other scale selection algorithms dor weights on Qwen 3.5 9B." → "In Table 1 we compare Local-Hessian against other scale-selection algorithms for weights on Qwen3.5-9B." (also keep the model naming consistent with `Qwen3.5-9B` used elsewhere).

docs/source/_static/announcements.css
```diff
@@ -142,3 +142,65 @@
 .toctree-wrapper.compound:empty {
   display: none;
 }
+
+/* Numbered equations: Sphinx's default `float: right` on the equation number
+   cannot share a line with MathJax's full-width display block, so the number
+   lands above the equation. Anchor it to the right of the equation instead. */
+div.math {
+  position: relative;
```
cjluo-nv, line 150: > _Bot comment._

`announcements.css` is registered in `conf.py` `html_css_files`, so it applies to the whole site, not just announcement pages. These new `div.math` / `div.math span.eqno` / `.yue .math-wrapper` rules will therefore change equation layout and spacing on every page with display math (e.g. `autoquantize.rst` and the guides). The PR body says the rendered output hasn't been checked with a full `sphinx-build` — please verify those other pages before merge, or scope these rules to the announcement pages.

docs/source/announcements/local-hessian.rst
```diff
@@ -0,0 +1,335 @@
...
+candidates per block, minimizing :math:`E(s)` jointly means searching
+:math:`M^{C_{\mathrm{in}}/16}` combinations -- this is not tractable. So we
+choose each block's scale in isolation, against the output error that
+block alone contributes. For block :math:`b`,
+
+.. math::
+   :label: lh-block-error
+
+   E_b(s_b) = \Delta(W_b,s_b)\,(X_bX_b^{\top})\,\Delta(W_b,s_b)^{\top},
+
```
Fridah-nv, line 97: We actually dropped the cross-block sum so the algorithm is a block-diagonal approximation of the real error (what `local` refer to in the name). This paragraph is explaining this exactly, but I feel it can be clearer, like ` we add up per-block damage and ignore how errors in different blocks interfere.`

docs/source/announcements/local-hessian.rst
```diff
@@ -0,0 +1,335 @@
...
+       --calib_seq 2048 \
+       --batch_size 1 \
+       --export_path <export_dir>
+
+.. note::
+
+   Calibration is layerwise: each layer is calibrated on the fake-quantized
+   outputs of the one before it, so later layers get an activation
+   distribution that accounts for the quantization of the earlier layers.
+
```
Fridah-nv, line 305: How much do we understand about keeping `layerwise=True` and `qdq_from_prev=True` for local-hessian? Asking because in the code/doc of local-hessian we are not mentioning these details. If we have results show that local-hessian works better with layerwise and bs=1, we shall update the API docstring as well

meenchen (approved):

docs/source/announcements/local-hessian.rst
realAsma: > _:robot: Bot comment._

Addressed in 040c80f34. I corrected the spelling and grammar, removed the duplicate `while` construction, and fixed the subject-verb agreement in the scale-selection paragraph.

docs/source/announcements/local-hessian.rst
realAsma: > _:robot: Bot comment._

Addressed in 040c80f34. The Local-Hessian link now uses a Sphinx `:func:` reference to `modelopt.torch.quantization.model_calib.local_hessian_calibrate`; the other generated API links were made consistent as well.

docs/source/announcements/local-hessian.rst
realAsma: > _:robot: Bot comment._

Addressed in 040c80f34. The sentence now uses the correct comparison wording and the consistent `Qwen3.5-9B` model name.

docs/source/announcements/local-hessian.rst
realAsma: How much do we understand about keeping layerwise=True and qdq_from_prev=True for local-hessian? -> I not have A/B tests - We should make this the default recommendation based on our understanding from GPTQ.

docs/source/announcements/local-hessian.rst
realAsma: > _:robot: Bot comment._

Addressed in 4e993078d. The blog now explains that Local-Hessian makes the search tractable with a block-diagonal approximation: it sums each block's output error independently and ignores interactions between quantization errors in different blocks.

docs/source/announcements/local-hessian.rst
realAsma: > _:robot: Bot comment._

Addressed in 4e993078d. The GPTQ and Local-Hessian API docs now recommend layerwise calibration with batch size 1, prior-layer QDQ activations are the layerwise default, and enabled max/MSE layerwise recipes explicitly retain `get_qdq_activations_from_prev_layer: false`.

docs/source/_static/announcements.css
realAsma: verified

Fridah-nv (approved):
LGTM, thanks!

docs/source/announcements/local-hessian.rst
```diff
@@ -0,0 +1,330 @@
...
+See :ref:`quant-cfg` for how to write the ``quant_cfg`` field.
+
+To reproduce the published Qwen3.8-27B checkpoint end to end:
+
+.. code-block:: bash
+
+   python examples/hf_ptq/hf_ptq.py \
+       --pyt_ckpt_path Qwen/Qwen3.8-27B \
+       --recipe modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_local_hessian-fp8_attn-kv_fp8_cast.yaml \
```
cjluo-nv: > _Bot comment._

This recipe path still does not exist: `modelopt_recipes/models/Qwen/` only contains `Qwen3.8-2.4T-A95B`, so the documented reproduce command fails. Either land this after the recipe from #2363 merges, or drop the `--recipe` line and point at `NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG` instead.

modelopt/torch/quantization/config.py
```diff
@@ -723,13 +723,14 @@ class LayerwiseConfig(ModeloptBaseConfig):
     )
 
     get_qdq_activations_from_prev_layer: bool = ModeloptField(
-        default=False,
+        default=True,
```
cjluo-nv, line 726: > _Bot comment._

Flipping this default to `True` for every algorithm is backward-breaking for all layerwise max/mse users, and this PR immediately pins `get_qdq_activations_from_prev_layer: false` in four shipped recipes to preserve their behavior — which suggests the new default is wrong for max/mse. Consider keeping the per-algorithm default (move the removed `_gptq_qdq_default` validator onto `GPTQCalibConfig` + `LocalHessianCalibConfig`) rather than flipping globally, especially given there are no A/B results backing the change.

cjluo-nv (commented):

Comment: the prose fixes landed, but the blog still documents a `--recipe` path that does not exist, and the PR now carries an undescribed backward-breaking calibration default flip.

Needs action:
- [ ] Fix or drop the `--recipe` path in `docs/source/announcements/local-hessian.rst:291`; only `modelopt_recipes/models/Qwen/Qwen3.8-2.4T-A95B/` exists, so the reproduce command fails as written (land after #2363).
- [ ] Justify or scope the `get_qdq_activations_from_prev_layer` default flip in `modelopt/torch/quantization/config.py` — four shipped max/mse layerwise recipes had to be pinned back to `false`, which argues for keeping it per-algorithm (GPTQ/Local-Hessian) instead of global.
- [ ] Update the PR body (still says "Documentation only; no code paths change") or split the config/mode/model_calib default change into its own PR.

No action needed:
- :heavy_check_mark: Resolved since the last review: the prose typos, the `layerwise_calibrate` → `local_hessian_calibrate` `:func:` anchor, the block-diagonal wording, and the site-wide CSS check (author confirmed verified).
- Test edits in `test_config_validation.py` are justified by the default change; coverage is preserved.

modelopt/torch/quantization/config.py
cjluo-nv, line 726: > _Bot comment._

The default is now `True` for every layerwise algorithm, but this same PR adds `get_qdq_activations_from_prev_layer: false` to four shipped max/mse recipes to keep their behaviour — and the description right below still says "We recommend setting this to True for GPTQ and Local-Hessian". That combination reads as "the new global default is wrong for max/mse". Consider restoring the per-algorithm default (move the removed `_gptq_qdq_default` validator onto `GPTQCalibConfig` + `LocalHessianCalibConfig`) so AWQ/SmoothQuant/max/mse layerwise users aren't silently changed without A/B data.

modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max.yaml
```diff
@@ -0,0 +1,51 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# QuantizeConfig for the precision assignment exported by a 5.5-bit NVFP4-max AutoQuantize sweep.
+
+# modelopt-schema: modelopt.torch.quantization.config.QuantizeConfig
+imports:
```
cjluo-nv, line 19: > _Bot comment._

This file is an `$import` fragment (a bare `QuantizeConfig`, no `metadata`/`quantize`), but it is named like a recipe and sits next to one in `models/<org>/<id>/ptq/`. `--recipe models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max` — the max baseline the blog compares against — fails with a `metadata.recipe_type` error. Other fragments in recipe dirs use a distinguishing name (`*.quant_cfg.yaml`, `disabled_quantizers.yaml`); please follow that, or promote this to a real recipe with its own `metadata`.

cjluo-nv (commented):

Comment: the missing recipe now ships in this PR, but the global `get_qdq_activations_from_prev_layer` default flip is still unscoped and the PR body still describes this as docs-only.

Needs action:
- [ ] :speech_balloon: Author replied "no A/B tests — default recommendation from our GPTQ understanding" — still needs the flip scoped to `GPTQCalibConfig`/`LocalHessianCalibConfig` (or evidence), since this PR pins `false` back into four shipped max/mse recipes, which argues the new global default is wrong for them.
- [ ] Update the PR body: it still says "Documentation only; no code paths change" while the PR changes `config.py`/`mode.py`/`model_calib.py` defaults and adds three recipes — or split the default change out.
- [ ] Rename or clearly mark `modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max.yaml`; it has no `metadata`, so `--recipe .../nvfp4_w4a4_mlp_fp8_attn_max` fails even though it sits beside a loadable recipe.

No action needed:
- :heavy_check_mark: Resolved since the last review: the nonexistent `--recipe` path (recipe now shipped), the prose typos, the `local_hessian_calibrate` anchor, the block-diagonal wording, and the site-wide CSS check.
- `test_config_validation.py` edits are justified by the default change; coverage is preserved, and shipped recipes stay covered by the existing load smoke tests.

modelopt/torch/quantization/config.py
realAsma: > _:robot: Bot comment._

Thanks for the careful read, but we're keeping the global default as `True` here — this is intentional rather than an oversight.

The reasoning is a design choice about what the layerwise abstraction is for, not an empirical claim. Layerwise calibration exists principally to serve GPTQ and Local-Hessian, whose error propagation depends on prior-layer QDQ activations, so `get_qdq_activations_from_prev_layer=True` is the behavior that matches what enabling layerwise is normally asking for. Making that the default of `LayerwiseConfig` keeps the abstraction coherent rather than requiring every principal consumer to opt back into it.

On the four shipped recipes pinning `false`: that is the intended reading of them, not evidence against the default. max/mse are algorithm-specific exceptions that want full-precision activations for subsequent layers, so they state that explicitly in the recipe. They remain free to use layerwise calibration; they just carry the flag.

To be straight about what this is and isn't: we have no A/B data, and we are not claiming `True` produces better quality for max/mse. It is a default chosen from the design intent of layerwise calibration. Because it does change behavior for existing layerwise max/mse/AWQ/SmoothQuant users who never set the flag, it is documented as a backward-breaking change in `CHANGELOG.rst`, and setting `get_qdq_activations_from_prev_layer=False` restores the previous behavior.

Leaving this thread open for you or the maintainers to weigh in if you still disagree.

modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max.yaml
realAsma: > _:robot: Bot comment._

Leaving this one as-is. The file is deliberately an internal `$import` fragment, not a recipe that was accidentally left unloadable.

`nvfp4_w4a4_mlp_fp8_attn_max.yaml` is a bare `QuantizeConfig` holding the precision assignment exported by the AutoQuantize sweep. Its only consumer is the sibling `nvfp4_w4a4_mlp_fp8_attn_local_hessian.yaml`, which imports it as its `quantize` base. That sibling is the user-facing, runnable entry point, and it is the one documented in `modelopt_recipes/ptq.md` and referenced by the blog. So the absence of `metadata.recipe_type` is expected: the fragment is not meant to be a supported top-level `--recipe` target, and `--recipe .../nvfp4_w4a4_mlp_fp8_attn_max` failing is the correct outcome rather than a regression.

That said, your point about signalling isn't wrong — a reader scanning the directory can't tell the two apart by name. I'm not claiming the current placement is beyond improvement. If maintainers want stronger internal-fragment signalling here, naming or placement can be revisited separately; I'd rather do that as its own change than fold it into this PR.

Leaving the thread open for maintainers to weigh in.

modelopt_recipes/general/ptq/nvfp4_default-kv_none-local_hessian.yaml
```diff
@@ -0,0 +1,27 @@
...
+imports:
+  preset: configs/ptq/presets/model/nvfp4_w4a4_weight_local_hessian
+
+metadata:
+  recipe_type: ptq
+  description: >-
+    Applies NVFP4 W4A4 with static weight scales, dynamic inputs, KV quantizers disabled, and
+    local-Hessian weight calibration with an FP8-scale sweep.
+quantize:
+  $import: preset
```
cjluo-nv, line 27: > _Bot comment._

The imported preset `configs/ptq/presets/model/nvfp4_w4a4_weight_local_hessian` declares only `method: local_hessian` + `fp8_scale_sweep: true` — no `layerwise` block — so `LayerwiseConfig.enable` stays `False` here. That contradicts three other places this PR adds: the `local_hessian_calibrate` docstring ("We recommend using Local-Hessian with layerwise calibration enabled"), the blog's usage snippet, and the Qwen3.8-27B recipe. The GPTQ sibling `nvfp4_default-kv_none-gptq.yaml` does enable it. Since this is the model-agnostic entry point `ptq.md` points readers at, please either override the algorithm block here with `layerwise: {enable: true}` (inline keys merge over `$import`, as the Qwen recipe does) or note the deliberate gap in `ptq.md`.

cjluo-nv (commented):

Comment: the blog itself is in good shape, but the bundled global calibration default flip is still unresolved and the new shipped `local_hessian` recipe ships the configuration the blog tells users not to use.

Needs action:
- [ ] :speech_balloon: Author replied the `get_qdq_activations_from_prev_layer=True` flip is a design choice with no A/B data and left the thread for maintainers — still needs an explicit maintainer decision, since this PR pins `false` back into four shipped max/mse recipes in `modelopt/torch/quantization/config.py`.
- [ ] Enable layerwise in `modelopt_recipes/general/ptq/nvfp4_default-kv_none-local_hessian.yaml` (its imported preset has no `layerwise` block, so `enable` stays `False`), or state in `ptq.md` why this entry point deliberately differs from the blog's recommendation and the GPTQ sibling.
- [ ] Update the PR body: the Testing section still says "Documentation only; no code paths change" while the diff changes `config.py`/`mode.py`/`model_calib.py` defaults and adds three recipes.
- [ ] :speech_balloon: Author declined renaming `modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max.yaml` — fine to defer, but confirm a maintainer accepts a non-loadable fragment sitting in a recipe directory.

No action needed:
- :heavy_check_mark: Resolved since the last review: the missing `--recipe` target now ships, plus earlier prose, anchor and CSS fixes.
- `test_config_validation.py` edits are justified by the default change; coverage preserved.

modelopt_recipes/general/ptq/nvfp4_default-kv_none-local_hessian.yaml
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Good catch, fixed in `8fdf7352b`. Rather than overriding the general recipe, I enabled layerwise calibration in the shared preset `configs/ptq/presets/model/nvfp4_w4a4_weight_local_hessian.yaml`, so the change reaches every consumer: the `general/ptq/nvfp4_default-kv_none-local_hessian` recipe, the exported `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`, and `--qformat nvfp4_w4a4_weight_local_hessian`.

All three previously resolved to `layerwise.enable=False`, which was inconsistent with the layerwise configuration used in the blog and recommended by the Local-Hessian API documentation. We did not run an A/B comparison, so this is a configuration-consistency fix rather than an accuracy claim.

One consequence worth stating plainly: `layerwise_calibrate` raises when a model has no identifiable decoder layers or when `forward_loop` is `None`. Callers of the public config in those situations previously fell back silently to a non-layerwise pass and will now fail loudly. That tradeoff is called out in the PR description because the silent fallback masked this configuration mismatch.

I left `checkpoint_dir` out of the shared preset because it would hardcode an output path into every consumer, and left `get_qdq_activations_from_prev_layer` implicit because it already defaults to `True`. The focused preset test landed with the fix; `c832f1169` additionally covers both the `--qformat` mapping and the separately loaded exported public config.

realAsma: > _:robot: Bot comment._

Good catch, fixed in `8fdf7352b`. Rather than overriding the general recipe, I enabled layerwise calibration in the shared preset `configs/ptq/presets/model/nvfp4_w4a4_weight_local_hessian.yaml`, so the change reaches every consumer: the `general/ptq/nvfp4_default-kv_none-local_hessian` recipe, the exported `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`, and `--qformat nvfp4_w4a4_weight_local_hessian`.

All three previously resolved to `layerwise.enable=False`, which was inconsistent with the layerwise configuration used in the blog and recommended by the Local-Hessian API documentation. We did not run an A/B comparison, so this is a configuration-consistency fix rather than an accuracy claim.

One consequence worth stating plainly: `layerwise_calibrate` raises when a model has no identifiable decoder layers or when `forward_loop` is `None`. Callers of the public config in those situations previously fell back silently to a non-layerwise pass and will now fail loudly. That tradeoff is called out in the PR description because the silent fallback masked this configuration mismatch.

I left `checkpoint_dir` out of the shared preset because it would hardcode an output path into every consumer, and left `get_qdq_activations_from_prev_layer` implicit because it already defaults to `True`. The focused preset test landed with the fix; `c832f1169` additionally covers both the `--qformat` mapping and the separately loaded exported public config.

shengliangxu (approved):
LGTM

modelopt_recipes/configs/ptq/presets/model/nvfp4_w4a4_weight_local_hessian.yaml
```diff
@@ -25,6 +25,8 @@ imports:
 algorithm:
   method: local_hessian
   fp8_scale_sweep: true
+  layerwise:
+    enable: true
```
cjluo-nv: > _Bot comment._

This propagates into `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`, which `tests/gpu/torch/quantization/test_quantize_cuda.py::test_quantize` runs against `SimpleLinear` — a plain `nn.Sequential` with no registered decoder layers. `wrapped_calib_func` now takes the layerwise branch, and `layerwise_calibrate` raises `ValueError("Could not find transformer layers in model...")` before calibration; there is no fallback to the non-layerwise pass. That GPU case should start failing.

Either drop the exported constant from that parametrization (or override `layerwise.enable: false` there with a comment), or set layerwise in the consuming recipes rather than the shared preset so the public config keeps working on models without decoder layers.

cjluo-nv (commented):

Comment: the preset now enabling layerwise fixes the blog/recipe mismatch but breaks an existing GPU test, and the second backward-breaking change never made it into `CHANGELOG.rst`.

Needs action:
- [ ] Update `tests/gpu/torch/quantization/test_quantize_cuda.py`: `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG` now carries `layerwise.enable: true`, so `layerwise_calibrate` raises `ValueError("Could not find transformer layers")` on `SimpleLinear`. See inline comment.
- [ ] Add a `Backward Breaking Changes` entry in `CHANGELOG.rst` for the preset flip (`--qformat nvfp4_w4a4_weight_local_hessian` and the exported config now require decoder layers and a `forward_loop`); the PR body claims it is there but only the QDQ default is.
- [ ] :speech_balloon: Author replied the global `get_qdq_activations_from_prev_layer=True` is a deliberate design default with no A/B data — still needs an explicit maintainer ack on the thread, since four shipped max/mse recipes pin `false` back.
- [ ] :speech_balloon: Author declined renaming `modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max.yaml` — fine to defer, but confirm a maintainer accepts a non-loadable `$import` fragment in a recipe directory.

No action needed:
- :heavy_check_mark: Resolved since the last review: the local-Hessian preset now enables layerwise (with `test_local_hessian_preset_enables_layerwise_calibration`), the PR body no longer claims docs-only, plus the earlier prose, anchor and CSS items.
- `test_config_validation.py` edits are justified by the default change; coverage preserved.

modelopt_recipes/general/ptq/nvfp4_default-kv_none-local_hessian.yaml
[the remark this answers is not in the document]
realAsma: > _:robot: Bot comment._

Correction to my earlier reply: I have reverted that approach.

Enabling layerwise in the shared preset propagated into the exported `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`, and that broke `tests/gpu/torch/quantization/test_quantize_cuda.py::test_quantize[config13-SimpleLinear]`: `SimpleLinear` is a plain `nn.Sequential`, so `layerwise_calibrate` raised `ValueError: Could not find transformer layers in model`. Chenjie caught this on the preset thread and was right; I reproduced the failure locally before reverting.

In `22f3f9e1b` the preset is restored byte-identically and layerwise remains only in `models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_local_hessian.yaml`, which already had it—that is the recipe the blog's numbers come from. The shared preset, general recipe, exported config, and `--qformat` value therefore keep layerwise off and remain usable on models without identifiable decoder layers. Please disregard the "reaches every consumer" framing and the raise-instead-of-silent-fallback tradeoff in my earlier reply; both described the approach now withdrawn.

That leaves the original point open: `general/ptq/nvfp4_default-kv_none-local_hessian` again resolves to `layerwise.enable=False`, so the model-agnostic entry point still does not match the blog's recommendation. Fixing it in that recipe rather than the preset is the narrower option and avoids the public-config breakage; I am flagging that for a maintainer call rather than changing scope again unilaterally.

A test now pins both halves: layerwise on for the Qwen recipe, off for the preset, exported config, and `--qformat` mapping.

realAsma: > _:robot: Bot comment._

Correction to my earlier reply: I have reverted that approach.

Enabling layerwise in the shared preset propagated into the exported `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`, and that broke `tests/gpu/torch/quantization/test_quantize_cuda.py::test_quantize[config13-SimpleLinear]`: `SimpleLinear` is a plain `nn.Sequential`, so `layerwise_calibrate` raised `ValueError: Could not find transformer layers in model`. Chenjie caught this on the preset thread and was right; I reproduced the failure locally before reverting.

In `22f3f9e1b` the preset is restored byte-identically and layerwise remains only in `models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_local_hessian.yaml`, which already had it—that is the recipe the blog's numbers come from. The shared preset, general recipe, exported config, and `--qformat` value therefore keep layerwise off and remain usable on models without identifiable decoder layers. Please disregard the "reaches every consumer" framing and the raise-instead-of-silent-fallback tradeoff in my earlier reply; both described the approach now withdrawn.

That leaves the original point open: `general/ptq/nvfp4_default-kv_none-local_hessian` again resolves to `layerwise.enable=False`, so the model-agnostic entry point still does not match the blog's recommendation. Fixing it in that recipe rather than the preset is the narrower option and avoids the public-config breakage; I am flagging that for a maintainer call rather than changing scope again unilaterally.

A test now pins both halves: layerwise on for the Qwen recipe, off for the preset, exported config, and `--qformat` mapping.

modelopt_recipes/configs/ptq/presets/model/nvfp4_w4a4_weight_local_hessian.yaml
realAsma: > _:robot: Bot comment._

You are right, and thank you—I reproduced it before changing anything: `test_quantize[config13-SimpleLinear]` failed with `ValueError: Could not find transformer layers in model` at `model_calib.py:2089`, exactly as you described. `config13` is `mtq.NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`, and there is no fallback to the non-layerwise pass.

I took your second option: layerwise is set in the consuming recipe rather than the shared preset. In `22f3f9e1b` the preset is restored byte-identically to its pre-change content, and layerwise remains only in `models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_local_hessian.yaml`, which already carried it. The public config, `--qformat nvfp4_w4a4_weight_local_hessian`, and the general recipe therefore keep working on models without decoder layers, and that GPU case passes again.

I also added a regression test that pins both directions—layerwise on for the Qwen recipe, off for the preset, exported constant, and `--qformat` mapping—so re-adding it to the shared preset fails in unit tests rather than only on a GPU runner.

The tradeoff is that `general/ptq/nvfp4_default-kv_none-local_hessian` goes back to `layerwise.enable=False`, which reopens the point raised on the other thread. I am leaving that for a maintainer call rather than moving scope again.

cjluo-nv (commented):

Nudge: the GPU-test regression is fixed by reverting the preset, but three items the author explicitly deferred still need a maintainer decision before merge.

Needs action:
- [ ] :speech_balloon: Author replied the global `get_qdq_activations_from_prev_layer=True` in `modelopt/torch/quantization/config.py:726` is a deliberate design default with no A/B data — still needs an explicit maintainer ack, since the same PR pins `false` back into four shipped max/mse recipes.
- [ ] Decide the layerwise gap on `modelopt_recipes/general/ptq/nvfp4_default-kv_none-local_hessian.yaml`: it resolves to `layerwise.enable=False` while the blog and `local_hessian_calibrate` recommend enabling it. Either set it in that recipe or note the deliberate gap in `ptq.md`.
- [ ] :speech_balloon: Author declined renaming `modelopt_recipes/models/Qwen/Qwen3.8-27B/ptq/nvfp4_w4a4_mlp_fp8_attn_max.yaml` — fine to defer, but confirm a maintainer accepts a non-loadable `$import` fragment inside a recipe directory.

No action needed:
- :heavy_check_mark: Resolved since the last review: the preset is restored byte-identically so `NVFP4_W4A4_WEIGHT_LOCAL_HESSIAN_CFG`/`SimpleLinear` passes again (pinned by `test_local_hessian_layerwise_is_scoped_to_the_model_recipe`), plus the earlier prose, anchor, CSS and PR-body items.
- Blog/README/`index.rst` dates agree (2026-09-09); `ptq.md`'s 25→26 count matches `general/ptq/`.
- `test_config_validation.py` edits are justified by the default change; coverage preserved.

modelopt/torch/quantization/config.py
cjluo-nv: is it impacting GPTQ and local hessian only?

cjluo-nv (approved):

PR #2420, opened 2026-09-12
Fix AutoCast unknown dimension metadata

haoxiz-nvidia:
### What does this PR do?

Fix a problem for autocast that caused Qwen2.5-0.5B-Instruct to have a shape mismatch problem during inference.

During dynamic shape computations, original code will replace number with unk, for example:

> RoPE Slice outputs: [batch, 2, sequence, 32]
> RoPE Concat output: [batch, 2, sequence, 64]
> Mul_3 output:       [batch, 2, sequence, 64]

to

> Slice:  [batch, unk, sequence, unk]  # actually 2 × 32
> Concat: [batch, unk, sequence, unk]  # actually 2 × 64
> Mul_3:  [batch, unk, sequence, unk]  # actually 2 × 64

ORT believes these identically named dimensions are equal. Its memory planner consequently selects a previously allocated 32-wide buffer for the 64-wide Mul_3 result. At execution, ORT discovers the actual shapes disagree and raises the error.

This fix is to remove the dim so Shape inference can then:
* Restore numeric dimensions it can prove.
* Establish meaningful symbolic relationships where possible.
* Leave other dimensions anonymous or give them distinct generated symbols.

This prevents false equality between 32-wide and 64-wide tensors that caused error.

### Usage

N/A

### Testing

Add a test case 

- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?:  N/A

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dynamic shape handling during model conversion by clearing concrete dimension values without replacing them with an `"unk"` alias.
  - Fixed compatibility for models with symbolic batch and sequence dimensions, including rotary-position embedding (RoPE) graphs.
  - Preserved correct output shapes and rotation values when converted models run with ONNX Runtime.

- **Tests**
  - Added regression coverage for dynamic RoPE-shaped models converted to FP16.


vishalpandya1990:
Is this specific to ROPE part, or applicable to any dynamic-shape containing model? Also, is this a regression in modelopt AutoCast logic?

PR #2426, opened 2026-09-14
Deprecate the single-format quantization CLI flags in favour of --recipe

shengliangxu:
### What does this PR do?

Type of change: deprecation

Deprecates the single-format quantization CLI flags in favour of `--recipe`. Passing one now emits a `DeprecationWarning`; nothing else changes.

| script | flags |
|---|---|
| `examples/hf_ptq` | `--qformat`, `--kv_cache_qformat` |
| `examples/megatron_bridge/quantize.py` | `--quant_cfg`, `--kv_cache_quant`, `--weight_only` |
| `examples/torch_onnx` | `--qformat` |

`--recipe` was already authoritative over all six — silently on `hf_ptq`, and with a runtime warning on `megatron_bridge` — and `modelopt/recipe/presets.py` already records the intent in a comment: *"the long-term direction is to retire `--qformat` / `--kv_cache_qformat` in favour of `--recipe`"*. This makes that a real deprecation.

A recipe carries the quantization config, the calibration algorithm and the KV-cache setting in one file, so they cannot drift apart the way separate flags can. That drift is not hypothetical: the preset path applies no MTP exclusion while the recipe unit `default_disabled_quantizers` disables `mtp.*`, so the same model quantizes differently depending on which entry point was used.

#### The warning fires only when a flag is actually passed

`RecipeSupersededAction` is an `argparse.Action`, and argparse invokes an action only for options present on the command line — never for a default. That matters because several of these default to a *quantizing* value (`--qformat fp8`, `--kv_cache_qformat fp8_cast`); warning on the defaults would fire on every run, including runs that correctly use `--recipe` and never mention the flag.

`examples/speculative_decoding/scripts/quantize_drafter.py` keeps `--qformat` undeprecated: it has no `--recipe`, so there would be nothing to migrate to.

### Usage

```bash
# deprecated
python examples/hf_ptq/hf_ptq.py --pyt_ckpt_path <ckpt> --qformat nvfp4 --kv_cache_qformat fp8_cast

# replacement
python examples/hf_ptq/hf_ptq.py --pyt_ckpt_path <ckpt> \
    --recipe general/ptq/nvfp4_experts_only-kv_fp8_cast
```

### Testing

Three tests in `tests/examples/hf_ptq/test_hf_ptq_args.py`, all passing:

- passing `--qformat` / `--kv_cache_qformat` raises `DeprecationWarning` and still parses the value;
- omitting them raises nothing and leaves the defaults (`fp8`, `fp8_cast`) untouched;
- the action stays wired to both flags, so a future edit cannot drop it while leaving the help text.

Defaults and parsed values were diffed against `main` and are unchanged — the action stores exactly what `store` / `store_true` would have. `ruff` findings are at parity with `main` on every changed file.

- Is this change backward compatible?: ✅ — the flags still work, they only warn.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: ✅ — under 0.48.0 Deprecations.

### Additional Information

Draft: the removal release for these flags is not decided here, only the deprecation.

## Summary by CodeRabbit

* **Deprecations**
  * Legacy quantization CLI options now issue visible `FutureWarning` messages only when explicitly provided.
  * Use `--recipe` instead of deprecated options in Hugging Face PTQ, Megatron-Bridge, and torch-to-ONNX workflows.
  * Existing option values, defaults, and parsing behavior remain unchanged.
  * Weight AutoQuantize recipes without an explicit `kv_cache` setting continue to use `--kv_cache_qformat` as a fallback.


kevalmorabia97 (approved):

modelopt/recipe/presets.py
```diff
@@ -95,3 +98,29 @@ def load_quant_cfg_choices(subdir: str) -> dict[str, dict[str, Any]]:
...
+    so warning unconditionally would fire on every run, including runs that correctly use
+    ``--recipe`` and never mention the deprecated flag.
+
+    Handles both value-taking flags and ``store_true`` ones; for the latter pass
+    ``nargs=0, const=True``.
+    """
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        """Warn that this flag is deprecated, then store the value as usual."""
+        warnings.warn(
```
cjluo-nv, line 125: > _Bot comment._

This warning will not be visible to users in practice. Python's default filters are `default::DeprecationWarning:__main__` plus `ignore::DeprecationWarning`; the `__main__` filter matches the module of the frame selected by `stacklevel`. With `stacklevel=2` inside an `argparse.Action.__call__`, that frame is argparse's own `take_action`, i.e. module `argparse` — so the warning is ignored and `python hf_ptq.py --qformat nvfp4` prints nothing.

The tests don't catch this because `pytest.warns` installs `simplefilter("always")` and pytest's default `-W` config re-enables `DeprecationWarning`.

Options: use `FutureWarning` (the documented category for warnings aimed at end users of an application, shown by default), or emit through `warn_rank_0`/a plain stderr print. Whichever you pick, please add a coverage check that the message reaches stderr under *default* filters (e.g. a subprocess invocation of one of the scripts).

examples/megatron_bridge/quantize.py
```diff
@@ -137,25 +142,37 @@ def get_args() -> argparse.Namespace:
...
+        help=(
+            "(deprecated: use --recipe) KV-cache quantization config to apply on top of "
+            "--quant_cfg. Ignored when --recipe is set."
+        ),
     )
     parser.add_argument(
         "--weight_only",
-        action="store_true",
-        help="Disable input (activation) quantization, i.e. weight-only quantization.",
+        action=RecipeSupersededAction,
```
cjluo-nv, line 168: > _Bot comment._

This is the one flag whose action semantics actually change (`store_true` → custom action with `nargs=0, const=True`), and it has no test. A direct unit test of `RecipeSupersededAction` in `tests/unit/recipe/test_presets.py` — parser with a `nargs=0, const=True` flag, asserting `args.weight_only is True` when passed and `False` when omitted — would cover the `self.const if self.nargs == 0 else values` branch that only this call site exercises.

cjluo-nv (commented):

Comment: the new `DeprecationWarning` is emitted from inside argparse, so Python's default filters suppress it in real runs — users will never see the deprecation this PR exists to deliver.

Needs action:
- [ ] Make the warning visible in a normal run: raise `FutureWarning` (shown by default) or route it through `warn_rank_0`/stderr in `modelopt/recipe/presets.py`; `DeprecationWarning` from a non-`__main__` frame is ignored, and pytest's own filters hide this in the tests.
- [ ] Add a check that the warning actually reaches stderr under default filters (subprocess run of one example script), since `pytest.warns` passes either way.
- [ ] Cover the `nargs=0, const=True` path (`--weight_only`, the only flag whose action semantics changed from `store_true`) with a unit test of `RecipeSupersededAction` in `tests/unit/recipe/test_presets.py`.
- [ ] Replace or drop `test_recipe_superseded_action_is_wired_to_both_flags` — it asserts on `inspect.getsource` text and leaves dead code (unused `parser`/`del parser`, unused `sys.argv` patch, function-local `import inspect`). Inspect `parser._actions` instead.
- [ ] Fix the CHANGELOG entry: it says "all five" while listing six flags.

No action needed:
- Custom action is reasonable: argparse's `deprecated=True` needs 3.13, min supported is 3.10.

tests/examples/hf_ptq/test_hf_ptq_args.py
```diff
@@ -817,3 +818,52 @@ def test_untracked_runs_write_no_experiment_json(monkeypatch, example_utils, tmp
...
+    """The defaults quantize (--qformat fp8, --kv_cache_qformat fp8_cast), so warning on every
+    run -- including runs that correctly pass --recipe -- would be pure noise. argparse only
+    invokes an action for options actually present, which is what keeps this quiet."""
+    _, args = _parse_hf_ptq_args(monkeypatch, "--pyt_ckpt_path", "/models/Qwen3-0.6B")
+    deprecations = [w for w in recwarn if issubclass(w.category, DeprecationWarning)]
+    assert not [w for w in deprecations if "is deprecated" in str(w.message)]
+    # and the defaults themselves are untouched by the deprecation wiring
+    assert args.qformat == "fp8"
+    assert args.kv_cache_qformat == "fp8_cast"
+
```
cjluo-nv, line 849: > _Bot comment._

This test asserts on the *source text* of `parse_args` via `inspect.getsource` and string slicing, which breaks on any reformatting (e.g. the action moving below `help=`) without the behavior changing. It also carries dead code: `parser = argparse.ArgumentParser()` / `del parser`, the unused `sys.argv` monkeypatch, and a function-local `import inspect` (imports belong at the top of the file).

If the intent is "the action stays wired", introspect the real parser instead — e.g. assert `isinstance(action, RecipeSupersededAction)` for the `qformat`/`kv_cache_qformat` entries of `parser._actions` — or drop this test, since the two behavioral tests above already fail if the action is removed.

cjluo-nv (commented):

Approving: every concern from the previous review is fixed in the diff, and the change stays small, backward compatible and well tested.

No action needed:
- :heavy_check_mark: Resolved since the last review: `DeprecationWarning` → `FutureWarning` in `modelopt/recipe/presets.py`, a subprocess test proving the warning reaches stderr under default filters, a direct `nargs=0, const=True` unit test for `--weight_only`, the `inspect.getsource` test replaced with `parser._actions` introspection, and the CHANGELOG "all five"/six-flag mismatch corrected.
- Design is settled: a single `argparse.Action` shared by three CLIs, needed because argparse's `deprecated=True` requires 3.13 while `pyproject.toml` pins `>=3.10`.
- Not blocking: `examples/hf_ptq/scripts/huggingface_example.sh` still passes `--qformat` on the non-recipe path, so example runs will now print the deprecation — expected, but worth migrating the shipped scripts/READMEs in a follow-up.

Complex PR: spans 6 directories (≥ 5); 1 existing test file modified or removed. Looping in a human for approval.

ajrasane (approved):

sugunav14 (approved):

PR #2431, opened 2026-09-14
Add pyrefly tool config section

shengliangxu:
### What does this PR do?

Type of change: Misc (developer tooling / editor config)

Adds the first `[tool.pyrefly]` configuration section to `pyproject.toml`. Specifically, the change **enables pyrefly's fallback search path** (`enable-fallback-search-path = true`) so that example scripts' sibling imports are resolved.

Example scripts import sibling helper modules as top-level modules — e.g. `from example_utils import ...` in `examples/hf_ptq/hf_ptq.py`, and `cast_mxfp4_to_nvfp4`. At runtime these resolve because Python puts the script's own directory on `sys.path`, but pyrefly infers the import root as the repo root and flags them as `missing-import`. Enabling the fallback search path makes pyrefly walk up from each importing file's directory to resolve imports the same way — fixing every example dir's sibling imports without enumerating each one in `search-path` (and with no cross-directory name collisions).

Config-only and inert for anyone without pyrefly: `[tool.*]` tables are namespaced, so other tools ignore it; the repo already disables even-better-toml schema validation, and the `check-toml` pre-commit hook validates syntax only.

### Usage

Not an API change. With the pyrefly VS Code extension (or CLI), open `examples/hf_ptq/hf_ptq.py` — `example_utils` now resolves instead of raising `Cannot find module 'example_utils'`.

### Testing

- `pyrefly check examples/hf_ptq/hf_ptq.py` with the auto-discovered config: `missing-import` errors for `example_utils` and `cast_mxfp4_to_nvfp4` → **0**.
- `pyrefly dump-config` confirms the config is picked up and the fallback search path is active (`[…/examples/hf_ptq, …/examples, …/main]`).
- Confirmed inert for non-pyrefly setups (namespaced `[tool.*]`, even-better-toml schema disabled, `check-toml` = syntax-only).

- Did you write any new necessary tests?: N/A (config-only, no runtime code)
- Did you update Changelog?: N/A (developer tooling; not a user-facing feature/fix)


pyproject.toml
```diff
@@ -373,3 +373,11 @@ exclude_lines = [
     # Don't complain about abstract methods, they aren't run
     "@(abc\\.)?abstractmethod",
 ]
+
+[tool.pyrefly]
```
kevalmorabia97, line 377: do we need to also install the dependency in dev deps?

PR #2439, opened 2026-09-15
Document the nvfp4_act_headroom calibration variant in ptq.md

cjluo-nv:
### What does this PR do?

Type of change: documentation

`general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml` appears in the shipped-recipes
table in `modelopt_recipes/ptq.md`, but the **Calibration variants** section —
which documents `max`, `mse`, `input_scale1`, `gptq`, and the `layerwise`
variants — had no entry for it. Someone scanning that section for "which
calibration do I pick when NVFP4 W4A4 regresses?" only found `mse`, which
searches **weight** scales and so cannot help when the loss comes from
activation clipping.

This adds the missing entry: the scale formula
(`amax = max(rho * anchor, upper)`) and its defaults, the fact that it costs one
calibration pass and exports a standard NVFP4 checkpoint with coverage identical
to `nvfp4_default-kv_fp8_cast`, and the symptoms that should route you here
rather than to a weight-side calibration — an A16 ablation clears the regression
while `mse` does not, the symptom is behavioral (verbose or runaway generations,
hitting the generation cap) rather than a flat score drop, inference contexts run
longer than the calibration set, or a few rare blocks dominate the activation
error. MoE experts-only scopes are called out as the common case.

It also extends step 3 of **Choosing a general recipe** so the escalation path
reads `mse` first, then `nvfp4_act_headroom` when the evidence points at
activations rather than weights.

**Evidence.** The guidance comes from a GLM-5.3-Flash NVFP4 experts-only W4A4
root-cause study on SciCode (temperature 1.0), which established causally that
activation quantization at the routed-expert `down_proj` input drove a large
generation-length blow-up. Swapping `max` for `nvfp4_act_headroom` cut the median
generation-length regression versus source from +38% to +19% and the mean from
+19% to +4%, with no capped generations. The entry states plainly that this was
the best strict-W4A4 result in that study but still missed the p50/p75
near-lossless gate, so headroom is presented as a strong first lever for
activation-driven regressions rather than a guaranteed fix, with a note that
`rho` should be swept.

### Usage

No API or recipe change; the recipe already ships. This PR only documents when
to select it over plain `max`:

```python

cfg = load_recipe("general/ptq/nvfp4_act_headroom-kv_fp8_cast")
```

### Testing

Docs-only change; no code paths touched, so no new or updated tests.

- `pre-commit run --files modelopt_recipes/ptq.md` — all applicable hooks pass,
  including `markdownlint-cli2` and `check-modelopt-recipes`.
- Re-read the rendered section to confirm the new bullet nests correctly in the
  existing `Calibration variants` list and that surrounding entries are unchanged.
- Cross-checked every claim against the implementation
  (`modelopt/torch/quantization/calib/nvfp4_act_headroom.py`), the recipe YAML,
  and the existing `CHANGELOG.rst` entry, so the documented defaults
  (`anchor_percentile=1`, `upper_percentile=99.99`, `rho=16384`) and the
  NVFP4-input-quantizer-only scope match the code.

### Additional Information

The entry deliberately does not sell this on accuracy: in that study the
quantized subtask accuracy (56.80%) was *above* the source checkpoint (51.18%),
so what headroom recovered was generation-length behavior, not accuracy.

## Summary by CodeRabbit

* **New Features**
  * Added an NVFP4 activation headroom calibration option with configurable percentile-based scaling.
  * Supports standard NVFP4 checkpoint export with a single calibration pass.
  * Applies to dynamic-block NVFP4 activation quantizers while keeping weight-scale configuration independent.
  * Added guidance for addressing activation-related W4A4 accuracy regressions, including calibration coverage and recipe selection.


meenchen (approved):

Approving: docs-only addition whose technical claims I verified line-by-line against the implementation, recipe YAML, and existing tests.

- Defaults (`anchor_percentile=1`, `upper_percentile=99.99`, `rho=16384`) and the `amax = max(rho * anchor, upper)` formula match `calib/nvfp4_act_headroom.py` and `general/ptq/nvfp4_act_headroom-kv_fp8_cast.yaml`.
- "NVFP4 input quantizers only / no-op for FP8 and weight-only" matches `_is_nvfp4_dynamic_input_quantizer` and the empty-swap warning in `model_calib.nvfp4_act_headroom_calibrate`; the `weight_scale_algorithm` `max`/`mse`/`local_hessian` dispatch matches `_run_weight_scale_calibration`.
- No licensing changes, no code paths touched, no test edits.

No action needed.

PR #2441, opened 2026-09-15
docs: fix evaluation output token budget guidance

chadvoegele:
### What does this PR do?

Type of change: documentation.

Fix `max_new_tokens` guidance:
- Non-reasoning: 16,384 tokens, reduced for smaller context/output limits.
- Reasoning: maximum output allowed, as disclosed by the model creators.
- A model-card budget explicitly used for the applicable evaluation can override these defaults.

Remove generic 64K reasoning and same-family token fallbacks, and the rule to pick the highest number anywhere in a card. Align the output-budget example and validation guidance. Other sampling and deployment settings are unchanged.

### Usage

No API changes. Follow the evaluation skill's output-token rule when preparing configs.

### Testing

- Focused pre-commit hooks and `git diff --check` passed.
- **Before PR:** the available pre-PR baseline (`28dc117594ee98c8875e8e0c9bb4496a4e259475`, September 15) passed **0/1 scored Harbor unit trials**. It generated `max_new_tokens: 32768`; the verifier required `131072`. No five-trial pre-PR baseline was found. Post-PR startup errors are excluded from behavioral pass rates; this historical comparison is not a controlled five-versus-five experiment.
- Rebuilt the temporary agent image at narrowed revision `37fe654ff29b645f77a549d28d21196a640be428` and ran exactly five fresh NEL-config Harbor unit trials against the unchanged task/verifier: **4/5 passed (80%)**, with zero infrastructure errors.
- **5/5 generated `max_new_tokens: 131072`**, temperature `1.0`, and top-p `0.95`. The failed trial substituted a hosted model ID for the exact requested BF16 model; verification failed with `unexpected model ID set`, not a token-budget error. No corrective reruns or model evaluation submissions.
- Earlier broader revision: 135 deterministic tests passed. Five trials are a small sample, not a reliability guarantee.

- Is this change backward compatible?: N/A — skill guidance.
- Copied code or new dependency?: N/A
- New necessary tests?: N/A — documentation; narrowed-revision behavioral results above.
- Changelog updated?: N/A
- Claude approval?: No.

### Additional Information

Fixes the output-token guidance issue documented in [Nemotron 3 Super GPQA reasoning-token limit (internal analysis)](https://gitlab-master.nvidia.com/cvoegele/lab/-/blob/main/wiki/day0_jul10_analysis/day0_jul10_analysis.md?plain=1#L217).

Prepared by **Chad's Agent**.


chadvoegele:
**Chad's Agent:** I checked the Harbor trajectories to trace where `max_new_tokens: 131072` came from.

The BF16 model card explicitly links its evaluation section to NVIDIA's model-specific NEL configs and reproducibility tutorial. In trial 1, the agent read that link, listed the directory through GitHub's API, then fetched [`local_nemotron-3-super-120b-a12b.yaml`](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/packages/nemo-evaluator-launcher/examples/nemotron/nemotron-3-super/local_nemotron-3-super-120b-a12b.yaml) and [`reproducibility.md`](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/packages/nemo-evaluator-launcher/examples/nemotron/nemotron-3-super/reproducibility.md). Both explicitly specify `131072`.

All five trials fetched the published config and wrote that value into `evaluation.nemo_evaluator_config.config.params.max_new_tokens`. This was a published evaluation budget discovered through model-card research—not a guessed filename, a context-length inference, or an automatic NEL default. It is the recipe's shared generation default used by GPQA, not a uniquely GPQA-specific setting.

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -326,13 +328,10 @@ overridden task is reported under sampling params it did not use.
...
-5. **Forbidden:** writing `max_new_tokens: <generic_default>` with a "card not yet checked" comment. Either fetch and apply, or fetch and confirm silence.
-6. **A higher cap doesn't fix runaway reasoning.** On hard tasks (e.g. HLE) a non-terminating model just rambles to the larger cap (~80% length-capped at 131072), and the cap only helps if deployment `--max-model-len > prompt + max_new_tokens` (else generation is silently clipped — AA-LCR's ~120K input leaves little room). Treat such tasks as low-confidence.
+1. **Read the HF model card before setting the value.** Identify the model's reasoning mode, creator-disclosed maximum output length, and any `max_tokens` / `max_new_tokens` used for the applicable evaluation. Cite the source and rationale in a config comment; quickstart examples are not evaluation budgets.
+2. **Non-reasoning:** use **16384**, lowered for smaller output caps or available context after input tokens. **Reasoning:** use the maximum output length allowed and disclosed by the model creators. Do not substitute 65536, a same-family budget, or the context-window size for an undisclosed output maximum.
```
cjluo-nv, line 332: > _Bot comment._

Rule 2 rules out 65536, same-family budgets and the context window, but the old rule 4 fallback ("if the card is genuinely silent … surface the silence to the user") is gone. Many cards in `references/nvfp4-modelcard-sampling.md` have `—` in `max_num_tokens`, so "creator-disclosed maximum output length" is frequently unavailable. Please state the resolution path (ask the user / leave `???` and escalate) so the agent isn't left without a value.

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -326,13 +328,10 @@ overridden task is reported under sampling params it did not use.
...
+3. **An explicit model-card budget used for the applicable evaluation can override these defaults.** Do not apply another benchmark's budget or choose the highest number mentioned. Keep one top-level `max_new_tokens` (no per-task overrides); conflicting benchmark-specific budgets require separate configs.
+4. Verify that the output budget fits alongside the input, including accumulated multi-turn history. Surface conflicts rather than silently clipping a creator/evaluation budget. Check `finish_reason: length` after the run (`references/run-validation.md`); a higher cap does not fix runaway reasoning.
```
cjluo-nv, line 334: > _Bot comment._

This replaces a concrete, hard-won constraint (`n_turns × max_new_tokens + prompt < max_model_len`, plus the SciCode HTTP 400 / AA-LCR ~120K examples) with "verify that the output budget fits". Worth keeping the formula and at least one example — with the new rule a reasoning model will often get `max_new_tokens: 131072` against the template's `--max-model-len 131072`, which cannot fit any prompt, and "surface conflicts" doesn't say which side gives.

plugins/modelopt/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -207,7 +207,7 @@ export:
       model: CHANGEME-served-model-name
       temperature: '1.0'
       top_p: '0.95'
-      max_new_tokens: '65536'
+      max_new_tokens: ???
```
cjluo-nv, line 210: > _Bot comment._

Sibling tags are quoted strings (`'1.0'`, `'0.95'`) and the comment below stresses keeping quotes; use `max_new_tokens: '???'` so the filled-in value keeps the string form MLflow tags expect.

cjluo-nv (commented):

Direction is right (card-disclosed budget, eval-specific override only), but the rewritten rule drops two things it needed to keep: a fallback when no output maximum is disclosed, and the concrete multi-turn/context fit constraint.

Needs action:
- [ ] Say what to do when a reasoning card discloses no output maximum — `SKILL.md` rule 2 forbids 65536, family values and the context window, and the old "surface the silence to the user" fallback is gone, so the agent has no resolution path. Mirror it in `references/model-card-research.md`.
- [ ] Restore the concrete fit constraint in `SKILL.md` rule 4 (`n_turns × max_new_tokens + prompt < max_model_len`, SciCode HTTP 400 precedent); "verify the budget fits" loses the arithmetic.
- [ ] Reconcile "use the disclosed maximum" (often 131072+) with the template's `--max-model-len 131072` and `analyze-results.md` item 5 — say whether to raise max-model-len or lower the budget, not just "surface conflicts".
- [ ] Quote the placeholder in `example_eval.yaml` tags (`max_new_tokens: '???'`) to match the neighbouring quoted `temperature`/`top_p` tags and the keep-the-quotes comment.

plugins/modelopt/skills/evaluation/recipes/examples/example_eval.yaml
```diff
@@ -132,7 +132,7 @@ evaluation:
         parallelism: ???  # Number of concurrent requests per each benchmark
         request_timeout: 3600
         max_retries: 10
-        max_new_tokens: 65536  # 64K for reasoning models; use 16384 (16K) for non-reasoning; prefer model card value
+        max_new_tokens: ???  # Resolve per SKILL.md's token-budget rule; match the export description/tag below
```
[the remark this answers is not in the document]
chadvoegele: **Chad's Agent — suggestion for review (not applied):** Reasonable small improvement, though the existing Step 3 reference already requires provenance. I suggest extending this comment with `cite the creator/evaluation source and rationale when filling in the value`, rather than adding more placeholder fields. That also covers sources linked from the model card, such as the published NEL recipe used in all five trials.

plugins/modelopt/skills/evaluation/references/nvfp4-modelcard-sampling.md
```diff
@@ -15,6 +15,10 @@ is silent or ambiguous. It does not replace reading the card — see
 
 ## Lookup
 
+**The lookup and family priors below govern `temperature` / `top_p`, not output
+budgets.** For `max_new_tokens`, follow SKILL.md Step 3's token-budget rule;
+verify table values against the card and the applicable evaluation.
```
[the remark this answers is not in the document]
chadvoegele: **Chad's Agent — suggestion for review (not applied):** The new heading already explicitly excludes output budgets from family fallbacks, so I see this as a clarity improvement rather than a missing policy. Suggest changing `Exact row, resolved per field` to `Sampling columns (temperature/top_p), resolved per field` and `No row` to `No sampling row`. Keep output budgets governed solely by Step 3; no need to rewrite the table or broaden the PR.

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -326,13 +328,10 @@ overridden task is reported under sampling params it did not use.
...
+1. **Read the HF model card before setting the value.** Identify the model's reasoning mode, creator-disclosed maximum output length, and any `max_tokens` / `max_new_tokens` used for the applicable evaluation. Cite the source and rationale in a config comment; quickstart examples are not evaluation budgets.
+2. **Non-reasoning:** use **16384**, lowered for smaller output caps or available context after input tokens. **Reasoning:** use the maximum output length allowed and disclosed by the model creators. Do not substitute 65536, a same-family budget, or the context-window size for an undisclosed output maximum.
+3. **An explicit model-card budget used for the applicable evaluation can override these defaults.** Do not apply another benchmark's budget or choose the highest number mentioned. Keep one top-level `max_new_tokens` (no per-task overrides); conflicting benchmark-specific budgets require separate configs.
```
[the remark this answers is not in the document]
chadvoegele: **Chad's Agent — suggestion for review (not applied):** The adapter caveat is valid: the Gym examples remove both outgoing cap fields, so a YAML value alone does not prove that the requested budget reaches the model. Suggest one sentence in the token validation guidance: `Verify the effective output limit after adapters/interceptors; document intentional omission of the request cap and the remaining server limits.` I would not equate removal with truly unbounded output or require a particular payload field universally: server defaults/context and harness-specific controls still apply. Keep this PR to the validation warning; do not change Gym recipes without checking their request path.

plugins/modelopt/skills/evaluation/SKILL.md
chadvoegele: **Chad's Agent — suggestion for review (not applied):** There is a real unresolved case, but I would not restore ask-user/escalation language: its removal was explicitly requested for this PR. Suggest first following the exact model card's linked evaluation recipes/configs (the successful Nemotron trials found 131072 that way). If no authoritative budget is available, leave `???` unresolved and do not treat the config as launch-ready, rather than inventing a value. This makes the failure mode explicit without reintroducing a user prompt or a generic fallback.

plugins/modelopt/skills/evaluation/SKILL.md
chadvoegele: **Chad's Agent — suggestion for review (not applied):** Agree that a concrete capacity check helps. I would use `current input tokens (including retained history) + requested output tokens <= max_model_len`, rather than restore `n_turns × max_new_tokens + prompt`: that is only a conservative worst case and ignores actual turn lengths, tool output, and truncation/compaction. A useful example is that a 131072 output budget cannot fit any nonempty prompt in a 131072 context. Prefer a larger supported serving context; if unavailable, mark the requested budget infeasible instead of silently shrinking it. No generic context override is needed.

plugins/modelopt/skills/evaluation/recipes/examples/example_eval.yaml
chadvoegele: **Chad's Agent — suggestion for review (not applied):** Agree: quote the export tag placeholder as `max_new_tokens: '???'` and retain quotes when replacing it, matching the neighboring MLflow tags. Keep the actual evaluation parameter numeric. Quoting is a type/style cue for the completed tag, not a way to make an unresolved OmegaConf placeholder valid.

chadvoegele:
## Chad's Agent — independent review

Reviewed head `637a1026cd4078e5879cd14871132a5442251288`.

**No actionable findings introduced by this diff.**

Checks performed:
- Read all five changed files, the entire diff, repository instructions, and surrounding evaluation/validation guidance.
- Verified Nemotron 3 Super's model-card-linked [GPQA evaluation config](https://github.com/NVIDIA-NeMo/Evaluator/blob/aac6fbaf5e9fbeaa06639fcd8bef20a67c741783/packages/nemo-evaluator-launcher/examples/nemotron/nemotron-3-super/local_nemotron-3-super-120b-a12b.yaml#L86-L119) uses `max_new_tokens: 131072`, distinct from quickstart budgets and context length.
- Parsed the example with OmegaConf; confirmed both budget placeholders are mandatory. Rendered the GPQA command with 16384 and 131072: both propagate unchanged, with no token-removing adapter in this template.
- Inspected client token mapping and exercised vLLM's context-clipping function in isolation: a 131072 output budget needs additional prompt capacity, as the revised rule requires.
- `git diff --check` passed; shared checkout unchanged.

Limits: implementation checks used local Evaluator/NeMo-Skills/vLLM sources, not the pinned evaluation containers. No GPU evaluations or Harbor behavioral trials were rerun; this does not independently verify the PR's reported pass rate.

PR #2447, opened 2026-09-16
[OMNIML-5899] Export IQ checkpoints from HF and Megatron

hychiang-git:
## Summary

- add IQ format metadata and packed-weight export
- support Hugging Face and TP=1 Megatron export paths
- reject fused-MoE IQ export until a deployment loader owns its packed layout
- document the shaped `uint8` weight contract and the fused-expert boundary
- add Hugging Face, Megatron, metadata, and fused-expert export tests

## PR split

This work is split into four focused PRs. Each PR targets `main` and owns a disjoint file set:

1. **Kernel** — [#2448: Add CUDA kernels for IQ packing](https://github.com/NVIDIA/Model-Optimizer/pull/2448)
2. **Quantization** — [#2446: Add IQ quantization codecs and backend](https://github.com/NVIDIA/Model-Optimizer/pull/2446)
3. **Export** — [#2447: Export IQ checkpoints from HF and Megatron](https://github.com/NVIDIA/Model-Optimizer/pull/2447)
4. **Recipes** — [#2449: Add IQ post-training quantization recipes](https://github.com/NVIDIA/Model-Optimizer/pull/2449)

The required merge order is #2448, #2446, #2447, then #2449.

## Scope

This PR owns only export code, deployment documentation, and export tests. It targets `main` and should merge after #2448 and #2446. It does not contain kernel, codec/backend, or recipe files.

## Deployment consumer boundary

Dense weights and individually named expert weights use the documented shaped `uint8` contract. Megatron fused-MoE IQ export is intentionally rejected with `NotImplementedError`: its payload would have shape `[num_experts, out_features, in_features // 256, payload_bytes]`, and no deployment loader in this stack currently owns that layout. Support should be enabled only with a loader integration test.

## Test coverage

- [Hugging Face packed-weight export](https://github.com/NVIDIA/Model-Optimizer/blob/11cd58d907465933f5a552bc1a8065f84c9ba3b1/tests/unit/torch/export/test_export_weight.py)
- [quantization metadata](https://github.com/NVIDIA/Model-Optimizer/blob/11cd58d907465933f5a552bc1a8065f84c9ba3b1/tests/unit/torch/export/test_get_quantization.py)
- [Megatron unified export and fused-MoE rejection](https://github.com/NVIDIA/Model-Optimizer/blob/11cd58d907465933f5a552bc1a8065f84c9ba3b1/tests/gpu_megatron/torch/export/test_unified_export_megatron.py)

## Validation

- all pre-commit hooks pass for the changed files
- 89 focused Hugging Face export, metadata, and fused-expert tests pass locally
- direct checks cover both fused-MoE export entry points for IQ1_S and IQ2_XS
- Megatron GPU execution remains delegated to GPU CI
- restricted-term scan passes

## Summary by CodeRabbit

* **New Features**
  * Added support for IQ1_S and IQ2_XS GGML quantization formats in unified Hugging Face and Megatron exports.
  * Added quantization metadata, tensor-shape recovery, packing details, and IQ2_XS size documentation.
  * Added validation for required block sizes and tensor parallelism settings.

* **Limitations**
  * Fused-MoE and GPT-OSS IQ expert packing are not supported.
  * IQ exports require standard `weight` attributes in Hugging Face models.


modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1846,7 +1910,9 @@ def _pack_name_remapping(self, module, prefix, layer_type=None, is_mtp=False, tr
             merged_input_scale = None
 
         # Save the merged weights
-        if merged_weight_scale is None:
+        if qformat in (QUANTIZATION_IQ1_S, QUANTIZATION_IQ2_XS):
+            self._state_dict.update(self._get_iq_weight_state(prefix, merged_weight, qformat))
```
cjluo-nv: > _Bot comment._

`merged_weight` has already been transposed to `[E, in, out]` above, so `quantize_iq*` blocks the last dim — the *output* features — into 256-element GGML blocks. Everywhere else (HF `_export_quantized_weight`, `_name_remapping`, `_gated_mlp_slicing`) the blocks run along the input dim, which is also what the `weight_quantizer` used during calibration/fake-quant. That means the exported experts are quantized on a different axis than the model was calibrated for, and `out % 256 != 0` would raise from `validate_weight` even when `in` is a valid multiple. Please pack before the transpose (and re-pack/reshape into the HF layout), or document why the output-dim grouping is correct here. Same applies to `_pack_name_remapping_gpt_oss`, where the gate/up interleave also happens on that last dim before packing.

modelopt/torch/export/quant_utils.py
```diff
@@ -1152,6 +1167,7 @@ def _export_key(key: str) -> str:
         # (pre_quant_scale is the AWQ / NVFP4_AWQ / SVDQuant companion, renamed in the KV-cache pass.)
         weight_suffixes = (
             "weight",
+            "weight_shape",
```

`_BASE_SKIP_KEYS` already contains `"weight_shape"`, so any key containing it is dropped before the tied-weight dedup runs and can never appear in `post_state_dict` — the `members` filter `if ak in post_state_dict` makes this entry dead. IQ export also discards the logical-shape tensor (`packed_weight, _ = quantize_iq(...)`), which the new test asserts. Suggest removing this line.

modelopt/torch/export/quant_utils.py
```diff
@@ -722,6 +729,14 @@ def process_layer_quant_config(layer_config_dict):
                 "quant_algo": "MXFP8",
                 "group_size": block_size_value,
             }
+        elif v in (QUANTIZATION_IQ1_S, QUANTIZATION_IQ2_XS):
+            payload_bytes = 50 if v == QUANTIZATION_IQ1_S else 74
```

`50` / `74` / `256` are re-derived here, again in `convert_hf_config._quant_algo_to_group_config`, and again in the docs. `modelopt.torch.quantization.ggml` already exports `IQ1_S_BLOCK_BYTES`, `IQ2_XS_BLOCK_BYTES`, `IQ1_S_EFFECTIVE_BITS`, `IQ2_XS_EFFECTIVE_BITS` and `GGML_BLOCK_SIZE` — importing them keeps the metadata in sync with the codec if a payload layout ever changes. Also note `group_size` is hardcoded to 256 here while the sibling branches use `block_size_value` from the recorded `awq_block_size`; asserting they agree would catch a misconfigured quantizer.

cjluo-nv (commented):

Commenting: the Megatron packed-expert paths pack IQ blocks after the `[E, out, in] -> [E, in, out]` transpose, so blocks form along the wrong axis, and five of the six new Megatron branches have no test.

Needs action:
- [ ] Confirm the block axis in `_pack_name_remapping` / `_pack_name_remapping_gpt_oss` (`unified_export_megatron.py:1914`): packing after the transpose groups 256 weights along the output dim, unlike fake-quant and the HF path. Pack before the transpose or explain why not.
- [ ] Add tests for the untested IQ branches: `_gated_mlp_slicing`, `_grouped_mlp_slicing`, `_qkv_slicing`, `_gated_delta_net_slicing`, `_pack_name_remapping` — only `_name_remapping` is covered.
- [ ] Replace the hardcoded `50`/`74`/`1.5625`/`2.3125`/`256` in `quant_utils.py:733` and `convert_hf_config.py:121` with `IQ*_BLOCK_BYTES` / `IQ*_EFFECTIVE_BITS` / `GGML_BLOCK_SIZE` from `modelopt.torch.quantization.ggml`.
- [ ] Drop the `"weight_shape"` entry added to `weight_suffixes` in `quant_utils.py:1170` — `_BASE_SKIP_KEYS` already removes any `weight_shape` key, and IQ export never emits one.

No action needed:
- Merge order: this stack sits on #2446, whose llama.cpp/ggml MIT provenance is still awaiting OSRB sign-off.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1842,7 +1906,9 @@ def _pack_name_remapping(self, module, prefix, layer_type=None, is_mtp=False, tr
...
```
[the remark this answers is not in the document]
hychiang-git: Fixed in f2d3a834a. Fused experts are now packed per expert while each logical weight is still [out, in], so 256-value blocks remain on the contraction axis. Packed CPU payloads are stacked without a post-pack transpose. Exact packed-byte coverage was added for the standard and GPT-OSS expert paths.

modelopt/torch/export/unified_export_hf.py
```diff
@@ -621,6 +624,21 @@ def _export_quantized_weight(
...
+            raise NotImplementedError(
+                "IQ unified export currently supports modules with a standard 'weight' "
+                f"attribute, got {weight_name!r} on {type(sub_module).__name__}"
+            )
+        quantize_iq = (
+            quantize_iq1_s if quantization_format == QUANTIZATION_IQ1_S else quantize_iq2_xs
+        )
+        packed_weight, _ = quantize_iq(weight.to(dtype))
+        delattr(sub_module, weight_name)
+        sub_module.register_buffer("weight", packed_weight)
```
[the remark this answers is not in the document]
hychiang-git: Fixed in f2d3a834a. IQ export now replaces weight with nn.Parameter(..., requires_grad=False), matching the other packed formats. The fused-expert buffer special case and its test were removed.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -1027,6 +1042,7 @@ def _get_weight_bias(
         module: torch.nn.Module,
         dtype: torch.dtype = torch.float16,
         name_to_value: dict[str, torch.Tensor] | None = None,
+        keep_weight_device: bool = False,
```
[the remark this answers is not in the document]
hychiang-git: Fixed in f2d3a834a. Each expert is packed immediately and moved to CPU before stacking. The export no longer builds a full fused floating-point stack and contiguous transpose on GPU.

modelopt/torch/export/quant_utils.py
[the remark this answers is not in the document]
hychiang-git: The dead weight_shape suffix was removed in f2d3a834a. The proposed logical-shape helper changes the codec API and belongs in #2446, so it is intentionally kept out of this export PR.

modelopt/torch/export/convert_hf_config.py
```diff
@@ -117,6 +117,19 @@ def _quant_algo_to_group_config(quant_algo: str, group_size: int | None = None)
...
+        return {
+            "weights": {
+                "dynamic": False,
+                "num_bits": 1 if quant_algo == "IQ1_S" else 2,
+                "effective_bits": effective_bits,
+                "type": "int",
+                "group_size": 256,
+                "packing": "ggml",
+                "block_payload_bytes": payload_bytes,
+            }
```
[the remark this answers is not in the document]
hychiang-git: Fixed in f2d3a834a. IQ metadata no longer presents the payload as a compressed-tensors integer weights group. Uniform exports carry ModelOpt-owned top-level format metadata, while mixed exports use an IQ group without a weights schema. Tests cover both forms.

modelopt/torch/export/unified_export_megatron.py
hychiang-git: Fixed in f2d3a834a by packing each expert before any layout transpose. For GPT-OSS, gate/up interleaving happens on the logical output rows before packing, then the CPU payloads are stacked. Exact-byte tests cover both expert paths.

modelopt/torch/export/quant_utils.py
hychiang-git: Removed in f2d3a834a. IQ export does not emit a separate weight_shape tensor, so the tied-weight suffix entry was dead.

modelopt/torch/export/quant_utils.py
hychiang-git: Fixed in f2d3a834a. Export metadata now imports the per-format block-byte, block-size, and effective-bit constants from the codec package. It also verifies that the recorded block size matches the format constant, with a mismatch regression test.

hychiang-git:
Addressed review 5229455007 in f2d3a834a:

- corrected the fused-expert block axis and avoided full GPU expert-stack copies;
- added exact packed-byte coverage for gated MLP, grouped MLP, QKV, GatedDeltaNet, standard packed experts, and GPT-OSS packed experts;
- sourced export metadata from the codec constants and added block-size validation;
- removed the dead weight_shape suffix.

Validation: 89 focused unit tests passed, direct Megatron slicing checks passed, and repository pre-commit hooks passed. The full Megatron test harness was not available in the local environment; CI is running.

cjluo-nv (commented):

Nudge: every concern from the last round is fixed in `f2d3a834a`, but the stack still needs a human for the ggml licensing call and the unanswered design question.

Needs action:
- [ ] Get OSRB/legal sign-off on the llama.cpp/ggml MIT codebook and the new `LICENSE` entry (carried in from #2446) before this stack merges.
- [ ] Explain in the PR body (or #2446's) why a new `modelopt/torch/quantization/ggml/` package is preferred over `BaseQuantizedTensor` subclasses in `modelopt/torch/quantization/qtensor/` — the diff under review adds both the package and its export consumers.
- [ ] Confirm the deployment consumer accepts fused-MoE IQ payloads in Megatron order `[E, out, in//256, bytes]` (`_pack_name_remapping`), since every other format transposes to HF `[E, in, out]`; the `3_unified_hf.rst` note documents it but no loader is exercised.
- [ ] Confirm merge order against #2446 / #2448.

No action needed:
- :heavy_check_mark: Resolved since the last review: the fused-expert block axis (now packed per expert before any transpose, with exact-byte tests for the standard and GPT-OSS paths), tests for the five previously uncovered Megatron IQ branches, codec-sourced block/byte/effective-bit constants plus a block-size mismatch check, the dead `weight_shape` suffix, the `Parameter`-vs-buffer regression, and the fused-expert GPU stack allocation.

hychiang-git:
Addressed the fused-MoE deployment-consumer concern in 1b89d3e26.

Rather than emitting an unverified `[num_experts, out_features, in_features // 256, payload_bytes]` tensor, both Megatron fused-expert export entry points now raise `NotImplementedError` for IQ1_S and IQ2_XS before writing any checkpoint state. Dense weights and individually named expert weights remain supported. The deployment documentation now states this boundary, and the Megatron export tests cover both rejection paths.

Fused-MoE IQ export can be enabled later when a deployment loader owns this layout and an integration test exercises it. The PR body also records the stack order: #2448, then #2446, then #2447.

hychiang-git:
/ok to test [1b89d3e](https://github.com/NVIDIA/Model-Optimizer/pull/2447/commits/1b89d3e263377e039c98e18f9755bcbc89179f22)

cjluo-nv (commented):

Nudge: the fused-MoE layout question is settled by an explicit rejection with tests, but licensing sign-off and one guard gap still need a human.

Needs action:
- [ ] Get OSRB/legal sign-off on the vendored ggml MIT codebook (`quantization/ggml/iq1_s.py`, `iq2_xs.py`) and the new `LICENSE` entry, carried in from #2446.
- [ ] Harden the TP>1 IQ check in `unified_export_megatron.py:319`: `_get_quantization_format(self.model)` returns only the first non-`NONE` format, so a mixed-format model whose IQ layers come later bypasses the guard and packs TP-sharded weights. Scan per-layer formats, or state why mixed IQ cannot occur.
- [ ] Confirm merge order against #2446 / #2448, and that #2446's body carries the `ggml/` vs `qtensor/` design rationale this PR depends on.

No action needed:
- :heavy_check_mark: Resolved since the last review: fused-MoE IQ export now raises `NotImplementedError` in both `_pack_name_remapping` and `_pack_name_remapping_gpt_oss` before any state is written, with tests on both entry points and the boundary documented in `3_unified_hf.rst`.

modelopt/torch/export/unified_export_megatron.py
```diff
@@ -312,10 +316,19 @@ def save_pretrained(
         is_last_stage_main_rank = pp_rank == pp_size - 1 and tp_rank == 0
         is_writer_rank = self._is_sidecar_writer_rank(is_last_stage_main_rank)
 
+        quantization_format = self._get_quantization_format(self.model)
+        if (
+            quantization_format in (QUANTIZATION_IQ1_S, QUANTIZATION_IQ2_XS)
```
cjluo-nv: > _Bot comment._

Still open from the previous round. `_get_quantization_format` walks the module tree and returns the *first* non-`QUANTIZATION_NONE` format, so on a mixed-format model whose IQ layers come after e.g. an FP8 layer this condition is false and the export proceeds to pack TP-sharded weights — either a confusing divisibility error or a silently sharded IQ payload.

Either scan the effective per-layer formats here (the same walk `get_quant_config` does), or gate on `get_tensor_model_parallel_world_size() != 1` first and then check whether *any* module resolves to an IQ format. If mixed IQ is structurally impossible with the shipped recipes, a one-line comment saying so would close this out.

modelopt/torch/export/quant_utils.py
```diff
@@ -474,6 +484,11 @@ def _get_quantization_from_layer(layer, quantizer_attr_names: QuantizerAttrNames
             return QUANTIZATION_W4A8_AWQ
 
         # Handle individual num_bits cases
+        if weight_quantizer.num_bits in (QUANTIZATION_IQ1_S, QUANTIZATION_IQ2_XS):
+            if weight_quantizer.backend != "ggml":
```
cjluo-nv, line 518: > _Bot comment._

This branch validates the backend but not the activation side, and both exporters return early for IQ before they collect `input_scale`: `_get_quantized_state` (`unified_export_megatron.py:1121`) returns above the `input_scale` block *and* above the existing `raise ValueError("Detect pre_quant_scale! ...")`, and `_export_quantized_weight` (`unified_export_hf.py:627`) returns before its own `input_scale` registration.

So a W-IQ + A-FP8 (or AWQ-smoothed) config calibrates cleanly and then exports as weight-only with `quant_algo: IQ2_XS` and no activation entry — nothing downstream can detect the loss. Since IQ is weight-only by construction, make that explicit right here, next to the backend check:

```python
if input_quantizer is not None and input_quantizer.is_enabled:
    raise ValueError(f"{weight_quantizer.num_bits} is weight-only; disable the input quantizer")
if getattr(input_quantizer, "_pre_quant_scale", None) is not None:
    raise ValueError("pre_quant_scale (SmoothQuant/AWQ) is not supported with IQ formats")
```

Both export paths then inherit the guard, and the two `is_iq` early returns become honestly weight-only.

modelopt/torch/export/convert_hf_config.py
```diff
@@ -209,6 +239,9 @@ def convert_hf_quant_config_format(input_config: dict[str, Any]) -> dict[str, An
             "targets": ["Linear"],
         }
         new_config["config_groups"] = {"group_0": config_group_details}
+    elif quant_algo_value in ("IQ1_S", "IQ2_XS"):
+        iq_metadata = _quant_algo_to_group_config(quant_algo_value)
```
cjluo-nv: > _Bot comment._

`_quant_algo_to_group_config` already validates `group_size` against the format's block size and raises on a mismatch, but this call site drops the caller's value, so an input config carrying `IQ2_XS` with `group_size: 128` is silently rewritten to 256 instead of raising. Pass it through:

```python
iq_metadata = _quant_algo_to_group_config(
    quant_algo_value, original_quantization_details.get("group_size")
)
```

The `MIXED_PRECISION` branch below already forwards `layer_group_size`, so this also keeps the two paths consistent. Worth a test alongside `test_iq_quantization_config_rejects_mismatched_block_size`, which covers the `process_layer_quant_config` side only.

cjluo-nv (commented):

Commenting: the TP>1 IQ guard I flagged last round is unchanged, and the new IQ branch in `get_quantization_format` still lets a quantized activation path export silently as weight-only.

Needs action:
- [ ] Harden the TP>1 check at `unified_export_megatron.py:319`: `_get_quantization_format(self.model)` returns only the first non-`NONE` format, so a mixed-format model whose IQ layers come later bypasses it and packs TP-sharded weights. Scan per-layer formats, or state why mixed IQ cannot occur.
- [ ] Reject an enabled `input_quantizer` / `_pre_quant_scale` in the IQ branch of `_get_quantization_from_layer` (`quant_utils.py:487`); both export paths return before `input_scale` collection and the existing `pre_quant_scale` guard. See inline.
- [ ] Forward the caller's `group_size` into `_quant_algo_to_group_config` at `convert_hf_config.py:242` so an invalid value raises instead of being rewritten to 256.
- [ ] Reword the `IQ1_S`/`IQ2_XS` bullets in `3_unified_hf.rst:53` — "importance-aware" implies imatrix support these encoders do not have.

No action needed:
- :heavy_check_mark: Resolved since the last review: fused-MoE rejection with tests on both entry points, and the merge order now recorded in the PR body.
- The ggml codebook / `LICENSE` OSRB call sits in #2446, outside this file set.

hychiang-git:
To address the failing checks:
1. Land #2448 — Kernel
2. Update/rebase #2446 onto main; rerun and land it
3. Update/rebase #2447 onto main; rerun its checks
4. Then handle #2449

cjluo-nv (commented):

Commenting: all four items from the last round are still present in the code at this commit — the rebase onto a `main` that now carries `modelopt/torch/quantization/ggml/` is the only visible change.

Needs action:
- [ ] Harden the TP>1 IQ guard at `unified_export_megatron.py:319`: `_get_quantization_format(self.model)` returns only the first non-`NONE` format, so a mixed-format model whose IQ layers come after e.g. an FP8 layer bypasses it and packs TP-sharded weights. Scan per-layer formats, or state why mixed IQ cannot occur.
- [ ] Reject an enabled `input_quantizer` / `_pre_quant_scale` in the IQ branch of `_get_quantization_from_layer` (`quant_utils.py:488`); both export paths return before `input_scale` collection and the existing `pre_quant_scale` guard, so W-IQ + A-FP8 exports silently as weight-only.
- [ ] Forward the caller's `group_size` into `_quant_algo_to_group_config` at `convert_hf_config.py:242` so an invalid value raises instead of being rewritten to 256; add a test alongside `test_iq_quantization_config_rejects_mismatched_block_size`.
- [ ] Reword the `IQ1_S`/`IQ2_XS` bullets in `3_unified_hf.rst:53` — "importance-aware" implies imatrix support these encoders do not have.

No action needed:
- :heavy_check_mark: Resolved earlier and still good: fused-MoE IQ rejection with tests on both entry points, and `ggml/` now resolving on this branch so the export imports load.

modelopt/torch/export/quant_utils.py
```diff
@@ -440,6 +450,27 @@ def get_weight_block_size(module: nn.Module, weight_name: str = "weight") -> int
...
+
+    This reads ``num_bits`` directly rather than resolving each layer's full format, so an
+    unrelated unsupported quantizer elsewhere in the model cannot turn the check into an error.
+    """
+    for weight_name in weight_attr_names(module):
+        weight_quantizer = representative_weight_quantizer(module, weight_name)
+        if (
+            weight_quantizer is not None
+            and weight_quantizer.is_enabled
+            and weight_quantizer.num_bits in (QUANTIZATION_IQ1_S, QUANTIZATION_IQ2_XS)
```

`representative_weight_quantizer` can return a `SequentialQuantizer` (W4A8_AWQ), and that class only delegates `fake_quant` / `is_enabled` / `amax` to its members — `num_bits` falls through to `nn.Module.__getattr__` and raises `AttributeError`. `is_enabled` delegates to `self[0]`, so the `and` chain does reach `.num_bits` for an enabled sequential quantizer.

This is now on the unconditional path: `save_pretrained` calls `uses_iq_quantization(self.model)` for every Megatron export, and W4A8_AWQ is otherwise supported there (`process_layer_quant_config` has a `w4a8_awq` branch, `to_quantized_weight` packs it). So a W4A8_AWQ Megatron export that works today would fail with an unrelated `AttributeError`.

The sibling helper in this file already handles it (`get_weight_block_size`: `if isinstance(weight_quantizer, SequentialQuantizer): weight_quantizer = weight_quantizer[0]`). Mirroring that, or `getattr(weight_quantizer, "num_bits", None) in (...)`, is enough — IQ is never a sequential format, so either reads as "not IQ". Worth a regression test with a W4A8-style config alongside `test_uses_iq_quantization_false_without_iq_layers`.

cjluo-nv (commented):

Commenting: all four items from the last round are fixed, but the new `uses_iq_quantization` helper reads `num_bits` off a quantizer that may be a `SequentialQuantizer`, which has no such attribute.

Needs action:
- [ ] Unwrap `SequentialQuantizer` (or use `getattr(..., "num_bits", None)`) in `uses_iq_quantization` (`quant_utils.py:468`) — `save_pretrained` now calls it on every Megatron export, so a W4A8_AWQ model raises `AttributeError` before any format dispatch. See inline.
- [ ] Add a non-IQ `SequentialQuantizer` case to `test_uses_iq_quantization_false_without_iq_layers`; the current coverage is FP8/IQ `TensorQuantizer` only.
- [ ] Confirm or fix rank-uniformity of the new guard (`unified_export_megatron.py:321`): `uses_iq_quantization(self.model)` is rank-local, so a PP stage with no IQ layer skips the raise and blocks in `_gather_exclude_modules()` while its peers exit. Checking TP size first, or all-reducing the flag, closes it.

No action needed:
- :heavy_check_mark: Resolved since the last review: the TP>1 mixed-format bypass, the W-IQ + enabled-activation-quantizer silent weight-only export (plus `pre_quant_scale`), `group_size` forwarding at `convert_hf_config.py:242` with a test, and the "importance-aware" wording in `3_unified_hf.rst`.

cjluo-nv (commented):

Nudge: all three items from the last round are fixed, but the fused-MoE rejection still has the rank-divergence problem the author just solved for the TP guard.

Needs action:
- [ ] Make `_reject_unsupported_fused_iq_export` rank-uniform in `unified_export_megatron.py`: it raises from inside the per-expert loop, so a PP stage (or EP rank) owning no fused expert skips the raise and blocks in `_gather_exclude_modules()` while peers exit. Reuse the `_any_rank_uses_iq_quantization()` pattern, or hoist it next to the TP guard.
- [ ] Confirm `save_pretrained` is entered by every rank before the new unconditional `all_gather_object` in `_any_rank_uses_iq_quantization()` — it now runs on every Megatron export, IQ or not.
- [ ] Report GPU/Megatron CI green on the rebased branch, and confirm the #2448 → #2446 → this merge order now that `quantization/ggml/` is on `main`.

No action needed:
- :heavy_check_mark: Resolved since the last review: the `SequentialQuantizer` `AttributeError` in `uses_iq_quantization` (now `getattr(..., "num_bits", None)`), its regression test, the rank-local TP>1 guard, `group_size` forwarding in `convert_hf_config.py`, and the "importance-aware" doc wording.

cjluo-nv (commented):

Nudge: the rank-divergence fix from the last round landed cleanly, but the fused-expert guard is now broader than the problem it solves and CI/merge order still needs confirming.

Needs action:
- [ ] Confirm the over-rejection in `save_pretrained` is intended: `_mappings_pack_fused_experts(all_mcore_hf_export_mapping[self.arch])` keys off the architecture's rule table, so an IQ dense-layer export on any fused-expert arch (Llama4, GPT-OSS) raises even when the experts themselves are BF16/FP8. Narrow it to the experts' own format, or say why mixed IQ+fused cannot occur.
- [ ] Add a test that `save_pretrained` raises the fused-MoE `NotImplementedError` — only `_mappings_pack_fused_experts` and the per-expert `_reject_unsupported_fused_iq_export` paths are covered today.
- [ ] Report GPU/Megatron CI green on the rebased branch and confirm the #2448 → #2446 → this merge order.

No action needed:
- :heavy_check_mark: Resolved since the last review: the rank-divergent fused-MoE rejection (now hoisted next to the TP=1 guard on the rank-uniform rule table, with unit tests), and the `_any_rank_uses_iq_quantization()` all_gather — `save_pretrained` already runs unconditional barriers/all-gathers, so every rank must enter it.
- ggml/llama.cpp OSRB sign-off sits in #2446, outside this file set.

cjluo-nv (commented):

Nudge: the fused-expert over-rejection is gone, but moving the guard back into the per-expert loop re-opens the rank-divergence hang the previous round had closed.

Needs action:
- [ ] :speech_balloon: Author reverted to a per-expert `_reject_unsupported_fused_iq_export` and documented why the arch rule table alone can't gate it — still needs rank-uniformity: an EP/PP rank owning no fused expert never enters the loop, skips the raise, and blocks in `_gather_exclude_modules()` while peers exit. Gate on the rank-uniform `use_packed_local_experts` / `pack_name_remapping*` rule AND an `all_gather`ed "experts are IQ" flag (the `_any_rank_uses_iq_quantization()` pattern), or state in the PR body that the hang is acceptable.
- [ ] Report GPU/Megatron CI green on the rebased branch and confirm the #2448 → #2446 → this merge order.

No action needed:
- :heavy_check_mark: Resolved since the last review: IQ dense-layer export on fused-expert architectures (Llama4/GPT-OSS) is no longer rejected, the guard now keys off the experts' own format, and both fused entry points are covered by tests asserting `_state_dict == {}`. The earlier ask for a `save_pretrained`-level fused-MoE test is moot now that the check no longer lives there.
- ggml/llama.cpp OSRB sign-off sits in #2446, outside this file set.

cjluo-nv (approved):

modelopt/torch/export/quant_utils.py
```diff
@@ -440,6 +450,30 @@ def get_weight_block_size(module: nn.Module, weight_name: str = "weight") -> int
...
+
+    This reads ``num_bits`` directly rather than resolving each layer's full format, so an
+    unrelated unsupported quantizer elsewhere in the model cannot turn the check into an error.
+    """
+    for weight_name in weight_attr_names(module):
+        weight_quantizer = representative_weight_quantizer(module, weight_name)
+        # getattr: a SequentialQuantizer has is_enabled but no num_bits, and is never IQ --
+        # IQ is a single quantizer with backend="ggml".
+        if (
+            weight_quantizer is not None
```
cjluo-nv, line 474: > _Bot comment._

This walk may not see grouped-MoE experts. `weight_attr_names(module)` yields `"weight"` only when `module.weight is not None`, and for other names only when `representative_weight_quantizer(module, name)` finds `<name>_weight_quantizer`. A `TEGroupedLinear` normally has no `weight` attribute (`_grouped_mlp_slicing` does `has_weight = hasattr(module, "weight")` and later `delattr`s the one it temporarily assigns) and holds `weight0..weightN` with a single `weight_quantizer` that is a `GroupedQuantizer` — so `weight_attr_names` yields nothing here, and the recursion into `weight_quantizer` / its child `TensorQuantizer`s has no weight either.

Net effect: a model whose IQ quantizers live only on grouped MoE experts (an expert-only recipe is a plausible shape for 1–2 bit formats) returns `False`, `save_pretrained` skips both the TP=1 and PP=1 raises, and `_grouped_mlp_slicing` then packs TP/ETP-sharded expert weights as if they were whole — the exact failure this guard exists to prevent.

`representative_weight_quantizer` already unwraps a `GroupedQuantizer` to `q[0]`, so handling this is mostly a matter of also considering `weight0` (or iterating the `GroupedQuantizer` members directly) when no standard weight attr is present. If grouped experts can never be the only IQ layers in a shipped recipe, a one-line comment saying so would close this out instead.

cjluo-nv (commented):

Nudge: last round's fused-expert rank-divergence hang is properly closed by the new rank-uniform PP=1/TP=1 guards, but the guard's detector looks blind to grouped-MoE quantizers.

Needs action:
- [ ] Confirm or fix `uses_iq_quantization` in `quant_utils.py` for `TEGroupedLinear`: it iterates `weight_attr_names`, which yields nothing for a module carrying `weight0..N` plus a `GroupedQuantizer`, so an experts-only IQ model returns `False` and TP>1 is never rejected. See inline.
- [ ] Add a test for the new PP>1 `NotImplementedError` in `tests/gpu_megatron/torch/export/test_unified_export_megatron.py`; only `test_megatron_iq_export_rejects_tensor_parallelism` covers the TP branch.
- [ ] Report GPU/Megatron CI green on the rebased branch and confirm the #2448 → #2446 → this merge order now that `quantization/ggml/` is on `main`.

No action needed:
- :heavy_check_mark: Resolved since the last review: the fused-MoE rank divergence — `_any_rank_uses_iq_quantization()` now decides TP=1 and the new PP=1 requirement uniformly before any collective, and the residual `ep_size > num_experts` case is documented in `_reject_unsupported_fused_iq_export`.

PR #2449, opened 2026-09-16
[OMNIML-5899] Add IQ post-training quantization recipes

hychiang-git:
## Summary

- add numerics configs for IQ1_S and IQ2_XS
- add model presets and general PTQ recipes
- document the supported weight-shape requirement
- add the release-note entry

## PR split

This work is split into four focused PRs. Each PR targets `main` and owns a disjoint file set:

1. **Kernel** — [#2448: Add CUDA kernels for IQ packing](https://github.com/NVIDIA/Model-Optimizer/pull/2448)
2. **Quantization** — [#2446: Add IQ quantization codecs and backend](https://github.com/NVIDIA/Model-Optimizer/pull/2446)
3. **Export** — [#2447: Export IQ checkpoints from HF and Megatron](https://github.com/NVIDIA/Model-Optimizer/pull/2447)
4. **Recipes** — [#2449: Add IQ post-training quantization recipes](https://github.com/NVIDIA/Model-Optimizer/pull/2449)

The required merge order is #2448, #2446, #2447, then #2449.

## Scope

This PR targets `main` and owns only eight recipe configuration, recipe documentation, and changelog files. It should merge after the kernel, quantization, and export PRs. It does not contain kernel, codec/backend, or export files.

## Integration contracts

- `block_sizes: {-1: 256}` records the native packed-block contract; it does not drive the GGML fake-quant scale search. After #2447 lands, the export path reads `TensorQuantizer.block_sizes[-1]` through `get_weight_block_size`, validates it against the format block size, and records `group_size: 256` in checkpoint metadata.
- The model presets intentionally expose `--qformat iq1_s` and `--qformat iq2_xs` in `hf_ptq`. This PR remains last in the merge order so those choices land only after #2447 provides packed-checkpoint export.

## Validation

- both shipped IQ recipes load with the expected backend and no unused search option
- YAML, Markdown, and changelog hooks pass for all 8 changed files
- restricted-term scan passes

## Summary by CodeRabbit

* **New Features**
  * Added GGML-compatible IQ1_S and IQ2_XS weight-only quantization using 256-value blocks.
  * Added ready-to-use PTQ recipes requiring no calibration data.
  * Added model presets for applying these formats to eligible linear layers.
  * Added support for exporting quantized models as separate packed checkpoints.

* **Documentation**
  * Updated PTQ documentation with activation requirements, dimension constraints, simulated quantization behavior, recipe coverage, and packed-checkpoint export guidance.


cjluo-nv (commented):

Nudge: the recipes are correct and reuse the existing numerics/preset/`$import` system (no new abstraction, so the design gate is satisfied), but the new YAMLs drop the canonical license header and nothing pins the hand-computed `effective_bits`.

Needs action:
- [ ] Replace the two-line SPDX header in the four new YAMLs (`configs/numerics/iq{1_s,2_xs}.yaml`, `configs/ptq/presets/model/iq{1_s,2_xs}.yaml`, `general/ptq/iq{1_s,2_xs}.yaml`) with the full `LICENSE_HEADER` text every sibling recipe YAML carries — or state that the short form is intentional.
- [ ] Add a test asserting each shipped IQ recipe resolves to `backend: ggml`, `num_bits: iq1_s`/`iq2_xs` and `effective_bits == IQ1_S_EFFECTIVE_BITS`/`IQ2_XS_EFFECTIVE_BITS`; the "Keep in sync with IQ1_S_BLOCK_BYTES" comment is currently unenforced.
- [ ] Confirm `block_sizes: {-1: 256}` is actually consumed (the ggml backend derives its own super-block scale and ignores `block_sizes`) — say in the PR body which consumer reads it, e.g. the export path.
- [ ] Confirm that landing the presets under `configs/ptq/presets/model/` intentionally adds `--qformat iq1_s`/`iq2_xs` to `hf_ptq`, given packed export only exists after #2447 merges.

No action needed:
- Shipped-recipe smoke tests (`tests/unit/recipe/test_loader.py`, `test_presets.py`) auto-discover the new files, so load/schema validation is covered.

hychiang-git:
Addressed in [`a2849c6f3`](https://github.com/NVIDIA/Model-Optimizer/commit/a2849c6f31f9088e4e13031b0ab601271b673672):

- Restored the full Apache-2.0 header in all six new recipe YAML files.
- Added a parameterized recipe-contract test that pins `backend`, `num_bits`, `block_sizes[-1]`, and `effective_bits` to the IQ implementation constants.
- Documented in the PR body that `block_sizes[-1]` is an export metadata/validation contract: #2447 reads it through `get_weight_block_size`, validates the native 256-value block size, and records it in checkpoint metadata. It does not drive the fake-quant scale search.
- Confirmed that exposing `--qformat iq1_s` and `--qformat iq2_xs` is intentional. This PR remains last in the merge order so #2447 provides packed-checkpoint export first.

Validation: 360 recipe tests passed; recipe validation, formatting, lint, license, and the remaining changed-file pre-commit checks passed.

cjluo-nv (commented):

Approving: every concern from the last round is resolved in `a2849c6f3`, and the remaining open items are one-line doc polish already recorded by other reviewers.

No action needed:
- :heavy_check_mark: Resolved since the last review: the four new YAMLs now carry the canonical `LICENSE_HEADER` verbatim; `test_iq_recipe_matches_packing_contract` pins `backend`, `num_bits`, `block_sizes[-1]` and `effective_bits` to `IQ*_BLOCK_SIZE`/`IQ*_EFFECTIVE_BITS`; the PR body now names the export consumer of `block_sizes[-1]` and confirms the `--qformat iq1_s`/`iq2_xs` exposure is intentional behind the stated merge order.
- Recipe count checks out — `general/ptq/` holds 28 YAMLs and `ptq.md` says 28.
- Minor and non-blocking: in `ptq.md`, the IQ rows sit under the `W4A16` weight-only heading and "packed checkpoint export is added separately" goes stale once #2447 lands ahead of this PR. Worth a one-line touch-up if you re-push.

Large PR: spans 5 directories (≥ 5). The review came back clean, so this is an LGTM — a human should take the final look and approve.

Testing: No test plan or testing section found in the PR description.

*Suggested test plan:*
- Run `tests/unit/recipe/test_presets.py` (especially the new `test_iq_recipe_matches_packing_contract` parametrization) and confirm the imported `IQ*_BLOCK_SIZE` / `IQ*_EFFECTIVE_BITS` constants exist and match the YAML values (1.5625 / 2.3125, block 256) — verify the test fails if a YAML value is edited.
- Load `general/ptq/iq1_s` and `iq2_xs` and validate the resolved `QuantizeConfig` against the schema: weight quantizer enabled with `backend: ggml`, input/output quantizers disabled, `algorithm: null` (no calibration), and no unused/ignored search or scale-search keys.
- Do a short end-to-end `mtq.quantize` smoke run on a tiny model with each recipe (no calib data) and confirm weights are fake-quantized, `TensorQuantizer.block_sizes[-1] == 256`, and outputs are finite/deterministic.
- Confirm the documented shape constraint: a linear whose last dim is not divisible by 256 raises a clear error (or is excluded), and one divisible by 256 quantizes successfully.
- Verify the `--qformat iq1_s` / `iq2_xs` choices are actually registered in `presets.QUANT_CFG_CHOICES` / `hf_ptq` and that `hf_ptq` runs at least to the quantize

modelopt_recipes/configs/numerics/iq1_s.yaml
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# IQ1_S weight quantizer using the built-in fixed-scale codebook search.
+
+# modelopt-schema: modelopt.torch.quantization.config.QuantizerAttributeConfig
+num_bits: iq1_s
+# Storage cost: 50 packed bytes * 8 / 256 weights. Keep in sync with IQ1_S_BLOCK_BYTES.
+effective_bits: 1.5625
```
cjluo-nv, line 23: why do we need to record this? Is it needed for autoquant?

hychiang-git: Yes. effective_bits is cost metadata for AutoQuant/compression estimation; it does not affect packing or numerical quantization. Since num_bits is the string iq1_s, the generic estimator cannot infer its storage cost. This records the actual 50 × 8 / 256 = 1.5625 bits per weight. I’ll clarify the YAML comment. https://github.com/NVIDIA/Model-Optimizer/blob/eaee22741ac66d0c6aed6a711e2da1d44245ca6d/modelopt/torch/quantization/algorithms.py#L184-L195

cjluo-nv (commented):

Comment: the new `ggml/common.py` cache change swaps a weakref for a strong reference to the weight view, which can pin full-precision storage in offloaded/meta-device flows, and it lands in a file this PR's stated scope excludes.

Needs action:
- [ ] Bound the cache lifetime in `modelopt/torch/quantization/ggml/common.py:41` — hold a weakref to `inputs._base` (the parameter) instead of the view, or clear `_quantizer_cache`, so an offloaded/meta-swapped weight is not kept alive. See inline comment.
- [ ] Move the `common.py` + `test_ggml_backend.py` cache fix into #2446 (which owns those files and already has the weakref flagged), or update the PR body — it still claims 8 recipe/doc files and "no codec/backend files".
- [ ] Reconcile the `tests/examples/hf_ptq/test_llm_ptq.py` comment with `algorithm: null` (no calibration forward), and confirm 18m of 2xH100 CI for two params is intended now that packing is cached.
- [ ] Fix the two doc nits in `modelopt_recipes/ptq.md`: the IQ rows sit under the `W4A16` weight-only heading, and the calibration column says "GGML IQ auto search" for a recipe with no calibration pass.

No action needed:
- :heavy_check_mark: Resolved since the last review: canonical Apache headers on the six YAMLs, `test_iq_recipe_matches_packing_contract` pinning `block_sizes`/`effective_bits`, and the stale "export added separately" wording.
- Recipes reuse the existing numerics/preset/`$import` layering — no new abstraction, so the design gate stays satisfied.

modelopt/torch/quantization/ggml/common.py
```diff
@@ -27,7 +26,19 @@
...
+    inside ``input_key`` cannot be recycled by a later allocation while this entry is live --
+    which is the hazard a weakref was meant to cover. It costs nothing in practice: the storage
+    belongs to the module's weight, which outlives the quantizer anyway.
+
+    A weakref cannot do that job here. TensorQuantizer hands the backend a fresh view of the
+    weight on every forward, so a weakref to it dies as soon as that forward returns, and an
+    identity check against it never matches again.
+    """
+
+    input_tensor: torch.Tensor
```
cjluo-nv: > _Bot comment._

Holding a strong reference fixes the ABA hazard, but the docstring's "it costs nothing in practice: the storage belongs to the module's weight, which outlives the quantizer" is not true in a few flows this repo supports:

- accelerate CPU/disk offload re-materialises the weight per forward and drops it afterwards — the cache would keep one full-precision copy alive per quantized module, i.e. the whole model resident, which is exactly what offloading avoids;
- layerwise export / `set_module_tensor_to_device(..., "meta")` frees the weight, but the cache keeps the old storage until that quantizer is called again (it is never cleared otherwise);
- if the block reshape has to copy (non-contiguous parameter), `inputs` is not a view at all and the cache pins a genuine duplicate of the weight.

A weakref to the *base* rather than the view gets you both properties: `weakref.ref(inputs._base if inputs._base is not None else inputs)`, then treat a dead ref as a miss. The base is the parameter, so it survives across forwards (fixing the bug this change targets) and dies when the weight is freed, which is what rules out data_ptr recycling. Alternatively keep the strong ref but clear `_quantizer_cache` on `_apply`/device moves, and soften the docstring to state the pinning cost explicitly.

tests/examples/hf_ptq/test_llm_ptq.py
```diff
@@ -76,6 +76,19 @@ def test_ptq_whisper(command):
         PTQCommand(quant="int8_weight_only", kv_cache_quant="none"),
         PTQCommand(quant="int4_awq", kv_cache_quant="none"),
         PTQCommand(quant="w4a8_awq_beta", kv_cache_quant="none"),
+        # GGML IQ weight-only, recipe-driven. These encoders require every weight's input
```
cjluo-nv, line 79: > _Bot comment._

Two things to reconcile here:

1. The comment says the pack runs "once on the first calibration forward" — but both IQ presets set `algorithm:` to null, so there is no calibration pass; the first pack happens on the first real forward.
2. 18m17s on 2xH100 for TinyLlama is a lot of example-suite time to add, and surprising now that the packed weight is cached per quantizer. Worth confirming the CUDA extension (`get_cuda_ext_ggml`) is actually available in this CI image rather than falling back to the torch `_encode_blocks` search — and if the cost is unavoidable, consider keeping only one of the two formats in the example matrix since they exercise the same code path.

modelopt_recipes/ptq.md
```diff
@@ -136,6 +138,11 @@ activations and tensor-core math are what deliver the throughput.
 - **`mxfp4_mlp_weight_only`** — MXFP4 weights on MLP/MoE layers only, BF16
   activations. Needs no calibration forward pass; the QAT starting point for the
   GPT-OSS family (see `examples/gpt-oss`).
+- **`iq1_s` / `iq2_xs`** — GGML-compatible IQ1_S or IQ2_XS weights on all linear
```
cjluo-nv: > _Bot comment._

These two entries land under `#### Weight-only schemes (W4A16 — activations stay BF16)`, but they are W1A16/W2A16 — either generalise the heading (e.g. "Weight-only schemes (activations stay BF16)") or give the IQ formats their own sub-section.

While you are here: "all linear layers" overstates the preset (`default_disabled_quantizers` keeps `lm_head`, routers, `conv1d` and the vision branch in BF16 — the recipe's own `description` says "eligible linear layers"), and the table cell on line 61-62 advertises "GGML IQ auto search (no calibration)" for a recipe with `algorithm:` null; `none (no calibration)` matches the `mxfp4_mlp_weight_only` row above.

modelopt/torch/quantization/ggml/common.py
cjluo-nv: Done in `f11eb630a` — took the `weakref.ref(inputs._base)` option.

Confirmed `_base` is the `Parameter` itself and that its id is stable across forwards, so the identity check now matches and the cache actually hits; the ref still clears when the weight is freed, which keeps the offload/meta cases you listed intact. Verified both halves directly: alive while the param lives, cleared after it is dropped.

The strong ref was also hiding how much this was costing. On TinyLlama IQ1_S `hf_ptq` (2xH100, same command either side):

| | packer calls | packing time | wall clock |
|---|---|---|---|
| before | 15400 | 499.0 s | 8m57s |
| after | 154 (one per weight) | 6.2 s | 3m21s |

modelopt_recipes/ptq.md
cjluo-nv: Both done in `871d4ee9f`, plus the two extras you flagged in the same comment:

- heading is now `#### Weight-only schemes (activations stay BF16)`
- the calibration column reads `none (no calibration)`, matching the `mxfp4_mlp_weight_only` row
- `all linears` → `eligible linears` in the table, and the prose bullet now names what `default_disabled_quantizers` holds back (`lm_head`, MoE routers, `conv1d`, vision branch)

modelopt_recipes/configs/numerics/iq1_s.yaml
cjluo-nv: Comment expanded in `871d4ee9f` to answer the "is it needed for autoquant?" part directly, since the arithmetic was the only thing it explained:

```yaml
# Cost metadata for AutoQuantize's compression estimate only; it drives no packing or
# numerics. num_bits is the string "iq1_s", so the generic estimator cannot derive the
# storage cost: 50 packed bytes * 8 / 256 weights. Keep in sync with IQ1_S_BLOCK_BYTES.
effective_bits: 1.5625
```

ChenhanYu (approved):

tests/examples/hf_ptq/test_llm_ptq.py
```diff
@@ -76,6 +76,19 @@ def test_ptq_whisper(command):
...
+        # dimension to be a multiple of 256; TinyLlama's 2048 and 5632 both are.
+        # Measured 18m17s for the pair on 2xH100, over the 300s tests/examples default. The
```
cjluo-nv: > _Bot comment._

This comment is inaccurate on two counts, and it was flagged last round.

1. Both IQ presets set `algorithm:` to null, so `hf_ptq` logs "Dynamic quantization. Calibration skipped." — the first pack happens on the first real forward (the sample `generate()`), not on "the first calibration forward".
2. "Measured 18m17s for the pair on 2xH100" predates the `_PackedWeightCache` fix in this same PR; the PR body now reports 3m21s wall clock per run (down from 8m57s). Either re-measure and update the number, or drop it — as written the file argues for a 1800s timeout on evidence the PR itself invalidates.

While updating: worth confirming the `get_cuda_ext_ggml` extension is actually available in this CI image (otherwise the torch `_encode_blocks` search runs), and whether both formats need to be in the example matrix when they exercise the same backend path.

cjluo-nv (commented):

The weakref-to-`inputs._base` cache fix is correct and well covered, but the one prior concern about `tests/examples/hf_ptq/test_llm_ptq.py` is still unaddressed and its comment now contradicts the PR body's own numbers.

Needs action:
- [ ] Fix the comment in `tests/examples/hf_ptq/test_llm_ptq.py:79-83`: these recipes set `algorithm: null`, so there is no calibration forward, and "18m17s for the pair" predates the packing cache (PR body now measures 3m21s per run). See inline comment.
- [ ] Confirm 2x1800s of 2xH100 example-suite time is still intended post-cache, or drop one of the two IQ params — both exercise the same backend path.

No action needed:
- :heavy_check_mark: Resolved since the last review: the cache now holds a weakref to `inputs._base` (offload/meta flows no longer pinned), the PR body documents the widened scope into `ggml/common.py`, and `ptq.md` got the generic weight-only heading plus the `none (no calibration)` column.
- `test_ggml_weight_is_packed_once_across_forwards` genuinely pins the fix: `block_sizes: {-1: 256}` makes `is_static_block_quant` true, so the quantizer really does hand the backend a fresh reshape view each forward.
- Recipes reuse the existing numerics → preset → `general/ptq` `$import` layering; no new abstraction, so the design gate stays satisfied.

tests/examples/hf_ptq/test_llm_ptq.py
cjluo-nv: Both points were right, and the second one turned out to be the more important of the two.

**1. The comment was wrong.** Corrected in `e707df765`. There is no calibration pass — the repeated packing came from the sample `generate()` calls, one decode step per weight.

**2. 18m was not intended, and not necessary.** Instrumenting both formats showed packing was no longer the cost at all once the cache worked — IQ2_XS packs *faster* than IQ1_S and was still 5x slower end to end:

| | pack calls | packing | wall clock |
|---|---|---|---|
| IQ1_S | 154 | 6.2 s | 3m21s |
| IQ2_XS | 154 | 1.6 s | 17m59s |

The extension is loaded in both cases (`Loaded extension modelopt_cuda_ext_ggml`), so it was not a fallback to the torch search.

The cause was one constant serving two loops with opposite characteristics. `_DEFAULT_BLOCK_CHUNK_SIZE` bounds the torch encode fallback, which holds the large codebook-search temporaries, runs once per weight, and is skipped entirely on the CUDA path; IQ2_XS sets it to 256 rather than IQ1_S`s 1024 because its search sweeps sixteen local scales per grid tile. The **decode** shared it — and the decode has tiny temporaries, runs on every forward, and is never cached, so a small chunk only multiplies kernel launches. Decoding a 2048x5632 weight: 91.4 ms at 256 blocks, 22.9 ms at 1024, 5.8 ms at 4096.

The decode now takes its own `_DEFAULT_DECODE_CHUNK_SIZE`, threaded through `fake_quantize_with_cache`; the encode bounds are untouched, so the ceiling from `5b465df27` stays where it was aimed.

| | before | after |
|---|---|---|
| IQ1_S | 259.65 s | 94.64 s |
| IQ2_XS | 1081.05 s | 102.77 s |

Both now fit the 300s `tests/examples` default, so I dropped the explicit timeouts rather than resizing them — the pair is 3m17s.

On trimming to one format: I would keep both. Together they now cost less than either did alone, and the 4x gap between them is exactly the kind of thing only running both would have caught.

cjluo-nv (approved):

cjluo-nv (commented):

The prior open concern is fixed and the new decode-chunk work checks out, but the PR has grown into two more codec files owned by #2446 while its scope paragraph still says otherwise.

Needs action:
- [ ] Update the PR body's "Scope" paragraph — it still says twelve files / only `ggml/common.py` plus two test files, while the diff is 15 files and now also changes `ggml/iq1_s.py` and `ggml/iq2_xs.py`.
- [ ] Confirm with the stack owner that the `_DEFAULT_DECODE_CHUNK_SIZE` + sign-parity changes belong here rather than in #2446 (which owns those codec files and merges first), or rebase them there.
- [ ] Consider allowing `decode_chunk_size` in `ggml/backend.py`'s `backend_extra_args` allowlist — `block_chunk_size` is tunable per quantizer, the new decode knob is not.

No action needed:
- :heavy_check_mark: Resolved since the last review: the `tests/examples/hf_ptq/test_llm_ptq.py` comment now states `algorithm: null` (no calibration) with re-measured 95s/103s runs, and the explicit 1800s timeouts are gone.
- I verified the XOR-fold parity in `dequantize_iq2_xs` matches the 7-bit loop it replaces, and `test_iq2_xs_dequantizes_the_implied_eighth_sign_bit` plus the chunk-invariance test pin both changes. No existing tests were weakened.

cjluo-nv:
Addressed the three items from the latest review in `9a6ce1652`, plus the PR body.

**Scope paragraph** — rewritten. It now says sixteen files and names all four `ggml/` files (`common.py`, `iq1_s.py`, `iq2_xs.py`, `backend.py`) rather than just `common.py`.

**Whether the decode-chunk and parity work belongs here or in #2446** — confirmed with the stack owner: keep it here. #2446 is already merged, and both bugs are only observable through the end-to-end test this PR adds, so moving the fixes would separate each one from the test that demonstrates it. The body now states this explicitly rather than leaving the widened scope unexplained.

**`decode_chunk_size` in the `backend_extra_args` allowlist** — good catch, done. It was the one chunk bound a caller could not reach:

```python
unknown_args = set(extra_args) - {"block_chunk_size", "decode_chunk_size"}
```

`test_ggml_backend_forwards_chunk_sizes` is now parametrized over each knob alone and both together.

---

Separately, on a question worth recording since the decode default changed: **`decode_chunk_size` cannot alter any quantization result.** It does not reach the encoder — `block_chunk_size` still drives `quantize()`, and the packed GGML bytes are what export writes. The decode loop has no cross-chunk state; each iteration reads `blocks[start:stop]` and writes `decoded[start:stop]`, with no accumulator or reduction spanning chunks. Verified on a 2048x5632 weight, both formats:

| check | IQ1_S | IQ2_XS |
|---|---|---|
| decode bitwise-equal across chunks 256..45056 | yes | yes |
| decode bitwise-equal across chunks 1..32 | yes | yes |
| fake-quant output, old decode default vs new | yes | yes |
| torch encode invariant to its own chunk | yes | yes |

`torch.equal`, not `allclose`. `test_ggml_decode_is_invariant_to_chunk_size` pins it in CI. The only thing that changes is transient memory, +24 MiB to +56 MiB on that weight, freed immediately.

The sign-parity XOR fold in the same commit *is* a real numerics change to the decoder, and is covered separately: exhaustive over all 128 sign indices, and mutation-tested against two deliberately broken folds.

Tests: 52 passed in `tests/unit/torch/quantization/ -k "ggml or iq1 or iq2"`, 12 in `tests/unit/recipe/test_presets.py`.

cjluo-nv (commented):

Approving: the three open items from the last round are all addressed in the current diff, and the new code checks out.

No action needed:
- :heavy_check_mark: Resolved since the last review: the PR body's Scope paragraph now names all four `ggml/` files and records the stack-owner decision to keep the codec fixes here, and `decode_chunk_size` is in `ggml/backend.py`'s `backend_extra_args` allowlist with a parametrized `test_ggml_backend_forwards_chunk_sizes`.
- Verified `decode_chunk_size` reaches `dequantize_iq*` through `fake_quantize_with_cache` and is still validated by `validate_block_chunk_size`; `test_ggml_decode_is_invariant_to_chunk_size` pins that it cannot change numerics.
- The renamed `test_ggml_backend_forwards_block_chunk_size` is a justified edit — it gained the new knob and both-knobs cases, no coverage lost.
- New YAMLs carry the canonical `LICENSE_HEADER` text verbatim; no third-party code added in this diff.
- Trivial: the Scope paragraph says "three test files" while the diff touches four — not worth a re-push on its own.

Complex PR: spans 7 directories (≥ 5); 1 existing test file modified or removed. Looping in a human for approval.

PR #2451, opened 2026-09-16
[6771663] Preserve ONNX API output types when wiring casts

ajrasane:
### What does this PR do?

Type of change: Bug fix

Preserves the public ONNX graph I/O types captured at the API boundary when `PrecisionConverter` wires output casts. Type inference can change the working graph's output declaration before conversion; consulting that mutated declaration caused the required cast back to the original public type to be discarded and metadata restoration to fail.

The converter now derives its I/O type map from the preserved boundary metadata and uses that map when deciding whether a cast should become a public graph output. A regression test covers an FP32 output whose working declaration is inferred as FP16, and the changelog records the corrected behavior.

### Usage

```python
# No API changes are required. Existing conversions now preserve the original
# public I/O declarations when keep_io_types=True.
converted = convert_to_f16(model, keep_io_types=True)
```

### Testing

- Ran `pytest tests/unit/onnx/autocast/test_precisionconverter.py` (186 passed).
- Ran `pytest tests/unit/onnx/autocast` (249 passed).
- Ran Ruff check and format validation on the changed Python files.
- Verified the original minimal end-to-end reproduction with Python 3.12 and TensorRT 10.16.1.11; conversion now completes without the output-metadata restoration error.
- Verified the full CLI quantization path proceeds through the formerly failing one-Q/DQ scheme and successfully benchmarks the generated TensorRT engine.

### Additional Information

Tracking: [6771663]

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed ONNX FP16 conversion to preserve public graph output types when type inference changes internal declarations.
  - Ensured output casts are inserted correctly when preserving input/output types is enabled, maintaining compatibility for downstream consumers.
  - Validated that converted models retain the expected output types and remain compliant with ONNX validation.

- **Tests**
  - Added regression coverage confirming preserved output types, correct cast insertion, and valid ONNX model generation.


cjluo-nv (approved):

Approving: a small, correctly scoped fix that sources `original_network_io` from the boundary metadata instead of the post-inference working declaration, with a regression test and changelog entry.

No action needed.

- Checked the interaction with `_add_cast`'s existing `must_preserve_io_type` guard and `_sanity_check`; both now read the same boundary map, and the fallback when `original_network_io_metadata is None` (i.e. `keep_io_types=False`) reproduces the previous behavior exactly.
- Cast-up/cast-down ordering for an output present in both lists still resolves to the boundary type, which is the intended outcome.

PR #2461, opened 2026-09-17
Fix three windows issues: extension-build timeout, 0xc000001d crash, and config I/O encoding

shengliangxu:
## What

Three independent Windows issues, unrelated to each other beyond all being Windows.

1. **A per-test timeout** — the `modelopt_round_and_pack_ext` MSVC build ran inside a test's own clock. CI only.
2. **An illegal-instruction crash** — `0xc000001d`, a bf16 GEMM faulting on Intel hosts. CI only, and not our bug.
3. **An encoding bug** — YAML config read with the locale codepage rather than UTF-8. This one is user-visible, and the only one with a changelog entry.

The first two made the `windows` job fail often and for different reasons, which is why they arrived together. The third was found while reading the same code and is deliberately scoped to the config paths, not swept across the tree — see section 3.

## 1. The extension build was inside a test's timeout

```
tests/unit/onnx/quantization/test_qdq_utils.py::...::test_column_major_transformation_graph_structure
  +++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Captured stdout ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
INFO - Loading extension modelopt_round_and_pack_ext...
.rendered.modelopt_round_and_pack_ext.cpp
...
setuptools/_distutils/compilers/C/msvc.py, line 520, in compile
subprocess.py, line 1590, in _wait -> _winapi.WaitForSingleObject
```

`modelopt/onnx/quantization/extensions.py` calls `cppimport.imp` at **module import time**, and that module is imported lazily from inside `quant_utils.round_and_pack`. The first test needing it pays a full MSVC compile inside its own per-test timeout. Which test pays depends on collection order — which is why the failure looks like it wanders rather than pinning to one test.

`pyproject` sets `timeout_func_only`, so the per-test clock covers the test call only. Importing the module from a session-scoped autouse fixture moves the build outside it. `tests/gpu_megatron/conftest.py` already does this for the quant CUDA extensions.

It **cannot** reuse that helper, which is the non-obvious part: `load_cpp_extension` skips every quant extension when CUDA is unavailable — the case on the CPU-only Windows runner — so `precompile()` would warm nothing there. `modelopt_round_and_pack_ext` uses a different loader (`cppimport`) and is not CUDA-gated, which is exactly why it is the one that builds on that runner.

**Verified: the `windows` job passes on this PR.** That was the only way to test it — my environment has neither `onnxruntime` nor `cppimport`.

## 2. The illegal-instruction crash: a bf16 GEMM, not our code

`0xc000001d` is `STATUS_ILLEGAL_INSTRUCTION` (`3221225501` as an exit code). bf16 `linear`/`matmul`
on CPU dispatch through **oneDNN**, which by default selects the highest instruction set the host
advertises — Intel **AMX** on the Emerald Rapids machines in the Actions fleet. On those hosts that
path executes an instruction that faults with `#UD`, killing the whole pytest process. This is a
torch/oneDNN Windows issue; nothing in ModelOpt causes it and no PR introduced it.

The fix is one environment variable on this job: `ONEDNN_MAX_CPU_ISA=AVX2`. AVX2 is the ceiling the
AMD runners already operate at, and those have never shown the crash — so it is the setting with
evidence behind it rather than the highest one that might work. It changes which kernel runs, not
what is tested.

**Evidence**

- Six crashes across **five unrelated branches** between 2026-07-04 and 2026-09-17. Not new, and not
  attributable to any change of ours.
- Crash sites: `test_peft_save_restore` (four times), `test_unet_save_restore`, and
  `test_fp8_export_rejects_unsupported_dtype_conversion[mixed-format]`. **Five of the six run a bf16
  forward on CPU**: `create_tiny_llama_dir` sets `dtype=torch.bfloat16`, and `mixed-format` is the
  only parametrization in its file built on a bf16 128×128 `Linear` rather than a 4×4 one. The sixth,
  the UNet test from July, is fp32 and is not explained by this mechanism.
- **It tracks the host, not the test.** Within one job it reproduces every time, which is what made it
  look deterministic; across jobs it follows the CPU. The crashing job drew an Intel Xeon 8573C
  (AVX-512 + AMX). The next drew an AMD EPYC 7763 (Zen 3: neither) and the entire suite passed —
  including the test that had just crashed three times consecutively.
- `ATEN_CPU_CAPABILITY=default` did **not** suppress it, which fits: it governs ATen's own kernels,
  while oneDNN JIT-generates its own from runtime detection. `ONEDNN_MAX_CPU_ISA` is the documented
  knob for that.
- Independently reported elsewhere with the same signature — a bf16 GEMM in the Windows CPU torch
  build faulting `0xC000001D` on some runner CPUs, intermittently, at a comparable rate.

**What is not established:** precisely which instruction faults, and why AMX is unusable on a machine
that advertises it — most likely XSAVE tile state the hypervisor never enabled. Answering that needs a
minidump from an Intel host, and no run since the fix has drawn one. It does not block the mitigation:
the cap is justified by the host correlation above, not by knowing the opcode.

**No diagnostics ship with this.** Earlier revisions carried a canary that ran the same bf16 `Linear`
uncapped to prove the cap was load-bearing, plus `procdump -i` as a postmortem debugger and a minidump
parser to name the faulting module. All of it is removed. The canary never returned non-zero — every
run drew an AMD host — so there was no evidence it could detect anything, and an `exit=0` on an Intel
host would not have distinguished "this host is fine" from "the canary does not reproduce it". It
plausibly does not: the six real crashes all occurred deep inside full suite runs, not in a bare
forward in a fresh process. The rest is unreachable for a simpler reason — with the cap in place the
suite does not crash, so no dump is ever produced. If this recurs, the diagnostics and what each one
found are in this branch's history.

**Corrections to earlier revisions of this PR.** Two diagnostic steps claimed more than they did, and
are recorded here because they are gone from the diff and a reader would not otherwise see them: the
step named "Record CPU and torch dispatch capability" ran before the nox venv
existed, so it never queried torch at all; and the WER `LocalDumps` route, then a `procdump` wrapper,
both produced no dump — the wrapper attached to the nox parent while the crash was in the pytest
child. An `ATEN_CPU_CAPABILITY` sweep added at one point ran the whole suite in both arms rather than
the single test it named, because nox's `unit` session hardcodes `tests/unit` and drops posargs.

## 3. YAML config I/O in the shipped package

`modelopt/recipe/loader.py` reads recipe YAML inside a **user's** process, on Windows, under whatever locale that machine has — a UTF-8 recipe then fails to decode under a cp1252 locale with ModelOpt nowhere near a test run.

Ten call sites, all YAML config: the recipe loader, two ONNX autotune state files, two transformers config readers, the distill config, the puzzletron profile. YAML configs are the files most likely to carry non-ASCII — comments, model names, paths. `modelopt/torch/fastgen/loader.py` already used this form, so the convention predates the change.

**Deliberately not** the ~590 other encoding-less calls across tests, examples, plugins and tooling. An earlier revision of this PR did sweep them all, along with a preview-gated ruff rule and a custom AST pre-commit hook; that was reverted in 555cec1c5 — a large permanent tax for no user-visible gain. The history is left intact rather than squashed.

An earlier revision also ran the windows job under `PYTHONUTF8=1`. That is gone too, and it is worth saying why, because it looks like a free safety net. It never fixed an observed failure — there is no `UnicodeDecodeError` in any windows log, and main passes 3478 tests without it. Worse, it hides the defect this section exists to prevent: 139 encoding-less reads remain in `modelopt/`, and with CI in UTF-8 mode none of them can ever fail there, so the only place that could catch a missing `encoding=` in shipped code was blind to it. The cost of removing it is that a future test fixture carrying a non-ASCII byte will break the windows job — noise for files no user executes, and the price of CI being able to see the shipped-code case at all.

## Also here: the windows job no longer gates the merge

`unit-pr-required-check` keeps `windows` in `needs`, so the job still runs and its result is still
visible on every PR — but it is no longer part of the failure condition, so a windows failure does
not block a merge.

This was **requested by @kevalmorabia97** in review on this PR, not decided here:

> Windows tests are too flaky, not just these ones. What if we just exclude windows from the check
> here so it can still fail but allow to merge PR?
>
> [...] we mainly care about onnx features on windows so its not worth fixing each flaky torch test
> for windows but having them still run helps with basic sanity check of torch features in windows

Worth being explicit that this is **not** justified by the two fixes above. Those address the
extension-build timeout and the `0xc000001d` crash; the rationale for ungating is the broader
flakiness of the windows torch surface, which those two fixes do not claim to have exhausted. If the
view is that the gate should stay now that these two causes are fixed, that is a reasonable position
— it is a question for the CI owners rather than something this PR should settle quietly, and I am
happy to split the hunk into its own PR.


.github/workflows/unit_tests.yml
kevalmorabia97: Windows tests are too flaky, not just these ones. What if we just exclude windows from the check here so it can still faill but allow to merge PR?

shengliangxu: We can also do that, we have QA on windows right?

kevalmorabia97: Correct. Plus we mainly care about onnx features on windows so its not worth fixing each flaky torch test for windows but having them still run helps with basic sanity check of torch features in windows

shengliangxu: yeah, we can do that

shengliangxu: removed windows from required.

.github/scripts/name_faulting_module.py
```diff
@@ -0,0 +1,59 @@
...
+            continue
+        for m in mods:
+            base = getattr(m, "baseaddress", 0)
+            size = getattr(m, "size", 0)
+            if base <= addr < base + size:
+                print(f"  >>> FAULTING MODULE: {getattr(m, 'name', '?')}  (+0x{addr - base:x})")
+                break
+        else:
+            print(f"  >>> address 0x{addr:x} is in NO loaded module "
+                  "-- a jump into non-code memory, i.e. corruption, not a missing opcode")
```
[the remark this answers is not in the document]
shengliangxu: Moot now — `name_faulting_module.py` is deleted, along with the rest of the crash diagnostics, in f3563c9.

Recording the point anyway, because it was right and it would have bitten: "in no loaded module therefore corruption" was a false dichotomy, and JIT-generated code was exactly the case it would have misread. The root cause turned out to be oneDNN, which JIT-generates its kernels at runtime from CPU detection. So a dump from a faulting host would very plausibly have landed on an address in no loaded module, and the script would have confidently reported memory corruption for what is really a JIT code buffer — sending the next person down the wrong path with a definite-sounding message. Had the diagnostics stayed, this would have been worth fixing rather than skipping.

.github/workflows/unit_tests.yml
```diff
@@ -91,8 +91,103 @@ jobs:
...
+          Remove-Item Env:\DNNL_MAX_CPU_ISA -ErrorAction SilentlyContinue
+          $env:ONEDNN_VERBOSE = "1"
+          # The first version of this canary was a bare `a @ b` and printed no onednn_verbose
+          # line at all, meaning it never created a oneDNN primitive -- so a clean exit from it
+          # proved nothing. name_faulting_module.py's sibling runs the real operation instead.
+          & $py .github/scripts/bf16_canary.py 2>&1 |
+            Select-String -Pattern "onednn_verbose.*info|bf16 linear" | ForEach-Object { $_.Line }
+          $code = $LASTEXITCODE
+          Write-Host "uncapped-bf16-gemm exit=$code"
+          if ($code -eq 3221225501) { Write-Host "=> this host faults uncapped; the ISA cap is load-bearing here" }
```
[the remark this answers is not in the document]
shengliangxu: Moot — the step is deleted in f3563c9, along with the rest of the crash diagnostics.

On the substance: agreed that comparing `$code` to the decimal 3221225501 is fragile given `$LASTEXITCODE` is a signed Int32, and a mask would have been the right form. It happened not to misfire here — the canary never once returned non-zero — but "never returned non-zero" is precisely why the step is gone, so the comparison had no chance to be exercised either way.

tests/unit/conftest.py
```diff
@@ -25,3 +27,28 @@
...
+    ``tests/gpu_megatron/conftest.py``, which prebuilds the quant CUDA extensions for the same
+    reason -- but it cannot reuse that helper: ``load_cpp_extension`` skips every quant extension
+    when CUDA is unavailable, which is exactly the case on the CPU-only Windows runner, so
+    ``precompile()`` would warm nothing here.
+
+    Best-effort. The extension is an optimisation with a Python fallback -- ``extensions.py``
+    already swallows its own build failures -- so a failure to prebuild must not fail the session.
+    """
+    with contextlib.suppress(Exception):
+        import modelopt.onnx.quantization.extensions  # noqa: F401
```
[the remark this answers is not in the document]
shengliangxu: Already addressed, before this comment landed — the fixture lives in `tests/unit/onnx/conftest.py`, not `tests/unit/conftest.py`, so it is scoped to the ONNX tests and a focused run elsewhere under `tests/unit` never touches the cppimport cache. The Python fallback in `extensions.py` is untouched, and the build still happens in session setup so it stays outside the per-test timeout, which was the point of the fixture.

.github/workflows/unit_tests.yml
kevalmorabia97, line 1: is rest of the changes in this file for debugging or also to be merged?

kevalmorabia97 (approved):

.github/workflows/unit_tests.yml
shengliangxu: debugging code will be deleted, the reset of the files fix 3 common issues:

1.  one is encoding fix to utf-8, which is OS independent issue but our linux boxes always set to utf-8 system wide, but windows boxes may be not, therefore surface on windows
2. an illegal instruction failure, it's not on us, but on a third party library, we just limit the choice of the library build. Other debugging code has been removed.
3. timeout on compiling the onnx extension

.github/workflows/unit_tests.yml
```diff
@@ -201,9 +223,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Required unit tests did not succeed
+        # windows is deliberately absent from this condition. It still runs and is still visible,
+        # but it does not gate the merge: the job is flaky for reasons unrelated to the change
+        # under review -- an MSVC extension build racing the per-test timeout, and an
+        # 0xc000001d illegal instruction that depends on which CPU the runner draws. What we
+        # actually care about on that platform is the ONNX surface, so blocking every PR on
+        # unrelated torch flakiness costs more than it catches.
         if: >-
```
cjluo-nv, line 232: > _Bot comment._

Removing `needs.windows.result != 'success'` makes the Windows job advisory for every future PR, which is a broader policy change than the three fixes this PR describes — and it isn't mentioned in the PR body.

It also reads as inconsistent with the rest of the PR: the extension-build timeout is fixed by the new conftest prebuild, the 0xc000001d crash is mitigated by `ONEDNN_MAX_CPU_ISA=AVX2`, and the body states the windows job passes here. If both causes are addressed, the gate should be able to stay. If residual flakiness is expected, a narrower option (skip/quarantine the specific torch CPU tests on Windows, keep ONNX gating) preserves the Windows ONNX coverage the comment says you actually care about.

Please either drop this hunk or split it into its own PR with CI-codeowner sign-off.

cjluo-nv (commented):

Requesting changes mainly because the PR silently drops `windows` from the merge-gating required check — a CI policy change the description never mentions, and one that contradicts the PR's own claim that both flakiness causes are now fixed.

Needs action:
- [x] Justify or revert the removal of `needs.windows.result` from `unit-pr-required-check` in `.github/workflows/unit_tests.yml` — if the timeout and 0xc000001d causes are fixed and the job passes, the gate should stay; see inline comment.
- [x] Describe the gating change in the PR body/CHANGELOG so a CI codeowner can sign off on it explicitly rather than finding it in the diff.
- [x] Add a regression test for the `encoding="utf-8"` fix (e.g. load a recipe YAML with non-ASCII under `-X warn_default_encoding` / a patched locale in `tests/unit/`), since with `windows` non-gating nothing else can catch a future regression.

No action needed:
- The session-scoped prebuild fixture in `tests/unit/onnx/conftest.py` correctly mirrors `tests/gpu_megatron/conftest.py`, and `timeout_func_only = true` in `pyproject.toml` confirms setup is off the per-test clock; the PR body explains why `ext.precompile()` can't be reused.
- The `ONEDNN_MAX_CPU_ISA`/`DNNL_MAX_CPU_ISA` cap is CI-only and well documented in the workflow comment.

.github/workflows/unit_tests.yml
shengliangxu: You are right that it was undocumented, and I have fixed that — the PR body now has a section on it. On whether the hunk should exist at all, there is context that is easy to miss because it lives in a separate thread on this same file:

**@kevalmorabia97 asked for it in review on this PR**, and it was agreed there rather than decided unilaterally:

> Windows tests are too flaky, not just these ones. What if we just exclude windows from the check here so it can still fail but allow to merge PR?
>
> Correct. Plus we mainly care about onnx features on windows so its not worth fixing each flaky torch test for windows but having them still run helps with basic sanity check of torch features in windows

That does not make your objection wrong, and one part of your reasoning I want to answer directly rather than lean on the agreement. You say that if both causes are addressed the gate should be able to stay. The two fixes here do **not** support that: they address the extension-build timeout and the `0xc000001d` crash specifically, and neither claims to have exhausted the flakiness of the windows torch surface. Keval's stated rationale was explicitly the broader set — "not just these ones". So "both causes are fixed" is true of the two this PR names, not of the platform.

Where I think you are straightforwardly right: it is a policy change, it is broader than the three fixes in the title, and it was not in the body. Reviewers should not have to reconstruct that from a thread.

On the narrower alternative — quarantining specific torch CPU tests on Windows while keeping ONNX gating — that is a better end state than ungating wholesale, and it is the one I would argue for if we were starting fresh. What makes it awkward as a change here is that it needs a list of which tests to quarantine, and the evidence says that list is not stable: the `0xc000001d` crash alone has landed on `test_peft_save_restore`, `test_unet_save_restore`, and `test_fp8_export_rejects_unsupported_dtype_conversion[mixed-format]` across five branches since July, following the runner CPU rather than the test. A quarantine list built from today's failures would be stale by the next fleet change.

I am not going to resolve a disagreement between two reviewers by picking one, so: if you and @kevalmorabia97 land on dropping the hunk or splitting it with CI-codeowner sign-off, say so and I will do it immediately — splitting is maybe ten minutes of work. The three fixes stand on their own and do not depend on it.

shengliangxu: intentional change

PR #2464, opened 2026-09-17
docs(eval): align GDPVal/TB2.1/SWE-bench/MRCR skills with upstream configs

cjluo-nv:
### What does this PR do?

Type of change: documentation (agent skill) + a bug fix in the GDPVal SIF builder script

Realigns the `evaluation` skill's four benchmark recipes, config templates and the
GDPVal SIF builder with the canonical configs in
`nvidia-eval-factory-benchmarking` (`dl/JoC/competitive_evaluation/…`) at `main` =
`92b5474` (2026-09-17). The skill had drifted on all four benchmarks.

**GDPVal — upstream is now `GDPVal-AA-V2`, a different measurement from what the skill described**

- **Metric key moved.** Upstream certifies on the final-stage-scoped
  `comparison/stage_1/normalized_elo`. The flat `comparison/normalized_elo` is
  emitted *only* when that stage is exactly complete, so a few lost rollouts blank it
  while the real number sits in the stage key. The skill told you to read the flat key
  and treat its absence as a failed run.
- **Reference ladder is 12 models** (ELO 274–1468; anchors read off the AA board
  2026-09-09/11), with `human_gold` and `gptoss_20b` deliberately disabled. Anchors are
  held fixed by the Bradley-Terry fit, so a stale anchor shifts every candidate score.
- **Two-stage fit** (45 tasks vs all refs → 220 vs the 4 nearest) with
  `partial_completion` gates and `retry_inprocess`. Without those gates a single
  unresolved rollout forfeits stage 2 after stage 1 has already been paid for.
- **Judge is a sampled 3-member panel** with per-member `create_params_overrides`; the
  docs said "Gemini 3.1 Pro" as *the* judge.
- Gym pin → `df3e201d`, SIF → `python-3.13.gdpval.gym-80e4fc.sif`, and the pin now
  requires Python ≥ 3.13.14. Added a section on why those three move as one unit: wrong
  Python fails loudly, wrong SIF degrades deliverables silently.

**`scripts/gdpval-sif.sh` was broken by that pin (the one functional fix here)**

The v2 `gdpval.def` declares a `%files` section (419-package pin list, apt closure,
arm64 exclusions, sandbox verifier). Apptainer resolves `%files` sources against the
*build's CWD*, so the previous "download one file and build it" path could not work.
The script now stages the def plus every source named in its `%files` block into a
temp dir and builds from inside it — the list is parsed from the def, not hardcoded,
so a def that adds a file keeps working.

**MRCR**

- Canonical `parallelism` is **512**, deliberately *above* total server capacity so 1M-token
  requests queue at the endpoint. That invalidates the skill's
  `--max-num-seqs = ceil(parallelism / num_instances / DP)` rule, which was a coincidence
  at 256 and is wrong at 512.
- `limit_samples` now reaches the gym via a gated `++limit`, so MRCR *does* have a
  sample-limited canary — the exact opposite of GDPVal. Added the gate to the template.
- 128K is now its own upstream benchmark on the condensed
  `benchmark:`/`prepare_args`/`run_args` schema.

**Terminal-Bench 2.1 / SWE-bench Verified**

- `http_pairs_dump` is **first** in the interceptor chain for both (docs said last for TB2.1).
- Sandbox region comes from `HARBOR_ECS_REGION`, but the ECR coupling differs: TB2.1's repo
  *name* tracks the region (`harbor-<region>`), SWE-bench's is pinned to
  `us-west-2/harbor-swebench` regardless. Called out explicitly since it is easy to get backwards.
- Interceptor lists replace wholesale on merge — a leaf adding one entry must restate the chain.
- Dropped a stale "MiniMax-M2.7 uses concurrency 20" note; that config is gone, and the
  `clu` track that exists now runs `repeats: 3` / `concurrency: 10` and is **not** AA-comparable.

### Usage

No API change. Regenerating a config from the skill now yields the aligned values, e.g.:

```yaml
# recipes/examples/gym/example_gdpval.yaml
evaluation:
  env_vars:
    GDPVAL_CONTAINER_PATH: lit:/gdpval/sif/python-3.13.gdpval.gym-80e4fc.sif
    NEMO_GYM_MAX_ROLLOUT_ATTEMPTS: lit:3
  tasks:
    - name: nemo_gym
      nemo_evaluator_config:
        config:
          params:
            extra:
              nemo_gym:
                allow_limit_samples: false                 # limit_samples is inert on this path
                install_on_the_fly:
                  commit: df3e201d942f6397def1ec8c10037d29115fcbbd
```

### Testing

- `pytest plugins/modelopt/skills/evaluation/tests/test_nel_gdpval.py` — passes (launcher pin unchanged).
- `pre-commit run --files <changed>` — passes (yamlfmt, markdownlint, license, symlink sync).
- `gdpval-sif.sh` smoke-tested end to end against the real `df3e201d` def with a stub
  `apptainer`: verified it fetches the def, parses the `%files` block, stages all four
  sources, invokes the build with CWD inside the staging dir, publishes atomically, and
  cleans up. Also exercised `--check` (present / wrong-name) and the reuse path.
- All three example configs re-validated as parseable YAML, and the edited keys
  (`allow_limit_samples`, gym pin, SIF path, MRCR `parallelism`/`limit_samples` gate,
  TB2.1 sandbox + MLflow tags) re-read back from the parsed docs.
- Upstream values were read from a fresh clone of the eval-factory repo at `92b5474`; the
  `gdpval.def` `%files` claim was confirmed against `raw.githubusercontent.com` at the pinned SHA.

Not run: an actual GDPVal/MRCR/TB/SWE evaluation. These are skill docs and templates; the
numbers quoted are upstream's own recorded sign-off values, not new measurements.

### Additional Information

Upstream source of truth for each benchmark, for reviewers who want to diff:
`configs/benchmarks/{gdpval-aa-v2,gym/gdpval,mrcr,mrcr-128k,terminal-bench-2.1,swe-bench-verified}/`
plus `configs/shared/nel_next_containers.yaml`, in
`dl/JoC/competitive_evaluation/nvidia-eval-factory-benchmarking`.

One judgement call worth a reviewer's eye: the upstream SWE-bench `manifest.yaml` lists its
metric key as `mean/reward`, but the certificates for the harbor path record `pass@1`
(e.g. `0.516`). I kept `pass@1` as what to report for a nel-next run and documented
`mean/reward` as the Gym/direct-path key. If you know which one the leaderboard actually
reads, that note should be tightened.

The companion change to the internal ModelOpt repo's own eval configs is tracked separately.

## Summary by CodeRabbit

- **New Features**
  - Added support for GDPVal-AA v2, including comparison-mode evaluation, multi-stage scoring, judge configuration, and version-matched runtime artifacts.
  - Added region-aware Harbor deployment settings and expanded credentials/configuration options.
  - Added improved model-traffic capture, observability, replay, and MLflow metadata options.

- **Documentation**
  - Clarified GDPVal, MRCR, SWE-bench, and Terminal-Bench benchmark configurations, metrics, manifests, concurrency, and sample-limit behavior.
  - Documented separate 1M and 128K MRCR benchmarks and updated canonical runtime and deployment requirements.


shengliangxu (approved):

cjluo-nv:
Superseded by #2479, which rebases onto current main (post-#2470, which dropped GDPVal from this skill) and narrows the scope to Terminal-Bench 2.1, SWE-bench Verified and MRCR.

PR #2466, opened 2026-09-18
Fix: Release per-layer expert weights in layerwise export under offload

Fridah-nv:
### What does this PR do?

Type of change: Bug fix

`_export_fused_experts` splits a fused MoE experts module into per-expert holder submodules and attaches them to the live model:

```python
proj = nn.Module();  proj.weight = wrapper.weight   # packed U8
expert.add_module(proj_name, proj)
module.add_module(str(idx), expert)
_delete_fused_moe_source_attrs(module)
```

Those holders are plain `nn.Module`s built *inside* the weight-access window, so they carry no accelerate `_hf_hook`. `weight_access_and_writeback_context` closes by iterating the modules it collected at entry and calling `hook.post_forward()` through each one's own offload hook — the holders satisfy neither condition, so nothing ever returns them to meta.

Whole-model export never notices: one pass, write the state dict, exit. Layerwise export runs the same pass once per decoder layer, so every finished layer's packed experts stay resident. The two paths diverge because resident weights make `_delete_fused_moe_source_attrs` free real BF16 tensors 4x larger than their packed replacements, while offloaded ones were on meta and cost nothing — so the packed tensors are pure new residency.

Measured through unmodified `examples/hf_ptq/hf_ptq.py` on a Qwen3.5-MoE-shaped model (10 layers, 64 experts), printing `torch.cuda.memory_allocated()` after each exported layer:

| placement | per layer | over 9 layers |
| --- | --- | --- |
| offload | +0.052 GiB | 0.579 → 1.047 GiB |
| resident | -0.135 GiB | falls, as designed |

+0.052 GiB is exactly one layer's quantized experts: packed U8 100.7M/2 = 0.047 GiB plus FP8 block scales 100.7M/16 = 0.006 GiB. What survives each layer's offload window is 768 tensors / 54.0 MiB, all of it `mlp.experts.{i}.{gate,up,down}_proj.{weight,weight_scale,weight_scale_2,input_scale}`.

At real scale this is fatal rather than wasteful. Quantizing Qwen/Qwen3.8-2.4T-A95B (92 layers, 512 experts) leaks 12.9 GB packed + 1.6 GB scales = 14.5 GB per layer, so 92 layers need 1.33 TB that no budget on a 283 GB card or 952 GB host can absorb. The run died of CUDA OOM at layer 16/92 with `--max_gpu_memory_gb 240`, and at `--max_gpu_memory_gb 30` leaked the same 15 GB/layer onto the host instead.

**Fix:** `_export_fused_experts` records the holders it attached, and `release_exported_fused_experts()` drops them. `LayerwiseExporter.export_layer` calls it right after `save_file` — from that point the shard on disk is the artifact, and `finalize()` indexes shards by reading them back, never the layer. The release is kept separate from the split because the whole-model path still needs the holders: it reads them out of `model.state_dict()` at the very end.

### Usage

No API change. Existing layerwise export under offload just stops growing:

```bash
    --pyt_ckpt_path Qwen/Qwen3.8-2.4T-A95B \
    --qformat nvfp4 \
    --export_path /path/to/export \
    --max_gpu_memory_gb 240
```

### Testing

- `cuda_alloc` over 10 offloaded layers: 0.526 → 0.518 GiB (-0.008), was +0.468
- Exported checkpoint byte-identical to the unfixed run: 7841 tensors, 0 mismatches, max abs diff 0.0; `hf_quant_config.json` / `config.json` / index JSON identical
- `tests/gpu/torch/export/test_layerwise_export.py` — 31 passed
- `tests/gpu/torch/export/test_offload_export.py` — 4 passed
- `tests/unit/torch/export/test_offload_export.py` — 38 passed
- Full Qwen3.8-2.4T-A95B PTQ then completed all 92 layers: peak GPU 117 GB of 283, peak host RSS 71 GB of 952, flat across 50 consecutive layers at 23-26 s/layer

- Did you write any new necessary tests?: ❌ — the leak is a memory-residency property of the offload path; the existing layerwise and offload export suites cover the behavior that had to stay unchanged (byte-identical checkpoints), and they pass.
- Did you update [Changelog](https://github.com/NVIDIA/Model-Optimizer/blob/main/CHANGELOG.rst)?: N/A — `layerwise.export_dir` is new in the unreleased 0.48.0, so this bug was introduced and fixed within the same cycle.
- Did you get Claude approval on this PR?: ❌ — not yet.

### Additional Information

Draft: opening for early visibility while the checklist above is finished.

## Summary by CodeRabbit

* **Performance**
  * Reduced memory usage during layerwise and streaming export by releasing temporary exported tensors after persistence.
  * Cleans up temporary exported expert data after use, helping limit memory growth across layers.
* **Bug Fixes**
  * Preserves unrelated module buffers while clearing export-generated quantization data and expert modules after successful export operations.
  * Ensures buffers created during export are also cleared during cleanup without affecting export results when errors occur.


modelopt/torch/export/moe_utils.py
```diff
@@ -230,6 +230,39 @@ def _export_fused_experts(
     # 4. Remove fused params and quantizer lists — replaced by per-expert submodules
     _delete_fused_moe_source_attrs(module)
 
+    # Record the holders so a caller that has already persisted them can drop them;
+    # see release_exported_fused_experts for why only some callers may.
+    module._modelopt_exported_expert_children = tuple(str(idx) for idx in range(n))
+
+
+def release_exported_fused_experts(root: nn.Module) -> int:
```
cjluo-nv: > _Bot comment._

`_export_transformers_checkpoint_streaming` already solves this exact problem inline (the block after `_reconstruct_fused_moe_linear(layer_module)` that nulls CUDA buffers and, for modules without `_hf_hook`, CUDA params, with a long comment naming `_export_fused_experts` as the cause). This PR adds a second, differently-shaped mechanism for the same leak in a sibling export path.

Please either factor one helper both paths call — the marker-based release here is more surgical and could replace the `not hasattr(sub_mod, "_hf_hook")` heuristic there — or say in the PR body why the two should stay separate. Note the streaming version also drops CUDA *buffers* on hooked modules (`post_forward` runs with `offload_buffers=False`); if that category applies to the layerwise path too, this fix is incomplete for the non-expert linears' `weight_scale`/`input_scale`.

Also: no test covers the new public function. A CPU unit test asserting the holders are gone, the marker is cleared, the count is returned, and a second call is a no-op would be cheap.

modelopt/torch/export/moe_utils.py
```diff
@@ -230,6 +230,39 @@ def _export_fused_experts(
...
+    cannot put them back on meta when it closes: it iterates the modules present at entry
+    and re-offloads through each one's own hook, and these modules existed under neither
+    condition. A whole-model export never notices -- it packs every module once, writes the
+    state dict and exits. A layerwise export runs the same pass once per decoder layer, so
+    without this every finished layer's packed experts stay resident and the run grows by
+    one layer's quantized weights until it runs out of memory.
+
+    Returns the number of expert containers released.
+    """
+    released = 0
```
cjluo-nv: > _Bot comment._

`root.modules()` is a generator that iterates `self._modules.items()` after yielding `self`, so deleting children of the just-yielded module happens to be safe today — but it depends on that ordering inside `named_modules()`. Iterate `list(root.modules())` to make it robust.

Minor: the marker stores `tuple(str(idx) for idx in range(n))`, which is fully determined by `n`; storing `n` (or a plain `True`) would do. And in the whole-model path nothing ever releases, so `_modelopt_exported_expert_children` stays on the module for the model's lifetime — worth clearing there too if it isn't meant to persist.

modelopt/torch/export/layerwise_export.py
```diff
@@ -320,6 +321,11 @@ def export_layer(
 
         save_file(tensors, str(self._export_dir / layer_shard_name(layer_idx)))
 
+        # The shard on disk is the artifact from here on -- finalize() indexes the shards by
+        # reading them back, never the layer. Holding the packed experts as well costs one
+        # layer's quantized weights per layer, which no memory budget survives at 92 layers.
+        release_exported_fused_experts(layer_module)
```
cjluo-nv: > _Bot comment._

Good placement: this runs inside `persistent_materialization` (from `layerwise_calibrate`) and before `ckpt.save`, which with an exporter present has `save_layer_state=False` and so never reads `layer_module.state_dict()`. Worth stating that second constraint in the comment — a future change making the checkpointer read the layer after export would silently drop the expert tensors.

cjluo-nv (commented):

The fix looks correct for the leak it targets, but it adds a second cleanup mechanism next to the one `unified_export_hf_streaming.py` already has, and ships no test for the new public helper.

Needs action:
- [ ] Reconcile with the existing cleanup in `unified_export_hf_streaming.py` (the per-layer loop that nulls CUDA buffers and params on hook-less modules) — either share one helper or explain in the PR body why two mechanisms are right.
- [ ] Add a CPU unit test for `release_exported_fused_experts` (holders dropped, marker cleared, return count, second call a no-op); `tests/unit/torch/quantization/plugins/test_fused_experts.py` already builds a fused-experts module.
- [ ] Confirm the layerwise path does not also leak the scale buffers registered on hooked submodules (`weight_scale`, `input_scale`) — `AlignDevicesHook.post_forward` runs with `offload_buffers=False`, which is exactly category 1 in the streaming path's comment.
- [ ] Iterate `list(root.modules())` in `release_exported_fused_experts`; deleting children mid-generator only works by `named_modules()` internals.

cjluo-nv (commented):

Prior review items are all addressed; the one thing left is a behaviour change the unification introduced in the streaming path that the author should confirm.

Needs action:
- [ ] Confirm dropping the old `not hasattr(sub_mod, "_hf_hook")` parameter sweep is intentional: the new helper releases only *new buffers on pre-existing modules* plus marked expert holders, so the 3-D `weight` Parameter `_reconstruct_fused_moe_linear` restacks onto `_QuantMoELinear` (`plugins/huggingface.py`) now stays resident per layer where `unified_export_hf_streaming.py` used to free it.
- [ ] Clear `_modelopt_exported_expert_children` on the whole-model export paths (`unified_export_hf.py`), or say why leaving the private marker on the user's live model is fine — raised last round, still open.

No action needed:
- :heavy_check_mark: Resolved since the last review: the duplicate cleanup in `unified_export_hf_streaming.py` (now one shared helper), the missing buffer category (`weight_scale`/`input_scale` via the before/after diff), `list(root.modules())`, and the missing CPU tests (`test_offload_export.py`, `test_fused_experts.py`).
- Test edit in `test_fused_experts.py` is justified: assertions moved inside the new block, coverage intact plus a holders-released assertion.

cjluo-nv (commented):

Nudge: the fix and its tests look right, but the dangling expert marker on the non-layerwise export paths is still open and the PR is still marked draft.

Needs action:
- [ ] Clear `_modelopt_exported_expert_children` on the whole-model (`unified_export_hf.py`) and FSDP2 (`collect_export_tensors`) paths, or say why leaving the private marker on a live model is fine — raised twice, still unanswered.
- [ ] Decide on the FSDP2 packing loop in `unified_export_hf_streaming.py`: wrap it in `_release_exported_tensors` or add a one-line comment that per-unit holder residency is deliberately out of scope.
- [ ] Confirm the checklist is complete and take the PR out of draft.

No action needed:
- :heavy_check_mark: Resolved since the last review: the duplicated streaming cleanup (now one shared helper), the missing scale-buffer category, `list(root.modules())`, the missing CPU tests, the accidental public symbol, and the re-derived expert names.
- I verified dropping the old `not hasattr(sub_mod, "_hf_hook")` param sweep is safe: `_QuantMoELinear._setup` refuses accelerate-offloaded weights, so the offload streaming path never holds a hookless restacked 3-D parameter.
- Test edit in `test_fused_experts.py` is justified: assertions moved inside the new block, coverage intact plus a holders-released check.

Fridah-nv:
> ## Claude review — 1 IMPORTANT, 3 SUGGESTIONs
> Full scope: all 5 changed files (`modelopt/torch/export/{layerwise_export,moe_utils,unified_export_hf_streaming}.py` plus the 2 test files; +109/-55), and the call chain needed to judge them — `_dispatch_export_handler`, `_export_quantized_weight`, `_reconstruct_fused_moe_linear`, `_export_fused_experts`, `LayerwiseExporter.export_layer`/`_collect`/`finalize`, and `_packed_units_for_rank`.
> 
> ### What this revision gets right
> The redesign since `7431523` is a real improvement, and it closes the IMPORTANT finding from the last round:
> 
> * **Category 1 is now covered.** `_reconstruct_fused_moe_linear` registers `weight_scale`/`weight_scale_2`/`input_scale` on the _hooked_ `_QuantMoELinear` wrapper, which `post_forward` never meta-ifies (`offload_buffers=False`). The entry-time buffer-name snapshot catches those — the ~1.6 GB/layer half of the leak the previous revision missed.
> * **The name diff is strictly safer than the device sweep it replaces.** The old streaming code nulled _every_ CUDA buffer, which would have wiped `TensorQuantizer._amax` (`unified_export_hf.py:657` moves it to fp32 in place) and the `k_bmm_quantizer._amax`/`v_bmm_quantizer._amax` that `postprocess_state_dict` maps to `k_scale`/`v_scale`. Those are pre-existing names, so the diff leaves them alone. The new `rotary_emb_inv_freq` assertion pins exactly this.
> * **Dropping the hookless-parameter sweep is sound, not a narrowing.** `AlignDevicesHook.post_forward` enumerates the module's _live_ tensors rather than a map recorded at attach time, so the freshly-stacked 3-D `module.weight` from `_reconstruct_fused_moe_linear` is meta-ified even though it did not exist when the hook was installed. `recurse=place_submodules` is False, so the hookless holders are what remains — and deleting them by marker covers that exactly. I traced `proj.weight = wrapper.weight` (a real `nn.Parameter`, so genuinely resident) and confirmed `wrapper` itself is never attached to the tree.
> * **`list(root.modules())` fixes the mid-iteration mutation** flagged by CodeRabbit and the prior round. Stale entries revisited after deletion simply miss the marker and `continue`.
> * **Release ordering and liveness check out.** In streaming, `with (enable_weight_access_and_writeback(...), release_exported_tensors(...))` enters left-to-right so the snapshot is taken post-materialization, and exits in reverse so the release lands before `post_forward` and before `torch.cuda.empty_cache()`. In layerwise, `_collect` does `.detach().contiguous().cpu()`, so `tensors` never pins the GPU allocation and the release frees immediately.
> * **Nothing downstream observes the released state.** `finalize()` skips decoder-owned module ids and layer-name prefixes, `_write_index` re-reads shards from disk, and `state_dict()` skips `None` buffers. Consistent with your byte-identical checkpoint result.
> * No mode registration, config schema, or `modelopt_state` surface is touched; the CHANGELOG skip is correct per CONTRIBUTING (introduced and fixed in the same unreleased cycle).
> 
> ### Findings
> **CRITICAL: 0 · IMPORTANT: 1 · SUGGESTION: 3**
> 
> #### [IMPORTANT Performance] The FSDP2 packing loop has the same holder leak and is the one packing loop left uncovered
> `_packed_units_for_rank` (`unified_export_hf_streaming.py:591-607`) runs the identical sequence this PR just wrapped twice over:
> 
> ```python
> with enable_weight_access_and_writeback(module, model, names, writeback=True):
>     if not is_owner:
>         continue
>     for sub_name, sub_module in module.named_modules():
>         _dispatch_export_handler(full_name, sub_module, ctx)
>     _reconstruct_fused_moe_linear(module)
>     for key, tensor in module.state_dict().items():
>         owned.append((full_key, tensor.detach().contiguous().cpu()))
> ```
> 
> `writeback=True` redistributes **`module`'s own parameters**. It does not remove child modules that `add_module` attached during the window, so the per-expert holders — `proj.weight` packed U8 plus the `weight_scale`/`weight_scale_2`/`input_scale` buffers — stay resident on GPU after the unit is packed, and accumulate across every unit in `get_export_units(model)`. Same arithmetic as your table, against sharded GPU memory instead of an offload budget. The in-window comment ("packing rebinds the parameter, so the restore copies the untouched gathered data back and drops the packed tensor") is accurate for `module`'s own parameters and simply does not reach the new children.
> 
> This is **pre-existing**, not a regression, so scoping it out is a legitimate call. I raise it as IMPORTANT because the helper now exists, the fix is one line in a file this PR already edits, and as it stands the PR makes the streaming writer look uniformly fixed when one of its two packing paths is not:
> 
> ```python
> with (
>     enable_weight_access_and_writeback(module, model, names, writeback=True),
>     release_exported_tensors(module),
> ):
> ```
> 
> Note the interaction with `if not is_owner: continue` — non-owners never run the handlers, so the context manager is a no-op snapshot for them and the collective is unaffected. If you would rather not touch the FSDP2 path in a bug-fix PR, a one-line comment at that loop pointing at the gap would keep the next reader from assuming it is covered.
> 
> #### Suggestions (non-blocking, posted inline)
> * **`moe_utils.py:249-259`** — `before is None: continue` silently skips modules created inside the window (empty today only because `_export_fused_experts` is the sole creator and its holders are deleted separately); `buffers_before.get(name, set())` is strictly safer. Separately, `yield` sits outside a `try`/`finally`, which now also covers `save_file` on the resumable layerwise path.
> * **`moe_utils.py:241`** — `release_exported_tensors` becomes public API via `from .moe_utils import *` (`export/__init__.py:20`, no `__all__` in this module). Underscore it, or add an explicit `__all__`.
> * **`moe_utils.py:237`** — the marker is never cleared on the whole-model and FSDP2 paths, leaving a dangling private attribute that asserts a pending release which will never happen; and setting it after `_delete_fused_moe_source_attrs` leaves a window where attached holders are not yet trackable.
> 
> ### Testing note
> `pytest` could not be executed in this review environment, so the suites you list are unverified here and this review is static. The new `test_release_exported_tensors_drops_what_the_offload_window_leaves` is well aimed: it asserts `offload_buffers is False`, demonstrates the leak _inside_ the window, and pins both the release and the non-destructiveness. With the restructured `test_fused_experts.py` case (holders dropped on block exit), the prior round's request for a unit test on the new helper is satisfied.
> 
> ### Risk assessment
> **Low.** The diagnosis is accurate and unusually well evidenced, the mechanism is both more precise and less destructive than the code it replaces, and it cannot alter the exported artifact — every tensor is copied out before the release, which your byte-identical check confirms. The unification the last round asked for happened: one helper, two call sites, the duplicated streaming block deleted. The remaining gap is coverage of the third packing loop, not correctness of what is here.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

@sugunav14 Please take a look at the FSDP part and see if it makes sense and deserve efforts to fix it (buffer memory are small in most cases..)

cjluo-nv (commented):

Nudge: the marker concern is resolved and cleanup looks sound, but the previously raised FSDP2 scope decision remains pending.

Needs action:
- [ ] :speech_balloon: Author requested FSDP review from @sugunav14 — record the decision for `collect_export_tensors`: add tested cleanup, or explicitly document that per-unit holder residency is deferred.
- [ ] Update the PR body's marker-based description and “no new tests” checklist to match the snapshot-based helper and added coverage; confirm readiness and remove the draft designation when ready.

No action needed:
- :heavy_check_mark: Resolved since the last review: dangling markers are eliminated entirely; shared cleanup, scale-buffer release, snapshot traversal, private API, and CPU coverage remain in place.
- The `test_fused_experts.py` edit is justified: existing assertions remain inside the export window, with holder-removal assertions added afterward.
- Review was static; test execution was not independently verified.

modelopt/torch/export/moe_utils.py
```diff
@@ -231,6 +232,28 @@ def _export_fused_experts(
     _delete_fused_moe_source_attrs(module)
 
 
+@contextmanager
+def _release_exported_tensors(root: nn.Module):
```
shengliangxu: nit: we seems targeting an moe specific cleanup, but the code is very general purpose.

shengliangxu (approved):

modelopt/torch/export/moe_utils.py
Fridah-nv: Nice catch! Thank you

PR #2470, opened 2026-09-18
Drop GDPVal support from the evaluation skill

cjluo-nv:
### What does this PR do?

Type of change: deprecation (agent skill)

Removes GDPVal support from the `evaluation` agent skill. Its task recipe, example
config and Apptainer SIF build helper are deleted.

GDPVal was not cleanly separable, so this is not just a delete:

- **MRCR depended on GDPVal's infrastructure.** `scripts/nel-gdpval.sh` was a
  generic pinned-0.2.6 `nel` launcher that only happened to be GDPVal-named, and
  MRCR ran through it; `references/gym-gdpval.md` documented the gym bootstrap
  machinery (prepare/reap, `install_on_the_fly` pin↔container coupling, the trust
  env vars) that both examples share. So the shared parts are kept and renamed
  rather than dropped: `scripts/nel-gym.sh` and `references/gym.md`. The launcher
  pin itself is unchanged (0.2.6), as are its env-override semantics.
- **GDPVal is an AA-suite member**, so the "AA rule" in `SKILL.md` and
  `references/quantization-benchmarks.md` had to change. An "AA" / "Artificial
  Analysis" request now generates the `aa/` tasks as one multi-task config and
  nothing else. Both files now state that the resulting set omits GDPVal and is
  therefore not directly comparable to a published AA Index — report per-task
  scores rather than an aggregate. `SKILL.md` also tells the agent to say GDPVal is
  unsupported rather than reconstruct a config from an older copy of the skill.

Everything GDPVal-specific is gone from `references/gym.md`: the SIF sandbox and
its silent-unsandboxed-exec failure mode, the 3-member judge panel, rubric vs.
comparison scoring, the deliverables/MLflow `*cache*` trap as a GDPVal concern, and
Stirrup-agent deploy sizing. `recipes/env.example` loses `GDPVAL_SIF_DIR`,
`GDPVAL_MAX_TURNS` and `TAVILY_API_KEY`, and gains a documented
`NEMO_EVALUATOR_TRUST_UNLISTED_TASKS` (required by every gym task, previously only
mentioned in prose).

One correction carried along: the old reference said the bootstrap pins
`ray==2.49.2`, but `example_mrcr.yaml` actually pins to whatever ray version the
image already carries. `references/gym.md` now describes what the template does.

### Usage

Not an API change. The skill-facing surface that moved:

```text
scripts/nel-gdpval.sh      -> scripts/nel-gym.sh      (NEL_GDPVAL_* -> NEL_GYM_*)
[...]

PR #2479, opened 2026-09-18
docs(eval): align Terminal-Bench 2.1 / SWE-bench / MRCR with upstream configs

cjluo-nv:
### What does this PR do?

Aligns the `evaluation` skill's three remaining upstream-tracked benchmarks —
Terminal-Bench 2.1, SWE-bench Verified and MRCR — with the canonical configs in
`nvidia-eval-factory-benchmarking` (`dl/JoC/competitive_evaluation/…`) at `main` = `92b5474`.

GDPVal is deliberately out of scope: it was removed from this skill in #2470. This
replaces #2464, which predated that removal and covered GDPVal too.

**Terminal-Bench 2.1 / SWE-bench Verified**

- **`http_pairs_dump` is first in the interceptor chain for both.** The TB2.1 recipe said
  it was last, and the shared `nel-next.md` said the position was per-benchmark. It isn't
  any more — the model fragment upstream puts it first for both.
- **Sandbox region moved to `HARBOR_ECS_REGION`, and the two benchmarks don't behave the
  same.** TB2.1's ECR repo *name* tracks the region
  (`463701203462.dkr.ecr.<region>.amazonaws.com/harbor-<region>`), so moving the region
  moves both halves. SWE-bench's is pinned to `us-west-2/harbor-swebench` regardless of
  its sandbox region (default `us-east-2`). Easy to get backwards, so it's called out on
  both recipes and in the shared reference.
- **Interceptor lists replace wholesale on merge.** The reviewed SWE-bench leaf restates
  the model fragment's entire chain to add two entries. Adding `system_message` without
  re-listing the rest silently drops them — a scoring change that looks like nothing.
- **`proxy.model_traffic.capture_request_body` belongs on the service leaf**, never a
  shared block: a shared `services:` entry is alias-only with no `type` and fails the
  service discriminator for every other model's composed config.
- Canonical MLflow tags now carry `task_name: <slug>` and
  `nemo-evaluator-next-version` (from `configs/shared/nel_next_containers.yaml`), plus
  `log_config_params: true`.
- Dropped a stale "per-model configs override concurrency (MiniMax-M2.7 uses 20)" note —
  that config no longer exists. The separate customer-launch track that does exist runs
  `repeats: 3` / `concurrency: 10` and is **not** AA-comparable, which is worth saying
  explicitly so its numbers aren't borrowed.

**MRCR**

- **`parallelism` is 512, and it is not a server cap.** Upstream raised it from 256 so
  1M-token requests queue at the endpoint rather than leaving replicas idle. The skill's
  `--max-num-seqs = ceil(parallelism / num_instances / DP)` rule happened to give the
  right answer at 256 (`ceil(256/4/2) = 32`, which is what the golden served) and gives
  the wrong one at 512. Replaced with sizing from what a replica's KV can hold.
- **`limit_samples` now reaches the gym**, via a gated
  `{% if config.params.limit_samples is not none %}++limit=…{% endif %}`. So MRCR does
  have a sample-limited canary. The prepare pass still runs in full, so it isn't cheap,
  and `pass@1` from a limited run isn't a score — both stated.
- Adds the canonical `++observability_enabled=true` / `++model_call_capture_dir`
  overrides and a canary check for the resulting `artifacts/model_calls`.
- **128K is now its own upstream benchmark** (`configs/benchmarks/mrcr-128k/`) with its
  own manifest key and the condensed `benchmark:`/`prepare_args`/`run_args` schema
  against the `gym_runtime` container. Noted, with the warning that switching schema
  means switching container — the `pre_cmd` and inline `command:` blocks don't apply there.
- The golden's recorded `parallelism: 256` is kept as provenance for the reference
  `pass@1 = 26.91`, flagged as a throughput difference rather than a scoring one.

### Usage

No API change. Regenerating a config from the skill now yields the aligned values:

```yaml
# recipes/examples/gym/example_mrcr.yaml
evaluation:
  nemo_evaluator_config:
    config:
      params:
        parallelism: 512     # above server capacity on purpose; do NOT derive --max-num-seqs from it
        limit_samples:       # set N for a canary -> forwarded to the gym as ++limit

# recipes/examples/example_eval_next.yaml
benchmarks:
  - sandbox:
      region: ${HARBOR_ECS_REGION:-us-east-1}   # TB2.1's ECR repo name tracks this; SWE-bench's does not
```

### Testing

- `pytest plugins/modelopt/skills/evaluation/tests/test_nel_gym.py` — passes.
- `pre-commit run --files <changed>` — passes (yamlfmt, markdownlint, license, symlink sync).
- Both edited example configs re-validated as parseable YAML, with the changed keys
  (`parallelism`, `limit_samples`, the `++limit` jinja gate, TB2.1 sandbox block and
  MLflow tags) read back from the parsed documents.
- Upstream values read from a fresh clone of the eval-factory repo at `92b5474`, against
  `configs/benchmarks/{terminal-bench-2.1,swe-bench-verified,mrcr,mrcr-128k}/` plus
  `configs/shared/nel_next_containers.yaml` and the reviewed per-model leaves under
  `reviewed_configs/vpr/2026/`.

Not run: an actual TB2.1 / SWE-bench / MRCR evaluation. These are skill docs and
templates; the numbers quoted are upstream's own recorded sign-off values, not new
measurements.

### Additional Information

Replaces #2464 (closed — it predated the GDPVal removal in #2470).

Companion internal change: the `eval-config` skill's Step 3b/3d rows in
Model-Optimizer-Internal, tracked separately.

One judgement call worth a reviewer's eye: the upstream SWE-bench `manifest.yaml` lists
its metric key as `mean/reward`, but the certificates for the harbor path record `pass@1`
(e.g. `0.516`). I kept `pass@1` as what to report for a nel-next run and documented
`mean/reward` as the Gym/direct-path key, so the two are never silently compared. If you
know which one the leaderboard actually reads, that note should be tightened.

## Summary by CodeRabbit

* **Documentation**
  * Added `HARBOR_ECS_REGION` guidance for sandbox and AWS region configuration, including benchmark-specific repository behavior.
  * Clarified Terminal-Bench, SWE-bench, and MRCR evaluation setup, including routing, concurrency, parallelism, manifests, and certified metrics.
  * Expanded guidance for request-body capture, model-traffic observability, interceptor ordering, logging, and MLflow metadata.
  * Added support for optional MRCR canary sample limits and documented distinct 1M and 128K benchmark configurations.


meenchen (commented):

Comment: the new MRCR behavior still conflicts with instructions in the skill entry point and shared reference.

Needs action:
- [ ] Reconcile MRCR canary guidance in `SKILL.md` and `references/gym.md` with the newly forwarded `limit_samples`; see inline comment.
- [ ] Scope the 1M serving instructions in `SKILL.md` to that variant and document the 128K serving adjustment; see inline comment.
- [ ] Add regression coverage for rendering `example_mrcr.yaml` with null and numeric `limit_samples`, checking that `++limit` is respectively omitted or emitted.

No action needed:
- The design reuses existing templates and launcher paths. The PR explains why the upstream condensed schema needs additional bootstrap and why nel-next cannot use OmegaConf environment syntax.
- The reported end-to-end campaigns are useful validation; no existing tests were weakened.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -140,26 +153,17 @@ NVIDIA-internal: `modelopttools:eval-config` Step 3d names a working image.
...
-of the YAML — easier than re-pasting the whole folded scalar through `-o`:
-
-```yaml
-                collect_rollout_params: >-
-                  ...
-                  ++limit=5          # canary only — remove for the scored run
-```
-
-Then watch the first ~30 min of the real run:
+Set `limit_samples` (the template gates `++limit` on it, as canonical does) — in the YAML,
```
meenchen, line 166: > _Bot comment._

The new template forwarding makes this guidance correct, but the mandatory reading still says the opposite: `SKILL.md` MRCR step 6 says **`limit_samples` is inert on the gym path**, and `references/gym.md`'s launcher-pin procedure says it does not reach Gym and instructs users to edit `collect_rollout_params`. The PR body acknowledges this, but it is directly part of the behavior changed here rather than an unrelated follow-up. Update both to describe the MRCR forwarding and link to this canary procedure, so an agent does not bypass the newly supported limiter.

plugins/modelopt/skills/evaluation/SKILL.md
```diff
@@ -64,6 +64,7 @@ for an "AA" request. If the user asks for MRCR:
 2. **Pick the variant first** (`config_n3_1m` / `config_n3_128k` / `config`) — it
    sets the context cap, dataset *and* metric prefix; the three are not
    comparable; set it in **both** `data_prep_params` and `collect_rollout_params`.
+   Take the largest variant within the checkpoint's trained context — see the recipe.
```
meenchen, line 67: > _Bot comment._

The new largest-supported-variant rule permits 128K, but MRCR step 5 below still unconditionally prescribes `--max-model-len 1100000` plus `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1`. Following the entry point literally therefore retains the 1M serving envelope even after selecting 128K for a 262K checkpoint. Qualify step 5 as 1M-only and document the corresponding 128K serving setup (within trained context, with answer headroom and without the extrapolation override), rather than only changing the two dataset selectors.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
```diff
@@ -56,11 +59,21 @@ Sets the context cap, the dataset, **and the metric prefix**. The golden uses 1M
 | `benchmarks/mrcr/config_n3_128k.yaml` | 131,072 (gated NVIDIA) | `mrcr_n3_128k_benchmark_simple_agent` | 1 |
 | `benchmarks/mrcr/config.yaml` | none (`o200k_base`) | `mrcr_benchmark_simple_agent` | 4 |
 
+**Pick the largest variant within the checkpoint's trained context**
+(`max_position_embeddings`, under `text_config` on multimodal configs). The template's 1M
+follows the golden; on a 262K model, 1M (via `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1`) measures
+extrapolation, not long-context quality — use 128K.
```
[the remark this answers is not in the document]
cjluo-nv: Addressed in d49cc63b: the 128K option now changes the serving setup as well. `mrcr.md` (variant rule + serving section) and the template comment say to serve 128K at the checkpoint's trained context, with room above 131,072 for the uncapped answer, and to drop `VLLM_ALLOW_LONG_MAX_MODEL_LEN` instead of keeping `1100000`.

plugins/modelopt/skills/evaluation/recipes/tasks/gym/mrcr.md
cjluo-nv: Addressed in d49cc63b: SKILL.md step 6 now says the MRCR template forwards `limit_samples` as `++limit`, pointing to the recipe's Canary section. `references/gym.md` now says the launcher-level `limit_samples` reaches the gym only when the template gates it (as `example_mrcr.yaml` does), and otherwise to add `++limit=N`.

plugins/modelopt/skills/evaluation/SKILL.md
cjluo-nv: Addressed in d49cc63b: SKILL.md step 5 now applies `--max-model-len 1100000` + `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1` to 1M only. For 128K it says to serve at the trained context, leave answer headroom above 131,072 and skip the override. The recipe's serving section and the template comment say the same.

cjluo-nv:
Re the review checklist item on regression coverage: added `plugins/modelopt/skills/evaluation/tests/test_example_mrcr.py` in d49cc63b. It checks that `limit_samples` defaults to null and that `collect_rollout_params` renders no `++limit` for null and `++limit=5` for 5. It uses only the standard library so it runs in the CI skill-test job (which installs no YAML/Jinja), and adds a real Jinja render when `jinja2` is available. 6/6 skill tests pass; removing the gate from the template makes 3 of them fail.

The CodeRabbit outside-diff note on `SKILL.md:84` (stale "inert" canary guidance) is covered by the same commit.

plugins/modelopt/skills/evaluation/recipes/examples/example_eval_next.yaml
```diff
@@ -110,4 +122,12 @@ output:
...
       description: ???               # '<model> | T=1.0 top_p=0.95 | <benchmark> (timeout_strategy=…) | r8'
       # model/checkpoint_path/benchmark drive dashboard attribution (engine logs only a generic metric key); temperature/top_p mirror generation above.
-      tags: {framework: vllm, model: "???", checkpoint_path: "???", benchmark: "???", temperature: '1.0', top_p: '0.95'}
+      tags:
+        framework: vllm
+        model: "???"
+        checkpoint_path: "???"
+        benchmark: "???"
+        task_name: "terminal-bench-2.1"          # upstream benchmark slug (SWE-bench: swebench-verified)
+        nemo-evaluator-next-version: "0.5.0.1"   # eval_image version, so a score traces to its image
```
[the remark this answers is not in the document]
cjluo-nv: Upstream `configs/benchmarks/{terminal-bench-2.1,swe-bench-verified}/bench.yaml` set the tag key `nemo-evaluator-next-version: ${._containers.nemo_evaluator_next_version}`. The underscore form is the container-pin variable in `configs/shared/nel_next_containers.yaml`, not the tag. Keeping the hyphenated key so runs match canonical dashboard attribution; `references/nel-next.md` now names both.

meenchen (commented):

Comment: prior concerns are resolved, but the new inline comment in MRCR’s folded deployment command disables subsequent serving flags.

Needs action:
- [ ] Move the variant note above `deployment.command` in `recipes/examples/gym/example_mrcr.yaml`; see inline comment.
- [ ] Add a regression in `evaluation/tests/test_example_mrcr.py` verifying the folded deployment command preserves the KV-cache, reasoning-parser, and batching arguments when interpreted by the shell.

No action needed:
- :heavy_check_mark: Resolved since the last review: canary guidance, variant-specific serving guidance, limit-forwarding coverage, and the MLflow tag concern (the reply distinguishes the upstream tag from its container-pin variable).
- The design retains existing templates and launcher paths; the PR explains the explicit-versus-condensed schema and environment-syntax choices.
- No existing tests were weakened. The new test header matches `LICENSE_HEADER`.

plugins/modelopt/skills/evaluation/recipes/examples/gym/example_mrcr.yaml
```diff
@@ -105,7 +105,7 @@ deployment:
     --port ${deployment.port}
     --tensor-parallel-size 1
     --data-parallel-size 1
-    --max-model-len 1100000
+    --max-model-len 1100000          # 1M only; 128K: trained context, drop VLLM_ALLOW_LONG_MAX_MODEL_LEN
```
meenchen: > _Bot comment._

`deployment.command` is a YAML folded scalar (`>-`), so this `#` is command text, not a YAML comment. All following lines fold onto the same shell line: the shell comments out `--kv-cache-dtype fp8`, `--reasoning-parser`, prefix-caching and batching settings. This changes the intended serving configuration and can prevent the long-context deployment from fitting. Move this variant note outside the scalar, above `command:`, and add a regression checking that the folded command retains its trailing flags under shell parsing.

cjluo-nv: Good catch, fixed in 6727bef9. The 1M-only note is now in the comment block above `command:`, with a warning that `#` inside the folded scalar becomes shell text. `tests/test_example_mrcr.py` adds `test_deployment_command_keeps_all_flags_under_shell_parsing`, which folds the command, checks that `shlex.split(..., comments=True)` equals the comment-free split, and checks that `--kv-cache-dtype`, `--reasoning-parser`, `--enable-prefix-caching`, `--enable-chunked-prefill` and `--max-num-batched-tokens` are present. Re-adding the inline comment makes it fail.

meenchen (commented):

Approve: prior concerns are resolved, including the folded-command bug and regression coverage; canary guidance, variant-specific serving, and MLflow tag documentation are consistent.

No action needed.

Large PR: spans 7 directories (≥ 5). The review came back clean, so this is an LGTM — a human should take the final look and approve.

chadvoegele (approved):

hychiang-git (approved):

